PR #96, opened 2021-01-26
SIMS#65 - Student selects login

andrewsignori-aot:
- Created an endpoint in the API (api/config) to retrieve the authentication config (Keycloak) and other configurations in the future.
- Added buttons for Login/Register to the login screen that are gonna point to the same method for now.


abschwenker (commented):
Please connect to ticket. I believe you can do manually or though keywords in PR desc

sources/packages/web/src/services/AuthService.ts
```diff
@@ -33,6 +33,7 @@ export default async function(
   }
   keycloak.onTokenExpired = () => {
     console.log("onTokenExpired");
```
fwpushan, line 35: Do we need this logging?

sources/packages/api/src/main.ts
```diff
@@ -6,6 +6,11 @@ async function bootstrap() {
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
   app.setGlobalPrefix('api')
   const port = process.env.PORT || 3000;
+
+  if(process.env.NODE_ENV !== 'production'){
```
fwpushan, line 10: I think CROS should enable for all env. Because our api will be exposed through api gateway, so any third party app can access them future.

fwpushan (commented):
All looks good, just couple of comments.

sources/packages/api/src/main.ts
andrewsignori-aot: I would prefer to keep it as locked as possible for now and enable it when the time comes.
All discussions about API Gateway seem to be too initial and I am still not sure if this specific API would be exposed.

fwpushan (approved):

sources/packages/api/src/main.ts
```diff
@@ -6,6 +6,11 @@ async function bootstrap() {
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
   app.setGlobalPrefix('api')
   const port = process.env.PORT || 3000;
```
2011sandeepraj, line 8: Is there any merit in externalizing the default port ?

sources/packages/web/src/services/AuthService.ts
```diff
@@ -32,7 +32,7 @@ export default async function(
     console.log(`KC - init excp : ${excp}`);
   }
   keycloak.onTokenExpired = () => {
-    console.log("onTokenExpired");
+    store.dispatch("auth/logout");
```
2011sandeepraj, line 35: we dont have a story for auth/logout as yet so better to remove this one?

sources/packages/web/src/services/http/common/HttpBaseClient.ts
```diff
@@ -0,0 +1,9 @@
+import HttpClient from './HttpClient';
+
+export default abstract class HttpBaseClient
+{
+  protected apiClient = HttpClient;
+  handleRequestError(e:any){
+    console.log(e);
```
2011sandeepraj, line 7: may be we should add a technical debt for logging framework instead of console.logging

sources/packages/web/src/services/http/common/HttpClient.ts
```diff
@@ -0,0 +1,8 @@
+import axios from "axios";
+
+const apiUrl = process.env.NODE_ENV === 'production' ?
+    "/api" : "http://localhost:3000/api";
```
2011sandeepraj, line 4: localhost??

sources/packages/web/src/views/Login.vue
```diff
@@ -7,15 +7,22 @@
...
+          below to start your sign in/sign up.
         </h4>
       </div>
     </template>
     <template #footer>
       <Button
-        label="Login using BCSC"
+        label="Login"
+        icon="pi pi-check"
+        class="p-mr-2"
```
2011sandeepraj, line 19: p-mr-2, mr is for medium sized devices right? should we worry about medium sized devices right now?.

2011sandeepraj (commented):
Some comments

sources/packages/api/src/main.ts
andrewsignori-aot: @2011sandeepraj, this is not part of the changes in this PR. Maybe we can start a conversation about it outside this review.

sources/packages/web/src/services/AuthService.ts
andrewsignori-aot: The second "Technical task" in the story mentions it and we talked about it during our "dev planning" and that is why this line was added.
After the implementation of the story that you are working on, this will be no needed anymore because the entire "auth/logout" will be removed.

sources/packages/web/src/services/http/common/HttpBaseClient.ts
andrewsignori-aot: Yes, we can. I am gonna create one. The idea here was just to introduce an idea for a base class to support HTTP API calls since this was the first story executing an API call.

sources/packages/web/src/services/http/common/HttpClient.ts
andrewsignori-aot: If we are running the application outside our local env, the "/api" relative address will be used, otherwise, we will be using the localhost for our local development, right?

sources/packages/web/src/views/Login.vue
andrewsignori-aot: **mr** means "margin-right" here. It is adding spacing between the buttons as described in PrimeVue doc below:
https://www.primefaces.org/primevue/showcase/#/spacing

The below ones are the ones that you probably would be thinking about:

- sm: small screens e.g. phones
- md: medium screens e.g. tablets
- lg: large screens e.g. notebooks
- xl: larger screens .e.g monitors

sources/packages/web/src/services/AuthService.ts
2011sandeepraj: So do you want me to remove it with the code i am checkin in
?

andrewsignori-aot: Yes, if you did not update your PR yet. Otherwise, I will do it once I merge your changes.

2011sandeepraj: k please take care of it

2011sandeepraj (approved):
All good. Just take care of removing auth/login reference in one of the files.
LGTM.

PR #109, opened 2021-01-27
Renaming env variable from KEYCLOAK_URL to KEYCLOAK_AUTH_URL.

andrewsignori-aot:
It causes an issue in the previous deployment because api-deploy.yaml was using different names for the variable name and variable value.


2011sandeepraj (approved):
LGTM

Issue #78, opened 2021-01-18
Student Saves Mandatory Profile Information


mzmichellesmith:
**As a** *(Student)* **I want** *(I want to Save my profile information)* **So That** *(my account is created with SABC)*

**Additional context**
- This task completes the steps in the previous User Stories ( Authenticate, Provide Contact Info and Provide SIN)
- Once a SABC account is created, then the student can log in anytime (using their Services Card login)  to view their profile information.

**Acceptance Criteria**
- [x] Given (A Student, When (I have provided my Contact Information AND my SIN), Then (I will be prompted to Save my profile)
- [x] Given (A Student), When (I have saved my profile information), Then (an SABC account is created)

**Technical Tasks**
- [ ] Enable the API authentication (create new client on keyclock)
- [x] Design database for user/student (consider a team discussions for this initial effort)
- [ ] Create the API to create/update (POST/PUT or PATCH) the student
- [x] Create tests for data models
- [x] Create data model test strategy
- [ ] Create tests for api

Issue #68, opened 2021-01-08
Student Provides SABC Profile SIN Information on account set up


mzmichellesmith:
As a Student, I want to complete my account profile by providing the Mandatory SIN information required to set up my account so that I can have access to apply for SABC Financial Aid Programs

**Additional Context**
- BCSC will provide the following verified information. First Name, Last Name, Middle Name, Date of Birth, Email, and Gender 
- A Student will need to also provide the Mandatory Contact Information (Address and Phone Number) to complete Account (another user story)

**Acceptance Criteria**
- [x] Given (a Student), has provided their first successful authentication with BCSC), then (the student account will prompt the student to provide the following mandatory SIN information required to create an account:
- [x] Given (a Student), (provides their SIN), then (it must consist of 9 digits, and any spaces that are entered are to be ignored). 
- [x] Given (a Student), (provides their SIN), then (must pass the Luhn algorithm check. Failure to pass will result in an error message.)
- [x] Given (a Student), (provides the SIN, then the student will be prompted to save the information and the student account will be created.

Issue #21, opened 2020-12-11
Student Provides SABC Profile Contact Information on account set up


mzmichellesmith:
As a Student, I want to complete my account profile by providing the Mandatory Contact Information required to set up my account so that I can have access to apply for SABC Financial Aid Programs

**Additional Context**
- BCSC will provide the following verified information. Given Name, Last Name, Date of Birth, Email, and Gender 
- The system must support mononymous names. This means a person known by just one name, rather than a first name and surname.
- In order to complete a Student Account, the Student will also need to provide their Social Insurance Number (SIN) (another user story) 
- The system must support International Address'

**Acceptance Criteria**
- [ ] Given (a Student), has provided their first successful authentication with BCSC), then (the student account will prompt the student to provide the following mandatory Contact information required to create an account: Address and Phone number
- [ ] Given (a Student), (provides their address), then (it will consist of the following free text fields: address line 1, address line 2, City, Province/State, Country, and Postal/zip code.) 
- [ ] Given (a Student), (provides their phone number), then we must allow for international telephone numbers in free text field.)
- [ ] Given (a Student), (has Not provided the information in the mandatory fields), then the system will not allow the student to save the account if mandatory information is missing and the student account is not created and a student may not apply for Financial Aid.

PR #111, opened 2021-01-28
SIMS #21 Student contact information

andrewsignori-aot:
Added the contact information to the student profile and SIN number in preparation for #68.

The bellows validation frameworks were considered and the one on this PR is the VueValidate.

- https://vee-validate.logaretm.com/v4/
- https://vuelidate.js.org/

Veutify supports both frameworks as 3rd party validators, as shown here: https://vuetifyjs.com/en/components/forms/#vee-validate, but we may not need a 3rd party validator when we change to Veutify, as shown in the below image (copied from the above link).

image

Both seem to be well accepted in the community:
https://www.reddit.com/r/vuejs/comments/emcsoa/veevalidate_vs_vuelidate/

vee-validate has more contributors and more starts on GitHub.

The "Save" button is just displaying an alert with the data. This should be enough to unblock the #78 that @fwpushan is currently working on.
The SIN specific validation will be added together with #68.
The "address line 2" is not required.
I am thinking of doing a POC using also the vee-validate so the team can have a better idea and define the best one to continue to use alongside #68.

Please find below how the screen looks like now.
image

image

### DIFF
--- a/sources/packages/web/src/components/ValidatedInput.vue
+++ b/sources/packages/web/src/components/ValidatedInput.vue
@@ -1,0 +1,38 @@
+<template>
+  <div>
+    <div :class="{ 'validation-error': property.$invalid }">
+      <slot></slot>
+    </div>
+    <span v-if="property.$invalid">
+      <ul class="validation-error-list">
+        <li v-for="e in property.$errors" :key="e.$message">
+          {{ e.$message }}
+        </li>
+      </ul>
+    </span>
+  </div>
+</template>
+
+<script lang="ts">
+export default {
+  props: {
+    property: {
+      type: Object,
+      required: true
+    }
+  }
+};
+</script>
+
+<style lang="scss">
+.validation-error-list {
+  list-style-type: none;
+  color: red;
+  padding-inline-start: 0px;
+  margin-top: 5px !important;
+}
+
+.validation-error input {
+  border: 2px solid red;
+}
+</style>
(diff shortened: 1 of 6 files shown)


abschwenker (commented):
lookin good

sources/packages/web/src/components/ValidatedInput.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div>
+    <div :class="{ 'validation-error': property.$invalid }">
+      <slot></slot>
```
fwpushan, line 4: I thing slot is deprecated and v-slot is introduced. 
https://vuejs.org/v2/guide/components-slots.html#Deprecated-Syntax
Can you check

fwpushan (commented):
LGTM, only one concern.

sources/packages/web/src/components/ValidatedInput.vue
andrewsignori-aot: Yes, that makes sense. I will change it.

andrewsignori-aot: Hi @fwpushan, the page that you mentioned is related to Vue 2 where it states the new way to do it in Vue 2 in comparison with the previous version.
The slots on Vue 3 are to be used it is in the code right now.
https://v3.vuejs.org/guide/component-slots.html#slot-content

2011sandeepraj (approved):
LGTM

PR #155, opened 2021-02-06
Userstory67 student updates info at bc service card

2011sandeepraj:
When 
Userinfo from token doesnt exist in SABC, this saves a record in SABC
UserInfo  From token is different from that in SABC, this updates the info in SABC
UserInfo From token is same as that in SABC , then simply return the same information

### DIFF
--- a/sources/packages/api/src/route-controllers/BaseController.ts
+++ b/sources/packages/api/src/route-controllers/BaseController.ts
@@ -1,6 +1,5 @@
 export default class BaseController {
-
-    public handleRequestError(e: any) {
+  public handleRequestError(e: any) {
     console.log(e);
   }
 }
--- a/sources/packages/api/src/route-controllers/BaseController.ts
+++ b/sources/packages/api/src/route-controllers/BaseController.ts
@@ -1,0 +1,6 @@
+export default class BaseController {
+
+    public handleRequestError(e: any) {
+    console.log(e);
+  }
+}
--- a/sources/packages/api/src/route-controllers/user/model/user.dto.ts
+++ b/sources/packages/api/src/route-controllers/user/model/user.dto.ts
@@ -1,0 +1,11 @@
+import { IsNotEmpty } from "class-validator";
+
+export default class UserSyncInfoDto {
+  @IsNotEmpty()
+  firstName: string;
+  @IsNotEmpty()
+  lastName: string;
+  @IsNotEmpty()
+  email: string;
+
+}
--- a/sources/packages/api/src/route-controllers/user/model/user.dto.ts
+++ b/sources/packages/api/src/route-controllers/user/model/user.dto.ts
@@ -7,5 +7,4 @@ export default class UserSyncInfoDto {
   lastName: string;
   @IsNotEmpty()
   email: string;
-
 }
--- a/sources/packages/api/src/route-controllers/user/user.controller.ts
+++ b/sources/packages/api/src/route-controllers/user/user.controller.ts
@@ -1,7 +1,10 @@
-import { Controller, Get, Patch, Req } from "@nestjs/common";
+import { Controller, Get, Patch, Req
+  //, NotFoundException 
+} from "@nestjs/common";
 import { Request } from "express";
 import { AuthService, UserService } from "../../services";
 import BaseController from "../BaseController";
+import UserSyncInfoDto from "./model/user.dto"
 
 @Controller("users")
 export class UserController extends BaseController {
@@ -13,18 +16,21 @@ export class UserController extends BaseController {
   }
 
   @Get("/check-user")
-  async checkUser(@Req() request: Request) {
+  async checkUser(@Req() request: Request): Promise<boolean> {
     try {
       const userInfo = this.authService.parseAuthorizationHeader(
         request.headers.authorization,
       );
       const userInSABC = await this.service.getUser(userInfo.userName);
-      if (!userInSABC) {
-        //TODO: Set a 404 in response when no user is found.
-        //response.status(404);
-      } else {
-        return userInSABC;
+      console.log('checkuser response in controller')
+      console.dir(userInSABC)  
+        // if (!userInSABC) throw new NotFoundException('User Not Found')
+      if(!userInSABC){
+        return false
+      }  else {
+        return true
       }
+
     } catch (error) {
       this.handleRequestError(error);
       throw error;
@@ -32,7 +38,7 @@ export class UserController extends BaseController {
   }
 
   @Patch("/sync-user")
-  async synchronizeUserInfo(@Req() request: Request) {
+  async synchronizeUserInfo(@Req() request: Request):Promise<UserSyncInfoDto> {
     try {
       const userInfoBCServiceCard = this.authService.parseAuthorizationHeader(
         request.headers.authorization,
@@ -40,7 +46,15 @@ export class UserController extends BaseController {
       const syncedUser = await this.service.synchronizeUserInfo(
         userInfoBCServiceCard,
       );
-      return syncedUser;
+      const userSyncInfo = new UserSyncInfoDto();
+      userSyncInfo.firstName = syncedUser.firstName;
+      userSyncInfo.lastName = syncedUser.lastName;
+      userSyncInfo.email = syncedUser.email;
+
+      console.log('User Sync Info DTO ');
+      console.dir(userSyncInfo);
+
+      return userSyncInfo;
     } catch (error) {
       this.handleRequestError(error);
       throw error;
--- a/sources/packages/api/src/route-controllers/user/user.controller.ts
+++ b/sources/packages/api/src/route-controllers/user/user.controller.ts
@@ -1,4 +1,4 @@
-import { Controller, Get, Req } from "@nestjs/common";
+import { Controller, Get, Patch, Req } from "@nestjs/common";
 import { Request } from "express";
 import { AuthService, UserService } from "../../services";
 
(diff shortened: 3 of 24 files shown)


sources/packages/api/src/route-controllers/BaseController.ts
```diff
@@ -0,0 +1,5 @@
+export default class BaseController {
+  public handleRequestError(e: any) {
+    console.log(e);
```
fwpushan, line 3: Re-throw error, nest error mechanism will handle this end send 502 error, otherwise in error condition server become unresponsive.

sources/packages/api/src/services/user/user.service.ts
```diff
@@ -14,4 +15,26 @@ export class UserService extends DataModelService<User> {
       userName,
     });
   }
+
+  async synchronizeUserInfo(userInfoBCServiceCard: UserInfo): Promise<User> {
+    let userToSync = await this.getUser(userInfoBCServiceCard.userName);
+
+    //Initializing a new user if one is not found
+    if (!userToSync) {
+      userToSync = new User();
```
fwpushan, line 24: we should throw error here.

fwpushan (changes requested):

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -1,20 +1,49 @@
...
+        request.headers.authorization,
+      );
+      const userInSABC = await this.service.getUser(userInfo.userName);
+      if (!userInSABC) {
+        //TODO: Set a 404 in response when no user is found.
+        //response.status(404);
+      } else {
+        return userInSABC;
+      }
+    } catch (error) {
```
andrewsignori-aot, line 28: Can we check if there is a global error handler or create a tech debt for it. In general, I would avoid adding a try/catch block in every single controller method.

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -1,20 +1,49 @@
...
+  @Patch("/sync-user")
+  async synchronizeUserInfo(@Req() request: Request) {
+    try {
+      const userInfoBCServiceCard = this.authService.parseAuthorizationHeader(
+        request.headers.authorization,
+      );
+      const syncedUser = await this.service.synchronizeUserInfo(
+        userInfoBCServiceCard,
+      );
+      return syncedUser;
```
andrewsignori-aot, line 43: Is this the object from the database? In general, I believe that we should avoid returning objects direct from the database model.
My understanding is that we should create models to return the exact information needed for every single information being returned form the API and being submitted.
@fwpushan, is it of for you also?

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -1,20 +1,49 @@
...
-    );
+  async checkUser(@Req() request: Request) {
+    try {
+      const userInfo = this.authService.parseAuthorizationHeader(
+        request.headers.authorization,
+      );
+      const userInSABC = await this.service.getUser(userInfo.userName);
+      if (!userInSABC) {
+        //TODO: Set a 404 in response when no user is found.
+        //response.status(404);
```
andrewsignori-aot, line 24: If it is needed you can check on the student controller how different HTTP errors are returned, like below.
```suggestion
//response.status(404);
throw new HttpException(
        {
          status: HttpStatus.FORBIDDEN,
          message: "Wrong token",
        },
        HttpStatus.FORBIDDEN,
      );
```

sources/packages/web/src/services/http/UserApi.ts
```diff
@@ -9,13 +9,33 @@ export default interface User {
...
-  public async checkUser(): Promise<User> {
+  public async checkUser(): Promise<boolean> {
+    let userExists = false;
     try {
       const response = await this.apiClient.get(
         "users/check-user",
         this.addAuthHeader(),
       );
-      return response.data as User;
+      if (response.status === 200) {
```
andrewsignori-aot, line 19: We can add the same or similar npm that is used on API for HTTP Status Codes (e.g. HttpStatus.FORBIDDEN).

sources/packages/web/src/services/http/UserApi.ts
```diff
@@ -9,13 +9,33 @@ export default interface User {
...
+        const user = response.data as User;
+        //TODO: This should not ideally rely on id.
```
andrewsignori-aot, line 21: If the status code on API will be changed to "NOT FOUND", this check can be changed as you mentioned in the comments.

andrewsignori-aot (commented):
@2011sandeepraj, there are no major issues but please take a look in the comments.

sources/packages/api/src/route-controllers/BaseController.ts
2011sandeepraj: Its already thrown in the caller  i.e. user.controller.ts line 29.

sources/packages/api/src/route-controllers/user/user.controller.ts
2011sandeepraj: Done

sources/packages/api/src/route-controllers/user/user.controller.ts
2011sandeepraj: Created a Tech Debt:157
https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/157

sources/packages/api/src/services/user/user.service.ts
2011sandeepraj: We need to save the information so shouldn't be an error. Can discuss if needed

sources/packages/web/src/services/http/UserApi.ts
2011sandeepraj: There was a case which i am not able to reproduce now, where i got a user with id as blank. So this i think is still valid. Can discuss if needed.

sources/packages/api/src/route-controllers/user/model/user.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { IsNotEmpty } from "class-validator";
```
andrewsignori-aot, line 1: Why do we need this object? Are we receiving this data from the client?
I believe that this data should come from the token, right?

sources/packages/api/src/route-controllers/user/model/user.dto.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default class UserSyncInfoDto {
+  @IsNotEmpty()
+  firstName: string;
+  @IsNotEmpty()
+  lastName: string;
+  @IsNotEmpty()
+  email: string;
+
```
andrewsignori-aot, line 10: This blank line is supposed to be removed by the formatter. Is the formatter working properly?

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -1,20 +1,63 @@
...
+      );
+      const syncedUser = await this.service.synchronizeUserInfo(
+        userInfoBCServiceCard,
+      );
+      const userSyncInfo = new UserSyncInfoDto();
+      userSyncInfo.firstName = syncedUser.firstName;
+      userSyncInfo.lastName = syncedUser.lastName;
+      userSyncInfo.email = syncedUser.email;
+
+      console.log('User Sync Info DTO ');
```
andrewsignori-aot, line 54: Please remove the "console.log" before the final commit.

sources/packages/api/src/services/user/user.service.ts
```diff
@@ -14,4 +15,28 @@ export class UserService extends DataModelService<User> {
       userName,
     });
   }
+
+  async synchronizeUserInfo(userInfoBCServiceCard: UserInfo): Promise<User> {
+    let userToSync = await this.getUser(userInfoBCServiceCard.userName);
+
+    //Initializing a new user if one is not found
```
andrewsignori-aot, line 22: I believe that this comment is from a past code. Can you please remove it?

sources/packages/api/src/services/user/user.service.ts
```diff
@@ -14,4 +15,28 @@ export class UserService extends DataModelService<User> {
...
+    if (!userToSync) {
+        throw new Error(
+          `Not able to find a user using the username (bcsc name)`
```
andrewsignori-aot, line 25: Consider do not use the string interpolation syntax if there is no variable to be added in the text.

sources/packages/web/src/services/http/UserApi.ts
```diff
@@ -9,13 +9,26 @@ export default interface User {
...
+    } catch (error) {
+      this.handleRequestError(error);
+      throw error;
+    }
+  }
+
+  public async synchronizeUserInfo(): Promise<void> {
+    try {
+      /**
+       * TODO Dont quite agree with sending a null in a patch/put request but going with it for now
```
andrewsignori-aot, line 29: Please propose an alternative method that would make more sense or please consider remove this comment.

andrewsignori-aot (commented):

sources/packages/api/src/services/user/user.service.ts
fwpushan: @andrewsignori-aot Incase of multiline string or string with "/', we can use ``.

andrewsignori-aot (approved):
👍

2011sandeepraj:
Userstory67 student updates info at bc service card, info being sycned between SABC and BC Services Card

Issue #137, opened 2021-02-02
Secure API end point with KeyCloak Token Verification


fwpushan:
**Describe the task**
Secure API end point with KeyCloak Token Verification. SIMS API server can verKeyCloak client-side JWT token

**Acceptance Criteria**
- [x] SIMS API server can verKeyCloak client-side JWT token
- [x] Use api specific client to validate token
- [x] Create tests

**Additional context**

PR #159, opened 2021-02-09
SIMS #137 - Secure API end point with KeyCloak Token Verification

andrewsignori-aot:
Enabled the JWT "offline" validation on the API using the passport package. After a team discussion, we decided to not validate the received token on API on every request using a "bearer-only" client on the Keyclock.

Implemented global authentication and roles-based authentication, as described on Nestjs docs.
https://docs.nestjs.com/security/authentication
https://docs.nestjs.com/security/authorization

Implemented the @UserToken parameter decorator to provided typed access to user token information, as shown below.
image

Retrieving the public key for JWT validation direct from the Keycloak issuer URL (e.g. https://dev.oidc.gov.bc.ca/auth/realms/jxoe2o46).

The public key is gonna be retrieved and converted to a PEM format to later be consumed by passport JWT validation.
More about PEM format: https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail

Replaced the nest-keycloak-connect by the packages based os passport.
Removed env variables KEYCLOAK_CLIENT_API  and KEYCLOAK_CLIENT_SECRET that are no longer needed.

Moved the config endpoint alongside the other controllers.

All the endpoints will require an authentication token unless they are decorated with the "Public" decorator, as shown below.
image

The roles authorization is also ready to be used as shown below.
image


sources/packages/api/src/app.module.ts
```diff
@@ -2,18 +2,26 @@ import { Module } from "@nestjs/common";
 import { AppController } from "./app.controller";
 import { AppService } from "./app.service";
 import { DatabaseModule } from "./database/database.module";
-import { ConfigController } from "./config/config.controller";
+import { ConfigController } from "./route-controllers/config/config.controller";
```
fwpushan, line 5: Add config controller export in the route-controller/index.ts and then import the controller in line 7

fwpushan (commented):
Excellent job. Just only small comment. Is it possible to add a basic test suit to validate this flow. or we can add technical debt. I think an e2e test suit is required for this.

sources/packages/api/src/app.module.ts
andrewsignori-aot: Goot catch! Code changed to import the ConfigController and ConfigService using their respective index.ts files.

andrewsignori-aot:
> Excellent job. Just only small comment. Is it possible to add a basic test suit to validate this flow. or we can add technical debt. I think an e2e test suit is required for this.

Tech debt created: Secure API endpoint - E2E tests#174

sources/packages/api/src/auth/auth.module.ts
```diff
@@ -1,37 +1,29 @@
...
     },
     {
       provide: APP_GUARD,
-      useClass: RoleGuard,
-    },*/
+      useClass: RolesGuard,
+    },
   ],
 })
 export class AuthModule {}
```
2011sandeepraj, line 29: Remove the parseAuthorizationHeader methods

sources/packages/api/src/route-controllers/config/config.controller.ts
```diff
@@ -0,0 +1,22 @@
+import { Controller, Get } from "@nestjs/common";
+import { Public } from "src/auth/decorators/public.decorator";
+import { ConfigService } from "src/services/config/config.service";
+import { IConfig } from "./models/get-config.dto";
```
2011sandeepraj, line 4: Let's standardize import paths. Use Relative Paths

2011sandeepraj (commented):
Some Minor COmments

sources/packages/api/src/auth/auth.module.ts
andrewsignori-aot: Done, good catch 👍

sources/packages/api/src/route-controllers/config/config.controller.ts
andrewsignori-aot: Done!

andrewsignori-aot:
> Some Minor COmments

I went through the comments and are all done.

fwpushan (approved):

PR #173, opened 2021-02-09
[web][SIM-104] Starting Application and page redirection

fwpushan:
Add button for application start in Home View
Add placeholder for application page
Add page not found page
Include basic test
SIM-104

### DIFF
--- a/sources/Makefile
+++ b/sources/Makefile
@@ -82,8 +82,7 @@ postgres:
 # Local API only
 api-only:
 	@echo "+\n++ Make: Building and running API only"
-	@docker-compose -f docker-compose.yml up -d postgres 
-	sleep 5
+	@docker-compose -f docker-compose.yml up -d postgres
 	@docker-compose -f docker-compose.yml up api
 
 # Local App
--- a/sources/packages/web/src/router/index.ts
+++ b/sources/packages/web/src/router/index.ts
@@ -2,6 +2,8 @@ import { createRouter, createWebHistory, RouteRecordRaw } from "vue-router";
 import Home from "../views/Home.vue";
 import Login from "../views/Login.vue";
 import Student from "../views/Student.vue";
+import FAApplication from "../views/FAApplication.vue";
+import PageNotFound from "../views/PageNotFound.vue";
 
 const routes: Array<RouteRecordRaw> = [
   {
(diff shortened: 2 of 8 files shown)


sources/packages/web/src/router/index.ts
```diff
@@ -2,6 +2,8 @@ import { createRouter, createWebHistory, RouteRecordRaw } from "vue-router";
 import Home from "../views/Home.vue";
 import Login from "../views/Login.vue";
 import Student from "../views/Student.vue";
+import FAApplication from "../views/FAApplication.vue";
```
andrewsignori-aot, line 5: What does FA mean?

fwpushan: @andrewsignori-aot  Financial Aid Application

sources/packages/web/src/views/Home.vue
```diff
@@ -1,13 +1,31 @@
 <template>
-  <div>Home - Student Dashboard</div>
-  <div>
-    You are here because Account information for this user exists within SABC
+  <div style="padding-left: 10px">
```
andrewsignori-aot, line 2: The current framework has a set of classes to be used for spacing. The same as line 5.
This is a real minor comment because we are gonna change the framework, but it could be handy to use the already existing classes.
https://primefaces.org/primevue/showcase/#/spacing

andrewsignori-aot (approved):
Looks good to me 👍

sources/packages/web/src/views/Home.vue
fwpushan: FIXED

sources/Makefile
```diff
@@ -82,8 +82,7 @@ postgres:
 # Local API only
 api-only:
 	@echo "+\n++ Make: Building and running API only"
-	@docker-compose -f docker-compose.yml up -d postgres 
-	sleep 5
```
2011sandeepraj, line 86: Any reason we put this and now removed it?

2011sandeepraj (approved):
Approved but asked one quick question

sources/Makefile
fwpushan: In that particular I’m spinning up DB container and API container separately.
Initially I though a few send sleep will provide enough time to bootstrap db before API container try to connect it.
But in single terminal env that has no effect.
So, I remove it.

Issue #105, opened 2021-01-26
Student Confirms Profile information


mzmichellesmith:
**As a** *(Student)* **I want** *(confirm my profile information)* **So That** *(so that I can verify that my address is current on my application)*

**Additional Context**
- The Profile information is the source of Truth on the Student Profile- not the application
- 
**Acceptance Criteria**
- [ ] Given (A Student), When (A student initiates an application), Then they are confirming that Read-Only view of both BC Services Card and SABC Profile Information is correct.
- [ ] Given (A student), When (cannot confirm the Profile Information), Then (they will be redirected to the Student Profile Section).
- [ ] Given (A student), When (they confirm their profile info is correct), Then (they select 'Yes), 
- [ ] Given (A student), When (they confirm their profile info is incorrect), Then (they select 'No, I need to update my profile information')
- [ ] Given (A student), When (they are viewing the profile information ), Then (it is NOT displayed as editable (its greyed out) for both Services Card and SABC Profile Information

**Technical Tasks**
- [ ] Display the entire student profile as a read-only. Probably creating a component to hold this. - Sandeep
- [ ] Starting saving to DB the fields that are not being saved right now (e.g. DOB and Gender) - Andrew Signori
- [ ] Update the new fields on DB during user synchronization after the login. - Andrew Signori
- [ ] Expose in the API an endpoint to retrieve all the student data needed to populate all the fields. - Sandeep

andrewsignori-aot:
![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/e6674180-981a-4265-8ff5-0bbd01dd4fc9)

PR #180, opened 2021-02-10
SIMS #105 Adding birth date and gender to students table

andrewsignori-aot:
Birthdate and gender added to the student's table.
Birthdate and gender added to DB during student creation and also updated during student login.
Gender is not being validated to ensure that anything that comes from BCSC will be saved to DB and will not cause issues while the user is being registered in the portal.

This story is a partial delivery of the entire ticket.


sources/packages/api/src-sql/StudentsAddGenderAndDOB/Add-GenderAndDOB.sql
```diff
@@ -0,0 +1,24 @@
+-- Adding new column date_of_birth
+-- Add the column with a dummy default value that will be removed in the next command.
+ALTER TABLE sims.students
+    ADD COLUMN birth_date Date NOT NULL DEFAULT now();
```
2011sandeepraj, line 4: Is there a way to mention the format? DD-MM-YYYY or YY-MM-DD?

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -76,4 +78,44 @@ export class StudentService extends RecordDataModelService<Student> {
...
+    ) {
+      studentToSync.user.email = userInfo.email;
+      studentToSync.user.lastName = userInfo.lastName;
+      studentToSync.user.firstName = userInfo.givenNames;
+      mustSave = true;
+    }
+
+    let birthDate = new Date(userInfo.birthdate);
+    if (
+      birthDate !== studentToSync.birthdate ||
```
2011sandeepraj, line 106: This is going to be string comparison and not comparison of actual dates. Intentional?

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -31,7 +34,9 @@ export class UserController extends BaseController {
     @UserToken() userToken: IUserToken,
   ): Promise<UserSyncInfoDto> {
     try {
-      const syncedUser = await this.service.synchronizeUserInfo(userToken);
+      const syncedUser = await this.studentService.synchronizeUserInfo(
```
2011sandeepraj, line 37: This is syncedStudent and not syncedUser right? If yes, can we name it accordingly

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -1,13 +1,16 @@
...
 import UserSyncInfoDto from "./model/user.dto";
 import { UserToken } from "src/auth/decorators/userToken.decorator";
 import { IUserToken } from "src/auth/userToken.interface";
 
 @Controller("users")
 export class UserController extends BaseController {
-  constructor(private readonly service: UserService) {
+  constructor(
+    private readonly service: UserService,
+    private readonly studentService: StudentService,
```
2011sandeepraj, line 12: Because this is now so much to do with student and not user per se, i think it makes more sense to have all this code in student.controller.ts

sources/packages/api/src-sql/StudentsAddGenderAndDOB/Add-GenderAndDOB.sql
```diff
@@ -0,0 +1,24 @@
...
+
+-- Removing the dummy constraint created just to allow us
+-- to add a "NOT NULL" column in an exising table.
+ALTER TABLE sims.students
```
fwpushan, line 8: For safety purpose  Can we add IF EXISTS / IF NOT EXISTS clause.

sources/packages/api/src/types/userInfo.ts
```diff
@@ -3,4 +3,6 @@ export interface UserInfo {
   givenNames: string;
   email: string;
   userName: string;
+  birthdate: string;
```
2011sandeepraj, line 6: Because our model is organized differently, i just think it makes more sense to either introduce another studentInfo.ts with the additional fields instead of putting it all here in userInfo.ts
and
We can rename this userInfo.ts to call this as BaseUserInfo.ts

2011sandeepraj (changes requested):
Let me know what you think of comments/suggestions.

sources/packages/api/src/services/student/student.service.ts
andrewsignori-aot: Both objects are Date types. It is supposed to be a Date/Date comparison.
Any particular reason to you believe that it would be a string comparison? Am I missing something maybe?

sources/packages/api/src-sql/StudentsAddGenderAndDOB/Add-GenderAndDOB.sql
andrewsignori-aot: Can you be more specific about where the date format would be applied here?

sources/packages/api/src/route-controllers/user/user.controller.ts
andrewsignori-aot: I struggled a bit with the name also because it is gonna update data on the user(first name, last name, and email) and also on the student (DOB and gender).
So, we are basically using User data to update Users/Students table.
Maybe we can say this.studentService.synchronizeFromUserInfo, what do you think?

sources/packages/api/src/route-controllers/user/user.controller.ts
andrewsignori-aot: I struggled a bit with it also because it is gonna update data on the user(first name, last name, and email) and also on the student (DOB and gender).
We can move it to student, but then I believe that we should move it to student controller and rename the method from synchronizeUserInfo to synchronizeStudentInfo on API endpoint, what do you think?

sources/packages/api/src-sql/StudentsAddGenderAndDOB/Add-GenderAndDOB.sql
andrewsignori-aot: Added the "IF EXISTS / IF NOT" and tested the scripts executing more than one time in the db.

sources/packages/api/src/types/userInfo.ts
andrewsignori-aot: If you look from a database perspective our models are in a way that does not make sense to add these properties here.
That said, my understanding was that this object was representing our business model, where both properties belong to the user and not the student. Let's talk about it.

sources/packages/api/src/route-controllers/student/model/student-sync-info.dto.ts
```diff
@@ -0,0 +1,7 @@
+export default class StudentSyncInfoDto {
```
andrewsignori-aot, line 1: Added to keep the behavior already in place for sync user return this data.
It is a new class only because it was moved from User models to Student models.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -91,4 +93,21 @@ export class StudentController {
...
+  @Patch("/sync")
+  async synchronizeFromUserInfo(
+    @UserToken() userToken: IUserToken,
+  ): Promise<UserSyncInfoDto> {
+    const syncedUser = await this.studentService.synchronizeFromUserInfo(
+      userToken,
+    );
+
+    return {
+      firstName: syncedUser.user.firstName,
```
andrewsignori-aot, line 106: Keeping the same behavior of returning the data that was already in place when this method was in the User endpoint.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -15,12 +15,14 @@ import {
 } from "./models/student.dto";
 import { UserToken } from "../../auth/decorators/userToken.decorator";
 import { IUserToken } from "../../auth/userToken.interface";
+import UserSyncInfoDto from "./model/student-sync-info.dto";
```
2011sandeepraj, line 18: This will be StudentSyncInfoDto right?

2011sandeepraj, line 18: This wll be StudentSyncInfoDto right?

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -91,4 +93,21 @@ export class StudentController {
     // Save student
     return this.service.createStudent(userToken, payload);
   }
+
+  @Patch("/sync")
+  async synchronizeFromUserInfo(
+    @UserToken() userToken: IUserToken,
+  ): Promise<UserSyncInfoDto> {
```
2011sandeepraj, line 100: Read that patch shouldnt return body so please remove the return UserSyncInfoDto Promise

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -91,4 +93,21 @@ export class StudentController {
...
+    const syncedUser = await this.studentService.synchronizeFromUserInfo(
```
2011sandeepraj, line 101: better variable name syncedStudent? Just asking

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -1,7 +1,6 @@
 import { Controller, Get, Patch } from "@nestjs/common";
-import { UserService } from "../../services";
+import { StudentService, UserService } from "../../services";
```
2011sandeepraj, line 2: Unused import StudentService ?

2011sandeepraj (commented):
May be a few more changes

fwpushan (approved):
LGTM

sources/packages/api/src/route-controllers/student/student.controller.ts
andrewsignori-aot: Just kept it because after our last conversation about it you had a strong position to do it :D
I will be happy to remove it 👍

2011sandeepraj (approved):
All Good

Issue #144, opened 2021-02-03
Student Views 'Are you Ready to Apply' Application Welcome Page


mzmichellesmith:
**As a** *(Student)* **I want** * Know upfront what I need to provide in an application)* **So That** *(i can have everything I need prepared to complete the application)*

**Additional Context**
- This page provides the information the student will need in order to complete their Application
- Information is TBD (eg, CRA Notice of Assessment) 

**Acceptance Criteria**
- [x] Given ( A student), When (they are directed to the application welcome page), Then (they are greeted with a screen that says "Are you Ready to Apply)

PR #182, opened 2021-02-11
Fixing the issue causing the deploy to fail

andrewsignori-aot:
Fix for a minor issue not allowing the web project to build properly introduced during the SIMS #144.


andrewsignori-aot:
Going ahead with the merge because it is a quick fix and other devs could take a look at this PR later.

PR #185, opened 2021-02-11
[web][SIM-126] Citizen and Residency Questionnaire

fwpushan:
1. Questionnire page
2. Add Section + Question component

Added radio buttons for the below sections:
- Citizen and Residency
- Bankruptcy
- Relationship Status

image

image

image


sources/packages/web/src/views/PersonalInfoQuestionnaire.vue
```diff
@@ -0,0 +1,167 @@
...
+  bankruptcySelectedValue: string;
+  relationshipSelectedValue: string;
+}
+
+const yesNoOptions = [
+  { text: "Yes", value: "yes" },
+  { text: "No", value: "no" },
+];
+
+const questionnaire = {
```
andrewsignori-aot, line 99: Seems to me that the questions in this format are becoming harder to read and, since we do not have a requirement to then be dynamic, I am not sure if we should keep them in this format or add the text for "Title", "SubTitle" and "Questions" directly in the components.

fwpushan: I always recommend to write all your static text in constants file and bind them with UI but I agree, current structure is complex, I'm adding [technical debt](https://github.com/bcgov/SIMS/issues/188) for this page.

2011sandeepraj (approved):

PR #206, opened 2021-02-18
[web][SIM-160] Fix text for next button on Application/PI page

fwpushan:
--------------------------------------------------------------
1. Change button text
2. Auto fix lint issues
3. Disable explicit any warning
4. Fix some non auto fixable lint issues


sources/packages/web/src/components/fa-application/Question.vue
```diff
@@ -10,11 +10,14 @@
...
 <script lang="ts">
+interface QuestionProp {
+  text: string;
+}
 export default {
   props: {
     text: String,
   },
-  setup(prop: any) {
+  setup() {
```
andrewsignori-aot, line 20: Please, consider removing the setup method since it not being used.

sources/packages/web/src/components/fa-application/Section.vue
```diff
@@ -20,7 +20,7 @@ export default {
     title: String,
     subTitle: String,
   },
-  setup(props: any) {
+  setup() {
```
andrewsignori-aot, line 23: Please, consider removing the setup method since it not being used.

sources/packages/web/.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   rules: {
     "no-console": process.env.NODE_ENV === "production" ? "warn" : "off",
     "no-debugger": process.env.NODE_ENV === "production" ? "warn" : "off",
+    "@typescript-eslint/no-explicit-any": "off",
```
andrewsignori-aot, line 19: Should we really disable it? It will allow us to incorrectly leave other untyped variables in the code because all of them will be considered "any", right?
I would recommend keeping it.

andrewsignori-aot (commented):
Please take a look at the comments!

andrewsignori-aot (approved):

sources/packages/web/src/views/PersonalInfoQuestionnaire.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="p-component">
+  <div class="p-component p-m-8">
```
2011sandeepraj, line 2: Actually am unsure about adding m-* classes simply because we are not testing for them are we?
My thought is they are for mobile devices or smaller screens. Even i have been adding them unconvincingly. can you initiate a chat/discussion around this point please

sources/packages/web/src/views/PersonalInfoQuestionnaire.vue
```diff
@@ -149,7 +149,7 @@
       </Section>
       <HorizontalSeparator />
     </div>
-    <Button @click="onNext" class="p-m-6">Next</Button>
+    <Button @click="onNext" class="p-m-6">Next Section</Button>
```
2011sandeepraj, line 152: Are we not supposed to have a Previous Section button too as per the figma?

2011sandeepraj (commented):
Left a couple of comments

sources/packages/web/src/views/PersonalInfoQuestionnaire.vue
fwpushan: This is size class and for spacing only.

sources/packages/web/src/views/PersonalInfoQuestionnaire.vue
fwpushan: @2011sandeepraj , I will added that as per your recommendation but in SIM-160, there is no specific acceptance criteria for this.

2011sandeepraj: Thanks. Once added, consider the PR approved. Please go ahead with merge to main.

2011sandeepraj (approved):

PR #212, opened 2021-02-19
Sim152 singlesourceoftruth db

2011sandeepraj:
Functionality changed so that DB is taken as the source of truth
Made contacts as part of a subclass rather than parent level fields in StudentInfo
Renaming endpoints and classes to make them generic enough
Added Email to returned value
Added UTC to date formatter to avoid substraction of one day
Updated ConfirmStudentAidBCProfileInfo screen and Student screen to match new contracts


sources/packages/api/src/helpers/utilfunctions.ts
```diff
@@ -1,11 +1,12 @@
...
+import { StudentInfo } from "../route-controllers/student/models/student.dto";
 export default class Helper {
-  static mapAddressAttributes(fromObject: any, toObject: StudentConfirmInfo) {
-    toObject.addressLine1 = fromObject.addressLine1;
-    toObject.addressLine2 = fromObject.addressLine2;
-    toObject.city = fromObject.city;
-    toObject.provinceState = fromObject.province;
-    toObject.country = fromObject.country;
-    toObject.postalCode = fromObject.postalCode;
+  static mapAddressAttributes(fromObject: any, toObject: StudentInfo) {
```
fwpushan, line 3: I think there is type available for StudentContact. Please check [here](https://raw.githubusercontent.com/bcgov/SIMS/main/sources/packages/api/src/types/studentContact.ts)

sources/packages/api/src/helpers/utilfunctions.ts
```diff
@@ -1,11 +1,12 @@
...
+    toObject.contact.addressLine1 = fromObject.addressLine1;
+    toObject.contact.addressLine1 = fromObject.addressLine1;
```
fwpushan, line 5: duplicate of line 4

sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -1,16 +1,19 @@
...
-  addressLine2: string;
-  city: string;
-  provinceState: string;
-  country: string;
-  postalCode: string;
+  email: string;
   gender: string;
-  phoneNumber: string;
   dateOfBirth: Date;
+  contact = new (class contact {
```
fwpushan, line 8: Again type available for Student contact [check-here](https://raw.githubusercontent.com/bcgov/SIMS/main/sources/packages/api/src/types/studentContact.ts)

fwpushan (changes requested):
Only few comments.

sources/packages/api/src/route-controllers/student/models/student.dto.ts
andrewsignori-aot, line 8: I would prefer to not use syntax like "new (class contact".
I believe that just property = { ... } would be clear.

sources/packages/api/src/helpers/utilfunctions.ts
andrewsignori-aot, line 3: Is there a particular reason to have a mapper for this?
Seems to me that the only property that is different between both objects is "province".
If that is the case, I would recommend not create a helper only for this case.
If there is a strong reason to keep it, please change the "fromObject" to a specific type.

sources/packages/web/src/components/fa-application/ConfirmStudentAidBCProfileInfo.vue
```diff
@@ -35,19 +35,18 @@
...
         <label for="homeAddress" class="label-extra">Home Address</label>
         <div class="p-field p-col text-extra">
-          {{ studentConfirmInfo.addressLine1 }},<br />{{
-            studentConfirmInfo.addressLine2
-          }}<br />{{ studentConfirmInfo.city }}
-          {{ studentConfirmInfo.provinceState }}
-          {{ studentConfirmInfo.postalCode }}<br />{{
-            studentConfirmInfo.country
-          }}<br />
+          {{ studentContact.addressLine1 }},<br />
```
andrewsignori-aot, line 38: I believe that this should be a computed property or something that would allow us to, for instance, do not display a blank line for address line 2 when it was not present. If I am not wrong this is the only non-required field.

sources/packages/web/src/components/fa-application/ConfirmStudentAidBCProfileInfo.vue
```diff
@@ -71,16 +70,17 @@ import { StudentService } from "../../services/StudentService";
 export default {
   setup() {
     //Creating a reactive array
-    const studentConfirmInfo = ref([]);
+    const studentInfo = ref([]);
+    const studentContact = ref([]);
```
andrewsignori-aot, line 74: Do we need a second object to hold the contact information?
Can we use just studentInfo?

sources/packages/web/src/views/Student.vue
```diff
@@ -180,18 +179,17 @@ export default {
     },
   },
   setup(props: any) {
-    // Readonly student data from state.
-    const store = useStore();
     const router = useRouter();
     const toast = useToast();
-
-    const readonlyProfile = computed(() => store.state.student.profile);
```
andrewsignori-aot, line 188: How the read-only information is being loaded during the user creation? Am I am missing something in the new implementation?

andrewsignori-aot (changes requested):
Please review the comments.

sources/packages/api/src/helpers/utilfunctions.ts
2011sandeepraj: Thats a type. In the controller we are using models. The current code merges information from two services and prepares a data transfer object using the class in the model folder.

sources/packages/api/src/helpers/utilfunctions.ts
2011sandeepraj: fixed

sources/packages/api/src/helpers/utilfunctions.ts
2011sandeepraj: Got rid of the mapper

sources/packages/web/src/components/fa-application/ConfirmStudentAidBCProfileInfo.vue
2011sandeepraj: i tried it , and nested infromation was being a problem

2011sandeepraj: And to your comment on addressline2 and space, i have fixed it but that's not really related to this user story. So if there are more requirements around formatting of address(which i don't think there will be) we can tackle it later

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -13,12 +13,11 @@ import {
   CreateStudentDto,
   GetStudentContactDto,
   UpdateStudentContactDto,
-  StudentConfirmInfo,
 } from "./models/student.dto";
 import { UserToken } from "../../auth/decorators/userToken.decorator";
 import { IUserToken } from "../../auth/userToken.interface";
 import BaseController from "../BaseController";
-import Helper from "../../helpers/utilfunctions";
+import { StudentInfo } from "src/types/studentInfo";
```
fwpushan, line 20: Use absolute path here.

sources/packages/web/src/services/StudentService.ts
```diff
@@ -58,14 +57,16 @@ export class StudentService {
...
-    studentConfirmInfo.birthDateFormatted = Helper.formatDate(
-      studentConfirmInfo.dateOfBirth,
-    );
-    return studentConfirmInfo;
+  async getStudentInfo() {
+    const studentInfo = await ApiClient.Students.getStudentInfo();
+    const studentInfoAll = {
+      ...studentInfo,
+      //Formatting date received from api from 1998-03-24T00:00:00.000Z to March 24, 1998
+      birthDateFormatted: Helper.formatDate(studentInfo.dateOfBirth),
```
fwpushan, line 65: why two formatted birthDate? Can we keep date display formatting consistent? If require we can flag @roopfw

fwpushan (commented):
Only Few doubts, otherwise OK.

fwpushan (approved):

sources/packages/api/src/helpers/utilfunctions.ts
2011sandeepraj: CHanged as discussed

sources/packages/web/src/components/fa-application/ConfirmStudentAidBCProfileInfo.vue
2011sandeepraj: Actually not related to the user story directly, but have fixed it. If there are more formatting changes needed, can do that after this one is merged in a subsequent PR.

sources/packages/web/src/views/Student.vue
2011sandeepraj: FIxed it with the PR update

sources/packages/api/src/route-controllers/student/student.controller.ts
2011sandeepraj: Done

sources/packages/web/src/services/StudentService.ts
2011sandeepraj: For now, have done as per the figma. Can remove it when needed.

sources/packages/web/src/views/Student.vue
```diff
@@ -179,17 +176,30 @@ export default {
     },
   },
   setup(props: any) {
+    // Readonly student data from state.
+    const store = useStore();
     const router = useRouter();
     const toast = useToast();
     const readonlyProfile = ref({});
+    const birthDate = ref();
```
andrewsignori-aot, line 184: Is there a reason to birthDate and givennames, even being part of the readonlyProfile, have separated objects returned on setup to bind the UI?
My understanding is that readonlyProfile would be populated from the state or from DB, but their bind to the UI would be done using only readonlyProfile for the read-only fields.

andrewsignori-aot (changes requested):
Please check the comments about the read-only UI bind.

sources/packages/web/src/views/Student.vue
2011sandeepraj: The state doesn't have the field First Name, it has GivenNames as coming from Keycloak but in our db we save First name. Does this explain? if not we can have a quick chat on Monday

andrewsignori-aot: We can have a chat about this. The comment was not about the variables naming. The question is why birthDate and givennames (or first name) should be returned as separated objects in the setup.

andrewsignori-aot (approved):
👍

Issue #169, opened 2021-02-09
Student Provides Program Information- Start Date and End Date


mzmichellesmith:
**As a** *(Student)* **I want** *(to provide my Education Program Start and End Date)* **So That** *(I can accurately be assessed for Financial Aid)

*Additional Context**
- The default answer will be blank, the field is mandatory on 'Submit' (Another User Story)
- All of the Institution and Program Information is listed on One Page

**Figma Wireframe Link**
[Loan Application - Institution/Program ](https://www.figma.com/file/VykmBDwobGoGvK9M5wgOvN/FreshWorks-Studios-StudentAid-BC-App-Design-System-WIP?node-id=123%3A1234)

**Acceptance Criteria**
- [x] Given (The Student), When (provides their program start-date ), Then (they should be day/month/year).
- [x] Given (The Student), When (provides their program end-date), Then (they should be day/month/year).

PR #213, opened 2021-02-19
[web][SIM-168] Add option to search education program

fwpushan:
------------------------------------------------------
1. Add new EducationProgramList.vue component and integrate with SelectProgram page
2. Add some state management
3. Also added code for SIM - #169 

Screenshot 2021-02-20 at 3 14 27 AM

### DIFF
--- a/sources/packages/web/src/services/InstituteService.ts
+++ b/sources/packages/web/src/services/InstituteService.ts
@@ -3,6 +3,11 @@ export interface Institute {
   code?: string;
 }
 
+export interface EducationProgram {
+  name: string;
+  code?: string;
+}
+
 export class InstituteService {
   // Share Instance
   private static instance: InstituteService;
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/services/InstituteService.ts
```diff
@@ -3,6 +3,11 @@ export interface Institute {
   code?: string;
 }
 
+export interface EducationProgram {
+  name: string;
+  code?: string;
```
andrewsignori-aot, line 8: Is there a reason to assume that this would be nullable?

andrewsignori-aot (approved):
👍

sources/packages/web/src/services/InstituteService.ts
fwpushan: For now this is placeholder only, we will revisit this after detailed API and data modelling.

2011sandeepraj (approved):
LGTM

Issue #174, opened 2021-02-09
Secure API endpoint - E2E tests


andrewsignori-aot:
After SIMS #137 is finalized we need to add some E2E tests to test the authentications flows.
Using the SUPER framework.

Test the authentication layer.

Authenticated Users 
Role Access
Public Access

PR #220, opened 2021-02-22
SIMS #174 - Auth e2e test

andrewsignori-aot:
Enabled E2E test to test authentication scenarios as shown below, following the recommendations here: https://docs.nestjs.com/fundamentals/testing#end-to-end-testing

In general, the E2E tests are executed in two steps:

1. Get a valid token from Keyclock;
2. Try to authenticate to a fake API (but with real functionality) controller (AuthTestController) 

image

Key observations:

- To allow us to test the E2E authentication cycle, a dummy user (student_e2e_test) was added to the Keycloak DEV instance.
- This user is used to retrieve the token from Keycloak and authenticate the API endpoints.
- The AuthTestController controller was created to be the target for all E2E tests related to the authentication. This controller is not part of the final API (not available on DEV/TEST/PROD) and it is being used only for the E2E tests (see comments on the file auth-test.controller.ts). 
- The Keycloak endpoints are retrieved from the discovery endpoint ".well-known/openid-configuration".
- To run the tests locally the ENV variable E2E_TEST_STUDENT_PASSWORD need to be defined in a local .env file, to avoid add the Keyclock dummy user password in the source code.
- It is not part of this PR the enablement of E2E on the CI/CD. We need to do this for UnitTest/E2E Test/Lint, etc.
- Along the process, some changes affected the Unit Tests already in place and some changes were needed to keep them working.
image


sources/packages/api/src/services/config/config.service.ts
```diff
@@ -9,14 +9,20 @@ export class ConfigService {
         url: process.env.KEYCLOAK_AUTH_URL,
         realm: process.env.KEYCLOAK_REALM,
         clientId: process.env.KEYCLOAK_CLIENT,
-        issuerUrl: this.getIssuerEndpoint(),
+        OpenIdConfigurationUrl: this.getOpenIdConfigurationUrl(),
+      },
+      e2eTest: {
```
fwpushan, line 14: Do we need to provide this information here? I suggest, somewhere in test or testHelper

fwpushan (changes requested):
Overall structure and KeyCloak Auth related changes are good but my only concerns is there are few test related config inside main application sources.

sources/packages/api/src/types/config.ts
```diff
@@ -1,10 +1,20 @@
 export interface IConfig {
   auth: IAuthConfig;
+  e2eTest: Ie2eTest;
```
fwpushan, line 3: Can we move this into test or testHelpers?

sources/packages/api/src/types/config.ts
```diff
@@ -1,10 +1,20 @@
...
 
 export interface IAuthConfig {
   url: string;
   realm: string;
   clientId: string;
-  issuerUrl: string;
+  OpenIdConfigurationUrl: string;
+}
+
+export interface Ie2eTest {
```
fwpushan, line 13: Can we move all test related config into test or testHelpers ?

fwpushan (changes requested):

sources/packages/api/src/types/config.ts
```diff
@@ -6,5 +6,5 @@ export interface IAuthConfig {
   url: string;
   realm: string;
   clientId: string;
-  issuerUrl: string;
+  OpenIdConfigurationUrl: string;
```
fwpushan, line 9: why this property name starts with Capital letter?

fwpushan (commented):
Please check there is only a minor comment.

sources/packages/api/src/auth/roles.enum.ts
```diff
@@ -3,4 +3,5 @@
  */
 export enum Role {
   Student = "StudentRole",
+  DummyRole = "DummyRole",
```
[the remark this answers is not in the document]
andrewsignori-aot: It is removed!

sources/packages/api/src/services/config/config.service.ts
andrewsignori-aot: I removed it from the main source as we talked.

sources/packages/api/src/types/config.ts
andrewsignori-aot: Because it is wrong :D
I am gonna change to the proper case.

andrewsignori-aot: Fixed!

fwpushan (approved):

andrewsignori-aot:
Went through a peer review chat with @2011sandeepraj and got the approval to execute the merge.

sources/packages/api/src/types/config.ts
andrewsignori-aot: Done!

sources/packages/api/src/types/config.ts
andrewsignori-aot: Done!

PR #236, opened 2021-02-23
193AddedPD_VERIFIED column

2011sandeepraj:
Added pd_verified to the database

### DIFF
--- a/docs/howto/db-migration.md
+++ b/docs/howto/db-migration.md
@@ -13,11 +13,17 @@
    `queryRunner.query()` method inside up method.
    See e.g. in `SIMS\sources\packages\api\src\database\migrations\1614107103138-StudentPDVerified.ts\1614107103138-StudentPDVerified.ts`
 
-5. If applicable Add the sql that is intended to be run using
-   queryRunner.query() method inside down method. This will be run when database is being wiped out.
+5. Add the sql that is intended to be run using
+   queryRunner.query() method inside down method. This will be run when database is going to be rolled back and/or database cleanup.
    See e.g. in `SIMS\sources\packages\api\src\database\migrations\1614107103138-StudentPDVerified.ts\1614107103138-StudentPDVerified.ts`
 
-6. Run the command to clean out all the volumes(database)
+6. Test the migrations by running
+
+   ```bash
+   $ make local
+   ```
+
+   Migrations can be tested by recreating the database
 
    ```bash
    $ make local-clean
--- a/docs/howto/db-migration.md
+++ b/docs/howto/db-migration.md
@@ -1,3 +1,26 @@
 # Database Migration
 
-TODO: Add content here!
+1. Checkout code locally cd to `sources/packages/api`
+
+2. Run the command
+   ```bash
+   $ npx typeorm migration:create -n "addStudentPDColumn"
+   ```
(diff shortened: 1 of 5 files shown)


docs/howto/db-migration.md
```diff
@@ -1,3 +1,26 @@
...
+   $ npx typeorm migration:create -n "addStudentPDColumn"
+   ```
+3. The above command will generate a migration file containing method stubs for async up and async down methods under the following package
+   `sources/packages/api/src/database/migrations`
+
+4. Add the sql that is intended to be run using
+   `queryRunner.query()` method inside up method.
+   See e.g. in `SIMS\sources\packages\api\src\database\migrations\1614107103138-StudentPDVerified.ts\1614107103138-StudentPDVerified.ts`
+
+5. If applicable Add the sql that is intended to be run using
```
andrewsignori-aot, line 16: I would recommend removing the "If applicable". We should always add a rollback. Not adding would be a major exception scenario.

docs/howto/db-migration.md
```diff
@@ -1,3 +1,26 @@
...
+   queryRunner.query() method inside down method. This will be run when database is being wiped out.
```
andrewsignori-aot, line 17: The rollback runs when the migration is "rolled back", not necessarily during a database cleanup.

docs/howto/db-migration.md
```diff
@@ -1,3 +1,26 @@
...
+   See e.g. in `SIMS\sources\packages\api\src\database\migrations\1614107103138-StudentPDVerified.ts\1614107103138-StudentPDVerified.ts`
+
+6. Run the command to clean out all the volumes(database)
+
+   ```bash
+   $ make local-clean
```
andrewsignori-aot, line 23: That is only necessary if we want to test all the migrations at once. To test the particular one, we can just run the application with the "make-local".

sources/packages/api/src-sql/StudentPDVerified/Add-StudentPDVerified.sql
```diff
@@ -0,0 +1,13 @@
+
+-- Add the column with pd_verified.
+ALTER TABLE students
+    ADD COLUMN IF NOT EXISTS pd_verified BOOLEAN DEFAULT ;
+   
+-- -- Removing the dummy constraint created just to allow us
```
andrewsignori-aot, line 6: Please remove the comment code.

sources/packages/api/src-sql/StudentPDVerified/Add-StudentPDVerified.sql
```diff
@@ -0,0 +1,13 @@
+
+-- Add the column with pd_verified.
+ALTER TABLE students
```
andrewsignori-aot, line 3: I believe that there is a problem with the syntax.
The correct command would be:
ALTER TABLE students
    ADD COLUMN IF NOT EXISTS pd_verified BOOLEAN;

andrewsignori-aot (changes requested):
Really nice working creating the documentation 👍 
Please take a look at the comments and ensure that the migration script is working as expected.

andrewsignori-aot (approved):
After a chat with @2011sandeepraj he updated the "up" script with the correct file and it looks good, hence approving this PR.

fwpushan (approved):

PR #241, opened 2021-02-24
SIMS# 162 - Student Provides Financial Information- Personal Income- Last Year Tax Return

andrewsignori-aot:
- Added the "2020 Tax Return Income" to the "Financial Information" page, as shown below.
- Used the currency component from PrimeVue.
- Created a small component to convert texts with line breaks in HTML paragraphs to avoiding adding HTML tags inside the current JSON text sources.

image


sources/packages/web/src/components/fa-application/ContentGroup.vue
```diff
@@ -0,0 +1,13 @@
+<template>
```
fwpushan, line 1: This component should be in global scope....

sources/packages/web/src/components/fa-application/StringToHtmlParagraphs.vue
```diff
@@ -0,0 +1,28 @@
+<template>
```
fwpushan, line 1: This component should be in global scope

sources/packages/web/src/views/FinancialInfo.vue
```diff
@@ -1,5 +1,64 @@
 <template>
-  <div class="p-component p-m-8">
-    This is the Financial Information page!
+  <div class="p-component">
+    <div class="p-card p-m-4">
+      <Section title="More questions to come..." />
+      <ContentGroup>
+        <Section :title="questions.previousYearTaxReturn.title">
+          <template #sub-title>
+            <StringToHtmlParagraphs
```
fwpushan, line 8: Nice

fwpushan (approved):
👍

sources/packages/web/src/components/fa-application/StringToHtmlParagraphs.vue
andrewsignori-aot: I will move it. I was not sure it would be good to promote it as a global right now, but it is a quick change and I will do it.

sources/packages/web/src/components/fa-application/ContentGroup.vue
andrewsignori-aot: Moved!

sources/packages/web/src/components/fa-application/StringToHtmlParagraphs.vue
andrewsignori-aot: Done!

sources/packages/web/src/views/FinancialInfo.vue
```diff
@@ -1,5 +1,64 @@
...
+import Section from "../components/fa-application/Section.vue";
+import Question from "../components/fa-application/Question.vue";
+import ContentGroup from "../components/ContentGroup.vue";
+import StringToHtmlParagraphs from "../components/fa-application/StringToHtmlParagraphs.vue";
+
+const questions = {
+  previousYearTaxReturn: {
+    title: "2020 Tax Return Income",
+    subTitle:
+      "Enter your reported total income from line 15000 of your 2020 Income Tax Return. This income will be matched with Canada Revenue Agency records, which may affect your assessment of need and/or grant eligibility.\nIf you did not file a 2020 Income Tax Return, enter your total income from all sources both inside AND outside of Canada",
```
2011sandeepraj, line 39: Can this go into some constants file

sources/packages/web/src/components/fa-application/StringToHtmlParagraphs.vue
2011sandeepraj, line 1: I think you can add a test for this component.

2011sandeepraj (changes requested):
Some changes requested

sources/packages/web/src/views/FinancialInfo.vue
andrewsignori-aot: It can, but right now it is following the same pattern for the other similar pages.
We still need to have discussions about where these texts will be retrieved.
Are they going to be retrieved from the API? Are they are going to be changed from time to time?

2011sandeepraj: Understood. Thanks

sources/packages/web/src/components/fa-application/StringToHtmlParagraphs.vue
2011sandeepraj: about my review comment requesting that a test be added for the component you created – On 2nd thoughts, i think the logic inside the component is very simple so this may NOT need to be tested. So leave it to you.. If you want to add a test go ahead and will again REview the PR . If you dont, want to add the test you can consider the PR merged!

andrewsignori-aot: I agree that a test could be added here, it would be nice to keep adding as much as possible.
I created the StringToHtmlParagraphs.spec.ts and fixed one test that was failing because a component was moved.

2011sandeepraj (approved):
LGTM

PR #242, opened 2021-02-24
SIMS 190 -Part 1 - Refactoring interfaces in web into a common folder to avoid repeated uses across web code

2011sandeepraj:
Part 2 of this ticket will be submitted via  different PR


fwpushan (changes requested):
I think the root should be **types** and then you can put sub directory called contract (**types/contract**). Then we can put any non-contract type also in types folder.

andrewsignori-aot (commented):
Nice work having this organized.
Is it possible to normalize the interface names? Some of them have the suffix "Contract" and some don't.

andrewsignori-aot:
> Nice work having this organized.
> Is it possible to normalize the interface names? Some of them have the suffix "Contract" and some don't.

Nevermind about the "normalize" thing, it was my bad 👍

andrewsignori-aot (approved):
👍

fwpushan (approved):
👍

PR #243, opened 2021-02-24
[api][.docker][SIM-193] Create custom docker file for Postgres local and creating db using script

fwpushan:
1. Custom docker file for local postgres 
2. Custom shell script to create db in local postgres 
3. Create Archive DB service 
4.  Complete Service Class
5. Update test query and hook
6. Create e2e tests for archive db service
7. Setup local env for direct application run
8. Update tests to get env setup


sources/packages/api/src/services/archive-db/archive-db.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const config: PostgresConnectionOptions = require("../../../ormconfig");
+
+@Injectable()
+export class ArchiveDbService {
+  private _connection?: Connection;
+
+  private async init() {
+    try {
+      this._connection = await createConnection({
```
andrewsignori-aot, line 13: You are not setting the db name here, right?
Is it going to be dedicated to sfaz DB or generic for any other raw SQL query to any other possible db?

andrewsignori-aot (approved):
Hi Pushan, really nice work with the service and with setting up the archive DB locally. I believe that we missed that while planning this.

sources/packages/api/src/services/archive-db/archive-db.service.ts
fwpushan: This was wrong, I will fix this.

sources/packages/api/src/database/__tests__/student.model.spec.ts
```diff
@@ -7,6 +8,7 @@ import { StudentService } from "../../services";
 describe("Test student model", () => {
   let connection: Connection;
   beforeAll(async () => {
+    // console.dir(process.env);
```
andrewsignori-aot, line 11: Please remove the commented code before the merge.

sources/packages/api/src/services/archive-db/archive-db.service.spec.ts
```diff
@@ -0,0 +1,37 @@
+require("../../../env_setup");
+import { Test, TestingModule } from '@nestjs/testing';
+import * as dayjs from 'dayjs';
+import { ArchiveDbService } from './archive-db.service';
+
+describe('ArchiveDbService', () => {
```
andrewsignori-aot, line 6: The file has a mix of single quotes and double-quotes.  Can you please format it?

sources/packages/api/src/services/archive-db/archive-db.service.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  it('should be defined', () => {
+    expect(service).toBeDefined();
+  });
+
+  it('should init', async () => {
+    expect(service.connection).toBeDefined();
+  });
+
+  it('should get status', async () => {
+    const r: Array<{ permanent_disability_flg: string | null }> = await service.getIndividualPDStatus({ birthdate: dayjs("2000-03-11 00:00:00").toDate(), sin: '123456799' });
```
andrewsignori-aot, line 31: Can you please elaborate on why we need to introduce the package "dayjs" to format the date here?

sources/packages/api/src-sql/ArchiveDB/SetupDB.sql
```diff
@@ -0,0 +1,10 @@
+-- Individual table
+CREATE TABLE IF NOT EXISTS individual ( 
+  individual_idx serial primary key,
+  sin numeric(22,9),
+  permanent_disability_flg varchar(1),
+  date_of_birth timestamp without time zone
+);
+
+-- Insert Test Item
+INSERT INTO individual(sin,permanent_disability_flg,date_of_birth) VALUES (123456799.000000000,'Y','2000-03-11 00:00:00');
```
andrewsignori-aot, line 10: I would recommend that the test item should be inserted during the E2E test and removed once the E2E is done.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -49,6 +50,12 @@ export class StudentService extends RecordDataModelService<Student> {
       phone: otherInfo.phone,
     };
     student.user = user;
+
+    // Get PD status from Archive DB
+    const result: { permanent_disability_flg: string | null } = await this.archiveDB.getIndividualPDStatus(student);
```
andrewsignori-aot, line 55: I think that it is fine for now, but in the future, I would prefer to create interfaces for the returning objects from the ArchiveDB methods.

andrewsignori-aot (commented):
Really great work setting up the ArchiveDB in the local DEV env and also creating the ArchiveDB service. I left minor comments only but the works look really great 👍

sources/packages/api/src-sql/ArchiveDB/SetupDB.sql
fwpushan: This whole items are available for local env only. We can keep them

sources/packages/api/src/services/student/student.service.ts
fwpushan: I will create interface.

sources/packages/api/src/services/archive-db/archive-db.service.spec.ts
fwpushan: 1. This is to format day into yyyymmdd format, native Date object can do that but for dayjs it is much easy.
2. For date formatting we must need some utility, normally we use **moment**, dayjs is stripped out version for moment for example using traditional date object, we need to do something like this
```js
and it subtracts a day
    const options = {
      year: "numeric",
      month: "long",
      day: "numeric",
      timeZone: "UTC",
    };
    const formattedDate = new Date(date);
    return formattedDate.toLocaleDateString("en-CA", options);
    ```
with Moment or dayjs this whole things become single liner.

File: sources/packages/api/src-sql/ArchiveDB/SetupDB.sql (hunk 5 above)
Comment (fwpushan) [reply sources/packages/api/src-sql/ArchiveDB/SetupDB.sql:10]: For remote e2e testing I've no idea how to proceed.

Comment (andrewsignori-aot) [reply sources/packages/api/src-sql/ArchiveDB/SetupDB.sql:10]: In my option, if it belongs to the E2E test and if is just gonna be used for the E2E test it should be created during the E2E test and deleted once the E2E is done.
It would be:
1 - Insert the dummy record (using a raw SQL query) before the test;
2 - Execute the E2E test using the dummy record inserted in step 1;
3 - Delete the dummy record created in step 1;
It would work for the remote E2E also, right?

File: sources/packages/api/src/testHelpers/testDBSetup.ts [hunk 7]
```diff
@@ -2,11 +2,16 @@ import { createConnection, getConnection } from "typeorm";
...
 export async function setupDB() {
-  const connection = await createConnection({
-    ...config,
-    logging: ["error", "warn", "info"],
-  });
-  return connection;
+  try {
+    const connection = await createConnection({
+      ...config,
+      logging: ["error", "warn", "info"],
```
2011sandeepraj, line 8: Just a question Why are we setting logger here?

2011sandeepraj (approved):
LGTM

sources/packages/api/src/services/archive-db/archive-db.service.ts
```diff
@@ -0,0 +1,60 @@
...
+    try {
+      await this.init();
+      return this._connection.query(raw, parameters);
+    } catch (excp) {
+      console.error(`Unable to query in archive db for ${excp}`);
+      return null;
+    }
+  }
+
+  async getIndividualPDStatus(
```
andrewsignori-aot, line 51: Does it make sense to return an array here?
Would not be the case to return the unique record found or null?

sources/packages/api/src/services/archive-db/archive-db.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      birthdate: dayjs(testData[0].dob).toDate(),
+      sin: testData[0].sin,
+    });
+    expect(r).toBeDefined();
+    expect(r.length).toBeGreaterThan(0);
+    expect(r[0].disability).toBeDefined();
+    expect(r[0].disability).toEqual("Y");
+  });
+
+  it("should get status with permanentDisabilityFlag = null", async () => {
```
andrewsignori-aot, line 74: Do we have a test to check what happens when the user is not found on DB?
Maybe would be good to have one.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -49,6 +54,12 @@ export class StudentService extends RecordDataModelService<Student> {
       phone: otherInfo.phone,
     };
     student.user = user;
+
+    // Get PD status from Archive DB
+    const result: StudentLegacyData [] = await this.archiveDB.getIndividualPDStatus(student);
+    if (result[0].disability === "Y") {
```
andrewsignori-aot, line 60: Does it will gonna work if the user is not found on DB?

andrewsignori-aot (commented):
Do we have the proper scenarios to test for when the user is not in the Archive DB?

PR #244, opened 2021-02-25
Added ExceptionRaiser and ErrorHelper in Api Layer

2011sandeepraj:
Part 2 of refactoring ticket 190. Added an ExceptionRaiser and ErrorHelper.
Also used standard exception classes available in nest framework rather than using HttpException with codes.


andrewsignori-aot (approved):
👍

fwpushan (approved):
👍

PR #247, opened 2021-02-27
165 student provides financial information travel costs

2011sandeepraj:
image

Expected functionality:
If No is selected to "living situation" question, the travel question is shown. If a yes is selected, the section asking for cost of one return trip home is shown.

Reviewers please note a couple of points:
1. questions-financial.ts: I want to have all the questions here. the file FinancialInfo.vue is getting big and hard to navigate. If the devs agree, I can move all questions for questions-financial into this file.
2. RadioButtonList.ts: This now has a default yes/no array. So all places where "options:yesNoOptions" was found to be used has been removed


sources/packages/web/src/helpers/questions-financial.ts
```diff
@@ -0,0 +1,16 @@
+export const questionsFin = {
```
andrewsignori-aot, line 1: I do not see a problem moving the questions, but I would prefer that they are moved inside a specific folder for fa-application, and please move all others also.

sources/packages/web/src/views/FinancialInfo.vue
```diff
@@ -49,6 +49,48 @@
...
+
+              <ContentGroup
+                v-if="financialInfoState.travelCostsValue === 'yes'"
+              >
+                <Section>
+                  <Question :text="questionsFin.returnTripCosts.question">
+                    <InputNumber
+                      class="p-m-2"
+                      mode="currency"
+                      currency="CAD"
```
andrewsignori-aot, line 82: Please take a look to ensure that the currency symbol is "$". It seems to be the requirement.

sources/packages/web/src/views/FinancialInfo.vue
```diff
@@ -79,23 +130,21 @@ const questions = {
...
     HorizontalSeparator,
     StringToHtmlParagraphs,
+    RadioButtonList,
     ReadMoreContainer,
   },
   setup() {
     const financialInfoState = reactive({} as FinancialInfoState);
     return {
       questions,
+      questionsFin,
```
andrewsignori-aot, line 147: Please move all the questions to the new source.

andrewsignori-aot (changes requested):
Please check the comments.

sources/packages/web/src/helpers/questions-financial.ts
fwpushan: @andrewsignori-aot @2011sandeepraj 
1. I've created one [technical debt](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/188) to manage static texts of the application and questionnaire.
2.  We have one placeholder for design challenge on questionnaire model, that may change whole section.  So, I will recommend, keep this portion in view/component.

fwpushan (approved):
LGTM to me

sources/.docker/postgres/create_db.sh
```diff
@@ -18,4 +18,4 @@ if [ -n "$POSTGRES_MULTIPLE_DATABASES" ]; then
     create_user_and_database $db
   done
   echo "Multiple databases created"
-fi
+fi
```
andrewsignori-aot, line 21: Please avoid adding this to the commit if it is not necessary.

sources/packages/web/src/validators/ValidatorConfigUI.ts
```diff
@@ -6,5 +6,7 @@ export default function configValidationRules() {
   // A list of available rules to be used could be found on
   // https://vee-validate.logaretm.com/v4/guide/global-validators#available-rules
   defineRule("required", required);
+  console.log(`In validator Config UI `);
```
andrewsignori-aot, line 9: Please remove commented lines.

andrewsignori-aot (commented):
Nice work moving the questions to the constants.
Please take a look at the minor comments.

andrewsignori-aot (approved):
After fixing the minor comments it should be good to go 👍

sources/.docker/postgres/create_db.sh
2011sandeepraj: Unable to

PR #252, opened 2021-03-01
Avoiding auto convert to CRLF when commit

andrewsignori-aot:
Changing the line breaks for create_db.sh from CRLF to LF to fix an issue during local postgres db creation.


fwpushan (approved):
With new file my local deployment is working fine. 👍

andrewsignori-aot:
It does not seem to be needed.

Issue #171, opened 2021-02-09
Student Selects 'Next Page' on Financial Information Page.


mzmichellesmith:
**As a** *(Student)* **I want** *(continue on to the next page in my Application )* **So That** *(I can determine if I qualify for Financial Aid)

**Additional Context**
- enter text here
- enter text here

**Acceptance Criteria**
- [x] Given (A Student), When (Selects 'Next Page' on the Financial Information Application Page), Then (they are redirected to the Confirm Submission Page next).

PR #253, opened 2021-03-02
SIMS #171 - Student Selects 'Next Page' on Financial Information Page.

andrewsignori-aot:
- Created the place holder for the "Confirm Submission" page.
- Created a default component for the footer buttons to have them aligned better to the left and right, as shown in the below image. This component has just the intention to avoid code duplication along with the pages. I don't believe that this will be the final component that we are going to be using, hence I did not spend time on unit testing it.

image

image


fwpushan (approved):
👍

2011sandeepraj (approved):

PR #258, opened 2021-03-02
Modified network policy config to use KNP

abschwenker:
Changed to use KNP and changed NSP to be wide open, effectively disabling it. Changes made due to deprecation of Aporeto


andrewsignori-aot (approved):
👍

Issue #235, opened 2021-02-23
Create Common logging system for API


fwpushan:
**Describe the task**
- Create one simple single interface logging service and infrastructure for logging. (Storage is out of scope)
- As requested during the review of the [PR#96](https://github.com/bcgov/SIMS/pull/96) we are creating this task to look into a way to handle
Global Network failures like API call errors.

**Additional Context**
Log to Console
Not putting try catch

PR #259, opened 2021-03-03
[api][SIM-#235] Logger and global error handler

fwpushan:
---------------------------------------------------
1. Create LoggerService and Module
2. Create Decorator and interface to inject logger
4. Create Global error filter
5. Update classes with proper logger
6. Lint fixes


sources/packages/api/src/main.ts
```diff
@@ -1,20 +1,41 @@
...
+  // Getting application port
   const port = process.env.PORT || 3000;
 
+  // Enabling cors for non prod env
   if (process.env.NODE_ENV !== "production") {
     app.enableCors();
   }
 
+  // Setting express middleware for req
+  app.use(LoggerService.apiLogger);
```
andrewsignori-aot, line 37: Is the idea to log all requests even on PROD env?
Right now we have only conole.log as the final log, but depending on how would be the final log storage maybe we do not want to log every request every time.
Is it a placeholder to be expanded in the future or even be enabled/disabled upon configuration?

sources/packages/api/src/database/database.service.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  it("should be defined", () => {
+    expect(service).toBeDefined();
+  });
+
+  it("should create service", async () => {
+    const connection = getConnection();
+    const svc = new DatabaseService(connection);
+    expect(svc).toBeDefined();
+    expect(svc.logger()).toBeDefined();
```
andrewsignori-aot, line 27: Is it possible, using the current design, to replace the current logger implementation with a mocked object?
In the future, if our logger is saving to a file or a database, and I want to execute a unit test is some service that uses the logger, is it possible to have a mocked logger to test this service?

fwpushan: It is partial test of the concepts, but my idea is to mock it through LoggerService ....

fwpushan: Currently Nest logger is automatically disabled in test env

sources/packages/api/src/main.ts
fwpushan: YES, we need discussion on this topic. And I'm totally agree, logging each request could be overhead if persistent log storage introduce....

sources/packages/api/src/common/loggable.ts
```diff
@@ -0,0 +1,21 @@
+import { LoggerService } from "../logger/logger.service";
+
+export interface Loggable {
+  logger(): LoggerService;
```
andrewsignori-aot, line 4: Is it possible to define a property instead of a method for the interface?

2011sandeepraj (changes requested):
As discussed, 
Base out classes we can. 
Inject LoggerService instead of having classes implement Loggable

sources/packages/api/src/common/loggable.ts
fwpushan: DONE

sources/packages/api/src/database/data.model.service.ts
```diff
@@ -1,8 +1,10 @@
+import { LoggerService } from "../logger/logger.service";
```
andrewsignori-aot, line 1: Can you remove it, please?

sources/packages/api/src/services/auth/keycloak/keycloak.service.ts
```diff
@@ -49,7 +54,7 @@ export class KeycloakService {
       return response.data as OpenIdConfig;
     } catch (ex) {
       // TODO: Add a logger.
```
andrewsignori-aot, line 56: I believe that we can remove the // TODO: Add a logger.

andrewsignori-aot (approved):
👍

Issue #255, opened 2021-03-02
Change Student Authentication


mzmichellesmith:
Technical Tasks 
--------------------------------------------------

- [x] Reorganize Folder Structure (#254 )
- [ ] Refactor current auth service to handle multiple persona. 
- [ ] Develop service and code for institution auth

PR #273, opened 2021-03-08
[api][web][SIMS-#255] Change Student Auth Flow

fwpushan:
-------------------
1. Update API to fetch multiple KC Clients
2. Update web to init AuthService on StudentApp root
3. Update web app to manage multiple clients
4. Fix any student path load issue

### DIFF
--- a/sources/packages/api/src/services/config/config.service.ts
+++ b/sources/packages/api/src/services/config/config.service.ts
@@ -8,7 +8,10 @@ export class ConfigService {
       auth: {
         url: process.env.KEYCLOAK_AUTH_URL,
         realm: process.env.KEYCLOAK_REALM,
-        clientId: process.env.KEYCLOAK_CLIENT,
+        clientIds: {
+          student: process.env.KEYCLOAK_CLIENT,
+          institute: "",
+        },
         openIdConfigurationUrl: this.getOpenIdConfigurationUrl(),
       },
     };
--- a/sources/packages/web/src/router/StudentRoutes.ts
+++ b/sources/packages/web/src/router/StudentRoutes.ts
@@ -9,6 +9,7 @@ import SelectProgram from "../views/student/financial-aid-application/SelectProg
 import FinancialInfo from "../views/student/financial-aid-application/FinancialInfo.vue";
 import ConfirmSubmission from "../views/student/financial-aid-application/ConfirmSubmission.vue";
 import { StudentRoutesConst } from "../constants/routes/RouteConstants";
+import { AppConfigService } from "@/services/AppConfigService";
 
 export const studentRoutes: Array<RouteRecordRaw> = [
   {
@@ -25,6 +26,15 @@ export const studentRoutes: Array<RouteRecordRaw> = [
         path: "login",
         name: StudentRoutesConst.LOGIN,
         component: Login,
+        beforeEnter: (to, from, next) => {
+          if (!AppConfigService.shared.authService?.authenticated) {
+            next()
+          } else {
+            next({
+              name: StudentRoutesConst.STUDENT_DASHBOARD
+            })
+          }
+        }
       },
       {
         path: "student-profile",
--- a/sources/packages/web/src/router/StudentRoutes.ts
+++ b/sources/packages/web/src/router/StudentRoutes.ts
@@ -27,14 +27,25 @@ export const studentRoutes: Array<RouteRecordRaw> = [
         name: StudentRoutesConst.LOGIN,
         component: Login,
         beforeEnter: (to, from, next) => {
-          if (!AppConfigService.shared.authService?.authenticated) {
-            next()
+          // Check Auth service is available or not
+          if (AppConfigService.shared.authService) {
+            const auth =
+              AppConfigService.shared.authService?.authenticated || false;
+            if (!auth) {
+              // Allowing to load login iff when user is not authenticated
+              next();
+            } else {
+              next({
+                name: StudentRoutesConst.STUDENT_DASHBOARD,
+              });
+            }
           } else {
+            // Auth service is not initialize so loading to student root
             next({
-              name: StudentRoutesConst.STUDENT_DASHBOARD
-            })
+              name: StudentRoutesConst.APP_STUDENT,
+            });
           }
-        }
+        },
       },
       {
         path: "student-profile",
--- a/sources/packages/web/src/types/contracts/ConfigContract.ts
+++ b/sources/packages/web/src/types/contracts/ConfigContract.ts
@@ -2,12 +2,15 @@ export interface GetConfig {
   auth: AuthConfig;
 }
 
-export type ClientIdType = "student" | "institute";
+export enum ClientIdType {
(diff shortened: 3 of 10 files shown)


sources/packages/web/src/views/student/AppStudent.vue
```diff
@@ -18,25 +18,43 @@ import { StudentService } from "../../services/StudentService";
...
-        name: "StudentDashboard",
-      });
     } else {
-      /* User doesn't exist in SABC Database and so redirect the user to Student Profile page
+      // TODO:
+      // - Try to implement a role based processing
+      // Get path
+      const path = route.path;
+      console.log("Not ok if called before KC");
+      if (path.includes("/student")) {
```
andrewsignori-aot, line 38: Since the AppStudent component is loaded only under "/student" path, is there any specific reason to check if the current path includes "/student"?

sources/packages/web/src/views/student/AppStudent.vue
```diff
@@ -18,25 +18,43 @@ import { StudentService } from "../../services/StudentService";
   },
 })
 export default class AppStudent extends Vue {
```
andrewsignori-aot, line 20: Consider refactor this component to be a "function component" instead of a class component to be aligned with the others along with this application. I am suggesting this in this PR because it is receiving more logic and it seems to be a good time to do it.

sources/packages/web/src/views/student/AppStudent.vue
```diff
@@ -18,25 +18,43 @@ import { StudentService } from "../../services/StudentService";
...
+      // Get path
+      const path = route.path;
+      console.log("Not ok if called before KC");
+      if (path.includes("/student")) {
+        if (await UserService.shared.checkUser()) {
+          if (path.includes("/student")) {
+            await StudentService.shared.synchronizeFromUserInfo();
+
+            // Not allowing to load login page
+            if (path.includes("login")) {
```
andrewsignori-aot, line 44: I would recommend do this type of check inside the beforeEach event on router.

sources/packages/web/src/types/contracts/ConfigContract.ts
```diff
@@ -2,17 +2,15 @@ export interface GetConfig {
   auth: AuthConfig;
 }
 
+export type ClientIdType = "student" | "institute";
```
andrewsignori-aot, line 5: Please consider changing it to a string enum.

sources/packages/api/src/services/config/config.service.ts
```diff
@@ -8,7 +8,10 @@ export class ConfigService {
       auth: {
         url: process.env.KEYCLOAK_AUTH_URL,
         realm: process.env.KEYCLOAK_REALM,
-        clientId: process.env.KEYCLOAK_CLIENT,
+        clientIds: {
+          student: process.env.KEYCLOAK_CLIENT,
```
andrewsignori-aot, line 12: Please consider renaming ENV variable KEYCLOAK_CLIENT to be more student explicit.

andrewsignori-aot (commented):

sources/packages/web/src/views/student/AppStudent.vue
fwpushan: Done

sources/packages/web/src/views/student/AppStudent.vue
fwpushan: DONE

sources/packages/api/src/services/config/config.service.ts
fwpushan: Will do when we introduce new env variable

sources/packages/web/src/types/contracts/ConfigContract.ts
fwpushan: DONE

sources/packages/web/src/views/student/AppStudent.vue
fwpushan: DONE

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -9,6 +9,7 @@ import SelectProgram from "../views/student/financial-aid-application/SelectProg
 import FinancialInfo from "../views/student/financial-aid-application/FinancialInfo.vue";
 import ConfirmSubmission from "../views/student/financial-aid-application/ConfirmSubmission.vue";
 import { StudentRoutesConst } from "../constants/routes/RouteConstants";
+import { AppConfigService } from "@/services/AppConfigService";
```
2011sandeepraj, line 12: Use of @  - i know its a vue3 thing but not a standard for us. And it may some day interfere with jest

2011sandeepraj (commented):
Very minor comments. Have a look and consider the PR approved.

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -25,6 +26,15 @@ export const studentRoutes: Array<RouteRecordRaw> = [
         path: "login",
         name: StudentRoutesConst.LOGIN,
         component: Login,
+        beforeEnter: (to, from, next) => {
+          if (!AppConfigService.shared.authService?.authenticated) {
```
andrewsignori-aot, line 30: It could be ok for now, but what I have been mentioning is that we should move to a model like below.
image

sources/packages/web/src/views/student/AppStudent.vue
andrewsignori-aot: I am still seeing it as a class component "export default class AppStudent extends Vue".
Am I missing something?

sources/packages/web/src/views/student/AppStudent.vue
```diff
@@ -12,32 +12,43 @@ import { AppConfigService } from "../../services/AppConfigService";
...
-        name: StudentRoutesConst.STUDENT_DASHBOARD,
-      });
     } else {
-      /* User doesn't exist in SABC Database and so redirect the user to Student Profile page
+      // TODO:
+      // - Try to implement a role based processing
+      // Get path
+      if (await UserService.shared.checkUser()) {
+        await StudentService.shared.synchronizeFromUserInfo();
+        if (route.path === "/student") {
```
andrewsignori-aot, line 40: Is there a reason why we need to check if the path contains "/student" here?
This component will be loaded only under "/student" path, right?

andrewsignori-aot (approved):
Please take a look at the comments. The places where I pointed that some modifications would be recommended are likely to receive new changes soon, hence I am ok if you want to proceed with the merge.

sources/packages/web/src/router/StudentRoutes.ts
fwpushan: This is going beyond login page, directly to KeyCloak? Do we really need that?

sources/packages/web/src/views/student/AppStudent.vue
fwpushan: Sorry, I missed this part.

Issue #172, opened 2021-02-09
Institution User can View Read Only BCeID Information on account set up


mzmichellesmith:
**As a** *(Institution User)* **I want** *( want to view the Read-Only Business BCeid information, So That *(I can confidently ensure that SABC has the correct information.

**Additional Context**
- The information from Business BCeID includes Given Names, Last Name, and Institution
- When Business BCeID information is updated, then the information will be updated on the SABC account).- Another User Story
-The information from Business BCeID is viewed as read-only.

**Acceptance Criteria**
- [x] (The Institution User), When (has been redirected to the SABC Site after authentication, Then they can view the information that has been received from the Business BCeID site (First Name, Last Name, and Legal Institution Name (NOT GUIDs)
- [x] Given (The  Institution User has been redirected to the SABC Site ), When (they are viewing the Business BCeID information) Then, this information is there but it greyed out and may not be updated.

Technical Tasks 
--------------------------------------------------

- [ ] Create webservice call to retrieve additional info (SOAP Client)
- [ ] Display required info
- [ ] ...
- [ ] ...
- [ ] ...
- [ ] ....

Reference Docs
--------------------------------------------------
https://github.com/bcgov/ocp-sso/wiki/3.2-Identity-Provider-Attribute-Mapping

andrewsignori-aot:
Please find below some documentation for BCeID Attribute Mapping

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/8fe2b927-6ce2-4278-aa11-4f8572c17877)
Source: https://github.com/bcgov/ocp-sso/wiki/3.2-Identity-Provider-Attribute-Mapping

PR #287, opened 2021-03-10
[OpenShift][api][SIMS #172]BCeID Soap Call to get account details

andrewsignori-aot:
Partial implementation of the SIMS #172

Implemented
- BCeID Web Service SOAP request to get BCeID Information (first name, surname, email, institution legal name, etc.);
- E2E Test to test the BCeID Soap call considering the "happy path";
- Created an OpenShift secret file to store sensitive information for any API related secret;
- Create a new API endpoint "/api/users/bceid-account" that returns the BCeID account information for the currently logged user.
To be implemented
- More E2E tests to test some some "not so happy path" scenarios;
- Check if the SOAP call can be done with the "user guid" instead of "user name" (under investigation);
- Check possible performance improvements on soap client creation;
- Documentation on git with an overview about BCeID Web Service usage;


sources/packages/api/src/services/bceid/bceid.service.ts
```diff
@@ -0,0 +1,148 @@
...
+  public async getAccountDetails(userName: string): Promise<AccountDetails> {
+    var client = await this.getSoapClient();
+    // SOAP call body to execute the getAccountDetail request.
+    var body = {
+      accountDetailRequest: {
+        onlineServiceId: this.bceidConfig.onlineServiceId,
+        // Internal indicate that the user being provided
+        // on requesterUserGuid is an user that belongs
+        // to the internal gov network.
+        requesterAccountTypeCode: "Internal",
```
fwpushan, line 40: Can we use some constant here?

fwpushan (approved):
Only few minor suggestions. 👍 . Nice work

sources/packages/api/src/services/bceid/bceid.service.ts
andrewsignori-aot: Sure, I will change it!

andrewsignori-aot: Done!

configs/env-example
```diff
@@ -20,5 +20,21 @@ KEYCLOAK_REALM=
 KEYCLOAK_CLIENT=
 
 # E2E Test
-E2E_TEST_STUDENT_PASSWORD=ADD_THE_PASSWORD_HERE_TO_RUN_LOCALLY
 E2E_TEST_STUDENT_USERNAME=student_e2e_test
+E2E_TEST_STUDENT_PASSWORD=ADD_THE_PASSWORD_HERE_TO_RUN_LOCALLY
+
+E2E_TEST_BCeID_USERNAME=SIMS_E2ETests
+E2E_TEST_BCeID_PASSWORD=ADD_THE_PASSWORD_HERE_TO_RUN_LOCALLY
```
2011sandeepraj, line 27: As we hav called line 24 as STUDENT_PASSWORD, i think we shold call this  INSTITUTION_PASSWORD. Same for Username

devops/Makefile
```diff
@@ -80,6 +79,14 @@ build-patroni:
 	@oc -n $(BUILD_NAMESPACE) process -f $(BUILD_TEMPLATE_PATH)  -p NAME=patroni -p TAG="12-latest" -p BASE_IMAGE_NAME=postgres -p BASE_IMAGE_TAG="12" -p SOURCE_REPOSITORY_URL="https://github.com/BCDevOps/platform-services.git" -p SOURCE_REPOSITORY_REF="master" -p SOURCE_CONTEXT_DIR="apps/pgsql/patroni/docker" | oc -n $(BUILD_NAMESPACE) apply -f -
 	@oc -n $(BUILD_NAMESPACE) start-build bc/patroni --wait
 
+init-api:
+	@echo "++\n Init Api in $(NAMESPACE) \n++"
+	test -n $(API_INIT_BCeID_WEB_SERVICE_ONLINE_SERVICE_ID)
+	test -n $(API_INIT_BCeID_WEB_SERVICE_REQUESTER_USER_GUID)
+	test -n $(API_INIT_BCeID_WEB_SERVICE_AUTH_USER_NAME)
+	test -n $(API_INIT_REQ_BCeID_WEB_SERVICE_AUTH_USER_PASSWORD)
+	@oc -n $(NAMESPACE) process -f openshift/api-pre-req.yml -p NAME=api -p BCeID_WEB_SERVICE_ONLINE_SERVICE_ID=$(API_INIT_BCeID_WEB_SERVICE_ONLINE_SERVICE_ID) -p BCeID_WEB_SERVICE_REQUESTER_USER_GUID=$(API_INIT_BCeID_WEB_SERVICE_REQUESTER_USER_GUID) -p BCeID_WEB_SERVICE_AUTH_USER_NAME=$(API_INIT_BCeID_WEB_SERVICE_AUTH_USER_NAME) -p BCeID_WEB_SERVICE_AUTH_USER_PASSWORD=$(API_INIT_REQ_BCeID_WEB_SERVICE_AUTH_USER_PASSWORD) | oc -n $(NAMESPACE) apply -f -
```
2011sandeepraj, line 88: Do we need to provide the oc -n ${namespace} twice? Once at the start once at the end?

sources/docker-compose.yml
```diff
@@ -24,6 +24,13 @@ services:
       - ENVIRONMENT=local
       - E2E_TEST_STUDENT_USERNAME=${E2E_TEST_STUDENT_USERNAME}
       - E2E_TEST_STUDENT_PASSWORD=${E2E_TEST_STUDENT_PASSWORD}
+      - E2E_TEST_BCeID_USERNAME=${E2E_TEST_BCeID_USERNAME}
```
2011sandeepraj, line 27: Same comment as previous ones, on the username and password . Can we have it like some mention of INSTITUTION there

2011sandeepraj: Or INSTITUTIONUSER

sources/packages/api/src/services/bceid/account-details.model.ts
```diff
@@ -0,0 +1,17 @@
...
+  displayName: string;
+  firstname: string;
+  surname: string;
+  email: string;
+}
+
+export interface InstitutionDetails {
+  guid: string;
+  legalName: string;
+}
```
2011sandeepraj, line 17: Confused between this and bceid-account.model.ts. Lets talk about it

sources/packages/api/src/services/bceid/bceid.service.ts
```diff
@@ -0,0 +1,152 @@
...
+  private bceidConfig: BCeIDConfig;
+  constructor(private readonly config: ConfigService) {
+    this.bceidConfig = this.config.getConfig().bceid;
+  }
+
+  /**
+   * This method allows you to retrieve the details of a single BCeID account.
+   * This method respects the information sharing rules. To perform a query, there must be an
+   * information sharing rule for the online service (onlineServiceId) with the specified requester
+   * user account type (requesterAccountTypeCode) and target user account type (accountTypeCode).
```
2011sandeepraj, line 30: i'd put this as 
"user account type (requesterAccountTypeCode - Business/Internal)"
same for target user account type

sources/packages/api/src/services/bceid/bceid.service.ts
```diff
@@ -0,0 +1,152 @@
...
+        // on requesterUserGuid is an user that belongs
+        // to the internal gov network.
+        requesterAccountTypeCode: BCeIDAccountTypeCodes.Internal,
+        // The user guid of the user on the internal gov network.
+        requesterUserGuid: this.bceidConfig.requesterUserGuid,
+        userId: userName,
+        // Type of the user account to search for the userId
+        // parameter provided above. Only business BCeID are
+        // on the scope of the application.
+        accountTypeCode: BCeIDAccountTypeCodes.Business,
```
2011sandeepraj, line 51: Am confused between accountTypeCode and Request AcountTypeCode. Lets talk on this

sources/packages/api/src/services/config/config.service.ts
```diff
@@ -11,6 +11,15 @@ export class ConfigService {
         clientId: process.env.KEYCLOAK_CLIENT,
         openIdConfigurationUrl: this.getOpenIdConfigurationUrl(),
       },
+      bceid: {
+        wsdlEndpoint: process.env.BCeID_WEB_SERVICE_WSDL,
+        onlineServiceId: process.env.BCeID_WEB_SERVICE_ONLINE_SERVICE_ID,
+        requesterUserGuid: process.env.BCeID_WEB_SERVICE_REQUESTER_USER_GUID,
+        credential: {
+          userName: process.env.BCeID_WEB_SERVICE_AUTH_USER_NAME,
+          password: process.env.BCeID_WEB_SERVICE_AUTH_USER_PASSWORD,
```
2011sandeepraj, line 20: Need to understand this

sources/packages/api/src/types/config.ts
```diff
@@ -8,3 +9,41 @@ export interface IAuthConfig {
   clientId: string;
   openIdConfigurationUrl: string;
 }
+
+/**
+ * BCeID configuration required to execute the SOAP requests
+ * to the BCeID Web Service.
+ */
+export interface BCeIDConfig {
```
2011sandeepraj, line 17: Rename to IBCeIDConfig

sources/packages/api/test/users.e2e-spec.ts
```diff
@@ -0,0 +1,62 @@
+require("../env_setup");
```
2011sandeepraj, line 1: Should we have anotherfolder under test to segragate tests for institution users and students?

sources/packages/api/src/utilities/auth-utils.spec.ts
```diff
@@ -0,0 +1,22 @@
+// Target module
```
2011sandeepraj, line 1: Keep inside test folder?

2011sandeepraj (changes requested):
Requesting Changes and Discussions

configs/env-example
andrewsignori-aot: This user/password is not related to an institution. This user could be used to any other E2E test that requires a valid BCeID User.
Under this scenario, do you still believe that we should change it? If yes, I would avoid adding "institution" to it.

devops/Makefile
andrewsignori-aot: To invoke the make command it will be provided only once (e.g. "make init-api NAMESPACE=<some_namespace>").
Here we have two OC commands being executed (process/apply) and both are providing the NAMESPACE parameter, following the same that others in the same makefile are doing.

sources/docker-compose.yml
andrewsignori-aot: This user/password is not related to an institution. This user could be used to any other E2E test that requires a valid BCeID User.
Under this scenario, do you still believe that we should change it? If yes, I would avoid adding "institution" to it.

sources/packages/api/src/services/bceid/bceid.models.ts
```diff
@@ -0,0 +1,107 @@
+/**
```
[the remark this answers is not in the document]
andrewsignori-aot: It is supposed to contain any model not only the ones for responses. For now, it is in the majority for response models, but it also has the BCeIDAccountTypeCodes, for instance, that is used for requests.

sources/packages/api/src/services/bceid/bceid.service.ts
andrewsignori-aot: Both variables are using an Enum that makes it clear what are the possible values to be used. The enum items also have a comment on each to explain the idea behind it. I think that move the option from the enum and adding it into the parenthesis is not needed in this case.

sources/packages/api/src/services/bceid/bceid.service.ts
andrewsignori-aot: I believe that you are talking about requesterAccountTypeCode and accountTypeCode, right?
- requesterAccountTypeCode: identification from the user doing the request. It will be used by BCeID Web Service to check permissions to access data.
- accountTypeCode: the type of the user account that we provided in the parameter userId.

sources/packages/api/src/types/config.ts
andrewsignori-aot: We do not have a pattern in place to dictate this and so far it seems that the majority of the interface is not using the "I".

sources/packages/api/test/users.e2e-spec.ts
andrewsignori-aot: This method is not institution-specific. It is under the same user endpoint already in use for students.
We can discuss how to organize on folder based on the endpoint, maybe.

sources/packages/api/src/utilities/auth-utils.spec.ts
andrewsignori-aot: The unit tests for the API are alongside the files that they are testing. We can change it, but it should be a change for all others also. Let's talk about it.

sources/packages/api/src/services/bceid/account-details.model.ts
andrewsignori-aot: Renamed to dto.

andrewsignori-aot:
Worked with @2011sandeepraj in a meeting to discuss the questions and comments.
He provided his approval for the PR during the meeting.

Issue #221, opened 2021-02-22
Institution User Authenticates with Business BCeID


mzmichellesmith:
**As a** *(Institution User)* **I want** *(Authenticate my Identity with Business BCeID)* **So That** *(ensure only the right people have access to my Institution Account)*

**Additional Context**
-SABC requires that an Institution Users identification is validated against government-issued ID and authentication. The current secure method for this Business BCeID.
-All attributes from BCeID are read-only and the user must go to Business BCeID to change.
-If the Institution User does not have a Business BCeID or they are having issues with the Business BCeID the Institution User will need to contact Business BCeID to get one.
Information from Business BCeID will be read-only on the SABC Account.
The information from Business BCeID includes Given Names, Last Name, User GUID, Institution Legal Name, Institution GUID.
When Business BCeID information is updated, then the information will be updated on the SABC account).- Another User Story

**Acceptance Criteria**
- [x] Given (an Institution User), When (has successfully authenticated with Business BCeID ), Then (site will redirect the Institution User back to the SABC site to complete their SABC Institution Profile.
- [x] Given the Institution User does not have a Business BCeID, Then (the Institution User will be directed to the Business BCeID site to activate their BCeID and their SABC account is NOT created until their Business BCeID is authenticated.
 - [x] Given The Institution Users authentication is NOT successful, then (the Institution user will remain on the Business BCeID site until it is authenticated.

Technical Reference:
https://github.com/bcgov/ocp-sso/wiki/3.2-Identity-Provider-Attribute-Mapping

Technical Tasks 
--------------------------------------------------

- [ ] Create Client
- [ ] Create Client Scope
- [ ] Create Attributes
- [ ] ONLY use minimal attributes to get user back to application
- [ ] Only the feilds that are already available
- [ ] ....

PR #307, opened 2021-03-12
[web][SIMS #221] Institution Login Flow

fwpushan:
--------
1. Connect institution login flow component
2. Create and implement 403 page
3. Add placeholder WIP page
4. Some auto lint fixes
----


sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -1,25 +1,68 @@
...
         name: InstitutionRoutesConst.INSTITUTION_DASHBOARD,
         component: InstitutionDashboard,
       },
+      {
+        path: AppRoutes.InstitutionProfile,
+        name: InstitutionRoutesConst.INSTITUTION_PROFILE,
+        component: WorkInProgress,
+      },
     ],
+    beforeEnter: (to, from, next) => {
```
andrewsignori-aot, line 36: I would consider two things for the "beforeEnter" method.
1 - Move the method definition outside the array for better readability.
2 - Since we have a RouterHelper, maybe this logic could be moved there.

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -1,25 +1,68 @@
...
+              next({
+                name: InstitutionRoutesConst.LOGIN,
+              });
+              break;
+            case AuthStatus.RedirectHome:
+              next({
+                name: InstitutionRoutesConst.INSTITUTION_DASHBOARD,
+              });
+              break;
+            case AuthStatus.ForbiddenUser:
```
andrewsignori-aot, line 58: Maybe would be good to have a default option in this switch.
Would it be good to maybe consider changing the default one to be the SharedRouteConst.FORBIDDEN_USER.

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -1,25 +1,68 @@
...
+              next({
+                name: SharedRouteConst.FORBIDDEN_USER,
+              });
+              break;
+          }
+        })
+        .catch(e => console.error(e));
```
andrewsignori-aot, line 65: I believe that we should throw the exception here and let the application be broken. In the future, we will probably have a generic error page.
My point is that this code must run with success, if not, we should not silence the exception.

sources/packages/web/src/views/ForbiddenUser.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <h1>Forbidden User 403</h1>
+  <h2>Please logout and try with different user type</h2>
```
2011sandeepraj, line 3: May be a good thing to have a link to logout button here.

2011sandeepraj (approved):
LGTM

sources/packages/web/src/services/AppConfigService.ts
```diff
@@ -75,10 +82,75 @@ export class AppConfigService {
...
+    }
     if (this._config) {
+      this._authClientType = type;
       this.authService = await AuthService(this._config, type);
     } else {
       throw new Error("Unable to load application: server is not responding");
     }
   }
+
+  roles(): string[] {
```
andrewsignori-aot, line 96: If we are not using this code, please consider removing it.

sources/packages/web/src/services/AppConfigService.ts
```diff
@@ -75,10 +82,75 @@ export class AppConfigService {
...
+        // 1. Role based solution is not working
+        // 2. For Logged in user KeyCloak clients are acting same
+        // 3. Add some more info on token to identify user type or userName for BCeID users
+        // 4. Now Just implementing one temporary work around
+        let validUser = false;
+        if (this.authService?.tokenParsed) {
+          const token = this.authService?.tokenParsed as any;
+          switch (options.type) {
+            case ClientIdType.INSTITUTION: {
+              if (token["bceid_user_name"]) {
```
andrewsignori-aot, line 125: @abschwenker created the below property on the token that could be used instead "bceid_user_name".

image

andrewsignori-aot (changes requested):
I make it explicit in some places that they are only recommendations and not blockers for PR approval.
Please replace the token property "bceid_user_name" with the new one "IDP" just added by @abschwenker, otherwise it will not work as expected.

sources/packages/web/src/router/InstitutionRoutes.ts
fwpushan: I will suggest Route Helper to handle this but as refactor of code with better access management.

sources/packages/web/src/views/ForbiddenUser.vue
fwpushan: This is placeholder for now, I think we must design it properly in future.

sources/packages/web/src/services/AppConfigService.ts
fwpushan: DONE

sources/packages/web/src/services/AppConfigService.ts
fwpushan: DONE

sources/packages/web/src/router/InstitutionRoutes.ts
fwpushan: DONE

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -86,5 +91,36 @@ export const studentRoutes: Array<RouteRecordRaw> = [
...
+              next({
+                name: StudentRoutesConst.LOGIN,
+              });
+              break;
+            case AuthStatus.RedirectHome:
+              next({
+                name: StudentRoutesConst.STUDENT_DASHBOARD,
+              });
+              break;
+            case AuthStatus.ForbiddenUser:
```
andrewsignori-aot, line 116: Please change in the same way as you did for Institution.

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -86,5 +91,36 @@ export const studentRoutes: Array<RouteRecordRaw> = [
...
+              next({
+                name: SharedRouteConst.FORBIDDEN_USER,
+              });
+              break;
+          }
+        })
+        .catch(e => console.error(e));
```
andrewsignori-aot, line 123: Please change in the same way as you did for Institution.

PR #310, opened 2021-03-12
[web][SIMS #172]Navbar to Institution Portal

andrewsignori-aot:
Added Navbar to Institution Portal, as shown below.

image


2011sandeepraj (approved):
LGTM

fwpushan (approved):
👍 . One small recommendation, Can we create profile button as reusable component, because that will I've similar user for institution.

Issue #231, opened 2021-02-23
Enable Unit and E2E Tests on CI/CD


andrewsignori-aot:
We already have some unit tests and E2E tests on the application and we need to configure than to run alongside our CI/CD pipeline.

Cypress is out of scope

PR #313, opened 2021-03-15
[api][SIMS #231] Enable Tests on Github Action CI/CD

fwpushan:
-----
1. Create new make rule to run api test on docker env
2. Update tests accordingly
3. Add new github test workflow
----
N.B. For Reviewers 
-------
**Currently most of the controllers and services required database connection, so to avoid multiple connection and speed up the test execution, change subject initialisation beforEach to beforAll**


.github/workflows/test.yml
```diff
@@ -0,0 +1,93 @@
...
+          echo Git Build ID: ${{ github.event.number }}
+          echo Git Pull Request Ref: ${{ github.event.pull_request.head.sha }}
+  test:
+    name: Test API in Docker env
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.merged == false }}
+    env:
+      PROJECT_NAME: aest-sims
+      BUILD_ID: ${{ github.event.number }}
+      POSTGRES_PASSWORD: bcgov!
```
andrewsignori-aot, line 25: Even that this password is used only for the test DB, consider moving to a secret.

.github/workflows/test.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/checkout@v1
+      - name: Docker Pull
+        working-directory: "./sources/"
+        run: docker-compose pull
+      # In this step, this action saves a list of existing images,
+      # the cache is created without them in the post run.
+      # It also restores the cache if it exists.
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        # Ignore the failure of a step and avoid terminating the job.
+        continue-on-error: true
```
andrewsignori-aot, line 59: Could you please elaborate on why we should ignore the error here?

.github/workflows/test.yml
```diff
@@ -0,0 +1,93 @@
...
+      # Deploy the api
+      - name: Build and run API in Docker
+        working-directory: "./sources/"
+        run: |
+          make local-api-detached
+      - name: Sleep for 10 seconds
```
andrewsignori-aot, line 65: Could you please elaborate on why we should wait for 10s here?

.github/workflows/test.yml
fwpushan: Same info is already exposed in sources/make file as default value .....

.github/workflows/test.yml
fwpushan: This is docker caching, this is to speedup docker build process, if that fails, still our tests run but with some extra time.

.github/workflows/test.yml
fwpushan: Currently, booting up local db host taking sometime in Github action machine adding a little bit wait resolving that.

sources/packages/api/src/services/auth/keycloak/keycloak.service.ts
```diff
@@ -110,6 +110,7 @@ export class KeycloakService {
     } catch (ex) {
       // TODO: Add a logger.
       this.logger.error(ex);
+      console.error(ex);
```
andrewsignori-aot, line 113: I believe that you can remove it now.

2011sandeepraj (commented):
LGTM

sources/packages/api/test/auth.e2e-spec.ts
```diff
@@ -11,6 +11,8 @@ import {
 import { AuthTestController } from "../src/testHelpers/controllers/auth-test/auth-test.controller";
 import { KeycloakService } from "../src/services/auth/keycloak/keycloak.service";
 
+jest.setTimeout(10000);
```
andrewsignori-aot, line 14: Was it really necessary or was it more a local issue?

andrewsignori-aot (approved):
👍

PR #316, opened 2021-03-16
[api][SIMS#217] institution user completes institutions account profile information

2011sandeepraj:


configs/env-example
```diff
@@ -27,14 +27,14 @@ E2E_TEST_STUDENT_PASSWORD=ADD_THE_PASSWORD_HERE_TO_RUN_LOCALLY
...
-# BCeID Web Service
+# BCeID Web Service on OC we do not the BCeID_ properties set
 BCeID_WEB_SERVICE_WSDL=https://gws1.test.bceid.ca/webservices/Client/V10/BCeIDService.asmx?wsdl
 BCeID_WEB_SERVICE_ONLINE_SERVICE_ID=
 BCeID_WEB_SERVICE_REQUESTER_USER_GUID=
 BCeID_WEB_SERVICE_AUTH_USER_NAME=
 BCeID_WEB_SERVICE_AUTH_USER_PASSWORD=
 
-# Devops/openshift Make api-init
+# Devops/openshift Make api-init - on  OC we only need API_ ones to run our application and not the BCeID_ ones
```
andrewsignori-aot, line 37: To me "# Devops/openshift Make api-init" already says from where the variables belong to and it is clear that they belong to "Devops/openshift".
If you want to more clear, I don't think that "on OC" is making it clear.
Maybe you can add a comment mentioning that the first set of variables are used only to run the application locally.

sources/packages/api/src-sql/Institution/Create-Institution.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE IF NOT EXISTS institutions(
+  id SERIAL PRIMARY KEY, 
+  legal_operating_name VARCHAR(64) NOT NULL,
```
andrewsignori-aot, line 3: Maybe it would be better to check the length that these fields should have. Is 64 something present in the requirement?
If not, maybe @abschwenker could have a better idea.

sources/packages/api/src-sql/Institution/Create-Institution.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE IF NOT EXISTS institutions(
+  id SERIAL PRIMARY KEY, 
+  legal_operating_name VARCHAR(64) NOT NULL, 
+  operating_name VARCHAR(64) NULL,  
+  primary_phone VARCHAR(64) NOT NULL,  
+  primary_email VARCHAR(64) NOT NULL,  
+  institution_type VARCHAR(64) NOT NULL,  
+  website VARCHAR(64) NOT NULL,  
+  regulating_body VARCHAR(64) NOT NULL,  
+  established_date  Date NOT NULL DEFAULT now(),
```
andrewsignori-aot, line 10: Is the "DEFAULT now()" a requirement here, if not, please remove it.

sources/packages/api/src-sql/Institution/Create-Institution.sql
```diff
@@ -0,0 +1,45 @@
...
+  primary_phone VARCHAR(64) NOT NULL,  
+  primary_email VARCHAR(64) NOT NULL,  
+  institution_type VARCHAR(64) NOT NULL,  
+  website VARCHAR(64) NOT NULL,  
+  regulating_body VARCHAR(64) NOT NULL,  
+  established_date  Date NOT NULL DEFAULT now(), 
+  primary_contact JSONB NOT NULL,
+  legal_authority_contact JSONB NOT NULL,  
+  institution_address JSONB NOT NULL,  
+  institution_mailing_address JSONB,
```
andrewsignori-aot, line 14: This is no longer part of the requirement. I was not considering adding it right now on the UI.

sources/packages/api/src-sql/Institution/Create-Institution.sql
```diff
@@ -0,0 +1,45 @@
...
+  -- Reference Columns
+  user_id INT REFERENCES users(id) ON DELETE CASCADE,
```
andrewsignori-aot, line 16: As per my understanding, the institution will not have a one-to-one relationship with the "users" table. Many BCeIDs will have access and "share the same institution record".
Maybe we need further discussions on this, but right now, one-to-one doesn't seem right to me.

sources/packages/api/src-sql/Institution/Create-Institution.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE IF NOT EXISTS institutions(
+  id SERIAL PRIMARY KEY, 
+  legal_operating_name VARCHAR(64) NOT NULL, 
+  operating_name VARCHAR(64) NULL,  
+  primary_phone VARCHAR(64) NOT NULL,  
+  primary_email VARCHAR(64) NOT NULL,  
+  institution_type VARCHAR(64) NOT NULL,
```
andrewsignori-aot, line 7: I am not considering this field in the UI.
@abschwenker, are we good to add this field or are there some discussions about it?

sources/packages/api/src/route-controllers/institution/institution-req.json
```diff
@@ -0,0 +1,30 @@
+{
```
andrewsignori-aot, line 1: Please remove this file from the final commit.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  Body,
+  Controller,
+  Get,
+  Post,
+  UnprocessableEntityException,
+} from "@nestjs/common";
+
```
andrewsignori-aot, line 8: Are these black lines in the middle of the imports a missing format in the file?

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+import { IUserToken } from "../../auth/userToken.interface";
+import BaseController from "../BaseController";
+
+@Controller("institution")
+export class InstitutionController extends BaseController {
+  constructor(
+    private readonly userService: UserService,
+    private readonly institutionService: InstitutionService,
+  ) {
+    super();
```
andrewsignori-aot, line 23: Please check if the "super" needs to be called here. I think that this could be removed.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  @Post()
+  async create(
+    @Body() payload: CreateInstitutionDto,
+    @UserToken() userToken: IUserToken,
+  ) {
+    // Check user exists or not
+    const existingUser = await this.userService.getUser(userToken.userName);
+    if (existingUser) {
+      throw new UnprocessableEntityException("Institution User already exists");
```
andrewsignori-aot, line 34: I would prefer thown something like a "BadRequestException" here.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    // Save student
+    return this.institutionService.createInstitution(userToken, payload);
```
andrewsignori-aot, line 38: The read-only data that is retrieved from BCeID Web Service should not be expected to be received from the API client.
User first name, last name, institution business GUID, and institution operating name should be retrieved during the institution creation.

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
```diff
@@ -0,0 +1,85 @@
+import { IsNotEmpty, IsOptional } from "class-validator";
+
+export class CreateInstitutionDto {
+  @IsNotEmpty()
+  legalOperatingName: string;
```
andrewsignori-aot, line 5: The read-only data that is retrieved from BCeID Web Service should not be expected to be received from the API client.
User first name, last name, institution business GUID, and institution operating name should be retrieved during the institution creation.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -0,0 +1,89 @@
...
+@Injectable()
+export class InstitutionService extends RecordDataModelService<Institution> {
+  @InjectLogger()
+  logger: LoggerService;
+  constructor(@Inject("Connection") connection: Connection) {
+    super(connection.getRepository(Institution));
+    this.logger.log("[Created]");
+  }
+
+  async getInstitutionByUserName(userName: string): Promise<Institution> {
```
andrewsignori-aot, line 19: I don't believe that we are going to have a one-to-one relationship between institutions and users.
It will probably require further discussion.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
andrewsignori-aot, line 38: It seems that the entire Institution entity is returned here, if it is, please do not return it.

configs/env-example
2011sandeepraj: Done

sources/packages/api/src-sql/Institution/Create-Institution.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE IF NOT EXISTS institutions(
+  id SERIAL PRIMARY KEY, 
+  legal_operating_name VARCHAR(64) NOT NULL, 
+  operating_name VARCHAR(64) NULL,  
+  primary_phone VARCHAR(64) NOT NULL,  
+  primary_email VARCHAR(64) NOT NULL,  
+  institution_type VARCHAR(64) NOT NULL,  
+  website VARCHAR(64) NOT NULL,  
+  regulating_body VARCHAR(64) NOT NULL,
```
andrewsignori-aot, line 9: Are we gonna be creating an enum for this or it will accept any string received from the UI?

sources/packages/api/Dockerfile.dev
```diff
@@ -36,4 +36,4 @@ CMD [ "npm", "run", "docker:start" ]
 
 # Sample line to run the E2E tests for now.
 # E2E tests will be enabled on CI/CD on upcoming tasks.
-# CMD [ "npm", "run", "test:e2e" ]
\ No newline at end of file
+#CMD [ "npm", "run", "test:e2e" ]
```
2011sandeepraj, line 39: Reintroduced the space

sources/packages/api/src-sql/Institution/Create-Institution.sql
abschwenker: I'm not sure the max length BCeID could provide. I can reach out

sources/packages/api/src/route-controllers/institution/institution.controller.ts
2011sandeepraj: This isnt a BadRequest, right? 
UnprocessableEntityException is being used in Student controller for the same scenario. 
We can have a new exception of our own, if needed.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
2011sandeepraj: I dint get this comment - lets discuss

2011sandeepraj: hmmm what is the problem in returning?  - Data security? Serialization overhead?Should we return void?
Student Controller returns the whole thing too?

sources/packages/api/src/route-controllers/institution/institution.controller.ts
2011sandeepraj: Fixed

sources/packages/api/src/route-controllers/institution/institution.controller.ts
2011sandeepraj: Done

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
2011sandeepraj: DIdnt get this , lets discuss

sources/packages/api/src/route-controllers/institution/institution.controller.ts
andrewsignori-aot: The main concern is that we are going to be returning an entire object from the other layer when we have already talked about returning only using Dto, so it falls under security also.
If the student controller is doing it we need to change it, in my option.

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: My thought is that Enum should be at the front end. It will come to API as a string.

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: Lets discuss

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: Yes we need discussion

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
andrewsignori-aot: The data that is received from BCeID Web Service and is displayed to the client in the UI, should again be retrieved during the user creation in the API. We should only allow the client to update the data that he is able to edit.
The source of truth for the read-only data is BCeID Web Service and in my option, the data is going to be persisted to our DB should come from BCeID Web Service and not from the UI.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
andrewsignori-aot: The data that is received from BCeID Web Service and is displayed to the client in the UI, should again be retrieved during the user creation in the API. We should only allow the client to update the data that he is able to edit.
The source of truth for the read-only data is BCeID Web Service and in my option, the data is going to be persisted to our DB should come from BCeID Web Service and not from the UI.

sources/packages/api/src-sql/Institution/Create-Institution.sql
andrewsignori-aot: @abschwenker, it is not about BCeID itself, it is about all other text fields here.

sources/packages/api/src-sql/Institution/Create-Institution.sql
andrewsignori-aot: I believe that the API should know the values. How are we going to enforce that the values received by the API are valid?

sources/packages/api/src-sql/Institution/Create-Institution.sql
abschwenker: 64 for field lengths is good

sources/packages/api/src-sql/Institution/Create-Institution.sql
abschwenker: Confirmed with ticket - Please leave out for now

sources/packages/api/src-sql/Institution/Create-Institution.sql
abschwenker: Correct "Many BCeIDs will have access and "share the same institution record". We need to meet to review further

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -0,0 +1,34 @@
...
+import { UserToken } from "../../auth/decorators/userToken.decorator";
+import { IUserToken } from "../../auth/userToken.interface";
+import BaseController from "../BaseController";
+
+@Controller("institution")
+export class InstitutionController extends BaseController {
+  constructor(
+    private readonly userService: UserService,
+    private readonly institutionService: InstitutionService,
+  ) {}
```
2011sandeepraj, line 18: after removing super call, get this error - 

src/route-controllers/institution/institution.controller.ts:15:3 - error TS2377: Constructors for derived classes must contain a 'super' call.

15   constructor(
     ~~~~~~~~~~~~
16     private readonly userService: UserService,
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
17     private readonly institutionService: InstitutionService,
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
18   ) {}
   ~~~~~~

Found 1 error(s).

2011sandeepraj: @andrewsignori-aot @fwpushan

fwpushan: If you base controller super is required ...

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: Done

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: For Temp Solution, it was agreed between the team that having an enum on the front end and API should have a string. There is the possibility of wrong values being sent into the API and wrong things will be stored. This will be revisited later.

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: Done

sources/packages/api/src-sql/Institution/Create-Institution.sql
2011sandeepraj: Done

sources/packages/api/src/route-controllers/institution/institution.controller.ts
2011sandeepraj: Done

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
2011sandeepraj: Done

sources/packages/api/src/route-controllers/institution/institution.controller.ts
2011sandeepraj: Returnign void

sources/packages/api/test/institution.e2e-spec.ts
```diff
@@ -0,0 +1,73 @@
+require("../env_setup");
+import { Test, TestingModule } from "@nestjs/testing";
+import { HttpStatus, INestApplication } from "@nestjs/common";
+import * as request from "supertest";
+import { AppModule } from "../src/app.module";
+import { KeycloakConfig } from "../src/auth/keycloakConfig";
+import { KeycloakService } from "../src/services/auth/keycloak/keycloak.service";
+
+describe("Users controller (e2e)", () => {
```
fwpushan, line 9: This should be Institution Controller

fwpushan (changes requested):
Please check review comments.

sources/packages/api/src/database/entities/institution.model.ts
```diff
@@ -0,0 +1,88 @@
...
+  })
+  legalAuthorityContact: LegalAuthorityContact;
+
+  @Column({
+    name: "institution_address",
+    type: "jsonb",
+  })
+  institutionAddress: InstitutionAddress;
+
+  @ManyToMany(() => User, { eager: true, cascade: true })
```
andrewsignori-aot, line 81: I would avoid adding the "eager: true".
I believe that when loading an institution we should request it if needed.

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
```diff
@@ -0,0 +1,72 @@
...
+  primaryEmail: string;
+
+  @IsNotEmpty()
+  website: string;
+
+  @IsNotEmpty()
+  regulatingBody: string;
+
+  @IsNotEmpty()
+  establishedDate: string;
```
andrewsignori-aot, line 23: In the future, we need to review the validations that UI is doing and API is not.
This field is not supposed to be a date in the future, for instance.
Don't need to change it right now.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -0,0 +1,87 @@
...
+import { CreateInstitutionDto } from "../../route-controllers/institution/models/institution.dto";
+import { LoggerService } from "../../logger/logger.service";
+import { BCeIDService } from "../bceid/bceid.service";
+import { InjectLogger } from "../../common";
+
+@Injectable()
+export class InstitutionService extends RecordDataModelService<Institution> {
+  @InjectLogger()
+  logger: LoggerService;
+  @Inject()
```
andrewsignori-aot, line 15: Can you please inject this in the constructor as the other ones? I think that the logger would be the only exception.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -0,0 +1,87 @@
...
+  async createInstitution(
+    userInfo: UserInfo,
+    institutionDto: CreateInstitutionDto,
+  ): Promise<Institution> {
+    const institution = this.create();
+    const user = new User();
+    const account = await this.bceidService.getAccountDetails(
+      userInfo.idp_user_name,
+    );
+    console.log("here");
```
andrewsignori-aot, line 32: Please remove the "console.log".

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -0,0 +1,87 @@
...
+    console.dir(account);
+    if (account == null) {
+      //TODO: This scenario occurs when basic BCeID users try to push the bceid account into our application.
+      //Ideally this should be handled at the login page where a Basic BCeID user should NOT be allowed to enter the portal
```
andrewsignori-aot, line 36: A ticket was created in the backlog to block BCeIDs. Maybe you could remove the comments if you think that they are not needed anymore.

sources/packages/api/test/institution.e2e-spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  //IN progress- Will unskip once there is a delete user in place to that this test can be run idempotently.
+  describe.skip("save institution", () => {
```
andrewsignori-aot, line 36: Please remove the skip before commit.

andrewsignori-aot (commented):

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -112,15 +112,15 @@ export class StudentController extends BaseController {
   async create(
     @Body() payload: CreateStudentDto,
     @UserToken() userToken: IUserToken,
-  ) {
+  ): Promise<void> {
```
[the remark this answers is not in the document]
2011sandeepraj: Me and Signori discussed this and unless it is needed, there is no point in sending the student dto back. If we wanted to, because they were being used in Client like so, we need to figure out which fields are needed. At this moment, sending a 201 CREATED is good enough. Let me know if devs need to discuss this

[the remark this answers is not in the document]
andrewsignori-aot: @fwpushan, my understanding is that POST doesn't need to return anything else that, maybe, an identifier that will allow the API consumer to have access to the newly created resource. We should avoid return unnecessary data.
Previously this method was returning the model from DB. We can consider this change as an improvement in comparison with the previous way.

sources/packages/api/src/services/institution/institution.service.ts
2011sandeepraj: Done

sources/packages/api/src/services/institution/institution.service.ts
2011sandeepraj: Modified

sources/packages/api/src/database/entities/institution.model.ts
2011sandeepraj: Is there any case for our application to eager:true?

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
2011sandeepraj: Didnt really get this. It is a string anyways in the dto. Its a date in the model.

sources/packages/api/src/services/institution/institution.service.ts
2011sandeepraj: Make Sense. Done

sources/packages/api/src/route-controllers/student/student.controller.ts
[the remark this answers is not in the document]
fwpushan: OK, Normally rest API specify a [location header for newly created resources](https://restfulapi.net/http-status-201-created/), if no such endpoint exists, normally people send newly created object. But if you, people think that is unnecessary, I'm ok with that.

sources/packages/api/src/database/entities/institution.model.ts
andrewsignori-aot: None that I can see how now. My understanding of this "eager: true" is that every time that we load an institution record all users-related records will be loaded. In general, I would avoid loading from DB more data that we need.

fwpushan: I agree with @andrewsignori-aot Eager will additional overhead on institution data fetch operation.

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
andrewsignori-aot: The requirement says "Established Date (may not be a future date)".
It doesn't matter if the API is accepting a string of data. From a business perspective, it is a date and it should not be in the future. The API should validate this and any other similar validation. We should never rely only on the UI to execute the validations, right?
Please note that I did not request this to be changed right now because we have other similar issues like this, for instance, the SIN validation that is happening only in the UI.

sources/packages/api/src/database/entities/institution.model.ts
2011sandeepraj: Done

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
2011sandeepraj: Holding it for now.

sources/packages/api/src/route-controllers/student/student.controller.ts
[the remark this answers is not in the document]
2011sandeepraj: Thanks.

sources/packages/api/test/institution.e2e-spec.ts
2011sandeepraj: Discussed, Removed comment but left the skip because of BCeID calls

fwpushan (approved):
👍

andrewsignori-aot (approved):
👍

sources/packages/api/src/services/institution/institution.service.ts
2011sandeepraj: DOne

sources/packages/api/Dockerfile.dev
```diff
@@ -28,12 +28,12 @@ RUN npm run build
...
 
 # Sample line to run the unit tests tests for now.
 # Unit tests will be enabled on CI/CD on upcoming tasks.
 # CMD [ "npm", "run", "test:local" ]
 
 # Sample line to run the E2E tests for now.
 # E2E tests will be enabled on CI/CD on upcoming tasks.
-# CMD [ "npm", "run", "test:e2e" ]
\ No newline at end of file
+CMD [ "npm", "run", "test:e2e" ]
```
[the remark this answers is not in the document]
2011sandeepraj: Done

Issue #217, opened 2021-02-22
Institution User Completes Institutions Account Profile Information


mzmichellesmith:
**As ** *(Institution User)* **I want** *(Complete my Institutions Account Profile Information)* **So That** *(I can communicate with SABC).

**Additional Context**
- enter text here

**Acceptance Criteria**
- [x] Given (Institution User), When (is completing an Institution Account Profile ), Then (they must provide the user email)

Given (Institution User), When (is completing an Institution Account Profile ), Then (they must provide the institutions:
- [x] Optional Institution Operating Name (this is different than their Legal name- which comes from Business BCeID) or they select my Legal Operating Name.
- [x] Institution Address: Address line 1, Address Line 2 (Optional), City, Province, Country, Postal Code
- [x] Institution Primary Phone Number
- [x] Institution Primary Email
- [x] Intitution Website
- [x] Established Date (may not be a future date)
- [x] Institutions Legal Authorized Authorities Contact Information- First Name, Last Name, Email, Phone Number
- [x] Instituiontions Primary Contact Information- First Name, Last Name, Email, Phone Number
- [x] Institution Regulating body that governs the institutions- A drop-down List (Private Act of BC Legislature, ICBC, DQAB, PTIB, ITA *update to include full name)

- [x] Given (An Institution User), When (selects save and all mandatory fields are not complete), Then (they are notified of mandatory fields required to complete the Profile
- [x] Given (An Institution User), When (selects save and mandatory fields are complete), Then (the profile information is saved)

Technical Tasks 
--------------------------------------------------

- [ ] Parse and display Token and Additional Information from BCeID
- [ ] Create fields for user input
- [ ] Show notification of mandatory fields on save
- [ ] Save to DB on complete
- [ ] ...
- [ ] ....

abschwenker:
removed 2nd address field due to inconsistencies from stories and wireframe. Can add it with a future story if need

PR #320, opened 2021-03-18
[web][SIMS #217]Institution Profile Creation

andrewsignori-aot:
Created the Institution profile page, as shown below. The page is big enough to break it into more components, but due to uncertainties about the final layout, it is not broken so far.

Renamed a couple of places that were still using "Institute" instead of "Institution".

image
image
image


sources/packages/web/package.json
```diff
@@ -24,7 +24,8 @@
     "vue-class-component": "^8.0.0-0",
     "vue-router": "^4.0.0-0",
     "vuex": "^4.0.0-0",
-    "vuex-class": "^0.3.2"
+    "vuex-class": "^0.3.2",
+    "yup": "^0.32.9"
```
fwpushan, line 28: Do we use this?

sources/packages/web/src/services/InstitutionService.ts
```diff
@@ -0,0 +1,67 @@
...
+          code: "np",
+        },
+      ];
+      return programs;
+    }
+    return [];
+  }
+
+  public async createInstitution(institutionProfile: InstitutionProfileState) {
+    const institutionDto = {
```
fwpushan, line 40: Is there any specific reason for no use of type?

fwpushan (approved):
:+1 , Only one comment.

sources/packages/web/src/types/contracts/institution/ProfileContract.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface ContactInfo {
+  firstName: string;
+  lastName: string;
+  email: string;
+  phoneNumber: string;
+}
+
+export interface Address {
```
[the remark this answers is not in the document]
andrewsignori-aot: Yes, but I don't know if we should.
I kept it separated to allow that they evolve in different ways.
If we figure out that the addresses are going to follow the same interface we can reuse the same.
Does it make sense?

sources/packages/web/src/types/contracts/institution/ProfileContract.ts
```diff
@@ -0,0 +1,28 @@
+export interface InstitutionProfileState {
+  studentEmail: string;
+  operatingName: string;
+  primaryPhoneNumber: string;
+  primaryEmail: string;
+  institutionWebsite: string;
+  regulatingBody: string;
+  estabilishdDate: Date;
```
2011sandeepraj, line 8: spelling mistake established

sources/packages/web/src/types/contracts/institution/ProfileContract.ts
```diff
@@ -0,0 +1,28 @@
...
+  primaryContact: ContactInfo;
+  legalContact: ContactInfo;
+  primaryAddress: Address;
```
2011sandeepraj, line 11: i'd use Different Interfaces contract from primary Contact and legal Contact just because they may evolve differently.

2011sandeepraj (approved):
Just a couple minor things.

sources/packages/web/src/views/institution/InstitutionProfile.vue
```diff
@@ -12,42 +12,340 @@
...
+
+// These options are temporarily hardcoded here as per team decision.
+// These should be moved to or shared between API/web layer.
+const regulatingBodyOptions = [
+  { name: "Private Act of BC Legislature", value: "private-act" },
+  { name: "ICBC", value: "icbc" },
+  { name: "DQAB", value: "dqab" },
+  { name: "PTIB", value: "ptib" },
+  { name: "ITA", value: "ita" },
+];
```
2011sandeepraj, line 349: regulating body options can also come from a service else like the institution names

sources/packages/web/src/views/institution/InstitutionProfile.vue
```diff
@@ -12,42 +12,340 @@
...
+          StudentAid BC’s Privacy policy.
+        </p>
+        <HorizontalSeparator />
+        <Button
+          type="submit"
+          label="Next"
+          icon="pi pi-save"
+          class="p-mt-4"
+          :disabled="isSubmitting"
+        ></Button>
```
2011sandeepraj, line 320: The look and feel in the screen shot underconversation page look odd. May be i am seeing something wrong, but will the look of text input boxes be similar to other screens in the applications?

sources/packages/web/src/services/InstitutionService.ts
andrewsignori-aot: Changed!

sources/packages/web/src/views/institution/InstitutionProfile.vue
andrewsignori-aot: Textboxes and components, in general, are the same as the other pages in the portal.
Some text color may differ from the current wireframes, but it is not part of the requirements so far to them be exactly the same.

sources/packages/web/src/views/institution/InstitutionProfile.vue
andrewsignori-aot: Institution names are going to be retrieved from a dynamic list from the API.
As per the team decision, the "Regulating body" is going to be static and the API is not even validating it for now, which means that future changes will be introduced.

sources/packages/web/src/types/contracts/institution/ProfileContract.ts
andrewsignori-aot: I agree. Let's keep the web and API as-is for now and once we have a more solid requirement for the Institution profile we can review the web and API for these fields.

sources/packages/web/src/types/contracts/institution/ProfileContract.ts
andrewsignori-aot: Fixed!

Issue #175, opened 2021-02-09
Refactor Student Profile to Forms IO


mzmichellesmith:
** Additional Context**

- [ ] We don't care about versioning 
- [ ] Includes the Create and Update of the Profile

**Technical Tasks**

- [x] Replace Student Profile Screen and turn it into Forms IO
- [x] Using the same API structure in the backend.

PR #332, opened 2021-04-05
[web][SIMS-#175] Forms.io implementation of Student Profile

fwpushan:
----------
1. Integrate with new form
2. Create temp route + view

form-profile-edit


2011sandeepraj (commented):
Zenhub issue 175 mentions
"Conduct Server Validations" but i dont see any code pertaining to that  in this ticket. Are you submitting another PR for it?

If yes, please consider this PR approved.

fwpushan:
@2011sandeepraj , Last-time we don't implement server-side validation because we don't have any clear way to share code between both packages (api/web). So, why that acceptance rule is added, I need clarification regarding that.

Issue #138, opened 2021-02-02
Student (Matt) Submits Financial Aid Application


mzmichellesmith:
**As a** *(Student)* **I want** *(Submit my Application)* **So That** *( provide my personal information to SABC)

**Additional Context**
- Save as JSON
- Basic on-screen validations per initial user stories

**Acceptance Criteria**
- [x] Given (A student), When (selects 'Submit' and they haven't completed the mandatory information), Then (they will be alerted of any mandatory fields required to submit). 
- [x] Given (A Student), When (selects 'Submit' and they have completed all the mandatory fields), Then (they are notified that the page has been submitted)

PR #344, opened 2021-04-08
[api][web][SIMS #138]Submit student application

andrewsignori-aot:
Created a new Vue page to render the dynamic form for student application: **/student/dynamic-application-form**
Created a Vue route to load a previously save application: **/student/dynamic-application-form/<application_id>**
Saving all data from submit to the database using a jsonb column.
Created a new controller to save and retrieve the application data.

Changed the page layout to be more aligned to what we have on Figma, which means a fixed size centralized on the screen, as shown below.

For now, the database table is named "applications". Please let me know what do you think would work best. The same is valid for the application controller on the API.

Added authentication to formio requests due to the authorization error while retrieving the form definition that was imported using Postman. The authentication is going to be used also when we need to execute the server validations where we would be probably using the formio API also to revalidate the client validations on formio JSON definition.

Server validations will be done in a separated PR.

image

image


sources/packages/api/src-sql/Applications/Create-Application.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE IF NOT EXISTS applications (
```
2011sandeepraj, line 1: Are these going to be Applications in Draft or Fully Complete and Submitted Applications? If it can be either, then Applications is a good table name. If not, then Draft_Applications or Submitted_Applications may be better names

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+import BaseController from "../BaseController";
+import { CreateApplicationDto } from "./models/application.model";
+
+@Controller("application")
+export class ApplicationController extends BaseController {
+  constructor(private readonly applicationService: ApplicationService) {
+    super();
+  }
+
+  @Get(":id/data")
```
2011sandeepraj, line 21: :id is enough i guess
Consider removing the word "data" from the path

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+    @Param("id") applicationId: string,
+    @UserToken() userToken: IUserToken,
+  ): Promise<any> {
+    const application = await this.applicationService.getApplicationById(
+      applicationId,
+      userToken.userName,
+    );
+    if (!application) {
+      throw new NotFoundException(
+        `Application id ${applicationId} was not found.`,
```
2011sandeepraj, line 32: Consider rewording the error message to "No Application Found". Dont want to log /give out application id"

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -0,0 +1,6 @@
+import { IsNotEmpty } from "class-validator";
+
+export class CreateApplicationDto {
+  @IsNotEmpty()
+  data: any;
```
2011sandeepraj, line 5: Can we have a Type here instead of any?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  async createApplication(
+    userInfo: UserInfo,
+    applicationDto: CreateApplicationDto,
+  ): Promise<Application> {
+    // Find the student realted to the logged user.
+    const student = await this.studentService.getStudentByUserName(
+      userInfo.userName,
+    );
+
```
2011sandeepraj, line 32: If no student is found?

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -65,9 +66,10 @@ export const studentRoutes: Array<RouteRecordRaw> = [
         component: StudentProfile,
       },
       {
-        path: "dynamic-application-form",
+        path: "dynamic-application-form/:id?",
```
2011sandeepraj, line 69: This is going to be visible in the browser as "dynamic-applcation-form/:id" right? Can we name it just "application form". User shouldn't care/know about the "dynamic"

2011sandeepraj (commented):
Please see comments

sources/packages/api/src-sql/Applications/Create-Application.sql
andrewsignori-aot: I believe that we are going to keep the draft and submitted in the same table, but right now we have no clue about it.
Right now the only goal in have something concrete to prove that we can save the formio data to our DB and retrieve it in an easy way to repopulate the form.

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: Yes, it is enough. I added the "data" to make it explicit that we are getting only part of the application, the "data" part.
In the future, if we are going to add another get to retrieve another application fields this could be kept as is.

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: Please note that the aplication Id is the one received from the user, not an internal one.
We are not disclosing any information here.

sources/packages/api/src/route-controllers/application/models/application.model.ts
andrewsignori-aot: This is part of the main idea to have dynamic forms generated using formio. This parameter should be able to receive anything sent by the client.
The next step that I am working on is to have it validated in the same way that happened on the client side.

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: At this point, the user is already authenticated and the user must exist in the DB, but I agree that it would be good to have a check here and maybe throw an exception with more context.

sources/packages/web/src/router/StudentRoutes.ts
andrewsignori-aot: The ":id" is not going to be visible in the browser.
The URL is going to be something like "dynamic-applcation-form/1" or "dynamic-applcation-form/2".
There is no requirement for it right now and this was introduced to us to be able to test and demo that we can save and retrieve the dynamic data from DB.

2011sandeepraj: The point here was about the word "dynamic"

andrewsignori-aot: Oh, I got it now, I can rename it. I am not sure if It should replace the current one on Vue or create a new one. I will adjust the name and keep the current one alive for now.

andrewsignori-aot: Route renamed to "application-form" for now as requested.

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Added a check and a better exception in the case something bad happens and the user is not present.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -0,0 +1,123 @@
...
+        console.error(error);
+      }
+    };
+
+    onMounted(async () => {
+      if (props.id) {
+        const dataToLoad = await ApiClient.Application.getApplicationData(
+          props.id,
+        );
+        console.dir(dataToLoad);
```
fwpushan, line 51: **Remove this console dir/log**

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -0,0 +1,123 @@
...
+      type: String,
+      required: false,
+    },
+  },
+  setup(props: any) {
+    const initialData = ref({});
+
+    const submitted = async (args: any) => {
+      if (props.id) {
+        console.log("Must update");
```
fwpushan, line 35: Remove this log

fwpushan (commented):
LGTM to me, just few comments.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -0,0 +1,117 @@
...
+        initialData.value = dataToLoad;
+      } else {
+        //Get the student info from api call
+        const studentInfo = await StudentService.shared.getStudentInfo();
+        initialData.value = {
+          studentFullName: `${studentInfo.firstName} ${studentInfo.lastName}`,
+          studentGender: studentInfo.gender,
+          studentDateOfBirth: studentInfo.birthDateFormatted,
+          studentPhoneNumber: studentInfo.contact.phone,
+          studentHomeAddress: "111-234 Esquimalt Rd Victoria BC Canada",
```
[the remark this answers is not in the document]
andrewsignori-aot: That was by mistake, I have already changed it.
I just concatenated the values for now, in the same way as they were before, but I believe that we should have a "Formatted Address" available in a more centralized way in the future. Maybe the API could return it.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+export class ApplicationController extends BaseController {
+  constructor(private readonly applicationService: ApplicationService) {
+    super();
+  }
+
+  @Get(":id/data")
+  async getByApplicationId(
+    @Param("id") applicationId: string,
+    @UserToken() userToken: IUserToken,
+  ): Promise<any> {
```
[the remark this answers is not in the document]
andrewsignori-aot: Added a DTO with data property.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
andrewsignori-aot: Removed as requested.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
andrewsignori-aot: Removed as requested.

fwpushan (approved):
👍

PR #351, opened 2021-04-09
Sims323 basic bceid

2011sandeepraj:
Not Allow Basic BCeID users to login to the institution portal


sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -35,18 +34,22 @@ export class UserController extends BaseController {
...
-        email: account.user.email,
-      },
-      institution: {
-        guid: account.institution.guid,
-        legalName: account.institution.legalName,
-      },
-    };
-  }
+    if (account == null) {
+      return {} as BCeIDDetailsDto;
```
andrewsignori-aot, line 38: Can you elaborate more about why is it better to return an empty object instead of null?
It could also return a 404.

sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -20,12 +20,12 @@ export default {
   setup() {
     const router = useRouter();
     const isAuthReady = ref(false);
-
     // Mounding hook
     onMounted(async () => {
       await AppConfigService.shared.initAuthService(ClientIdType.INSTITUTION);
       isAuthReady.value = true;
       const auth = AppConfigService.shared.authService?.authenticated ?? false;
+
```
andrewsignori-aot, line 28: Please consider not committing this file if the only change is the spaces.

sources/packages/web/src/views/institution/Login.vue
```diff
@@ -31,17 +36,21 @@
...
 import { AppConfigService } from "../../services/AppConfigService";
+import { useRoute } from "vue-router";
 
 export default {
   components: {},
-  setup() {
+
+  setup(props: any) {
+    const route = useRoute();
+    const basicBCeID = route.query.basicBCeID;
```
andrewsignori-aot, line 46: This mapping could be done directly into the routes enabling the "props: true". In this way, no additional code is needed here. In the example below the ":id?" is an optional parameter that will be mapped to an "id" property in the component.

```suggestion
{
        path: "dynamic-application-form/:id?",
        name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM,
        component: DynamicStudentApp,
        props: true,
      },
```

fwpushan (changes requested):
Please change this to regular PR and also address issues raised by @andrewsignori-aot.

sources/packages/api/src/route-controllers/user/user.controller.ts
2011sandeepraj: Returning 404 from API forces me to do this in the web/src/services/http/UserApi.ts and i do not like what i am having to do in catch block. Check this issue https://github.com/axios/axios/issues/960

public async bceidAccount(headers?: any): Promise<BCeIDDetailsDto | null> {
    try {
      const response = await this.apiClient.get(
        "users/bceid-account",
        headers || this.addAuthHeader(),
      );
      return response.data as BCeIDDetailsDto;
    } catch (error) {
      const errorObject = JSON.parse(JSON.stringify(error));
      if (errorObject.message.includes("404")) {
        return null;
      } else {
        this.handleRequestError(error);
        throw error;
      }
    }
  }
  
  Returning a null

sources/packages/web/src/views/institution/Login.vue
2011sandeepraj: In the interest of time, Can you approve it so that i can merge this PR and i can submit another PR for this small change?

sources/packages/web/src/views/institution/AppInstitution.vue
2011sandeepraj: Done

sources/packages/api/src/route-controllers/user/user.controller.ts
fwpushan: @2011sandeepraj , There is **error.response** object which has **status** property which gives you exact https status code. Try to explore **error.response** object.

Even you can directly check **error.message**, JSON parsing  is pure unnecessary overhead.

sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -20,7 +20,6 @@ export default {
   setup() {
     const router = useRouter();
     const isAuthReady = ref(false);
-
```
andrewsignori-aot, line 23: Please avoid committing the file if there is no change.

andrewsignori-aot (approved):
👍

fwpushan (approved):

sources/packages/web/src/views/institution/Login.vue
2011sandeepraj: We were not able to do it right now.

Issue #125, opened 2021-02-01
Student  (Matt) Saves A DRAFT Application


mzmichellesmith:
**As a** *(Student)* **I want** *(save a Draft of my application)* **So That** *(I can come back and complete it later)*

**Additional Context**
-The default answer will be blank, the field is mandatory on 'Submit' (Another User Story)
-Saved information is just a draft and does not include validations
- All of the Personal Information is listed on One Page

**Acceptance Criteria**
- [x] Given (A student), When (want to save an application), Then (they select 'Save Information')
- [x] Given (A student), When (select save information), Then (it saves a draft application)

**Technical Tasks**

- [x]  Research Form IO Component Options

PR #352, opened 2021-04-09
[don't merge][api][web][SIMS-#125] Generic Save as draft option in any form

fwpushan:
------
1. Create generic users draft table and model entity
2. Create api and service for CURD operation on users-draft
3. Integrate draft service with forms.io component
---

NB. This PR is only for review purpose. **DON'T MERGE**

### DIFF
--- a/sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
+++ b/sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
@@ -1,0 +1,28 @@
+-- User Draft Table
+CREATE TABLE IF NOT EXISTS users_draft (
+  id SERIAL PRIMARY KEY,
+  -- Reference column
+  user_name VARCHAR(300) NOT NULL UNIQUE REFERENCES users(user_name) ON DELETE CASCADE,
+  -- Data
+  data JSONB NOT NULL,
+
+  -- Form path
+  form_path VARCHAR(100) NOT NULL,
+
+  -- Audit timestamp
+  created_at timestamp without time zone NOT NULL DEFAULT now(),
+  updated_at timestamp without time zone NOT NULL DEFAULT now()
+
+);
+
+-- Indices
+CREATE INDEX IF NOT EXISTS idx_users_draft_name_20210409_8_30 ON users_draft(user_name);
+
+-- Comments
+COMMENT ON TABLE users_draft IS 'A common table to store all forms draft per user. The data in the table is not validated and forms.io compatible';
+COMMENT ON COLUMN users_draft.id IS 'Auto-generated sequential primary key column';
+COMMENT ON COLUMN users_draft.user_name IS 'Reference column to user table through user name';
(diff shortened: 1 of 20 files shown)


sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
```diff
@@ -0,0 +1,28 @@
+-- User Draft Table
+CREATE TABLE IF NOT EXISTS users_draft (
+  id SERIAL PRIMARY KEY,
+  -- Reference column
+  user_name VARCHAR(300) NOT NULL UNIQUE REFERENCES users(user_name) ON DELETE CASCADE,
```
andrewsignori-aot, line 5: The student table is referencing the user through user_id, it would be good to keep the same reference field here. Is that any particular reason to create the relationship using the user_name?

sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
```diff
@@ -0,0 +1,28 @@
...
+  -- Data
+  data JSONB NOT NULL,
+
+  -- Form path
+  form_path VARCHAR(100) NOT NULL,
```
andrewsignori-aot, line 10: I believe that, if we are going to use this generic table to store applications, for instance, the same user could have more than one draft at the same time. We need to check this possibility and maybe adjust accordingly.

sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
fwpushan: I welcome this comment, thanks for raising this. I did it because to check associated user we don't need to create any join query to fetch a draft for any user. Overall performance will be quick. And user_name is technically a business primary key of user table with index.

sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
fwpushan: Here, my design decision was one draft per unique form. But yes there is a huge scope of improvement.

sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
andrewsignori-aot: I see. My point is that we should have this as a pattern and do it in all the places then.
I am not sure if we are not going to end up having the user id loaded from DB also if we are saving roles to our DB, for instance. Maybe we end up going to DB to load user roles and probably the user id, what would be available then to execute all the queries.

sources/packages/api/src-sql/UsersDraft/Create-UsersDraft.sql
andrewsignori-aot: I totally agree. We do not have a solid requirement yet. In general, I liked the design 👍

Issue #222, opened 2021-02-22
Institution User Logs in to SABC Account


mzmichellesmith:
**As a** *(Institution User)* **I want** *log in to my created SABC account in order to access SABC services.

**Additional Context**
- An Institution can Log in anytime to their SABC account using their Business BCeID
- This is the original User logging in- not a new user

**Acceptance Criteria**
- [x] Given (An Institution User), When, (Logs in), Then, (they will be directed to their account Home Welcome Page).
- [x]  Given (An Institution User) When, (Business BCeID Authentication is NOT successful, Then (the institution User will be directed to the Business BCeID site to authenticate their services card. They cannot log in to their SABC Account until their Business BCeID is authenticated.

**Technical Tasks**
- [ ] Check DB for user GUID and if exists allow access to landing page

PR #362, opened 2021-04-13
[web][api][SIMS-#222] Existing BCeID User Login

fwpushan:
-------------------
1. Redirect existing BCeID user to home screen
2. Create a flag based mock BCeID account response for local development without VPN.
----


andrewsignori-aot (approved):
👍

Issue #369, opened 2021-04-14
Refactor Institution Profile to Forms IO


mzmichellesmith:
** Additional Context**

- [ ] We don't care about versioning 

**Technical Tasks**

- [ ] Replace Institution Profile Screen and turn it into Forms IO
- [ ] Using the same API structure in the backend. 
- [ ] Remove API validation and add FormIO Dryrun validation

Issue #223, opened 2021-02-22
Institution User (Zoe) Updates SABC Profile Information


mzmichellesmith:
**As a** *Institution User, I want to update my SABC Profile information so that I can ensure my profile information is up to date

**Additional Context**
- Role Authority to update the profile information is not yet defined.
The information they can update includes:
- [ ] Optional Institution Operating Name (this is different than their Legal name- which comes from Business BCeID) or they select my Legal Operating Name.
- [ ] Institution Address: Address line 1, Address Line 2, City, Province, Country, Postal Code
- [ ] Optional Institution Mailing Address: Address line 1, Address Line 2, City, Province, Country, Postal Code. If the Mailing Address is the same they can select "same as" Address.
- [ ] Institution Primary Phone Number
- [ ] Institution Primary Email
- [ ] Intitution Website
- [ ] Established Date (may not be a future date)
- [ ] Institutions Legal Authorized Authorities Contact Information- First Name, Last Name, Email, Phone Number
- [ ] Instituiontions Primary Contact Information- First Name, Last Name, Email, Phone Number
- [ ] Institution Regulating body that governs the institutions- A drop-down List (Private Act of BC Legislature, ICBC, DQAB, PTIB, ITA *update to include full name)

**Acceptance Criteria**
- [ ] Given (an Institution User), has successfully authenticated their SABC account with their Business BCeID, Then the Institution User can update their SABC profile information 
- [ ] Given (an Institution User) updates the information in the mandatory fields), then the Institution User will be prompted to save the information and the Institution account will be updated.
- [ ]  Given (an Institution User), (has Not saved the updated the information in the mandatory fields in the correct format), then the system will NOT save the updated information.

PR #372, opened 2021-04-16
[api][web][SIMS-#369][SIMS-#223] Refactor Institution Profile page with Forms.io and Handle Profile Update

fwpushan:
------
1. Create dynamic form based vue component
2. Update route
3. Setup basic readonly view
4. Add some error logging in api/form service
---
Issue Link: #369 , #223


sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -1,11 +1,14 @@
...
   Controller,
+  Get,
+  HttpCode,
+  Patch,
   Post,
   UnprocessableEntityException,
 } from "@nestjs/common";
 import { InstitutionService, UserService } from "../../services";
-import { CreateInstitutionDto } from "./models/institution.dto";
+import { CreateInstitutionDto, InstituteDto } from "./models/institution.dto";
```
andrewsignori-aot, line 11: Please try to keep "Institution" instead of "Institute". It seems to be more accurate.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -33,4 +36,17 @@ export class InstitutionController extends BaseController {
...
+  @Patch()
+  async update(
+    @Body() payload: InstituteDto,
+    @UserToken() userToken: IUserToken,
+  ) {
+    await this.institutionService.updateInstitution(userToken, payload);
+  }
+
+  @Get()
+  async institutionDetail(@UserToken() token: IUserToken) {
```
andrewsignori-aot, line 49: Can we try to add the explicit return types always as possible? I think that we should take it as a pattern. Please let me know your thoughts on this.

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
```diff
@@ -67,3 +69,10 @@ export class CreateInstitutionDto {
   @IsNotEmpty()
   postalCode: string;
 }
+
+export class InstituteDto extends PartialType(CreateInstitutionDto) {}
```
andrewsignori-aot, line 73: Please consider renaming it to "Institution" instead of "Institute".

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
```diff
@@ -67,3 +69,10 @@ export class CreateInstitutionDto {
...
+
+export interface InstitutionDetailDto {
+  institution: InstituteDto;
```
andrewsignori-aot, line 76: I totally understand these both objects being together here for the current scenario, but I hope that in the future when we have a more solid idea about the UI/UX, we are going to be dealing with them separately.

fwpushan: I agree.

sources/packages/api/src/route-controllers/institution/models/institution.dto.ts
fwpushan: DONE

sources/packages/api/src/route-controllers/institution/institution.controller.ts
fwpushan: DONE. I missed that part if any method is not returning that mean either void or Promise<void> else it should return proper type or any

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -16,11 +25,45 @@ export class InstitutionService extends RecordDataModelService<Institution> {
   constructor(
     @Inject("Connection") connection: Connection,
     private readonly bceidService: BCeIDService,
+    private readonly userService: UserService,
   ) {
     super(connection.getRepository(Institution));
     this.logger.log("[Created]");
   }
 
+  toDtoObject(institutionEntity: Institution): InstituteDto {
```
andrewsignori-aot: I would consider moving this logic to the InstituteDto and have it be responsible for creating the Dto from a Institution object.
In the future, if every dto have its own logic, it will prevent us from adding all to conversion to the service layer.
I would also be considering adding a library like the below one to do the mappings to us. It will handle the majority of the heavy working by conversion and allow us to configure what is not resolved by convention.
https://github.com/loedeman/AutoMapper/wiki/Getting-started

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -52,7 +95,70 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+    return await this.save(institution);
+  }
+
+  async updateInstitution(userInfo: UserInfo, institutionDto: InstituteDto) {
+    const account = await this.bceidService.getAccountDetails(
+      userInfo.idp_user_name,
+    );
+
+    if (account == null) {
+      //This scenario occurs if basic BCeID users try to push the bceid account into our application.
```
andrewsignori-aot, line 136: This scenario should no longer be valid because we are blocking Basic BCeID to authenticate.
We can keep the code but maybe change the comment.

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
```diff
@@ -0,0 +1,132 @@
+<template>
+  <Message severity="info">
```
andrewsignori-aot, line 2: Can we move this warning message to the formio definition?
I think that this would be one thing that would be good to be dynamic.
If the student profile is not dynamic, it would be good to change there also in the future.

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
```diff
@@ -0,0 +1,132 @@
...
+  setup(props: any) {
+    // Hooks
+    const toast = useToast();
+    const router = useRouter();
+    // Data-bind
+    const initialData = ref({});
+
+    // Event handler
+    const changed = async () => {
+      console.log("changed");
```
andrewsignori-aot: Please consider removing the "changed" if there is no usage for it right now.

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
```diff
@@ -0,0 +1,132 @@
...
+        router.push({
+          name: InstitutionRoutesConst.INSTITUTION_DASHBOARD,
+        });
+      }
+    };
+
+    // Hooks
+    onMounted(async () => {
+      if (props.editMode) {
+        const resp: any = await InstitutionService.shared.getDetail();
```
andrewsignori-aot, line 95: Is there a particular reason why we would not return the InstitutionDetailDto directly on getDetail?

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
```diff
@@ -0,0 +1,132 @@
...
+
+        const detail: InstitutionDetailDto = resp.data;
+        const bceidAccount = detail.account;
+        initialData.value = {
+          userfFrstName: bceidAccount?.user.firstname,
+          userLastName: bceidAccount?.user.surname,
+          userEmail: bceidAccount?.user.email,
```
andrewsignori-aot, line 102: Just double-checking, the user email while editing should come from our DB, if I understand correctly, the username here was replaced in the API for the one from our DB, right?

andrewsignori-aot (commented):
Please take a look at the comments along with the code.

sources/packages/api/src/services/form/form.service.ts
```diff
@@ -99,19 +104,38 @@ export class FormService {
...
-        `Error while retrieving formio authentication token. Error ${authRequest.status}: ${authRequest.statusText}`,
+      if (authRequest.status !== HttpStatus.OK) {
+        throw new Error(
+          `Error while retrieving formio authentication token. Error ${authRequest.status}: ${authRequest.statusText}`,
+        );
+      }
+
+      return authRequest;
+    } catch (excp) {
+      this.logger.error(`Received exception while getting form SA token`);
```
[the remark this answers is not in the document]
fwpushan: This is axios level exception, in case of any http exception axios throw exception not any proper response. You can get error response through `excp.response` and `excp.data`.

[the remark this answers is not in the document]
fwpushan: @andrewsignori-aot I will remove the code in line 118, the catch block is handling all generic error condition.

sources/packages/api/src/services/institution/institution.service.ts
fwpushan: 👍

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
fwpushan: I will fix this

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
fwpushan: I think form.io should only manage form rendering and handling events. If require, we can develop another reusable view with accessory view, then singular responsibility principle will be maintained.

andrewsignori-aot: I don't think that the singular responsibility principle applies to this here.
The message is on the top just because we as developers decided to put it there. It belongs to the form context and it maybe would be rendered right in the middle of the form in the future. This is probably where the ministry could have some benefits of doing changes without code changes.
Create a new formio container only to hold the message seems that we would be creating a "static" place where the message would be displayed.

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
andrewsignori-aot: Can you please elaborate on why it was marked as resolved?

fwpushan: The response is Axios response object. response. data is actual response data which is `InstitutionDetailDto `

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
fwpushan: But message is only attached with profile forms only, even no use for application forms.

fwpushan: I will create one technical debt to handle this change for both of the profile (student, institution) info.

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
fwpushan: I will fix this.

andrewsignori-aot (approved):
👍

Issue #233, opened 2021-02-23
Institution User (Zoe) Updates their Business BCeID Information


mzmichellesmith:
**As a** *(Institution User)* **I want** *(my updated information at Business BCeID to be updated on my SABC Profile / Account so that my current accurate information is provided

**Additional Context**
-The information from Business BCeID includes First Name, Last Name, and Institution as Read Only

**Acceptance Criteria**
- [ ] Given (an Institution User) When (updates their information at Business BCeid and they log in the next time (First Name, Last Name, and Institution), Then (this information will be updated on their SABC Account automagically)

PR #389, opened 2021-04-23
[api][web][SIMS-#233] Sync BCeID Account Info

fwpushan:
-------
1. Create new method to fetch institution using userName
2. Create sync flow
---

### DIFF
--- a/sources/packages/api/src/services/institution/institution.service.spec.ts
+++ b/sources/packages/api/src/services/institution/institution.service.spec.ts
@@ -81,9 +81,7 @@ describe("InstitutionService", () => {
   it("should get institution using user name", async () => {
     const [institution, user] = await factory(userService, service);
 
-    const result = await service.getInstitute({
-      userName: user.userName,
-    });
+    const result = await service.getInstituteByUserName(user.userName);
     expect(result).toBeDefined();
     expect(result.id).toEqual(institution.id);
     expect(result.guid).toEqual(institution.guid);
--- a/sources/packages/api/src/services/institution/institution.service.spec.ts
+++ b/sources/packages/api/src/services/institution/institution.service.spec.ts
@@ -1,13 +1,62 @@
 require("../../../env_setup");
 import { Test, TestingModule } from "@nestjs/testing";
+import * as faker from "faker";
 import { InstitutionService } from "./institution.service";
 import { DatabaseModule } from "../../database/database.module";
 import { BCeIDService } from "../bceid/bceid.service";
 import { ConfigService } from "../config/config.service";
 import { UserService } from "../user/user.service";
+import { Institution, User } from "../../database/entities";
+import { DatabaseService } from "../../database/database.service";
+
+const factory = async (
+  userService: UserService,
+  service: InstitutionService,
+): Promise<[Institution, User]> => {
+  const user = userService.create();
+  user.email = faker.internet.email();
+  user.firstName = faker.name.firstName();
+  user.lastName = faker.name.lastName();
+  user.userName = `${faker.random.uuid()}@bceid`;
+  const institution = service.create();
+  institution.legalOperatingName = faker.company.companyName();
+  institution.operatingName = institution.legalOperatingName;
+  institution.guid = faker.random.uuid();
+  institution.establishedDate = faker.date.past();
+  institution.website = faker.internet.url();
+  institution.institutionAddress = {
+    phone: faker.phone.phoneNumber(),
+    addressLine1: faker.address.streetAddress(),
+    city: "Victoria",
+    country: "Canada",
+    postalCode: "V8V1M9",
+    provinceState: "BC",
+  };
+  institution.primaryEmail = faker.internet.email();
+  institution.primaryPhone = faker.phone.phoneNumber();
+  institution.institutionPrimaryContact = {
+    primaryContactEmail: faker.internet.email(),
+    primaryContactLastName: faker.name.lastName(),
+    primaryContactFirstName: faker.name.firstName(),
+    primaryContactPhone: faker.phone.phoneNumber(),
+  };
+  institution.legalAuthorityContact = {
+    legalAuthorityEmail: faker.internet.email(),
+    legalAuthorityLastName: faker.name.lastName(),
+    legalAuthorityFirstName: faker.name.firstName(),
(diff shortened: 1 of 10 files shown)


sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -94,24 +94,18 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
-      );
-      throw new InternalServerErrorException(
-        `Unable to fetch account information from BCeID`,
-      );
-    }
+  async getInstitute(userInfo: { userName: string }): Promise<Institution> {
+    const query = this.repo
+      .createQueryBuilder("institution")
+      .leftJoinAndSelect("institution.users", "users");
+    const institution = await query
```
andrewsignori-aot, line 101: Can you please elaborate on why we need to break the query into two objects here?

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -94,24 +94,18 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
-    if (account == null) {
-      // Due to any connection or any other integration error retrieval of account info failed.
-      this.logger.error(
-        "Account information could not be retrieved from BCeID",
-      );
-      throw new InternalServerErrorException(
-        `Unable to fetch account information from BCeID`,
-      );
-    }
+  async getInstitute(userInfo: { userName: string }): Promise<Institution> {
```
andrewsignori-aot, line 97: Can this method be named as something like getInstitutionByUserName and have the parameter as userName only?
`async getInstitutionByUserName(userName: string): Promise<Institution>`

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -157,13 +151,25 @@ export class InstitutionService extends RecordDataModelService<Institution> {
     return await this.save(institution);
   }
 
+  async syncInstitution(userInfo: UserInfo): Promise<void> {
+    const account = await this.bceidService.getAccountDetails(
+      userInfo.idp_user_name,
+    );
+    const institutionEntity = await this.getInstitute(userInfo);
+    institutionEntity.guid = account.institution.guid;
```
andrewsignori-aot, line 159: Please consider not updating the institution guid and, please add validation to check if the institution guid on account object is the same institution guid retrieved from DB to be updated.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -157,13 +151,25 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+    institutionEntity.legalOperatingName = account.institution.legalName;
+    await this.save(institutionEntity);
+    const user = await this.userService.getUser(userInfo.userName);
```
andrewsignori-aot, line 162: If the user is already loaded in the institutionEntity, please consider changing the user object inside institutionEntity and executing only one save operation.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -157,13 +151,25 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
   async institutionDetail(userInfo: UserInfo): Promise<InstitutionDetailDto> {
     const account = await this.bceidService.getAccountDetails(
       userInfo.idp_user_name,
     );
-    const institutionEntity = await this.repo.findOne({
-      guid: account.institution.guid,
-    });
+    const institutionEntity = await this.getInstitute(userInfo);
 
     const user = await this.userService.getUser(userInfo.userName);
```
andrewsignori-aot, line 175: I think that I missed it in the previous PR, but I believe that the conversion to DTO should happen in the controller layer or anywhere outside this service class.

sources/packages/api/src/services/institution/institution.service.spec.ts
```diff
@@ -1,13 +1,62 @@
...
+    legalAuthorityPhone: faker.phone.phoneNumber(),
+  };
+  institution.regulatingBody = "ICBC";
+  institution.users = [user];
+  await service.save(institution);
+
+  return [institution, user];
+};
 
 describe("InstitutionService", () => {
```
andrewsignori-aot, line 56: This service is now interacting with the database, are we going to keep those "Unit Tests" executing always to it should be good to skip them?

andrewsignori-aot (commented):
Please take a look in the comments.

sources/packages/api/src/services/institution/institution.service.ts
fwpushan: DONE

sources/packages/api/src/services/institution/institution.service.ts
fwpushan: The egar loading working, so FIXED

sources/packages/api/src/services/institution/institution.service.ts
fwpushan: I like mapper concept, but that require additional learning, so skip this for current works. Mapper could be part of entity.

fwpushan: This code is outdated, I don't know why git is not reflecting that

fwpushan: I don't agree for controller, but I agree there should be a mapping layer.

fwpushan: But mapping logic is there from the beginning, in this PR there is little change on that portion. Refactoring them will be totally change the task estimation.

sources/packages/api/src/services/institution/institution.service.spec.ts
fwpushan: I will include this as integration test later.

fwpushan: Test skipped

sources/packages/api/src/services/institution/institution.service.ts
fwpushan: DONE

sources/packages/api/src/services/institution/institution.service.ts
fwpushan: DONE

PR #425, opened 2021-05-07
[api][web][sims-120] Add institution location form and API to save the data to sims data base after form validation

ann-aot:
…save the data to sims data base after form validation

### DIFF
--- a/sources/packages/api/src-sql/Institution/Create-Institutions-Users.sql
+++ b/sources/packages/api/src-sql/Institution/Create-Institutions-Users.sql
@@ -16,3 +16,4 @@ COMMENT ON TABLE institutions_users IS 'This is a pivot table to capture the rel
 COMMENT ON COLUMN institutions_users.id IS 'Auto-generated sequential primary key column';
 COMMENT ON COLUMN institutions_users.user_id IS 'Foreign key reference to users table users';
 COMMENT ON COLUMN institutions_users.institution_id IS 'Foreign key reference to users table institutions';  
+1
--- a/sources/packages/api/src-sql/Institution/Create-Institutions-Users.sql
+++ b/sources/packages/api/src-sql/Institution/Create-Institutions-Users.sql
@@ -15,5 +15,4 @@ CREATE TABLE IF NOT EXISTS institutions_users(
 COMMENT ON TABLE institutions_users IS 'This is a pivot table to capture the relationship between users table and institutions table';
 COMMENT ON COLUMN institutions_users.id IS 'Auto-generated sequential primary key column';
 COMMENT ON COLUMN institutions_users.user_id IS 'Foreign key reference to users table users';
-COMMENT ON COLUMN institutions_users.institution_id IS 'Foreign key reference to users table institutions';  
-1
+COMMENT ON COLUMN institutions_users.institution_id IS 'Foreign key reference to users table institutions';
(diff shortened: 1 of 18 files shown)


sources/packages/api/src-sql/Institution/Create-Institutions-Users.sql
```diff
@@ -16,3 +16,4 @@ COMMENT ON TABLE institutions_users IS 'This is a pivot table to capture the rel
 COMMENT ON COLUMN institutions_users.id IS 'Auto-generated sequential primary key column';
 COMMENT ON COLUMN institutions_users.user_id IS 'Foreign key reference to users table users';
 COMMENT ON COLUMN institutions_users.institution_id IS 'Foreign key reference to users table institutions';  
+1
```
andrewsignori-aot, line 19: Please consider not including this file in the PR if the only change is the line break.

sources/packages/api/src/types/institutionLocation.ts
```diff
@@ -0,0 +1,26 @@
+export interface InstitutionLocationType {
+  address1: string;
+  address2?: string;
+  applicationId: string;
+  city: string;
+  country: string;
+  locationName: string;
+  postalZipCode: string;
+  provinceState: string;
+  submit: boolean;
```
andrewsignori-aot, line 10: Please consider removing the submit from here since it is not part of the business model.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -27,4 +32,26 @@ export class InstitutionLocationsController extends BaseController {
       data: location.data,
     };
   }
+
+  @Post()
+  async create(
+    @Body() payload: InstitutionLocationType,
+    @UserToken() userToken: IUserToken,
+  ): Promise<GetInstitutionLocationDto> {
```
andrewsignori-aot, line 40: I believe that this method should return a Promise<number> since it is returning only the id.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -27,4 +32,26 @@ export class InstitutionLocationsController extends BaseController {
       data: location.data,
     };
   }
+
+  @Post()
+  async create(
+    @Body() payload: InstitutionLocationType,
```
andrewsignori-aot, line 38: We are trying to keep all objects received/returned by the API as DTO. I would recommend moving this to the DTO level.

sources/packages/web/src/types/contracts/institution/AddInstitutionLocation.ts
```diff
@@ -0,0 +1,11 @@
+export interface Institutionlocation {
+  address1: string;
+  address2?: string;
+  applicationId: string;
+  city: string;
+  country: string;
+  locationName: string;
+  postalZipCode: string;
+  provinceState: string;
+  submit: boolean;
```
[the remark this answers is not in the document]
ann-aot: okay, at any point we won't be using the submit variable right? okay I will remove it

sources/packages/api/src/types/institutionLocation.ts
```diff
@@ -0,0 +1,14 @@
+export interface ValidatedInstitutionLocation {
+  data: {
+    locationName: string,
+    address1: string,
+    address2?: string,
+    city: string,
+    provinceState: string,
+    postalZipCode: string,
+    country: string,
+    applicationId: string,
```
andrewsignori-aot, line 10: Is there any particular reason these properties be present here?
- applicationId: string,
- pid: string,
- process_pid: string

andrewsignori-aot (approved):
Great work @ann-aot, looks good 👍

fwpushan (approved):

Issue #226, opened 2021-02-22
Enable Search for BCeID webservices


mzmichellesmith:
Enable Search --Create new SOAP call to BCeID Webservice (this requires gov vpn)
Enable Search Bypass (Non-vpn) -This is an environment variable to bypass the required VPN
Enable E2E Tests

Andrew Signori has access to Live code to reference if need

PR #437, opened 2021-05-15
[api][SIMS #226]Enable Search for BCeID webservices

andrewsignori-aot:
- Created SOAP request to retrieve all the BCeID accounts that belong to a particular institution. The implementation is retrieving a maximum of 500 records. With the help of @abschwenker, we were able to check that the max amount of records currently on production is around 270 accounts. So far the decision is to retrieve the first page (500 maximum).
- Create a mocked version of BCeIDService to be used when "DUMMY_BCeID_ACCOUNT_RESPONSE=yes";
- Created mocked files with the full content from BCeID Web Service on "Teams>Files>BCeID Web Service Mockups" to allow the development without the Gov VPN enabled. IMPORTANT: the mocked classes will not be part of the production runtime code.
- Documentation updated

### DIFF
--- a/sources/packages/api/src/services/bceid/bceid.service.provider.ts
+++ b/sources/packages/api/src/services/bceid/bceid.service.provider.ts
@@ -1,0 +1,15 @@
+import { BCeIDService } from "..";
+import { BCeIDServiceMock } from "../bceid/bceid.service.mock";
+
+/**
+ * Used to allow developers to consume a mocked version of
+ * BCeID WebService that makes it possible to run the
+ * solution locally without a VPN.
+ */
+export const BCeIDServiceProvider = {
+  provide: BCeIDService,
+  useClass:
+    process.env.DUMMY_BCeID_ACCOUNT_RESPONSE === "yes"
+      ? BCeIDServiceMock
+      : BCeIDService,
+};
--- a/sources/packages/api/src/services/bceid/bceid.service.ts
+++ b/sources/packages/api/src/services/bceid/bceid.service.ts
@@ -6,9 +6,16 @@ import { LoggerService } from "../../logger/logger.service";
 import { ConfigService } from "../config/config.service";
 import { AccountDetails } from "./account-details.model";
 import {
+  SearchAccountOptions,
+  SearchBCeIDAccountResult,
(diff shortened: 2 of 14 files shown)


sources/packages/api/src/services/bceid/bceid.service.ts
```diff
@@ -101,6 +93,77 @@ export class BCeIDService {
...
+      // Destructuring the result of the SOAP request to get the
+      // first item on the array where the parsed js object is.
+      const [result] = await client.searchBCeIDAccountAsync(body);
+      const response = result.searchBCeIDAccountResult as ResponseBase;
+      this.ensureSuccessStatusResult(response);
+      // Array os items from SOAP response that contains all accounts.
+      const accounts = result.searchBCeIDAccountResult.accountList.BCeIDAccount;
+      const pagination = result.searchBCeIDAccountResult.pagination;
+
+      const accountResults = accounts.map((account: any) => {
```
fwpushan, line 139: Can we create any type regarding account?

fwpushan (approved):
LGTM, one minor recommendation.

ann-aot (approved):

sources/packages/api/src/services/bceid/bceid.service.provider.ts
```diff
@@ -0,0 +1,15 @@
+import { BCeIDService } from "..";
+import { BCeIDServiceMock } from "../bceid/bceid.service.mock";
+
+/**
+ * Used to allow developers to consume a mocked version of
+ * BCeID WebService that makes it possible to run the
+ * solution locally without a VPN.
+ */
+export const BCeIDServiceProvider = {
```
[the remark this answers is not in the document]
andrewsignori-aot: Thanks 😃

sources/packages/api/src/services/bceid/bceid.service.ts
andrewsignori-aot: I did not add before because the SOAP structure that is mapped is a little complex in terms of nesting objects but since it could bring more clarity to the code, I created 5 new types to map this. Thanks for the recommendation 👍

guru-aot (approved):

PR #438, opened 2021-05-17
[web][api][sims-391]- Institution Location Summary page

ann-aot:
[web][api][sims-391]- Institution Location Summary page

Created API and Vue Component for Institution Summary Page  with Vuetify3.

TODO: when Vuetify 3 stable version is released.
1. I have used ` v-sheet` since `v-card` was not in vuetify3 alpha version. Once stable version is released `v-card` or veutify3 equivalent component can be used. 
2. I have used ` v-badge` with a dummy icon since `v-chip` was not in vuetify3 alpha version. Once stable version is released `v-chip` or veutify3 equivalent component can be used. 

TODO:
3. `DESIGNATED`  tag is static , in future need to make it dynamic.
image
Screenshot 3.1
4.  Screenshot 4.1 data are dummy values, need to make it dynamic.
image
Screenshot 4.1

### DIFF
--- a/sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
+++ b/sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
@@ -16,6 +16,7 @@ import {
 import { UserToken } from "../../auth/decorators/userToken.decorator";
 import { IUserToken } from "../../auth/userToken.interface";
 import { FormsFlowService } from "../../services/forms-flow/forms-flow.service";
+import { InstitutionLocationsDetails } from "../../types";
 
 @Controller("institution/location")
 export class InstitutionLocationsController extends BaseController {
@@ -88,7 +89,7 @@ export class InstitutionLocationsController extends BaseController {
   @Get()
   async getAllInstitutionLocations(
     @UserToken() userToken: IUserToken,
-  ): Promise<any> {
+  ): Promise<InstitutionLocationsDetails[]> {
     // get all institution locations.
     const Institutionlocations = await this.locationService.getAllInstitutionlocations(
       userToken
--- a/sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
+++ b/sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
@@ -11,19 +11,19 @@ import BaseController from "../BaseController";
 import { InstitutionLocationService, FormService } from "../../services";
 import {
   GetInstitutionLocationDto,
+  InstitutionLocationsDetailsDto,
   InstitutionLocationTypeDto,
 } from "./models/institution-location.dto";
 import { UserToken } from "../../auth/decorators/userToken.decorator";
 import { IUserToken } from "../../auth/userToken.interface";
-import { FormsFlowService } from "../../services/forms-flow/forms-flow.service";
-import { InstitutionLocationsDetails } from "../../types";
-
+import { FormsFlowService, InstitutionService } from "../../services";
 @Controller("institution/location")
 export class InstitutionLocationsController extends BaseController {
   constructor(
     private readonly locationService: InstitutionLocationService,
     private readonly formService: FormService,
     private readonly formsFlowService: FormsFlowService,
+    private readonly institutionService: InstitutionService
   ) {
     super();
   }
@@ -62,10 +62,18 @@ export class InstitutionLocationsController extends BaseController {
         "Not able to create the institution location due to an invalid request.",
       );
     }
+    
+    //To retrive institution id
+    const institutionDetails = await this.institutionService.getInstituteByUserName(
+      userToken.userName,
+    );
+    if (!institutionDetails) {
+      throw new UnprocessableEntityException("Not able to find a institution associated with the current user name.");
+    }
 
     // If the data is valid the location is saved to SIMS DB.
     const createdInstitutionlocation = await this.locationService.createtLocation(
-      userToken,
+      institutionDetails.id,
       dryRunSubmissionResult.data,
     );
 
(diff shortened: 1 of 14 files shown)


sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -52,4 +52,22 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+    );
+    if (!institutionDetails) {
+      throw new Error(
+        "Not able to find a institution associated with the current user name.",
+      );
+    }
+    return this.repo
+      .createQueryBuilder("institution_location")
+      .select(['institution_location.name', 'institution_location.data', 'institution.institutionPrimaryContact'])
+      .leftJoin("institution_location.institution", "institution", "institution.id = institution_location.institution_id")
```
andrewsignori-aot, line 69: I believe that TypeOrm would be guessing what the relationship looks like and we don`t need to explicitly define it. Can you check if it would be possible?
Maybe something like below would be enough:

`
.leftJoin("institution_location.institution", "institution")
`

sources/packages/web/src/assets/css/institution.css
```diff
@@ -0,0 +1,11 @@
+.color-grey {
+  color: #686868;
+}
+.location-box {
+  border-radius: 10px;
+  border: 1px solid #d8d2d2;
+  padding: 20px;
+}
+.color-blue {
+  color: #1763b5;;
```
andrewsignori-aot, line 10: Please remove the double ";;" in the end.

sources/packages/web/src/main.ts
```diff
@@ -25,6 +25,7 @@ import InputNumber from "primevue/inputnumber";
 import ProgressSpinner from "primevue/progressspinner";
 import { AppConfigService } from "./services/AppConfigService";
 import configValidationRules from "./validators/ValidatorConfigUI";
+import "./assets/css/institution.css";
```
andrewsignori-aot, line 28: Just as a recommendation, is it a global style, or is it only needed for the locations page?
If everything in the style belongs only to the scope of the location page, please considering moving it to a scoped CSS block inside locations view.

sources/packages/web/src/services/http/InstitutionLocation.ts
```diff
@@ -18,4 +18,19 @@ export class InstitutionLocationApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async allInstitutionLocationsApi(): Promise<any> {
```
andrewsignori-aot, line 22: Please consider adding the return type to the "Promise<any>".

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -52,4 +52,22 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
 
     return await this.repo.save(newLocation);
   }
+
+  async getAllInstitutionlocations(userInfo: UserInfo): Promise<any> {
```
andrewsignori-aot, line 56: Please consider adding an explicit type to be returned instead of any.

andrewsignori-aot (commented):
Overall looks good, nice work 👍 
Please take a look at some comments.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -84,4 +84,15 @@ export class InstitutionLocationsController extends BaseController {
 
     return createdInstitutionlocation.id;
   }
+
+  @Get()
+  async getAllInstitutionLocations(
+    @UserToken() userToken: IUserToken,
+  ): Promise<any> {
```
fwpushan, line 91: Try to return a typed response.

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -52,4 +52,22 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+    //To retrive institution id
+    const institutionDetails = await this.institutionService.getInstituteByUserName(
+      userInfo.userName,
+    );
+    if (!institutionDetails) {
+      throw new Error(
```
fwpushan, line 62: Throw proper Http Error, here 422 un-processable. entity is more appropriate. In nest js corresponding exception class is available.

sources/packages/api/src/services/institution-location/institution-location.service.ts
fwpushan: I agree with @andrewsignori-aot . `JoinColumn` decorator already define that relationship mapping.

fwpushan (commented):
Mostly perfect. Please check comments.

sources/packages/api/src/services/institution-location/institution-location.service.ts
andrewsignori-aot: I am not sure if the HTTP errors should be returned from the services layer. I would expect the HTTP errors to be returned from the controller layer only.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -88,7 +89,7 @@ export class InstitutionLocationsController extends BaseController {
   @Get()
   async getAllInstitutionLocations(
     @UserToken() userToken: IUserToken,
-  ): Promise<any> {
+  ): Promise<InstitutionLocationsDetails[]> {
```
andrewsignori-aot, line 92: API should return only specific DTO objects to avoid that changes on other layers consuming the same object expose more fields than what the API should be exposing.

fwpushan (approved):
LGTM to me, please address the comments of other reviewer.

andrewsignori-aot (approved):
👍

abschwenker (approved):

guru-aot (approved):

Issue #227, opened 2021-02-22
Institution Administrator (Zoe) authorizes Users at Institution Account


mzmichellesmith:
**As a** *(Institution Administrator)* **I want** *(to be able to add a role to authorize users to SABC who are registered with my institution at BCeID)* **So That** *(Other institution employees can log in)*

**Additional Context**
- Access type for Users is Granted per "Institution Location" & "Intuition Account"
- Users can only be provided access to a location when their UserID (GUID) exists within a “Business Account” at BCeID. The business account name must match the “Legal Operating Name” of the Institution Account
- An Institution location, and Institution Account can have multiple users
- Roles are not dependent on User Type

High-level User TYPE (Access Level)  descriptions:
1) **Administrator**:  This role can only be assigned on the Institution Account. This role is able to view/edit program information requests, view/create programs, view/create designation agreements, and view students as well as confirm enrolment for _all locations_. Can also view/edit institution account details and institution location details. Manage users for the institution account and institution location.
2) **Location Manager**: This role can only be assigned on an Institution Location. This role is able to view/edit program information requests, view/create programs, and view students as well as confirm enrolment for the authorized _location level_. Can also view institution location details, and manage Institution Location users.
3) **User**: This role can only be assigned on an Institution Location. This role is able to view/edit program information requests, view/create programs, and view students as well as confirm enrolment for the authorized _location level_. Can view institution location details. 

High-level ROLE descriptions:
1) **Legal Signing authority**: Is able to view/edit program information requests, view/create programs, view/create/**sign** designation agreements, and view students as well as confirm enrolment for _all locations_. Can also view/edit institution profile, locations, and manage users. This is the only role allowed to sign designation agreements. To be assigned this role you must be an "Administrator"
2) **Primary Contact**: Only 1 primary contact can exist per "Institution Location" & "Intuition Account". This roll will be used displayed on other pages (other user stories)

**Acceptance Criteria**
- [x] Given (institution administrator), When (assigning roles), Then (can VIEW a list of users authorized  at my institution account)
- [x] Given (institution administrator), When (managing user roles), Then ( can ASSIGN Business BCeID users to an Institution account)

**Tasks**
- [ ] Create navigation to page
- [ ] UI Table with: Name, BCeID, Access Level, Role
- [ ] Create add user popup based on table values to add user
- [ ] Create DB Migration for users authorization table
- [ ] Review comments in https://github.com/bcgov/SIMS/pull/464

PR #450, opened 2021-05-21
[api][SIMS-#227] Update Creating of Institution User API

fwpushan:
(#227 )
------
1. Add user info to payload
2. Adding actual user while creating IU
3. Update tests
--

### DIFF
--- a/sources/packages/api/src/route-controllers/institution/institution.controller.ts
+++ b/sources/packages/api/src/route-controllers/institution/institution.controller.ts
@@ -7,6 +7,7 @@ import {
   UnprocessableEntityException,
 } from "@nestjs/common";
 import {
+  BCeIDService,
   InstitutionLocationService,
   InstitutionService,
   UserService,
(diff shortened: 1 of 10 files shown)


sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -133,20 +133,26 @@ export class InstitutionController extends BaseController {
       );
     }
 
+    // Create User
+    const userEntity = this.userService.create();
```
andrewsignori-aot, line 137: I believe that we should ensure that this user belongs to the institution that he is being added to.
I would recommend executing a bceidService.getAccountDetails to ensure the relationship on BCeID and we can also get the firstName, lastName from BCeID response.

andrewsignori-aot (commented):
Please take a look at the comment about the user creation.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
fwpushan: are we getting bceid user name?

fwpushan: Resolved

andrewsignori-aot (approved):
👍

ann-aot (approved):

PR #458, opened 2021-05-25
[web][api][SIMS-#227] Crete User table and add remove user api

fwpushan:
----
1. Create user table to display all users of institution
2. Create api to delete user from institution (Currently using hard delete)
--

### DIFF
--- a/sources/packages/api/src/route-controllers/institution/institution.controller.ts
+++ b/sources/packages/api/src/route-controllers/institution/institution.controller.ts
@@ -1,7 +1,10 @@
 import {
   Body,
   Controller,
+  Delete,
+  ForbiddenException,
   Get,
+  Param,
   Patch,
   Post,
   UnprocessableEntityException,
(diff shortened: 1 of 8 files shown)


sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -313,4 +313,13 @@ export class InstitutionService extends RecordDataModelService<Institution> {
         .map((roleObject) => roleObject.role),
     };
   }
+
+  async removeUser(institutionUser: InstitutionUser) {
+    await this.institutionUserRepo.remove(institutionUser);
```
andrewsignori-aot, line 318: Please consider changing both queries to be executed in Parallel.
Please considering decorating the method with a transaction.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -167,4 +170,22 @@ export class InstitutionController extends BaseController {
...
     return this.institutionService.getUserTypesAndRoles();
   }
+
+  @Delete("/user/:id")
+  async removeUserAuth(
+    @Param("id") id: string,
+    @UserToken() token: IUserToken,
+  ) {
+    // Get auth
+    const institutionUser = await this.institutionService.getUser(+id);
```
andrewsignori-aot, line 180: Is this method ensuring that the user being deleted belongs to the current institution?

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -167,4 +170,22 @@ export class InstitutionController extends BaseController {
   async getUserTypesAndRoles(): Promise<InstitutionUserTypeAndRoleResponseDto> {
     return this.institutionService.getUserTypesAndRoles();
   }
+
+  @Delete("/user/:id")
+  async removeUserAuth(
+    @Param("id") id: string,
```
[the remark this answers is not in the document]
fwpushan: All the example in nestjs is saying @Param() decorator returns string ....

sources/packages/api/src/route-controllers/institution/institution.controller.ts
fwpushan: 👍

sources/packages/api/src/route-controllers/institution/institution.controller.ts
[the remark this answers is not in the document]
andrewsignori-aot: It seems that we already have an example working on our API, as shown below.
image

PR #464, opened 2021-05-26
[api][web][SIMS-#227] Fix API and UI table display issue

fwpushan:
----
1. Fix api to accept multiple location
2. Fix UI table layout issue
--

### DIFF
--- a/sources/packages/api/src/route-controllers/institution/institution.controller.ts
+++ b/sources/packages/api/src/route-controllers/institution/institution.controller.ts
@@ -27,7 +27,7 @@ import { InstitutionUserRespDto } from "./models/institution.user.res.dto";
 import { InstitutionUserAuthDto } from "./models/institution-user-auth.dto";
 import { InstitutionUserRole, InstitutionUserType } from "../../types";
 import { InstitutionUserTypeAndRoleResponseDto } from "./models/institution-user-type-role.res.dto";
-import { User } from "src/database/entities";
+import { User } from "../../database/entities";
 
 @Controller("institution")
 export class InstitutionController extends BaseController {
@@ -141,12 +141,13 @@ export class InstitutionController extends BaseController {
 
     let userEntity: User;
 
+    // Get user if userGuid is supplied
     if (body.userGuid) {
       userEntity = await this.userService.getUser(`${body.userGuid}@bceid`);
     }
 
     if (!userEntity) {
-      // Get user details
+      // Get user details from BCeID
       const accountDetails = await this.accountService.getAccountDetails(
         body.userId,
       );
@@ -157,9 +158,11 @@ export class InstitutionController extends BaseController {
       }
       const userName = `${accountDetails.user.guid}@bceid`;
 
+      // Check user exists or not with same user name
       userEntity = await this.userService.getUser(userName);
 
       if (!userEntity) {
+        // User not exists
         // Create User
         userEntity = this.userService.create();
         userEntity.email = accountDetails.user.email;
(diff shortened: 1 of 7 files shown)


sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -27,6 +27,7 @@ import { InstitutionUserRespDto } from "./models/institution.user.res.dto";
 import { InstitutionUserAuthDto } from "./models/institution-user-auth.dto";
 import { InstitutionUserRole, InstitutionUserType } from "../../types";
 import { InstitutionUserTypeAndRoleResponseDto } from "./models/institution-user-type-role.res.dto";
+import { User } from "src/database/entities";
```
andrewsignori-aot, line 30: Please change to a relative path.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -138,22 +139,35 @@ export class InstitutionController extends BaseController {
...
+      userEntity = await this.userService.getUser(`${body.userGuid}@bceid`);
     }
 
-    // Create User
-    const userEntity = this.userService.create();
-    userEntity.email = accountDetails.user.email;
-    userEntity.firstName = accountDetails.user.firstname;
-    userEntity.lastName = accountDetails.user.surname;
-    userEntity.userName = `${accountDetails.user.guid}@bceid`;
+    if (!userEntity) {
```
andrewsignori-aot, line 148: Can you please elaborate in which scenario the userEntity will be null at this stage?

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -138,22 +139,35 @@ export class InstitutionController extends BaseController {
...
+      // Get user details
+      const accountDetails = await this.accountService.getAccountDetails(
```
andrewsignori-aot, line 150: This method is supposed to be called all the time, right?
Why is it being called only if "userEntity" is not defined?

andrewsignori-aot (commented):
Please see the comments.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
fwpushan: Fixed

sources/packages/api/src/route-controllers/institution/institution.controller.ts
fwpushan: Please check code comments.

andrewsignori-aot: Still not clear to me. We should always check for "Get user details from BCeID". Why is it inside an "if (!userEntity)"?

fwpushan: BCeID service call is require to get user info... if same user already exits we skip that call and create auth association.

andrewsignori-aot: How do we ensure that the user belongs to the institution that it is being associated with?

abschwenker (approved):
Approved to merge due to team changes. Will note comments for review of outstanding comments

sources/packages/web/src/views/institution/InstitutionUserDetails.vue
```diff
@@ -131,7 +139,7 @@ export default {
     const invalidUserType = ref(false);
     const users = ref([] as InstitutionUserViewModel[]);
```
ann-aot, line 140: I guess, we missed the update `users` after a new user is added

Issue #392, opened 2021-04-26
Institution Admin User (Zoe) wants to View & Edit an Institution's Account Location Details


mzmichellesmith:
Clean Up --Ensure Edit after add below

**As a** *(Institution Admin User)* **I want** *(Edit An Institution's Location Details )* **So That** *(I can ensure SABC has the most current information about my location)*

**Additional Context**
- This does not need to be submited to formio. Information will be saved to SIMSDB and not trigger a re-approval

**Acceptance Criteria**
- [x] Given (an Institution Admin User), When (they select the 'Edit Button on the Locations summary page"), Then (they are DIRECTED to the 'Edit Location Details Page')

VIEW
- [x] Given (an Institution Admin User), When, (they View the "Edit Locations Page")Then, (they can SEE the Locations Address (Location Name, Address, City, Postal / Zip Code, Province'/ State and Country)
- [x] Given (an Institution Admin User), When, (they View the "Edit Locations Page")Then, (They can SEE the Saved Changes.) 

EDIT
- [x] Given (an Institution Admin User), When, (want to Edit their location details ")Then, (they can EDIT all fields) 

https://www.figma.com/file/VykmBDwobGoGvK9M5wgOvN/StudentAid-BC-App-Design-System-(WIP!!)---FreshWorks-Studios?node-id=21%3A451

abschwenker:
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/62cc5a9f-527e-4c20-8689-8c4b1f3dbf5e)

PR #474, opened 2021-05-31
Update Institution Location as part of #392

guru-aot:
1. Created Edit location screen in formsflow.ai
2. Integrated the vue component with the Edit location screen to load the data in formsflow.ai
2. Created API to get single location with locationId (This is used to fetch the particular location on click of 'Edit' in Manage Institutions location screen)
3. Created API to update the single location with locationId and the updated data in the table without forms submission in formsflow.ai database
4. Updated toast message in the Manage Institutions location screen on updating
5. Updated fields data in the Manage Institutions location screen

### DIFF
--- a/sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
+++ b/sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
@@ -2,6 +2,7 @@ import {
   Controller,
   Get,
   Post,
+  Patch,
   NotFoundException,
   Param,
   Body,
@@ -26,24 +27,6 @@ export class InstitutionLocationsController extends BaseController {
     private readonly institutionService: InstitutionService
   ) {
     super();
-  }
-
-  @Get(":id")
-  async getInstitutionLocation(
-    @Param("id") id: number,
-  ): Promise<GetInstitutionLocationDto> {
-    const location = await this.locationService.getById(id);
-    if (!location) {
-      throw new NotFoundException(
-        "Not able to retrieve the institution location.",
-      );
-    }
-
-    return {
-      id: location.id,
-      name: location.name,
-      data: location.data,
-    };
   }
 
   @Post()
@@ -94,6 +77,48 @@ export class InstitutionLocationsController extends BaseController {
     return createdInstitutionlocation.id;
   }
 
+  @Patch(":locationId")
+  async update(
+    @Param("locationId") locationId: number,
+    @Body() payload: InstitutionLocationTypeDto,
+    @UserToken() userToken: IUserToken
+    ): Promise<number> {
+      // Validate the location data that will be saved to SIMS DB.
+      const dryRunSubmissionResult = await this.formService.dryRunSubmission(
+        "institutionlocationupdation",
+        payload,
+      );
+  
+      if (!dryRunSubmissionResult.valid) {
+        throw new UnprocessableEntityException(
+          "Not able to update the institution location due to an invalid request.",
+        );
+      }
+      
+      //To retrive institution id
+      const institutionDetails = await this.institutionService.getInstituteByUserName(
+        userToken.userName,
+      );
+      if (!institutionDetails) {
+        throw new UnprocessableEntityException("Not able to find a institution associated with the current user name.");
+      }
+  
+      // If the data is valid the location is updated to SIMS DB.
+      const updatedInstitutionlocation = await this.locationService.updateLocation(
+        locationId,
+        institutionDetails.id,
+        dryRunSubmissionResult.data,
+      );
+  
+      // Save a form to formio to handle the location approval.
+      const submissionResult = await this.formService.submission(
+        "institutionlocation",
+        payload,
+      );
+
+      return updatedInstitutionlocation.id;
+    }
+
   @Get()
   async getAllInstitutionLocations(
     @UserToken() userToken: IUserToken,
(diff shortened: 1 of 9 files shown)


ann-aot (approved):
LGTM

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -94,6 +77,48 @@ export class InstitutionLocationsController extends BaseController {
...
 
+  @Patch(":locationId")
+  async update(
+    @Param("locationId") locationId: number,
+    @Body() payload: InstitutionLocationTypeDto,
+    @UserToken() userToken: IUserToken
+    ): Promise<number> {
+      // Validate the location data that will be saved to SIMS DB.
+      const dryRunSubmissionResult = await this.formService.dryRunSubmission(
+        "institutionlocationupdation",
```
andrewsignori-aot, line 88: Is "updation" a valid word? I think that just "update" should be fine.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -94,6 +77,48 @@ export class InstitutionLocationsController extends BaseController {
...
+      }
+  
+      // If the data is valid the location is updated to SIMS DB.
+      const updatedInstitutionlocation = await this.locationService.updateLocation(
+        locationId,
+        institutionDetails.id,
+        dryRunSubmissionResult.data,
+      );
+  
+      // Save a form to formio to handle the location approval.
```
andrewsignori-aot, line 113: I would recommend not send the data to formio at this moment. There is no requirement for it right now.
The insert location is executing it right now only because it was part of the POC.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -94,6 +77,48 @@ export class InstitutionLocationsController extends BaseController {
     return createdInstitutionlocation.id;
   }
 
+  @Patch(":locationId")
+  async update(
+    @Param("locationId") locationId: number,
```
andrewsignori-aot, line 82: This method needs to validate if the user updating the location has access to it.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -94,6 +77,48 @@ export class InstitutionLocationsController extends BaseController {
...
+      );
+  
+      if (!dryRunSubmissionResult.valid) {
+        throw new UnprocessableEntityException(
+          "Not able to update the institution location due to an invalid request.",
+        );
+      }
+      
+      //To retrive institution id
+      const institutionDetails = await this.institutionService.getInstituteByUserName(
```
andrewsignori-aot, line 99: This method needs to validate if the user updating the location has access to it.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -111,4 +136,24 @@ export class InstitutionLocationsController extends BaseController {
...
+    //To retrive institution id
+    const institutionDetails = await this.institutionService.getInstituteByUserName(
+      userToken.userName,
+    );
+    if (!institutionDetails) {
+      throw new UnprocessableEntityException("Not able to find the Location associated.");
+    }
+    // get all institution locations.
+    const Institutionlocations = await this.locationService.getInstitutionLocation(
+      institutionDetails.id,
```
andrewsignori-aot, line 153: It seems that it will ensure that the location being requested belongs to the institution but the code is not validating if the user has access to the specific location.
we can add the validation or create a tech debit to load into the userToken an array of all locations that the user has access to and validate based on it.

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -43,6 +43,26 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
     return await this.repo.save(newLocation);
   }
 
+  async updateLocation(locationId: number, institution_id: number, data: ValidatedInstitutionLocation): Promise<any> {
```
andrewsignori-aot, line 46: Please be explicit on the type being returned here: Promise<any>.

andrewsignori-aot, line 46: Can we try to keep some consistency when declaring the variables names, like "locationId" vs "institution_id".
I noticed that the files already have some inconsistencies, but it would be good to try to keep the same like "locationId".

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -51,4 +71,14 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
       .where('institution.id = :Id', { Id: institution_id})
       .getMany();
   }
+
+  async getInstitutionLocation(institution_id: number, location_id: number): Promise<InstitutionLocationsDetailsDto> {
+    return this.repo
+      .createQueryBuilder("institution_location")
+      .select(['institution_location.name', 'institution_location.data', 'institution.institutionPrimaryContact', 'institution_location.id'])
+      .leftJoin("institution_location.institution", "institution")
+      .where('institution.id = :Id and institution_Location.id = :locationId', { Id: institution_id, locationId: location_id})
```
andrewsignori-aot, line 80: Please try to keep the consistency while declaring the variables: for instance "Id" vs "id".
I would recommend that line 80 be adjusted to: 
```
{ id: institutionId, locationId: locationId})
```

andrewsignori-aot (commented):
In general, the changes looks nice. Please take a look at the comments.
Can you please check if the files are being formatted? It seems to be that the code is not following the same format.

abschwenker (commented):
lgtm

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
andrewsignori-aot: Created #476 to handle it.

andrewsignori-aot (approved):
👍

Issue #407, opened 2021-04-28
Send Receive Students Information to CRA


mzmichellesmith:
Tasks
- Create DB migration for sin_validated (Yes,No,Fail) -table is ok
- Create timed Task to manage jobs to call SIMSApi to generate file
- Create service for reading response file
- Update sin_validated (Yes,No,Fail)

PR #475, opened 2021-05-31
[api][SIMS #407]Send Receive Students Information to CRA

andrewsignori-aot:
- Created service for reading the response files from CRA SFTP;
- Created DB migration for valid_sin on students table;
- Created 2 unit tests, using a mocked SFTP response, to have at least one test for creating SIN validation files and one to download and parse the SIN validation responses.

The process will read all files on the CRA Response file and process all records. In case of an error on a particular record or entire file, the process will continue the execution and provide a summary in the end.

Sample API response for a SIN validation processing.
image


sources/packages/api/src/cra-integration/cra-files/cra-file-header.ts
```diff
@@ -30,4 +31,14 @@ export class CRAFileHeader implements CRAFileLine {
     header.append("0");
     return header.toString();
   }
+
+  public static CreateFromLine(line: string): CRAFileHeader {
+    const header = new CRAFileHeader();
+    header.transactionCode = line.substr(0, 4) as TransactionCodes;
+    header.processDate = dayjs(line.substr(28, 8), DATE_FORMAT).toDate();
```
ann-aot, line 38: what is in any scenario line.substr(28, 8) is a null or empty string, will dayjs raise an error?

sources/packages/api/src/cra-integration/cra-files/cra-file-header.ts
```diff
@@ -30,4 +31,14 @@ export class CRAFileHeader implements CRAFileLine {
...
+    header.programAreaCode = line.substr(37, 4);
+    header.environmentCode = line.substr(41, 1);
+    header.sequence = parseInt(line.substr(42, 5));
```
ann-aot, line 41: will it to good to check the character line.substr(42, 5) is a digit before doing parseInt- just in any scenario what if it's not a digit. or to add  a exception handling

sources/packages/api/src/cra-integration/cra-files/cra-file-header.ts
andrewsignori-aot: The lines are going to be a fixed 147 long string. The date is part of the CRA response and is not optional, it will be in the response unless something really bad happens from CRA side.
If it happens, the date will be invalid, as shown below, but it will not throw an exception. Right now, it will not cause an issue in the process also.

image

sources/packages/api/src/cra-integration/cra-files/cra-file-header.ts
andrewsignori-aot: The files are fixed-sized and there is no scenario where it will not be a number, it is a mandatory field. In case of something really bad happen, like if CRA send a file with bad format, it will throw an error that will be captured in the processing logs.
IMHO we should not add contingencies to the code for a scenario that is not supposed to happen.
Please let me know if you are comfortable with the explanation or if we need further discussion.

ann-aot (approved):

PR #484, opened 2021-06-05
[web][api][sims-#336] Remove/Edit Institution User Role Access

ann-aot:
1. Add refresh after a user is added to the institution
2. Made Add User and Edit user as vue components
3. Edit user permission to change user permissions, Admin or at least 1 location is required (Logic -> remove existing association/s and add new association/s)
4. Add Active/Deactivate column in the user's table and Active/Deactivate  functionality in the frontend (`is_active` field is  `true` by default)
5. Block login when the user is Deactivated
6. Update existing search queries with Active column (for some search queries both active and disabled users are needed, so I didn't update whose queries, rest I believe I have updated, please let me know if I missed any)


sources/packages/web/src/components/partial-view/student/NavBar.vue
```diff
@@ -43,7 +43,7 @@ export default {
   setup(props: any) {
     const menuItems: any = [];
     const isAuthenticated = computed(
-      () => AppConfigService.shared.authService?.authenticated === true,
+      () => AppConfigService.shared.authService?.authenticated === true
```
ann-aot, line 46: Prettier - Code formatter did this

sources/packages/web/src/views/institution/Login.vue
```diff
@@ -5,13 +5,18 @@
       <div>
         <h1>Welcome to StudentAid BC</h1>
         <h4>
-          We are using BCeID for Authentication. Please click on Login/Register
-          buttons below to start your sign in/sign up with your Business BCeID.
+          We are using BCeID for Authentication. Please click on Login/Register buttons
```
ann-aot, line 8: its due to Prettier - Code formatter

sources/packages/web/src/views/institution/Login.vue
```diff
@@ -5,13 +5,18 @@
...
+          below to start your sign in/sign up with your Business BCeID.
         </h4>
       </div>
       <Message severity="error" v-if="basicBCeID">
-        No such Business account has been found with BCeID. Please login with
-        your Business BCeId
+        No such Business account has been found with BCeID. Please login with your
+        Business BCeId
+      </Message>
```
ann-aot, line 15: due to Prettier - Code formatter

sources/packages/api/src-sql/User/Add-col-user-isactive.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "users" ADD COLUMN IF NOT EXISTS "is_active" BOOLEAN NULL DEFAULT TRUE;
```
andrewsignori-aot, line 1: I believe that "is_active" should not be nullable.

sources/packages/api/src/database/migrations/1622595885302-UpdateUserIsActive.ts
```diff
@@ -0,0 +1,20 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
```
andrewsignori-aot, line 1: By the format of these files and the black lines along with the code, I assume that they are not formatted.
Can you please check if the "format on save" is configure on VS code? It is supposed to get rid of the blank lines and keep the formatting consistency between devs.

sources/packages/api/src-sql/User/Add-col-user-isactive.sql
ann-aot: that I made the default value as TRUE, If that doesn't make i will update the code

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -149,4 +157,79 @@ export class InstitutionController extends BaseController {
...
+  async getInstitutionLocationByUserName(
+    @Param("userName") userName: string,
+  ): Promise<InstitutionUser> {
+    // Get institutionUser
+    const institutionUser = await this.institutionService.getInstitutionUserByUserName(
+      userName,
+    );
+    if (!institutionUser) {
+      throw new UnprocessableEntityException("Not able to retrieve the institution location user.");
+    } else if (!institutionUser.user.isActive){
```
andrewsignori-aot, line 171: We can leave this check for now, but this should be handled globally by the API.
I am planning to introduce this with the PR that I am working on.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -149,4 +157,79 @@ export class InstitutionController extends BaseController {
   async getUserTypesAndRoles(): Promise<InstitutionUserTypeAndRoleResponseDto> {
     return this.institutionService.getUserTypesAndRoles();
   }
+
+  @Get("/user/:userName")
+  async getInstitutionLocationByUserName(
+    @Param("userName") userName: string,
+  ): Promise<InstitutionUser> {
```
andrewsignori-aot, line 164: Please avoid returning the entire entities directly. Please create some DTO with the relevant information.

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -17,13 +18,17 @@ export class UserController extends BaseController {
   }
 
   @Get("/check-user")
-  async checkUser(@UserToken() userToken: IUserToken): Promise<boolean> {
+  async checkUser(@UserToken() userToken: IUserToken): Promise<string> {
```
andrewsignori-aot, line 21: IMHO it would be better to keep returning boolean and change 28 to return boolean true.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -217,6 +216,7 @@ export class InstitutionService extends RecordDataModelService<Institution> {
       .leftJoin("institution.users", "institutionUsers")
       .leftJoin("institutionUsers.user", "user")
       .where("user.userName = :userName", { userName })
+      .andWhere("(user.isActive = :isActive )", {isActive: true})
```
andrewsignori-aot, line 219: Please consider change the line style to the same from line 218.

```
.andWhere("user.isActive = :isActive", {isActive: true})
```

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -396,4 +396,44 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+      if (!authType) {
+        throw new Error(
+          "The combination of user type and user role is not valid.",
+        );
+      }
+      newAuthorization.authType = authType;
+      newAuthorizationEntries.push(newAuthorization)
+    }
+    // delete existing associations
+    await this.deleteAssociationByUserID(institutionUser)
```
andrewsignori-aot, line 436: I believe that it is possible to execute both commands together.
It would be good to have them executed in "same transaction" in case of the save fail (line 437), all the authentication records would be lost.

sources/packages/api/src/services/user/user.service.ts
```diff
@@ -14,4 +14,18 @@ export class UserService extends DataModelService<User> {
...
+    return this.repo
+    .createQueryBuilder()
+    .update(User)
+    .set({ isActive: isActive})
+    .where("id = :id", { id: userId })
+    .execute();
+  }
+
+  async getActiveUser(userName: string) {
+    return this.repo.findOne({ where: {userName:userName, isActive: true} });
```
andrewsignori-aot, line 28: I believe that the line could be simplified as below. If you think that it is better, please adjust it.

```
return this.repo.findOne({userName:userName, isActive: true});
```

sources/packages/web/src/components/institutions/modals/AddInstitutionUserModal.vue
```diff
@@ -0,0 +1,237 @@
+<template>
+  <!-- Add user -->
+  <Dialog
+    v-if="showAddUser"
+    header="Add User to Account"
+    v-model:visible="display"
```
andrewsignori-aot, line 6: Please consider removing the v-model to be in alignment with the other bindings.

sources/packages/web/src/components/institutions/modals/AddInstitutionUserModal.vue
```diff
@@ -0,0 +1,237 @@
...
+      <v-btn color="primary" @click="submitAddUser()">
+        Add User
+        <v-icon right>mdi-account </v-icon>
+      </v-btn>
+    </template>
+  </Dialog>
+</template>
+
+<script lang="ts">
+/* eslint-disable */
```
andrewsignori-aot, line 100: Can you please elaborate on why the eslint is disabled here?

sources/packages/web/src/components/institutions/modals/AddInstitutionUserModal.vue
```diff
@@ -0,0 +1,237 @@
...
+    const getInstitutionLocationList = async () => {
+      //Get Institution Locations
+      institutionLocationList.value = await InstitutionService.shared.getAllInstitutionLocations();
+    };
+
+    const closeAddUser = () => {
+      selectUser.value = { name: "", code: "", id: "" };
+      isAdmin.value = false;
+      invalidName.value = false;
+      getInstitutionLocationList();
```
andrewsignori-aot, line 147: Does it need an await?

sources/packages/web/src/components/institutions/modals/AddInstitutionUserModal.vue
```diff
@@ -0,0 +1,237 @@
...
+        }
+        if (payLoad.value?.location?.length === 0) {
+          invalidUserType.value = true;
+        }
+      }
+    };
+
+    onMounted(async () => {
+      // call institution location
+      getInstitutionLocationList();
```
andrewsignori-aot, line 213: Does it need an await?

sources/packages/web/src/components/institutions/modals/EditInstitutionUserModal.vue
```diff
@@ -0,0 +1,264 @@
...
+    const closeEditUser = async () => {
+      context.emit("updateShowEditInstitutionModal");
+      invalidUserType.value = false;
+      await getInstitutionUserDetails();
+      checkIsAdmin();
+      await getAndSetAllLocations();
+    };
+    const checkIsAdmin = () => {
+      isAdmin.value =
+        userData.value?.authorizations?.filter((el: any) => el?.authType?.type == "admin")
```
andrewsignori-aot, line 138: The expression could be simplified with the use of "some" instead of "filter" as below.
IMHO this should be resolved by API or some centralized logic. Don't need to change it right the way.

```
userData.value?.authorizations?.some(el => el?.authType?.type === "admin")
```

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -34,7 +34,13 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
         path: AppRoutes.LoginWithBusinessBCeID,
         name: InstitutionRoutesConst.LOGIN_WITH_BUSINESS_BCEID,
         component: Login,
-        props: { showBasicBCeIDMessage: true },
+        props: { showBasicBCeIDMessage: true, showBasicDisabledUserMessage: false },
```
andrewsignori-aot, line 37: I believe that "Basic" in this context doesn't make too much sense. I would recommend just to just "showDisabledUserMessage".

sources/packages/web/src/services/AppConfigService.ts
```diff
@@ -101,6 +101,7 @@ export class AppConfigService {
     type: ClientIdType,
     authService?: KeyCloak.KeycloakInstance,
     isBasicBCeID?: boolean,
+    isDisabled?: boolean
```
andrewsignori-aot, line 104: Please consider renaming to something that indicates that the user is disabled, like "isUserDisabled".

sources/packages/web/src/services/AuthService.ts
```diff
@@ -45,7 +45,15 @@ export default async function(
...
             isForbiddenUser = true;
-          }
+          } else if (await UserService.shared.checkUser(authHeader) === "Disabled"){
+            await AppConfigService.shared.logout(
+              ClientIdType.INSTITUTION,
+              keycloak,
+              false,
+              true
+            );
+            isForbiddenUser = true;
```
andrewsignori-aot, line 55: Do you need to control if the user is forbidden at this point?

sources/packages/web/src/services/InstitutionService.ts
```diff
@@ -170,14 +171,16 @@ export class InstitutionService {
     return ApiClient.Institution.getUserTypeAndRoles();
   }
 
-  public async createUser(data: InstitutionUser): Promise<void> {
+  public async prepareUserPayload(isNew:boolean, data:InstitutionUserAuthDetails) {
```
andrewsignori-aot, line 174: Is this method being used outside the service? If not, please consider changing to private.
I saw a similar one on Vue page, is it possible to move that logic here?

sources/packages/web/src/services/http/UserApi.ts
```diff
@@ -2,13 +2,13 @@ import { BCeIDDetailsDto, BCeIDAccountsDto } from "../../types/contracts/UserCon
 import HttpBaseClient from "./common/HttpBaseClient";
 
 export class UserApi extends HttpBaseClient {
-  public async checkUser(): Promise<boolean> {
+  public async checkUser(headers?: any): Promise<string> {
```
andrewsignori-aot, line 5: If the strings are enum like, please consider creating the enum.

andrewsignori-aot (commented):
Great work doing all the heavy lifting to this big-ticket 👍 
Please ensure that the files are correctly formatted (ESLint+Prettier).
For modals management, I saw another approach that uses a Promise to display the modal and await to if be resolved/closed. It helps also with the "refreshing" operations from the page that is showing the modal.
it would be something like the below snippet. I will not consider this a blocker, but we can talk about it in another opportunity if you think that it is worth 😁 

```
aync addNewUser(){
    await myModal.Show();
    await refreshUserList();
}
```

sources/packages/web/src/services/AuthService.ts
ann-aot: image
isForbiddenUser = true; was set `true` beacuse of this condition

andrewsignori-aot: The is fine, but if the user is forbidden, are we not logging off the user and redirecting it to log in before the code reaches this area? Is this "throw new Error" needed?

abschwenker (commented):
Please have a quick look at Code Smells, other than that. LGTM

sources/packages/web/src/services/AuthService.ts
ann-aot: yes, we are logging off and not allowing the user to login.

ann-aot: "throw new Error"  was already there. that why I set the flag to true. and the function is the default function. if it doesn't make any sense, we can remove it

sources/packages/api/src/route-controllers/user/user.controller.ts
```diff
@@ -23,8 +23,8 @@ export class UserController extends BaseController {
       if (!userInSABC) {
         return false;
       } else {
-        return true;
-      }
+        return true
```
andrewsignori-aot, line 26: This file doesn't seem that is formatted. Can you check, please?

sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -38,11 +37,13 @@ export default {
         });
       } else {
         if (await UserService.shared.checkUser()) {
-          await InstitutionService.shared.sync();
-          if (route.path === AppRoutes.InstitutionRoot) {
-            router.push({
-              name: InstitutionRoutesConst.INSTITUTION_DASHBOARD,
-            });
+          if (await UserService.shared.checkActiveUser()) {
```
andrewsignori-aot, line 40: Do we need to check if the user is active at this point?
I believe that if it is not the user will be logged of before reaching this point, doesn't it?
f not, keep it, and please consider this comment.

andrewsignori-aot (commented):

sources/packages/web/src/views/institution/AppInstitution.vue
ann-aot: INSTITUTION_DASHBOARD should only be shown to active users,  right? that's why I add that check there

andrewsignori-aot (approved):
👍

abschwenker (approved):

PR #491, opened 2021-06-08
[api][SIMS #336] Adding API Authorizations Capabilities

andrewsignori-aot:
Added more authorizations capabilities to the API as described below.

The authorizations are in majority based on the Nestjs guards, more info about it https://docs.nestjs.com/guards

- Created **AllowAuthorizedParty** decorator. Used to restrict access to the API endpoints accordingly to the azp token property. It means that we can restrict the access to the users based on the Keycloak client used to authenticate the user.
Please see the sample below of how to use it.

image

- Created **IsInstitutionAdmin** decorator. Used to restrict access to the API endpoints to only institution administrators.
Please see the sample below of how to use it.

image

- Created **HasLocationAccess** decorator. Used to restrict access to the API endpoints considering the location-specific permissions. This decorator can use a LocationId parameter from the URL to figure out which location permission should be validated.
Please see the sample below of how to use it.

image

- Added a global validation to block an inactive user to access the API and created the **AllowInactiveUser** decorator to handle some exceptions, as below.

image

While dealing with institutions we can use the IInstitutionUserToken to have access to all authorizations information that could also be used inside the controllers.

image

UserId and InstituionId are also available on the token.


ann-aot (commented):
LGTM @andrewsignori-aot  great work :+1:

ann-aot (approved):
LGTM @andrewsignori-aot  great work :+1:

guru-aot (approved):

Issue #376, opened 2021-04-20
Add Institution User Access type authorization to pages


mzmichellesmith:
**Acceptance Criteria**
- [x] Manage institution (Including Institution Details, Manage Locations, Manage Users, Manage Designation) is accessible only by (Institution Admin)
- [x] Location (Including Dashboard, Location Details, Manage Programs, Manage Student Applications) is accessible by (Institution Admin, Location-Manager (for that location), User (for that location))
- [x] Location (Including: Manage Users) is accessible by (Institution Admin, Location-Manager (for that location))

**Tasks**
- [ ] Save locations/roles info to Vue Store and load locations left menu from Vue Store also.
- [ ] Restrict routes
- [ ] Remove UI links
- [ ] Review API permissions

PR #496, opened 2021-06-14
[web][api][SIMS-#376]-Add Institution User Access type authorization to pages->Save locations/roles info to Vue Store and load locations left menu from Vue Store

ann-aot:
1. Added vue stores with vuex (vuex updated to 4.0.1)
2. Save locations/roles and logged in user info to Vue Store 
3. load locations left menu from Vue Store.
4. removed previous API call for fetch logged in user location to display in the sidebar and related functions 
5.  moved the icon to left in the `ADD NEW LOCATION` and  `ADD NEW USER`  button in Manageuser and ManageLocation file
6. some files are updated due to formatting.
7. update location store after adding a new institution location and after editing a location


sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -57,20 +57,20 @@ export class InstitutionLocationsController extends BaseController {
     }
 
     //To retrive institution id
-    const institutionDetails = await this.institutionService.getInstituteByUserName(
-      userToken.userName,
-    );
+    const institutionDetails =
```
andrewsignori-aot, line 60: Just checking, is there any change here related to the PR or just the formatting?

ann-aot: ya, it's just formatting

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -259,4 +264,50 @@ export class InstitutionController extends BaseController {
...
+    // Get logged in user and location details with auth
+    const institutionUser =
+      await this.institutionService.getInstitutionUserByUserName(
+        token.userName,
+      );
+    const user = {
+      user: {
+        firstName: institutionUser.user?.firstName,
+        lastName: institutionUser.user?.lastName,
+        userName: token.userName ?? institutionUser.user?.userName,
```
andrewsignori-aot, line 281: I believe that we should not expose this in the API. Is there any use of this outside the API?
The API will always have this information on the koken and the Vue application or other possible API client should not send it to the API.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -259,4 +264,50 @@ export class InstitutionController extends BaseController {
...
+        isActive: institutionUser.user?.isActive,
+        isAdmin: token.authorizations.isAdmin,
+        id: institutionUser.user?.id,
```
andrewsignori-aot, line 284: Same from the previous comment, I believe that we should not expose this outside the API.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -259,4 +264,50 @@ export class InstitutionController extends BaseController {
...
+        email: token.email ?? institutionUser.user?.email,
```
andrewsignori-aot, line 285: We must rely on the email from our database. The user can change the email received from BCeID, so our DB is our source of truth.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
ann-aot: client-side token has logged in user's username only. In header, we need logged in user information. 
image

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -259,4 +264,50 @@ export class InstitutionController extends BaseController {
...
+      },
+    };
+    const LocationAuth = {
+      authorizations: token.authorizations,
+    };
+
+    return {
+      ...user,
+      ...LocationAuth,
```
andrewsignori-aot, line 294: Please consider not exposing the InstitutionUserAuthorizations directly in the API.
In the future, if someone adds some sensitive data to InstitutionUserAuthorizations it could be automatically exposed by mistake. That's why we try to enforce the DTOs' creation to ensure that future maintenance and changes do not expose more than what API is supposed to do.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -259,4 +264,50 @@ export class InstitutionController extends BaseController {
...
+  @Get("/my-locations")
+  async getMyInstitutionLocations(
+    @UserToken() userToken: IInstitutionUserToken,
+  ): Promise<InstitutionLocationsDetailsDto[]> {
+    // get all institution locations that user has access too.
+    if (userToken.authorizations.isAdmin()) {
+      return await this.locationService.getAllInstitutionlocations(
+        userToken?.authorizations?.institutionId,
+      );
+    } else {
```
andrewsignori-aot, line 307: Just a recommendation, you can simplify the code by removing the "else".

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -116,19 +116,9 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
-              LEFT JOIN       sims.institution_user_auth INSTITUTIONUSERAUTH 
-              ON              institutionuserauth.institution_user_id = institutionuser.id 
-              LEFT OUTER JOIN sims.institution_locations INSTITUTIONLOCATIONS 
-              ON              institutionlocations.institution_id = institutionuser.institution_id 
-              AND             (institutionlocations.id = institutionuserauth.institution_location_id 
-              OR              institutionuserauth.institution_location_id IS NULL )
-              WHERE           users.user_name = '${userName}'`);
+  async getMyInstitutionlocations(
+    locationIds: number[],
+  ): Promise<InstitutionLocationsDetailsDto[]> {
```
andrewsignori-aot, line 116: I believe that this method should return the entity returned from the DB.
If we are going to consume this method anywhere else it would be good to not be restricted to what the API is returning to this specific method through the DTO.
IMHO, the service layer should be more generic.

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
```diff
@@ -85,7 +91,17 @@ export default {
         });
       }
     };
-    onMounted(getuserLocationList);
+    watch(
+      () => userLocationList.value,
+      async () => {
+        // get user details
+        await getuserLocationList();
```
andrewsignori-aot, line 98: It seems that this method doesn't need to be async anymore. Can you check, please?
Also, it seems that there is no reason to return it on the Setup method, can you check it also, please?

sources/packages/web/src/store/modules/institution/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+  InstitutionLocationsDetails,
+} from "@/types";
+
+export const actions: ActionTree<InstitutionLocationState, RootState> = {
+  initialize(): boolean {
+    console.log("Initializing the institution store...");
+    return true;
+  },
+  getUserInstitutionDetails(context): Promise<InstitutionUserAndAuthDetails> {
+    return new Promise((resolve, reject) => {
```
andrewsignori-aot, line 16: IMHO, we should use the async/await pattern in the same way that we have been using in others areas of the application. Is there any particular reason to use Promise callbacks here?

sources/packages/web/src/store/modules/institution/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+      InstitutionService.shared
+        .getMyInstitutionDetails()
+        .then(resultComment => {
+          context.commit("setmyInstitutionAndUserDetailsState", resultComment);
```
andrewsignori-aot, line 20: I think that "my" should start with upper case like "setMyInstitutionAndUserDetailsState"

sources/packages/web/src/store/modules/institution/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+import { InstitutionService } from "@/services/InstitutionService";
+import {
+  InstitutionLocationState,
+  RootState,
+  InstitutionUserAndAuthDetails,
+  InstitutionLocationsDetails,
+} from "@/types";
+
+export const actions: ActionTree<InstitutionLocationState, RootState> = {
+  initialize(): boolean {
```
andrewsignori-aot, line 11: Is this method being used anywhere?

sources/packages/web/src/types/StoreState.ts
```diff
@@ -0,0 +1,13 @@
...
+  InstitutionLocationsDetails,
+} from "@/types";
+
+export interface RootState {
+  version: string;
+}
+
+export interface InstitutionLocationState {
+  myInstitutionAndUserDetailsState: InstitutionUserAndAuthDetails;
+  myInstitutionLocationsState: InstitutionLocationsDetails;
```
andrewsignori-aot, line 12: Are the InstitutionLocationsDetails interface being reused here because we need all the data or just to avoid create a specific type.
IMO, we should add to state what is needed. If we don't need the entire object, please consider creating a specific object to hold the state.

sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -51,6 +53,11 @@ export default {
           });
         }
       }
+      if (isAuthenticated.value) {
+        store.dispatch("institution/initialize");
```
andrewsignori-aot, line 57: Maybe the initialize could be used to call getUserInstitutionDetails and getUserInstitutionLocationDetails in parallel.
Instead of calling 3 dispatch methods here, there will only one dispatch call internally calling the API twice, but in parallel.

andrewsignori-aot (commented):
Great work enabling the Vue store and retrieving all the data 👍 

sources/packages/web/src/store/modules/institution/actions.ts
ann-aot: image
`Unnecessary try/catch wrapper` error is throwing when I use async/await with try catch

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -278,15 +282,24 @@ export class InstitutionController extends BaseController {
...
         isActive: institutionUser.user?.isActive,
         isAdmin: token.authorizations.isAdmin,
-        id: institutionUser.user?.id,
-        email: token.email ?? institutionUser.user?.email,
+        email: institutionUser.user?.email,
       },
     };
     const LocationAuth = {
-      authorizations: token.authorizations,
+      authorizations: {
```
andrewsignori-aot, line 291: Please rename the variable "LocationAuth" to "locationAuth".

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -259,4 +268,72 @@ export class InstitutionController extends BaseController {
...
     );
   }
+
+  @Get("/my-details")
+  async getMyDetails(
+    @UserToken() token: IInstitutionUserToken,
+  ): Promise<InstitutionUserAndAuthDetailsDto> {
+    // Get logged in user and location details with auth
+    const institutionUser =
+      await this.institutionService.getInstitutionUserByUserName(
```
andrewsignori-aot, line 278: It seems that there is nothing being used from the institution data.
Please consider retrieving only the user from DB.

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
```diff
@@ -85,10 +91,21 @@ export default {
         });
       }
     };
-    onMounted(getuserLocationList);
+    watch(
+      () => userLocationList.value,
+      async () => {
```
andrewsignori-aot, line 96: I believe that the async could be removed here now, right?

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
```diff
@@ -85,10 +91,21 @@ export default {
...
+        // get user details
+        getuserLocationList();
+      }
+    );
+    onMounted(() => {
+      // get user details
+      if (Array.isArray(userLocationList.value)) {
```
andrewsignori-aot, line 103: Can you elaborate on why this check is needed?

sources/packages/web/src/store/modules/institution/institution.ts
```diff
@@ -0,0 +1,27 @@
...
+  InstitutionUserAndAuthDetailsForStore,
+  InstitutionLocationsDetailsForStore,
+} from "@/types";
+
+import { actions } from "./actions";
+import { getters } from "./getters";
+import { mutations } from "./mutations";
+
+export const state: InstitutionLocationState = {
+  myInstitutionAndUserDetailsState: {} as InstitutionUserAndAuthDetailsForStore,
```
andrewsignori-aot, line 15: Just a recommendation...
These 2 variables (myInstitutionAndUserDetailsState and myInstitutionLocationsState) seems to be too verbose, IMO.
If you agree, myInstitutionLocationsState could be renamed to something like "locationState" since it will be accessed like below.

```
store.state.institution.locationState
```

myInstitutionAndUserDetailsState could be renamed and have their inner objects moved as below.

```
store.state.institution.userState
store.state.institution.authorizationsState
```

andrewsignori-aot (commented):
Few more comments. Thanks for the changes so far 👍

andrewsignori-aot (approved):
Looks good 👍

guru-aot (approved):

Issue #470, opened 2021-05-27
Institution User Views and edits Education Program Information details


mzmichellesmith:
**As a** *(Institution location)* **I want** *(view and edit information on my education program)* **So That** *(I can maintain my program information)*

**Additional Context**
See #279 for add program
This uses the same form as "Add Program"

**Acceptance Criteria**
- [x] Given (an Institution User), When (selecting edit program), Then (they have the ability to view and edit all program information)
- [x] Given (an Institution User), When (entering program information), Then (they have to complete all mandatory fields)
- [x] Given (An Institution User), When (selects submit), Then (the program information is saved)
- [x] All fields are mandatory

**Tasks**
- [ ] Render form
- [ ] Submit to SIMSDB and formio for new application

Figma: https://www.figma.com/file/VykmBDwobGoGvK9M5wgOvN/StudentAid-BC-App-Design-System-(WIP!!)---FreshWorks-Studios?node-id=1884%3A20472

PR #500, opened 2021-06-16
[web][api][SIMS #470]Institution User Views and edits Education Program Information details

andrewsignori-aot:
- Enabled the editing program page.
- Fix some columns on DB that were declared as BIT and should be BOOLEAN.
- Manually fixed OpenShift DEV.


sources/packages/web/src/services/http/EducationProgramApi.ts
```diff
@@ -1,8 +1,23 @@
...
 
 export class EducationProgramApi extends HttpBaseClient {
-  public async createProgram(createProgramDto: any): Promise<void> {
+  public async getProgram(programId: number): Promise<void> {
+    try {
+      const response = await this.apiClient.get(
+        `institution/education-program/${programId}`,
+        this.addAuthHeader(),
+      );
+      return response.data;
```
ann-aot, line 11: function is having a `void` promise and data is returning

sources/packages/web/src/services/EducationProgramService.ts
```diff
@@ -9,10 +9,21 @@ export class EducationProgramService {
     return this.instance || (this.instance = new this());
   }
 
-  public async createProgram(data: any): Promise<void> {
+  public async getProgram(programId: number): Promise<any> {
```
ann-aot, line 12: instead of promise `any` can we create a interface

andrewsignori-aot: Yes, I added, but right now it is not gonna pass in Sonar Cloud due to a duplication constraint.
Right now, we are trying to figure out how to configure the Sonal Cloud to ignore it.
The other way would be to share the DTOs between API and web, which we are not doing right now.
I don't like any, but it is the only way to have the PR passing the gates right now.

abschwenker (approved):

ann-aot (commented):
LGTM 👍 Good work, added some comments

sources/packages/web/src/services/http/EducationProgramApi.ts
andrewsignori-aot: Fixed to any for now until we figure out the duplication issue with Sonar Cloud or we start to share the DTOs between web/api.

PR #503, opened 2021-06-17
[web][api][SIMS-#376]-Add Institution User Access type authorization to pages

ann-aot:
1. Restrict routes
2. Remove UI links
3. Review API permissions
4. corrected typo and some files changed due to formatting

### Rout gaurds enabled:
**_meta:_**
 1. **requiresAuth** : set `true` if login required for a page
 2. **clientType**: to differentiate between each client type 
 3. **userTypes**: add userType that are allowed to view the Page image _SCREENSHOT 3.1_
 4. **checkAllowedLocation**: if the page needs a check by location ID. (For eg, the User page inside Locations in the Homesidebar (_SCREENSHOT 4.1_) 
image
_SCREENSHOT 4.1_
should only access by the location manager of that location and Admin), then add the location manager userType in the list,
Note:
 location ID is taken from the params of the router, so this should only be added to routers that have a  location ID.

image
_SCREENSHOT 4.2_

Please let me know if you need a call, to discuss it


sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -67,37 +67,38 @@ export class InstitutionLocationsController extends BaseController {
...
       formId: submissionResult.formId,
       formUrl: submissionResult.absolutePath,
       submissionId: submissionResult.submissionId,
     });
 
-    return createdInstitutionlocation.id;
+    return createdInstitutionLocation.id;
   }
 
+  @IsInstitutionAdmin()
```
andrewsignori-aot, line 93: The decorators will work in "AND" logic, so this will require admin users and will not let a non-admin have access to the method. When the decorator @HasLocationAccess is present, it will assume also that, if the user is an admin, he will have access to the method.
I would say that we need only the @HasLocationAccess for this method because it is the most restrictive one and also will alow admin to have access to it.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -133,7 +132,7 @@ export class InstitutionLocationsController extends BaseController {
     }
     // get all institution locations.
     const InstitutionLocationData =
-      await this.locationService.getAllInstitutionlocations(
+      await this.locationService.getAllInstitutionLocations(
```
andrewsignori-aot, line 135: Not part of your changes, but here we don`t need the code in lines 125-131 because we can use the institution id from the token directly in line 136. Up to you if you want to handle this with this PR.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
ann-aot: institution location should only be updated/edited by admin user right @andrewsignori-aot ?

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -230,6 +238,12 @@ export class InstitutionController extends BaseController {
     } else if (!institutionUser.user.isActive) {
       throw new UnprocessableEntityException("Not an Active User.");
     }
+    // checking if user belong to logged-in users institution
```
andrewsignori-aot, line 241: Now that we have the IsInstitutionAdmin added we can remove the lines 234-240 because if the user is not present or is not active it will be blocked before reach the controller.
Please consider removing the above-mentioned lines.

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -247,22 +261,27 @@ export class InstitutionController extends BaseController {
...
-    // We can pass the institution id to getInstitutionUserByUserName to ensure it.
-    // The institution id is present on token.
     const institutionUser =
       await this.institutionService.getInstitutionUserByUserName(userName);
     if (!institutionUser) {
       throw new UnprocessableEntityException(
         "user does not exist in the system.",
       );
     }
+    // checking if user belong to logged-in users institution
```
andrewsignori-aot, line 279: Now that we have the IsInstitutionAdmin added we can remove the lines 274-277.
Please consider removing the above-mentioned lines.

sources/packages/api/src/route-controllers/institution/models/institution.user.res.dto.ts
```diff
@@ -40,5 +40,11 @@ export interface InstitutionUserAuth {
 }
 export interface InstitutionUserAndAuthDetailsDto {
   authorizations: InstitutionUserAuth;
-  user: Pick<User, "firstName" | "lastName" | "isActive" | "email">;
+  user: {
```
andrewsignori-aot, line 43: I personally like this way more than the previous one 👍

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
```diff
@@ -22,15 +24,23 @@ export default {
   setup() {
     const store = useStore();
     const router = useRouter();
-
     const userLocationList = computed(() => store.state.institution.locationState);
+    const isAdmin = computed(() => store.state.institution.userState?.isAdmin);
```
andrewsignori-aot, line 28: The code looks good, but as a recommendation for now or the future, I believe that we can use the composition API to aggregate these common-auth properties under the same hook, like below. In this way we would avoid doing the below line over and over again.
```
const isAdmin = computed(() => store.state.institution.userState?.isAdmin);
```

I believe that we could have something like:

```
const auth = useInstitutionAuth();
```
where auth would be the computed properties like isAdmin and userAuth(authorizations);
Please consider it as a recommendation. We can handle this in the future if appropriate.

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -67,6 +68,10 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
               default: InstitutionDashboard,
               sidebar: InstitutionHomeSideBar,
             },
+            meta: {
+              requiresAuth: true,
```
andrewsignori-aot, line 72: Would make sense to invert the logic here? I mean, consider that everything requires authentication unless that explicit defined as "requiresAuth: false".
I believe that the majority of pages will be under authentication, so we would avoid writing "requiresAuth: true" and if some dev forgets it on a page that is public, it will be easily noted.

sources/packages/web/src/router/index.ts
```diff
@@ -12,5 +20,40 @@ const router = createRouter({
   history: createWebHistory(process.env.BASE_URL),
   routes,
 });
+router.beforeEach((to, from, next) => {
+  const isAuthenticated = AppConfigService.shared.authService;
```
andrewsignori-aot, line 24: This can be moved inside the first if.

sources/packages/web/src/types/contracts/InstitutionRouteMeta.ts
```diff
@@ -0,0 +1,12 @@
+export enum InstitutionRouteUserType {
```
andrewsignori-aot, line 1: Can you please elaborate on why do we need 2 enums here?
Can they evolve differently in the future?

sources/packages/web/src/router/InstitutionRoutes.ts
ann-aot: Actually, for `requiresAuth: false` or if `requiresAuth` is not given, the logic to allow `next()`, if `requiresAuth: true `, logic to check, whether the user is authenticated or not, if authenticated a set of checks else redirecting to `Login` page, let me know if this logic doesn't make sense

sources/packages/web/src/services/UserAuthorizationService.ts
```diff
@@ -0,0 +1,126 @@
...
+  ) {
+    const isInstitutionAdmin = store.getters["institution/myDetails"]?.isAdmin;
+    const institutionUserTypes: InstitutionUserAuthRolesAndLocation[] =
+      store.getters["institution/myAuthorizationDetails"].authorizations;
+    /* Check is user a Institution Admin and Institution 
+    Admin users has access to the requested page*/
+
+    if (isInstitutionAdmin) {
+      return allowedTypeList.some(
+        type => type === InstitutionRouteUserType.ADMIN,
```
andrewsignori-aot, line 27: If the user is already defined as an admin in the state, why do we need to check the permissions again?

sources/packages/web/src/services/UserAuthorizationService.ts
```diff
@@ -0,0 +1,126 @@
...
+      );
+    } else {
```
andrewsignori-aot, line 29: This else could be removed to simplify the logic.

sources/packages/web/src/services/UserAuthorizationService.ts
```diff
@@ -0,0 +1,126 @@
...
+              InstitutionRouteUserType.USER,
+              InstitutionUserTypes.user,
+              urlParams.locationId,
+            )
+          )
+            return true;
+
+          return false;
+        });
+      } else {
```
andrewsignori-aot, line 62: This else could be removed to simplify the logic.

sources/packages/web/src/types/contracts/InstitutionRouteMeta.ts
ann-aot: `InstitutionRouteUserType` for router meta, just difference in values with `Institution_` prefix, ya, we can use the same enum

sources/packages/web/src/services/UserAuthorizationService.ts
```diff
@@ -0,0 +1,126 @@
...
+            )
+          )
+            return true;
+
+          return false;
+        });
+      }
+    }
+  }
+  // this.checkUserTypeIsAllowed(type, InstitutionRouteUserType.USER, urlParams?.locationId)
```
andrewsignori-aot, line 93: Please remove the commented line.

sources/packages/web/src/services/UserAuthorizationService.ts
```diff
@@ -0,0 +1,126 @@
...
+  public checkUserTypeIsAllowed(
+    allowedUserType: string,
+    checkUserTypeFromRouter: string,
```
andrewsignori-aot, line 96: Consider replacing the types for the enum ones.

andrewsignori-aot (commented):
Great work 👍 
The metadata organization looks really good. Please check the comments.

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
andrewsignori-aot: Yes, you are right! So, please remove the other decorator for @HasInstitutionAccess because the @IsAdmin will take precedence.

sources/packages/web/src/services/UserAuthorizationService.ts
ann-aot: image
else added beacuse of this

sources/packages/web/src/services/UserAuthorizationService.ts
ann-aot: image
else added beacuse of this

sources/packages/web/src/router/InstitutionRoutes.ts
andrewsignori-aot: There is nothing wrong with the logic, my recommendation was to, in case the requiresAuth is not present, consider that by default the authentication is required and just change the logic to execute the authentication check to something like below. In this way, we do not need to explicitly set "requiresAuth: true". That is a recommendation only 😊 

```
if(metadat.requiresAuth !== false)
```

sources/packages/web/src/services/UserAuthorizationService.ts
ann-aot: no, it's the reverse logic here, if the state of the user is Admin then its checking whether the page is accessible to admin user type.

abschwenker (commented):
LGTM

guru-aot:
LGTM, great work @ann-aot

guru-aot (approved):

andrewsignori-aot (approved):
👍

PR #506, opened 2021-06-17
[improvement]Cached form.io token

andrewsignori-aot:
Adding the Form.IO token to a cache to avoid requesting it multiple times while it is still valid.
The token will be stored locally and once it is expired or to about to expire (30s) it will be renewed.


sources/packages/api/src/services/form/form.service.ts
```diff
@@ -119,14 +130,48 @@ export class FormService {
...
+  /**
+   * Tries get the token from the cache if it is not expired
+   * or about to be expired, otherwise get a new one.
+   * @returns Cached token or a new one.
+   */
+  private async tryGetCachedToken(): Promise<{ token: any; decoded: any }> {
+    const needRenewToken =
+      !this.cachedToken || // If the token is not present, retrieve it.
+      needRenewJwtToken(
+        +this.cachedToken.decoded["exp"],
```
ann-aot, line 159: there is a `+` before this.cachedToken.decoded["exp"].

andrewsignori-aot: This is to convert the string to a number in a "shortcut" way.
https://www.codegrepper.com/code-examples/typescript/how+to+convert+string+to+number+in+typescript

image

ann-aot (approved):
LGTM

guru-aot (approved):
LGTM

Issue #471, opened 2021-05-27
Institution User Views and edits Education offering details


abschwenker:
**As a** *(Institution location)* **I want** *(view and edit information on my education program offering)* **So That** *(I can maintain my program offering information)*

**Additional Context**
See #421 for add offering
This uses the same form as "Add Program"

**Acceptance Criteria**
- [x] Given (an Institution User), When (selecting edit offering), Then (they have the ability to view and edit all offering information)
- [x] Given (an Institution User), When (entering offering information), Then (they have to complete all mandatory fields)
- [x] Given (An Institution User), When (selects submit), Then (the offering information is saved)
- [x] All fields are mandatory

**Tasks**
- [x] Render form
- [x] Submit to SIMSDB

Figma: https://www.figma.com/file/VykmBDwobGoGvK9M5wgOvN/StudentAid-BC-App-Design-System-(WIP!!)---FreshWorks-Studios?node-id=1884%3A21118

PR #512, opened 2021-06-18
[web][SIMS #471]Institution User Views and edits Education offering details

guru-aot:
Institution User Views and edits Education offering details#471

Given (an Institution User), When (selecting edit offering), Then (they have the ability to view and edit all offering information)
Given (an Institution User), When (entering offering information), Then (they have to complete all mandatory fields)
Given (An Institution User), When (selects submit), Then (the offering information is saved)
All fields are mandatory

### DIFF
--- a/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
+++ b/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
@@ -93,7 +93,25 @@ export class EducationProgramOfferingController {
         "Not able to find a Education Program Offering associated with the current Education Program, Location and offering.",
       );
     }
-    return offering;
+    return {
+      id: offering.id,
+      name: offering.name,
+      studyStartDate: offering.studyStartDate,
+      studyEndDate: offering.studyEndDate,
+      breakStartDate: offering.breakStartDate,
+      breakEndDate: offering.breakEndDate,
+      actualTuitionCosts: offering.actualTuitionCosts,
+      programRelatedCosts: offering.programRelatedCosts,
+      mandatoryFees: offering.mandatoryFees,
+      exceptionalExpenses: offering.exceptionalExpenses,
+      tuitionRemittanceRequestedAmount:
+        offering.tuitionRemittanceRequestedAmount,
+      offeringDelivered: offering.offeringDelivered,
+      lacksStudyDates: offering.lacksStudyDates,
+      lacksStudyBreaks: offering.lacksStudyBreaks,
+      lacksFixedCosts: offering.lacksFixedCosts,
+      tuitionRemittanceRequested: offering.tuitionRemittanceRequested,
+    };
   }
 
   @HasLocationAccess("locationId")
--- a/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
+++ b/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
@@ -13,6 +13,7 @@ import {
   CreateEducationProgramOfferingDto,
   EducationProgramOfferingDto,
   ProgramOfferingDto,
+  UpdateEducationProgramOfferingDto,
 } from "./models/education-program-offering.dto";
 import { FormNames } from "../../services/form/constants";
 import { EducationProgramOfferingService, FormService } from "../../services";
@@ -94,5 +95,35 @@ export class EducationProgramOfferingController {
       );
     }
     return offering;
+  }
+
+  @HasLocationAccess("locationId")
+  @Patch(
+    "location/:locationId/education-program/:programId/offering/:offeringId",
+  )
+  async updateProgramOffering(
+    @Body() payload: UpdateEducationProgramOfferingDto,
+    @Param("locationId") locationId: number,
+    @Param("programId") programId?: number,
+    @Param("offeringId") offeringId?: number,
+  ): Promise<number> {
+    const updatingResult = await this.formService.dryRunSubmission(
+      FormNames.EducationProgramOffering,
+      payload,
+    );
+
+    if (!updatingResult.valid) {
+      throw new UnprocessableEntityException(
+        "Not able to a update a program offering due to an invalid request.",
+      );
+    }
+
+    const updateProgramOffering = await this.programOfferingService.updateEducationProgramOffering(
+      locationId,
+      programId,
+      offeringId,
+      payload,
+    );
+    return updateProgramOffering.affected;
   }
 }
--- a/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
+++ b/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
@@ -2,6 +2,7 @@ import {
   Body,
   Controller,
   Post,
+  Patch,
   Param,
   Get,
   UnprocessableEntityException,
@@ -11,7 +12,8 @@ import { AllowAuthorizedParty, HasLocationAccess } from "../../auth/decorators";
 import {
   CreateEducationProgramOfferingDto,
   EducationProgramOfferingDto,
-} from "./models/create-education-program-offering.dto";
+  ProgramOfferingDto,
+} from "./models/education-program-offering.dto";
 import { FormNames } from "../../services/form/constants";
 import { EducationProgramOfferingService, FormService } from "../../services";
 
@@ -71,5 +73,26 @@ export class EducationProgramOfferingController {
       studyDates: offering.studyDates,
       offeringDelivered: offering.offeringDelivered,
     }));
+  }
+
+  @HasLocationAccess("locationId")
+  @Get("location/:locationId/education-program/:programId/offering/:offeringId")
+  async getProgramOffering(
+    @Param("locationId") locationId: number,
+    @Param("programId") programId: number,
+    @Param("offeringId") offeringId: number,
+  ): Promise<ProgramOfferingDto> {
+    //To retrive Education program offering corresponding to ProgramId and LocationId
+    const offering = await this.programOfferingService.getProgramOffering(
+      locationId,
+      programId,
+      offeringId,
+    );
+    if (!offering) {
+      throw new UnprocessableEntityException(
+        "Not able to find a Education Program Offering associated with the current Education Program, Location and offering.",
+      );
+    }
+    return offering;
   }
 }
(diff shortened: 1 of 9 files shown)


sources/packages/api/src/route-controllers/education-program-offering/models/education-program-offering.dto.ts
```diff
@@ -0,0 +1,61 @@
...
+  exceptionalExpenses: number;
+  tuitionRemittanceRequestedAmount: number;
+  offeringDelivered: string;
+  lacksStudyDates: boolean;
+  lacksStudyBreaks: boolean;
+  lacksFixedCosts: boolean;
+  tuitionRemittanceRequested: string;
+}
+
+export interface UpdateEducationProgramOfferingDto {
```
andrewsignori-aot, line 19: Do we need 2 DTOs with exactly the same fields? I believe that they will evolve together and we can keep only one.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -72,4 +75,55 @@ export class EducationProgramOfferingController {
...
+      );
+    }
+
+    const updateProgramOffering = await this.programOfferingService.updateEducationProgramOffering(
+      locationId,
+      programId,
+      offeringId,
+      payload,
+    );
+    return updateProgramOffering.affected;
```
andrewsignori-aot, line 127: So far, I believe that we are not returning anything for Patch/Put.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -72,4 +75,55 @@ export class EducationProgramOfferingController {
...
+      locationId,
+      programId,
+      offeringId,
+    );
+    if (!offering) {
+      throw new UnprocessableEntityException(
+        "Not able to find a Education Program Offering associated with the current Education Program, Location and offering.",
+      );
+    }
+    return offering;
```
andrewsignori-aot, line 97: I believe that the entire object from DB entity will be returned here. The DTO will not prevent that the fields that are not part of the DTO be returned.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -95,4 +101,94 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+        "offerings.tuitionRemittanceRequestedAmount",
+        "offerings.offeringDelivered",
+        "offerings.lacksStudyDates",
+        "offerings.lacksStudyBreaks",
+        "offerings.lacksFixedCosts",
+        "offerings.tuitionRemittanceRequested",
+      ])
+      .innerJoin("offerings.educationProgram", "educationProgram")
+      .innerJoin("offerings.institutionLocation", "institutionLocation")
+      .where(
```
andrewsignori-aot, line 141: Just a recommendation, the andWhere could be used to break the 3 where condition into 3 lines.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
guru-aot: Yeah we are not returning, but this is a good practice to return the updated rows and error handle if it fails. I did the same for InsitutionLocation patch

sources/packages/api/src/route-controllers/education-program-offering/models/education-program-offering.dto.ts
guru-aot: done

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -72,4 +74,55 @@ export class EducationProgramOfferingController {
...
+  }
+
+  @HasLocationAccess("locationId")
+  @Patch(
```
andrewsignori-aot, line 100: Open for discussion, maybe PUT would be appropriate because we are replacing basically the entire entity data.

sources/packages/web/src/services/http/EducationProgramOfferingApi.ts
```diff
@@ -35,4 +35,39 @@ export class EducationProgramOfferingApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async getProgramOffering(
+    locationId: number,
+    programId: number,
+    offeringId: number,
+  ): Promise<Offering> {
```
andrewsignori-aot, line 43: I would recommend renaming it to OfferingDto to be aligned with the others.

sources/packages/web/src/views/institution/locations/programs/LocationProgramOffering.vue
```diff
@@ -27,26 +29,59 @@ export default {
...
         router.push({
           name: InstitutionRoutesConst.VIEW_LOCATION_PROGRAMS,
         });
-        toast.add({
-          severity: "success",
-          summary: "Created!",
-          detail: "Education Offering created successfully!",
-          life: 5000,
-        });
+        if (props.offeringId) {
```
andrewsignori-aot, line 70: The logic to display the toast notification could be moved inside the previous "if (props.offeringId)" after line 59 and 65 to avoid checking the same condition again.
Or the toast notification could be simplified to something generic for create/edit like "Education Offering saved successfully!"

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
guru-aot: done

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -95,4 +98,94 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+  }
+
+  /**
+   * Creates a new education program offering at program level
+   * @param locationId location id to associate the new program offering.
+   * @param programId program id to associate the new program offering.
+   * @param educationProgramOffering Information used to create the program offering.
+   * @returns Education program offering created.
+   */
+  async updateEducationProgramOffering(
```
andrewsignori-aot, line 156: Is the logic for insert/update different in a way that justifies the creation of 2 different methods? Maybe we can have just one method to d both.

andrewsignori-aot (commented):

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
guru-aot: I believe, the difference is obvious and in API updates, I feel Patch works better than PUT as we do not know all the entities of the particular dto. I do agree with you that we are basically replacing the entire data, but Patch has a control of updating with Primary key of the DTO not just deleting the existing row and creating a new one altogether.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
guru-aot: We may not know, it would be always better to create 2 different methods rather having one and adding and editing dto's/methods/services when there is a difference between create and update.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
guru-aot: Please elaborate on this, I got only the DTO values that are returned.

ann-aot: `offering` is a direct DB entity, while returning we may need to prepare the DTO and return

guru-aot: done

ann-aot (approved):
LGTM 👍

abschwenker (approved):

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
andrewsignori-aot: After a conversation, we agreed to continue with the PATCH.
Besides that, I am not sure from what is the origin of the statement has origin: "_but Patch has a control of updating with Primary key of the DTO not just deleting the existing row and creating a new one altogether._". Can you point some reference with definition?

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
andrewsignori-aot: The affect rows are supposed to be always 1 and the HTTP status is supposed to return the status os the operation.
When you mention that it is a good practice to return in this way, is there any reference that supports it?

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -25,35 +28,13 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
   async createEducationProgramOffering(
     locationId: number,
     programId: number,
-    educationProgramOffering: CreateEducationProgramOfferingDto,
+    educationProgramOffering: SaveEducationProgramOfferingDto,
```
andrewsignori-aot, line 31: We should not save directly the DTO received from the payload in the controller. In this way, if the API receives any other field that is not supposed to be saved it will be persisted. It basically opens the entire table to be updated without any control from the API. Please revert it as it was before.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
andrewsignori-aot: As per our conversation, this will not be considered a blocker, but IMHO we are preparing the application for something that is not an actual requirement and does not seems to be something that will be coming, and, in case it comes, it will not be hard to change it. Either way, all good to keep it.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
guru-aot: This was creating code duplication error in Sonar check. Thats why created a method. Is that ok?

andrewsignori-aot (approved):
👍

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
guru-aot: https://en.wikipedia.org/wiki/Patch_verb#PUT_vs_PATCH_vs_POST

PR #515, opened 2021-06-21
Feature/invoke application workflow

andrewsignori-aot:
- Created a new endpoint "/api/system-access/application/:id" with full data access to any application data. This endpoint is accessible only through "forms-flow-bpm" client.
- Renamed CRA endpoints to be also under the same "system-access" URL to be aligned to the endpoints that require full access and are limited to specific clients.
- Replace the fullapplication form.IO definition with the more completed on "SFAA2022-23", as per the conversation with @abschwenker.
- Created a helper service to keep all the workflow interactions and made it easy to deal with, like shown below.

```
await this.workflow.startApplicationAssessment(createdApplication.id);
```

- Adjusted field names on personal information read-only section to adapt to the new form, as shown below (still missing PD status).
image


abschwenker (approved):

abschwenker (commented):
Please review sonarcloud, other than that LGTM

sources/packages/api/src/database/data.model.service.ts
```diff
@@ -25,8 +24,13 @@ export class DataModelService<DataModel extends BaseModel>  {
...
 
-  async findById(id: number) {
-    return this.repo.findByIds([id]);
+  async findById(id: number): Promise<DataModel> {
+    const result = await this.repo.findByIds([id]);
+    if (result?.length === 1) {
+      return result[0];
+    }
+
+    return null;
```
ann-aot, line 33: do we need to give `Promise<DataModel|null>`, since we are returning null in line 33

andrewsignori-aot: Yes, it makes sense. Code changed and comments added.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+    @Param("id") applicationId: number,
+  ): Promise<GetApplicationDataDto> {
+    const application = await this.applicationService.findById(applicationId);
+    if (!application) {
+      throw new NotFoundException(
+        `Application id ${applicationId} was not found.`,
+      );
+    }
+
+    return { data: application.data };
```
ann-aot, line 29: `application.data` is direct db entry right

andrewsignori-aot: Yes, this represents all dynamic data saved from the Student Application. We should be able to add/remove questions to the Student Application and the workflow should be able to have access to all the data.
This endpoint will not be exposed to any other client like Student/Institution.
Does it make sense?

andrewsignori-aot:
> Please review sonarcloud, other than that LGTM

Done!

ann-aot (approved):
LGTM 👍

Issue #358, opened 2021-04-12
Populate Program Information on Student Application


mzmichellesmith:
Could be up to 2000 Institution Locations

**Tasks**

- [ ] Populate Institution drop down with all institution locations
- [ ] Populate Programs information with all programs that have at least 1 offering at that location
- [ ] Populate Offerings information with all offerings belonging to that program for that location

**Technical Notes**
- [ ] Loading from the Vue application into the form definition
- [ ] Do API changes to accommodate loading

PR #524, opened 2021-06-29
[web][api][SIMS #358]Feature/populate program info 358

andrewsignori-aot:
- Populate the dropdowns for Locations, Programs, and Offerings;
- Created a new composable folder to keep Vue composition functions;
- Created a useFormioUtils to help to deal with Form.IO-related tasks, like find components and interact with them. The idea is to keep expanding this as needed.
- Created a useFormioDropdownLoader to help load dropdowns in the forms. The idea is, for instance, if we need to load a list of locations in different forms, we can reuse the same logic in different forms. Even if the logic is not going to be used in another form, it still helps reduce the amount of login in the Vue Setup method.
- Created useAuth and useInstitutionAuth to illustrate better the idea of sharing the coding using composition methods.
- We need to filter the offerings per program year also, but we are not doing it right now (not part of this PR).
- The dropdowns are listing the Locations, Programs, and Offerings names only. It will be probably changed in the future but there is no requirement right now about what we should display on each one.
- As discussed in the grooming/planning, trying to load 2000 locations at once results in something around 100Kb of data being download. Once it is deployed to DEV we can run one more download test, but after a conversation with @abschwenker, it seems that we are going to proceed with this approach for now.


ann-aot (approved):
LGTM

abschwenker (approved):
LGTM

Issue #548, opened 2021-07-07
Authentication changes for SIMS-API to Camunda and back


abschwenker:
**Additional details**
the triggering of the workflow is using a service account and we should be using the client authentication 

**Tasks**
- [x] Change the authentication to use client authentication and not the service account

PR #550, opened 2021-07-08
[api][SIMS #548] Authentication changes for SIMS-API to Camunda and back

andrewsignori-aot:
- Changed the authentication to interact with Camunda API to use the client/secret from sims-api instead of the service account.
- Removed the service account services
- Created a generic cache to be used to keep the tokens from third-party integrations like FormsFlow, Form.IO, ATBC, ect.
- Changed Form.IO authentication to use the new generic token cache.
- Added some unit tests.


ann-aot (approved):

abschwenker (approved):

guru-aot (approved):
LGTM

PR #557, opened 2021-07-14
Feature/#545-Program Year Selection when Starting FT/PT Application

guru-aot:
- Create a page when clicking "start new application" show a page with the option to select a PY based on your application start date.
- Upon selecting the year view the correct form (eg. SFAA-22/23)
- Create DB table to store year
- form definition to load
- API call

image


sources/packages/api/src-sql/ProgramYear/Create-program-year.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS program_year(
+  id SERIAL PRIMARY KEY,
+  program_year VARCHAR(100) NOT NULL,
+  program_year_desc VARCHAR(250) NOT NULL,
+  form_name VARCHAR(100) NOT NULL,
+  active_indicator BOOLEAN NOT NULL DEFAULT TRUE,
```
andrewsignori-aot, line 6: I would recommend following the same name convention already in place for the users table "is_active".

sources/packages/api/src-sql/ProgramYear/Create-program-year.sql
```diff
@@ -0,0 +1,34 @@
...
+
+  -- Audit columns
+  created_at timestamp without time zone NOT NULL DEFAULT now(),
+  updated_at timestamp without time zone NOT NULL DEFAULT now(),
+  creator INT NULL DEFAULT NULL REFERENCES users(id) ON DELETE SET NULL,
+  modifier INT NULL DEFAULT NULL REFERENCES users(id) ON DELETE SET NULL
+);
+
+-- Comments
+COMMENT ON TABLE program_year IS 'Program Year table to manage students program year of the submitted application';
```
andrewsignori-aot, line 16: I believe that the "submitted" word does not apply here.

sources/packages/api/src-sql/ProgramYear/Create-program-year.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS program_year(
+  id SERIAL PRIMARY KEY,
+  program_year VARCHAR(100) NOT NULL,
```
andrewsignori-aot, line 3: Maybe would be good to add a UNIQUE constraint here.

sources/packages/api/src-sql/ProgramYear/Create-program-year.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS program_year(
```
andrewsignori-aot, line 1: In general, the table names are plural. Can you please change the table name to "program_years" and change the rollback script as well.

sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -44,3 +44,12 @@ export class UpdateStudentContactDto {
   @IsNotEmpty()
   postalCode: string;
 }
+
+export class ProgramYearDto {
```
andrewsignori-aot, line 48: The DTO validations don't apply for response DTOs.
Please, remove the decorators and convert the class to an interface.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +140,21 @@ export class StudentController extends BaseController {
...
+
+  @Get("/program-year")
+  async getProgramYears(): Promise<ProgramYearDto[]> {
+    const programYears = await this.programYearService.getProgramYears();
+
+    if (!programYears) {
+      throw new NotFoundException(`Program Years are not found.`);
+    }
+
+    return programYears.map((programYear) => {
```
andrewsignori-aot, line 152: As a recommendation, the logic could be simplified to:
```
return programYears.map((programYear) => ({
      programYear: programYear.programYear,
      programYearDesc: programYear.programYearDesc,
      formName: programYear.formName,
    }));
```

sources/packages/web/src/services/StudentService.ts
```diff
@@ -57,4 +58,8 @@ export class StudentService {
   async synchronizeFromUserInfo(): Promise<void> {
     return await ApiClient.Students.synchronizeFromUserInfo();
   }
+
+  async getProgramYears(): Promise<ProgramYear[] | null> {
```
andrewsignori-aot, line 62: I do not see a reason to expect the return to be null in this case. We must ensure that there will be always a value for the program year, otherwise, just leave it thrown an exception, IMHO.

sources/packages/web/src/services/http/StudentApi.ts
```diff
@@ -62,4 +63,13 @@ export class StudentApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async getProgramYears(): Promise<ProgramYear[]> {
+    const response = await this.apiClient.get(
+      "students/program-year",
```
andrewsignori-aot, line 69: The program year is not a student-reated concept, it broader than (IMO). I would recommend moving it to a dedicated controller.

sources/packages/web/src/views/student/StudentDashboard.vue
```diff
@@ -25,15 +45,43 @@
...
+  setup(props: any, context: SetupContext) {
+    const programYearList = ref();
+    const onYearChange = (event: any) => {
+      context.emit("update:formName", event.value);
+      context.emit("change", event);
+    };
+    onMounted(async () => {
+      const programYears = await StudentService.shared.getProgramYears();
+      programYearList.value = programYears
+        ? programYears?.map((el: any) => ({
```
andrewsignori-aot, line 66: Please consider changing this logic with the assumption that there will be aways a program year available. Without an active program year, the system is useless. It will simplify this logic a bit.

andrewsignori-aot: I have seen many times that we are creating ".map(el" where "el" probably means "element".
I got a review about it mentioning that it would be better to have something specific instead of just "el".
Can we all follow that?
@ann-aot FYIA

andrewsignori-aot: Please assign the correct type instead of "any" since it is available.

sources/packages/web/src/views/student/StudentDashboard.vue
```diff
@@ -6,13 +6,33 @@
     <template #title>
       Welcome {{ user.givenNames }}, let's get started
     </template>
-    <template #content> Please start your Financial Aid application </template>
+    <template #content>
+      Please start your Financial Aid application
+      <br />
+      <br />
+      <Dropdown
```
andrewsignori-aot, line 13: I was expecting that this dropdown would be added to the applications page.
I think that we should point the dashboard to the applications page and add the dropdown there.

sources/packages/web/src/views/student/StudentDashboard.vue
```diff
@@ -6,13 +6,33 @@
...
       <v-btn
         color="primary"
         class="p-button-raised"
+        :disabled="!selectedForm"
         @click="
-          $router.push({ name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM })
+          $router.push({
+            name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM,
+            params: {
+              selectedForm: selectedForm,
```
andrewsignori-aot, line 33: This could be simplified to:
```
params: { selectedForm }
```

sources/packages/web/src/views/student/StudentDashboard.vue
```diff
@@ -25,15 +45,43 @@
...
 
     return {
-      user,
       StudentRoutesConst,
+      programYearList,
+      user,
+      onYearChange,
+    };
+  },
+  data(props: any) {
```
andrewsignori-aot, line 82: Can you please elaborate on why it is needed or why we cannot resolve it inside the setup method?

andrewsignori-aot (commented):
Nice work with the migrations and e2e work from Vue till API.
Please take a look at the comments, code smells, and the tests that are failing.

abschwenker (commented):
Please review tests, other than that LGTM

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -82,7 +82,7 @@ export const studentRoutes: Array<RouteRecordRaw> = [
         },
       },
       {
-        path: "application-form/:id?",
+        path: "application-form/:id?/:selectedForm",
```
ann-aot, line 85: as a recommendation , can we add the path in `https://github.com/bcgov/SIMS/blob/main/sources/packages/web/src/types/AppRoutes.ts` and call here

ann-aot (commented):
Good Work @guru-aot 👍 
Please take a look at the comments, code smells

sources/packages/web/src/services/http/StudentApi.ts
guru-aot: Restructured

sources/packages/web/src/views/student/StudentDashboard.vue
guru-aot: done

sources/packages/api/src-sql/ProgramYear/Create-initial-program-year.sql
```diff
@@ -0,0 +1,9 @@
+-- Inserting initial values for 'program_year' table. This will be part of table creation migration. So, no drop migration sql, this data will destroy with table in drop migration
+INSERT INTO program_years(program_year,program_year_desc, form_name)
+VALUES('2021-2022','Study starting between August 01, 2021 and July 31, 2022', 'SFAA2021-22');
```
andrewsignori-aot, line 3: We will need to provide Camunda a program year "code", like "2021-2022", to be used on DMN, right?
So, I believe that we need keep the format in sync like "2021-2022" vs "2021/22". Does it make sense?

sources/packages/api/src/route-controllers/program-year/program-year.controller.ts
```diff
@@ -0,0 +1,25 @@
+import { Controller, Get, NotFoundException } from "@nestjs/common";
+import { ProgramYearService } from "../../services";
+import { ProgramYearDto } from "./models/program-year.dto";
+import BaseController from "../BaseController";
+@Controller("program-year")
```
andrewsignori-aot, line 5: Please add the proper authorization. It could be only students for now and we can add institutions later when they start consuming it.

sources/packages/web/src/views/student/StudentDashboard.vue
```diff
@@ -6,34 +6,22 @@
...
 </template>
 <script lang="ts">
+import { StudentRoutesConst } from "../../constants/routes/RouteConstants";
 import { computed } from "vue";
 import { useStore } from "vuex";
-import { StudentRoutesConst } from "../../constants/routes/RouteConstants";
-
 export default {
-  setup() {
+  setup(props: any) {
```
andrewsignori-aot, line 19: Please remove the props if it is not used.

ann-aot (approved):

abschwenker (approved):

andrewsignori-aot (approved):
👍

Issue #337, opened 2021-04-06
Request Permanent Disability 


mzmichellesmith:
Students can request permanent disability status at the account level 

SABC Provides First Name, Last Name, DOB and SIN to ATBC
Permanent Disability status is pending until status is received back from AT BC

**Acceptance Criteria**
- [ ] Given (A Student), When (they want to request PD), Then (they can select Request Permanent Disability Status)
- [ ] Given (A Student), When (they select PD Status), Then, we send First Name, Last Name and DOB and SIN to AT BC and the Students PD Status is set to Pending)
- [ ] Given (ATBC), When (have confirmed PD Status), Then, (the students PD Status is set to 'PD Confirmed /' PD Denied'
- [ ] Given (ATBC), When (have confirmed PD Status), Then, (we capture the date of status change

**Tasks**
- [ ] Create button on profile to "Apply for PD status"
- [ ] Create API calls to create client (Record) 
- [ ] Create API calls to Retrieve client status
- [ ] Create table for PD Status with "PD Status" "Date Sent" "Date updated"\- [ ] Update "PD Status"
- [ ] Scheduler to Update "PD Status"
- [ ] Update on Account Creation PD lookup (Code exists for PD lookup on SFASDB on account creation)
- [ ] Create New Openshift Secrets

roopfw:
Initial Discussion Workflow: https://www.figma.com/file/lkLlMiqSYamaaHIaB5CtaW/StudentAidBC-PD-Workflows?node-id=0%3A1

abschwenker:
The pd consultation was moved from GET to POST
The client is now accepting e-mail (required), mobile phone (optional) and address (optional)

For PD consultation you need just to send the same information as before but in JSON format with the parameter id like this:
{
  "id": <value>
}

For the client this is the new format:

{
"SIN":"123456789",
"firstName": "Client",
"lastName": "Test",
"birthDate": "MM/DD/YYYY",
"email": "someone@example.com",
"mobile": "(999)-999-9999",
"address": {
"street": "1234 Fake Street",
"city": "Vancouver",
"province": "BC",
"country": "CA",
"postalCode": "V6P 2T1"
}
}

abschwenker:
PD Notes.docx in SharePoint-DEVS

PR #560, opened 2021-07-15
[web][api][sims-#337] Request Permanent Disability

ann-aot:
Students can request permanent disability status at the account level
1. Create a button on the profile to "Apply for PD status"
2. Create API calls to create client (Record) 
3. Create API calls to Retrieve client status 
4. Create a table for PD Status with  "Date Sent" "Date updated"
5.  Update "PD Status"
6. Update on Account Creation PD lookup (Code exists for PD lookup on SFASDB on account creation)
7. 1 code smells are due to TODO comment


sources/packages/api/src/database/entities/student.model.ts
```diff
@@ -54,4 +54,22 @@ export class Student extends RecordDataModel {
...
+  StudentPDSentAt?: Date;
+
+  @Column({
+    name: "pd_date_update",
+    nullable: true,
+  })
+  StudentPDUpdateAt?: Date;
+
+  @Column({
+    name: "pd_status",
```
andrewsignori-aot, line 71: What is the difference between pd_verified and pd_status?

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -64,6 +67,12 @@ export class StudentController extends BaseController {
         provinceState: existingStudent.contactInfo.addresses[0].province,
         phone: existingStudent.contactInfo.phone,
       },
+      sin: existingStudent.sin,
```
andrewsignori-aot, line 70: The student SIN should not be returned from the API.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -64,6 +67,12 @@ export class StudentController extends BaseController {
...
+      pdVerified: existingStudent.studentPDVerified,
+      validSin: existingStudent.validSIN ?? false,
```
andrewsignori-aot, line 72: When validSIN is null it means that it was not validated yet, we should not assume that it is false.
false indicates that it was validated and is not valid.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
   ): Promise<void> {
     await this.studentService.synchronizeFromUserInfo(userToken);
   }
+
+  @Get("/apply-PD-status")
```
andrewsignori-aot, line 147: Please change the HTTP verb to patch.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
...
+      const payload: ATBCCreateClientPayload = {
+        SIN: existingStudent?.sin,
+        firstName: userToken?.givenNames,
+        lastName: userToken?.lastName,
+        birthDate: userToken?.birthdate,
+        email: userToken?.email,
+      };
+      // api to create student profile in ATBC
+      const response = await this.atbcService.ATBCcreateClient(payload);
+      if (response?.code === 200) {
```
andrewsignori-aot, line 182: I believe that there is no reason for checking for 200 here. If the response is not a 200 Axios is going to throw an exception, right?

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
...
+    //check pd status in db, student should only allowed to check PD status once
+    // existingStudent?.StudentPDSentAt is set when student apply for PD Status first
+    // StudentPDStatus is null  before PD checker update status
+    // StudentPDStatus is true if PD Confirmed by ATBC
+    // StudentPDStatus is false if PD Denied by ATBC
+    // if student has a valid only, he/she should allow for a PD check
+    // if studentPDVerified is true from `SFASDB`, then no need to apply for PD to ATBC
+
+    if (
+      (existingStudent?.validSIN ?? false) &&
```
andrewsignori-aot, line 167: I believe that this logic could be simplified, like below. It will validate if not null and true.
```
(existingStudent?.validSIN)
```
Besides that, it seems that there are many redundant checks, can we narrow it down to validSIN and some other one that dictates if the request was made or not?

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
...
+      // api to create student profile in ATBC
+      const response = await this.atbcService.ATBCcreateClient(payload);
+      if (response?.code === 200) {
+        //code to update PD Sent Date
+        await this.studentService.updatePDSentDate(existingStudent?.id);
+      }
+    } else {
+      let msg = "";
+      if (!(existingStudent?.validSIN ?? false))
+        msg = `${userToken.userName}'s sin is not valid`;
```
andrewsignori-aot, line 189: If validSIN is null we should say that the validation was not done yet.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
...
+    // existingStudent?.StudentPDSentAt is set when student apply for PD Status first
+    // StudentPDStatus is null  before PD checker update status
+    // StudentPDStatus is true if PD Confirmed by ATBC
+    // StudentPDStatus is false if PD Denied by ATBC
+    // if student has a valid only, he/she should allow for a PD check
+    // if studentPDVerified is true from `SFASDB`, then no need to apply for PD to ATBC
+
+    if (
+      (existingStudent?.validSIN ?? false) &&
+      !existingStudent?.StudentPDSentAt &&
```
andrewsignori-aot, line 168: Please consider removing the null("?"). The method already ensures that the existingStudent is valid on the line 153.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
...
+      existingStudent?.StudentPDStatus === null &&
+      !existingStudent?.studentPDVerified
+    ) {
+      // create client payload
+      const payload: ATBCCreateClientPayload = {
+        SIN: existingStudent?.sin,
+        firstName: userToken?.givenNames,
```
andrewsignori-aot, line 175: I believe that our database should be the source of truth for these data. Please consider getting these data from or database.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -134,4 +143,108 @@ export class StudentController extends BaseController {
...
+      if (!(existingStudent?.validSIN ?? false))
+        msg = `${userToken.userName}'s sin is not valid`;
+      if (existingStudent?.StudentPDSentAt)
+        msg = `${userToken.userName}'s already applied for PD status`;
+      if (
+        existingStudent?.StudentPDStatus !== null ||
+        existingStudent?.studentPDVerified
+      )
+        msg = `${userToken.userName}'s PD status already captured`;
+      throw new NotAcceptableException(msg);
```
andrewsignori-aot, line 197: I don`t believe that NotAcceptableException applies here, please check if it makes sense; https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
I would recommend a UnprocessableEntityException or BadRequestException.

andrewsignori-aot: In general, do we need to provide this level of detail in the exception? I would recommend log this on API and just throw the exception with a generic message.
The user is only supposed to make this call if he fits into the proper conditions to do it. The UI will not show the option if the user is not allowed to execute it.
If this method is executed and the user does not qualifies for it, it must be because he tries to bypass the UI rules in some way.

sources/packages/api/src/database/entities/student.model.ts
ann-aot: pd_verified is from sfas DB and pd_status is from atbc

sources/packages/api/src/route-controllers/student/student.controller.ts
ann-aot: oh..okay. Thanks for that (Y)

sources/packages/api/src/database/entities/student.model.ts
abschwenker: Sorry for the confusion, we can assume that we will use 1 status column, no need for a 2nd. pd_verified that comes from sfasDB does actually come from ATBC. So 1 field should be fine. Thoughts?

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -174,4 +173,39 @@ export class StudentService extends RecordDataModelService<Student> {
...
+  /**
+   * Update the PD Sent Date
+   * @param studentId
+   */
+  async updatePDSentDate(studentId: number): Promise<Student> {
+    // get the Student Object
+    const studentToUpdate = await this.repo.findOne({
+      id: studentId,
+    });
+    if (studentToUpdate) {
```
andrewsignori-aot, line 186: I would recommend throwing an exception if the student is not found, for instance calling findOneOrFail.
If the student is not found it will execute quietly, but it seems to be an error.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -174,4 +173,39 @@ export class StudentService extends RecordDataModelService<Student> {
...
+   */
+  async updatePDStatusNDate(
+    studentId: number,
+    status: boolean,
+  ): Promise<Student> {
+    // get the Student Object
+    const studentToUpdate = await this.repo.findOne({
+      id: studentId,
+    });
+    if (studentToUpdate) {
```
andrewsignori-aot, line 205: I would recommend throwing an exception if the student is not found, for instance calling findOneOrFail.
If the student is not found it will execute quietly, but it seems to be an error.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -174,4 +173,39 @@ export class StudentService extends RecordDataModelService<Student> {
...
+   * Update the PD Sent Date
+   * @param studentId
+   */
+  async updatePDSentDate(studentId: number): Promise<Student> {
+    // get the Student Object
+    const studentToUpdate = await this.repo.findOne({
+      id: studentId,
+    });
+    if (studentToUpdate) {
+      studentToUpdate.StudentPDSentAt = new Date();
```
andrewsignori-aot, line 187: I believe that we should save the Data as UTC. Can you check please?

sources/packages/api/src/types/config.ts
```diff
@@ -87,3 +88,12 @@ export interface CRAIntegrationConfig {
   programAreaCode: string;
   environmentCode: string;
 }
+
+export interface ATBCIntegrationConfig {
+  ATBCClientCreateEndpoint: string;
```
andrewsignori-aot, line 93: Usually I believe that we are creating the env variables for the root URL and defining the endpoints based on the root URL. Is there a need to have the endpoints also as env variables?

sources/packages/api/src/database/entities/student.model.ts
ann-aot: okay, that make sense

sources/packages/api/src/types/config.ts
ann-aot: since it's an external API endpoint, there may be a chance of change in endpoint right, and if there is any scenario, without any code changes, we can update the endpoint. if that doesn't make sense, then I can update the code

andrewsignori-aot: I got the point but, if they change the endpoints, we will be changing at least our CICD to accommodate the new endpoint names. When we go for production we should not expect such changes to happen, otherwise, it would be a major backward compatibility issue. I would recommend keeping just the root URL as a parameter that is what is going to be different per environment. Always open to chat about it if you prefer :smile:

sources/packages/api/src/database/entities/student.model.ts
```diff
@@ -54,4 +54,16 @@ export class Student extends RecordDataModel {
     referencedColumnName: ColumnNames.ID,
   })
   user: User;
+
+  @Column({
+    name: "pd_date_sent",
+    nullable: true,
+  })
+  StudentPDSentAt?: Date;
```
andrewsignori-aot, line 62: Please follow the same naming conventions from others properties starting the names with lower case.

sources/packages/api/src/database/entities/student.model.ts
```diff
@@ -54,4 +54,16 @@ export class Student extends RecordDataModel {
...
+
+  @Column({
+    name: "pd_date_update",
+    nullable: true,
+  })
+  StudentPDUpdateAt?: Date;
```
andrewsignori-aot, line 68: Please follow the same naming conventions from others properties starting the names with lower case.

sources/packages/api/src/route-controllers/atbc/atbc.system.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+        await this.studentService.getStudentsAppliedForPD();
+
+      // Executes the processing of each student in parallel.
+      const pdProcess = studentAppliedPD.map((eachStudent) =>
+        this.atbcService.PDCheckerProcess(eachStudent),
+      );
+      // Waits for all the parallel processes to be finished.
+      await Promise.all(pdProcess);
+      return {
+        message: "PD checker ran successfully",
```
andrewsignori-aot, line 41: I would recommend not return the status messages from API.
I believe that the API should return the proper HTTP status and the Vue application could display the message accordingly. I would remove the ATBCPDResponseDto.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -156,6 +162,53 @@ export class StudentController extends BaseController {
...
+      throw new NotFoundException(
+        `No user record was found with for student ${userToken.userName}`,
+      );
+    }
+    //check pd status in db, student should only allowed to check PD status once
+    // existingStudent?.StudentPDSentAt is set when student apply for PD Status first
+    // studentPDVerified is null  before PD checker update status
+    // studentPDVerified is true if PD Confirmed by ATBC OR is true from `SFASDB
+    // studentPDVerified is false if PD Denied by ATBC
+    // if student has a valid only, he/she should allow for a PD check
```
andrewsignori-aot, line 190: Should the SIN be mentioned here: "if student has a valid SIN only"?

sources/packages/api/src/route-controllers/atbc/atbc.system.controller.ts
ann-aot: Andrew, this function is for camunda scheduler, Vue application is not calling this

sources/packages/api/src/services/atbc/atbc.controller.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+import { StudentService, ArchiveDbService, ATBCService } from "../../services";
+import { KeycloakConfig } from "../../auth/keycloakConfig";
+import { KeycloakService } from "../auth/keycloak/keycloak.service";
+import { ConfigService } from "../config/config.service";
+import { HttpStatus, INestApplication } from "@nestjs/common";
+import { Test, TestingModule } from "@nestjs/testing";
+import * as request from "supertest";
+import { ATBCCreateClientResponse } from "../../types";
+
+describe("Test student model", () => {
```
andrewsignori-aot, line 15: Please consider renaming this to something for meaningful for ATBC context, like "Test ATBC Controller"

sources/packages/api/src/services/atbc/atbc.controller.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      {},
+    ).compile();
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+  afterAll(async () => {
+    await closeDB();
+  });
+
+  it("should save student model object with user relationship and address jsonb", async () => {
```
andrewsignori-aot, line 42: I believe that the main validation on this test would be the "apply-pd-status". Please consider changing it to something like: "should return an HTTP 200 status when applying for PD and student is valid"

sources/packages/api/src/services/atbc/atbc.controller.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    };
+    const simsUser = new User();
+    simsUser.userName = faker.random.uuid();
+    simsUser.email = faker.internet.email();
+    simsUser.firstName = faker.name.firstName();
+    simsUser.lastName = faker.name.lastName();
+    fakestudent.user = simsUser;
+
+    // Save the student in SIMS
+    await studentService.save(fakestudent);
```
andrewsignori-aot, line 71: Here the student is being saved but the "getStudentByUserName" is also being mocked. Are both being used?
I believe that the mocked method will make the database not be used at all.

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    private readonly studentService: StudentService,
+  ) {
+    this.logger.log("[Created]");
+  }
+  /**
+   * Allow a singleton access to the ATBCService for areas
+   * of the application that are not part of the dependency
+   * injection framework.
+   */
+  public static get shared(): ATBCService {
```
andrewsignori-aot, line 33: With a few exceptions, services created to be used alongside Nestjs do not need to have singletons created. Please remove it and ensure that it is being injected.
By default, the services injected by Nest will be singletons unless that we do not want it and change its default behavior using decorators.
More info: https://docs.nestjs.com/fundamentals/injection-scopes

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    return ATBCService.shared;
+  }
+
+  get config(): ATBCIntegrationConfig {
+    return this.configService.getConfig().ATBCIntegration;
```
andrewsignori-aot, line 38: If there is no reason for this be public, please consider keeping it private for now.

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+   * Executes the token request
+   * @returns bearer header.
+   */
+  private async getConfig(): Promise<ATBCHeader> {
+    const token = await this.getAuthToken();
+    return {
+      headers: { Authorization: `Bearer ${token}` },
+      // (NOTE: this will disable client verification)
+      // TODO: add certificate for PROD
+      httpsAgent: new (require("https").Agent)({
```
andrewsignori-aot, line 50: This is really important and we cannot forget about it. Please create a tech debit for this.

sources/packages/api/src/services/atbc/atbc.controller.spec.ts
ann-aot: `getStudentByUserName` is mocked because the username is from BCSC, the fake user created in the DB doesn't have a username from BCSC, that is why it is mocked. The fake user in DB is created to update the PD sent date, `studentPDSentAt` column

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Create client on ATBC.
+   * @returns the result of a success full authentication or throws an exception
+   * in case the result is anything different from HTTP 200 code.
+   */
+  public async ATBCCreateClient(
+    payload: ATBCCreateClientPayload,
+  ): Promise<ATBCCreateClientResponse> {
+    try {
+      const config = await this.getConfig();
```
andrewsignori-aot, line 102: It seems that this variable is not being used, can you check and remove it, please.

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+      const apiEndpoint =
+        this.config.ATBCEndpoint.slice(-1) === "/"
```
andrewsignori-aot, line 104: I would recommend do not add more complexity to check the "/" every time. It will not be good repeating this logic in all places needed. Please consider that the root endpoint will not have the "/" or resolve this logic on config service to be centralized.

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+  private async ATBCPDChecker(
+    payload: ATBCPDCheckerPayload,
+  ): Promise<ATBCPDCheckerResponse> {
+    try {
+      const config = await this.getConfig();
+      const apiEndpoint =
+        this.config.ATBCEndpoint.slice(-1) === "/"
+          ? `${this.config.ATBCEndpoint}pd`
+          : `${this.config.ATBCEndpoint}/pd`;
+      const res = await axios.post(`${apiEndpoint}`, payload, config);
```
andrewsignori-aot, line 129: No need for string interpolation here since it is just one variable.

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+      // try {
+      const response = await this.ATBCPDChecker(payload);
+      // e9yStatusId === 1 , PD Confirmed
+      // e9yStatusId === 2 , PD Denied
+      // e9yStatusId === 0 , Processing, In Queue
+
+      this.logger.log(
+        `PD Status for student ${eachStudent.id}, status ${response?.e9yStatusId} ${response?.e9yStatus}, `,
+      );
+      if (response?.e9yStatusId === 1 || response?.e9yStatusId === 2) {
```
andrewsignori-aot, line 158: Please avoid doing the comparison with numbers. Considering creating an enum to keep the below codes.
The comments help, but the values should be better expressed in enum/constants.
// e9yStatusId === 1 , PD Confirmed
// e9yStatusId === 2 , PD Denied
// e9yStatusId === 0 , Processing, In Queue

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+      // api to check the student PD status in ATBC
+      this.logger.log(`Check for PD status of student ${eachStudent.id}`);
+      // try {
+      const response = await this.ATBCPDChecker(payload);
+      // e9yStatusId === 1 , PD Confirmed
+      // e9yStatusId === 2 , PD Denied
+      // e9yStatusId === 0 , Processing, In Queue
+
+      this.logger.log(
+        `PD Status for student ${eachStudent.id}, status ${response?.e9yStatusId} ${response?.e9yStatus}, `,
```
andrewsignori-aot, line 156: Nice work logging the processing, can you plase add something when the status is "Processing, In Queue"?

sources/packages/api/src/services/atbc/atbc.service.ts
```diff
@@ -0,0 +1,180 @@
...
+  public async PDCheckerProcess(eachStudent: Student) {
+    try {
+      this.logger.log(`Creating Payload for student id ${eachStudent.id}`);
+      // create PD checker payload
+      const payload: ATBCPDCheckerPayload = {
+        id: eachStudent.sin,
+      };
+      // api to check the student PD status in ATBC
+      this.logger.log(`Check for PD status of student ${eachStudent.id}`);
+      // try {
```
andrewsignori-aot, line 149: Is this some leftover comment?

guru-aot (commented):
Please check on the merge conflicts with @andrewsignori-aot in file sources/packages/api/src/route-controllers/student/student.controller.ts

sources/packages/api/src/services/auth/token-cache.service.models.ts
```diff
@@ -2,3 +2,11 @@ export interface TokenCacheResponse {
   accessToken: string;
   expiresIn: number;
 }
+export interface ATBCAuthTokenResponse {
+  accessToken: string;
+}
+
+export interface ATBCHeader {
+  headers?: { Authorization: string };
```
andrewsignori-aot, line 10: I believe that headers should not be optional, right? Please consider removing the "?".

sources/packages/api/src/services/config/config.service.ts
```diff
@@ -56,6 +56,13 @@ export class ConfigService {
         programAreaCode: process.env.CRA_PROGRAM_AREA_CODE,
         environmentCode: process.env.CRA_ENVIRONMENT_CODE,
       },
+      ATBCIntegration: {
+        ATBCloginEndpoint: process.env.ATBC_LOGIN_ENDPOINT,
```
andrewsignori-aot, line 60: Please keep the same pattern as the others in the same object: "ATBCLoginEndpoint"

sources/packages/api/src/types/config.ts
```diff
@@ -87,3 +88,11 @@ export interface CRAIntegrationConfig {
   programAreaCode: string;
   environmentCode: string;
 }
+
+export interface ATBCIntegrationConfig {
+  ATBCloginEndpoint: string;
```
andrewsignori-aot, line 93: Please consider keeping the same pattern from other variables in the same object: "ATBCLoginEndpoint".

sources/packages/web/src/services/StudentService.ts
```diff
@@ -57,4 +58,8 @@ export class StudentService {
   async synchronizeFromUserInfo(): Promise<void> {
     return await ApiClient.Students.synchronizeFromUserInfo();
   }
+
+  async applyForPDStatus() {
```
andrewsignori-aot, line 62: Please make the return type explicit.

sources/packages/web/src/services/http/StudentApi.ts
```diff
@@ -62,4 +62,17 @@ export class StudentApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async applyForPDStatus() {
```
andrewsignori-aot, line 66: Please make the return type explicit.

sources/packages/web/src/types/contracts/StudentContract.ts
```diff
@@ -49,3 +54,8 @@ export interface StudentProfile {
   country: string;
   postalCode: string;
 }
+
+export interface StudentFormInfo extends StudentInfo {
+  birthDateFormatted: string;
```
andrewsignori-aot, line 59: Can you please elaborate on where this was needed for this particular ticket?

sources/packages/web/src/views/student/StudentProfile.vue
```diff
@@ -5,6 +5,48 @@
...
+        <ProgressSpinner style="width:30px;height:25px" strokeWidth="10"
+      /></span>
+    </v-btn>
+  </span>
+  <span v-else>
+    <Message
+      severity="warn"
+      :closable="false"
+      v-if="
+        studentAllInfo?.pdSentDate &&
```
andrewsignori-aot, line 27: Just as a recommendation, can we move this logic to a computed property in the setup method to simplify the Vue template?

andrewsignori-aot: Do we need the null propagation check for these properties?
Is there a scenario where they will not be present? If not, please remove the "?".

sources/packages/web/src/views/student/StudentProfile.vue
```diff
@@ -5,6 +5,48 @@
...
+          studentAllInfo?.pdUpdatedDate === null &&
+          studentAllInfo?.pdVerified === null
+      "
+    >
+      <strong>PD Status: Pending</strong>
+    </Message>
+    <Message
+      severity="success"
+      :closable="false"
+      v-if="studentAllInfo?.pdVerified === true"
```
andrewsignori-aot, line 37: While demoing it, please remember to mention that students validated through SFAS DB will also be displaying the status in the same way 😉

sources/packages/web/src/views/student/StudentProfile.vue
```diff
@@ -5,6 +5,48 @@
...
+    >
+      Apply for PD status
+      <span v-if="disableBtn">
+        &nbsp;&nbsp;
+        <ProgressSpinner style="width:30px;height:25px" strokeWidth="10"
+      /></span>
+    </v-btn>
+  </span>
+  <span v-else>
+    <Message
```
andrewsignori-aot, line 23: Should we add a message for when the SIN is not validated yet and the student must wait for this process to happen?

sources/packages/web/src/views/student/StudentProfile.vue
```diff
@@ -56,15 +99,53 @@ export default {
...
+      } catch (error) {
+        toast.add({
+          severity: "error",
+          summary: "Unexpected error",
+          detail:
+            "An error happened during the apply PD process. Please try after sometime.",
+          life: 5000,
+        });
+      }
+      await getStudentInfo();
```
andrewsignori-aot, line 138: The API calls could happen in parallel here. Please consider changing it.

sources/packages/api/src/route-controllers/atbc/atbc.system.controller.ts
andrewsignori-aot: Yes, I mentioned Vue, my bad, but I don`t see a reason to return a message to Camunda here. I believe that HTTP status should be ok.

sources/packages/api/src/services/atbc/atbc.service.ts
ann-aot: actually, `${response?.e9yStatus}` is has the status from API response
image

sources/packages/api/src/services/atbc/atbc.service.ts
ann-aot: no, maybe i can edit the comment like, ``Check starting for PD status of student ${eachStudent.id}``

sources/packages/api/src/services/atbc/atbc.controller.spec.ts
andrewsignori-aot: The user name on token is supposed to be "student_e2e_test". If the fake user name is created using this username it should be aligned with what the UserToken will be receiving. Does it make sense?
The token, in the last instance has the "key cloack user name and info" not BCSC. Whet happens is, when the user is authenticated using BCSC, a Key Clock user is created using BCSC info.

sources/packages/api/src/services/auth/token-cache.service.models.ts
ann-aot: I am resuing the interface for login and other ATBC endpoints, okay, I will create separate header interface

sources/packages/web/src/types/contracts/StudentContract.ts
ann-aot: https://github.com/bcgov/SIMS/blob/main/sources/packages/web/src/views/student/Student.vue#L200 is calling https://github.com/bcgov/SIMS/blob/main/sources/packages/web/src/services/StudentService.ts#L42 (this code was already there without interface, I just added interface, since it was throwing me errors)

sources/packages/web/src/views/student/StudentProfile.vue
abschwenker: We will have further requirements on what a student can do when their SIN is not validated. Not required at this time but there will be a requirement for it in the future

ann-aot: that's a good point, @abschwenker  do we need to show msg to student if their SIN is not validated yet

sources/packages/web/src/views/student/StudentProfile.vue
ann-aot: actually, `getStudentInfo` function update `studentAllInfo` variable  and this variable is used in `appliedPDButton` function, that why, I added await,

sources/packages/api/src/services/atbc/atbc.service.ts
andrewsignori-aot: Yes, make sense.

sources/packages/api/src/services/atbc/atbc.service.ts
andrewsignori-aot: Nice, thanks for clarifying it 😄

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -177,4 +177,66 @@ export class StudentService extends RecordDataModelService<Student> {
...
+  async updatePDSentDate(studentId: number): Promise<Student> {
+    // get the Student Object
+    const studentToUpdate = await this.repo.findOneOrFail({
+      id: studentId,
+    });
+    if (studentToUpdate) {
+      const now = new Date();
+      // Date in UTC
+      studentToUpdate.studentPDSentAt = new Date(
+        now.getUTCFullYear(),
```
andrewsignori-aot, line 194: Please use the dayjs library already in API to do this type of manipulation: https://day.js.org/docs/en/plugin/utc

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -177,4 +177,66 @@ export class StudentService extends RecordDataModelService<Student> {
...
+    // get the Student Object
+    const studentToUpdate = await this.repo.findOneOrFail({
+      id: studentId,
+    });
+    if (studentToUpdate) {
+      const now = new Date();
+      studentToUpdate.studentPDVerified = status;
+      // Date in UTC format
+      studentToUpdate.studentPDUpdateAt = new Date(
+        now.getUTCFullYear(),
```
andrewsignori-aot, line 223: Please use the dayjs library already in API to do this type of manipulation: https://day.js.org/docs/en/plugin/utc

sources/packages/web/src/views/student/StudentProfile.vue
andrewsignori-aot: await must be here but, since there is no dependency between getStudentInfo and appliedPDButton you can do something like:
```
const getStudentInfoResult = getStudentInfo(); 
const appliedPDButtonResult = appliedPDButton();
await Promise.All([getStudentInfoResult , appliedPDButtonResult ]);
```

andrewsignori-aot (approved):
Great work with tests and all 👍

guru-aot (approved):
Nice work @ann-aot

abschwenker (approved):
Please review tests, other than that, lgtm

Issue #517, opened 2021-06-21
Review & Improve APIs - restrict institution Admin from accessing/ editing locations or location related features of other institutions


ann-aot:
- [x] restrict institution Admin from accessing/ editing locations or location related features of other institutions

PR #579, opened 2021-07-23
[API][sims-#517]  restrict institution Admin from accessing/ editing locations or location related features of other institutions

ann-aot:
1. Restrict institution Admin from accessing/ editing locations or location related features of other institutions
2.  Restrict institution Admin from accessing/ editing/ creating offerings, and programs related features of other institutions

### DIFF
--- a/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
+++ b/sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
@@ -6,23 +6,35 @@ import {
   Param,
   Get,
   UnprocessableEntityException,
+  ForbiddenException,
 } from "@nestjs/common";
 import { AuthorizedParties } from "../../auth/authorized-parties.enum";
-import { AllowAuthorizedParty, HasLocationAccess } from "../../auth/decorators";
+import {
+  AllowAuthorizedParty,
+  HasLocationAccess,
+  UserToken,
+} from "../../auth/decorators";
 import {
   SaveEducationProgramOfferingDto,
   EducationProgramOfferingDto,
   ProgramOfferingDto,
 } from "./models/education-program-offering.dto";
 import { FormNames } from "../../services/form/constants";
-import { EducationProgramOfferingService, FormService } from "../../services";
+import {
+  EducationProgramOfferingService,
+  FormService,
+  InstitutionLocationService,
+  EducationProgramService,
+} from "../../services";
 import { OptionItem } from "../../types";
-
+import { IInstitutionUserToken } from "../../auth/userToken.interface";
 @Controller("institution/offering")
 export class EducationProgramOfferingController {
   constructor(
     private readonly programOfferingService: EducationProgramOfferingService,
     private readonly formService: FormService,
+    private readonly locationService: InstitutionLocationService,
+    private readonly programService: EducationProgramService,
   ) {}
 
   @AllowAuthorizedParty(AuthorizedParties.institution)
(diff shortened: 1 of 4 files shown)


sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -32,7 +45,23 @@ export class EducationProgramOfferingController {
...
+      userToken.authorizations.institutionId !== requestedLoc.institution.id
+    ) {
+      throw new ForbiddenException();
+    }
+    const requestProgram = await this.programService.getInstitutionProgram(
+      programId,
+      userToken.authorizations.institutionId,
+    );
+    if (!requestProgram) {
+      throw new ForbiddenException();
```
andrewsignori-aot, line 63: At this point, the institution access was already validated, I would recommend a throw a NotFoundException.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -32,7 +45,23 @@ export class EducationProgramOfferingController {
     @Body() payload: SaveEducationProgramOfferingDto,
     @Param("locationId") locationId: number,
     @Param("programId") programId: number,
+    @UserToken() userToken: IInstitutionUserToken,
   ): Promise<number> {
+    const requestedLoc = await this.locationService.getInstitutionLocationById(
+      locationId,
+    );
+    if (
+      userToken.authorizations.institutionId !== requestedLoc.institution.id
```
andrewsignori-aot, line 54: The fixes in this way look good to complete this ticket, but I believe that we should move this check inside the InstitutionLocationGuard. What is happening now is that we are not checking the locations for admins.
To fix it in the InstitutionLocationGuard, I believe that we should.

1. Load the locations Ids when the user is an Admin on JwtStrategy
2. Remove the below code from InstitutionLocationGuard and let the location ids be validated for Admin also.
```
if (authorizations.isAdmin()) {
   return true;
}
```

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -24,6 +24,20 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+    return this.repo
+      .createQueryBuilder("institution_location")
+      .select([
+        "institution_location.name",
+        "institution_location.data",
+        "institution_location.id",
+        "institution",
+      ])
+      .leftJoin("institution_location.institution", "institution")
+      .where("institution_location.id = :id", { id: id })
```
andrewsignori-aot, line 38: This could be simplified as below:
```
.where("institution_location.id = :id", { id })
```

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -24,6 +24,20 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+  async getInstitutionLocationById(id: number): Promise<InstitutionLocation> {
+    return this.repo
+      .createQueryBuilder("institution_location")
+      .select([
+        "institution_location.name",
+        "institution_location.data",
+        "institution_location.id",
+        "institution",
+      ])
+      .leftJoin("institution_location.institution", "institution")
```
andrewsignori-aot, line 37: It seems that there is no need to return the entire institution record here. Please consider returning only what is need for now.
I believe that you can also use the leftJoinAndSelect, if you think that it would be better.

abschwenker (approved):
LGTM, once reviews are done please merge

andrewsignori-aot (approved):
👍

guru-aot (approved):
LGTM

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -32,7 +46,23 @@ export class EducationProgramOfferingController {
     @Body() payload: SaveEducationProgramOfferingDto,
     @Param("locationId") locationId: number,
     @Param("programId") programId: number,
+    @UserToken() userToken: IInstitutionUserToken,
   ): Promise<number> {
+    const requestedLoc = await this.locationService.getInstitutionLocationById(
+      locationId,
+    );
+    if (
```
dheepak-aot, line 54: Code from line 54-58 has been repeated in multiple places. 
Hence this piece of code can be added to a method and called. It will reduce the redundant code.

PR #580, opened 2021-07-26
Feature/#526 notice of assessment - API only

guru-aot:
- Determine table/Column for assessment
- Create Migration for "assessment" table/Column
- Create API for "assessment" table SAVE (for Camunda to save to SIMSDB)
- API call to get Assessment Details for Vue application

Only the Save and Get API for the assessment is created

### DIFF
--- a/sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
+++ b/sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
@@ -5,13 +5,13 @@ export class UpdateApplicationsAddAssessment1626994719103 implements MigrationIn
 
     public async up(queryRunner: QueryRunner): Promise<void> {
         await queryRunner.query(
-            getSQLFileData("Add-col-assessment", "Applications"),
+            getSQLFileData("Add-col-assessment.sql", "Applications"),
           );
     }
 
     public async down(queryRunner: QueryRunner): Promise<void> {
         await queryRunner.query(
-            getSQLFileData("Drop-col-assessment", "Applications"),
+            getSQLFileData("Drop-col-assessment.sql", "Applications"),
           );
     }
 
--- a/sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
+++ b/sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
@@ -1,0 +1,18 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+import { getSQLFileData } from "../../utilities";
+
+export class UpdateApplicationsAddAssessment1626994719103 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            getSQLFileData("Add-col-assessment", "Applications"),
+          );
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            getSQLFileData("Drop-col-assessment", "Applications"),
+          );
+    }
+
+}
--- a/sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
+++ b/sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
@@ -1,18 +1,17 @@
-import {MigrationInterface, QueryRunner} from "typeorm";
+import { MigrationInterface, QueryRunner } from "typeorm";
 import { getSQLFileData } from "../../utilities";
 
-export class UpdateApplicationsAddAssessment1626994719103 implements MigrationInterface {
+export class UpdateApplicationsAddAssessment1626994719103
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Add-col-assessment.sql", "Applications"),
+    );
+  }
 
-    public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(
-            getSQLFileData("Add-col-assessment.sql", "Applications"),
-          );
-    }
-
-    public async down(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(
-            getSQLFileData("Drop-col-assessment.sql", "Applications"),
-          );
-    }
-
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Drop-col-assessment.sql", "Applications"),
+    );
+  }
 }
--- a/sources/packages/api/src/route-controllers/application/application.system.controller.ts
+++ b/sources/packages/api/src/route-controllers/application/application.system.controller.ts
@@ -1,6 +1,6 @@
-import { Controller, Get, NotFoundException, Param } from "@nestjs/common";
+import { Controller, Get, Patch, NotFoundException, Param, Body } from "@nestjs/common";
 import { ApplicationService } from "../../services";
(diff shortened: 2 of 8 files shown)


sources/packages/api/src/database/migrations/1626994719103-UpdateApplicationsAddAssessment.ts
```diff
@@ -0,0 +1,18 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+import { getSQLFileData } from "../../utilities";
+
```
andrewsignori-aot, line 3: The formatter doesn't seem to be working, please check and format the file accordingly.
Please check on others files on API also.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -28,4 +28,32 @@ export class ApplicationSystemController {
 
     return { data: application.data };
   }
+
+  @Patch(":applicationId/assessment")
+  async updateAssessmentInApplication(
+    @Body() assessment: ApplicationAssessmentDTO,
+    @Param("applicationId") applicationId: number,
+  ): Promise<number> {
+    const application = await this.getByApplicationId(applicationId);
```
andrewsignori-aot, line 37: I would recommend do not retrieve the entire record to just update on a particular column.
Since we have the application id and the value to be updated, we can just update the column, IMO.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -28,4 +28,32 @@ export class ApplicationSystemController {
...
+    const updateAssessmentInApplication = await this.applicationService.updateAssessmentInApplication(
+      applicationId,
+      application,
+      assessment,
+    );
+    return updateAssessmentInApplication.affected;
+  }
+
+  @Get(":applicationId/assessment")
```
andrewsignori-aot, line 46: Will this method be consumed by Camunda?
It seems that this would be a method being consumed by Vue application. If yes, please move to the correct controller with the proper authorization.

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -16,3 +16,30 @@ export interface GetApplicationDataDto {
    */
   data: any;
 }
+
+export interface ApplicationAssessmentDTO {
```
andrewsignori-aot, line 20: IMO, since all this data will be flexible, I would not mind not express all individual properties.
If all the data will be consumed only by a Form.IO form and only Form.IO form will know the expected fields, I would say to express it as flexible as possible.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
guru-aot: yeah, I will move it

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -53,4 +58,42 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      total_provincial_award: assessment.total_provincial_award,
+    };
+    return application;
+  }
+
+  async getAssessmentByApplicationId(applicationId: number): Promise<any> {
+    return this.repo
+      .createQueryBuilder("application")
+      .select("application.assessment")
+      .where("application.id = :applicationIdParam", {
```
andrewsignori-aot, line 94: This could be simplified as below:
```
.where("application.id = :applicationId", { applicationId })
```

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -53,4 +58,42 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      weeks: assessment.weeks,
+      federal_assessment_need: assessment.federal_assessment_need,
+      provincial_assessment_need: assessment.provincial_assessment_need,
+      total_federal_award: assessment.total_federal_award,
+      total_provincial_award: assessment.total_provincial_award,
+    };
+    return application;
+  }
+
+  async getAssessmentByApplicationId(applicationId: number): Promise<any> {
```
andrewsignori-aot, line 90: Please express what is being returned instead of any in this case.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -53,4 +58,42 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+  async updateAssessmentInApplication(
+    applicationId: number,
+    application: GetApplicationDataDto,
+    assessment: ApplicationAssessmentDTO,
+  ): Promise<UpdateResult> {
+    const populatedApplication = await this.populateApplication(
+      application,
+      assessment,
+    );
+    return this.repo.update(applicationId, populatedApplication);
```
andrewsignori-aot, line 71: The code doesn't need all the applications from the database to perform an update.
Please consider performing the update in some way similar to the below.
```
return this.repo.update(applicationId, { assessment });
```

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -53,4 +58,42 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+  }
+
+  async populateApplication(
```
andrewsignori-aot, line 74: Please consider removing this method.

andrewsignori-aot (changes requested):

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -16,3 +16,30 @@ export interface GetApplicationDataDto {
...
+export class ApplicationDto {
+  /**
+   * Application dynamic data.
+   */
+  data: any;
+  /**
+   * Array of unique file names to be associated
+   * with this application.
+   */
+  associatedFiles: string[];
```
andrewsignori-aot, line 40: Can you please elaborate on why data and associatedFiles are declared here?

sources/packages/api/src/route-controllers/application/models/application.model.ts
guru-aot: I feel it's better to fetch it individually, as its easy to show case in the VUE application. Anyhow, when we need to show additional columns, we need to do a code change irrespective on the form side.

sources/packages/api/src/route-controllers/application/models/application.model.ts
guru-aot: Removed it

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -53,4 +56,29 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+  async getAssessmentByApplicationId(
+    applicationId: number,
+  ): Promise<ApplicationAssessmentDTO> {
+    const application = await this.repo
+      .createQueryBuilder("application")
+      .select("application.assessment")
+      .where("application.id = :applicationId", { applicationId })
+      .getOne();
+    return {
+      weeks: application.assessment.weeks,
```
andrewsignori-aot, line 76: We should not be expressing these as they are supposed to be dynamic and flexible.

sources/packages/api/src/route-controllers/application/models/application.model.ts
andrewsignori-aot: Not quite agree. When we need more column/data, we can change the workflow to save more data and we can change the Form.IO definition, no Vue code change is needed at all.
Making those expressed in this way seems to be against what we talked about in the meeting where we defined it.

sources/packages/api/src/services/application/application.service.ts
guru-aot: Do you want to return any? i.e. the assessment object itself?

andrewsignori-aot (approved):
👍

abschwenker (approved):

ann-aot (approved):
LGTM 👍

Issue #530, opened 2021-06-29
Create Student Application Number


abschwenker:
**Describe the task**
When an application is submitted create a DB generated value (to ensure no duplicates are created)

**Acceptance Criteria**
- [x] Application number must 10 Characters
- [x] Application number must be in the format ((Program year/Program year) eg. 2122) YYYY######

**Additional context**
- Add any other context about the task here.
- Or here

PR #587, opened 2021-07-26
[api][sims-#530] Application Number added

ann-aot:
When an application is submitted create a DB generated value (to ensure no duplicates are created)
1. Application number must 10 Characters
2. Application number must be in the format ((Program year/Program year) eg. 2122) YYYY######
3. Code Smell is due to TODO (for this ticket program year is static, we need to make it dynamic)

### DIFF
--- a/sources/packages/api/src-sql/ApplicationNumber/Create-application-number.sql
+++ b/sources/packages/api/src-sql/ApplicationNumber/Create-application-number.sql
@@ -1,0 +1,7 @@
+
+-- Add the column application_number for application.
+ALTER TABLE applications ADD 
+    COLUMN IF NOT EXISTS application_number VARCHAR(10);
+
+-- ## Comments
+COMMENT ON COLUMN applications.application_number IS 'Unique 10 digit application number';
--- a/sources/packages/api/src-sql/ApplicationNumber/Create-application-number.sql
+++ b/sources/packages/api/src-sql/ApplicationNumber/Create-application-number.sql
@@ -1,7 +1,7 @@
 
 -- Add the column application_number for application.
 ALTER TABLE applications ADD 
-    COLUMN IF NOT EXISTS application_number VARCHAR(10);
+    COLUMN IF NOT EXISTS application_number CHAR(10);
 
 -- ## Comments
 COMMENT ON COLUMN applications.application_number IS 'Unique 10 digit application number';
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -23,6 +23,12 @@ export class Application extends RecordDataModel {
   })
   data: any;
 
+  @Column({
+    name: "application_number",
+    nullable: false,
+  })
+  applicationNumber: string;
+
   @OneToOne(() => Student, { eager: false, cascade: true })
   @JoinColumn({
     name: "student_id",
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -25,7 +25,6 @@ export class Application extends RecordDataModel {
 
   @Column({
     name: "application_number",
-    nullable: false,
   })
   applicationNumber: string;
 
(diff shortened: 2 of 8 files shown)


sources/packages/api/src-sql/ApplicationNumber/Create-application-number.sql
```diff
@@ -0,0 +1,7 @@
+
+-- Add the column application_number for application.
+ALTER TABLE applications ADD 
+    COLUMN IF NOT EXISTS application_number VARCHAR(10);
```
andrewsignori-aot, line 4: IMO, a length shorter than 10 would be inconsistent. I believe that the type should be CHAR(10) instead.

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -23,6 +23,12 @@ export class Application extends RecordDataModel {
   })
   data: any;
 
+  @Column({
+    name: "application_number",
+    nullable: false,
```
andrewsignori-aot, line 28: So far, we believe that the application number will be generated only during application submission what means that while saving a draft application the number will not be generated and the column must accept null.
I would recommend removing the "nullable: false" to not lead a developer to false assumptions.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -86,13 +88,26 @@ export class ApplicationController extends BaseController {
         payload.associatedFiles,
       );
     }
-
+    // TODO:remove the static program year and add dynamic year, from program year table
```
andrewsignori-aot, line 91: I would recommend moving this logic inside the ApplicationService to keep the controller cleaner.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -86,13 +88,26 @@ export class ApplicationController extends BaseController {
...
+      sequenceName,
+      async (nextSequenceNumber: number) => {
+        nextApplicationSequence = nextSequenceNumber;
+      },
+    );
+    if (!nextApplicationSequence) {
+      throw new BadRequestException();
+    }
+    const applicationNumber =
+      parseInt(sequenceName) * 10 ** 6 + (nextApplicationSequence - 1);
```
andrewsignori-aot, line 104: Please consider that the sequenceName could be a string. I do not see a benefit considering that it will be a number.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -86,13 +88,26 @@ export class ApplicationController extends BaseController {
...
+    submissionResult.data.applicationNumber = applicationNumber;
```
andrewsignori-aot, line 105: The submissionResult.data is supposed to be the result from the application dynamic form data.
Please pass the applicationNumber to the createApplication using a separated variable.

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: I believe that the logic could be simplified as below:
```
const MAX_APPLICATION_NUMBER_LENGTH = 10;
const sequenceNumberSize = MAX_APPLICATION_NUMBER_LENGTH - sequenceName.length;
const applicationNumber = sequenceName + nextApplicationSequence.padStart(sequenceNumberSize, "0") ;
```

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -8,6 +8,7 @@ export interface CreateApplicationDto {
    * with this application.
    */
   associatedFiles: string[];
+  applicationNumber: string
```
andrewsignori-aot, line 11: This DTO is being shared for API and service classes, but the applicationNumber doesn`t belong to the API layer.
I would recommend creating a new DTO for the service or passing the info to the service with parameters.

andrewsignori-aot (changes requested):

sources/packages/api/src/database/entities/application.model.ts
ann-aot: okay, So the draft application won't have the application number?

andrewsignori-aot (approved):
👍

abschwenker (approved):
LGTM

guru-aot (approved):
LGTM

dheepak-aot (approved):
Approved

Issue #564, opened 2021-07-19
Program Data in workflow - API Call to get Offering data for Camunda


abschwenker:
**Additional Context**
-When a student selects "offering or program not found" on the application. Camunda or the SIMS api should determine if a PIR is required. Camunda will need to retrieve the offering data from the SIMSAPI in order to assess the application

**Acceptance Criteria**
- [x] Create API to Save PIR Status
- [x] Create Migration to store PIR Status and Offering ID
- [x] Create API to Get Offering data using Offering ID from Camunda
- [x] Change workflow to use Offering Data from API Call

Newly added
- [x] Create Migration save the workflow ID to application table;
- [x] Create API endpoint to set the Offering ID (when PIR requested) and allow the workflow to advance

abschwenker:
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/a59673e5-536e-461c-93e5-7666f0314776)

abschwenker:
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/d684dbbc-6f22-4c99-ae62-ff1e9bb0ac88)

PR #588, opened 2021-07-27
[api][SIMS #564]Program Data in workflow - API Call to get Offering data for Camunda

andrewsignori-aot:
- Create API routes to allow Camunda to retrieve data and to update data back.
- Create assessment_workflow_id on sims.applications table to store the workflow id started when the application is submitted.
- Created location_id to store the location associated with the application. When an application is created without an offering we will need a way to know what location will be handling the PIR.
- Created fake entities to prepare scenarios for e2e tests on the database.
- Created the below e2e tests.
image


sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -93,11 +94,28 @@ export class ApplicationController extends BaseController {
...
+    // TODO: Once we have the application status we should run the create/associate
+    // under a DB transation to ensure that, if the workflow fails to start we would
+    // be rolling back the transaction and returning an error to the student.
+    const workflowAssociationResult =
+      await this.applicationService.associateAssessmentWorkflow(
+        createdApplication.id,
+        assessmentWorflow.id,
+      );
+
+    if (workflowAssociationResult.affected !== 1) {
```
ann-aot, line 111: can we add enum for these statuses, or add a comment, like to quickly understand, what does status 1 and 0 means

andrewsignori-aot: I believe that an enum does not make sense because 1 just represents the expected number of affected rows. I will add a comment.

andrewsignori-aot: Comment added.

sources/packages/api/src-sql/Applications/Add-ProgramInfo.sql
```diff
@@ -0,0 +1,31 @@
...
+
+COMMENT ON COLUMN sims.applications.offering_id
+    IS 'Offering associated with the application.';
+
+-- Create pir_status
+ALTER TABLE sims.applications
+    ADD COLUMN IF NOT EXISTS pir_status sims.program_info_status;
+
+COMMENT ON COLUMN sims.applications.pir_status
+    IS 'Status of the Program Information Request(PIR) (e.g. requested, in progress, not required).';
```
abschwenker, line 31: Status? 'required', 'not required', 'completed'

abschwenker (commented):
1 minor language comment tweak, non-impactful but a bit confusing for someone not as involved

sources/packages/api/src-sql/Applications/Add-ProgramInfo.sql
andrewsignori-aot: Comment updated to 'Status of the Program Information Request(PIR) (e.g. required, not required, completed).'

abschwenker (approved):
LGTM

ann-aot (approved):
👍

guru-aot (approved):
LGTM, and the extensive end to end test cases are good to understand.
Have some doubts on the connection parameters, will discuss with you to understand.

PR #596, opened 2021-07-28
Vue changes to import form

guru-aot:
As discussed, I have hard coded 112 as application id to fetch and test the form from formsflow.ai in dev. 
This change integrated the vue routes with the form in formsflow.ai

### DIFF
--- a/sources/packages/web/src/services/http/ApplicationApi.ts
+++ b/sources/packages/web/src/services/http/ApplicationApi.ts
@@ -26,4 +26,17 @@ export class ApplicationApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async getNOA(applicationId: any): Promise<any> {
+    try {
+      const response = await this.apiClient.get(
+        `application/${applicationId}/assessment`,
+        this.addAuthHeader(),
+      );
+      return response.data.data;
+    } catch (error) {
+      this.handleRequestError(error);
+      throw error;
+    }
+  }
 }
--- a/sources/packages/web/src/services/http/ApplicationApi.ts
+++ b/sources/packages/web/src/services/http/ApplicationApi.ts
@@ -30,7 +30,7 @@ export class ApplicationApi extends HttpBaseClient {
   /**
    * Retrieve the Notice of Assessment (NOA) for a particular application.
    */
-  public async getNOA(applicationId: any): Promise<any> {
+  public async getNOA(applicationId: number): Promise<any> {
     try {
       const response = await this.apiClient.get(
         `application/${applicationId}/assessment`,
--- a/sources/packages/web/src/services/http/ApplicationApi.ts
+++ b/sources/packages/web/src/services/http/ApplicationApi.ts
@@ -27,6 +27,9 @@ export class ApplicationApi extends HttpBaseClient {
     }
   }
 
+  /**
+   * Retrieve the Notice of Assessment (NOA) for a particular application.
+   */
   public async getNOA(applicationId: any): Promise<any> {
     try {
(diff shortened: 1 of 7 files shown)


sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -26,4 +26,17 @@ export class ApplicationApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async getNOA(applicationId: any): Promise<any> {
```
andrewsignori-aot, line 30: When using an acronym please consider adding some comments that just describe of the acronym means.
For instance:
```
/**
   * Retrieve the Notice of Assessment (NOA) for a particular application.
   */
```

sources/packages/web/src/views/student/financial-aid-application/Applications.vue
```diff
@@ -31,6 +31,21 @@
     <v-icon size="25" class="mr-2">mdi-text-box-plus</v-icon>
     Start New Application
   </v-btn>
+  <v-btn
```
andrewsignori-aot, line 34: Since @ann-aot changes to application summary will be done inside the sprint and we can demo this accessing directly the URL, I would consider not adding this.
If you think that is still worth you can keep it, but we need to remove it as soon it is integrated into the application summary list, up to you 😉 .

andrewsignori-aot (approved):
Nice 👍

sources/packages/web/src/views/student/NoticeOfAssessment.vue
```diff
@@ -0,0 +1,35 @@
...
+    applicationId: {
+      type: Number,
+      required: true,
+    },
+  },
+  setup(props: any) {
+    // Hooks
+    const initialData = ref({});
+    onMounted(async () => {
+      initialData.value = await ApiClient.Application.getNOA(
```
ann-aot, line 26: I think, you missed the service and you are directly calling the APIclient, actually, we are following the method like 'vue call a service , and the service will call the APIclient

sources/packages/web/src/services/http/ApplicationApi.ts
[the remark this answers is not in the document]
guru-aot: appplicationId should be a number. Promise<any> is used, as the route itself will be returning any, in the future we may add more objects and eventually no code changes are needed

sources/packages/web/src/views/student/NoticeOfAssessment.vue
guru-aot: But I had hard time finding a service for application, so I had to use StudentService. That's the reason I went via the API Client

andrewsignori-aot: I agree with @ann-aot, yes, please create a service and use it.

sources/packages/web/src/services/http/ApplicationApi.ts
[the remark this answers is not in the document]
andrewsignori-aot: applicationId really should be a number but the Promise<any> could stay, IMO. This data has no defined structure and should be flexible and interpreted only by the Form.IO definition.

sources/packages/web/src/views/student/NoticeOfAssessment.vue
ann-aot: ya @guru-aot , you can create a sep service for applications or of it student-specific, you can add in student service

ann-aot (approved):
👍

abschwenker (approved):

Issue #319, opened 2021-03-18
Student views application summary`


roopfw:
**As a** *(Student)* **I want** *(view my financial aid application)* **So That** *(I can view the status of my current application and find historical information on my past applications)*

**Additional Context**

**Acceptance Criteria**
- [x] Given (Student), When (Click "Applications" in the top nav bar), Then (Display a list of applications) -Simple List
- [x]  "applications summary page button to "Edit Application"

roopfw:
Waiting for walkthrough from Stuart Timewell. Was supposed to happen today but got pushed to Friday.

PR #597, opened 2021-07-29
[api][vue][sims-#319] Student views application summary

ann-aot:
As a (Student) I want (view my financial aid application) So That (I can view the status of my current application and find historical information on my past applications)

1. Given (Student), When (Click "Applications" in the top nav bar), Then (Display a list of applications) -Simple List
2. "applications summary page button to "Edit Application"
image

image

image


abschwenker (commented):
please review tests, other than that lgtm

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -15,4 +15,27 @@ export interface GetApplicationDataDto {
    * Application dynamic data.
    */
   data: any;
-}
\ No newline at end of file
+}
```
andrewsignori-aot, line 18: Please add a blank line to be formatted in the same way as others.

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -15,4 +15,27 @@ export interface GetApplicationDataDto {
...
+  id: number;
+  applicationName: string;
+  award: string;
+  status: string;
+}
+
+export interface ApplicationFormData {
+  selectedLocation: number;
+  selectedProgram?: number;
+  offeringIWillBeAttending?: number;
```
andrewsignori-aot, line 32: I think that we need to rename this variable all over the place.
We are using selectedLocation, selectedProgram and I believe that offeringIWillBeAttending should be selectedOffering.
Thoughts?

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -308,4 +313,48 @@ export class StudentController extends BaseController {
 
     stream.pipe(response);
   }
+
+  @Get("application-summary")
+  async getStudentApplicationSummary(
+    @UserToken() userToken: IUserToken,
+  ): Promise<StudentApplicationDTO[]> {
+    const existingStudent = await this.studentService.getStudentByUserName(
```
andrewsignori-aot, line 321: Now that we have the student id on token and we have a method in the student service to getStudentByUserId, I think that we should start to use the search by id (primary key) instead of the username, IMO.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -308,4 +313,48 @@ export class StudentController extends BaseController {
...
+      userToken.userName,
+    );
+    if (!existingStudent) {
```
andrewsignori-aot, line 324: I believe that we can add the student id to and "Student Token" and move the validation if the student is present to a decorator. We can create a tech debit for it. The idea would be to try to keep the controller clean.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -308,4 +313,48 @@ export class StudentController extends BaseController {
...
+      );
+    }
+    const application = await this.applicationService.getAllStudentApplications(
+      existingStudent.id,
+    );
+    let response = [] as StudentApplicationDTO[];
+    for await (const element of application) {
+      let offering = undefined;
+      if (
+        element.data?.selectedProgram &&
```
andrewsignori-aot, line 336: No need for a null check on data since it is "not null" on the database.

sources/packages/api/src/route-controllers/application/models/application.model.ts
ann-aot: yes, that will be good, if we are updating this variable, we need to update in form definition  and workflow also

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -308,4 +313,48 @@ export class StudentController extends BaseController {
...
+        element.data?.offeringIWillBeAttending
```
andrewsignori-aot, line 337: The offering that students will be attending should come from the offering_id on the application table. This is the field that will be resolved by the PIR or populated by the workflow if it is present in application data.
@abschwenker, is there any concern about it? The only drawback is that the offering_id could be null till it is populated by the workflow or resolved by the PIR, but I believe that offering_id should be the source of thuth.

sources/packages/api/src/route-controllers/student/student.controller.ts
ann-aot: i agree

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -308,4 +313,48 @@ export class StudentController extends BaseController {
...
+    if (!existingStudent) {
+      throw new NotFoundException(
+        `No student was found with the student name ${userToken.userName}`,
+      );
+    }
+    const application = await this.applicationService.getAllStudentApplications(
+      existingStudent.id,
+    );
+    let response = [] as StudentApplicationDTO[];
+    for await (const element of application) {
```
andrewsignori-aot, line 333: The performance impact of doing it in for loop awaiting for every single item is huge and it will be getting worse as the list gets bigger. As commented below, if we are going to use the offering_id from applications table, I believe that we can resolve all that is needed is a DB query.

andrewsignori-aot: Please change the element to a more meaningful name.

sources/packages/web/src/types/AppRoutes.ts
```diff
@@ -24,5 +24,7 @@ export enum AppRoutes {
   InstitutionManageUsers = "manage-users",
   DisabledUser = "login/disabled-user",
   StudentApplication = "application-form/:id?/:selectedForm",
+  StudentApplicationSummary = "my-application-summary",
```
andrewsignori-aot, line 27: It seems that these routes are not aligned: StudentApplication, StudentApplicationSummary, and StudentEditApplication.
StudentEditApplication is just a placeholder for now, right?

sources/packages/web/src/views/student/StudentApplicationSummary.vue
```diff
@@ -0,0 +1,86 @@
...
+          id: id,
+        },
+      });
+    };
+    onMounted(async () => {
+      myApplications.value = await StudentService.shared.getAllStudentApplications();
+    });
+    return { myApplications, goToApplication };
+  },
+  methods: {
```
andrewsignori-aot, line 79: Please move this method inside the Setup and add the explicit return type.

sources/packages/web/src/views/student/StudentEditApplication.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <v-btn color="primary" class="float-right" @click="toggle"
+    >Application Options
+    <v-icon size="25"> mdi-arrow-down-bold-circle</v-icon></v-btn
+  >
+  <Menu class="mt-n15" ref="menu" :model="items" :popup="true" />
```
andrewsignori-aot, line 6: Is there any other Vue component that we can use instead of this PrimeVue?

andrewsignori-aot (changes requested):
Please review the comments. The most important one is about the way the applications are retrieved in the API. We can discuss as needed and maybe we need some inputs from @abschwenker.

sources/packages/web/src/views/student/StudentEditApplication.vue
ann-aot: @andrewsignori-aot I checked this before using primevue, vuetify3 alfa, does not support v-menu as of now

sources/packages/api/src/route-controllers/application/models/application.model.ts
andrewsignori-aot: I am changing the workflow in my current task. I can update the workflow and the form definition, does it sounds good?
Can you please add the below change to your review also to complete the renaming effort?
image

sources/packages/api/src/route-controllers/student/student.controller.ts
ann-aot: we are not going to load the whole data at once, as it will cause performance issues, we will be using pagination with page calls?

sources/packages/web/src/types/AppRoutes.ts
ann-aot: as discussed closing this thread

sources/packages/api/src/route-controllers/application/models/application.model.ts
ann-aot: okay changing to `selectedOffering`

sources/packages/api/src/route-controllers/student/student.controller.ts
abschwenker: Null does seem logical until he PIR is complete because the Offering ID will contain the final source of truth

andrewsignori-aot (approved):
👍

abschwenker (approved):

guru-aot (approved):
LGTM

PR #600, opened 2021-07-29
Updated JSON and data field change in NOA

guru-aot:


andrewsignori-aot (changes requested):
I noticed that the json on databse is represented as below:
```
{
    "weeks": 52,
    "total_federal_award": 22893,
    "total_provincial_award": 2656.8911627906964,
    "federal_assessment_need": 4642.227906976743,
    "provincial_assessment_need": 4642.227906976743
}
```
Please rename the variable to be in accordance with the others that we have in SIMS DB and also in the form definitions (e.g. provincialAssessmentNeed).

Why the structure above is not being used as a form data directly?
Instead of having the custom code below
image

we could be accessing the form data directly (e.g. data.provincialAssessmentNeed).
I would recommend using custom code only when it is needed. I am not convinced that this would be the case.
IMO, we should try to keep the form definition as dry as possible.

guru-aot:
> provincialAssessmentNeed

sources/packages/forms/src/form-definitions/noticeofassessment.json
```diff
@@ -0,0 +1,5593 @@
...
+          "persistent": false,
+          "inputFormat": "html",
+          "protected": false,
+          "dbIndex": false,
+          "case": "",
+          "encrypted": false,
+          "redrawOn": "",
+          "clearOnHide": true,
+          "customDefaultValue": "",
+          "calculateValue": "let eligibleAmount=JSON.stringify(data.totalFederalAward+data.totalProvincialAward);\nvalue='<h1><B>You are eligible to receive $'+eligibleAmount+' for '+data.weeks+' weeks</B></h1>';",
```
andrewsignori-aot, line 80: Is there a reason to keep the JSON.stringify here?

andrewsignori-aot (approved):
👍

abschwenker (approved):

ann-aot (approved):

PR #615, opened 2021-08-03
[web][vue-sims#529]Institution User views Student Application Summary- watch added

ann-aot:
watch added to update the summary list

### DIFF
--- a/sources/packages/web/src/views/institution/LocationApplications.vue
+++ b/sources/packages/web/src/views/institution/LocationApplications.vue
@@ -68,7 +68,7 @@
 </template>
 
 <script lang="ts">
-import { onMounted, ref } from "vue";
+import { onMounted, ref, watch } from "vue";
 import { ApplicationService } from "../../services/ApplicationService";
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/views/institution/LocationApplications.vue
```diff
@@ -98,6 +98,15 @@ export default {
     const goToViewApplication = (applicationId: number) => {
       console.log(applicationId);
     };
+    watch(
+      () => props.locationId,
+      async currValue => {
```
andrewsignori-aot, line 103: Can you please create one method and call it on watch and onMounted?
The method is supposed to be called in the same way for both, right?

ann-aot: okay, I didn't create a common function because there was only one line of operation. anyways, I have updated the code.

abschwenker (approved):

guru-aot (approved):
LGTM

andrewsignori-aot (approved):
👍

Issue #589, opened 2021-07-27
Institution User can Add Program Intensity to Ed Program and Offering


michesmi:
**As a** *(Institution User)* **I want** *(to add information on Program Intensity)* **So That** *(I can confirm which offering a student will be attending)*

**Additional Context**
- New Fields have been added to the "Create New Program" Wireframes and flow to account for Program Intensity. This is indicated as NEW in Orange on the Figma.
- Offerings (sometimes called Intakes in the Figma) will remain Offering until user testing confirms otherwise (Likely October- we will determine then if it's a priority to change for MVP.
- Related Ticket- Students can Select Program Intensity #610

**Acceptance Criteria**
**Create New Program Page**:
- [x] New Question:  "Are Students able to take this on a part-time basis? A part-time program has a course load between 20 and 59%. A full-time program must have a course load of:  60% or greater OR  Between 40 and 60% for students with a permanent disability. _(Yes / No)_ 
 
**"Add Offerings" Page**
-  [x] New Question, "How will this be offered" (User can select " _Part-time" or "Full-time")

**Add Additional Offerings" Page** This page is not built. so don't worry about it. 

** "View Program" Page**
- [x] Summary Type: (Part-time/ Full-time) offerings
- [x] Institution Users can Offering List- Type Column- Full-time or Part-time listed

PR #619, opened 2021-08-06
[web][api][sims-#589]Institution User can Add Program Intensity to Ed Program and Offering

ann-aot:
Institution User can Add Program Intensity to Ed Program and Offering#589
1. new Intensity questions added to Program and Offering Form and update formio forms.json
2. new fields and migrations added in program and offerings
3. display the types in offerings view and offering summary page
4. updated edit offering and program initial data

image

image

image


sources/packages/api/src-sql/EducationPrograms/Create-part-time-basis-program.sql
```diff
@@ -0,0 +1,10 @@
+-- Add the column part_time_basis_program for Program with a dummy default value.
+ALTER TABLE sims.education_programs ADD 
+    COLUMN IF NOT EXISTS part_time_basis_program VARCHAR(50) NOT NULL DEFAULT 'no';
+COMMENT ON COLUMN sims.education_programs.part_time_basis_program IS 'If part_time_basis_program is yes, then the program is both Full-Time and Part-Time, if part_time_basis_program is no, then the program is only Full-Time basis';
```
guru-aot, line 4: Suggestion on column rename: Can we remove basis in part_time_basis_program? it would be better if, we have part_time_program as yes or no.

sources/packages/api/src-sql/EducationProgramsOfferings/Create-offered-type.sql
```diff
@@ -0,0 +1,9 @@
+-- Add the column offered_type for Offering with a dummy default value.
+ALTER TABLE sims.education_programs_offerings ADD 
+    COLUMN IF NOT EXISTS offered_type VARCHAR(50) NOT NULL DEFAULT 'fullTime';
+COMMENT ON COLUMN sims.education_programs_offerings.offered_type IS 'offered_type decides if offering is Full-Time and Part-Time';
```
guru-aot, line 4: If the values are going to be Full-Time or Part-Time, can we add it as a enum type and assign sims.offerings_type here?

ann-aot: mm, here values are coming from the formio and this default value is only for the migration purpose, it's a dummy value. I don't find the actual use of enum here. correct me if I misinterpreted it @guru-aot , let me know your thoughts @andrewsignori-aot

sources/packages/api/src-sql/EducationPrograms/Create-part-time-basis-program.sql
ann-aot: mm, do you find it confusing? it's pretty straight forward right?okay, let me know if you want me to update the field name @guru-aot

sources/packages/api/src-sql/EducationProgramsOfferings/Create-offered-type.sql
guru-aot: I get it, it comes from formio. That's the real reason, I want to validate it along the db side too. I feel using these values we will create logics for the application going further. @andrewsignori-aot Thoughts please

sources/packages/api/src-sql/EducationPrograms/Create-part-time-basis-program.sql
guru-aot: I felt its quite large, its fine, just a suggestion.

andrewsignori-aot: Just adding my thoughts also 😄...
I would suggest that the column should be called something like "intensity" and instead of yes/no we can save something like "partTime"/"fullTimePartTime" or whatever makes sense from database/api perspective.
The UI would still be the same with the Yes/No.

abschwenker: For the program I think it makes sense that if we have intensity with part time or parttimefulltime. Thoughts?

sources/packages/api/src-sql/EducationProgramsOfferings/Create-offered-type.sql
abschwenker: For offering I think an appropriate name would be intensity with values of Full Time Part Time. Thoughts?

andrewsignori-aot: > I get it, it comes from formio. That's the real reason, I want to validate it along the db side too. I feel using these values we will create logics for the application going further. @andrewsignori-aot Thoughts please

Yes, we will probably be using this to apply some logic on API and/or workflow. I started a conversation with @abschwenker about when creating or not a Postgres enum, for instance. Let's review it in the upcoming meeting at 2:30 PM.

abschwenker (approved):

sources/packages/api/src-sql/EducationProgramsOfferings/Create-valid-intensity.sql
```diff
@@ -0,0 +1,9 @@
+-- Add the column valid_intensity for Offering with a dummy default value.
+ALTER TABLE sims.education_programs_offerings ADD 
+    COLUMN IF NOT EXISTS valid_intensity sims.valid_intensity NOT NULL DEFAULT 'fullTime';
```
andrewsignori-aot, line 3: I am not able to understand why the name starts with "valid". If for program is "program_intensity" I would expect something like "offering_intensity" or "intensity". Can you please elaborate if there is a reason to call it a "valid_intensity"?

sources/packages/api/src-sql/Types/Create-ProgramIntensities.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TYPE sims.program_intensity
+  AS ENUM ('fullTimePartTime', 'fullTime');
```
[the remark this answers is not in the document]
andrewsignori-aot: > Please see naming convention, though this relates to a yes/no question I believe we should still maintain consistency. Thoughts?

Sorry @abschwenker I did not get it. From API/database perspective I believe that it should be a yes/no value unless we change the name of the column.
Besides that, the I thought that the idea would be save the type using friendly  name that could be directly displayed in the UI if need, as below.
```
CREATE TYPE sims.program_intensity
  AS ENUM ('Full Time and Part Time', 'Full Time');
```

sources/packages/api/src/route-controllers/education-program/education-program.controller.ts
```diff
@@ -109,6 +110,10 @@ export class EducationProgramController {
       eslEligibility: program.eslEligibility,
       hasJointInstitution: program.hasJointInstitution,
       hasJointDesignatedInstitution: program.hasJointDesignatedInstitution,
+      programIntensity:
+        program.programIntensity === ProgramIntensity.fullTimePartTime
+          ? "yes"
```
andrewsignori-aot, line 115: Please avoid doing it on API. It is a UI matter that could be resolved on Form.IO definition assigning the proper values to options on Form.IO.

sources/packages/api/src/route-controllers/education-program/models/save-education-program.dto.ts
```diff
@@ -19,6 +19,7 @@ export interface EducationProgramDto {
   eslEligibility: string;
   hasJointInstitution: string;
   hasJointDesignatedInstitution: string;
+  programIntensity: string;
```
andrewsignori-aot, line 22: Please consider using enum and use an IsEnum class validator.

andrewsignori-aot, line 22: Please use the enum whatever it is possible to make the type explicit.

sources/packages/api/src/services/education-program/education-program.service.models.ts
```diff
@@ -24,6 +25,7 @@ export interface SaveEducationProgram {
   hasJointInstitution: string;
   hasJointDesignatedInstitution: string;
   approvalStatus: string;
+  programIntensity: string;
```
andrewsignori-aot, line 28: Please use the enum and use the IsEnum class validator.

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -79,6 +80,10 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
       educationProgram.hasJointDesignatedInstitution;
     program.approvalStatus = educationProgram.approvalStatus;
     program.institution = { id: educationProgram.institutionId } as Institution;
+    if (educationProgram.programIntensity === "yes")
```
andrewsignori-aot, line 83: Please avoid resolving it on API. I would recommend changing the values on Form.IO definition.

sources/packages/forms/src/form-definitions/educationprogram.json
```diff
@@ -320,6 +320,95 @@
...
+      "hidden": false,
+      "hideLabel": false,
+      "autofocus": false,
+      "disabled": false,
+      "tableView": false,
+      "modalEdit": false,
+      "values": [
+        {
+          "label": "Yes",
+          "value": "yes",
```
andrewsignori-aot, line 341: I would recommend using the values from DB type.

sources/packages/forms/src/form-definitions/educationprogram.json
```diff
@@ -320,6 +320,95 @@
...
+          "shortcut": ""
+        },
+        {
+          "label": "No",
+          "value": "no",
```
andrewsignori-aot, line 346: I would recommend using the values from DB type.

sources/packages/web/src/views/institution/locations/programs/LocationProgramView.vue
```diff
@@ -78,6 +97,18 @@
             header="Study Dates"
             :sortable="true"
           ></Column>
+          <Column field="validIntensity" header="Type" :sortable="true"
+            ><template #body="slotProps">
+              <span
+                v-if="slotProps.data.validIntensity === ValidIntensity.partTime"
```
andrewsignori-aot, line 103: If the names of the types on DB are user-friendly please consider displaying them directly here.

guru-aot (approved):
LGTM

sources/packages/api/src-sql/Types/Create-ProgramIntensities.sql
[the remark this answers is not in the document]
abschwenker: Yes agreed, yes/no is just the label where everywhere else will use the ENUM value

sources/packages/api/src-sql/EducationProgramsOfferings/Create-valid-intensity.sql
abschwenker: offering_intensity is preferred

andrewsignori-aot (approved):
Great work 👍 
I noticed that you adjusted also the form definition to use "onlyAvailableItems": true 😉

abschwenker (approved):
LGTM

Issue #533, opened 2021-06-29
Institution User VIEWS Student Financial Aid Application Page for Program Information Request


michesmi:
**As a** *(Institution  User* **I want** *(View the information a Student provided on a Program and Offering )* **So That** *(I can determine If I should Complete a Program Information Request or Deny It)*

**Additional Context**
- Complete PIR and Deny Request are separate User Stories.

**Acceptance Criteria**
- [x] Given an (institution User), When (they are on the Student Application Summary page and select 'View' on an application that says 'Infor Request  ), Then (they are directed to the View 'Program Information Request Page' on behalf of that specific Student).
- [x] Given an (institution User), When (they are viewing Financial Aid Application on behalf of a specific student ), Then (they can VIEW the Following:
    - [x] An infobox at the top of the page, "This applicant indicated that the program or offering they would like to attend is not listed on StudentAid BC  You will need to confirm the program and offering information entered by the applicant."
    - [x] Show Basic Student Information Summary including Students Name,  Application Number, Location the Student is wanting to attend, Program Study Dates the student wishes to attend. 
    - [x]  Applicant Entered Details: Program Applicant will be attending: (_No Program Selected -means the student had to manually input information_). 
         - [x] Applicant Entered Program:  (_Provide text student manually input)_
          - [x] Applicant Entered Description:  (_Provide text student manually input)_
     - [x] Offering Applicant will be attending: (_No Offering Selected -means the student had to manually input information_). 
          - [x] Applicant Entered Offering:  (_Provide text student manually input)_
- [x] Location (_Provide the location the student selected)_
- [x] Application Type  (_Provide application type the student selected)_  (Not Possible Yet)
- [x] Study Period  (_Provide study period the student selected)_
- [x] Application Number  (_#530)_
- [x] Application Status (_On this page it will say, "Program Information Request" )_
- [x] Given an (institution User), When (they have viewed the information on the page  ), Then (they can determine if they wish to 'Complete Information Request' or 'Deny Request' under Application Actions at the top of the page).

Issue #103, opened 2021-01-26
Institution User CREATES and COMPLETES New Program Offering at Program Information Request


mzmichellesmith:
**As a** *(Institution User* **I want** *(create a new Program offering when a student selects a Program Information Request)* **So That** *(the student can be approved for Financial Aid)

**Additional Information**

** Acceptance Criteria** 
- [x] Given an (institution User), When (can not complete a Program information Request from previously approved Program Offerings ), Then (they will need to Create a New Offering).
      - [ ] 'Create New Offering" If the offering the student would like to attend is not currently entered in StudentAid BC you may enter a new one by clicking the “Create New Offering Button” You may also choose to option to make the offering available to future applicants."  _(Provide Create New Offering Button and provide the pop-up section, " Create New Offering')_ 
** New Offering** ( _Provide the Add Offerings Page)_ User Story #421
- [x] Option is provided to add this Offering for the specific student for one time only or for Selected Education Program for ALL Students.  _(check box)_
- [x] Offering Availability- Given the Institution User Selects "add offering for all students) this offering is now available for all students to selects.
- [x] Provide of Summary view of what the institution user created/ selected for the offering.
- [x] When the User completes the Program Info Request- the User is redirected back to the Student Application Summary Page and the Status is changed to _( Completed)_ 

**As a** *(Institution  User* **I want** *(Complete a Program Information Request)* **So That** *(so that a student can receive Financial aid for my program)*

**Additional Context**

**Acceptance Criteria**
- [x] Given an (institution User), When (they are on the 'View Financial Aid Application' Page and select 'Complete Program Information Page' ), Then (they are directed to the  'Program Information Request' page on behalf of that specific Student).
** Program and Offering**

** No Program or Offering Selected**
      - [ ] _Provide the following if the Student could Not select Program OR Offering_ - 
            - [ ] 'Select Program' (_If the Institution User can determine which Program the student is intending to take based on the information provided, then they can select a previously approved Education Program from the list._ _(Perhaps the student missed it)_
             - [ ] ' Select Offering' "Select an offering that the student will be attending below." (_If the Institution User can determine which Offering the student is intending to take based on the information provided, then they can select a previously approved Education Program Offering from the list._

**Program Selected but No Offering**
      - [ ] __Provide the following if the Student Selected a Program but not an Offering_
[...]

PR #622, opened 2021-08-06
[web][api][SIMS #533][SIMS #103]Complete PIR

andrewsignori-aot:
- Partial delivery for #533 and #103
- Added PIR Program to student applications table;
- Prepare the API to start saving the PIR Program (need to update the workflow to start saving it);
- Updated e2e tests to check the program id;
- Created the form definition to complete/view a PIR (form will be added in the next PR);
- Loading PIR to be completed;
- Able to complete a PIR with a missing program and missing offering;
- Able to complete a PIR creating a new offering (custom or application-specific);
- Moved PIR-related components under PIR routes on Vue application.

TODO:
- Adjustments on the form definition;
- Adjustment on assessment workflow;
- Add proper side menu for PIR;
- More and more testing.


sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -226,7 +226,26 @@ export class EducationProgramOfferingController {
...
+    return offerings.map((offering) => ({
+      id: offering.id,
+      description: offering.name,
+    }));
+  }
+
+  @AllowAuthorizedParty(AuthorizedParties.institution)
+  @HasLocationAccess("locationId")
+  @Get("location/:locationId/education-program/:programId/offerings-list")
+  async getProgramOfferingsForLocationForInstitution(
```
abschwenker, line 248: This is what is causing 1 of the 2 code smells. We are duplicating intentionally because this "HasLocationAccess". Can you please confirm?

andrewsignori-aot: The short answer is yes. We need a method to return a list of programs (the same for offerings) and this method is consumed by students and institutions.
Students will have access to all the programs/offerings while institutions are going to have access only to their respective programs and offerings.
So, we have two methods with different authorization. To avoid impact the returned payload size, we are also keeping the data returned so far as only a key/value. Besides that, I would imagine these methods returning different data for students and institutions, which, besides only authorization, would justify the two methods.
Always open to suggestions 😉

abschwenker: Thats exactly what I thought. At this point I don't see an alternative. You can disregard these 2 codesmells

ann-aot (approved):

abschwenker (approved):
LGTM. Approved, please ignore the 2 code smells

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -142,19 +143,19 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+    const programOffering = this.populateProgramOffering(
       locationId,
       programId,
       educationProgramOffering,
     );
     return this.repo.update(offeringId, programOffering);
   }
 
-  async populateProgramOffering(
+  populateProgramOffering(
```
guru-aot, line 154: Cool I agree, dont know why I added it in the first place 👍

sources/packages/web/src/assets/css/base.css
```diff
@@ -1,15 +1,14 @@
...
-}
-
-.bold-text {
-  font-style: normal;
-  font-weight: 700;
-  line-height: 36px;
-  letter-spacing: -0.2px;
-  color: #2965c5;
-}
+html * {
```
guru-aot, line 1: Why we changed this? is this part of your Story?

sources/packages/web/src/services/http/EducationProgramOfferingApi.ts
```diff
@@ -3,17 +3,25 @@ import { OfferingDTO } from "../../types/contracts/OfferingContact";
...
-  ): Promise<void> {
+  ): Promise<number> {
     try {
-      await this.apiClient.post(
+      const response = await this.apiClient.post(
         `institution/offering/location/${locationId}/education-program/${programId}`,
         createProgramOfferingDto,
         this.addAuthHeader(),
       );
+      return +response.data;
```
guru-aot, line 24: +response.data?

guru-aot (commented):
Very Big PR, 👍 Nice work @andrewsignori-aot . Its a lot of work and end to end implementations

sources/packages/web/src/assets/css/base.css
andrewsignori-aot: Not quite part of this, but it was not part of any other specific one when it was introduced 😄 
This is just an effort here and there to try to bring some styles from Figma.
In this PR I added the formio-shared.css as an idea to try to start to centralized the styles used along with the form definitions in only one place instead of configuring one by one, as shown below. (@ann-aot FYI)
image

sources/packages/web/src/services/http/EducationProgramOfferingApi.ts
andrewsignori-aot: That's is the equivalent to parseInt(something) or Number(something).

sources/packages/api/src-sql/EducationProgramsOfferings/Add-OfferingType.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    sims.education_programs_offerings
+ADD
+    COLUMN offering_type sims.offering_types NOT NULL DEFAULT 'Public';
```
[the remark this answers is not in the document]
andrewsignori-aot: Yes, I also had second thoughts about it. I think that would be better to remove the default 👍

guru-aot (approved):
LGTM

PR #625, opened 2021-08-10
Feature/#617 confirmation of enrolment API

guru-aot:
-  Migration for coe_status and assessment_status column and types
- Create a button for "Confirm Assessment" on NoA
- On Click update assessement_status to complete
- Update application_status to Enrollment
- On Click update coe_status to required
- API call methods in application controller to update status of application, assessment and coe is created


sources/packages/api/src/database/migrations/1628282476832-CoeStatusType.ts
```diff
@@ -0,0 +1,12 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../../utilities";
+
+export class CoeStatusType1628282476832 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(getSQLFileData("Create-CoeStatus.sql", "Types"));
```
ann-aot, line 6: I think, we agreed that .sql name will be `hyphen-case`

sources/packages/api/src/database/migrations/1628282476832-CoeStatusType.ts
```diff
@@ -0,0 +1,12 @@
...
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(getSQLFileData("Drop-CoeStatus.sql", "Types"));
```
ann-aot, line 10: I think, we agreed that .sql name will be `hyphen-case`

ann-aot, line 10: pls mind renaming the .sql name in `hyphen-case`

sources/packages/api/src/database/migrations/1628282484750-AssessmentStatusType.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../../utilities";
+
+export class AssessmentStatusType1628282484750 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Create-AssessmentStatus.sql", "Types"),
```
ann-aot, line 7: pls mind renaming ALL .sql name in `hyphen-case`

sources/packages/api/src/database/entities/assessment-status.type.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export enum AssessmentStatus {
+  /**
+   * An Assessment is required
+   */
+  required = "Required",
+  /**
+   * An Assessment is not required, not used in our current workflow but having it as an placeholder
+   */
+  notRequired = "Not Required",
```
andrewsignori-aot, line 12: IMO, if it is something that we would likely have in the future we can leave it here, but I am not sure if it is the case.
@abschwenker, is there any possible scenario where the status "Not Required" would be used?

sources/packages/api/src/database/migrations/1628282476832-CoeStatusType.ts
andrewsignori-aot: Please adjust all the .sql files as per @ann-aot comments.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -134,4 +136,24 @@ export class ApplicationController extends BaseController {
...
+  ): Promise<any> {
+    const updateResult = await this.applicationService.studentConfirmAssessment(
+      applicationId,
+    );
+    if (!updateResult) {
+      throw new UnprocessableEntityException(
+        `Confirmation of Assessment for the application id ${applicationId} failed.`,
+      );
+    }
+    return updateResult;
```
andrewsignori-aot, line 157: Is there a reason why we are returning the updateResult here?
If the method is executed with success it should return a proper HTTP status, if there is an error the proper HTTP status should also be returned.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -134,4 +136,24 @@ export class ApplicationController extends BaseController {
...
+
+  /**
+   * Confirm Assessment of a Student.
+   * @param applicationId application id to be updated.
+   */
+  @AllowAuthorizedParty(AuthorizedParties.student)
+  @Patch(":applicationId/confirm-assessment")
+  async studentConfirmAssessment(
+    @Param("applicationId") applicationId: number,
+  ): Promise<any> {
```
andrewsignori-aot, line 148: If there is a need to return something here, please create a DTO.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -134,4 +136,24 @@ export class ApplicationController extends BaseController {
 
     return assessment;
   }
+
+  /**
+   * Confirm Assessment of a Student.
+   * @param applicationId application id to be updated.
+   */
+  @AllowAuthorizedParty(AuthorizedParties.student)
+  @Patch(":applicationId/confirm-assessment")
```
andrewsignori-aot, line 145: The controller or the service should check at some point if the application being updated belongs to the students.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -164,6 +166,88 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    );
+  }
+
+  /**
+   * Updates overall Application status.
+   * @param applicationId application id to be updated.
+   * @param status status of the Application.
+   * @returns COE status update result.
+   */
+  async studentConfirmAssessment(applicationId: number): Promise<any> {
```
andrewsignori-aot, line 231: I got the idea to reuse the same methods to updates the statuses, but IMO, executing them in parallel will not ensure that the update on DB happens successfully for all of them. One of them could fail and leave the record in an undesired state.
I would recommend do a simple update altogether or execute them in a transaction (I would prefer the first one 😉).

sources/packages/web/src/services/ApplicationService.ts
```diff
@@ -14,4 +14,7 @@ export class ApplicationService {
   public async getPIRSummary(locationId: number): Promise<PIRSummaryDTO[]> {
     return ApiClient.Application.getPIRSummary(locationId);
   }
+  public async confirmationOfAssessment(applicationId: number): Promise<any> {
```
andrewsignori-aot, line 17: IMO this method should not return any data. The success or error could be expressed as HTTP codes.

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -56,4 +56,18 @@ export class ApplicationApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async confirmationOfAssessment(applicationId: number): Promise<any> {
```
andrewsignori-aot, line 60: IMO this method should not return any data. The success or error could be expressed as HTTP codes.

sources/packages/web/src/views/student/NoticeOfAssessment.vue
```diff
@@ -22,12 +30,18 @@ export default {
   setup(props: any) {
     // Hooks
     const initialData = ref({});
+    const confirmAssessment = async () => {
+      await ApplicationService.shared.confirmationOfAssessment(
```
andrewsignori-aot, line 34: Please consider adding some toast notification for success or error. It is probably not part of the ticket but it will help us troubleshooting.

andrewsignori-aot (changes requested):
Nice work with the controllers and statuses updates.
Please take a look at the comments especially the one about authorization.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -134,4 +136,24 @@ export class ApplicationController extends BaseController {
...
+  async studentConfirmAssessment(
+    @Param("applicationId") applicationId: number,
+  ): Promise<any> {
+    const updateResult = await this.applicationService.studentConfirmAssessment(
+      applicationId,
+    );
+    if (!updateResult) {
```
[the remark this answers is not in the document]
guru-aot: When the update fails in the service

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Just to clarify, the Promise.all will wait for all promises to be returned, but it will not ensure at any moment that all had success or that all fail or that one fails and two worked.

sources/packages/api/src/database/entities/assessment-status.type.ts
andrewsignori-aot: I believe that I misunderstood the conversation that we had in the past. My understanding was that the sub statuses would be the same because they would be following a similar flow. IMHO, they must represent the flow (or what it will probably be) and they can evolve differently. In the future, if we keep states that are not used it could cause confusion, but the comments added will minimize it.
Question, if we found that we need a new status for assessment, for instance, are we going to update all the others ones?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -195,6 +197,82 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+   * @param applicationId application id to be updated.
+   * @param status status of the Application.
+   * @returns COE status update result.
+   */
+  async studentConfirmAssessment(
+    applicationId: number,
+    student: Student,
+  ): Promise<UpdateResult> {
+    return this.repo.update(
+      { id: applicationId, student: student },
```
andrewsignori-aot, line 267: We don't need to use the entire Student object to select the application.
I would recommend receiving in the parameter only what is need, in this case, the student id.
The "where" would be like below.
```
return this.repo.update(
      { id: applicationId, student: { id: studentId }}
```

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -135,4 +137,35 @@ export class ApplicationController extends BaseController {
...
     return assessment;
   }
+
+  /**
+   * Confirm Assessment of a Student.
+   * @param applicationId application id to be updated.
+   */
+  @AllowAuthorizedParty(AuthorizedParties.student)
+  @Patch(":applicationId/confirm-assessment")
+  async studentConfirmAssessment(
```
andrewsignori-aot, line 147: I think about one more scenario, in order for the confirmation of the assessment to happen, should the record be in some specific state?
I believe that we need to introduce a check to verify if the application is in the expected state to have the assessment confirmed. Does it make sense @abschwenker?

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -42,4 +42,17 @@ export class ApplicationApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async confirmationOfAssessment(applicationId: number): Promise<void> {
```
andrewsignori-aot, line 46: Can we have this method renamed to confirmAssessment? It seems to be more aligned with the others and with the API call.

sources/packages/web/src/services/ApplicationService.ts
```diff
@@ -11,4 +11,7 @@ export class ApplicationService {
   public async getNOA(applicationId: number): Promise<any> {
     return ApiClient.Application.getNOA(applicationId);
   }
+  public async confirmationOfAssessment(applicationId: number): Promise<void> {
```
andrewsignori-aot, line 14: Can we have this method renamed to confirmAssessment? It seems to be more aligned with the others and with the API call.

sources/packages/web/src/views/student/NoticeOfAssessment.vue
```diff
@@ -21,13 +30,24 @@ export default {
   },
   setup(props: any) {
     // Hooks
+    const toast = useToastMessage();
     const initialData = ref({});
+    const confirmAssessment = async () => {
+      await ApplicationService.shared.confirmationOfAssessment(
+        props.applicationId,
+      );
+      toast.success(
```
andrewsignori-aot, line 39: Please add an error message as well.

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: I think application_status to be in 'Assessment'. Is that what you meant?

ann-aot (approved):

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: I am really not sure. I just believe that the student cannot change the status at any time and the API must restrict it.
What you mentioned makes sense to me, but I was looking to get some business knowledge from @abschwenker about it.

abschwenker: For a student to confirm their assessment the application must be in application_status="Assessment". Does that answer the question?

sources/packages/api/src/database/entities/assessment-status.type.ts
abschwenker: Depending what they evolve to they may need to be maintained separately. At this point I see a use case for each of the status's. For this PR and right now we can go under the assumption we will be using these 4 states for each status. However future changes may prove me wrong.

andrewsignori-aot (approved):
Thanks for taking care of all the comments, really appreciate it 👍

abschwenker (approved):

PR #634, opened 2021-08-12
Updated query to not null check

guru-aot:


andrewsignori-aot (approved):
Good catch! Thanks for taking care of the PIR also 👍

abschwenker (approved):

ann-aot (approved):

Issue #558, opened 2021-07-15
Dynamically validate form off submitted PY


abschwenker:
Include PY Drop down requirements
Include "active years" requirements
Currently submitted data, but need to use the PY + Form name data

PR #642, opened 2021-08-16
#558 program year dynamic validation - Save Program Year ID to the ApplicationFeature -Form Integration

guru-aot:
- Save Program Year ID to the Application
- Along side with Application Submission send the Program Year ID
- Validate if PY is Actiive
- Validates if the program year is active, if not throw a BadRequestException or similar.
- Use the form definition to validate the Student Application submission instead of the hard-coded value.

### DIFF
--- a/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
+++ b/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
@@ -2,8 +2,6 @@
 ALTER TABLE
     sims.applications
 ADD
-    COLUMN IF NOT EXISTS program_year_id INT REFERENCES sims.program_years(id) ON DELETE
-SET
-    NULL;
+    COLUMN IF NOT EXISTS program_year_id INT REFERENCES sims.program_years(id) ON DELETE CASCADE;
 
-COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active and the application is Submitted and not in Draft state';
+COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active';
--- a/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
+++ b/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
@@ -1,0 +1,9 @@
+-- Create program_year_id
+ALTER TABLE
+    sims.applications
+ADD
+    COLUMN IF NOT EXISTS program_year_id INT REFERENCES sims.program_years(id) ON DELETE
+SET
+    NULL;
+
+COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active and the application is Submitted and not in Draft state';
--- a/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
+++ b/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
@@ -2,6 +2,12 @@
 ALTER TABLE
     sims.applications
 ADD
-    COLUMN IF NOT EXISTS program_year_id INT REFERENCES sims.program_years(id) ON DELETE CASCADE;
+    COLUMN IF NOT EXISTS program_year_id INT NOT NULL REFERENCES sims.program_years(id) DEFAULT 2;
 
-COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active';
+COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active. The default 2 is the Program Year 2022-2023, this is used to ensure not null constraint';
+
+--Default constraint for the program_year_id is removed after the not null constraint is enforced
+ALTER TABLE
+    sims.applications
+ALTER
+    COLUMN program_year_id DROP DEFAULT;    
--- a/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
+++ b/sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
@@ -1,10 +1,10 @@
--- Create program_year_id
+-- Create program_year_id -  The default 2 is the Program Year 2022-2023, this is used to ensure not null constraint. 
 ALTER TABLE
     sims.applications
 ADD
     COLUMN IF NOT EXISTS program_year_id INT NOT NULL REFERENCES sims.program_years(id) DEFAULT 2;
 
-COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active. The default 2 is the Program Year 2022-2023, this is used to ensure not null constraint';
+COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active.';
 
 --Default constraint for the program_year_id is removed after the not null constraint is enforced
 ALTER TABLE
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -20,6 +20,7 @@ import { AssessmentStatus } from "./assessment-status.type";
 import { COEStatus } from "./coe-status.type";
 import { RecordDataModel } from "./record.model";
 import { Student } from "./student.model";
+import { ProgramYear } from "./program-year.model";
 
 @Entity({ name: TableNames.Applications })
 export class Application extends RecordDataModel {
@@ -75,7 +76,11 @@ export class Application extends RecordDataModel {
     referencedColumnName: ColumnNames.ID,
   })
   pirProgram?: EducationProgram;
-
+  @JoinColumn({
+    name: "program_year_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  programYearId?: ProgramYear;
   @ManyToOne(() => EducationProgramOffering, {
     eager: false,
     cascade: true,
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -89,7 +89,7 @@ export class Application extends RecordDataModel {
     name: "program_year_id",
     referencedColumnName: ColumnNames.ID,
   })
-  programYear?: ProgramYear;
+  programYear: ProgramYear;
 
   @ManyToOne(() => EducationProgramOffering, {
     eager: false,
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -76,11 +76,21 @@ export class Application extends RecordDataModel {
     referencedColumnName: ColumnNames.ID,
   })
   pirProgram?: EducationProgram;
+
+  /**
+   * References the program year related to the application.
+   * This will be populated only when an active program year application is Submitted
+   */
+  @ManyToOne(() => ProgramYear, {
+    eager: false,
+    cascade: true,
+  })
   @JoinColumn({
     name: "program_year_id",
     referencedColumnName: ColumnNames.ID,
   })
-  programYearId?: ProgramYear;
+  programYear?: ProgramYear;
+
   @ManyToOne(() => EducationProgramOffering, {
     eager: false,
     cascade: true,
--- a/sources/packages/api/src/route-controllers/application/application.controller.ts
+++ b/sources/packages/api/src/route-controllers/application/application.controller.ts
@@ -68,8 +68,8 @@ export class ApplicationController extends BaseController {
     @Body() payload: CreateApplicationDto,
     @UserToken() userToken: IUserToken,
   ): Promise<number> {
-    const programYear = await this.programYearService.validateProgramYear(
-      payload.selectedForm,
+    const programYear = await this.programYearService.getProgramYear(
+      payload.programYearId,
     );
 
     if (!programYear) {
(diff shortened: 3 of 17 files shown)


sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
```diff
@@ -0,0 +1,9 @@
+-- Create program_year_id
+ALTER TABLE
+    sims.applications
+ADD
+    COLUMN IF NOT EXISTS program_year_id INT REFERENCES sims.program_years(id) ON DELETE
```
andrewsignori-aot, line 5: I believe that this column should not allow NULL values. Is there a scenario where it could be null?

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
```diff
@@ -0,0 +1,9 @@
...
+SET
+    NULL;
+
+COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active and the application is Submitted and not in Draft state';
```
andrewsignori-aot, line 9: Draft applications need a program year associated with. Since the draft application is also coming from a Form.IO definition, that came from a program year, we should be saving the program year also during the draft application.

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
guru-aot: Yeah when the column is in Draft state.

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
guru-aot: I dont think, we need to calculate the program year in draft state, we can discuss in a call to confirm this. As per discussion with @abschwenker , we came up with a requirement to save it while its submitted. @abschwenker Please confirm

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -76,6 +77,20 @@ export class Application extends RecordDataModel {
...
+   */
+  @ManyToOne(() => ProgramYear, {
+    eager: false,
+    cascade: true,
+  })
+  @JoinColumn({
+    name: "program_year_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  programYear?: ProgramYear;
```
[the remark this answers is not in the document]
guru-aot: Lets discuss this until we come to a confirmation. Lets discuss

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -66,8 +68,17 @@ export class ApplicationController extends BaseController {
...
+    );
+
+    if (!programYear) {
+      throw new BadRequestException(
+        "Program Year is not active. Not able to create an application invalid request",
+      );
+    }
     const submissionResult = await this.formService.dryRunSubmission(
-      "SFAA2022-23",
+      payload.selectedForm,
```
andrewsignori-aot, line 81: Define the program year based on the form name is only valid if the column "programYear.formName" is unique in the database. IMO, or the column should be unique or we should send the program year id in the payload instead of the form name.

sources/packages/api/src/services/program-year/program-year.service.ts
```diff
@@ -16,4 +16,20 @@ export class ProgramYearService extends RecordDataModelService<ProgramYear> {
...
+
+  /**
+   * Validate Program Year of an Application,
+   * if the ProgramYear is not active
+   *  return null
+   * else r
+   *  return the ProgramYear row
+   * @param selectedForm Selected Form of the application
+   */
+  async validateProgramYear(selectedForm: string): Promise<ProgramYear> {
```
andrewsignori-aot, line 28: IMO this method is not validating the program year, it is retrieving the data. 
On the controller, it is being used to execute some validation, but this method in the service is not executing a validation, right? Could you please consider rename this method?

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: We are sending the form name, as its not changed inside the form, please verify sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue. We are dynamically loading the form from the db data, using the drop down. Its better to have the selected form(as it is not tampered in anyway), rather than having the program Year.

sources/packages/api/src/services/program-year/program-year.service.ts
guru-aot: Is renaming it to selectProgramYear is fine?

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
andrewsignori-aot: Why it will be null in the draft state?

guru-aot: Its not passed in Draft state, so it will be null. That is the part of the requirement.

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: That was not what we talked about and agreed on the planning/grooming. I am aware that we have the form name, but we talked about it and we decided to have the program id. That's why the AC was defined as below.
What does it means "as it is not tampered in anyway"?

image

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
andrewsignori-aot: Where is it in the requirement?

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: We are saving the program year id, but we are using the selected form to retrieve it. Yeah Submission and not in draft state.

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
guru-aot: "Along side with Application Submission send the Program Year ID"
 
 During submission , program Year id needs to be sent and saved. Not during draft state

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
abschwenker: @guru-aot  not sure if you had a look at how the draft applications now work, but a "save draft" is now a create/update in our db so there is some core information that will be be required. A draft submission should have the PY saved with it because once created it must match without the possibility to be matched to another PY

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: The sentence mentions the application submission in the sense of the HTTP request being submitted to the API, not the application being saved as submitted. Besides that, we talked about it, and we talked in different meetings. If there is a team decision to keep it as is, not a problem.

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
guru-aot: Yeah, when we submit the draft, the Program Year validation will happen, but the question is do we need to have the PY saved while the application is saved in draft state. Because last time, when we had a discussion, we agreed upon saving the PY only on successful submission and validation.

abschwenker: yes we should. There will also be draft cleanup tasks we have to do and it would be really easy if the PY was saved with draft to move all drafts belonging to a PY

sources/packages/api/src/services/program-year/program-year.service.ts
andrewsignori-aot: It seems that the methods retrieving data are using more getSomething, like:
getProgramYearByFormName
getProgramYear
,,,

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: I dont think, its the HTTP request submission. Then we would have clearly mentioned it as draft and submitted state submissions. Let's discuss

sources/packages/api/src/database/entities/application.model.ts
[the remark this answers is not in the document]
ann-aot: I would also like to join the discussion

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
guru-aot: Okay then if we agree on having the PY in Draft state saving, I can change my column to not null. For now, I have done my changes for the Submitted state. For Draft state integration, it has to be taken care. Fair enough? @abschwenker @andrewsignori-aot ?

sources/packages/api/src/services/program-year/program-year.service.ts
guru-aot: The main reason why I named the method like validateProgramYear is, it actually validates if the programYear is active or not by returning the complete column. But its fine, if you want me to rename

sources/packages/api/src-sql/Applications/Add-col-program-year-id.sql
```diff
@@ -0,0 +1,13 @@
+-- Create program_year_id
+ALTER TABLE
+    sims.applications
+ADD
+    COLUMN IF NOT EXISTS program_year_id INT NOT NULL REFERENCES sims.program_years(id) DEFAULT 2;
+
+COMMENT ON COLUMN sims.applications.program_year_id IS 'References the program year related to the application, this will be saved only when the ProgramYear is active. The default 2 is the Program Year 2022-2023, this is used to ensure not null constraint';
```
andrewsignori-aot, line 7: The comment about the Id 2 used should not be present in the column comments because it will no longer exists after the "DROP DEFAULT" right below.
Please, add a comment to the migration script, not to the column comment.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -66,8 +68,17 @@ export class ApplicationController extends BaseController {
     @Body() payload: CreateApplicationDto,
     @UserToken() userToken: IUserToken,
   ): Promise<number> {
+    const programYear = await this.programYearService.getProgramYear(
+      payload.programYearId,
+    );
+
+    if (!programYear) {
+      throw new BadRequestException(
```
andrewsignori-aot, line 76: IMO option would be more a 422 Unprocessable Entity exception.
I would say that we should use Bad Request when the request is not well-formed, like the ones that nest throws when "class-validators" fails. In this case, the request was understood, but the data could not be processed.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -66,8 +68,17 @@ export class ApplicationController extends BaseController {
     @Body() payload: CreateApplicationDto,
     @UserToken() userToken: IUserToken,
   ): Promise<number> {
+    const programYear = await this.programYearService.getProgramYear(
```
andrewsignori-aot, line 71: Thanks for renaming the method, but could we please use getActiveProgramYear instead?

sources/packages/api/src/services/program-year/program-year.service.ts
```diff
@@ -16,4 +16,20 @@ export class ProgramYearService extends RecordDataModelService<ProgramYear> {
       .orderBy("programYear.id", "DESC")
       .getMany();
   }
+
+  /**
+   * Validate Program Year of an Application,
```
andrewsignori-aot, line 21: Please adjust the comments since the method is not executing a validation.

sources/packages/web/src/views/student/financial-aid-application/Applications.vue
```diff
@@ -20,7 +20,8 @@
         $router.push({
           name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM,
           params: {
-            selectedForm: formName,
+            selectedForm: formName.formName,
```
andrewsignori-aot, line 23: Please consider renaming the variable name to something like "programYear", since it no longer represents the formName.

sources/packages/web/src/views/student/financial-aid-application/Applications.vue
```diff
@@ -50,7 +51,7 @@ export default {
       programYearList.value = programYears.map((programYear: ProgramYear) => ({
         name:
           "(" + programYear.programYear + ") - " + programYear.programYearDesc,
-        code: programYear.formName,
+        code: programYear,
```
andrewsignori-aot, line 54: Please consider renaming the variable "code" to something like "programYear" since it no longer represents the code.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -16,7 +16,7 @@
 
 <script lang="ts">
 import formio from "../../../components/generic/formio.vue";
-import { onMounted, ref } from "vue";
+import { onMounted, ref, computed } from "vue";
```
andrewsignori-aot, line 19: Please check if the "computed" is being used here, if not, please remove it.

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: I actually did as per the requirement, but fair enough

 Validates if the program year is active, if not throw a BadRequestException or similar.

andrewsignori-aot (approved):
Looks good to me 👍 
Thanks for taking care of the comments 😉

dheepak-aot (approved):

abschwenker (approved):

ann-aot (approved):
👍

Issue #605, opened 2021-07-30
Student can Cancel their FT/PT Application


michesmi:
**As a** *(Student)* **I want** *(Cancel my FT/PT Application)* **So That** *(I am not issued a loan that I cant commit to paying)*

**Additional Context**
-  Students wont be able to edit or re-submit a Cancelled Application in the future but they can View it.

**Acceptance Criteria**
- [x] Given (a Student), When (selects 'Cancel" when viewing their Fulltime/ Parttime application details ), Then (provide the pop-up, "Cancel Application", " Are you sure you want to cancel your application? You’ll still be able to view canceled applications but can’t make further edits".)
- [x] AND provide the options to 'Cancel' and 'Cancel Application' (in red))
    - [x] If the Student Selects 'Cancel' then navigate back to the Application Details page
    - [x] If the Student Selects 'Cancel Application' then, change the status of the application to 'Cancelled'
         - [x] AND provide the pop-up, "For your information" "This application was canceled on _(INSERT DATE_).  View application (with the option to view the canceled application) 
         - [x] AND provide the pop-up in green, "Your application is now canceled" 
         - [x] AND The application now appears as 'Cancelled' on the 'Student Application Summary' Page

  TASKS

- [x] API to update SIMS DB to canceled
- [x] API to call Camunda to Stop process_id
- [x] Display canceled 'state' on the application details page

PR #643, opened 2021-08-16
[vue][api][sims-#605] Student can Cancel their FT/PT Application

ann-aot:
1. API to update SIMS DB to canceled
2. API to call Camunda to Stop process_id
3.  Display canceled 'state' on the application details page

image

image

image

image

image


sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -140,4 +140,10 @@ export class Application extends RecordDataModel {
     type: "enum",
   })
   coeStatus: COEStatus;
+
+  @Column({
+    name: "application_status_updated_on",
+    nullable: true,
```
andrewsignori-aot, line 146: Since the column is not null on DB, please remove it.

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -140,4 +140,10 @@ export class Application extends RecordDataModel {
...
+  })
+  applicationStatusUpdatedOn?: Date;
```
andrewsignori-aot, line 148: Please remove the nullable option also.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -180,4 +184,60 @@ export class ApplicationController extends BaseController {
...
+        "The user is not associated with a student.",
+      );
+    }
+    const studentApplication =
+      await this.applicationService.getStudentApplicationStatus(
+        applicationId,
+        student,
+      );
+
+    if (!studentApplication) {
```
andrewsignori-aot, line 214: I would recommend doing the student/application validation as below, in the same way already in place on method _getByApplicationId_.
```
 const application =
      await this.applicationService.getApplicationByIdAndUserName(
        applicationId,
        userToken.userName,
      );
    if (!application) {
      throw new NotFoundException(
        `Application id ${applicationId} was not found.`,
      );
    }
```

sources/packages/api/src/database/entities/application.model.ts
ann-aot: Ya, nice catch

sources/packages/api/src/database/entities/application.model.ts
ann-aot: sure, Ya, nice catch

abschwenker (approved):
LGTM

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -402,4 +408,19 @@ export class ApplicationService extends RecordDataModelService<Application> {
       .addOrderBy("application.applicationNumber")
       .getMany();
   }
+  /**
+   * get Student Application.
+   * @param applicationId application id.
+   * @param student student id.
+   * @returns student Application object.
+   */
+  async getStudentApplicationStatus(
```
andrewsignori-aot, line 417: We already have a method that retrieves an application by user and id (getApplicationByIdAndUserName), is it possible to use it instead or maybe make it more generic and just rename it to getStudentApplication.

sources/packages/api/src/services/workflow/workflow.service.ts
```diff
@@ -44,4 +44,21 @@ export class WorkflowService {
...
+   */
+  async delete(assessmentWorkflowId: string): Promise<number> {
+    const deleteURL = `${this.config.ruleEngineUrl}/process-instance/${assessmentWorkflowId}`;
+    const token = await this.tokensService.simsApiClient.getToken();
+    const response = await axios.delete(deleteURL, {
+      headers: {
+        Authorization: `Bearer ${token}`,
+      },
+    });
+    return response.status;
```
andrewsignori-aot, line 62: I would recommend not returning a number. If I am not wrong, if the status is different from 2xx Axios is going to throw an exception, which means that this method will return always a 2xx.
Can you please consider changing it to return a Promise<void>?

sources/packages/api/src/services/workflow/workflow-actions.service.ts
```diff
@@ -59,6 +59,23 @@ export class WorkflowActionsService {
...
   }
 
+  /**
+   * delete application assessment.
+   * @param assessmentWorkflowId workflow Id to be deleted.
+   * @returns status code.
+   */
+  async deleteApplicationAssessment(
+    assessmentWorkflowId: string,
+  ): Promise<number> {
```
andrewsignori-aot, line 69: I would recommend returning a Promise<void> in this case. If I am not wrong, if the status is different from 2xx Axios is going to throw an exception, which means that this method will return always a 2xx.

sources/packages/api/src/utilities/common.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as utc from "dayjs/plugin/utc"; // import plugin
+import * as localizedFormat from "dayjs/plugin/localizedFormat"; // import plugin
+dayjs.extend(utc);
+dayjs.extend(localizedFormat);
+/**
+ * get utc date time now
+ * @returns date now in utc
+ */
+export const getUTCNow = () => {
+  return new Date(dayjs().utc().format());
```
andrewsignori-aot, line 14: I noticed that dayjs has also a method like below. Is it the same as "new Date(dayjs().utc().format())"?
```
dayjs().utc().toDate()
```

sources/packages/web/src/constants/routes/RouteConstants.ts
```diff
@@ -16,7 +16,7 @@ export const StudentRoutesConst = {
   NOTIFICATIONS: Symbol(),
   NOTIFICATIONS_SETTINGS: Symbol(),
   STUDENT_APPLICATION_SUMMARY: Symbol(),
-  STUDENT_EDIT_APPLICATION: Symbol(),
+  STUDEN_APPLICATION_DETAILS: Symbol(),
```
andrewsignori-aot, line 19: Is it supposed to be "STUDENT_APPLICATION_DETAILS"?

sources/packages/web/src/main.ts
```diff
@@ -5,8 +5,8 @@ import "primevue/resources/themes/md-light-indigo/theme.css";
 import "primevue/resources/primevue.min.css";
 import "primeicons/primeicons.css";
 import "primeflex/primeflex.css";
-
 import { createApp } from "vue";
+
```
andrewsignori-aot, line 9: If this is the only change in the file, please consider removing it from the final commit.

sources/packages/api/src/services/workflow/workflow-actions.service.ts
ann-aot: yes, it will throw exception,

sources/packages/api/src/utilities/common.ts
ann-aot: ya @andrewsignori-aot, both are the same, toDate() will be date object fed to new Date(), .format() is dates string fed to new Date(). new Date() accepts both date string and date object. both are correct

sources/packages/web/src/constants/routes/RouteConstants.ts
ann-aot: yes @andrewsignori-aot , it is the application details page, Application edit is a different page. That's what i understood from Figma. correct me if I am wrong @abschwenker

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -180,4 +184,60 @@ export class ApplicationController extends BaseController {
...
+  @Patch(":applicationId/update-status")
+  async updateStudentApplicationStatus(
+    @UserToken() userToken: IUserToken,
+    @Param("applicationId") applicationId: number,
+    @Body() payload: ApplicationStatusToBeUpdatedDto,
+  ): Promise<void> {
+    const student = await this.studentService.getStudentByUserId(
+      userToken.userId,
+    );
+    if (!student) {
```
andrewsignori-aot, line 203: I understanding that this check is more specific, but IMHO the check already in place on "getApplicationByIdAndUserName" would be enough to ensure the authorization in this case, since we ensure that the application belongs to the user accessing it.

andrewsignori-aot (approved):
Thanks for taking care of the comments, looks good 👍

sources/packages/web/src/constants/routes/RouteConstants.ts
abschwenker: I think you missed the T on Student :)

ann-aot: ya, sorry . pls ignore the comment :)

sources/packages/api/src/route-controllers/application/application.controller.ts
ann-aot: ya, for `getApplicationByIdAndUserName` the check is not need.  for the previous method, I was using the `student`. update it.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -180,4 +184,60 @@ export class ApplicationController extends BaseController {
...
     }
   }
+
+  /**
+   * Generic method to update all Student Application status from frontend.
+   * @param applicationId application id to be updated.
+   * @body payload contains the status, that need to be updated
+   */
+  @AllowAuthorizedParty(AuthorizedParties.student)
+  @Patch(":applicationId/update-status")
```
dheepak-aot, line 194: IMO,  @Patch("/status/:applicationId") is preferred REST endpoint

guru-aot: I guess, we have similar controller endpoints, we have a preference of passing the applicationId before the type of process the controller does.

Some Examples:
============
@Patch(":applicationId/submit")
@Get(":applicationId/assessment")
@Patch(":applicationId/confirm-assessment")

ann-aot: IMO, :applicationId should come after `application/`. @andrewsignori-aot  @guru-aot  thoughts?

guru-aot: @dheepak-aot the main controller has the /application so all the other endpoints will be

@patch(":applicationId/submit") - /application/:applicationId/submit
@get(":applicationId/assessment") - /application/:applicationId/assessment
@patch(":applicationId/confirm-assessment") - /application/:applicationId/confirm-assessment

I guess, you might have missed the appended controller endpoint.

Now I feel, it make sense after appending the /application. Am I right? Also its the preferred way of REST definition.

dheepak-aot: @ann-aot @guru-aot Agreed that :applicationId  should come after application. My understand was wrong about this controller initially.
But still verb(s) shouldn't be part of URI. like update, or create. IMO @Patch(":applicationId/status") will be better.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -180,4 +184,52 @@ export class ApplicationController extends BaseController {
...
+
+    if (!studentApplication) {
+      throw new UnprocessableEntityException(
+        `Application ${applicationId} associated with requested student does not exist.`,
+      );
+    }
+    // delete workflow if the payload status is cancelled
+    // workflow doest not exists for draft application
+    if (
+      payload.applicationStatus === ApplicationStatus.cancelled &&
```
guru-aot, line 214: Just a suggestion, always try to have the constant first, in a IF check (constant === variable). 

if(payload && ApplicationStatus.cancelled === payload.applicationStatus && ApplicationStatus.draft !== studentApplication.applicationStatus && studentApplication.assessmentWorkflowId) {

}

sources/packages/web/src/types/contracts/students/ApplicationContract.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Possible status of a Student Application.
+ */
+export enum ApplicationStatus {
+  /**
+   * The application is in draft state and is not Submitted
+   */
+  draft = "Draft",
+  /**
+   * The application has been submitted by the application
```
guru-aot, line 10: Please change it to "The application has been submitted by the student"

sources/packages/web/src/views/student/StudentApplicationDetails.vue
```diff
@@ -0,0 +1,169 @@
...
+      :applicationId="id"
+      @showHideCancelApplication="showHideCancelApplication"
+      @reloadData="getApplicationDetails"
+    />
+
+    <v-container class="pt-12">
+      <div
+        class="bg-white application-info-border"
+        v-if="
+          applicationDetails.applicationStatus == ApplicationStatus.cancelled
```
guru-aot, line 40: == or === ?

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: I agree, patch(":applicationId/status"). I think you are right @dheepak-aot

guru-aot (commented):
Minor changes requested, overall nice work @ann-aot . Thanks for taking care of the UTC conversions and saving.

sources/packages/web/src/types/contracts/students/ApplicationContract.ts
ann-aot: I literally copy-pasted your interface in API :). sure, will do

sources/packages/web/src/views/student/StudentApplicationDetails.vue
ann-aot: mm, ya we need to check the type too. updating

sources/packages/api/src/route-controllers/application/application.controller.ts
ann-aot: mm, Yoda notation, it's just a style of programming to avoid errors, ya sure, we can follow that

ann-aot: okay, I thought that since keys in the interface are mandatory, it will fail if a null payload comes. okay, it's not true, will add the condition for that

sources/packages/api/src/route-controllers/application/application.controller.ts
ann-aot: updated @dheepak-aot

dheepak-aot: Thanks @ann-aot. Great work overall

dheepak-aot (approved):
LGTM.

guru-aot (approved):
LGTM

PR #650, opened 2021-08-18
[vue][api][sims-#605] Student can Cancel their FT/PT Application and add edit application

ann-aot:
1. corrected cancel action 
2. added edit application in summary and details page


andrewsignori-aot (approved):
👍

PR #656, opened 2021-08-19
Updating getApplicationAPI for camunda

guru-aot:
Updated getApplication API for camunda to get Program Year from the application

### DIFF
--- a/sources/packages/api/src/route-controllers/application/models/application.model.ts
+++ b/sources/packages/api/src/route-controllers/application/models/application.model.ts
@@ -38,7 +38,7 @@ export interface ApplicationDataDto {
   /**
    * Program Year is added to integrate the application in camunda workflows
    */
-  programYear: String;
+  programYear: string;
 }
 
 export interface StudentApplicationDTO {
--- a/sources/packages/api/src/route-controllers/application/models/application.model.ts
+++ b/sources/packages/api/src/route-controllers/application/models/application.model.ts
@@ -35,6 +35,10 @@ export interface ApplicationDataDto {
    * Application dynamic data.
    */
   data: any;
+  /**
+   * Program Year is added to integrate the application in camunda workflows
+   */
+  programYear: String;
 }
 
 export interface StudentApplicationDTO {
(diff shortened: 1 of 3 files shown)


ann-aot (approved):
looks good 👍 . pls look code smells

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -35,6 +35,10 @@ export interface ApplicationDataDto {
    * Application dynamic data.
    */
   data: any;
+  /**
+   * Program Year is added to integrate the application in camunda workflows
+   */
+  programYear: String;
```
ann-aot, line 41: 'string` small 's'

andrewsignori-aot (approved):
Looks good to me 👍

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -289,6 +289,18 @@ export class ApplicationService extends RecordDataModelService<Application> {
     return application;
   }
 
+  async getApplicationById(applicationId: number): Promise<Application> {
```
dheepak-aot, line 292: Just a suggestion. We get application by id and program year but the method name describes only id. Service method name could be changed.

andrewsignori-aot: IMO this method could keep the generic name as it is. If we need to get more information we can add more items to the select. I would not recommend creating a specific getApplicationX to every single select that we need unless there is a huge performance impact for a particular feature.

guru-aot: Agreed @andrewsignori-aot . Actually we are not getting the application with respect to the program year. The where clause contains if the program year is active. We can have a discussion if you want.

dheepak-aot: Not required @guru-aot @andrewsignori-aot. I am good.

dheepak-aot (approved):
LGTM

abschwenker (approved):
LGTM

Issue #532, opened 2021-06-29
Institution User Edits Student Enrolment  (CoE)


michesmi:
As an (Institution User) I want (to edit Program or Offering Information ) So That (I can confirm enrolment for a student to receive Financial Aid from SABC.

Additional Context
- Editing program information in CoE will cause a PIR to be required leading to a new Assessment
- An Institution User may also deny Enrolment when a student is not enrolled in their program #606
- Confirm Enrolment #166

Acceptance Criteria
- [x] On confirmation of enrolment. When clicking "Edit Program Information" show model of confirmation.
----
- [x] Duplicate row and file associations
- [x] On Duplicate, copy Application number and provide date in "application_status_updated_on"
- [x] add application_status of "Overwritten" (add a good comment)
- [x] populate pir_status "Required" and offeringid (status's) from the "Overwritten" application to the "Submitted" Application
- [x] trigger workflow again (same way as application submit)
- [x] Application status should be "Submitted"
- [x] Restrict API to not be able to return "Overwritten" (Application Summary Page, and check other constraints for APIs)
- [x] Delete Camunda Process Instance ID (Created already) and Start new Process (/start)

3 End states for an application
- Completed
- Cancelled
- Overwritten

PR #657, opened 2021-08-20
[api][SIMS #532]Institution User Edits Student Enrolment (CoE)

andrewsignori-aot:
- Partial delivery for ticket #532 ;
- No Vue changes included;
- Created the new enum value for application status called "Overwritten";
- Added @RelationId properties to application entity to allow the retrieval of many foreign keys without the need to create a join to satisfy Typeorm requirements. In this way, we can have easy access to foreign keys;
- Create a centralized method to start the application workflow and assign the workflow id to the application. This method is being shared between the submit application method (already existent) and the new one added to COE;
- Create a new API endpoint to:
    - Clone the Student Application;
    - Delete the current workflow;
    - Start a new workflow.
    
This API endpoint was created as a **POST** because it is creating a new resource and because **the operation is NOT idempotent** (open for suggestions 😉)
The endpoint also has the "rollback" appended because the COE is not being edited, from an API perspective, it is canceling the process and rolling back the assessment to a previous phase. It was also tricky to find a good way to name it, please let me know if someone has a better idea.

### DIFF
--- a/sources/packages/api/src/database/entities/application-status.type.ts
+++ b/sources/packages/api/src/database/entities/application-status.type.ts
@@ -38,6 +38,9 @@ export enum ApplicationStatus {
    * an edit on Confirmation of Enrollment that forces the assessment to
    * be reevaluated. Is this case the application is cloned and and the
    * old version is marked as 'Overwritten'.
+   * An Overwritten application should never be modified, once an application
+   * is Overwritten and a clone/new version is created all edits should take
+   * place on new record.
    */
   overwritten = "Overwritten",
 }
--- a/sources/packages/api/src/database/entities/application-status.type.ts
+++ b/sources/packages/api/src/database/entities/application-status.type.ts
@@ -30,7 +30,14 @@ export enum ApplicationStatus {
    */
   completed = "Completed",
   /**
-   * TThe application has been cancelled by the student
+   * The application has been cancelled by the student
    */
   cancelled = "Cancelled",
+  /**
+   * The application was replaced by a new version due to some event like
+   * an edit on Confirmation of Enrollment that forces the assessment to
+   * be reevaluated. Is this case the application is cloned and and the
+   * old version is marked as 'Overwritten'.
+   */
+  overwritten = "Overwritten",
 }
--- a/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
+++ b/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
@@ -1,14 +1,20 @@
-import { Controller, Param, Get } from "@nestjs/common";
+import {
+  Controller,
+  Param,
+  Get,
+  Post,
+  NotFoundException,
+  UnprocessableEntityException,
+} from "@nestjs/common";
 import { HasLocationAccess, AllowAuthorizedParty } from "../../auth/decorators";
 import { AuthorizedParties } from "../../auth/authorized-parties.enum";
 import {
   ApplicationService,
-  InstitutionService,
-  InstitutionLocationService,
+  INVALID_OPERATION_IN_THE_CURRENT_STATUS,
+  APPLICATION_NOT_FOUND,
+  WorkflowActionsService,
 } from "../../services";
 import { Application } from "../../database/entities";
-import { UserToken } from "../../auth/decorators/userToken.decorator";
-import { IInstitutionUserToken } from "../../auth/userToken.interface";
 import { COESummaryDTO } from "../application/models/application.model";
 
 @AllowAuthorizedParty(AuthorizedParties.institution)
@@ -16,8 +22,7 @@ import { COESummaryDTO } from "../application/models/application.model";
 export class ConfirmationOfEnrollmentController {
   constructor(
     private readonly applicationService: ApplicationService,
-    private readonly institutionService: InstitutionService,
-    private readonly locationService: InstitutionLocationService,
+    private readonly workflow: WorkflowActionsService,
   ) {}
 
   /**
@@ -26,12 +31,10 @@ export class ConfirmationOfEnrollmentController {
    * @param locationId location that is completing the COE.
    * @returns student application list of an institution location
    */
-  @AllowAuthorizedParty(AuthorizedParties.institution)
   @HasLocationAccess("locationId")
   @Get(":locationId/confirmation-of-enrollment")
   async getCOESummary(
     @Param("locationId") locationId: number,
-    @UserToken() userToken: IInstitutionUserToken,
   ): Promise<COESummaryDTO[]> {
     const applications = await this.applicationService.getCOEApplications(
       locationId,
@@ -47,5 +50,48 @@ export class ConfirmationOfEnrollmentController {
         lastName: eachApplication.student.user.lastName,
       };
     }) as COESummaryDTO[];
+  }
+
+  /**
+   * Creates a new Student Application overriding the current one
+   * in order to rollback the process and start the assessment
+   * all over again.
+   * @param locationId location id executing the COE rollback.
+   * @param applicationId application to be rolled back.
+   * @returns the id of the newly created Student Application.
+   */
+  @HasLocationAccess("locationId")
+  @Post(":locationId/confirmation-of-enrollment/:applicationId/rollback")
+  async startCOERollback(
+    @Param("locationId") locationId: number,
+    @Param("applicationId") applicationId: number,
+  ): Promise<number> {
+    try {
+      const result = await this.applicationService.overrideApplicationForCOE(
+        locationId,
+        applicationId,
+      );
+
+      if (result.overriddenApplication.assessmentWorkflowId) {
+        await this.workflow.deleteApplicationAssessment(
+          result.overriddenApplication.assessmentWorkflowId,
+        );
+      }
+
+      await this.applicationService.startApplicationAssessment(
+        result.createdApplication.id,
+      );
+
+      return result.createdApplication.id;
+    } catch (error) {
+      switch (error.name) {
+        case APPLICATION_NOT_FOUND:
+          throw new NotFoundException(error.message);
+        case INVALID_OPERATION_IN_THE_CURRENT_STATUS:
+          throw new UnprocessableEntityException(error.message);
+        default:
+          throw error;
+      }
+    }
   }
 }
--- a/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
+++ b/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
@@ -53,15 +53,17 @@ export class ConfirmationOfEnrollmentController {
   }
 
   /**
-   * Creates a new Student Application overriding the current one
-   * in order to rollback the process and start the assessment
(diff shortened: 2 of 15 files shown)


sources/packages/api/src/services/application/application.service.ts
```diff
@@ -638,4 +646,118 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        `Student Application is not in the expected status. The application must be in application status '${ApplicationStatus.enrollment}' and COE status '${COEStatus.required}' to be override.`,
+        INVALID_OPERATION_IN_THE_CURRENT_STATUS,
+      );
+    }
+
+    // Change status of the current application being overwritten.
+    appToOverride.applicationStatus = ApplicationStatus.overwritten;
+    appToOverride.applicationStatusUpdatedOn = getUTCNow();
+
+    // Create a new application based in the previous one.
```
abschwenker, line 688: Wording Suggestion: Create a new application record based off Overwritten record to rollback state of application

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -638,4 +646,118 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    await this.repo.save([appToOverride, newApplication]);
+
+    return {
+      overriddenApplication: appToOverride,
+      createdApplication: newApplication,
+    };
+  }
+
+  /**
+   * Starts an application assessment workflow associating
```
abschwenker, line 726: Question, is this specific for a rollback application or generic? If specific could be please review comments? Happy to discuss

abschwenker (commented):
Looks great! Please review comments. They are all comment comments, I see this being a headscratcher down the road so comments will be very helpful

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -26,12 +31,10 @@ export class ConfirmationOfEnrollmentController {
    * @param locationId location that is completing the COE.
    * @returns student application list of an institution location
    */
-  @AllowAuthorizedParty(AuthorizedParties.institution)
```
ann-aot, line 29: why is `AuthorizedParties.institution` removed

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -48,4 +51,47 @@ export class ConfirmationOfEnrollmentController {
...
+  /**
+   * Creates a new Student Application overriding the current one
+   * in order to rollback the process and start the assessment
+   * all over again.
+   * @param locationId location id executing the COE rollback.
+   * @param applicationId application to be rolled back.
+   * @returns the id of the newly created Student Application.
+   */
+  @HasLocationAccess("locationId")
+  @Post(":locationId/confirmation-of-enrollment/:applicationId/rollback")
```
ann-aot, line 64: how about `@Post(":locationId/confirmation-of-enrollment/application/:applicationId/rollback")` @andrewsignori-aot  @guru-aot  @dheepak-aot thoughts?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -638,4 +646,129 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+
+    // Change status of the current application being overwritten.
+    appToOverride.applicationStatus = ApplicationStatus.overwritten;
+    appToOverride.applicationStatusUpdatedOn = getUTCNow();
+
+    // Create a new application based in the previous one.
+    const newApplication = new Application();
+    newApplication.data = appToOverride.data;
+    newApplication.applicationNumber = appToOverride.applicationNumber;
+    newApplication.student = {
```
ann-aot, line 695: I think we can directly assign the object,  `newApplication.student = appToOverride.student`. can you check

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -638,4 +646,129 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+  async startApplicationAssessment(applicationId: number) {
+    const application = await this.repo.findOneOrFail(applicationId);
+    if (!application) {
+      throw new CustomNamedError(
+        "Student Application not found.",
+        APPLICATION_NOT_FOUND,
+      );
+    }
+
+    if (application.applicationStatus !== ApplicationStatus.submitted) {
```
ann-aot, line 751: got a suggestion in my last PR to use `Yoda notation` from @guru-aot . May be we can follow that in our application. just suggestion

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: This is being used during application submission and during COE Edit (this PR), that the one described in the PR description as below. Please let me know if further comments are needed.
"Create a centralized method to start the application workflow and assign the workflow id to the application. This method is being shared between the submit application method (already existent) and the new one added to COE;'

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
andrewsignori-aot: It is was present at the controller level and at the method level and there is no need to keep both.
I kept it only at the controller level.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
andrewsignori-aot: That sounds more descriptive and easy to understand to me. Happy to do the change 😉

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
ann-aot: got it 👍

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Here I took a different approach introducing the @RelationId to the relationships.
The main query does not contain the joins and the relations objects are not being loaded, in this way, we have direct access to the foreign keys that need to be associated with the cloned record. Please let me know if it makes sense.

image

sources/packages/api/src/database/entities/application-status.type.ts
```diff
@@ -30,7 +30,14 @@ export enum ApplicationStatus {
...
+   * The application has been cancelled by the student
    */
   cancelled = "Cancelled",
+  /**
+   * The application was replaced by a new version due to some event like
+   * an edit on Confirmation of Enrollment that forces the assessment to
+   * be reevaluated. Is this case the application is cloned and and the
+   * old version is marked as 'Overwritten'.
+   */
+  overwritten = "Overwritten",
```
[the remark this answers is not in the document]
andrewsignori-aot: Done

sources/packages/api/src/database/entities/application-status.type.ts
```diff
@@ -30,7 +30,14 @@ export enum ApplicationStatus {
    */
   completed = "Completed",
   /**
-   * TThe application has been cancelled by the student
+   * The application has been cancelled by the student
```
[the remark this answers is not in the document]
andrewsignori-aot: Done

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -638,4 +646,118 @@ export class ApplicationService extends RecordDataModelService<Application> {
       });
     return query.getOne();
   }
+
+  /**
+   * Creates a new Student Application overriding the current one
```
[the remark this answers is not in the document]
andrewsignori-aot: Done

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -48,4 +51,47 @@ export class ConfirmationOfEnrollmentController {
       };
     }) as COESummaryDTO[];
   }
+
+  /**
+   * Creates a new Student Application overriding the current one
```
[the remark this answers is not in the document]
andrewsignori-aot: Done

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Done

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
andrewsignori-aot: I did the change, @guru-aot and @dheepak-aot, please let me know if you have some different ideas or suggestions.

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: I don't mind if we have a proper justification. Why it would be a benefit to use it in typescript?

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
dheepak-aot: @andrewsignori-aot I'm good with Ann's suggestion.

sources/packages/api/src/services/application/application.service.ts
dheepak-aot: I see what you are doing here. with Cascade set to true on One-to-One, if we don't supply value for "student" instead just populating studentId, we assume studentId will override and take care "student: Student" doesn't become null.

And what happens if there we populate "student: Student" and does not supply studentId? I hope in either of scenario, it should not delete/set null to foreign key.

I asked this because I have less hands-on in typeorm. And it is slightly different from other orm frameworks which i previously worked.

sources/packages/api/src/services/application/application.service.ts
abschwenker: Nope I think that's good then. thank you for claifying

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: All good @dheepak-aot 😃 
So, the @RelationId works in only one direction as described [here](https://github.com/typeorm/typeorm/blob/master/docs/decorator-reference.md#relationid)
"_Relation id is used only for representation. The underlying relation is not added/removed/changed when chaining the value._"

It is just a shortcut to avoid adding a join to the query when we just need the id of the foreign key. We can use it to read the values but, to create relationships while saving the record we still need to do it in the usual way using the relationship object, student in this case.

@guru-aot this is what we talked about in the last sprint. I believe that we have good advantages doing it is this way and not adding the join just to get the ids 😉

dheepak-aot (approved):
LGTM

guru-aot (approved):
LGTM

ann-aot (approved):

ann-aot (commented):
LGTM

abschwenker (approved):
LGTM

PR #662, opened 2021-08-24
[vue][api][sims-#166] Institution User Confirms Student Enrolment (CoE)

ann-aot:
1. create a form definition to confirm COE
2. On click of "view" on the confirmation of enrolment summary screen for a required application show a form. This form is the same as the PIR form EXCEPT the student supplied program and offering information
2. form's fields are always read-only
3. On "Confirm Enrolment" update coe_status to "Submitted" and call camunda to continue the flow
4. add camunda wait message when waiting for coe_status to 'completed'
5. use camunda workflow message to update coe_status
6. use camunda workflow to update application_status to completed

image

image

image

image

image

### DIFF
--- a/sources/packages/web/src/components/institutions/modals/ConfirmCOEModal.vue
+++ b/sources/packages/web/src/components/institutions/modals/ConfirmCOEModal.vue
@@ -1,0 +1,85 @@
+<template>
+  <Dialog
+    v-if="showModal"
+    :visible="showModal"
+    :modal="true"
+    :style="{ width: '50vw' }"
+    :closable="false"
+  >
+    <template #header
+      ><p class="font-weight-bold h5">
+        Do you want to confirm this application?
+      </p></template
+    >
+    <p>
+      Confirming enrolment verifies this applicant as attending your institution
+      and will allow funding to be dispersed.
+    </p>
+    <template #footer>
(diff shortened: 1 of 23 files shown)


sources/packages/web/src/components/institutions/modals/ConfirmCOEModal.vue
```diff
@@ -0,0 +1,85 @@
...
+    :style="{ width: '50vw' }"
+    :closable="false"
+  >
+    <template #header
+      ><p class="font-weight-bold h5">
+        Do you want to confirm this application?
+      </p></template
+    >
+    <p>
+      Confirming enrolment verifies this applicant as attending your institution
```
abschwenker, line 15: please change as to is

abschwenker (changes requested):
I will have a look into code duplication from sonarqube. Can you please change so enrollment has 2 L's, I noticed in some cases you used 1.

sources/packages/web/src/components/institutions/modals/ConfirmCOEModal.vue
ann-aot: these contents, I copy-pasted from Figma. yes, I will update 👍

PR #675, opened 2021-08-30
[web][api][SIMS #532]Institution User Edits Student Enrolment (CoE)

andrewsignori-aot:
- Added the call in the COE to "Edit the Program Information". The menu item was renamed after a conversation with @abschwenker 
image

- Restrict API queries to do not consider Overwiritten applications while retrieving data from DB. Many queries on application service were changed.

- Added below modal dialog for confirmation of the CoE Edit
image

- Created a generic modal dialog based on Veutify and based on Promises instead of events (please let me know your thoughts about it).
The generic component has a type that allows us to introduce more features and also Vue slots that allow full customization of header, content, and footer, as shown below.

image

image

### DIFF
--- a/sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
+++ b/sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
@@ -46,4 +46,20 @@ export class ConfirmationOfEnrollmentApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async rollbackCOE(
+    locationId: number,
+    applicationId: number,
+  ): Promise<void> {
+    try {
+      await this.apiClient.post(
+        `institution/location/${locationId}/confirmation-of-enrollment/application/${applicationId}/rollback`,
+        null,
+        this.addAuthHeader(),
+      );
+    } catch (error) {
+      this.handleRequestError(error);
+      throw error;
+    }
+  }
 }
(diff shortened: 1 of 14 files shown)


sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
```diff
@@ -46,4 +46,20 @@ export class ConfirmationOfEnrollmentApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async rollbackCOE(
+    locationId: number,
+    applicationId: number,
+  ): Promise<void> {
+    try {
+      await this.apiClient.post(
```
guru-aot, line 55: is it possible to have a GET instead of a POST? it does not have a payload, so wanted to check, is there a reason to have it as post call?

guru-aot (approved):
LGTM, other than the doubt on the POST call and the failing Test cases

sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
guru-aot: Cool i get it now, thanks

andrewsignori-aot:
> LGTM, other than the doubt on the POST call and the failing Test cases

Needed some additional work on E2E tests. I just updated the PR with some modifications. I added a restriction to once the application is in some final state (Overwritten, Cancelled, Completed) it will no longer allow the change in the status and it caused some issues.

ann-aot (approved):
LGTM 👍 Liked the `ModalDialogBase` generic component with flexible slots

Issue #613, opened 2021-08-03
Student - View application (Read-Only)


michesmi:
As an (student) I want (view my application) So That (I can know what i submitted to SABC)

**Additional Context**
-

**Acceptance Criteria**
- [x] Add a button to "view application" (This is already there as a placeholder)
- [x] From the application details page. When a student clicks view. Show them a READ-ONLY version of the submitted application. This option will always be available

PR #678, opened 2021-08-30
Feature/[SIMS-#613]-Student - View application (Read-Only)

guru-aot:
From the application details page. When a student clicks view. Show them a READ-ONLY version of the submitted application. This option will always be available

- Removed Save Draft and Submit Application Button
- Made all the fields in the application readonly

Screen Shot 2021-08-30 at 3 29 43 PM

### DIFF
--- a/sources/packages/web/src/components/generic/formio.vue
+++ b/sources/packages/web/src/components/generic/formio.vue
@@ -20,6 +20,9 @@ export default {
     data: {
       type: Object,
     },
+    readOnly: {
+      type: String,
+    },
   },
   setup(props: any, context: SetupContext) {
     const formioContainerRef = ref(null);
@@ -76,6 +79,7 @@ export default {
         formDefinition.data,
         {
           fileService: new FormUploadService(),
+          readOnly: props.readOnly,
         },
       );
 
--- a/sources/packages/web/src/types/AppRoutes.ts
+++ b/sources/packages/web/src/types/AppRoutes.ts
@@ -28,6 +28,7 @@ export enum AppRoutes {
   InstitutionManageUsers = "manage-users",
   DisabledUser = "login/disabled-user",
   StudentApplication = "application-form/:selectedForm/programYear/:programYearId/application/:id",
+  StudentApplicationView = "application-form/:selectedForm/programYear/:programYearId/application/:id/:readOnly",
   StudentApplicationSummary = "my-application-summary",
   StudentApplicationDetails = "application/:id",
   Assessment = "application/:applicationId/assessment",
--- a/sources/packages/web/src/views/student/StudentApplicationDetails.vue
+++ b/sources/packages/web/src/views/student/StudentApplicationDetails.vue
@@ -1,5 +1,5 @@
-<template
-  ><div class="p-m-4">
+<template>
+  <div class="p-m-4">
     <h5 class="text-muted">
       <a @click="goBack()">
         <v-icon left> mdi-arrow-left </v-icon> Back to Applications</a
@@ -123,6 +123,18 @@ export default {
         },
       });
     };
+    const viewApplicaion = async () => {
+      await getProgramYear();
+      router.push({
+        name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM_VIEW,
+        params: {
+          selectedForm: programYear.value.formName,
+          programYearId: programYear.value.programYearId,
+          id: props.id,
+          readOnly: "readOnly",
+        },
+      });
+    };
     const loadMenu = () => {
       items.value = [
         {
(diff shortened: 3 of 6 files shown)


sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -86,6 +87,9 @@ export default {
       type: String,
       required: true,
     },
+    readOnly: {
+      type: String,
+    },
```
ann-aot, line 92: did you check, what happens when `readOnly is ""` (when `readOnly` is not passed)?

sources/packages/web/src/components/generic/formio.vue
```diff
@@ -20,6 +20,9 @@ export default {
     data: {
       type: Object,
     },
+    readOnly: {
+      type: String,
+    },
```
ann-aot, line 25: did you check, what happens when `readOnly is ""` (when `readOnly` is not passed)?

sources/packages/web/src/components/generic/formio.vue
```diff
@@ -76,6 +79,7 @@ export default {
         formDefinition.data,
         {
           fileService: new FormUploadService(),
+          readOnly: props.readOnly,
```
ann-aot, line 82: did you check, what happens when `readOnly is ""` (when `readOnly` is not passed)?

ann-aot (commented):
Again, you missed PR title :)

sources/packages/web/src/types/AppRoutes.ts
```diff
@@ -28,6 +28,7 @@ export enum AppRoutes {
   InstitutionManageUsers = "manage-users",
   DisabledUser = "login/disabled-user",
   StudentApplication = "application-form/:selectedForm/programYear/:programYearId/application/:id",
+  StudentApplicationView = "application-form/:selectedForm/programYear/:programYearId/application/:id/:readOnly",
```
andrewsignori-aot, line 31: Do we need a new route for this? Maybe we just need to change the ":readOnly" to be optional or, if there will be a different route, do we need ":readOnly" as a parameter?

sources/packages/web/src/views/student/StudentApplicationDetails.vue
```diff
@@ -136,6 +148,9 @@ export default {
         {
           label: "View",
           icon: "pi pi-fw pi-folder-open",
+          command: () => {
```
andrewsignori-aot, line 151: This could be simplified to:
```
command: viewApplicaion
```

sources/packages/web/src/components/generic/formio.vue
andrewsignori-aot: I would also be looking forward to knowing what would be the default value.
Whatever it is it could be the default value of the property also.

guru-aot: Yes @ann-aot I tested, you can check it by running locally, the edit application, where the readonly is null as it is not passed

sources/packages/web/src/types/AppRoutes.ts
guru-aot: I created it optional, but do not want mess with the edit route as its used for more than one place.

sources/packages/web/src/views/student/StudentApplicationDetails.vue
guru-aot: just copied from edit application :) will change both to sync

sources/packages/web/src/types/AppRoutes.ts
andrewsignori-aot: For the first point "I created it optional, but do not want mess with the edit route as its used for more than one place", it will not mess with current routes, it is backward compatible 😉 

For the second point "I prefer to have it 2 and pass the parameter, as we can use it in the form to show/hide the buttons", the property name is readOnly and it is receiving a string "readOnly" that is not being used in any check. The check is based in the fact that there is something or not in the parameter. So, if I provide "anything" in the readOnly parameter it will assume that the form will be readonly, right?

guru-aot: I agree with the first point, the second point I intended to create that way. Is that wrong?

andrewsignori-aot: I would not say wrong, but to me, it is weird that I would access the URL like below and the form will be readonly 😉 
```
application-form/THEFROM/programYear/1/application/126/I_WANT_THE_FORM_TO_BE_EDITABLE
```

andrewsignori-aot: Please do not consider this issue as a blocker.

andrewsignori-aot (approved):
👍

abschwenker (approved):
LGTM

Issue #664, opened 2021-08-25
Add institution Code on Institution Location


michesmi:
**Describe the task**
- We Include this is the files we send and receive with CSFA and is used as an identifier for an institution

**Acceptance Criteria**
- [x] 4 Upper Case Letters free field on the institution location page
- [x]  This is a required field- (going to be nullable in the database.
- [x] Institution Locations can share the same code.

**Tasks**
- [x] Migration to add column institution_code    (nullable)
- [x] Add field on form
- [x] Add to API to save/retrieve
- [x] Manage Institution -> Location Summary -> Add&Edit should use the form "Institution Location"

PR #679, opened 2021-08-31
[web][api][SIMS #664] Add institution Code on Institution Location

ann-aot:
- 4 Upper Case Letters free field on the institution location page
-  This is a required field- (going to be nullable in the database.
-  Institution Locations can share the same code.
- Migration to add column institution_code (nullable)
- Add field on form
- Add to API to save/retrieve
- Manage Institution -> Location Summary -> Add&Edit should use the form "Institution Location"


sources/packages/api/src/database/entities/institution-location.model.ts
```diff
@@ -40,4 +40,10 @@ export class InstitutionLocation extends RecordDataModel {
     referencedColumnName: ColumnNames.ID,
   })
   institution: Institution;
+
+  @Column({
+    name: "institution_code",
+    nullable: false,
```
andrewsignori-aot, line 46: Could you please remove the "nullable: false" or change it to nullable true, since it accepts null?

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -66,7 +66,6 @@ export class InstitutionLocationsController extends BaseController {
         "Not able to find a institution associated with the current user name.",
       );
     }
-
     // If the data is valid the location is saved to SIMS DB.
```
andrewsignori-aot, line 69: Please consider do not include this file in the PR if there is no modifications related to this PR.

sources/packages/api/src/route-controllers/institution-locations/models/institution-location.dto.ts
```diff
@@ -14,6 +14,7 @@ export interface InstitutionLocationTypeDto {
   locationName: string;
   postalZipCode: string;
   provinceState: string;
+  institutionCode: string;
```
andrewsignori-aot, line 17: This DTO is used to create/update(POST/PATCH) the institution.
During the POST I can see a dry run that is running the validations, but this does not happen on update/PATCH.
Can you please add the dry/run on the update also?
The reason that I am asking is this PR is because the new field, during the update, is not receiving any server validation and it would fail only on the database if it has more than 4 characters, what means that I can still save invalid data.

andrewsignori-aot (commented):
Overall looks good but please take a look at the server validations during the update. I believe that we have an existing gap that I also believe that we can fix in this PR.

sources/packages/api/src/route-controllers/institution-locations/models/institution-location.dto.ts
ann-aot: DB wont allow more than 4 characters, but it will allow less than 4 character and lower case and all other characters, yes, it make sens to add dry/run in patch. updating

sources/packages/api/src/route-controllers/institution-locations/institution-locations.controller.ts
```diff
@@ -48,7 +48,7 @@ export class InstitutionLocationsController extends BaseController {
   ): Promise<number> {
     // Validate the location data that will be saved to SIMS DB.
     const dryRunSubmissionResult = await this.formService.dryRunSubmission(
-      "institutionlocationcreation",
+      "institutionlocation",
```
andrewsignori-aot, line 51: I would avoid changing it on this PR. The POST must include the email from the user also.

andrewsignori-aot, line 51: What is the difference between the two Form definitions?
If we are changing it we need also to add the definition in Git.

sources/packages/forms/src/form-definitions/institutionlocation.json
```diff
@@ -0,0 +1,1613 @@
+{
+    "_id": "612fa7e794d56e2366d54163",
```
andrewsignori-aot, line 2: Please remove the id since it makes sense only in the environment where it was generated.

sources/packages/forms/src/form-definitions/institutionlocation.json
```diff
@@ -0,0 +1,1613 @@
...
+        }
+    ],
+    "display": "form",
+    "title": "Institution Location",
+    "name": "institutionlocation",
+    "path": "institutionlocation",
+    "machineName": "institutionlocation",
+    "created": "2021-06-28T22:06:38.576Z",
+    "modified": "2021-09-01T16:18:47.856Z",
+    "access": [
```
andrewsignori-aot, line 1600: Please remove the access since it makes sense only in the environment where it was generated.

andrewsignori-aot (approved):
Looks good. Nice work organizing the form definitions also 👍

dheepak-aot (approved):
LGTM

PR #680, opened 2021-08-31
[web][api][SIMS #532]Fix modal dialog and user name

andrewsignori-aot:
- Found an issue on the modal dialog for COE when the user clicks outside the modal or cancels the model by clicking in ESC.
- Found an issue while displaying the user name for mononymous users.


abschwenker (approved):
LGTM

guru-aot (approved):
LGTM

sources/packages/api/src/utilities/auth-utils.ts
```diff
@@ -54,14 +54,16 @@ export function setGlobalPipes(app: INestApplication) {
...
  * @returns user full name.
  */
 export function getUserFullName(user: {
   firstName: string;
   lastName: string;
 }): string {
-  return `${user.firstName} ${user.lastName}`.trim();
+  return `${(user.firstName ?? "").trim()} ${(
+    user.lastName ?? ""
+  ).trim()}`.trim();
```
dheepak-aot, line 68: trim() is used for $firstName and $lastName already, so is it again required at the end?

andrewsignori-aot: Yes, because there is a white space in the middle that should be removed in case firstName or lastName are null or empty.

dheepak-aot: Okay. Makes sense. Thanks

dheepak-aot (commented):
LGTM

Issue #339, opened 2021-04-06
Form validation error message not showing


mzmichellesmith:
Additional context
Ticket #534 removed the application validation errors on submit application

Tasks
- [x] ensure submit application validation errors show up for the student
- [x] ensure the formio submit successful message is removed

PR #691, opened 2021-09-02
[vue-css][sims-#339] hide only success formio alert, show danger alerts

ann-aot:
- `alert` is a common class, because of that  all alerts were hidden, updated CSS only to hide success alert

formio is using bootstrap and it accepts bootstrap class if we want to add bootstrap alter-success class to any form component, it won't work. but we can override it by adding an extra class to show the alert-success or we can create a custom alert-success  equivalent to bootstrap success-alert

### DIFF
--- a/sources/packages/web/src/assets/css/formio-shared.css
+++ b/sources/packages/web/src/assets/css/formio-shared.css
@@ -48,6 +48,6 @@
   line-height: 18px;
 }
 
-.ff-form-container .alert{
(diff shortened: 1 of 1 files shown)


abschwenker (approved):

dheepak-aot (approved):

andrewsignori-aot (approved):
👍

guru-aot (approved):
LGTM

Issue #395, opened 2021-04-27
API modifications - Verify Existing Institution On institution login


mzmichellesmith:
**Task 1:**

When an institution exists and a new institution user authenticates before they are added to the institution
User=New
Institution=Exists
- [x] New user logging into an institution (before being added by their admin) should be redirected back to the login saying that you have not been added

**Task 2:**
When an institution exists and a new institution user authenticates before they are added to the institution
User=New
Institution=New
- [x] New user and new institution logging in should be directed to complete institution account profile and created as an institution account admin after the account is created

**Task 3:**
- [x] Review/Remove Duplicate "check user" on files AuthService.ts and AppInstitution.vue, as shown below.

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/1eaf2fe3-51ef-46b3-80f1-6097299965db)

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/3e1d20bf-94aa-46fb-a72e-73ecc9eb5228)

PR #692, opened 2021-09-02
[web] [api] [SIMS #395] API modifications - Verify Existing Institution On institution login

dheepak-aot:
- New HTTP head API added to check if an Institution exist
- Auth service logic modified to handle the following scenario (User does not exist in our system after successful bceid authentication for a new and existing institution)
- Complexity of the AuthService in web resulted in Sonar violation. Hence created a private method which handles institution navigation flow.
- Duplicate logic in AppInstitution.Vue has been removed.


sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
```diff
@@ -8,11 +8,19 @@
...
       <formio
+        v-if="editMode"
         formName="institutionprofile"
         :data="initialData"
         @loaded="formLoaded"
         @submitted="submitted"
       ></formio>
+      <formio
+        v-if="!editMode"
+        formName="institutionprofilecreation"
```
ann-aot, line 19: can we add something like this `:formName="editMode ? 'institutionprofile' : 'institutionprofilecreation'"` and reduce LOC

sources/packages/web/src/services/http/Institution.ts
```diff
@@ -115,4 +115,23 @@ export class InstitutionApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async checkIfExist(guid: string, headers: any): Promise<boolean> {
+    try {
+      const resp: AxiosResponse<void> = await this.apiClient.head(
+        "institution/" + guid,
```
ann-aot, line 122: can we use **Template literals** for the sake of consistency, as we follow  **Template literals** in all other Axios calls

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
dheepak-aot: This is exactly how I initially coded but formio is unable to read expressions when they are written in param(s). I got a 400 error.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -408,6 +408,10 @@ export class InstitutionService extends RecordDataModelService<Institution> {
     });
   }
 
+  async doesExist(guid: string): Promise<number> {
+    return this.repo.count({ guid: guid });
```
andrewsignori-aot, line 412: Please avoid using counting to check if a record exists. The count will force the query to process all the records in the table instead of stopping on the first match. Please consider changing it to search using a "where" like syntax.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -408,6 +408,10 @@ export class InstitutionService extends RecordDataModelService<Institution> {
     });
   }
 
+  async doesExist(guid: string): Promise<number> {
```
andrewsignori-aot, line 411: I would recommend changing the return type to a boolean. A boolean value should fit better the question "doesExist".

sources/packages/web/jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   transform: {
     "^.+\\.vue$": "vue-jest",
   },
+  transformIgnorePatterns: ["/node_modules/(?!primevue)"],
```
andrewsignori-aot, line 6: Good catch having this fixed 😉

sources/packages/web/src/services/http/Institution.ts
```diff
@@ -115,4 +115,23 @@ export class InstitutionApi extends HttpBaseClient {
...
+        headers,
+      );
+      if (200 === resp.status) {
```
andrewsignori-aot, line 125: Don't need to check for 200 at this point. If the request was not successful Axios will throw an exception, so at this point, it is safe to return true.

sources/packages/web/src/services/http/UserApi.ts
```diff
@@ -20,7 +20,7 @@ export class UserApi extends HttpBaseClient {
     }
   }
 
-  public async bceidAccount(headers?: any): Promise<BCeIDDetailsDto | null> {
+  public async bceidAccount(headers?: any): Promise<BCeIDDetailsDto> {
```
andrewsignori-aot, line 23: I noticed that this method is also returning "null" from the API side. Is there a reason to change it in the Vue application?

sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -92,21 +88,6 @@ export default {
         router.push({
```
andrewsignori-aot, line 88: Do we still need this check here or the guard will take care of it?

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
ann-aot: @dheepak-aot , i checkout to this branch in my local and tried its working for me

andrewsignori-aot: Please consider reducing the amount of code as @ann-aot suggested.
I would recommend creating a computed value to return the form name with the logic that Ann proposed inside. Both ways work and I would be good either way 😄

sources/packages/web/src/types/AppRoutes.ts
```diff
@@ -27,6 +27,7 @@ export enum AppRoutes {
   ManageInstitutionDesignation = "manage-desgination",
   InstitutionManageUsers = "manage-users",
   DisabledUser = "login/disabled-user",
+  UnknownUser = "login/unknown-user",
```
andrewsignori-aot, line 30: "UnknownUser" is a bit tricky IMO. We know the user, the user is valid, the user belongs to that institution and the user knows it.
I would change it to something that indicates that the user is valid but is not allowed to access the system and he needs permission from the administrator.
The variable name could be something like:
- userNotAllowedInApplication
- userNotInManageUsers
It is not a blocker if the Team is ok with it.

sources/packages/web/src/views/institution/Login.vue
```diff
@@ -33,6 +33,10 @@
       Disabled User - you dont have access to the system. Please contact
       Administrator for more informations.
     </Message>
+    <Message severity="error" class="mx-2" v-if="unknownUser">
+      Institution already exists. Please contact Administrator for more
```
andrewsignori-aot, line 37: I believe that we could be a bit more clear about the fact that the user exists and it is valid on BCeID but he is not allowed to access the application.
I would recommend something lie below:

_"The user was validated with success but it is not currently allowed to have access to this application. Please contact the Administrator for more information."_

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
dheepak-aot: @andrewsignori-aot Yes but as I mentioned, expression was throwing error so I stepped back.  I will have  screen share with @ann-aot and see what's going wrong at my side.

sources/packages/web/src/views/institution/Login.vue
dheepak-aot: Thanks. This make more sense. I will change to this.(with slight modification)
"The user was validated successfully but not currently allowed to have access to this application. Please contact the Administrator for more information."

dheepak-aot: If that is ok

sources/packages/web/src/services/UserService.ts
```diff
@@ -18,14 +18,8 @@ export class UserService {
...
 
-  async getBCeIDAccountDetails(
-    authHeader?: any,
-  ): Promise<BCeIDDetailsDto | null> {
-    try {
-      return await ApiClient.User.bceidAccount(authHeader);
-    } catch (excp) {
-      return null;
-    }
+  async getBCeIDAccountDetails(authHeader?: any): Promise<BCeIDDetailsDto> {
```
andrewsignori-aot, line 21: Removing the option to this method no longer returns null will affect the below code?
image

sources/packages/web/src/services/AuthService.ts
```diff
@@ -4,8 +4,58 @@ import { AppConfig, ClientIdType } from "../types/contracts/ConfigContract";
...
+import { InstitutionService } from "./InstitutionService";
+import router from "../router";
+import { InstitutionRoutesConst } from "@/constants/routes/RouteConstants";
+import { BCeIDDetailsDto } from "@/types";
 let keycloak: Keycloak.KeycloakInstance;
 
+async function navigateForInstitution(
+  authHeader: any,
+  bceIdAccountDetails: BCeIDDetailsDto,
+) {
```
andrewsignori-aot, line 16: Please add the explicit return type Promise<boolean>.

sources/packages/web/src/views/institution/Login.vue
andrewsignori-aot: I liked it 👍 
I believe that there is just a "is" missing in the middle of it. Either way, it sounds good to me.
_"The user was validated successfully but is not currently allowed to have access to this application. Please contact the Administrator for more information."_

sources/packages/web/src/types/AppRoutes.ts
ann-aot: mmm, IMO, `unknown-user` seems more close, like a valid user, which is unknown to our application

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
dheepak-aot: It is fixed now. It was missing ":" before the param name.

sources/packages/web/src/types/AppRoutes.ts
andrewsignori-aot: It is fine for me 😉

guru-aot (approved):
LGTM

sources/packages/web/src/views/institution/DynamicInstitutionProfile.vue
dheepak-aot: @ann-aot param modified with expression

sources/packages/web/src/services/http/Institution.ts
dheepak-aot: Template literals has been used as discussed.

sources/packages/api/src/services/institution/institution.service.ts
dheepak-aot: As per discussion we are going to stick with count in this scenario.

sources/packages/api/src/services/institution/institution.service.ts
dheepak-aot: Modified the return type to boolean

sources/packages/web/src/services/http/Institution.ts
dheepak-aot: Fixed.

sources/packages/web/src/views/institution/Login.vue
dheepak-aot: Verbiage modified.

sources/packages/web/src/services/AuthService.ts
dheepak-aot: Promise has been explicitly defined.

ann-aot (approved):
LGTM 👍

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -408,6 +408,14 @@ export class InstitutionService extends RecordDataModelService<Institution> {
     });
   }
 
+  async doesExist(guid: string): Promise<boolean> {
```
andrewsignori-aot, line 411: @ann-aot, @guru-aot, and @dheepak-aot, I will not block this PR due to this, but please in the future consider not using the count to check if there is a row present in the DB. It will not have a major impact because the table will be small, but I will never consider this as the right approach if we need to do something in tables like sims.applications.
This is the below Postgres analysis when using count:

_select count(1) from sims.institutions where guid = 'ca813fd2-f9b7-4762-83f5-9eb1d244d022'_
image

image

This is the Postgres analysis when using a simple where:

_select 1 from sims.institutions where guid = 'ca813fd2-f9b7-4762-83f5-9eb1d244d022'_

image

image

sources/packages/web/src/views/institution/Login.vue
```diff
@@ -53,17 +58,23 @@ export default {
...
+    showUnknownUserMessage: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
   },
   setup(props: any) {
     const basicBCeID = ref(props.showBasicBCeIDMessage);
     const disabledUser = ref(props.showDisabledUserMessage);
+    const unknownUser = ref(props.showUnknownUserMessage);
```
andrewsignori-aot, line 70: I do not see any reason to use "ref" here since we can directly access the properties on the template.
Since the others are already following this it will not be a problem for this PR.
In the future, we do not need to use ref either return them on the setup method.

andrewsignori-aot (approved):
Good work cleaning the code on the AppIinstitution and ensuring that we can create new institutions now 👍 😉

abschwenker (approved):

Issue #611, opened 2021-08-03
Institution Admin Location Authorization


andrewsignori-aot:
**Additional Context**
While the decorator @HasLocationAccess is validating the access for Institution Admins, the validation for access to a particular location is not happening, it is just skipped for Institutions Admins because we don't have the complete list of institutions ids that Admins have access to.
In order to fix it we need to:

**Tasks**
@andrewsignori-aot  can give an overview if unclear
- [x] On the jwt.strategy.ts, while loading the information for Institution Admins, load also the locations ids that the users have access (all locations ids for that institution);
- [x] On institution-location.guard.ts, stop skipping the location verification for institution admins and check if the Institution Admin user has access to the particular location;
- [x] Search and adjust authorization where the decorator @HasLocationAccess is used and where the URL contains ":locationId"

From 582
- [x] Remove 
![image.png](https://images.zenhubusercontent.com/608872f7d0d25cdfe8dab75d/8068e109-c9af-4de4-b96d-265f7230891e) 
from sources/packages/api/src/auth/guards/institution-location.guard.ts 
- [x] Loads admin locations

andrewsignori-aot:
While the decorator **@HasLocationAccess** is validating the access for Institution Admins, the validation for access to a particular location is not happening, it is just skipped for Institutions Admins because we don't have the complete list of institutions ids that Admins have access to.
In order to fix it we need to:

1 - On the _jwt.strategy.ts_, while loading the information for Institution Admins, load also the locations ids that the users have access (all locations ids for that institution);
2 - On _institution-location.guard.ts_, stop skipping the location verification for institution admins and check if the Institution Admin user has access to the particular location;
3 - Search and adjust authorization where the decorator _@IsInstitutionAdmin()_ is used and where the URL contains "_:locationId_"

PR #694, opened 2021-09-02
[api][SIMS #611]Institution Admin Location Authorization

andrewsignori-aot:
- Load all locations IDs when an Institution Admin is authenticated;
- Decorator @HasLocationAccess is now validating if Institution Admins have access to his locations only;
- Removed non longer necessary authorizations checks;
- Created unit tests for institution-location.guard;
- Created test helpers to mock the HTTP context.


sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -144,4 +140,23 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+   * Gets all locations ids for a particular institution.
+   * This method is used during the login process and should be
+   * as lightweight as possible. Do not expand this query unless
+   * it is related to login/authorization process.
+   * @param institutionId institution id.
+   * @returns institution locations ids.
+   */
+  async getInstitutionLocationsIds(institutionId: number): Promise<number[]> {
+    const allLocations = await this.repo
+      .createQueryBuilder("locations")
```
ann-aot, line 154: table name is `institution_location` right? am I missing something?

andrewsignori-aot: This is not a table name, it is the alias to execute the query. IMO using the full table name as an alias doesn't make sense and it also makes the code longer and hard to read.
Please let me know if it makes sense.

ann-aot (approved):
LGTM 👍

guru-aot (approved):
LGTM, good job on the unit test cases

dheepak-aot (approved):
LGTM

abschwenker (approved):

Issue #606, opened 2021-07-30
Institution User Declines Request for (Confirmation of Enrolment) 


michesmi:
**As a** *(Institution User)* **I want** *(Declines a Confirmation of Enrolment Request)* **So That** *(only Confirmed Enrolled Students in my program receive Finacial Aid from SABC)*

**Additional Context**
Related Tickets Include:
- Confirmation of Enrolment #166
- Edit Student Application #607

**Acceptance Criteria**
- [x] Given (an Institution User, When (they select 'Decline Request' on the 'View Financial Aid Application page'), Then (provide the following pop-up "Decline Confirmation of Enrollment", "Declining this request will also stop this application from moving forward. If you would still like to decline this application please select the appropriate reason below. Your denial reason will be visible to the applicant" "Select the reason why you are declining this application"  and the following options as radio buttons:
   - [x] School is unable to confirm student identity, please confirm your student number or contact Financial Student Aid Office for more information
  - [x]  Study period dates selected are incorrect, please submit an application for each semester individually
  - [x]  Study period dates are incorrect, please contact Financial Aid Office for more information
  - [x]  Programs you are registered in is not eligible for Student Aid BC funding, please contact Financial Aid Office for more information
  - [x]  Our records indicate you are not registered, please re-submit the application after registration
  - [x]  Other
      - [x] Given the (Institution User) When, (Selects  'Other') Then  (provide, "Please explain your reason for declining this application" (_Long text box_)
- [x] AND, Provide the Option to 'Cancel' and 'Decline Request'
     - [x] Given the User selects 'Cancel', then direct the user back to the Application Summary page.
     - [x] Given the User selects ' Decline Request', then change the Confirmation of Enrolment to ' Declined' 
           - [ ] AND provide the pop-up in Red, "This confirmation of enrolment has been declined"  and (provide the reason selected from the drop-down. 
           - [ ] AND provide the pop-up in Green, "Enrolment Declined"

Tasks
- [x] On click of "view" on the confirmation of enrolment summary screen for a required application show a form. This form is the same as the PIR form EXCEPT the student supplied program and offering information (already built)
- [x] On "Decline Enrolment" update coe_status to "Declined" (see PIR for reference)
- [x] Migration for coe_declined_reason (see PIR for reference)
- [x] Types stored in table
- [x] Modal in Vue Application

PR #701, opened 2021-09-04
[web][api][SIMS-#606] Institution User Declines Request for (Confirmation of Enrolment)

ann-aot:
Tasks
- [x] On click of "view" on the confirmation of enrolment summary screen for a required application show a form. This form is the same as the PIR form EXCEPT the student supplied program and offering information (already built)
- [x] On "Decline Enrolment" update coe_status to "Declined" (see PIR for reference)
- [x] Migration for coe_declined_reason (see PIR for reference)
- [x] Types stored in a table
- [x] Modal in Vue Application
- [x] `declineconfirmationofenrollment` definition added
- [x] COE action information is now a component
- [x] Moved all COE modals insides COE folder in components
- [x] removed `sources/packages/web/src/types/contracts/students` as there was already `sources/packages/web/src/types/contracts/student/`
- [x] I didn't find any function to add dynamic options for formio radio component, so I fetched the component and added the dynamic options, and redrew the formio, let me know why you guys know any direct method or any better method
- [x] Sonarcloud duplication is due to the long interfaces added in `web` and `api`. let me know if you guys have any better way to avoid this Sonarcloud duplication


sources/packages/api/src-sql/COEDeniedReason/Create-coe-denied-reason.sql
```diff
@@ -0,0 +1,28 @@
+CREATE TABLE IF NOT EXISTS coe_denied_reason(
```
andrewsignori-aot, line 1: In the beginning, we were using all the tables with plural names and I believe that we lost track of it when few tables were created not following it.
I would recommend stop creating tables in singular and keep all of them from now on using the plural.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -207,4 +224,55 @@ export class ConfirmationOfEnrollmentController {
...
+        payload.coeDenyReasonId,
+        payload.otherReasonDesc,
+      );
+    } catch (error) {
+      if (error.name === COE_REQUEST_NOT_FOUND_ERROR) {
+        throw new UnprocessableEntityException(error.message);
+      }
+
+      throw new InternalServerErrorException(
+        "Error while denying a  Confirmation Of Enrollment (COE).",
```
andrewsignori-aot, line 258: Please remove the extra white spaces in the string.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -207,4 +224,55 @@ export class ConfirmationOfEnrollmentController {
...
+      );
+    }
+  }
+
+  /**
+   * Get all COE denied reason ,which are active
+   * @returns COE denied reason list
+   */
+  @Get("confirmation-of-enrollment/denial-reasons")
+  async getCOEDeniedReason(): Promise<COEDeniedReasonDto[]> {
```
andrewsignori-aot, line 268: Just a suggestion, we already have on API and Vue a key/value pair object name OptionItemDto to map this type of return. If this is something that could be changing in the future we can keep creating specific classes. I would be ok either way.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -207,4 +224,55 @@ export class ConfirmationOfEnrollmentController {
...
+    const coeDeniedReason =
+      await this.DeniedCOEReasonService.getCOEDeniedReasons();
+    return coeDeniedReason.map((eachCOEDeniedReason) => {
```
andrewsignori-aot, line 271: The code could be simplified as below
```
 return coeDeniedReason.map((eachCOEDeniedReason) => ({
     value: eachCOEDeniedReason.id,
     label: eachCOEDeniedReason.reason,
 }));    
```

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -185,4 +186,22 @@ export class Application extends RecordDataModel {
...
+  @JoinColumn({
+    name: "coe_denied_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  coeDeniedReason?: COEDeniedReason;
+
+  @Column({
+    name: "coe_denied_other_desc",
+  })
+  coeDeniedOtherDesc: string;
```
andrewsignori-aot, line 206: I believe that there is an "?" missing since the column is optional.
```
coeDeniedOtherDesc?: string;
```

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -916,6 +920,7 @@ export class ApplicationService extends RecordDataModelService<Application> {
       .innerJoinAndSelect("student.user", "user")
       .innerJoinAndSelect("application.offering", "offering")
       .innerJoinAndSelect("offering.educationProgram", "educationProgram")
+      .innerJoinAndSelect("application.coeDeniedReason", "coeDeniedReason")
```
andrewsignori-aot, line 923: Is it really supposed to be an inner join?
It seems that it is a left join since the reason will be something to be added later, does it make sense or am I missing something?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -950,4 +955,52 @@ export class ApplicationService extends RecordDataModelService<Application> {
       ) <= COE_WINDOW
     );
   }
+  /**
+   * Deny the Confirmation Of Enrollment(COE) for an Application.
+   * Updates only applications that have the COE status as required.
+   * @param applicationId application id to be updated.
+   * @param locationId location that is setting the offering.
```
andrewsignori-aot, line 962: It seems that "location that is setting the offering" does not apply to this method.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -950,4 +955,52 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+          ApplicationStatus.completed,
+          ApplicationStatus.overwritten,
+          ApplicationStatus.cancelled,
+        ]),
+      ),
+    });
+    if (!application) {
+      throw new CustomNamedError(
+        "Not able to find an application that requires a COE to be completed.",
+        COE_REQUEST_NOT_FOUND_ERROR,
```
andrewsignori-aot, line 989: I would suggest using the generic existing constant error INVALID_OPERATION_IN_THE_CURRENT_STATUS that, alongside the message will give the right error context and will allow the proper HTTP staus to be returned from the controller consuming this method.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -950,4 +955,52 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      );
+    }
+
+    application.coeDeniedReason = {
+      id: coeDeniedReasonId,
+    } as COEDeniedReason;
+    if (COE_DENIED_REASON_OTHER_ID === coeDeniedReasonId && !otherReasonDesc) {
+      throw new CustomNamedError(
+        "Other is selected as COE reason, specify the reason for the COE denial.",
+        COE_DENIED_REASON_NOT_FOUND_ERROR,
```
andrewsignori-aot, line 999: I noticed that this constant is created in a common/generic constant file and I am not sure what we are following here. The idea is to have all the constants under the same file because they are constants?
I would recommend creating the constants in a context where they are more meaningful.
For instance, if the idea is not to have the constants in the same service class where it will be mainly used, we can create a const file for that service.

sources/packages/api/src/services/coe-denied-reason/coe-denied-reason.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable, Inject } from "@nestjs/common";
+import { RecordDataModelService } from "../../database/data.model.service";
+import { Connection } from "typeorm";
+import { COEDeniedReason } from "../../database/entities/coe-denied-reason.model";
+
+@Injectable()
+export class COEDeniedReasonService extends RecordDataModelService<COEDeniedReason> {
+  constructor(@Inject("Connection") private readonly connection: Connection) {
```
andrewsignori-aot, line 8: Please remove the "private readonly" as it is not needed.
I noticed this along with many classes but it is probably generating a warning because the connection is not used anywhere besides the constructor.

sources/packages/api/src/utilities/constants.ts
```diff
@@ -2,3 +2,7 @@
 export const EXTENDED_DATE_FORMAT = "MMMM, D YYYY";
 export const PIR_DENIED_REASON_OTHER_ID = 1;
 export const COE_WINDOW = 21;
+export const COE_DENIED_REASON_OTHER_ID = 1;
+export const COE_REQUEST_NOT_FOUND_ERROR = "COE_REQUEST_NOT_FOUND_ERROR";
```
andrewsignori-aot, line 6: I would recommend move COE_DENIED_REASON_OTHER_ID and COE_REQUEST_NOT_FOUND_ERROR to the service class. I gave some further explanations on why in the previous comment about the same constant. Always open to discussion.

sources/packages/web/src/components/institutions/confirmation-of-enrollment/information.vue
```diff
@@ -0,0 +1,105 @@
...
+  setup(props: any) {
+    const content = ref("");
+    const containerClass = ref("");
+    const infoHeader = ref("");
+    const infoHeaderClass = ref("");
+    const iconColor = ref("");
+    const showInfo = ref(false);
+    const prepareContent = () => {
+      showInfo.value = false;
+      if (
```
andrewsignori-aot, line 36: Just as a recommendation, it could be simplified to avoid repeating "_props.data.applicationCOEStatus_" as below:
```
if([COEStatus.submitted, COEStatus.completed].includes(props.data.applicationCOEStatus)){
    // Do seomthing
}
```

sources/packages/web/src/services/ConfirmationOfEnrollmentService.ts
```diff
@@ -42,4 +47,20 @@ export class ConfirmationOfEnrollmentService {
...
   }
+
+  public async getCOEDenialReasons(): Promise<COEDeniedReasonDto> {
+    return ApiClient.ConfirmationOfEnrollment.getCOEDenialReasons();
+  }
+
+  public async denyConfirmationOfEnrollment(
+    locationId: number,
+    applicationId: number,
+    DenyCOEPayload: DenyConfirmationOfEnrollment,
```
andrewsignori-aot, line 58: Please adjust the parameter to match the other letter cases: _denyCOEPayload_ instead of _DenyCOEPayload_.

sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
```diff
@@ -62,4 +67,34 @@ export class ConfirmationOfEnrollmentApi extends HttpBaseClient {
       throw error;
     }
   }
+
+  public async getCOEDenialReasons(): Promise<COEDeniedReasonDto> {
+    try {
+      const response = await this.apiClient.get(
+        `institution/location/confirmation-of-enrollment/denial-reasons`,
```
andrewsignori-aot, line 74: Just as recommendation, since there is no string interpolation, the code could be changed as below:
```
"institution/location/confirmation-of-enrollment/denial-reasons"
```

sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
```diff
@@ -62,4 +67,34 @@ export class ConfirmationOfEnrollmentApi extends HttpBaseClient {
...
+    } catch (error) {
+      this.handleRequestError(error);
+      throw error;
+    }
+  }
+
+  public async denyConfirmationOfEnrollment(
+    locationId: number,
+    applicationId: number,
+    DenyCOEPayload: DenyConfirmationOfEnrollment,
```
andrewsignori-aot, line 87: Please rename _DenyCOEPayload_ to _denyCOEPayload_.

sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/ApplicationDetailsForCOE.vue
```diff
@@ -179,7 +136,29 @@ export default {
...
+        });
+      } catch {
+        toast.error(
+          "Unexpected error",
+          "An error happened while denying Confirmation of Enrollment.",
+        );
+      }
+    };
+    const denyProgramInformation = async () => {
+      await denyCOEModal.value.showModal();
```
andrewsignori-aot, line 160: IMO, if the modal is already awaiting a response, what is the benefit of registering/triggering an event?
All the event-related code could be removed and the feature would still be working but with less code.
If the idea is to keep the logic for submitting in a separate method, it could be done checking the response from the showDialog and calling _submitCOEDeny_ after line 160.

sources/packages/web/src/components/institutions/confirmation-of-enrollment/modals/ConfirmCOEDenyModal.vue
```diff
@@ -0,0 +1,92 @@
...
+  </ModalDialogBase>
+</template>
+
+<script lang="ts">
+import { ref } from "vue";
+import ModalDialogBase from "@/components/generic/ModalDialogBase.vue";
+import { useModalDialog, useFormioUtils } from "@/composables";
+import formio from "@/components/generic/formio.vue";
+import { ConfirmationOfEnrollmentService } from "@/services/ConfirmationOfEnrollmentService";
+export const COE_DENIAL_REASON_RADIO = "coeDenyReasonId";
```
andrewsignori-aot, line 42: Since this const is not being used anywhere else besides this class, it does not need to be exported.

andrewsignori-aot (commented):

sources/packages/api/src-sql/COEDeniedReason/Create-coe-denied-reason.sql
ann-aot: I was following the `pir_denied_reason` name of consistency, make sense, updating

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -207,4 +224,55 @@ export class ConfirmationOfEnrollmentController {
...
+  @Patch(
+    ":locationId/confirmation-of-enrollment/application/:applicationId/deny",
+  )
+  async denyConfirmationOfEnrollment(
+    @Param("locationId") locationId: number,
+    @Param("applicationId") applicationId: number,
+    @Body() payload: DenyConfirmationOfEnrollmentDto,
+  ): Promise<void> {
+    try {
+      await this.applicationService.setDeniedReasonForConfirmationOfEnrollment(
```
dheepak-aot, line 246: The name "COE" and "ConfirmationOfEnrollment" both used randomly. We can use consistently or at least use COE when the function/property names become very lengthy.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -31,6 +43,7 @@ export class ConfirmationOfEnrollmentController {
     private readonly applicationService: ApplicationService,
     private readonly workflow: WorkflowActionsService,
     private readonly locationService: InstitutionLocationService,
+    private readonly DeniedCOEReasonService: COEDeniedReasonService,
```
dheepak-aot, line 46: variables are advised to start with lowercase.

sources/packages/api/src/services/application/application.service.ts
dheepak-aot: Even I think it has to be left-join unless there is a specific reason to use inner-join here.

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -185,4 +186,22 @@ export class Application extends RecordDataModel {
     name: "pir_denied_other_desc",
   })
   pirDeniedOtherDesc: string;
+
+  @RelationId((application: Application) => application.coeDeniedReason)
+  coeDeniedId?: number;
+
+  @ManyToOne(() => COEDeniedReason, {
```
dheepak-aot, line 193: I haven't seen any advantage in using the relationship Object here. i.e while persisting/saving application we just assign an id to it. And in case of other again we just save the id and save coeDeniedOtherDesc as well. 
So can this reference just be 
@Column({
    name: "coe_denied_id",
    nullable: true,
  })
  coeDeniedReason: number;

we can skip the join in select as well if we just reference the column.

Correct me if I am wrong and this relation object is required for a purpose.

sources/packages/api/src/utilities/constants.ts
dheepak-aot: IMO. We can functionally/technically create multiple constant files and mange them. eg: application-constants.ts, coe-constants.ts. Just a suggestion.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
dheepak-aot: I recommend the same as we are following the same standard for almost other dropdowns as well.

sources/packages/api/src-sql/COEDeniedReason/Create-coe-denied-reason.sql
andrewsignori-aot: Yes, I got it. Thanks for doing the change.
@guru-aot and @dheepak-aot, please let's ensure that we are creating new tables with names in the plural.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
ann-aot: the radio button option accepts a list of objects with keys, `value` , `label` and `shortcuts`

sources/packages/api/src/services/application/application.service.ts
ann-aot: yes @andrewsignori-aot , the idea was to add all constants in the `utils/constants` file. right @guru-aot ?

sources/packages/api/src/utilities/constants.ts
ann-aot: the idea was to add all constants in the utils/constants file. right @guru-aot ? yes, @dheepak-aot  suggestions also sound good, if we predict we will have more constants in our application

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
andrewsignori-aot: The API can still return the OptionItemDto and the map could take care of providing to the radio button the object that it expects. We are not going to return the API DTO specific to populate a radio or dropdown. The API could return a specific key/value object and the adaptions could be done on Vue application.
Either way, it was a suggestion 😉

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: It does not seem right to me. If there is a constant that is globally used and it is generic, maybe, if there is not a specific file to add, we can add to a common/generic/altogether constants file, otherwise, why not create the constants near the folder/context that they belong to?

sources/packages/api/src/utilities/constants.ts
andrewsignori-aot: It does not seem right to me. If there is a constant that is globally used and it is generic, maybe, if there is not a specific file to add, we can add to a common/generic/altogether constants file, otherwise, why not create the constants near the folder/context that they belong to?
Following the same logic, are we going to add all the enums in the same file?
We also have database constants inside the database folder (because they belong to that context), do we need to move these constants inside the "common/generic/altogether constants file"?

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
ann-aot: to avoid extra code mapping, in the `web` , I returned the expected list of objects from the API :)

sources/packages/api/src/database/entities/application.model.ts
andrewsignori-aot: The advantage is to load an object and the related object with one command.
For instance, the below code will load the application and the related objects. It is also helpful while creating joins and loading some data from the main object and some data from related objects. Does it make sense?
```
await applicationRepository.findOne(testApplication.id, { relations: ["pirProgram", "offering", "location"] });
```

dheepak-aot: Yes, but if there is no requirement or need to load the related object, ORM relationship is not of any use. ORM Object mapping is required only when you read/edit the object data of relation object. In our case form is doing that part with the options list api and passes on the FK to persist.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
ann-aot: updated as disussed

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: > yes @andrewsignori-aot , the idea was to add all constants in the `utils/constants` file. right @guru-aot ?

As we talked we are not going to add a constant to a generic constants file just because it is a constant.
We already have some other constant files like database/constants.ts.
@ann-aot and I were talking and we can have a constant file to keep values that represent configurations that are not going to be exposed as environment variables but could be changed in the future, like the ones already in the file, as shown below. In this way, if the Ministry is looking to do a change in some hardcoded configurations they will be in the same place. We can rename the constant file to something like system-configuration-constants.ts
@guru-aot, @dheepak-aot, @ann-aot, does it make sense?

image

ann-aot: `system-configuration-constants.ts`, `configurable-constants.ts` or `application-configuration-constant.ts` make sense to me

andrewsignori-aot (approved):
Nice work 😉 Looks good 👍

sources/packages/api/src/services/application/application.service.ts
guru-aot: I agree with even for my existing story, I am creating a utilities constant where I am creating a Intitution_type for BC_Private in sources/packages/web/src/constants/utilities.ts. This is a separate file and is created for web package level.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
ann-aot: update lengthy function name with COE

sources/packages/api/src/services/application/application.service.ts
dheepak-aot: @andrewsignori-aot, I expressed same(similar) opinion in other comment. We can functionally/technically separate the constants file to make it more modular instead of having in same file. 
 
e.g
http-status-code.ts (technical)
application-constants.ts(functional)

sources/packages/api/src/database/entities/application.model.ts
ann-aot: another advantage, I felt was while saving, we can save the object directly like below
image

dheepak-aot: @ann-aot but here we are not leveraging the benefit of using relational Object.

guru-aot (approved):
LGTM

sources/packages/api/src/database/entities/application.model.ts
dheepak-aot: This is not a blocker, but we can definitely talk about it outside the scope of the PR about using object only when it is required.

dheepak-aot (approved):
LGTM. Good work overall.

abschwenker (approved):

PR #719, opened 2021-09-10
[api][SIMS-#518] Open Award Eligibility in Camunda (CSG- FT)  -> Add variables for camunda

ann-aot:
- [x] Add variables to below

(from Offering) Offering_Intensity (PT/FT)
(from program) credential_type (degree, etc)
(from program) program_length (12WeeksToLessThan1Year, etc)
(from institution_location) instiuttion_location_type (Bc Private, etc)
(from institution_location) iinstitution_location_Province (BC,AB, etc)
(from student) student PD status (Y/N)
- [x] updated InstitutionType as an relation Object in Application
- [x] refactored the generic external API endpoint


sources/packages/api/src/database/entities/institution.model.ts
```diff
@@ -78,9 +87,13 @@ export class Institution extends RecordDataModel {
   })
   users: InstitutionUser[];
 
-  @Column({
+  @RelationId((institution: Institution) => institution.institutionType)
+  institutionTypeId: number;
+
+  @ManyToOne(() => InstitutionType, { eager: false, cascade: true })
```
andrewsignori-aot, line 93: For this particular case, I believe that we should set the cascade to false.
The Institution Types are a fixed list that will not be changed normally and, if they changed, should not be in this way.

abschwenker (changes requested):
As discussed, lets change to 1 call for get application with required details. This way camunda is consistent and we can re-use

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -55,6 +54,38 @@ export class ApplicationSystemController {
...
+        mandatoryFees: application.offering.mandatoryFees,
+        exceptionalExpenses: application.offering.exceptionalExpenses,
+        tuitionRemittanceRequestedAmount:
+          application.offering.tuitionRemittanceRequestedAmount,
+        offeringDelivered: application.offering.offeringDelivered,
+        offeringIntensity: application.offering.offeringIntensity,
+      },
+      program: {
+        /**
+         * for now - if credential type is `other`, then the other is send to camunda
```
andrewsignori-aot, line 74: I was not able to follow the comment here. Can you elaborate more on it?

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -39,6 +72,31 @@ export interface ApplicationDataDto {
    * Program Year is added to integrate the application in camunda workflows
    */
   programYear: string;
+
+  /**
+   * Offering details fro student application
```
andrewsignori-aot, line 77: Please fix the typo "fro" on this and other comments.
Add a "." period to the sentence.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -324,16 +324,21 @@ export class ApplicationService extends RecordDataModelService<Application> {
     return application;
   }
 
+  /**
+   * get student application details with the application Id.
```
andrewsignori-aot, line 328: Please start the sentence with a upper case letter.

andrewsignori-aot (commented):
Overall looks good 👍 
Please just reply to some minor comments.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -324,16 +324,21 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
-    const application = this.repo
-      .createQueryBuilder("application")
-      .select(["application", "programYear.programYear"])
-      .innerJoin("application.programYear", "programYear")
-      .where("programYear.active = true")
-      .andWhere("application.id = :applicationId", {
-        applicationId,
-      });
-    return application.getOne();
+    return this.repo.findOne(applicationId, {
```
andrewsignori-aot, line 333: Can we be more specific in the select clause here? It seems to me that we are retrieving way more than we need.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
ann-aot: In the program, when `credential_type` is select as `other`, users have the option to enter a free text credential_type. 
image
as of now, we are not sending the free text `credential_type` in the response. In the future, if there is a required, will need to update accordingly

abschwenker (commented):
Please review tests, other than that lgtm

sources/packages/api/src/route-controllers/application/application.system.controller.ts
abschwenker: perhaps say something like:

if credential type is `other`, then the string 'other' is sent to camunda. we are not sending the free text credential type when 'other' is selected

dheepak-aot: IMO template literals(`other`) can be avoided in the comment.

andrewsignori-aot: @ann-aot, I do not see a reason to have this comment at all.
I would recommend removing it.

abschwenker (approved):

dheepak-aot (approved):
LGTM 👍

andrewsignori-aot (approved):

PR #733, opened 2021-09-16
[api][SIMS-#518] Open Award Eligibility in Camunda (CSG- FT) -> updated query and response

ann-aot:
relation data in the application is not available at every point so, added  `null` check in the response and updated query,
eg, 
image
at this point, the application won't be having offering, location info


sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -53,34 +53,36 @@ export class ApplicationSystemController {
...
+          application.location?.institution?.institutionType?.name,
       },
       location: {
-        institutionLocationProvince: application.location.data.address.province,
+        institutionLocationProvince:
+          application.location?.data?.address?.province,
       },
       student: {
-        studentPDStatus: application.student.studentPDVerified,
+        studentPDStatus: application.student?.studentPDVerified,
```
andrewsignori-aot, line 85: Student will never be nullable.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -53,34 +53,36 @@ export class ApplicationSystemController {
 
     return {
       data: application.data,
-      programYear: application.programYear.programYear,
+      programYear: application.programYear?.programYear,
```
andrewsignori-aot, line 56: Program year will never be nullable.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -343,17 +343,17 @@ export class ApplicationService extends RecordDataModelService<Application> {
         "institutionType",
         "student",
       ])
-      .innerJoin("application.programYear", "programYear")
-      .innerJoin("application.offering", "offering")
-      .innerJoin("application.pirProgram", "pirProgram")
-      .innerJoin("application.location", "location")
+      .leftJoin("application.programYear", "programYear")
```
andrewsignori-aot, line 346: Program year should be an inner join. An application should not even start the assessment without it.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -343,17 +343,17 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .leftJoin("application.offering", "offering")
+      .leftJoin("application.pirProgram", "pirProgram")
+      .leftJoin("application.location", "location")
       .leftJoin("location.institution", "institution")
       .leftJoin("institution.institutionType", "institutionType")
-      .innerJoin("application.student", "student")
+      .leftJoin("application.student", "student")
```
andrewsignori-aot, line 352: Student should be an inner join.

andrewsignori-aot (commented):
IMO, student and program year should never be null. An application should not be created without them.

andrewsignori-aot (approved):
looks good 👍

guru-aot (approved):
LGTM

Issue #640, opened 2021-08-13
Import PD status into Student Readonly Application Details


hellolynn-dsib:
**Additional Context**
When starting an application we should display the student's permanent disability status in the read-only application information

**Acceptance Criteria**
- [x] Status' from ATBC include: Yes, No, Not Requested, and Pending

**Tasks**
![MicrosoftTeams-image (4).png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/e6c57767-02c8-4535-8ba2-3692ac64f0d5)

PR #734, opened 2021-09-17
[API][Web][SIMS #640] Import PD status into Student Readonly Application Details

dheepak-aot:
Student PD status Enum created with following values
Yes - When pd_verified is true in student table
No - When pd_verified is false 
Not Requested - When pd_date_sent and pd_date_updated is null
Pending - When pd_date_sent is NOT null pd_date_updated is null

### DIFF
--- a/sources/packages/api/src/types/pdStatus.ts
+++ b/sources/packages/api/src/types/pdStatus.ts
@@ -4,6 +4,6 @@
 export enum StudentPDStatus {
   Yes = "Yes",
   No = "No",
-  Not_Requested = "Not Requested",
+  NotRequested = "Not Requested",
   Pending = "Pending",
 }
--- a/sources/packages/api/src/types/pdStatus.ts
+++ b/sources/packages/api/src/types/pdStatus.ts
@@ -1,0 +1,9 @@
+/**
+ * Enumeration for student permanent disability status.
+ */
+export enum StudentPDStatus {
+  Yes = "Yes",
+  No = "No",
+  Not_Requested = "Not Requested",
+  Pending = "Pending",
+}
--- a/sources/packages/api/src/utilities/student-utils.ts
+++ b/sources/packages/api/src/utilities/student-utils.ts
@@ -2,17 +2,19 @@ import { StudentPDStatus } from "../types/pdStatus";
 import { Student } from "../database/entities";
 
 /**
- * Determines the permanent disablility status from Student Entity.
+ * Determines the permanent disability status from Student Entity.
  * @param student
  * @returns PDStatus enum
  */
 export const determinePDStatus = (student: Student): StudentPDStatus => {
   if (student.studentPDVerified) {
     return StudentPDStatus.Yes;
-  } else if (student.studentPDSentAt) {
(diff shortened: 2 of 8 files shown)


sources/packages/api/src/utilities/student-utils.ts
```diff
@@ -0,0 +1,18 @@
+import { StudentPDStatus } from "../types/pdStatus";
+import { Student } from "../database/entities";
+
+/**
+ * Determines the permanent disablility status from Student Entity.
```
ann-aot, line 5: Typo `disability`

ann-aot (commented):
LGTM 👍 added a minor comment

sources/packages/api/src/types/pdStatus.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Enumeration for student permanent disability status.
+ */
+export enum StudentPDStatus {
+  Yes = "Yes",
+  No = "No",
+  Not_Requested = "Not Requested",
```
andrewsignori-aot, line 7: Please rename it to "NotRequested". We are not supposed to be using "_" in this case. If the application has some other place using it, it is probably wrong.

sources/packages/api/src/utilities/student-utils.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param student
+ * @returns PDStatus enum
+ */
+export const determinePDStatus = (student: Student): StudentPDStatus => {
+  if (student.studentPDVerified) {
+    return StudentPDStatus.Yes;
+  } else if (student.studentPDSentAt) {
```
andrewsignori-aot, line 12: I would recommend removing the else here because the return will already interrupt the method execution.
In general, for big methods or even small ones, to reduce the complexity (the same validated on SonarCloud) we can use the "Return Early" pattern. It implies the code and makes it easy to ready.
More info: https://medium.com/swlh/return-early-pattern-3d18a41bba8

@ann-aot @guru-aot, any thoughts?

sources/packages/api/src/utilities/student-utils.ts
```diff
@@ -0,0 +1,18 @@
...
+    return student.studentPDUpdateAt
+      ? StudentPDStatus.No
```
andrewsignori-aot, line 14: I would prefer having the "Yes/No" status defined exclusively on "student.studentPDVerified" and, if "student.studentPDVerified" is not defined, check the others conditions.

andrewsignori-aot (commented):
Please take a look at the comments 😉

sources/packages/api/src/utilities/student-utils.ts
guru-aot: Yes I agree with you @andrewsignori-aot

guru-aot (approved):
LGTM, please work on the other review Comments. Nice work @dheepak-aot

andrewsignori-aot (approved):
Nice work, looks good 👍

abschwenker (approved):
LGTM

ann-aot (approved):

Issue #594, opened 2021-07-27
Camunda and Devops changes to support mail and Dynamic Env


michesmi:
Additional Context
- Make changes to TEST before moving to dev to not interfere with current tasks

**Camunda Changes**
- [x] Create "retry event" for "Get Application"
- [x] 3 retries if it is failing, wait 1m between retries
- [x] Notify by email of fail
- [x] Change all schedulers and workflows to reference env URL variable below
**Devops Changes**
- [x] Create SIMS-API URL secret (add to secrets-param.yml and secrets.yml)
- [x] Refer secret in bpm-deploy.yml
- [x] Add mail_config.properties in bpm-deploy.yml

**Ensure we are following our current approach (Make file and Readme files should be updated)**

PR #739, opened 2021-09-21
Feature/#594 camunda devops

guru-aot:
#594 
- [x] Create SIMS-API URL secret (add to secrets-param.yml and secrets.yml)
- [x]  Refer secret in bpm-deploy.yml
- [x] Add mail_config.properties in bpm-deploy.yml


devops/openshift/forms-flow-ai/bpm-deploy.yml
```diff
@@ -3,6 +3,25 @@ kind: Template
 metadata:
   name: forms-flow-bpm
 objects:
+  - apiVersion: v1
+    data:
+      mail-config.properties: |
+        # send mails via SMTP
+        mail.transport.protocol=smtp
+        mail.smtp.host=${SMTP_SERVER}
+        mail.smtp.port=25
```
andrewsignori-aot, line 12: Is it ok to use a non SSL protect SMTP server? Is it for DEV only?

abschwenker: this is the common config for gov services. I will confirm if we have alternatives but this is commonly used

andrewsignori-aot (approved):
👍

abschwenker (approved):

devops/openshift/forms-flow-ai/bpm-deploy.yml
guru-aot: I am not sure, but we used the same server in our previous mail config too

ann-aot (approved):

devops/openshift/forms-flow-ai/bpm-deploy.yml
```diff
@@ -107,6 +126,8 @@ objects:
                       name: ${FORMIO_SECRET}
                 - name: DEBUG
                   value: "False"
+                - name: MAIL_CONFIG
```
dheepak-aot, line 129: name can be EMAIL_CONFIG instead of MAIL because we may have to use config related to physical mail(in future).

guru-aot: As the actual file name of the configuration is like /app/mail-config.properties. Its used this way. Not sure why they used it in camunda 👍

dheepak-aot (approved):

Issue #682, opened 2021-08-31
Remove "Save Draft" While Viewing Applications.


michesmi:
- [x] Do not display the "save/ submitted' buttons while the Student Application is in any state different from "Draft" and make the Form.IO also read-only (same behavior when the Student Application is visualized with the "/readonly" parameter appended to the URL).
- [x] Do not display "save draft" while Application Submission is in Progress
- [x] API should also block the above events

PR #741, opened 2021-09-21
[Api] [Vue] [SIMS #682] Hide Save-Draft and Submit button while viewing the application

dheepak-aot:
Form.io wrapper component readOnly attribute type changed to boolean

isReadOnly variable added which decides to show the form in read-only mode or not

API check for status added for submit application

API check is not required to be added for save draft as it is already in place

### DIFF
--- a/sources/packages/api/src/route-controllers/application/application.controller.ts
+++ b/sources/packages/api/src/route-controllers/application/application.controller.ts
@@ -18,6 +18,8 @@ import {
   ProgramYearService,
   APPLICATION_DRAFT_NOT_FOUND,
   MORE_THAN_ONE_APPLICATION_DRAFT_ERROR,
+  APPLICATION_NOT_FOUND,
+  APPLICATION_NOT_VALID,
 } from "../../services";
 import { IUserToken } from "../../auth/userToken.interface";
 import BaseController from "../BaseController";
@@ -106,32 +108,28 @@ export class ApplicationController extends BaseController {
       );
     }
 
-    const applicationToSubmit =
-      await this.applicationService.getApplicationByIdAndUser(
-        applicationId,
-        userToken.userId,
-      );
-    if (!applicationToSubmit) {
-      throw new NotFoundException("Student Application not found.");
-    }
-
-    if (ApplicationStatus.draft !== applicationToSubmit.applicationStatus) {
-      throw new UnprocessableEntityException(
-        "Application in draft status can only be submitted",
-      );
-    }
-
     const student = await this.studentService.getStudentByUserId(
       userToken.userId,
     );
-
-    await this.applicationService.submitApplication(
-      applicationId,
-      student.id,
-      programYear.id,
-      submissionResult.data.data,
-      payload.associatedFiles,
-    );
+    try {
+      await this.applicationService.submitApplication(
+        applicationId,
+        student.id,
+        programYear.id,
+        submissionResult.data.data,
+        payload.associatedFiles,
+      );
+    } catch (error) {
+      if (error.name === APPLICATION_NOT_FOUND) {
+        throw new NotFoundException(error.message);
+      }
+      if (error.name === APPLICATION_NOT_VALID) {
+        throw new UnprocessableEntityException(error.message);
+      }
+      throw new InternalServerErrorException(
+        "Unexpected error while updating the draft application.",
+      );
+    }
 
     this.applicationService.startApplicationAssessment(applicationId);
   }
--- a/sources/packages/api/src/route-controllers/application/application.controller.ts
+++ b/sources/packages/api/src/route-controllers/application/application.controller.ts
@@ -115,6 +115,12 @@ export class ApplicationController extends BaseController {
       throw new NotFoundException("Student Application not found.");
     }
 
+    if (ApplicationStatus.draft !== applicationToSubmit.applicationStatus) {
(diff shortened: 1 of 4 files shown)


sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -115,6 +115,12 @@ export class ApplicationController extends BaseController {
       throw new NotFoundException("Student Application not found.");
     }
 
+    if (ApplicationStatus.draft !== applicationToSubmit.applicationStatus) {
```
andrewsignori-aot, line 118: The method below "submitApplication" is already ensuring that the application will be in the draft state and throwing an exception.
I would recommend:
1 - From the service layer, change the current exception being thown 
from
```
if (!application) {
  throw new Error(
    "Student Application not found or it is not in the correct status to be submitted.",
  );
}
```

to

```
throw new CustomNamedError(
    "Student Application not found or it is not in the correct status to be submitted.",
    APPLICATION_NOT_FOUND,
  );
```

2 - Capture the exception in the controller and throw a NotFoundException.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -7,7 +7,8 @@
             <v-btn
               color="primary"
               class="mr-5"
-              v-show="!isFirstPage && !readOnly"
+              v-if="!isReadOnly"
+              v-show="!isFirstPage"
```
andrewsignori-aot, line 11: I believe that there is one AC missing. If the _submittingApplication_ is true, the draft button should not be visible, as per below AC
```
Do not display "save draft" while Application Submission is in Progress
```

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -111,6 +114,7 @@ export default {
     const isFirstPage = ref(true);
     const isLastPage = ref(false);
     let applicationWizard: any;
+    const isReadOnly = ref(false);
```
andrewsignori-aot, line 117: I would recommend starting with true. It seems to be safer in the sense that the user will not be able to interact with the form till the data is loaded.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -120,6 +124,9 @@ export default {
       ]);
       // TODO: Move formatted address to a common place in Vue app or API.
       // Adjust the spaces when optional fields are not present.
+      isReadOnly.value =
+        applicationData.applicationStatus !== ApplicationStatus.draft ||
+        props.readOnly;
```
andrewsignori-aot, line 129: I believe that the below expression must have the second part converted to a boolean (using double exclamations), as below.
```
isReadOnly.value =
        applicationData.applicationStatus !== ApplicationStatus.draft ||
        !!props.readOnly;
```

andrewsignori-aot (commented):
Nice work with the Form.IO read-only.
Please take a look at the comments 😉

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
dheepak-aot: I do not see a benefit of converting an object to boolean here as we take the advantage of objects being used in expression. The logic works perfectly with given verbose.

andrewsignori-aot: What type of object do you expect as a result in "isReadOnly.value"?

andrewsignori-aot: - This PR changed the Form.IO component readOnly property from string to boolean (I am assuming that was because it reflects better what it really is).
- isReadOnly.value if initialized as boolean, because I assume that it is better expressed as a boolean
- The mentioned expression could return a boolean or a string, so I was just assuming that it would be more accurate to keep the isReadOnly.value always populated with a boolean instead of sometimes as a string, as shown below.

image

image

ann-aot (commented):
LGTM 👍 pls address the other comments

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
dheepak-aot: I got it now. It is getting tricky with the scenario. Will modify accordingly

guru-aot (approved):
LGTM

abschwenker (approved):
LGTM, once comments addressed please merge

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: The error handling part is done as discussed.

andrewsignori-aot (approved):
Thanks for the adjustments. Great work in general 👍

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
dheepak-aot: Flag added to hide "Save Draft" while submission in progress

ann-aot (approved):

Issue #527, opened 2021-06-29
CRA Income Validation


michesmi:
**Acceptance Criteria**

Tasks
- [ ] Create DB migration for sin_validated (Yes,No,Fail) -table is ok
- [ ] Create timed Task to manage jobs to call SIMSApi to generate file
- [ ] Create service for reading response file
-  [ ] Update sin_validated (Yes,No,Fail)

PR #746, opened 2021-09-24
[api][SIMS #527]CRA Income Validation

andrewsignori-aot:
- Partial delivery of ticket #527 
- The main database migrations to retrieve which income verification must be executed and to populated the response are part of an upcoming PR;
- Created the structure to generate the files to request income verifications;
- Created the structure to read the response files with income verifications;
- Created the endpoint in the API to process the income verifications (similar to what is done for SIN validations);
- Added start/end date to the program year table in preparation for upcoming PR. The program year start date will be used as a reference year to define which tax year the income verification will happen;
- Some files in this PR are not new, they were only renamed for better understanding;
- SIN validations and income verifications are generated using 2 different files with the same structure but different codes. The response files for both look the same;
- Code smells are related to TODO for upcoming changes in the next PR.


ann-aot (commented):
LGTM 👍

sources/packages/api/src-sql/ProgramYear/Add-col-start-end-dates copy.sql
```diff
@@ -0,0 +1,13 @@
...
+  COLUMN IF NOT EXISTS start_date DATE NOT NULL DEFAULT NOW();
+
+ALTER TABLE
+  sims.program_years
+ADD
+  COLUMN IF NOT EXISTS end_date DATE NOT NULL DEFAULT NOW();
+
+COMMENT ON COLUMN sims.program_years.start_date IS 'Inclusive start date for the program year.';
+
+COMMENT ON COLUMN sims.program_years.end_date IS 'Inclusive end date for the program year.';
```
dheepak-aot, line 13: why is the file name suffixed with " copy"?

andrewsignori-aot: It is wrongly named. I will fix it 👍

andrewsignori-aot: Done!

ann-aot (approved):
👍

sources/packages/api/src/cra-integration/cra-files/cra-response-record-identification.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Common identification info for a CRA response file record.
+ * Please note that the numbers below (e.g. line.substr(0, 4)) represents
+ * the position of the information in a fixed text file format.
+ * The documentation about it is available on the document
+ * 'Income Verification Data Exchange Technical Guide BC'.
+ */
+export class CRAResponseRecordIdentification {
+  constructor(line: string, lineNumber: number) {
+    this.transactionCode = line.substr(0, 4) as TransactionCodes;
```
dheepak-aot, line 15: IMO, instead of doing the substring logic inside the constructor we can have a simple constructor and and return transactionCode sin and transactionSubCode as getter methods. 

like this 

constructor(line: string, lineNumber: number) {
    this.line = line;
    this.lineNumber = lineNumber;
}

private readonly transactionCode: TransactionCodes;
  /**
   * SIN value that is part of the common record identification.
   */
  private readonly sin: string;
  /**
   * Secondary codes used alongside the records.
   */
  private readonly transactionSubCode: TransactionSubCodes;
  /**
   * Original line read from the CRA response file.
   */
  protected readonly line: string;
  /**
   * Original line number where this record was present
   * in the CRA response file. Useful for log.
   */
  protected readonly lineNumber: number;

public getSIN(){
 return this.line.substr(4, 9);
}

public getTransactionCode(){
 return this.line.substr(0, 4) as TransactionCodes;
}

andrewsignori-aot: We have both ways in different files. Maybe we can do it all in one way or another.
I would prefer to keep this logic in the constructor because moving it to the get method will make the _substr_ method be executed every time that the property is accessed, does it make sense?
The only benefit, IMO, to defer the execution of the substring to the get method would be if the logic in the constructor is heavy and must be deferred or if the property would not likely be accessed.
Please let me know your thoughts about it.

dheepak-aot: Technically what you said makes sense. But practically would we call getSIN() more than once? because once it is called we assign it to const or map it to another object for transformation or else evaluate an expression.  So multiple calls to getSIN or any other getter is not practically required. 

There is no major difference between both the ways. So any of them should be fine.

dheepak-aot (approved):
LGTM. Good work @andrewsignori-aot 👍

abschwenker (approved):
LGTM nice work!

guru-aot (approved):
LGTM

Issue #722, opened 2021-09-13
Student Application Enhancements -Part 2


michesmi:
Related Ticket #683

Issue #393, opened 2021-04-26
Education Programs/ Offerings Enhancements Dev Part 2


mzmichellesmith:
Related Ticket #520

The following are updates required for Education Programs:

**Institution Location- Create New Program Page**
Text Change:
- [x] "Is the Minimum age 19 and up?" _(Y/N)_ 
- [x] If No, Change the wording of the banner-_Reveal the Banner,  "This program is ineligible for StudentAid BC funding", Students need to be 19 and up.  If you require an appeal, please continue and complete this application. StudentAidBC will review your application" Set Program Status to 'Pending'_

** Institution Location- Add Offering Page**
- [x] Hide the Feild- "No Offering Dates" (For Continuous Entry)" _(Added to the Icebox for future Solution-ing)_ Hardcode to False
- [x] Hide the Feild- - "No Fixed Costs" (Added to the Icebox for future Solution-ing)_ Hardcode to False
---No Database change required. Columns remain either nullable or not-nullable

**Student Application- Select your Program Page**
- [x] When the Student Selects the Program from the Dropdown list in "The Program I will be Attending", _Provide The details of the Program Name, Description, Credential, and Delivery Method_

PR #757, opened 2021-10-01
[api][web][#393]Education Programs/ Offerings Enhancements Dev Part 2

guru-aot:
#393 

Form changes for the below 3 files will be taken care by @dheepak-aot, as he is working on the same form already for the story #722

image

Please ignore the code smells, as its the repeat of the same code in the dto layer. I dont know to clear it, if somebody has suggestions, please share.

**Institution Location- Create New Program Page**
Text Change:
- [x] "Is the Minimum age 19 and up?" _(Y/N)_ 
- [x] If No, Change the wording of the banner-_Reveal the Banner,  "This program is ineligible for StudentAid BC funding", Students need to be 19 and up.  If you require an appeal, please continue and complete this application. StudentAidBC will review your application" Set Program Status to 'Pending'_

image

** Institution Location- Add Offering Page**
- [x] Hide the Feild- "No Offering Dates" (For Continuous Entry)" _(Added to the Icebox for future Solution-ing)_ Hardcode to False
- [x] Hide the Feild- - "No Fixed Costs" (Added to the Icebox for future Solution-ing)_ Hardcode to False
---No Database change required. Columns remain either nullable or not-nullable

image

**Student Application- Select your Program Page**
- [x] When the Student Selects the Program from the Dropdown list in "The Program I will be Attending", _Provide The details of the Program Name, Description, Credential, and Delivery Method_

image


sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -146,6 +149,27 @@ export class StudentController extends BaseController {
     );
   }
 
+  /**
+   * This returns only a part of the EducationProgram details to fetch
+   * for the of Student
+   * @param programId
+   * @returns
```
ann-aot, line 156: you missed this comment

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -146,6 +149,27 @@ export class StudentController extends BaseController {
...
+   */
+  @Get("/education-program/:programId/summary")
+  async get(
```
ann-aot, line 159: can you rename the function

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -40,6 +41,46 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+        "programs.name",
+        "programs.description",
+        "programs.credentialType",
+        "programs.credentialTypeOther",
+        "programs.deliveredOnSite",
+        "programs.deliveredOnline",
+      ])
+      .where("programs.id = :programId", { programId })
+      .getOne();
+
```
ann-aot, line 67: IMO, the query should be an individual function that returns the repo as promise, and the DTO response can be created inside the controller or as a sep function, then only we could reuse the query right? if the service is specifically for a DTO, then we cant reuse it right? thoughts @guru-aot  @andrewsignori-aot  @dheepak-aot

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -67,6 +67,30 @@ export function useFormioDropdownLoader() {
...
+    programId: number,
+    fieldId: string,
+  ) => {
+    const valueToBeLoaded = await EducationProgramService.shared.getStudentEducationProgram(
+      programId,
+    );
+    formioUtils.setComponentValue(
+      form,
+      fieldId,
+      `<h3><span style="background-color:transparent;color:#2965c5;"><strong>${valueToBeLoaded.name}</strong></span></h3>${valueToBeLoaded.description}
```
ann-aot, line 82: instead of inline style, you can create a class and use

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -67,6 +67,30 @@ export function useFormioDropdownLoader() {
...
+      <table style="width:100%">
```
ann-aot, line 83: why `<table>` ?

guru-aot: Without table, it looks weird and we had to use &nbsp for the spaces, this is to be in sync with the wireframe. If you have suggestions, please tell.

sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -52,3 +52,16 @@ export interface FileCreateDto {
   size: number;
   mimetype: string;
 }
+
+export class StudentEducationProgramDto {
```
andrewsignori-aot, line 56: Is this DTO used to send data to the API?
Using a class and the validators have effect only to submit data to the API. If this is not the case, please convert it to an interface and remove the validators.

sources/packages/api/src/services/education-program/education-program.service.ts
guru-aot: I feel, we should restrict the call in the service level as the student is not authorized to see other columns like the institution. Else we would have used the getInstitutionProgram method of the same service and would have created a separate controller reusing the service.

sources/packages/api/src/route-controllers/student/student.controller.ts
guru-aot: the complete route looks like api/students/education-program/44/summary. I feel its explanatory, saying the education-program summary for a student.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
guru-aot: image

ann-aot: https://getbootstrap.com/docs/4.0/layout/grid/, you can use bootstrap classess

sources/packages/api/src/route-controllers/student/student.controller.ts
ann-aot: the is student controller, IMO, If the function name is `get` , then router should be `api/student/:id`

sources/packages/web/src/composables/useFormioDropdownLoader.ts
guru-aot: So what's the difference?

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -40,6 +41,46 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+      .select([
+        "programs.id",
+        "programs.name",
+        "programs.description",
+        "programs.credentialType",
+        "programs.credentialTypeOther",
+        "programs.deliveredOnSite",
+        "programs.deliveredOnline",
+      ])
+      .where("programs.id = :programId", { programId })
```
andrewsignori-aot, line 65: I believe that we should restrict this a bit more. Can the code do the check if at least the program is approved?
```
.where("programs.approvalStatus = :approvalStatus", {
        approvalStatus: ApprovalStatus.approved,
      })
```

guru-aot: I agree, even for the list we are showing only the approved programs

sources/packages/api/src/services/education-program/education-program.service.ts
andrewsignori-aot: As @ann-aot mentioned, I believe that we has this conversation before and we had a meeting about it. Is there a reason to not follow it?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
ann-aot: IMO, it's not appropriate to use the table for these types of styles, in future if we want to add a custom styles or override an actual table, it will be difficult and break styles

sources/packages/api/src/route-controllers/student/student.controller.ts
guru-aot: the id of the student is not used in the query, its a plain query to fetch the education program description for any student irrespective. As its for the student, I have moved it to the student controller.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
guru-aot: OKay, let me change it.

sources/packages/api/src/route-controllers/student/student.controller.ts
ann-aot: Let me rephrase, IMO, it makes sense to give a function name as `get`, when the router is `api/student/:id`, here the case is different right?.  `get` is not an appropriate name for the function, in this scenario IMO. thoughts @andrewsignori-aot  @dheepak-aot

guru-aot: Ho you meant the get :). I get it now 👍 Thanks will rename

sources/packages/api/src/services/education-program/education-program.service.ts
guru-aot: I might have missed it, I think I followed what we discussed. If I have missed something, please explain or we can have a call.

andrewsignori-aot: Please follow what @ann-aot mentioned or organize a call and we can talk about it.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
andrewsignori-aot: I do not see a reason to keep this HTML on Vue application. Can you please move this HTML to the form definition?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -67,6 +67,30 @@ export function useFormioDropdownLoader() {
     );
   };
 
+  // Get Program description for the selected program and set to the field (programDesc) in formio
+  const loadProgramDesc = async (
```
andrewsignori-aot, line 71: This class is for methods related to loading dropdowns. Loading this information does not sound like something directly related to the dropdowns loading. Yes, the data is loaded in reaction to a dropdown event, but still not loading a data to a dropdown.

andrewsignori-aot (commented):
Please connect the issue with the ticket.
About the below comment in the description.
"_Please ignore the code smells, as its the repeat of the same code in the dto layer. I dont know to clear it, if somebody has suggestions, please share._"
SonarCloud is configured to ignore code duplications on ".dto" files. This expectation was added to SonarCloud with the intention of avoiding duplication warnings between API DTOs and Vue DTOs.
In the mentioned case, the duplication is API->API. Either way, renaming the class to something.dto will make SonarCloud ignore it.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -38,7 +40,11 @@ import { Readable } from "stream";
 import { defaultFileFilter, uploadLimits } from "../../utilities/upload-utils";
 import { StudentApplicationDTO } from "../application/models/application.model";
 import { Application } from "../../database/entities";
-import { determinePDStatus } from "../../utilities/student-utils";
+import {
+  determinePDStatus,
+  deliveryMethod,
+} from "../../utilities/student-utils";
+import { credentialTypeToDisplay } from "../../utilities/credential-type-utils";
```
ann-aot, line 47: i think u missed to push this file

guru-aot: you mean credentialTypeToDisplay? its already present. i just reused the utility

ann-aot: okay 👍

abschwenker (approved):

sources/packages/web/src/composables/useFormioComponentLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useFormioUtils } from ".";
+
+/**
+ * Common methods to load components data on Form.IO that could
+ * be reusable or at least simplify the form data load logic.
+ * @returns Methods to help loading the data into Form.IO forms.
+ */
+export function useFormioComponentLoader() {
+  const formioUtils = useFormioUtils();
+  // get offering date of the selected offering and set to the hidden field (selectedOfferingDate) in formio
```
andrewsignori-aot, line 12: As agreed, please start the sentence with a capital letter and end it with a period.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -146,6 +153,32 @@ export class StudentController extends BaseController {
     );
   }
 
+  /**
+   * This returns only a part of the EducationProgram details for the student
+   * @param programId
+   * @returns
+   */
+  @Get("/education-program/:programId/summary")
```
andrewsignori-aot, line 161: Do we need to have the "_summary_" appended here? I believe that this could be reused anytime that there is a need to retrieve the program data for a student.
Following this, the property "_credentialType_" on DTO could be renamed to "_credentialTypeToDisplay_" allowing the DTO in the future to return the real "_credentialType_" from the database.

dheepak-aot (approved):
LGTM. Please address the review comments.

sources/packages/api/src/route-controllers/student/student.controller.ts
guru-aot: removed the /summary

andrewsignori-aot (approved):
Looks good 👍 
Thanks for doing the changes and adjusting the Form.IO definition to remove the custom HTML.

PR #762, opened 2021-10-05
[SIMS #722] Student Application Enhancements - Part 2

dheepak-aot:
Number of form.io components created, modified and removed as per #722 
These changes have been done on form "SFAA2022-23". 
Form definitions for SFAA2021-22 and SFAA2023-24 are created by making a copy of SFAA2022-23 and changing the years respectively.
The form definition update implicitly has the change made by @guru-aot for #393


ann-aot (commented):
instead of this well component, pls use panel component and hide label to match up with figma, 
image

abschwenker:
> instead of this well component, pls use panel component and hide label to match up with figma, image

Ageed, this change was part of @guru-aot  however for consistency if we can change it to be like the rest it would be good

andrewsignori-aot (commented):
As we talked, please add the TXT extensions in the allowed list for upload on Form.IO and API.

Issue #668, opened 2021-08-26
Workflow for Decline Confirmation of Enrolment  (COE)  and Program Information Request (PIR)


michesmi:
**Additional Context**
- When a COE or PIR is declined, the workflow in Camunda sits at the message waiting. This ticket is to stop the workflow when a COE or PIR and declined.
- Once a COE or PIR is declined the student will either edit their application and resubmit or cancel it and restart
**Tasks**
 - [x] When PIR or COE is declined, stop camunda process (cancel/delete).
 - [x] API log but don't report error to user (cancel/delete).

PR #763, opened 2021-10-05
[SIMS #668][API]Workflow for Decline Confirmation of Enrolment (COE) and Program Information Request (PIR)

guru-aot:
#668 
**Additional Context**
- When a COE or PIR is declined, the workflow in Camunda sits at the message waiting. This ticket is to stop the workflow when a COE or PIR and declined.
- Once a COE or PIR is declined the student will either edit their application and resubmit or cancel it and restart
**Tasks**
 - [x] When PIR or COE is declined, stop camunda process (cancel/delete).
 - [x] API log but don't report error to user (cancel/delete).

### DIFF
--- a/sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
+++ b/sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
@@ -47,6 +47,7 @@ export class ProgramInfoRequestController {
     private readonly institutionService: InstitutionService,
     private readonly locationService: InstitutionLocationService,
     private readonly pirDeniedReasonService: PIRDeniedReasonService,
+    private readonly workflow: WorkflowActionsService,
     private readonly formService: FormService,
   ) {}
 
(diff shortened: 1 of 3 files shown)


sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
```diff
@@ -151,6 +148,17 @@ export class ProgramInfoRequestController {
         payload.pirDenyReasonId,
         payload.otherReasonDesc,
       );
+      const result =
```
andrewsignori-aot, line 151: Did you check if the method _this.applicationService.setDeniedReasonForProgramInfoRequest_ isn't returning the _assessmentWorkflowId_ that is needed?

guru-aot: No I did check, let me check

guru-aot: Wow, we are getting it, let me change and remove the unwanted method.

andrewsignori-aot (approved):
looks good 👍

dheepak-aot (approved):
LGTM.

abschwenker (approved):
LGTM

Issue #764, opened 2021-10-05
Program Regulatory not saving


abschwenker:
When creating a program the below field is not being saved to SIMSDB, saving as null
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/c717cff0-031f-4e10-8a89-4186aa74bc5b)

PR #765, opened 2021-10-05
[SIMS#764][Form]Program Regulatory not saving 

guru-aot:
#764 
When creating a program the below field is not being saved to SIMSDB, saving as null

Added the missed hidden field isBCPrivate

The bug was happening, during the dry run submission of the saveEducationProgram call. The isBCPrivate variable which is used to manipulate regulatoryBody field in the conditional logic was not present in the form definition, so after the dry run submission, the value for regulatory body was undefined.

### DIFF
--- a/sources/packages/forms/src/form-definitions/educationprogram.json
+++ b/sources/packages/forms/src/form-definitions/educationprogram.json
@@ -1075,6 +1075,76 @@
       "defaultValue": ""
     },
     {
+      "label": "isBCPrivate",
+      "customClass": "",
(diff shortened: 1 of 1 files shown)


abschwenker (commented):
Do you know how this became an issue? Wondering why this fix was required

abschwenker (approved):

abschwenker:
> Do you know how this became an issue? Wondering why this fix was required

answered in description

dheepak-aot (approved):

Issue #790, opened 2021-10-12
Enable Refresh Token


michesmi:
Additional Context
- While a student or institution or AEST or parent/partner is logged in we must automatically log off the individual after a set idle time
- While they user is about to be logged off, show a warning message asking if they wish to extend their time

Acceptance Criteria
- [x] Create mechanism to refresh users token for "active users"
- Active user = Any "user mouse move on portal" OR "user mouse click on portal" OR "user keyboard stroke on portal"
- [x] While a student is logged in, after 4.5min of idle time, show a warning box "You are about to logged off, do you wish to extend your time" "Yes" "No"  (30second timer). After 30 second, the user is logged off.
- [x] While a parent/partner is logged in, after 4.5min of idle time, show a warning box "You are about to logged off, do you wish to extend your time" "Yes" "No"  (30second timer). After 30 second, the user is logged off.
- [x] While a institution is logged in, after 4.5min of idle time, show a warning box "You are about to logged off, do you wish to extend your time" "Yes" "No"  (30second timer). After 30 second, the user is logged off.
- [x] While a aest staff is logged in, after 4.5min of idle time, show a warning box "You are about to logged off, do you wish to extend your time" "Yes" "No"  (30second timer). After 30 second, the user is logged off.

Technical Notes
- [x] Each portal have a configurable time
- [x] Refresh token task
- [x] Active/Inactive timeout check task
- [x] Adjust timer of expiration of tokens in keycloak

PR #828, opened 2021-10-29
[web][sims-#790] Enable Refresh Token and Set Idle time

ann-aot:
- Enable refresh token - auth token life is set to 5min(set in keylock for each client) and refreshes token life is 30min (The refresh token lifetime is controlled by the SSO Session Idle Setting. 30 minutes = 30 * 60 = 1800 seconds (the refresh_expires_in value)), here we have set an interval for every 1min, if authenticated and the interval is destroyed when not authenticated, during the 1 min interval, the application will check if the life of auth token is less than 3 min(i.e MINIMUM_TOKEN_VALIDITY ), if its less than 3min, new auth and refresh token will be created.
- checks, if a portal is idle for more than 4.5 ( `MINIMUM_IDLE_TIME_FOR_WARNING_*`) (image) if the application was idle (i.e no mouse event, no click event, and no keyboard event) for more than 4.5, show a warning box "You are about to logged off, do you wish to extend your time" "Yes" "No" (30second timer). After 30 seconds, the user is logged off.

> **LOGIC**
Last activity time is saved in a variable, and a 30second timer is set, which will check if the Last activity time is greater than equal to `MINIMUM_IDLE_TIME_FOR_WARNING_*`. if greater than  `MINIMUM_IDLE_TIME_FOR_WARNING_*`, then shows a warning box with a 30-second timer, if the user clicks `yes` or anywhere else in the application, time is extended and create a new auth and refresh token if the current auth token life is less than 3 min(i.e MINIMUM_TOKEN_VALIDITY ). if user click `no` or the user still keep idle, then the user will be logged off after 30seconds
image

**_Screenshots_**
image

**PS**
disabled and enable eslint 
image
beacuse it was throwing below eslint below (circular dependency),
image,
 let me know if you have better solution.


sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
...
+    const interval = ref();
+    const extendTimeModal = ref({} as ModalDialog<boolean>);
+    const startTimer = ref(false);
+    const { getClientType } = useClientLoader();
+
+    const clientType = computed(() => {
+      return getClientType(props.clientIdType);
+    });
+
+    const minimumIdleTime = computed(() => {
```
andrewsignori-aot, line 52: Here maybe we can use a similar approach to the one that we are using to get the Keycloak client name, as shown below. In this way, we do not need the 4 constants and there is no need to check each one in a switch.

image

ann-aot: added constant because of this task,
image and I confirmed with @abschwenker then added separate constants for each portal 
image

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
...
+          return MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER;
+        case ClientIdType.AEST:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_AEST;
+        default:
+          return 0.02;
+      }
+    });
+
+    const startIdleCheckerTimer = () => {
+      if (!route.path.includes(AppRoutes.Login)) {
```
andrewsignori-aot, line 68: I believe that this check could be replaced by a check if the user is authenticated or not, that is what really matters, right?

ann-aot: mm..ya that sounds logical

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
+<template>
```
andrewsignori-aot, line 1: Can we name the component as something else related to the logoff timeout control?
I got the idea that this will be at the root of the application, but we can build other ones that will also be on the root with different purposes.

ann-aot: mm, ya sure `IdleTimeChecker`. or do you guys  have any better name suggestion

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
...
+    const { executeRenewTokenIfExpired } = useAuth();
+    const route = useRoute();
+    const lastActivityLogin = ref(new Date());
+    const interval = ref();
+    const extendTimeModal = ref({} as ModalDialog<boolean>);
+    const startTimer = ref(false);
+    const { getClientType } = useClientLoader();
+
+    const clientType = computed(() => {
+      return getClientType(props.clientIdType);
```
andrewsignori-aot, line 49: It is not clear to me why do we need this method getClientType. Is direct conversation not possible?

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
...
+        lastActivityLogin.value = new Date();
+        startTimer.value = false;
+        clearInterval(interval.value);
+        startIdleCheckerTimer();
+        executeRenewTokenIfExpired();
+      }
+    };
+
+    const checkIdle = () => {
+      const idleTimeInMintutes =
```
andrewsignori-aot, line 86: We also have the _dayjs_ available on Vue. IMO operations on Dates are more clear using the library.

sources/packages/web/src/components/common/Root.vue
ann-aot: i agree, removed

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
...
+        case ClientIdType.AEST:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_AEST;
+        default:
+          return 0.02;
+      }
+    });
+
+    const startIdleCheckerTimer = () => {
+      if (!route.path.includes(AppRoutes.Login)) {
+        /* eslint-disable */
```
andrewsignori-aot, line 69: The eslint could be removed by changing the order of the methods.
image

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,115 @@
...
+        case ClientIdType.Student:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT;
+        case ClientIdType.Institution:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION;
+        case ClientIdType.SupportingUsers:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER;
+        case ClientIdType.AEST:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_AEST;
+        default:
+          return 0.02;
```
andrewsignori-aot, line 63: This default timeout seems too short. I would leave as 4.5 also.

sources/packages/web/src/components/common/Root.vue
andrewsignori-aot: I like it 😉

sources/packages/web/src/components/common/Root.vue
andrewsignori-aot: The way that I mentioned will allow the same configuration per client in the same way that the constants provided.

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -0,0 +1,116 @@
+<template>
```
andrewsignori-aot, line 1: The "Yes" button here seems to me as the wrong or dangerous choice for the user. IMO it would be the preferred.

image

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -0,0 +1,116 @@
...
+      required: false,
+    },
+    clientIdType: {
+      type: String,
+      required: true,
+      default: "" as ClientIdType,
+    },
+  },
+  setup(props: any) {
+    const countDown = ref(COUNT_DOWN_TIMER_FOR_LOGOUT);
```
andrewsignori-aot, line 51: Not an issue at all, but countdown I actually one word. It could be all lower case 😉

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -0,0 +1,116 @@
...
+import { COUNT_DOWN_TIMER_FOR_LOGOUT } from "@/constants/system-constants";
+import { ref, onUnmounted, watch } from "vue";
+import { ClientIdType } from "@/types/contracts/ConfigContract";
+
+export default {
+  components: {
+    ModalDialogBase,
+  },
+  props: {
+    startTimer: {
```
andrewsignori-aot, line 40: Is there a reason to this be controlled from outside the modal?
My understanding is that the countdown will start when the modal is displayed and reset/stop when it is closed. Does it make sense?

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -0,0 +1,116 @@
...
+
+    const dialogClosed = () => {
+      showDialog.value = false;
+      clearInterval(interval.value);
+      logoff();
+    };
+
+    const updateTimer = () => {
+      if (countDown.value > 0) {
+        countDown.value -= 1;
```
andrewsignori-aot, line 69: It could be simplified as below
```
countDown.value--;
```

sources/packages/web/src/constants/system-constants.ts
```diff
@@ -0,0 +1,12 @@
+// * MINIMUM_TOKEN_VALIDITY in seconds, i,e 180 seconds is 3 minutes
```
andrewsignori-aot, line 1: Can we normalize everything as minutes or seconds?

sources/packages/web/src/services/http/common/HttpBaseClient.ts
```diff
@@ -13,8 +14,14 @@ export default abstract class HttpBaseClient {
       throw new Error("User is not authenticated!");
     }
   }
+  static renewTokenIfExpired() {
+    if (AuthService.shared.keycloak?.isTokenExpired(MINIMUM_TOKEN_VALIDITY)) {
+      AuthService.shared.keycloak?.updateToken(MINIMUM_TOKEN_VALIDITY);
```
andrewsignori-aot, line 19: This method returns a promise. All chain off calls must be changed to await, right? Am I missing something?

sources/packages/web/src/services/http/common/HttpBaseClient.ts
```diff
@@ -13,8 +14,14 @@ export default abstract class HttpBaseClient {
       throw new Error("User is not authenticated!");
     }
   }
+  static renewTokenIfExpired() {
+    if (AuthService.shared.keycloak?.isTokenExpired(MINIMUM_TOKEN_VALIDITY)) {
```
andrewsignori-aot, line 18: What happens if the user stays active in the application without submitting any API call for 30min and the refresh token expires? For instance, filling the Student Application but not saving or submitting it?
The refresh token will expire, right?

sources/packages/web/src/components/common/Root.vue
ann-aot: My thought like, here default is an invalid scenario, so, it doesn't make sense to wait till 4.5 mini right? thoughts?

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
ann-aot: nice catch

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,118 @@
...
+      }
+    });
+
+    const startIdleCheckerTimer = () => {
+      if (isAuthenticated) {
+        /* eslint-disable */
+        interval.value = setInterval(checkIdle, 1000);
```
dheepak-aot, line 70: It throws error when we remove eslint-disable because checkIdle has been defined after startIdleCheckerTimer.
function as const gets defined sequentially unlike function definitions.

Now the issue is, if we define checkIdle before startIdleCheckerTimer, the previous error will get resolved, but checkIdle has -> confirmExtendTimeModal

image

Hence if i move confirmExtendTimeModal ahead of checkIdle like
image

Now I need to put startIdleCheckerTimer ahead CheckIdle and looks like a circular dependency. 
We can sync up to talk about it if needed.

sources/packages/web/src/components/common/Root.vue
```diff
@@ -0,0 +1,118 @@
...
+        startTimer.value = true;
+      }
+    };
+
+    onMounted(async () => {
+      startIdleCheckerTimer();
+    });
+
+    onUnmounted(() => {
+      clearInterval(interval.value);
```
dheepak-aot, line 103: why are clearing the interval on unmount ?

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
ann-aot: updated as discussed

sources/packages/web/src/components/common/Root.vue
ann-aot: we need to destroy the interval when the component is destroyed, in case any timer is running

sources/packages/web/src/services/http/common/HttpBaseClient.ts
dheepak-aot: I too have the same question.

sources/packages/web/src/components/common/Root.vue
dheepak-aot: The scope the interval is within application(Student, Institution ...) so won't it be destroyed when we move out of the scope?

dheepak-aot:
Great work overall @ann-aot 👏 Left few review comments

sources/packages/web/src/services/http/common/HttpBaseClient.ts
ann-aot: updated as discussed

sources/packages/web/src/components/common/Root.vue
ann-aot: that is the exact implementation done here

sources/packages/web/src/components/common/Root.vue
ann-aot: yes, it's because of circular dependency.  sure, we can syncup

abschwenker (approved):
just 1 typo, other than that LGTM

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -0,0 +1,112 @@
...
+  <ModalDialogBase
+    title="Confirm Extend Time"
+    dialogType="warning"
+    :showDialog="showDialog"
+    @click="extendTime"
+  >
+    <template v-slot:content>
+      <v-container class="p-component text-dark">
+        <p>
+          You are about to logged off, do you wish to extend your time?
```
abschwenker, line 11: You are about to be logged off, do you wish to extend your time? (typo, please add "be")

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -0,0 +1,104 @@
...
+          return MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER;
+        case ClientIdType.AEST:
+          return MINIMUM_IDLE_TIME_FOR_WARNING_AEST;
+        default:
+          return 0.02;
+      }
+    });
+
+    const startIdleCheckerTimer = () => {
+      if (isAuthenticated.value) {
```
andrewsignori-aot, line 57: Whatever starting the timer, would not be good to reset the timer?
```
lastActivityLogin.value = new Date();
```

sources/packages/web/src/components/common/Root.vue
andrewsignori-aot: If it is a default and invalid scenario you can just throw an exception then 😉

sources/packages/web/src/services/AuthService.ts
```diff
@@ -88,6 +90,7 @@ export class AuthService {
       });
 
       if (this.keycloak.authenticated) {
+        this.interval = setInterval(this.renewTokenIfExpired, 60000);
```
andrewsignori-aot, line 93: Please create a constant.

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/web/src/components/common/IdleTimeChecker.vue
ann-aot: updated

sources/packages/web/src/services/AuthService.ts
ann-aot: done

sources/packages/web/src/components/common/Root.vue
ann-aot: changed code,

sources/packages/web/src/components/common/Root.vue
dheepak-aot: Currently we couldn't find a solution for this. so we can move ahead with  /* eslint-disable */.
But trying to find solution will be more ideal.

dheepak-aot (approved):

andrewsignori-aot (approved):
Great work @ann-aot 👍 
Please just remove the class useClientLoader.ts as we talked and it is approved 😉

PR #830, opened 2021-11-01
[web][api]Fix/date issues students supporting users

andrewsignori-aot:
- Created a Typeorm transformer to allow us to map a Date column on DB to a Date object on the typescript. The time will be always defined as a local time zero. Typeorm would map a date-only DB column to a string, as per the references below. 
Some references:
- https://wanago.io/2021/03/15/postgresql-typeorm-date-time/
- https://stackoverflow.com/questions/62696628/how-can-i-create-columns-with-type-date-and-type-datetime-in-nestjs-with-typeorm
- https://github.com/typeorm/typeorm/issues/2176
image
- Birthdate on the students model was renamed to be in sync with supporting_users and database. Apparently, "birthDate" and "birthdate" are both acceptable, but since the DB column is named as birth_date it seems more logical use birthDate.
- Changed Date component on the search UI on supporting users because the PrimeVue calendar was not behaving as expected and it was causing a bad user experience.

### DIFF
--- a/sources/packages/api/src/database/transformers/date-only.transformer.ts
+++ b/sources/packages/api/src/database/transformers/date-only.transformer.ts
@@ -1,0 +1,41 @@
+import * as dayjs from "dayjs";
+import { getDateOnly } from "../../utilities";
+import { ValueTransformer } from "typeorm";
+export const TYPEORM_DATE_ONLY_FORMAT = "YYYY-MM-DD";
+
+/***
+ * Format a date in the same expected format that a date only
+ * string is received from Typeorm entity when mapping a
+ * Date only field on database.
+ */
+export function formatDateOnly(date?: Date): string | undefined {
+  if (date) {
+    return dayjs(date).format(TYPEORM_DATE_ONLY_FORMAT);
+  }
+  return undefined;
+}
+
+/***
+ * Allow that the a date only column on Postgres can
+ * be mapped to a Date object with local time always
+ * defined as zero.
+ * When mapping a Date only field from database Typeorm
+ * will use a string object. This transformer will allow
+ * the conversion to a javascript Date object to avoid the
+ * issues while converting a string with no time specified,
+ * as the one returned from the database.
+ */
+export const dateOnlyTransformer: ValueTransformer = {
+  from: (dbValue: any): any => {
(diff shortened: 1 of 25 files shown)


abschwenker (approved):
lgtm

dheepak-aot (approved):
LGTM

sources/packages/api/src/database/transformers/date-only.transformer.ts
```diff
@@ -0,0 +1,35 @@
...
+ * be mapped to a Date object with local time always
+ * defined as zero.
+ * When mapping a Date only field from database Typeorm
+ * will use a string object. This transformer will allow
+ * the conversion to a javascript Date object to avoid the
+ * issues while converting a string with no time specified,
+ * as the one returned from the database.
+ */
+export const dateOnlyTransformer: ValueTransformer = {
+  from: (dbValue: any): any => {
```
dheepak-aot, line 29: Just checking, is it required by ValueTransformer that we should use any. If else we can use the data type(string|Date).

andrewsignori-aot: I wouldn't say required, but I followed what was defined on ValueTransformer.

image

ann-aot (approved):
LGTM 👍 Nice that we found that issue

sources/packages/api/src/database/transformers/date-only.transformer.ts
andrewsignori-aot: Changed as we talked. Thanks for the recommendation, looks better now 😉

andrewsignori-aot:
> LGTM 👍 Nice that we found that issue

Thanks for the collaboration in helping testing and identify the gaps 😉

sources/packages/api/src/database/transformers/date-only.transformer.ts
dheepak-aot: Thanks for making the changes.

dheepak-aot:
LGTM again with dateOnlyTransformer. 👍

abschwenker (approved):
lgtm

guru-aot (approved):
LGTM @andrewsignori-aot

PR #837, opened 2021-11-03
Updated query

guru-aot:
Bug, which does not search with application number now works

image


andrewsignori-aot (approved):
Good catch 👍 
Looks good!

ann-aot (approved):
Nice Catch 👍 @guru-aot

dheepak-aot (approved):

Issue #70, opened 2021-01-08
Ministry User Views a Student Profile And Application Info


mzmichellesmith:
**As a** *(Ministry User )* **I want** *(View a Students Account)* **So That** *(I can look up information related to that Student)*

**Additional Context**
- Once a student account is successfully created, then it can be viewed by approved ministry users
- A ministry user will have the ability to search for student accounts using tombstone information including, last name, dob,
- A ministry user will be able to see the details of an individual student account. This will include all of the tombstone and contact information collected to create their account

**Acceptance Criteria**
- [ ] Given (A students account is created ), When (a Ministry User wants to view a specific students information), Then (they can see profile information, 
- [ ] Given (A Ministry User), When (Is View a Students Profile ), Then (The Students SIN will not be displayed)
- [ ] Given (A Ministry User), When (Is View a Students Profile ), Then (will not display on the ministry view, any other associated information (GUIDs) related to the credential that is used.
- [ ] Include PD status - Changed to not, as part of new UI design
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/ebd5e91f-b7cf-4dfb-8f2a-d9cdae9134ec)
- [ ] Given (A Ministry User), When (Is View a Students Profile ), Then (They can view a list of applications for that student below).
- [ ] Given (A Ministry User), When (View a Students Applications), Then (They can select an application and VIEW (ReadOnly) a student application

**Notes:**
- [ ] Create page to be rendered after "search for student" and select student
- [ ] Display profile tombstone information as per screenshot with PD status as per application (Not Requested, Pending, Etc)
- [ ] Display table of applications
- [ ] read-only view of a students application when select from table
- [ ] Investigate sharing API between student and ministry. Share or create a new 1 for ministry?

Issue #761, opened 2021-10-04
POC - Organize API's


michesmi:
- [ ] Determine approach to organizing API between user types

There was a conversation to segregate the API to create some route isolation between the different "portals".
api/aest/
api/students/
api/institutions/
api/supporting-users/
api/system-access/
 Implement of we discussed in the past, using the nest modules to try to register the global prefixes (https://docs.nestjs.com/recipes/router-module).
In this case the designation agreement will have a dedicated controller for each "portal" and different routes, as below

api/aest/designation-agreement
api/institutions/designation-agreement

andrewsignori-aot:
- RouterModule.register needed for POC not present in the current version of Nestjs (7);
- Needed upgrade Nestjs to version 8 (latest)
- After upgrading there we needed fix below issues
    - @Injection("Connection") removed
    - Entity models declaring enums must also declare enumName property;
    - Need adjust logger due to new changes on Nestjs version 8;
- Able to use the router module as expected and shown below
![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/9584e7d9-8b7a-452c-a222-292fe80c52b0)

**Steps used to upgrade Nestjs**
1. npm i rxjs@^7
2. npx npm-check-updates "/nestjs*/" -u 
3. npm i typescript@latest
4. npm i

PR #839, opened 2021-11-05
[Web][Api][SIMS #70] Ministry user view for Student Profile And Application

dheepak-aot:
- [x] Given (A students account is created ), When (a Ministry User wants to view a specific students information), Then (they can see profile information, 
- [x] Given (A Ministry User), When (Is View a Students Profile ), Then (The Students SIN will not be displayed)
- [x] Given (A Ministry User), When (Is View a Students Profile ), Then (will not display on the ministry view, any other associated information (GUIDs) related to the credential that is used.
- [x] Given (A Ministry User), When (Is View a Students Profile ), Then (They can view a list of applications for that student below).
- [x] Given (A Ministry User), When (View a Students Applications), Then (They can select an application and VIEW (ReadOnly) a student application

After student search by clicking "view" the student details page is displayed

image

Student details page consists of student profile info and list of application that belong to the student.
By clicking on any of these applications, application details page is displayed. 

image

Fixed the query for search by application.
image

**_Sonar code smells(2 of them) is because of TODO._**


sources/packages/api/src/services/application/application.service.ts
```diff
@@ -356,10 +356,19 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+  /**
+   * Fetches application by applicationId and userId|studentId.
+   * When studentId is supplied, studentId is preferred as search criteria.
+   * When studentId is not supplied, then userId is used.
+   * @param applicationId
+   * @param userId
+   * @param studentId
+   * @returns
+   */
   async getApplicationByIdAndUser(
```
ann-aot, line 368: Not a part of this PR, I found that this service is also used to retrieve student applications, 
image,
 and I didn't find any logic or restriction not to show `overwritten` application on the API side, the frontend has the logic to show the form in read-only mode ( when a student tries to access the application by editing the application id from URL). We need to address this right @devs?

sources/packages/web/src/assets/css/base.scss
```diff
@@ -57,3 +57,7 @@ html * {
 .p-md-4 {
   padding: 0.5rem !important;
 }
+
+.padding-bottom-30 {
+  padding-bottom: 30px;
```
ann-aot, line 62: i guess, you can use bootstrap classes instead, https://getbootstrap.com/docs/4.0/utilities/spacing/

sources/packages/web/src/views/aest/ApplicationDetails.vue
```diff
@@ -0,0 +1,122 @@
...
+      }}
+    </h2>
+    <formio
+      :formName="selectedForm"
+      :data="initialData"
+      :readOnly="true"
+      @loaded="formLoaded"
+    ></formio>
+    <v-row>
+      <v-col md="6">
```
ann-aot, line 23: I feel this is a kind of duplication with 
image, maybe we can make the form as a separate component and call here in read-only mode? thoughts @devs? just a thought

sources/packages/web/src/views/aest/ApplicationDetails.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <h5 class="text-muted">
+    <a @click="goBack()">
+      <v-icon left> mdi-arrow-left </v-icon> Back to Student Details</a
+    >
+  </h5>
+  <full-page-container v-if="applicationDetail.data">
```
ann-aot, line 7: I thought ` <full-page-container` is a root container

sources/packages/web/src/views/aest/StudentDetails.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+  <h5 class="text-muted">
+    <a @click="goBack()">
+      <v-icon left> mdi-arrow-left </v-icon> Back to Student Search</a
+    >
+  </h5>
+  <PDStatus :pdStatus="studentDetail.pdStatus" />
+  <full-page-container>
```
ann-aot, line 8: same here, I thought ` <full-page-container` is a root container

sources/packages/web/src/views/student/ApplicationStatus.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <Chip
```
ann-aot, line 2: vue 3 `v-badge` was released, did u try that, https://next.vuetifyjs.com/en/components/badges/#api

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Since this method is specific to returning the application for a user(student), I would assume that there is no need to return _overwritten_ application, as you mentioned @ann-aot.

sources/packages/web/src/views/student/StudentPDBanner.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <Message
+    severity="warn"
+    :closable="false"
+    v-if="pdStatus === StudentPDStatus.Pending.toString()"
```
ann-aot, line 5: why `toString()`? enum is already in string 
image

ann-aot (commented):
Added some comments. Good Work Dheepak

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -261,13 +261,13 @@ export class StudentService extends RecordDataModelService<Student> {
...
     appNumber: string,
   ): Promise<Student[]> {
-    const applicationExistsQuery = this.applicationRepo
-      .createQueryBuilder("application")
-      .where("application.applicationNumber Ilike :appNumber")
-      .andWhere("application.applicationStatus != :overwrittenStatus")
-      .select("1");
     const searchQuery = this.repo
       .createQueryBuilder("student")
+      .leftJoinAndSelect(
```
guru-aot, line 266: Thanks for this change

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -402,4 +384,30 @@ export class ApplicationController extends BaseController {
...
+   */
+  @AllowAuthorizedParty(AuthorizedParties.aest)
+  @Get("aest/:applicationId/:studentId")
+  async getByStudentAndApplicationId(
+    @Param("applicationId") applicationId: number,
+    @Param("studentId") studentId: number,
+  ): Promise<GetApplicationDataDto> {
+    const application = await this.applicationService.getApplicationByIdAndUser(
+      applicationId,
+      0,
```
andrewsignori-aot, line 403: It seems that the method could be called with or without the user id, which makes the userId an optional parameter.
Can you please change the method to _userId_ be an optional parameter?

sources/packages/api/src/services/student/student.service.ts
ann-aot, line 266: do we need all Application objects here, Maybe you can retrieve the needed columns

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -402,4 +384,30 @@ export class ApplicationController extends BaseController {
...
+
+  /**
+   * API to fetch application details by applicationId and studentId.
+   * This API will be used by ministry users.
+   * @param applicationId
+   * @param userId
+   * @returns Application details
+   */
+  @AllowAuthorizedParty(AuthorizedParties.aest)
+  @Get("aest/:applicationId/:studentId")
```
andrewsignori-aot, line 396: The URI here would be _api/application/aest/123/456/987654_, it does not look like the ones that we have been creating.
I believe that AEST/Student/Institution would be always at the most starting as possible.
I would assume that the API for this case would be something like
```
api/aest/student/123/application/456
```
@ann-aot @guru-aot thoughts?

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -402,4 +384,30 @@ export class ApplicationController extends BaseController {
...
   }
+
+  /**
+   * API to fetch application details by applicationId and studentId.
+   * This API will be used by ministry users.
+   * @param applicationId
+   * @param userId
+   * @returns Application details
+   */
+  @AllowAuthorizedParty(AuthorizedParties.aest)
```
andrewsignori-aot, line 395: Please add the group validation also for AEST.
```
@Groups(UserGroups.AESTUser)
```

sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -62,3 +64,29 @@ export interface StudentRestrictionDTO {
...
+export interface StudentDetailDTO {
+  firstName: string;
+  lastName: string;
+  email: string;
+  gender: string;
+  dateOfBirth: Date;
+  contact: StudentContact;
+  pdVerified: boolean;
+  pdStatus: StudentPDStatus;
+  applications: StudentApplicationSummary[];
```
andrewsignori-aot, line 79: I would recommend creating two separate API calls for this. They belong to two separated contexts, that, right now, are being displayed on the same UI but, from an API perspective, it does not matter much. Does it make sense?
Assume that we would be adding new features to the list, like sorting, pagination, etc, that will need new API calls. Why we would be retrieving Student+Applications all the time?

sources/packages/web/src/assets/css/base.scss
dheepak-aot: You are right. I wanted to use the bootstrap classes but used the quick one to check how it looks and it got left.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -501,4 +503,54 @@ export class StudentController extends BaseController {
...
     } as StudentRestrictionDTO;
   }
+
+  /**
+   * API to fetch student details by studentId.
+   * This API will be used by ministry users.
+   * @param studentId
+   * @returns Student Details
+   */
+  @AllowAuthorizedParty(AuthorizedParties.aest)
```
andrewsignori-aot, line 513: Please add the group validation also for AEST.
```
@Groups(UserGroups.AESTUser)
```

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -501,4 +503,54 @@ export class StudentController extends BaseController {
...
+  @Get("/aest/:studentId")
```
andrewsignori-aot, line 514: The URI here would be api/students/aest/123, it does not look like the ones that we have been creating.
I believe that AEST/Student/Institution would be always at the most starting as possible.
I would assume that the API for this case would be something like
```
api/aest/student/123
```
@ann-aot @guru-aot thoughts?

andrewsignori-aot (commented):
Please take a look at the comments. Still working on it.

sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -62,3 +64,29 @@ export interface StudentRestrictionDTO {
...
 }
+/**
+ * DTO object for student details.
+ */
+export interface StudentDetailDTO {
+  firstName: string;
+  lastName: string;
+  email: string;
+  gender: string;
+  dateOfBirth: Date;
```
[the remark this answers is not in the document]
andrewsignori-aot: @ann-aot, it is supposed to work either, unless we figure out something that tell us otherwise 😄 
@dheepak-aot, please ensure that the value being displayed reflects the DB date.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -501,4 +503,54 @@ export class StudentController extends BaseController {
...
+  async getStudentDetails(
+    @Param("studentId") studentId: number,
+  ): Promise<StudentDetailDTO> {
+    const student = await this.studentService.findById(studentId);
+    const address = student.contactInfo.addresses[0];
+    const applications =
+      await this.applicationService.getAllStudentApplications(studentId);
+    const studentApplications = !applications
+      ? ([] as StudentApplicationSummary[])
```
andrewsignori-aot, line 523: I believe that we could avoid the null check here and the ternary expression to make the code cleaner.
In the case there are no records, an empty array will be returned and the map will also be returning an empty array, right?

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -261,13 +261,13 @@ export class StudentService extends RecordDataModelService<Student> {
...
+        Application,
+        "application",
+        "application.student_id = student.id",
```
andrewsignori-aot, line 269: Please do not use database columns names on the queries.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -261,13 +261,13 @@ export class StudentService extends RecordDataModelService<Student> {
...
   ): Promise<Student[]> {
-    const applicationExistsQuery = this.applicationRepo
-      .createQueryBuilder("application")
-      .where("application.applicationNumber Ilike :appNumber")
-      .andWhere("application.applicationStatus != :overwrittenStatus")
-      .select("1");
     const searchQuery = this.repo
       .createQueryBuilder("student")
+      .leftJoinAndSelect(
+        Application,
```
andrewsignori-aot, line 267: In this query creating the proper join between students and applications?
What happens when we are searching only by the student names without providing an application number?
It seems that all the applications of the student will be part of the query, doesn't it?

sources/packages/api/src/route-controllers/student/models/student.dto.ts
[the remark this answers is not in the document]
dheepak-aot: Yeah I tested for this specific thing. It looks good.

sources/packages/web/src/services/AestService.ts
```diff
@@ -0,0 +1,53 @@
+import ApiClient from "@/services/http/ApiClient";
+import {
+  StudentDetail,
+  SearchStudentResp,
+  GetApplicationDataDto,
+} from "@/types";
+/**
+ * Client Service layer for Ministry APIs
+ */
+export class AestService {
```
andrewsignori-aot, line 10: Is this service supposed to hold all API calls related to AEST portal?

sources/packages/web/src/views/student/ApplicationStatus.vue
andrewsignori-aot: Just keep in mind the version of the Vuetify that we are using @ann-aot.
If we need to upgrade the version to use the new component we need to check if the basic functionalities are working and looking similar.
IMO, if it is available in the version that we are using, go for it, if we need to upgrade the Vuetify version, it is up to the developer.

sources/packages/web/src/views/aest/ApplicationDetails.vue
dheepak-aot: I felt it would involve some significant work to make this component as a common(FullTimeVue) and also considering the code refactoring that is done with this story like ApplicationStatus, StudentPD banner. I thought this work can be done as a technical debt to make it one component in Student, Ministry and wherever required.

sources/packages/web/src/views/aest/StudentDetails.vue
```diff
@@ -0,0 +1,116 @@
...
+    <h2 class="color-blue padding-bottom-30">Student Details</h2>
```
andrewsignori-aot, line 9: I believe that we can start using the same classes that were initially created to form.io for header and all.
From UI/UX perspective, form.io and Vue are the same, so sharing the same styles would be the way to go 😉 
@ann-aot @guru-aot @dheepak-aot, does it make sense?

sources/packages/web/src/views/aest/StudentDetails.vue
```diff
@@ -0,0 +1,116 @@
...
+    onMounted(async () => {
+      studentDetail.value = await AestService.shared.getStudentDetail(
+        props.studentId,
+      );
+      initialData.value = {
+        firstName: studentDetail.value.firstName,
+        lastName: studentDetail.value.lastName,
+        gender: studentDetail.value.gender,
+        email: studentDetail.value.email,
+        dateOfBirth: dateString(studentDetail.value.dateOfBirth),
```
andrewsignori-aot, line 87: Please ensure that this DOB is displayed in the same way as the Student Profile.

sources/packages/web/src/views/aest/ApplicationDetails.vue
dheepak-aot: I inherited this pattern from application details screen where arrows are outside the v-container. Anyhow we do not have the wire-frame for these screens yet. we can follow the wireframes once they are out.

sources/packages/web/src/views/aest/ApplicationDetails.vue
```diff
@@ -0,0 +1,122 @@
...
+    <h2 class="color-blue padding-bottom-30">
+      Student Application Details
+      {{
+        applicationDetail.applicationNumber
+          ? " - " + applicationDetail.applicationNumber
+          : ""
+      }}
+    </h2>
+    <formio
+      :formName="selectedForm"
```
andrewsignori-aot, line 17: How the cascading dropdowns are being loaded?

andrewsignori-aot (commented):

sources/packages/api/src/route-controllers/application/application.controller.ts
ann-aot: i agree it @andrewsignori-aot

sources/packages/web/src/views/aest/ApplicationDetails.vue
andrewsignori-aot: Maybe we can do it now or later but I do not see how StudentPD or ApplicationStatus could impact this decision. Maybe I am missing something.

sources/packages/web/src/views/aest/StudentDetails.vue
andrewsignori-aot: The _full-page-container_ is the white part of the page where the content leaves.
Things like headers are not supposed to leave on it as it is right now, **but we can extend the component to have one slot for header and one for the "body".**

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -373,32 +381,45 @@ export class StudentController extends BaseController {
...
     @Res() response: Response,
   ) {
-    const student = await this.studentService.getStudentByUserId(
-      userToken.userId,
-    );
+    let studentFile = undefined;
+    if (
+      AuthorizedParties.aest === userToken.authorizedParty &&
+      userToken.groups?.some(
+        (group) => UserGroups.AESTUser.toString() === group,
```
andrewsignori-aot, line 400: I believe that we can omit the "toString" in this case, since the enum is a string enum.

sources/packages/api/src/route-controllers/application/application.controller.ts
guru-aot: @dheepak-aot Yes, having an identifier before the actual values, will help us in better understanding of the APi, @andrewsignori-aot  I agree

sources/packages/api/src/services/application/application.service.ts
dheepak-aot: Feedback comment addressed.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -402,4 +384,30 @@ export class ApplicationController extends BaseController {
...
+  @Get("aest/:applicationId/:studentId")
+  async getByStudentAndApplicationId(
+    @Param("applicationId") applicationId: number,
+    @Param("studentId") studentId: number,
+  ): Promise<GetApplicationDataDto> {
```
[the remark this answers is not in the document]
dheepak-aot: BaseDTO created as discussed.

sources/packages/web/src/services/http/ApiClient.ts
```diff
@@ -30,6 +31,7 @@ const ApiClient = {
   ProgramInfoRequest: new ProgramInfoRequestApi(),
   ConfirmationOfEnrollment: new ConfirmationOfEnrollmentApi(),
   SupportingUserApi: new SupportingUserApi(),
+  AestApi: new AestApi(),
```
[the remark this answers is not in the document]
dheepak-aot: AestApi does not exist anymore.

sources/packages/web/src/assets/css/base.scss
dheepak-aot: Feedback comment addressed.

sources/packages/web/src/views/aest/ApplicationDetails.vue
dheepak-aot: Issue #761 has been created to address the Student Application dropdown fix and organize the APIs after discussing with @abschwenker

sources/packages/web/src/views/aest/ApplicationDetails.vue
dheepak-aot: I did not create any common component for header portion.

sources/packages/web/src/views/aest/StudentDetails.vue
dheepak-aot: I did not create any common component for header portion.

sources/packages/web/src/views/student/ApplicationStatus.vue
dheepak-aot: Did not try v-badge, can try in any of the upcoming stories related.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -72,24 +71,7 @@ export class ApplicationController extends BaseController {
...
-      applicationInstitutionName: application.location?.name,
-      applicationPIRStatus: application.pirStatus,
-      applicationAssessmentStatus: application.assessmentStatus,
-      applicationCOEStatus: application.coeStatus,
-      applicationFormName: application.programYear.formName,
-      applicationProgramYearID: application.programYear.id,
-      applicationPIRDeniedReason: getPIRDeniedReason(application),
-      applicationCOEDeniedReason: getCOEDeniedReason(application),
-    };
+    return transformToApplicationDto(application);
```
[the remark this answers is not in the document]
dheepak-aot: Feedback comment addressed.

sources/packages/web/src/views/student/StudentPDBanner.vue
dheepak-aot: Feedback comment addressed.

sources/packages/api/src/services/student/student.service.ts
dheepak-aot: Feedback comment addressed.

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: The API now looks like -> api/application/aest/123/student/456, api/application/aest/student/456 and api/student/aest/123.
To push aest before application and student, new controller is required. Given that discussion is going to happen on organizing the APIs, I want to keep the APIs in existing controllers itself. The pattern of continuous numbers in API is now fixed.

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: Feedback comment addressed.

sources/packages/api/src/route-controllers/student/student.controller.ts
dheepak-aot: Feedback comment addressed.

sources/packages/api/src/route-controllers/student/student.controller.ts
dheepak-aot: Given that discussion is going to happen on organizing the APIs, I want to keep the APIs in existing controllers itself. The pattern of continuous numbers in API is now fixed.

sources/packages/api/src/route-controllers/student/student.controller.ts
dheepak-aot: Feedback comment addressed.

sources/packages/api/src/services/student/student.service.ts
dheepak-aot: Feedback comment addressed.

sources/packages/web/src/services/AestService.ts
dheepak-aot: Removed as discussed

sources/packages/api/src/route-controllers/student/student.controller.ts
dheepak-aot: Feedback comment addressed.

sources/packages/web/src/views/aest/ApplicationDetails.vue
dheepak-aot: They are not loading. #761 has been created to address the Student Application dropdown fix and organize the APIs after discussing with @abschwenker

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -356,12 +360,26 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
     applicationId: number,
-    userId: number,
+    userId?: number,
+    studentId?: number,
   ): Promise<Application> {
-    const application = await this.repo
+    if (!userId && !studentId) {
+      throw new UnprocessableEntityException(
```
andrewsignori-aot, line 378: We are usually not returning HTTP exceptions from services.
The validation is really valid but I think that a regular Error exception would be better.
Not providing at least one would be more likely a developer mistake while consuming the method.

sources/packages/forms/src/form-definitions/studentprofilesummary.json
```diff
@@ -0,0 +1,2404 @@
...
+                },
+                {
+                    "label": "Columns",
+                    "columns": [
+                        {
+                            "components": [
+                                {
+                                    "label": "City",
+                                    "tag": "p",
+                                    "className": "font-weight-bold",
```
andrewsignori-aot, line 1714: I believe that we have a style for the labels also. Probably PIR, COR, etc are using it.

sources/packages/web/src/services/ApplicationService.ts
```diff
@@ -76,4 +78,32 @@ export class ApplicationService {
...
+      applicationId,
+      studentId,
+    );
+  }
+
+  /**
+   * Get application detail of given application
+   * @param applicationId
+   * @param userId
+   * @returns GetApplicationBaseDTO
```
andrewsignori-aot, line 102: The returning object in the comment is different from the actually returned.

sources/packages/web/src/views/aest/ApplicationDetails.vue
andrewsignori-aot: Just to be clear, the ticket is not about fixing dropdowns, it is about creating a common component to be shared among the Student and the Ministry portal, that yes, will fix the dropdowns, but also fixed everything else that is needed 😉

sources/packages/web/src/views/aest/ApplicationDetails.vue
andrewsignori-aot: Just to be clear, the ticket is not about fixing dropdowns, it is about creating a common component to be shared among the Student and the Ministry portal, that yes, will fix the dropdowns, but also fixed everything else that is needed 😉

andrewsignori-aot (commented):
Thanks for implementing the changes.

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -133,4 +135,34 @@ export class ApplicationApi extends HttpBaseClient {
...
+   * @param applicationId
+   * @param userId
+   * @returns
+   */
+  public async getApplicationDetails(
+    applicationId: number,
+    studentId: number,
+  ): Promise<GetApplicationBaseDTO> {
+    const response = await this.getCall(
+      `application/aest/${applicationId}/student/${studentId}`,
```
ann-aot, line 150: shouldn't it be ` aest/application/${applicationId}/student/${studentId}`,

sources/packages/web/src/types/contracts/student/ApplicationContract.ts
```diff
@@ -70,24 +70,30 @@ export interface ApplicationStatusToBeUpdatedDto {
...
   applicationStatus: ApplicationStatus;
-  applicationStatusUpdatedOn: string;
   applicationNumber: string;
+  applicationFormName: string;
+  applicationProgramYearID: number;
+}
+
+export interface GetApplicationDataDto extends GetApplicationBaseDTO {
+  /**
+   * Application dynamic data.
```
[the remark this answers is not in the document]
dheepak-aot: Comment updated.

sources/packages/forms/src/form-definitions/studentprofilesummary.json
dheepak-aot: As discussed will update class from global styles when we have the Figma and update this screen

sources/packages/web/src/services/ApplicationService.ts
dheepak-aot: comment updated

sources/packages/web/src/services/http/ApplicationApi.ts
dheepak-aot: As discussed, moving aest to the end of URI

ann-aot (approved):
LGTM 👍  Nice work @dheepak-aot . Thanks for addressing the comments 😀

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: I noticed that a CustomNamedError is created.
A simple Error would be ok IMO. I am resolving the topic either way.

andrewsignori-aot (approved):
Thanks for doing the changes 😉 
looks good 👍

guru-aot (approved):
LGTM, nice work on the changes @dheepak-aot

abschwenker (approved):

Issue #766, opened 2021-10-06
Data Migration POC


michesmi:
**Additional Context**
In October 2021, a decision was made that the new SIMS system that is replacing SFAS and SAIL will take over all new applications, as of the date of implementation.  (currently scheduled to be some time in 2022) In order to keep the new SIMS system up to date with a pre-defined set of relevant student, application or funding changes in the SFAS system, the following methodology will be used:
1.	A single large extract will be done from SFAS, creating a data file that SIMS will use to create a starting set of SFAS data that will be used by SIMS during assessment processing.

2.	On a regular basis (hourly, daily, or whatever is decided), a file will be created with a new set of data, for any individual/application data that has changed since that last time this process was run.  The same file layout will be sent as in the first large extract.

**Draft SFAS Documentation is in the DEVs Channel Files**

**Tasks**
- [x] Create 3 Tables in SIMSDB sims schema (use SIMS naming conventions)
- Individual
- Application
- Restrictions
- [x] Individual table must accept the first name and middle name as null to support mononymous names
- [x] First name and middle name will be concatenated to created the first name on Individual table
- [x] Use mechanism to read SFTP file (on Linda.dmz) and LOAD table according to SFAS documentation
- [x] The files will be downloaded ordered by name from oldest to newest. Necessary in the case that the same record (e.g. student) is present among multiple files being imported at the same time. For this case, the most recent file will execute the last update that will be most updated one
- [x] Conduct load test on current file (50,000 individuals with associated records)
- [x] Look into using a Postgres procedure for efficiency
- [x] Tables DO NOT need physical relationships, SIMS will use individual_idx
- [x] Create hourly scheduler to read files
- [x] How to delete or active/de-active records (Schwenker to ask Harry how he is planning to do updates)
- [x] Knowledge Transfer Sesh with the Team on the Outcome of the POC
- [x] Remove the sfas_db from local development resources
- [x] Change the PD status check during student creation to get the value from the new SFAS tables

PR #842, opened 2021-11-06
[api][SIMS #766]Data Migration POC

andrewsignori-aot:
- Create tables on our database to store individuals(it is how SFAS call students), applications, and restrictions;
- Read the sample file on SFTP with 183,954 records. The entire process took ~6min on the local dev machine. This amount of records is considerably bigger than the one to process further updates later, which means that the regular sync data is expected to be processed in less than 1min or a few seconds. We still need to execute the same test once the API is deployed to DEV env.
- The files will be imported on order;
- Any error on a file or record will interrupt the processes;
- Logged information that allows the follow-up of the import process, as shown below. In case an error happens, the line number of the file that contains the error will be listed.
image
- Returns a log summary as a result with the API call that contains the exact line where the process fails, as shown below. In case an error happens the process will be interrupted and the line that contains the error will be listed.
image


sources/packages/api/src-sql/SFASIndividuals/Create-sfas-individuals.sql
```diff
@@ -0,0 +1,72 @@
+CREATE TABLE IF NOT EXISTS sims.sfas_individuals (
+  id INT PRIMARY KEY,
+  first_name VARCHAR(50) NOT NULL,
```
abschwenker, line 3: nullable, due to mononymous names

abschwenker (changes requested):

configs/env-example
```diff
@@ -54,6 +54,9 @@ API_INIT_CRA_ENVIRONMENT_CODE=
 API_INIT_ATBC_USERNAME=
 API_INIT_ATBC_PASSWORD=
 API_INIT_ATBC_APP=
+API_INIT_MSFAA_REQUEST_FOLDER=
+API_INIT_MSFAA_PROVINCE_CODE=
```
guru-aot, line 58: Thanks @andrewsignori-aot

sources/packages/api/src-sql/SFASIndividuals/Create-sfas-individuals.sql
andrewsignori-aot: Changed to accept null.
It will be null when the first name and middle name are both nulls.

abschwenker (approved):

guru-aot (approved):

ann-aot (approved):
LGTM @andrewsignori-aot Good Work 👍

dheepak-aot (approved):
LGTM. Good work @andrewsignori-aot  👍

Issue #753, opened 2021-09-28
Schedule Student Disbursement


michesmi:
Additional Context
- On the NOA we display to the student when they are receiving funds
- These funds need to be calculated after the award is determined but displayed on the NOA
- These funds calculated to either have 1 disbursement amount or 2 based off the workflow attached
- Once scheduled a job will pick up these disbursements with dates to be disbursed
CSL(Canada Student loan)
CSG (Canada Student grant)
BCSL(BC Student loan)
BCSG (BC Student Grant)

**Acceptance Criteria**
- [ ] Add sub process in Camunda  for scheduling disbursements per workflow
- [x] Create DB structure (verify naming conventions)
**REVISIONS BELOW IN COMMENTS**
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/b1e68a51-f2a7-4da6-b30d-a5d26a1bd920)

**updates**
latest workflow - [https://app.zenhub.com/files/319429847/be430ad6-e773-45be-9b11-bfc70473ce71/download](https://app.zenhub.com/files/319429847/be430ad6-e773-45be-9b11-bfc70473ce71/download)

**Tasks**
- Create DB structure per attached, (verify naming conventions)
- Add after award calculation in camunda a process for creating a schedule and applicable grants (according to workflow)

**Rules**
- [ ] If either disbursement schedule is NULL when 2 are determined, only create 1 disbursement date (all upfront)

DRAFT Disbursement Model  (Oct 6, 2021) Updates OTW
[https://app.zenhub.com/files/319429847/a947d143-c960-4700-bee9-207a292f44f0/download](https://app.zenhub.com/files/319429847/a947d143-c960-4700-bee9-207a292f44f0/download)

abschwenker:
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/bb4760b5-5b08-4831-8625-e1900204b74d)

PR #852, opened 2021-11-15
[api][SIMS #753]Schedule Student Disbursement - API/Migrations

andrewsignori-aot:
- Created the migrations for disbursements schedules and their values;
- Created an API endpoint to insert the disbursements with the payload as defined below.
image

### DIFF
--- a/sources/packages/api/src/database/entities/disbursement-value-type.ts
+++ b/sources/packages/api/src/database/entities/disbursement-value-type.ts
@@ -24,7 +24,7 @@ export enum DisbursementValueType {
    */
   BCGrant = "BC Grant",
   /**
-   * Represents the sun of the total provincial grants that
+   * Represents the sum of the total provincial grants that
    * must be reported to ESDC. The individual BC Grants are not
    * reported individually, only the total.
    */
--- a/sources/packages/api/src/database/entities/disbursement-value-type.ts
+++ b/sources/packages/api/src/database/entities/disbursement-value-type.ts
@@ -1,0 +1,32 @@
+/**
+ * Generic types of loan and grants that could
+ * be associated with a disbursement.
+ */
+export enum DisbursementValueType {
+  /**
+   * Total of Canada Loan (federal loan).
+   */
+  CanadaLoan = "Canada Loan",
+  /**
+   * Represents different types of federal grants
+   * that will be reported to ESDC (e.g. CSGP, CSGD, CSGT).
+   */
+  CanadaGrant = "Canada Grant",
+  /**
+   * Total of BC Loan (provincial loan).
+   */
+  BCLoan = "BC Loan",
+  /**
+   * Represents different types of provincial grants.
+   * These grants are not reported individually to ESDC.
+   * Only the sum of all 'BC Grants' are reported as a single
+   * value defined as 'BC Total Grant'.
+   */
+  BCGrant = "BC Grant",
+  /**
+   * Represents the sun of the total provincial grants that
+   * must be reported to ESDC. The individual BC Grants are not
+   * reported individually, only the total.
+   */
+  BCTotalGrant = "BC Total Grant",
+}
--- a/sources/packages/api/src/database/entities/disbursement-values.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-values.model.ts
@@ -1,0 +1,65 @@
+import {
+  Column,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+  RelationId,
+} from "typeorm";
+import { ColumnNames, TableNames } from "../constant";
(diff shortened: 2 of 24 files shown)


sources/packages/api/src/database/entities/disbursement-values.model.ts
```diff
@@ -0,0 +1,65 @@
...
+    type: "enum",
+    nullable: false,
+  })
+  valueType: DisbursementValueType;
+  /**
+   * Specific code for the loan/grant to be disbursed (e.g. CSLF or BCSL).
+   */
+  @Column({
+    name: "value_code",
+    nullable: false,
```
ann-aot, line 37: just a general question. do we need to create an enum for code?

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    if (application.disbursementSchedules?.length > 0) {
+      throw new CustomNamedError(
+        `Disbursements were already created for this Student Application.`,
+        APPLICATION_NOT_VALID,
+      );
+    }
+
+    for (const disbursement of disbursements) {
+      console.log(disbursement.disbursementDate);
```
ann-aot, line 86: i think missed to remove console.log

ann-aot (commented):
LGTM 👍 Minor comments added

sources/packages/api/src/database/entities/disbursement-values.model.ts
andrewsignori-aot: The **_valueType_** enum created was created with the idea to allow us to identify what and how to map the grants and loans to be sent to ESDC.

The idea to leave the **_valueCode_** as a string is to allow the workflow to dynamically insert new grants/loans without any other code change on DB or API.

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
andrewsignori-aot: Fixed 😉

ann-aot (approved):
LGTM 👍  Thanks for sharing the work @andrewsignori-aot

abschwenker (approved):
lgtm

sources/packages/api/src/database/entities/disbursement-value-type.ts
```diff
@@ -0,0 +1,32 @@
...
+  BCLoan = "BC Loan",
+  /**
+   * Represents different types of provincial grants.
+   * These grants are not reported individually to ESDC.
+   * Only the sum of all 'BC Grants' are reported as a single
+   * value defined as 'BC Total Grant'.
+   */
+  BCGrant = "BC Grant",
+  /**
+   * Represents the sun of the total provincial grants that
```
dheepak-aot, line 27: typo, sum of

andrewsignori-aot: Sun makes more sense here because it will bring the light over all the grants 😃 
Good catch, I will fix it 👍

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -0,0 +1,120 @@
...
+      .select([
+        "application.id",
+        "application.applicationStatus",
+        "disbursementSchedules.id",
+      ])
+      .leftJoin("application.disbursementSchedules", "disbursementSchedules")
+      .where("application.id = :applicationId", { applicationId })
+      .getOne();
+
+    if (!application) {
```
guru-aot, line 64: Is it required to have throw custom error status for different application status and disbursements schedule? Or can we have the checks in the query to select only the application in the right status?

guru-aot (approved):
@andrewsignori-aot Nice work, I have only one question on the query fetch, very minor. Other than that LGTM overall :)

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
andrewsignori-aot: That is a super valid question. I followed what we did in another area and I would say that this is not what we need to follow everywhere.
The reason that I decided to be more specific for this particular method is:
1. This method will be called most likely only by the workflow and it is really unlikely to be called when the application is in the wrong state, which means that the record retrieved would be almost the time right;
2. In case of something really unusual happens and this method throws that particular exception, it could be used by technical people to troubleshoot this, likely a TBTB person. It means that, if this method would be likely returning the exception message to a Student or Institution user, I would say that only a more generic error message would be better and I would include the status in the query.
Does it make sense?

dheepak-aot (approved):

Issue #76, opened 2021-01-18
Ministry User Searches for Institution Account


mzmichellesmith:
**As a** *(Ministry User)* **I want** *(search for a an Instituion)* **So That** *(review designation details)

**Additional Context**
- [ ] Viewing Institution details on the Institution account is a future User Story. Wireframe Mock-up is on the way.
- [ ] Institution Account Details may include things like, Designation, Location Details, Programs, Offerings, Timeline? Users?, Restrictions.

**Acceptance Criteria**
- [x] From the Left Navigation on the Ministry Welcome page, the Ministry User Selects " Institutions" and is navigated to the Search Institutions Page
- [x] Ministry User Can Search for an Institution using: Legal Name or Operating Name  (Return list of Legal Name, Operating Name and Address)
- [x] Ministry Users can select the Institution Account they wish to view. Once they select they are navigated to that Institutions' Account Details.  

Notes:
-single search bar that searches Legal Name or Operating Name
-table displays Legal Name or Operating Name and address

PR #853, opened 2021-11-16
[API][Web][SIMS-#76]Ministry User search institution

guru-aot:
- [x] From the Left Navigation on the Ministry Welcome page, the Ministry User Selects " Institutions" and is navigated to the Search Institutions Page
- [x] Ministry User Can Search for an Institution using: Legal Name or Operating Name  (Return list of Legal Name, Operating Name and Address)
- [x] Ministry Users can select the Institution Account they wish to view. Once they select they are navigated to that Institutions' Account Details.  

image


sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -474,4 +474,30 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+   * @param operatingName operatingName of the institution.
+   * @returns Searched institution details.
+   */
+  async searchInstitution(
+    legalName: string,
+    operatingName: string,
+  ): Promise<Institution[]> {
+    const searchQuery = this.repo
+      .createQueryBuilder("institution")
+      .where("institution.establishedDate is not null");
```
guru-aot, line 490: This where clause is added to have a default not failing condition and add additional conditions dynamically in the below lines

sources/packages/api/src/route-controllers/institution/institution.controller.ts
```diff
@@ -405,4 +409,41 @@ export class InstitutionController extends BaseController {
...
+      throw new UnprocessableEntityException(
+        "Search with at least one search criteria",
+      );
+    }
+    const searchInstitutions = await this.institutionService.searchInstitution(
+      legalName,
+      operatingName,
+    );
+    return searchInstitutions.map((eachinstitution: Institution) => ({
+      id: eachinstitution.id,
```
ann-aot, line 436: camel case `eachInstitution`

sources/packages/web/src/router/AESTRoutes.ts
```diff
@@ -89,6 +91,30 @@ export const aestRoutes: Array<RouteRecordRaw> = [
...
         },
       },
+      {
+        path: AppRoutes.SearchInstitutions,
+        name: AESTRoutesConst.SEARCH_INSTITUTIONS,
+        components: {
+          default: SearchInstitutions,
+          sidebar: AESTHomeSideBar,
+        },
+        props: true,
```
ann-aot, line 101: why prop is `true` here?

sources/packages/api/src/services/institution/institution.service.ts
andrewsignori-aot: Alternatively, you can use a dummy where clause that will not cause DB processing, like below.
```
.where("1 = 1");
```

guru-aot: Yes I thought of it, but I thought, this route can be reused in the future, to fetch all the institution. So thought of a valid scenario. I can change, what you suggest??

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <full-page-container>
+    <h2 class="color-blue">Search Institution</h2>
+    <v-row class="mt-5">
+      <v-col><label for="legalName">Legal Name</label></v-col>
+      <v-col><label for="operatingName">Operating Name</label></v-col>
+      <v-col></v-col>
```
ann-aot, line 7: why empty 2 `<v-col></v-col>`. same in line 8, 13 ?

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
...
+      </Column>
+      <Column field="address" header="Address">
+        <template #body="slotProps">
+          <span class="mt-2">
+            {{ slotProps.data.address.addressLine1 }}
+            <br />
+          </span>
+          <span v-if="slotProps.data.address.addressLine2">
+            {{ slotProps.data.address.addressLine2 }}
+            <br />
```
ann-aot, line 55: we can use bootstrap spacing classes instead of <br/>. same in line 51, 61

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
...
+      );
+      if (institutions.value.length == 0) {
+        toast.warn(
+          "No Institutions found",
+          "No Institutions found for the given search criteria.",
+        );
+      }
+    };
+
+    const institutionsFound = computed(() => {
```
ann-aot, line 115: do we need a length check? we can directly check  `institutions. value` right? let me know if I am missing anything.

ann-aot (commented):
Added some comments @guru-aot

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -474,4 +474,30 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+   * @param legalName legalName of the institution.
+   * @param operatingName operatingName of the institution.
+   * @returns Searched institution details.
+   */
+  async searchInstitution(
+    legalName: string,
+    operatingName: string,
+  ): Promise<Institution[]> {
+    const searchQuery = this.repo
+      .createQueryBuilder("institution")
```
andrewsignori-aot, line 489: Can we please add only the necessary fields on the select?

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <full-page-container>
+    <h2 class="color-blue">Search Institution</h2>
+    <v-row class="mt-5">
+      <v-col><label for="legalName">Legal Name</label></v-col>
```
andrewsignori-aot, line 5: Labels and inputs are supposed to be grouped together. Can you adjust it? The search button also seems to be lost in the screen on the right. Can you try to adjust it?

image

There an simular example on _SupportingInformation.vue_. Please take a look if makes sense to follow it.

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
...
+      <Column field="operatingName" header="Operating Name" :sortable="true">
+        <template #body="slotProps">
+          <div class="p-text-capitalize">
+            {{ slotProps.data.operatingName }}
+          </div>
+        </template>
+      </Column>
+      <Column field="address" header="Address">
+        <template #body="slotProps">
+          <span class="mt-2">
```
andrewsignori-aot, line 49: Can we create a component to display the address in the same way all around the application?
If we create a common method to format it we can reuse it even on form.io forms.

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
...
+import { AESTRoutesConst } from "@/constants/routes/RouteConstants";
+import { SearchInstitutionResp } from "@/types";
+import { useToastMessage } from "@/composables";
+import FullPageContainer from "@/components/layouts/FullPageContainer.vue";
+
+export default {
+  components: {
+    FullPageContainer,
+  },
+
```
andrewsignori-aot, line 89: Please remove the blank line.

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,129 @@
...
+        name: AESTRoutesConst.INSTITUTION_DETAILS,
+        params: { institutionId: institutionId },
+      });
+    };
+    const searchInstitutions = async () => {
+      institutions.value = await InstitutionService.shared.searchInstitutions(
+        legalName.value,
+        operatingName.value,
+      );
+      if (institutions.value.length == 0) {
```
andrewsignori-aot, line 107: Please change to ===

andrewsignori-aot (commented):
Nice work 😉 
Please take a look at the comments and I believe that UI needs improvements.

sources/packages/api/src/services/institution/institution.service.ts
dheepak-aot, line 490: Is this default condition added to handle the scenario where none of the search criteria is entered? If so IMO I would recommend API throwing error where !legalName && !operatingName

guru-aot: If search criteria is not given, the search cannot be triggered from the UI, the search button is disabled.

sources/packages/web/src/services/http/Institution.ts
```diff
@@ -142,4 +145,27 @@ export class InstitutionApi extends HttpBaseClient {
...
+  ): Promise<SearchInstitutionResp[]> {
+    try {
+      let queryString = "";
+      if (legalName) {
+        queryString += `legalName=${legalName}&`;
+      }
+      if (operatingName) {
+        queryString += `operatingName=${operatingName}&`;
+      }
+      const institution = await this.apiClient.get(
```
dheepak-aot, line 161: We can leverage the getCall() method from parent class HttpBaseClient.

sources/packages/web/src/views/aest/SearchInstitutions.vue
guru-aot: As the response of  the InstitutionService.shared.searchInstitutions is going to be an empty array, if no value is returned. the Institution.value will be always true. That is the reason we have Institution.value.length defined.

sources/packages/api/src/services/institution/institution.service.ts
dheepak-aot: But what if someone plays with API? to secure the API.

dheepak-aot:
Good work @guru-aot 👍 Please look at the comments.

sources/packages/api/src/services/institution/institution.service.ts
guru-aot: :) Yes they can until they have AEST authorization and query params by design are always optional, else we would have used path params. If both the values are not given, all the rows will be returned.

dheepak-aot: IMO. Still API must validate the same what UI has done(value to at least one of the search params). @andrewsignori-aot  @ann-aot please share your opinion here.

andrewsignori-aot: I am not following the discussion. The API has a check to validate it, the same as the UI. What am I missing here?
image

dheepak-aot: My bad. I did not check the API 😒
Sorry.

sources/packages/web/src/views/aest/SearchInstitutions.vue
guru-aot: changed to a different input and removed empty cols

sources/packages/api/src/services/institution/institution.service.ts
andrewsignori-aot: All good @dheepak-aot, these are exactly the questions that we need keep asking 😉

sources/packages/web/src/helpers/utilfunctions.ts
```diff
@@ -24,4 +26,15 @@ export default class Helper {
     const formattedDate = new Date(date);
     return formattedDate.toLocaleDateString("en-CA", options);
   }
+
+  static addressLineItemList(address: Address) {
```
andrewsignori-aot, line 30: Please move the method to the useFormatters composable.
This helper file should be killed IMO.

sources/packages/web/src/helpers/utilfunctions.ts
```diff
@@ -24,4 +26,15 @@ export default class Helper {
...
+    return [
```
andrewsignori-aot, line 31: As we talked, I believe that this method should return a simple formatted string to be directly bound to Vue template or form.io definition.

sources/packages/web/src/views/aest/SearchInstitutions.vue
```diff
@@ -0,0 +1,125 @@
...
+        <template #body="slotProps">
+          <div class="p-text-capitalize">
+            {{ slotProps.data.operatingName }}
+          </div>
+        </template>
+      </Column>
+      <Column field="address" header="Address">
+        <template #body="slotProps">
+          <div
+            v-for="addressLineItem in addressLineItemList(
```
andrewsignori-aot, line 51: As we talked, I do not think that this is the way to go. It seems that we are overcomplicating something that is supposed to be simple.

dheepak-aot (approved):

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -66,5 +67,18 @@ export function useFormatters() {
     return 0;
   };
 
-  return { dateString, dateOnlyLongString, getDatesDiff };
+  const getFormattedAddress = (address: Address): string => {
+    const formattedAddress = [];
```
andrewsignori-aot, line 71: _formattedAddress_ is declared as an Array of any. Please change it to string.
image

andrewsignori-aot (approved):
Thank you for doing the changes.
Great work normalizing the address properties and creating the formatted 👍 🥇

ann-aot (approved):
LGTM 👍 
Good work @guru-aot  😄

abschwenker (approved):

Issue #639, opened 2021-08-13
Additional Ed Program and Offering Enhancements


michesmi:
**Institution User Creates Program and Offering Study Period**
**Create New Program Page**
** Overview**

_Mockups:_ https://www.figma.com/file/VykmBDwobGoGvK9M5wgOvN/StudentAid-BC-App-Design-System---FreshWorks-Studios?node-id=5925%3A45231

- [x] Provide the following text for guidance, " Submit your program to ensure it meets the eligibility for student funding
After the program is approved, you’ll get to add study periods and costs to it"
- [x] Program Description (_Is now optional)_ _(Updates to the following: Program Details, Student App Selection View, Student App, 'Not Listed' View)
- [x] Credential Type is now revise_d to be the following list, (_Updates to the following: Program Summary Table, Offering Summary Table & Student View)_ Revise credential type to the following:
      - Undergraduate Certificate 
      - Undergraduate Diploma 
      - Undergraduate Degree 
      - Graduate Certificate 
      - Graduate Diploma 
      - Graduate Degree / Master’s
      - Post-Graduate / Doctorate
      - Qualifying Studies 
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/9c2f811f-609d-40f5-8ad9-3b1be2133432)
- [x] Update the credential type values in Camunda(award-eligibility.bpmn)
- [x] Add Institution Program Code _(Optional) (Update Program Details)_

 ** Program Eligibility** (_This is a new heading)_
- [x] Are students able to take this program on a part-time basis?  Y/ N  _(Move this content into (?) - A part-time program has a course load between 20 and 59%. A full-time program must have a course load of: 60% or greater or Between 40 and 60% for students with a permanent disability)__
- [x] New Validation on Question- "Does this program include a minimum of 20 instructional hours per week?" _(Include validation for 20 hours and 15 for aviation.) The absolute minimum number of hours of study for a full-time program in a post-secondary institution is 20 instructional hours per week and 15 instructional hours per week for aviation
institutions._
 ![image.png](https://images.zenhubusercontent.com/60c7d7040a42d68fb4abaa59/9c1878d2-108c-445d-b1ef-d766dc283ae6)

**Entrance Requirements**
- [x] "What are the entrance requirements for this program? (Select all that apply)" _(Must Select at least 1)_ If none is selected, provide the pop-up, "An Entrance requirement is required" Entrance Requirement Options are
[...]

PR #857, opened 2021-11-19
[Api][Web][SIMS #639] Program and Offering Enhancements - Part 2

dheepak-aot:
**Program Details Page** 
- [x] Show new credential types
- [x] Show Institution Program Code
- [x] Update "Offerings" to "Study Period Offerings"

Revise credential type to the following:
      - Undergraduate Certificate 
      - Undergraduate Diploma 
      - Undergraduate Degree 
      - Graduate Certificate 
      - Graduate Diploma 
      - Graduate Degree / Master’s
      - Post-Graduate / Doctorate
      - Qualifying Studies 

Following inputs removed from education program(form, DB column, dtos) averageHoursStudy, admissionRequirement
and credentialTypeOther.


sources/packages/api/src/utilities/credential-type-utils.ts
```diff
@@ -1,9 +1,18 @@
...
-    return credentialTypeOther;
-  }
-  return credentialType;
-}
+/**
+ * Util to get the label of credential type.
+ * @param credentialType
+ * @returns credentialTypeToDisplay
+ */
+export const credentialTypeToDisplay = (credentialType: string): string => {
```
andrewsignori-aot, line 6: I liked the way that the mapping is built but, can we return something in case the mapping is not present, for instance, the _credentialType_ itself. If there is some change in the form.io definition a least this method will return something close to meaningful.

andrewsignori-aot (commented):
Great work, looks good 👍 
Please take a look at the comment. Besides that, it is good to go 😉

ann-aot (approved):
Looks good to me @dheepak-aot . Good Work 👍 . Thanks for the details 😄

sources/packages/api/src/route-controllers/education-program/models/summary-education-program.dto.ts
```diff
@@ -14,9 +15,11 @@ export interface SubsetEducationProgramDto {
   name: string;
   description: string;
   credentialType: string;
+  credentialTypeToDisplay: string;
   cipCode: string;
   nocCode: string;
   sabcCode: string;
   approvalStatus: string;
   programIntensity: ProgramIntensity;
+  institutionProgramCode?: string;
```
guru-aot, line 24: Please remove the ? for the institutionProgramCode, as this DTO is used only for the GET request.

guru-aot (approved):
LGTM, just a minor comment. Nice work @dheepak-aot

sources/packages/api/src/route-controllers/education-program/models/summary-education-program.dto.ts
andrewsignori-aot: I am not following. Why DTO for a GET request is a justification to not use the nullable "?"?
If the property is nullable it could be better expressed by making it explicit. Does it make sense?

guru-aot: I am pretty sure, I got a similar comment from @andrewsignori-aot in my previous PRs, so thought of suggesting :) . Its okay, if people are following any one type of DTO creation for GET.

abschwenker (approved):

sources/packages/api/src/route-controllers/education-program/models/summary-education-program.dto.ts
andrewsignori-aot: @guru-aot, I am pretty sure that we are not on the same page. I would be really, really interested in seeing this comment from any last PR.

guru-aot: We talked, thanks for the explanation @andrewsignori-aot

andrewsignori-aot (approved):
Looks good 👍

Issue #822, opened 2021-10-22
Populate NOA with all Assessment Values


michesmi:
Tasks

- [x] Based off the disbursement schedules populate NOA will values ("$####")
- [x]  show 'Not Eligable' awards  (shown as a dash currently)

PR #903, opened 2021-12-01
[Api][Web][SIMS #822] Populate NOA with all Assessment Values

dheepak-aot:
- [x] Based off the disbursement schedules populate NOA values
- [x]  show 'Not Eligible' awards
- [x] Map all the required variables for assessment from assessment.bpmn
- [x] Update assessment in application table with all required assessment values

- Application Query for NOA enhanced with disbursement and the disbursement data is converted to json with dynamic key(s).
- Dynamic keys are to make the data compatible with Award Breakdown in form. 
- The pattern of the dynamic key is disbursement{disbursement no i.e 1|2}{value code eg: csgd} -> disbursement1csgd.
- Camunda updated with all required inputs to for assessment payload. Also the all the amounts are rounded to 2 decimal places.

### DIFF
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -242,29 +242,36 @@ export class Application extends RecordDataModel {
 }
 
 /**
- * Interface for assessment payload
+ * Interface for assessment payload.
+ *
+ * This interface is to provide contract for the assessment payload
+ * which is stored to database by workflow.
+ * It is possible that more properties can be added to the assessment payload
+ * without updating this interface and displayed in NOA form.
+ * Whenever there is a source code update, please ensure that properties in this interface are in sync with
+ * assessment payload created by camunda workflow.
  */
 export interface Assessment {
   weeks: number;
-  totalFederalAward: string;
-  totalProvincialAward: string;
-  federalAssessmentNeed: string;
-  provincialAssessmentNeed: string;
-  tuitionCost: string;
-  booksAndSuppliesCost: string;
-  exceptionalEducationCost: string;
-  livingAllowance: string;
-  transportationCost: string;
-  childcareCost: string;
-  alimonyOrChildSupport: string;
-  secondResidenceCost: string;
-  partnerStudentLoanCost: string;
-  totalAssessedCost: string;
-  studentTotalFederalContribution: string;
-  studentTotalProvincialContribution: string;
-  partnerAssessedContribution: string;
-  parentAssessedContribution: string;
-  totalFederalContribution: string;
-  totalProvincialContribution: string;
-  otherAllowableCost: string;
+  totalFederalAward: number;
+  totalProvincialAward: number;
+  federalAssessmentNeed: number;
+  provincialAssessmentNeed: number;
+  tuitionCost: number;
+  booksAndSuppliesCost: number;
+  exceptionalEducationCost: number;
+  livingAllowance: number;
+  transportationCost: number;
+  childcareCost: number;
+  alimonyOrChildSupport: number;
+  secondResidenceCost: number;
+  partnerStudentLoanCost: number;
+  totalAssessedCost: number;
+  studentTotalFederalContribution: number;
+  studentTotalProvincialContribution: number;
+  partnerAssessedContribution: number;
+  parentAssessedContribution: number;
+  totalFederalContribution: number;
+  totalProvincialContribution: number;
+  otherAllowableCost: number;
 }
--- a/sources/packages/api/src/database/entities/application.model.ts
+++ b/sources/packages/api/src/database/entities/application.model.ts
@@ -49,7 +49,7 @@ export class Application extends RecordDataModel {
     type: "jsonb",
     nullable: true,
   })
-  assessment: any;
+  assessment: Assessment;
 
   @RelationId((application: Application) => application.student)
   studentId: number;
@@ -239,4 +239,32 @@ export class Application extends RecordDataModel {
     },
   )
   disbursementSchedules?: DisbursementSchedule[];
+}
+
+/**
+ * Interface for assessment payload
+ */
+export interface Assessment {
+  weeks: number;
+  totalFederalAward: string;
+  totalProvincialAward: string;
+  federalAssessmentNeed: string;
+  provincialAssessmentNeed: string;
+  tuitionCost: string;
+  booksAndSuppliesCost: string;
+  exceptionalEducationCost: string;
+  livingAllowance: string;
+  transportationCost: string;
+  childcareCost: string;
+  alimonyOrChildSupport: string;
+  secondResidenceCost: string;
+  partnerStudentLoanCost: string;
+  totalAssessedCost: string;
+  studentTotalFederalContribution: string;
+  studentTotalProvincialContribution: string;
+  partnerAssessedContribution: string;
+  parentAssessedContribution: string;
+  totalFederalContribution: string;
+  totalProvincialContribution: string;
+  otherAllowableCost: string;
 }
--- a/sources/packages/api/src/route-controllers/application/application.controller.ts
+++ b/sources/packages/api/src/route-controllers/application/application.controller.ts
@@ -263,14 +263,13 @@ export class ApplicationController extends BaseController {
     //Disbursement data is populated with dynamic key in a defined pattern to be compatible with form table.
     const disbursementDetails = {};
     application.disbursementSchedules.forEach((schedule, index) => {
-      const disbursementKey = `disbursement${index + 1}Date`;
+      const disbursementNumber = index + 1;
+      const disbursementKey = `disbursement${disbursementNumber}Date`;
       disbursementDetails[disbursementKey] = dateString(
         schedule.disbursementDate,
       );
       schedule.disbursementValues.forEach((disbursement) => {
-        const disbursementValueKey = `disbursement${
-          index + 1
-        }${disbursement.valueCode.toLowerCase()}`;
+        const disbursementValueKey = `disbursement${disbursementNumber}${disbursement.valueCode.toLowerCase()}`;
         disbursementDetails[disbursementValueKey] = disbursement.valueAmount;
       });
     });
--- a/sources/packages/api/src/route-controllers/application/application.controller.ts
+++ b/sources/packages/api/src/route-controllers/application/application.controller.ts
@@ -263,12 +263,15 @@ export class ApplicationController extends BaseController {
     //Disbursement data is populated with dynamic key in a defined pattern to be compatible with form table.
     const disbursementDetails = {};
     application.disbursementSchedules.forEach((schedule, index) => {
-      disbursementDetails["disbursement" + (index + 1) + "Date"] =
-        schedule.disbursementDate;
+      const disbursementKey = `disbursement${index + 1}Date`;
+      disbursementDetails[disbursementKey] = dateString(
+        schedule.disbursementDate,
+      );
       schedule.disbursementValues.forEach((disbursement) => {
-        disbursementDetails[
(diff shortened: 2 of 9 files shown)


sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -240,3 +240,31 @@ export class Application extends RecordDataModel {
   )
   disbursementSchedules?: DisbursementSchedule[];
 }
+
+/**
+ * Interface for assessment payload
+ */
+export interface Assessment {
+  weeks: number;
+  totalFederalAward: string;
```
andrewsignori-aot, line 249: Can you elaborate on the reason to keep all these values as strings?
Are they saved as strings to the assessment payload on _sims.applications_ table?

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -240,3 +240,31 @@ export class Application extends RecordDataModel {
...
+  totalProvincialAward: string;
```
andrewsignori-aot, line 250: As far as we know right now, are all these variables always present or we do know already some that could be nullable?

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -260,6 +260,17 @@ export class ApplicationController extends BaseController {
         `Assessment for the application id ${applicationId} was not calculated.`,
       );
     }
+    //Disbursement data is populated with dynamic key in a defined pattern to be compatible with form table.
+    const disbursementDetails = {};
+    application.disbursementSchedules.forEach((schedule, index) => {
+      disbursementDetails["disbursement" + (index + 1) + "Date"] =
```
andrewsignori-aot, line 266: I would say that we have been using more string interpolation instead of concatenation with "+".

andrewsignori-aot: Please consider creating a variable to keep the part being used as a disbursement identifier to keep it as one single point for a change.
```
const disbursementId = `disbursement${index + 1}`
```

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -260,6 +260,17 @@ export class ApplicationController extends BaseController {
         `Assessment for the application id ${applicationId} was not calculated.`,
       );
     }
+    //Disbursement data is populated with dynamic key in a defined pattern to be compatible with form table.
+    const disbursementDetails = {};
```
andrewsignori-aot, line 264: Nice solution with the dynamic data 😉

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -503,15 +503,22 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
       .innerJoin("student.user", "user")
       .innerJoin("application.offering", "offering")
       .innerJoin("offering.educationProgram", "educationProgram")
       .innerJoin("application.location", "location")
       .innerJoin("application.msfaaNumber", "msfaaNumber")
+      .innerJoin("application.disbursementSchedules", "disbursementSchedule")
+      .innerJoin("disbursementSchedule.disbursementValues", "disbursementValue")
       .where("application.id = :applicationId", { applicationId })
       .andWhere("student.id = :studentId", { studentId })
+      .addOrderBy("disbursementSchedule.disbursementDate")
```
andrewsignori-aot, line 521: Please replace the "addOrderBy" by "orderBy". In my understanding, the "addOrderBy" is supposed to be used when you already have an "orderBy" in place.

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -193,15 +194,19 @@ export interface ActiveApplicationDataDto {
...
-  assessment: any;
+  assessment: Assessment;
   applicationNumber: string;
   fullName: string;
   programName: string;
   locationName: string;
   offeringStudyStartDate: string;
   offeringStudyEndDate: string;
   msfaaNumber: string;
+  disbursement: any;
```
ann-aot, line 209: is this `disbursement` form data?

sources/packages/forms/src/form-definitions/noticeofassessment.json
```diff
@@ -1832,7 +1832,7 @@
                         "value": ""
                       }
                     ],
-                    "content": "$#######",
+                    "content": "{{data.disbursement?.disbursement1Date ? data.disbursement.disbursement1Date : ''}}",
```
andrewsignori-aot, line 1835: The expression could be simplified as below.
```
{{data.disbursement?.disbursement1Date ?? : ''}}
````

File: sources/packages/forms/src/form-definitions/noticeofassessment.json [hunk 8]
```diff
@@ -596,7 +596,7 @@
               "value": ""
             }
           ],
-          "content": "You are eligible for <b>{{data.eligibleAmount}}</b>",
+          "content": "You are eligible for <b>${{data.eligibleAmount}}</b>",
```
ann-aot, line 599: just wanted to check, do we need `$` ?

sources/packages/forms/src/form-definitions/noticeofassessment.json
```diff
@@ -6387,129 +6403,1760 @@
...
+                    "label": "CSLFTotal",
+                    "tag": "p",
+                    "className": "",
+                    "attrs": [
+                      {
+                        "attr": "",
+                        "value": ""
+                      }
+                    ],
+                    "content": "{{(data.disbursement?.disbursement1cslf ? data.disbursement.disbursement1cslf : 0) + (data.disbursement?.disbursement2cslf ? data.disbursement.disbursement2cslf : 0)}}",
```
andrewsignori-aot, line 7155: The expression can be simplified as below
````
{{(data.disbursement?.disbursement1cslf ?? 0) + (data.disbursement?.disbursement2cslf ?? 0)}}
````

File: sources/packages/api/src/route-controllers/application/models/application.model.ts (hunk 6 above)
Comment (dheepak-aot) [reply sources/packages/api/src/route-controllers/application/models/application.model.ts:209]: Yes @ann-aot.

File: sources/packages/forms/src/form-definitions/noticeofassessment.json (hunk 8 above)
Comment (dheepak-aot) [reply sources/packages/forms/src/form-definitions/noticeofassessment.json:599]: It is a dollar sign. We have $ sign present in the form as for e.g($####) so added the $ sign.

File: sources/packages/forms/src/form-definitions/noticeofassessment.json (hunk 7 above)
Comment (ann-aot) [reply sources/packages/forms/src/form-definitions/noticeofassessment.json:1835]: I believe we can also use `{{data.disbursement?.disbursement1Date}}  ??  ' '`, both ways will work, not sure about the better approach.

File: sources/packages/forms/src/form-definitions/noticeofassessment.json (hunk 8 above)
Comment (ann-aot) [reply sources/packages/forms/src/form-definitions/noticeofassessment.json:599]: okay got it.

File: sources/packages/web/src/types/contracts/student/ApplicationContract.ts [hunk 10]
```diff
@@ -139,3 +139,46 @@ export interface ApplicationSummaryDTO {
...
+  programName: string;
+  locationName: string;
+  offeringStudyStartDate: string;
+  offeringStudyEndDate: string;
+  msfaaNumber: string;
+  disbursement: any;
+}
+
+/**
+ * DTO for assessment payload
```
andrewsignori-aot, line 159: Can we add a comment on this DTO, here and in the API, to make it clear that although the properties are explicit, this represents a dynamic jsonB database value that could be dynamically expanded by the workflow adding more values that will also be available on Form definition to be consumed?

andrewsignori-aot (commented):
Great work on dealing with the disbursements values and keeping then dynamic 👍 

ann-aot (commented):
LGTM 👍 Good Work @dheepak-aot . added some minor comments

sources/packages/api/src/database/entities/application.model.ts
guru-aot: The main reason we had the structure as 'any' for the assessment column is we were not sure about the variables that are present or not(nullable). Just wanted to confirm, if we have come to a conclusion on the variables list?

guru-aot (changes requested):
Good work, please look into the comments.

sources/packages/api/src/database/entities/application.model.ts
dheepak-aot: Type of the assessment properties changed to number and Math.round used as discussed.

sources/packages/api/src/database/entities/application.model.ts
dheepak-aot: @guru-aot Still we do not have conclusion about which one(s) can be nullable or not. We added interface to provide a signature to the data, even if more properties are added to the assessment payload, the added ones will be part of the payload in a dynamic way.

dheepak-aot: An elaborated comment has been provided to this interface as discussed.

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: Code change has been made.

sources/packages/api/src/services/application/application.service.ts
dheepak-aot: Review comment addressed.

sources/packages/forms/src/form-definitions/noticeofassessment.json
dheepak-aot: avoiding ?? operator for the benefit that it is not supported in 2 browsers.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

sources/packages/forms/src/form-definitions/noticeofassessment.json
dheepak-aot: avoiding ?? operator for the benefit that it is not supported in 2 browsers.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

sources/packages/web/src/types/contracts/student/ApplicationContract.ts
dheepak-aot: An elaborated comment has been provided to this interface as discussed.

sources/packages/api/src/route-controllers/application/application.controller.ts
andrewsignori-aot: The first change related to the concatenation was made. The second one was not changed. The changed would be as shown below.
```javascript
const disbursementDetails = {};
application.disbursementSchedules.forEach((schedule, index) => {
  const disbursementKey = `disbursement${index + 1}`;
  const disbursementDateKey = `${disbursementKey}Date`;
  disbursementDetails[disbursementDateKey] = dateString(
    schedule.disbursementDate,
  );
  schedule.disbursementValues.forEach((disbursement) => {
    const disbursementValueKey = `${disbursementKey}${disbursement.valueCode.toLowerCase()}`;
    disbursementDetails[disbursementValueKey] = disbursement.valueAmount;
  });
});
```

sources/packages/forms/src/form-definitions/noticeofassessment.json
andrewsignori-aot: @dheepak-aot, this code is from the API, it will never gonna reach or be executed by the browser, only for Nodejs. The argument does not apply here.

ann-aot (approved):
Good Work @dheepak-aot 😄

andrewsignori-aot (approved):
Thanks for doing the adjustments. The NOA looks better than ever 😉

guru-aot (approved):
LGTM

Issue #747, opened 2021-09-27
Implement Federal Restrictions File 


michesmi:
[https://app.zenhub.com/files/319429847/1d27fecf-6edb-46a3-94b8-8bfcf1772f9c/download](https://app.zenhub.com/files/319429847/1d27fecf-6edb-46a3-94b8-8bfcf1772f9c/download)

**Additional Context**
- ESDC/CSFA send us a file daily containing a list of all restrictions for all students currently in their system
- As SABC we must process this file and take 1 of 2 actions:
---Add a restriction to a student that is on the file and not in our system
---Remove a restriction on a student that is no longer in that file

Tasks
- [ ] Download and Process restircitons file (transfered via MSFT process) and add or remove restrictions based off the records

- [ ] Test feasibility of processing this file in same way to ensure it will work (it is a lot bigger) - review options if not feasible

PR #914, opened 2021-12-03
[api][SIMS #747]Implement Federal Restrictions File

andrewsignori-aot:
- Created the table sims.federal_restrictions to store the most updated snapshot of all federal restrictions;
- Created raw SQL queries to perform bulk operations to update the student restrictions (not able to create the bulk operations as needed using Typeorm builder);
- The file being imported have some inconsistencies, like invalid dates, that are being sanitized and logged;
- Restrictions codes not found on the database are being added and reported on logs;
- All operations are executed on a single transaction;
- Created a file named "DCSLS.PBC.RESTR.LIST.D2021330.999.FAKE_WITH_INVALID_DATES" to allow the load test with data similar to PROD. The file is created under DEV Teams files on folder ESDC documentation and contains:
    - More than 1 million federal restrictions records with fake SIN, Last Names and DOB;
    - 2 invalid dates to simulate the same possible issues from PROD file;
    - 10 test users with valid BCSC login (CRA Test Users from DEVs Wiki)

Please find below a sample of what the log on the POD looks like:

image

Please find below a sample of what the log on API response looks like:

image


sources/packages/api/src-sql/RawSQL/FederalRestrictions/Bulk-update-active-restrictions.sql
```diff
@@ -0,0 +1,26 @@
+/*
+ * Update all active federal restrictions on table sims.student_restrictions
+ * that are still present of federal restrictions. This update is not
+ * critical but it will keep the created date and updated date representing
+ * in a more accurate way when the system received the restriction for the first
+ * time till when the same status was kept in the federal snapshot day after day.
+ */
+UPDATE
```
dheepak-aot, line 8: Is this update really required? Asking for clarification because, on federal restriction daily file, a restriction is repeatedly sent to mention that restriction still exist and not resolved, so functionally there is no update to the restriction. Am i missing something?

andrewsignori-aot: This update is not critical but keep in mind the update is happening on the student restrictions. Update the active restriction will give the system the ability to keep track of the last time that the restriction was received in the daily report from the federal file.
Consider the below scenario:
1 - New restriction is received on day 2021-12-01 -> (CreatedDate: 2021-12-01, Update Date: 2021-12-01)
2 - Same restrictions is still active on day next day -> (CreatedDate: 2021-12-01, Update Date: 2021-12-**02**) 
3 - Same restrictions is still active on day next day -> (CreatedDate: 2021-12-01, Update Date: 2021-12-**03**) 
So, IMO, it gives a better view that the restriction was created and the system is doing a follow-up on this while it is active. Does it make sense? So far, there is no reason to believe that this will bring a performance impact but is mentioned in the comments that it is not critical, so it could be removed in the future if there is some struggle, what I do not have reasons to believe right now that would be the case.
Please le me know if we need further discussions on this 😉

sources/packages/api/src-sql/RawSQL/FederalRestrictions/Bulk-insert-new-restrictions.sql
```diff
@@ -0,0 +1,26 @@
...
+ * that are not present and active in the table. The same federal restriction
+ * can be activated and deactivated multiple times for the same student,
+ * generating a new record for every time that the restriction changes its state.
+ */
+INSERT INTO
+    sims.student_restrictions(student_id, restriction_id, is_active)
+SELECT
+    fed_restrictions.student_id student_id,
+    fed_restrictions.restriction_id restriction_id,
+    true is_active
```
ann-aot, line 12: i didn't get this part @andrewsignori-aot `true is_active`.

andrewsignori-aot: The result of the select will be inserted, so I gave aliases to the three select columns to match the insert. I though that it would be clear in this way 😉 
image

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-files/fed-restriction-file-record.ts
```diff
@@ -0,0 +1,61 @@
...
+    return `${this.restrictionCode}${this.restrictionReasonCode}`.trim();
+  }
+
+  /**
+   * Assess the record to know in advance if some invalid data is present.
+   * @returns invalid data message, if any.
+   */
+  public getInvalidDataMessage(): string | null {
+    const errors: string[] = [];
+    if (isNaN(+this.sin)) {
```
ann-aot, line 48: if the `this.sin` is empty or `"           "`, it is considered as a valid sin, right, same in the below cases. or am I missing anything? or are we considering all these values are mandatory fields
image

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
```diff
@@ -0,0 +1,254 @@
...
+   * Import all the federal restrictions from the SFTP and process as below:
+   * 1 - If the restriction is present on federal data and not on the
+   * student data, create a new active restriction;
+   * 2 - If the restriction in present and active on the student data
+   * but it is not present on federal data, deactivate it;
+   * 3 - If the restriction is present on federal data and it is also
+   * present and active on student data, update the updated_at only.
+   */
+  async process(): Promise<ProcessSFTPResponseResult> {
+    //const results: ProcessSftpResponseResult[] = [];
```
ann-aot, line 47: pls remove the commented line

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
```diff
@@ -0,0 +1,254 @@
...
+    // Get the list of all files from SFTP ordered by file name.
+    const fileSearch = new RegExp(
+      `^${this.esdcConfig.environmentCode}CSLS.PBC.RESTR.LIST.D[\w]*\.[0-9]*`,
```
ann-aot, line 50: are not not including extenstion, `.DAT`

ann-aot (commented):
LGTM 👍  added some comments

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-files/fed-restriction-file-record.ts
andrewsignori-aot: Your test is missing one part. It should be as below
```
console.log(isNaN(parseInt("   ")))
```

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
andrewsignori-aot: Each file follows its own naming conventions. 
image

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
andrewsignori-aot: Removed!

sources/packages/api/src-sql/RawSQL/FederalRestrictions/Bulk-update-active-restrictions.sql
dheepak-aot: It makes sense when it is helpful to find that a restriction which is active has to be updated within last 24 hours otherwise the nightly job isn't updated the records correctly.

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
andrewsignori-aot: @abschwenker, can you please double-check if the files are going to come with or without .DAT?
Is there a pattern that we can assume for all files under ESDC?

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+    );
+
+    let result: ProcessSFTPResponseResult;
+    if (filePaths.length > 0) {
+      // Process only the most updated file.
+      result = await this.processAllRestrictions(
+        filePaths[filePaths.length - 1],
+      );
+      // If there are more than one file, delete it.
+      // Only the most updated file matters because it represents the entire data snapshot.
```
dheepak-aot, line 64: So is it possible that we can get more than one file in a 24-hour interval?

ann-aot (approved):
LGTM 👍 Good work @andrewsignori-aot

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+        } catch (error) {
+          result.errorsSummary.push(
+            `Error while deleting federal restrictions file: ${remoteFilePath}`,
+          );
+          result.errorsSummary.push(error);
+        }
+      }
+    } else {
+      result = new ProcessSFTPResponseResult();
+      result.processSummary.push(
```
dheepak-aot, line 77: Not a part of this story, but while we populate the process summary we can categorize it by adding a process summary code kind of to identify which ones needs to be acted upon and which ones doesn't.

andrewsignori-aot: The ones that need some action are usually added to the _errrosSummary_. Is it what you are talking about?
image

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
andrewsignori-aot: Ideally not, but if there are more than one we just need to process the most updated one since a single file represents the entire snapshot.

dheepak-aot (approved):
Great work @andrewsignori-aot on importing the bulk federal restrictions file in an efficient way 👏👏

sources/packages/api/src/esdc-integration/fed-restriction-integration/fed-restriction-processing.service.ts
andrewsignori-aot: Hi @abschwenker, here @dheepak-aot was mentioning that we need something to let the Ministry(TBTB) be aware of the possible action on the errors. It would be the next step to figure out if we are going to send some notifications from Camunda or we are going to run some queries on DB to send alerts, or somthing else. Do we have some sort of ticket to handle it?

guru-aot (approved):

PR #927, opened 2021-12-10
Feature/[API][SIMS-815]MSFAA number check

guru-aot:
- [x] Add check to MSFAA number creation and validity check to create a new number if it is a FT or PT program
- [x] Make referenceapplication notNULL
- [x] Review Married/Single data in MSFAA send file, Student Number is also sent to ESDC. **Create columns in application table and add marital status to MSFAA file**
- [x] Ability for MSFAA-Receive file to Cancel MSFAA. During the return file if we receive a cancel for an MSFAA, add attribute of canceled to msfaa number (new column) and add to check for valid MSFAA number


sources/packages/api/src/services/application/application.service.e2e-spec.ts
```diff
@@ -128,20 +133,20 @@ describe("ApplicationService", () => {
...
       try {
         const savedApplication = await applicationService.associateMSFAANumber(
           testApplication.id,
+          OfferingIntensity.fullTime,
         );
         expect(savedApplication.msfaaNumber).toBeTruthy();
         expect(savedApplication.msfaaNumber.id).toBe(testMSFAANumber.id);
       } finally {
-        await applicationRepository.remove(testApplication);
-        await msfaaNumberRepository.remove(testMSFAANumber);
```
guru-aot, line 144: Removed this as when the reference application is assigned, we cannot remove the column as its part of cascading in the model

sources/packages/api/src/services/application/application.service.e2e-spec.ts
```diff
@@ -103,20 +108,20 @@ describe("ApplicationService", () => {
...
```
guru-aot, line 119: Removed this as when the reference application is assigned, we cannot remove the column as its part of cascading in the model

sources/packages/api/src-sql/Applications/Add-col-relationship-status.sql
```diff
@@ -0,0 +1,13 @@
+-- Create relationship_status
```
ann-aot, line 1: may be you can give `add relationship_status`

sources/packages/api/src-sql/Applications/Add-col-relationship-status.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE
+    sims.applications
+ADD
+    COLUMN IF NOT EXISTS relationship_status VARCHAR(20) NOT NULL DEFAULT 'single';
```
ann-aot, line 5: how about creating enum for relationship statues? @devs thoughts?

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
```diff
@@ -0,0 +1,27 @@
...
+    sims.msfaa_numbers
+ADD
+    COLUMN IF NOT EXISTS new_issuing_province VARCHAR(2);
+
+COMMENT ON COLUMN sims.msfaa_numbers.new_issuing_province IS 'MSFAA Number New issuing Province.';
+
+ALTER TABLE
+    sims.msfaa_numbers
+ADD
+    COLUMN IF NOT EXISTS cancelled_date DATE;
```
ann-aot, line 19: is this a timestamp? or only date?

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
```diff
@@ -0,0 +1,27 @@
...
+
+COMMENT ON COLUMN sims.msfaa_numbers.cancelled_date IS 'MSFAA Number cancelled date';
+
+--Default constraint for the offering_intensity is removed after the not null constraint is enforced
```
ann-aot, line 23: i didn't get this part @guru-aot

sources/packages/api/src-sql/Applications/Add-col-relationship-status.sql
dheepak-aot: Agree with @ann-aot. It may be re-used for future as well.

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -189,6 +189,12 @@ export class Application extends RecordDataModel {
   })
   pirDeniedOtherDesc?: string;
 
+  @Column({
+    name: "relationship_status",
+    nullable: false,
+  })
```
ann-aot, line 195: line 194 and 196 is `contradicting`. did you mean `nullable: true`

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
```diff
@@ -0,0 +1,27 @@
...
+    sims.msfaa_numbers
+ADD
+    COLUMN IF NOT EXISTS offering_intensity VARCHAR(10) NOT NULL DEFAULT 'Full Time';
+
+COMMENT ON COLUMN sims.msfaa_numbers.offering_intensity IS 'Offering intensity for the MSFAA Number Part Time/ Full Time.';
+
+ALTER TABLE
+    sims.msfaa_numbers
+ADD
+    COLUMN IF NOT EXISTS new_issuing_province VARCHAR(2);
```
ann-aot, line 12: I guess `new_issuing_province` will always be 2 characters, if so, pls use 'CHAR`. 

Also for province too we can use enum or a table. thoughts @dheepak-aot  @abschwenker  @guru-aot

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
guru-aot: just the date, similar to service_provider_received_date from the response file

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
```diff
@@ -0,0 +1,27 @@
+-- Create offering_intensity
```
dheepak-aot, line 1: Just a suggestion, the file name and the columns added are slightly different. Can give a more general file name like Add-cols-for-msfaa or Add-cols.

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
guru-aot: I have a default value for the existing rows for offering_intensity, removing the default after populating them

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -189,6 +189,12 @@ export class Application extends RecordDataModel {
   })
   pirDeniedOtherDesc?: string;
 
+  @Column({
```
dheepak-aot, line 192: Other new entity models added have comment, but missing here. not a blocker. but nice to have.

sources/packages/api/src/database/entities/application.model.ts
guru-aot: it should not be nullable, changed line 196

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
guru-aot: If we know the values of the province and how they come in the file, its okay to create an enum, else it will error out if we are not sure while processing the file. Or it should be good with CHAR(2) as you said @ann-aot .

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
guru-aot: Yeah even I was skeptical, because I was doing it as an enhancement for MSFAA and also cancelled file changes in the response file. I will change it

sources/packages/api/src-sql/Applications/Add-col-relationship-status.sql
guru-aot: if that is the case, we should have values, as married, single and other. @abschwenker  please confirm

sources/packages/api/src-sql/MSFAANumbers/Add-col-offering-cancelled.sql
abschwenker: the file layouts only support 2 characters. But we wont be using these codes anywhere as its only for records. For this likely not worth using an ENUM type which if there is an issue will cause an issue with the insert

sources/packages/api/src/database/entities/application.model.ts
abschwenker: When a student saves as draft you will not be able to save the relationship status. So I believe it should be nullable

sources/packages/api/src-sql/Applications/Add-col-relationship-status.sql
abschwenker: since we are using this field is other areas and doing a translation between status and a "code" yes i would agree that ENUM would fit

sources/packages/api/src/database/entities/application.model.ts
guru-aot: yes you are right @abschwenker

abschwenker (commented):
Did I miss the student number column to work like relationship status?

sources/packages/api/src/esdc-integration/msfaa-integration/msfaa-response.service.ts
```diff
@@ -105,6 +118,30 @@ export class MSFAAResponseService {
     }
   }
 
+  /**
+   * Process the cancelled record of the MSFAA response file
+   * by updating the columns of MSFAA Numbers table
+   * @param cancelledRecord MSFAA received record with
+   * cancelledDate and newIssusingProvince.
+   */
+  private async processCancelledRecords(
```
dheepak-aot, line 127: the method can have a singular name, when I initially read the name, it looked it it process multiple records(cancellation records).

dheepak-aot:
Good work @guru-aot 👍. Few minor comments.

sources/packages/api/src-sql/Applications/Add-col-relationship-student-number.sql
```diff
@@ -0,0 +1,15 @@
+-- Add relationship_status
+ALTER TABLE
+    sims.applications
+ADD
+    COLUMN IF NOT EXISTS relationship_status sims.relationship_status;
```
dheepak-aot, line 5: Just to confirm, this is not a NOT NULL column anymore? in previous commit I saw a default value of Single to it.

guru-aot: Yes, i removed it as its nullable

ann-aot (approved):
LGTM @guru-aot . Thanks for addressing the comments. Good Work 👍

dheepak-aot (approved):
LGTM. 👍 thanks for making the change @guru-aot

abschwenker (approved):

Issue #357, opened 2021-04-12
Drop downs not loading selected value


mzmichellesmith:
**Additional Context**
- When a drop down is loaded in either read-only or edit the ID is shown and not the value that was selected. As far as I can tell this is happening in every view

**Acceptance criteria**
- [x] Solution way of loading selected value and not the ID when viewing a completed dropdown in view/edit mode. *Note this should also work when the data is no longer available

**SIMILAR TO TICKET 509**
REF: Read-only view should use labels not IDs
**Additional Context**

-Since we are saving and displaying based off PY and the offering dropdown uses the id that are enforced based off active PY non-active PY will not display correct
-source: https://github.com/bcgov/SIMS/pull/777

**Acceptance Criteria**
- [x] Read-only, applications for a non-active PY should display the offering label

- [x] PIR form in institution -> offering and program dropdown

PR #947, opened 2021-12-23
[Api][Web][SIMS #357] Drop downs not loading selected value#357

ann-aot:
- offering and program  dropdown fix for read-only mode in student application and  PIR form application
- icon console error fixed
image
image

image
- when `read-only` is removed from the URL, the form will be still in `read-only` mode
image

- students cant edit an application with inactive program year, irrespective for the application status
image
image


sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -376,13 +376,16 @@ export class ApplicationController extends BaseController {
...
+   * @query includeInActivePY, if includeInActivePY is true,
+   * then consider both active and inactive program year.
    * @returns program year details of the application
    */
   @AllowAuthorizedParty(AuthorizedParties.student)
   @Get(":applicationId/program-year")
   async programYearOfApplication(
     @UserToken() userToken: IUserToken,
     @Param("applicationId") applicationId: number,
+    @Query("includeInActivePY") includeInActivePY?: boolean,
```
dheepak-aot, line 388: This Query argument is repeated many times. Can you declare a constant and use?

sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
```diff
@@ -100,6 +100,7 @@ export class ProgramInfoRequestController {
     result.studentStudyStartDate = application.data.studystartDate;
     result.studentStudyEndDate = application.data.studyendDate;
     result.programYearId = application.programYear.id;
+    result.activeProgramYear = application.programYear.active;
```
dheepak-aot, line 103: When I read activeProgramYear it gave me a quick impression of a year instead of flag. Can we rename the variable to something like  isProgramYearActive or programYearActive ?

sources/packages/web/src/components/aest/StudentApplications.vue
```diff
@@ -153,6 +153,8 @@ export default {
     const programYear = ref({} as ProgramYearOfApplicationDto);
     const showModal = ref(false);
     const selectedApplicationId = ref(0);
+    const toast = useToastMessage();
+    const TOAST_ERROR_DISPLAY_TIME = 15000;
```
dheepak-aot, line 157: Why is the toast time manually set here? I was trying to search in #357 to find any reason. But couldn't get any.

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -124,12 +124,14 @@ export class ApplicationApi extends HttpBaseClient {
...
     try {
-      const response = await this.apiClient.get(
-        `application/${applicationId}/program-year`,
-        this.addAuthHeader(),
-      );
+      let url = `application/${applicationId}/program-year`;
+      if (includeInActivePY) {
+        url = `${url}?includeInActivePY=${includeInActivePY}`;
+      }
+      const response = await this.apiClient.get(url, this.addAuthHeader());
```
dheepak-aot, line 134: I know this is not part of PR, but can change to getCall() and reduce some code. Same for other APIs as well.

sources/packages/web/src/services/http/EducationProgramOfferingApi.ts
```diff
@@ -95,12 +95,14 @@ export class EducationProgramOfferingApi extends HttpBaseClient {
...
     programYearId: number,
     selectedIntensity: OfferingIntensity,
+    includeInActivePY?: boolean,
   ): Promise<OptionItemDto[]> {
     try {
-      const response = await this.apiClient.get(
-        `institution/offering/location/${locationId}/education-program/${programId}/program-year/${programYearId}/options-list?selectedIntensity=${selectedIntensity}`,
-        this.addAuthHeader(),
-      );
+      let url = `institution/offering/location/${locationId}/education-program/${programId}/program-year/${programYearId}/options-list?selectedIntensity=${selectedIntensity}`;
```
dheepak-aot, line 101: can we break the url to 2 lines for better readability?

sources/packages/web/src/views/student/StudentApplicationDetails.vue
```diff
@@ -133,25 +136,30 @@ export default {
...
+        router.push({
+          name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM,
+          params: {
+            selectedForm: programYear.value.formName,
+            programYearId: programYear.value.programYearId,
+            id: props.id,
+          },
+        });
+      } catch (error) {
+        toast.error("Program Year not active", "", TOAST_ERROR_DISPLAY_TIME);
```
dheepak-aot, line 158: You can pass undefined instead of empty string for second argument. I know our composble has signature of string only but the toast.add allows undefined.  We can update the signature of composable.
image

dheepak-aot:
Nice work @ann-aot . 👍. Just few minor comments. FYI I connected the issue #357 to PR.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -221,13 +224,15 @@ export class EducationProgramOfferingController {
     @Param("programId") programId: number,
     @Param("programYearId") programYearId: number,
     @Query("selectedIntensity") selectedIntensity: OfferingIntensity,
```
guru-aot, line 226: can we please move the selectedIntensity query parameter down and make includeInActivePY as mandatory query parameter, so that pass either true or false for the includeInActivePY. SO that we may not use undefined when calling this method .

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -253,12 +260,15 @@ export class EducationProgramOfferingController {
...
     @Param("programYearId") programYearId: number,
+    @Query("includeInActivePY") includeInActivePY?: boolean,
   ): Promise<OptionItem[]> {
     const offerings =
       await this.programOfferingService.getProgramOfferingsForLocation(
         locationId,
         programId,
         programYearId,
+        undefined,
+        includeInActivePY,
```
guru-aot, line 271: Can we please remove the use of undefined, so that change the signature of getProgramOfferingsForLocation, to have mandatory includeInActivePY.

sources/packages/web/src/views/student/StudentApplicationDetails.vue
```diff
@@ -133,25 +136,30 @@ export default {
...
-      router.push({
-        name: StudentRoutesConst.DYNAMIC_FINANCIAL_APP_FORM,
-        params: {
-          selectedForm: programYear.value.formName,
-          programYearId: programYear.value.programYearId,
-          id: props.id,
-        },
-      });
+      try {
+        await getProgramYear();
```
guru-aot, line 148: IMO, it would be better to pass the includeInActivePY variable as false, as it is defined as boolean.

guru-aot (commented):
LGTM, minor comments, lets discuss when you are back.

sources/packages/api/src/route-controllers/application/application.controller.ts
ann-aot: mm, I seriously don't think, there is a use of declaring const for params and query variable, do you have any strong reason @dheepak-aot, pls let me know, if I am missing  anything

sources/packages/web/src/components/aest/StudentApplications.vue
ann-aot: to increase the span of TOAST, same is also implemented in the supporting users by other developers

sources/packages/web/src/views/student/StudentApplicationDetails.vue
ann-aot: image

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: The reason I mentioned here is reason of a constant, because the same text is repeated several times and declaring it once and using constant will minimize the probability of typo and if at all it needs to be changed, just modifying at one place. 
I couldn't imagine any downside of doing it.
Let me know if I am missing anything.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
ann-aot: since `includeInActivePY` is a query parameter, there is no need to send, if it's `undefined` @guru-aot. the query parameters are optional

dheepak-aot (approved):
LGTM @ann-aot . Thanks for making the changes.

guru-aot (approved):
LGTM

Issue #846, opened 2021-11-10
Ministry User Adds, Views  or Resolves a Restriction on Student Account


michesmi:
Reference: "Restriction List vDec 8'21.xlsx"    *General

**As a** *(Ministry User* **I want** *(to: 
- add a restriction to an Account)* **So That** *(there is a hold put on so that I student cannot create an application, receive funding*
- View a restriction on an Account)* **So That** *(I can assist students when they need support)*
- Resolve a restriction on an Account)* **So That** *(The Restriction (Hold) disabling students from applying, receiving funding,*

**Additional Context**
- Related UX Ticket #773
- [ ] Restrictions work the same for Institutions- The Ministry user will just have different types to select from a drop down list. US not yet created 
- ADD THE SPREADSHEET

**Acceptance Criteria**
- [ ] Active restrictions should be at the top of default sort
- [ ] Review Restrictions spreadsheet and adjust DB/API where required to enhance backend to final requirements
**Add Restriction**
- [ ] From the Ministry User Views Institution Restriction Tab. The Ministry User selects "Add Restriction" and is directed to the " Add Restriction" Pop up.
![image.png](https://images.zenhubusercontent.com/60c7d7040a42d68fb4abaa59/44c62fcd-4ad1-40d7-a232-ce14974b1abf)
![image.png](https://images.zenhubusercontent.com/60c7d7040a42d68fb4abaa59/ca7a4ea5-6c8d-44e7-ac19-b121130befcd)
- [ ] Provide the text, "Add new Restriction" "Add a new restriction to this profile. Be sure to add notes with relevant decisions or actions to be taken on this account."
- [ ] Restriction Category- Set to General (for now) other types to come in the future post MVP.
- [ ] Restriction Reason-( Concationation of Restriction Code and Restriction Description) (Drop Down - Types: Verification, Academic, Federal (will come down automatically hopefully so can be left out of this list?) , Appeal (POST MVP), Application Change, Student Loan Delinquency, Designation (Institution Only), Other 
- [ ] Notes- Long Text Free
Capture the following:
- [ ] Date Added
- [ ] Status  (Active)
- [ ] Created by (Staff IDIR)

**Resolve Restriction**
![image.png](https://images.zenhubusercontent.com/60c7d7040a42d68fb4abaa59/86324dda-bcd1-4464-bf0c-ba4dada16238)
Show the Following:
- [ ]  Restriction Type
- [ ] Show the Reason 
- [ ] Show the Notes
[...]

PR #956, opened 2022-01-06
[API][SIMS #846] Minor fix to include restrictions without creator

dheepak-aot:
The Federal restrictions are created by importing file. Hence as of now, it does not have creator. Changing the query from inner join to left join to accommodate these restrictions.


ann-aot (approved):
LGTM @dheepak-aot

guru-aot (approved):
LGTM, @dheepak-aot

abschwenker (approved):

Issue #693, opened 2021-09-02
Student SIN Validation -Update User Experience 


michesmi:
Additional Context
- When a Student first creates their account their account is flagged to have their SIN validates with CRA. During this time until their SIN comes back as a match we need to restirct the student from doing certain things. (Similar to restrictions)

Acceptance Criteria
- [ ] While the students SIN validation is pending (sims.students.valid_sin=null) display yellow banner and stop the student from doing everything a restriction does, AND stop the student from applying for PD
- Text: "StudentAid BC is currently validating your account information with CRA. This process normally takes up to 3 hours. If you have not heard from us please contact us [link to contact]"
- [ ] While the students SIN validation is False (sims.students.valid_sin=false) display RED banner and stop the student from doing everything a restriction does, AND stop the student from applying for PD
- Text: "StudentAid BC has found an issue with your account information. Please ensure your account information matches CRA and log back in. If you have further issues please contact us"

PR #963, opened 2022-01-13
 [Api][Web][SIMS #693]Student SIN Validation

henryoforeh-aot:
For this PR, I have done the following:

- Created a decorator (**CheckSinValidation**) that is used in the Application Controller.
- Created sin validation guard to ensure that students cannot invoke certain endpoints when their Sin is null or invalid.
- Added **validSIN** attribute to the studentprofile class. This attribute is used to map the validSIN attribute from the response payload when the endpoint for studentprofile is called.
- Created an attribute called validSIN in the StudentState state object to store the value of validSIN in the redux store when the it is retrieved from via the backend service api call.
- Created a reusable banner component to read the value of the validSIN from the redux store and render the appropriate message with the appropriate message with the color.
ScreenShorInvalid
ScreenShotPending


sources/packages/api/src/auth/guards/sin-validation.guard.ts
```diff
@@ -0,0 +1,34 @@
...
+      CHECK_SIN_VALIDATION_KEY,
+      [context.getHandler(), context.getClass()],
+    );
+
+    if (!checkSinValidation) {
+      return true;
+    }
+    const { user } = context.switchToHttp().getRequest();
+    const userToken = user as IUserToken;
+    return this.studentService.getStudentSinStatus(userToken.userId);
```
dheepak-aot, line 32: await is missing here

sources/packages/api/src/auth/guards/sin-validation.guard.ts
```diff
@@ -0,0 +1,34 @@
...
+export class SinValidationGuard implements CanActivate {
+  constructor(
+    private reflector: Reflector,
+    private readonly studentService: StudentService,
+  ) {}
+  /**
+   * Implementation of canActivate method in CanActivate.
+   * With the user token, it checks for sin validation for the user.
+   * @param context
+   * @returns Promise<boolean>
```
ann-aot, line 19: can you add details to the `return` comment, something like `return true for valid SIN and false for invalid SIN`

sources/packages/api/src/auth/guards/sin-validation.guard.ts
```diff
@@ -0,0 +1,34 @@
...
+      CHECK_SIN_VALIDATION_KEY,
+      [context.getHandler(), context.getClass()],
+    );
+
+    if (!checkSinValidation) {
+      return true;
+    }
+    const { user } = context.switchToHttp().getRequest();
+    const userToken = user as IUserToken;
+    return await this.studentService.getStudentSinStatus(userToken.userId);
```
ann-aot, line 33: I think, when returning an async call `await` is not needed, consumer function will be awaiting I believe, correct me if I am wrong @dheepak-aot  @guru-aot

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -176,6 +176,20 @@ export class StudentService extends RecordDataModelService<Student> {
     return this.repo.find({ validSIN: null });
   }
 
+  /**
+   * Gets a student's valid sin status
+   * @returns Students SIN validation status.
```
ann-aot, line 181: you can `@parms ` too in the comment

sources/packages/api/src/auth/guards/sin-validation.guard.ts
guru-aot: Yes you are right @ann-aot

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -176,6 +176,20 @@ export class StudentService extends RecordDataModelService<Student> {
...
+   */
+  async getStudentSinStatus(userId: number): Promise<boolean | null> {
+    const sinStatus = await this.repo
+      .createQueryBuilder("student")
+      .leftJoinAndSelect("student.user", "user")
```
dheepak-aot, line 186: A student definitely has to be part of user, so innerJoin is more appropriate here. You can ignore andSelect from join as we do not require any user property here.

sources/packages/web/src/constants/message-constants.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * SIN Validation Message
+ */
+export const PENDING_SIN_MESSAGE =
```
ann-aot, line 4: 👍

sources/packages/web/src/store/modules/student/student.ts
```diff
@@ -10,16 +10,19 @@ export class StudentProfile {
   gender?: string;
   givenName?: string;
   middleName?: string;
+  validSIN: any;
```
ann-aot, line 13: can you add the proper type, here it will be boolean

sources/packages/api/src/auth/guards/sin-validation.guard.ts
```diff
@@ -0,0 +1,35 @@
+import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common";
+import { Reflector } from "@nestjs/core";
+import { StudentService } from "../../services";
+import { CHECK_SIN_VALIDATION_KEY } from "../decorators/check-sin-status.decorator";
+import { IUserToken } from "../userToken.interface";
+/**
+ * This guard validates an API for Valid Sin if it is decorated with @CheckRestriction.
+ */
+@Injectable()
+export class SinValidationGuard implements CanActivate {
```
guru-aot, line 10: Can you please change the class name to SINValidationGuard from SinValidationGuard.

sources/packages/api/src/auth/guards/sin-validation.guard.ts
```diff
@@ -0,0 +1,35 @@
+import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common";
+import { Reflector } from "@nestjs/core";
+import { StudentService } from "../../services";
+import { CHECK_SIN_VALIDATION_KEY } from "../decorators/check-sin-status.decorator";
+import { IUserToken } from "../userToken.interface";
+/**
+ * This guard validates an API for Valid Sin if it is decorated with @CheckRestriction.
```
guru-aot, line 7: Sin --> SIN, please change

sources/packages/api/src/auth/guards/sin-validation.guard.ts
dheepak-aot: @ann-aot you are right. Await is not needed actually because, in the service method await has been already used. Sonar caught it.

@henryoforeh-aot Sorry you can remove the await.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -176,6 +176,20 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      .where("student.userId = :userIdParam", { userIdParam: userId })
+      .select("student.validSin")
+      .getOne();
+    return sinStatus.validSIN;
```
ann-aot, line 190: actually, if you check other service methods, you can see that we are following like, service will be returning the DB entity as such, and on consumer side, the needed data is retrieved. we are doing like so that we can reuse the service layer

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -176,6 +176,20 @@ export class StudentService extends RecordDataModelService<Student> {
...
+  /**
+   * Gets a student's valid sin status
+   * @returns Students SIN validation status.
+   */
+  async getStudentSinStatus(userId: number): Promise<boolean | null> {
+    const sinStatus = await this.repo
+      .createQueryBuilder("student")
+      .leftJoinAndSelect("student.user", "user")
+      .where("student.userId = :userIdParam", { userIdParam: userId })
+      .select("student.validSin")
```
guru-aot, line 188: image
 Can we please change the validSin to "student.validSIN" as part of db model?

sources/packages/web/src/views/student/SINStatusBanner.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <Message :severity="severity">
```
dheepak-aot, line 2: The <Message> can be directly used in CheckValidSINBanner.vue instead of being wrapped inside SINStatusBanner.vue.

sources/packages/web/src/store/modules/student/student.ts
```diff
@@ -43,9 +46,17 @@ export const student = {
     SET_HAS_STUDENT_ACCOUNT(state: StudentState, hasStudentAccount: boolean) {
       state.hasStudentAccount = hasStudentAccount;
     },
+
+    SET_VALID_SIN(state: StudentState, validSIN: any) {
```
ann-aot, line 50: `validSIN: any`, please replace any this proper type

ann-aot (commented):
LGTM 👍 Added some comments

sources/packages/api/src/services/student/student.service.ts
ann-aot: yes, as the user will be already there, we can use `inner join`

ann-aot: actually, I didn't get the other 2 parts `You can ignore andSelect from join as we do not require any user property here.` and `like
.andWhere("user.isActive = :isActive", { isActive: true })
this join might be needed.`

dheepak-aot: @guru-aot thanks for asking this.
Looking at this service, the purpose it solely to get the SIN status, even a inactive user can have a verified SIN status and in that case this would return false for a user who has verified SIN. 

Looking at security aspect you have a point but, I assume if a user is inactive, then that student will not be able to login to our system. In this case there is no scope of showing any banner and blocking access at UI. Looking at the API level where we added this checkSINDecorator, inactive users cannot get past these APIs.

@abschwenker @ann-aot what is your opinion.

ann-aot: yes @dheepak-aot , inactive users will not be able to sign in

dheepak-aot: @ann-aot to answer you on the 2nd point what i meant is leftJoin**AndSelect**("student.user", "user") can be replaced with .innerJoin("student.user", "user")

dheepak-aot:
Good work at your first story @henryoforeh-aot 👍Added some comments.

sources/packages/api/src/services/student/student.service.ts
henryoforeh-aot: Correct, there is already a service that returns Studentinfo entity currently used on the calling side. This one was created specifically for the decorator to check validSIN status.

sources/packages/web/src/types/contracts/StudentContract.ts
```diff
@@ -91,6 +91,15 @@ export enum StudentPDStatus {
   Pending = "Pending",
 }
 
+/**
+ * Enumeration for student sin status.
+ */
+export enum SinStatusEnum {
```
guru-aot, line 97: Please change it to SINStatusEnum

ann-aot (approved):
LGTM 👍 good work @henryoforeh-aot  👍

sources/packages/web/src/store/modules/student/student.ts
```diff
@@ -10,19 +12,25 @@ export class StudentProfile {
...
+  sinStatus?: SinStatusEnum;
+  severity?: string;
+  message?: string;
 }
 
 export class StudentState {
   profile: StudentProfile = new StudentProfile();
   hasStudentAccount = false;
-  validSIN = false;
+  sinValidStatus: SINValidStatus = new SINValidStatus();
```
dheepak-aot, line 27: @henryoforeh-aot We could have used the SinStatusEnum here instead of SINValidStatus. Am I missing something ?

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -176,6 +176,21 @@ export class StudentService extends RecordDataModelService<Student> {
     return this.repo.find({ validSIN: null });
   }
 
+  /**
+   * Gets a student's valid sin status
+   * @returns Students SIN validation status.
+   * @param userId information needed to select the user.
+   */
+  async getStudentSinStatus(userId: number): Promise<Student> {
+    const sinStatus = await this.repo
```
dheepak-aot, line 185: you can directly return the await statement here. It will clear the sonar code smell that we got.

sources/packages/web/src/store/modules/student/student.ts
henryoforeh-aot: I created class called SINValidStatus to encapsulate the following attributes sinstatus, severity and message. This was done to save the state of those attributes and reduce logic in the CheckValidSINBanner.vue component.

dheepak-aot (approved):
LGTM @henryoforeh-aot 👍. Thanks for doing the changes.

guru-aot (approved):
LGTM, Nice work @henryoforeh-aot . Thanks for implementing the changes 👍

abschwenker (approved):
lgtm

Issue #917, opened 2021-12-06
Ministry User Views  Programs and Study Period Offering Details 


michesmi:
**Additional Context**
- This ticket does NOT require DB changes, DB changes will be a part of #894

**Acceptance Criteria**
- As a ministry user when viewing program summary:
- [x] Should display "No records found."   -Not a toast message
- [x] Update UI badge style in table
- [x] Review discrepancies with wireframe
- [x] Review pagination implementation
- As a ministry user when clicking "view" from program summary:
- [x] Create shared component for ministry and institution user program summary view (ministry user will have "view" buttons NOT "edit")
- [x] Update UI badges and date headers
- [x] Review discrepancies with wireframe
- As a ministry user when clicking "view" on the program summary program page PROGRAM:
- [x] display read-only view of program details page
- As a ministry user when clicking "view" on the program summary program page OFFERING:
- [x] `study period` in the offering summary should not include application specific offering? and the count should match the offering id inside?
- [x] display read-only view of offering details page
- [x] remove `locationName` from the route and take the name from API. (BUG - > EDIT OR CREATE offering  then click 'back all programs` from the view program page and location at the muted heading is undefined)
- [x] total offering count in summary include application specific offerings?

IrenaMilosevic1:
Reviewed Dev. Looks very good. Thanks Ann

PR #968, opened 2022-01-21
[Api][Web][SIMS#917] Ministry User Views Programs and Study Period Offering Details #917

ann-aot:
- [x] resolved PIR bug raised by Lisa.
- [x] added paginations for program and offering summary, added search by offering name after discussion with Lynn
- [x] update UI Badges
- [x] Program description was not properly displayed in the details page
- [x] added new classes for headings, sub classes
- [x] updated program summary query, now program with `0` offerings will also be listed in the AEST program summary.
- [x] resolved location name label bug in the heading(BUG - > EDIT OR CREATE offering then click 'back all programs` from the view program page and location at the muted heading is undefined)
- [x] image
the hyperlink was wrong , updated
- [x] code smell due to code comment and different controllers with same logics, for different portals
- screenshots
image
image
image
image
image
image
image
image
image
image
image
image


sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -301,53 +293,64 @@ export class EducationProgramOfferingController {
...
+  @Get(":offeringId/aest")
+  async getProgramOfferingForAEST(
+    @Param("offeringId") offeringId: number,
+  ): Promise<ProgramOfferingDto> {
+    //To retrieve Education program offering corresponding to ProgramId and LocationId
+    const offering = await this.programOfferingService.getOfferingById(
+      offeringId,
+    );
+    if (!offering) {
+      throw new UnprocessableEntityException(
```
andrewsignori-aot, line 350: I believe that it should be 404 because the id, that is part of the URL does not exists, right?

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -301,53 +293,64 @@ export class EducationProgramOfferingController {
...
+        "Not able to find a Education Program Offering associated with the current Education Program, Location and offering.",
```
andrewsignori-aot, line 351: The message seems to be going way further than it really is, IMO. Is it not just an "offering not found because the id does not exist" situation?

sources/packages/api/src/route-controllers/education-program/education-program.controller.ts
```diff
@@ -35,27 +47,42 @@ export class EducationProgramController {
...
+    @Query("searchProgramName") searchProgramName: string,
+    @Query("sortField") sortField: string,
+    @Query("sortOrder") sortOrder: FieldSortOrder,
     @UserToken() userToken: IInstitutionUserToken,
-  ): Promise<SummaryEducationProgramDto[]> {
-    const programs = await this.programService.getSummaryForLocation(
+    @Query("page") page = DEFAULT_PAGE_NUMBER,
+    @Query("pageLimit") pageLimit = DEFAULT_PAGE_LIMIT,
+  ): Promise<EducationProgramsSummaryPaginated> {
+    // [OfferingTypes.applicationSpecific] offerings are created during PIR, if required
```
andrewsignori-aot, line 73: Does this comment belongs to this scope?

andrewsignori-aot, line 73: Can this comment be expanded? 
// [OfferingTypes.applicationSpecific] offerings are created during PIR, if required, and they are supposed to be viewed only associated to the application that they were associated to during the PIR, hence they should not be displayed alongside with the public offerings.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
ann-aot: i agree, updating

sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
```diff
@@ -213,19 +213,9 @@ export class ProgramInfoRequestController {
...
-        )
-      ) {
-        throw new CustomNamedError(
-          "Offering Intensity does not match the students intensity",
-          OFFERING_INTENSITY_MISMATCH,
-        );
-      }
       // studyStartDate from payload is set as studyStartDate
       let studyStartDate = payload.studyStartDate;
+      let selectedOfferingIntensity = payload.offeringIntensity;
```
andrewsignori-aot, line 218: In my understating there is no scenario where during the PIR complete process there will be no offering available on "payload.selectedOffering". If the offering is not present the API can throw a unprocessable exception.
I am confused with the need of initialization of variables like studyStartDate and selectedOfferingIntensity.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.models.ts
```diff
@@ -10,17 +8,11 @@ export class EducationProgramOfferingModel {
...
-      return "No program dates";
-    } else {
-      return `${dayjs(this.studyStartDate).format(
-        EXTENDED_DATE_FORMAT,
-      )} - ${dayjs(this.studyEndDate).format(EXTENDED_DATE_FORMAT)}`;
-    }
-  }
 }
-
+export class PaginatedOffering {
```
andrewsignori-aot, line 12: Can we change it to be a generic interface and move to a generic place. I believe that this will be used for many other paginated results.

sources/packages/api/src/utilities/datatable-config.ts
```diff
@@ -10,3 +10,12 @@ export enum FieldSortOrder {
  */
 export const DEFAULT_PAGE_LIMIT = 10;
 export const DEFAULT_PAGE_NUMBER = 0;
+
+// pagination option
+export class PaginationOptions {
+  searchName?: string;
```
andrewsignori-aot, line 16: I believe that _searchName_ is too specific, can we rename it to _searchCriteria_?

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -70,17 +78,38 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+      // default sort and order
+      offeringsQuery.orderBy(
+        databaseFieldOfOfferingDataTable(DEFAULT_SORT_FIELD),
+        FieldSortOrder.ASC,
       );
     }
+    // pagination
+    offeringsQuery
+      .take(paginationOptions.pageLimit)
+      .skip(paginationOptions.page * paginationOptions.pageLimit);
```
andrewsignori-aot, line 107: I would say that it would be more meaningful call skip and then take, just a suggestion.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -70,17 +78,38 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
 
-    const queryResult = await offeringsQuery.getMany();
+    // result
+    const queryResult = await offeringsQuery.getManyAndCount();
 
-    return queryResult.map((educationProgramOffering) => {
+    const offerings = queryResult[0].map((educationProgramOffering) => {
```
andrewsignori-aot, line 112: Can you please destructor the array in line 110, something like below to avoid the [0] access?
```javascript
// result
const [records, count] = await offeringsQuery.getManyAndCount();
```

sources/packages/api/src/services/education-program/education-program.service.models.ts
```diff
@@ -16,17 +16,9 @@ export class EducationProgramsSummary {
...
-  name: string;
-  description: string;
-  credentialType: string;
-  cipCode: string;
-  nocCode: string;
-  sabcCode: string;
-  approvalStatus: string;
-  programIntensity: ProgramIntensity;
-  institutionProgramCode: string;
+export class EducationProgramsSummaryPaginated {
```
andrewsignori-aot, line 21: As mentioned earlier, I believe that we can create a generic interface to avoid creating the type over and over again.

sources/packages/api/src/utilities/database.ts
```diff
@@ -18,3 +18,23 @@ export async function configureIdleTransactionSessionTimeout(
...
+ * will return count without innerJoins), For those cases we can use this helper
+ * function to get the actual total count of the raw data
+ * @param sqlQuery the sql query (result of typeorm .getSql()).
+ * if we want the total count, then pass the sql query without the
+ * pagination logic (i.e, without .skip(), .take(), .limit() or .offset())
+ * @param parameter pass the parameter of the sql query.
+ * @return the total count
+ */
+export async function getRawCount(
+  repo: Repository<any>,
```
andrewsignori-aot, line 35: The generic repository can be declared with amore specific type then any. Can you please change it?

sources/packages/api/src/utilities/database.ts
```diff
@@ -18,3 +18,23 @@ export async function configureIdleTransactionSessionTimeout(
...
+  sqlQuery: string,
+  parameter: any[],
+): Promise<[{ count: number }]> {
```
andrewsignori-aot, line 38: Is there a reason to not return directly the number of records from this method?

sources/packages/api/src/utilities/database.ts
```diff
@@ -18,3 +18,23 @@ export async function configureIdleTransactionSessionTimeout(
 ): Promise<void> {
   queryRunner.query(`SET idle_in_transaction_session_timeout = '${seconds}s'`);
 }
+
+/**
+ * This helper to to get the total count of the Raw for the pagination.
```
andrewsignori-aot, line 23: There is a "to to"

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -154,24 +174,175 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
       )
-      .where("programs.institution.id = :institutionId", { institutionId })
-      .getRawMany();
+      .where("programs.institution.id = :institutionId", { institutionId });
+
+    const queryParams: any[] = [...offeringTypes, locationId, institutionId];
+    // program name search
+    if (paginationOptions.searchName) {
+      summaryResult.andWhere("programs.name Ilike :searchProgramName", {
+        searchProgramName: `%${paginationOptions.searchName}%`,
```
andrewsignori-aot, line 192: I believe that if the :searchProgramName is going t be provided as a global parameter in queryParams there is no need to provide it at line 192.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
ann-aot: make sense

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -154,24 +174,175 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+   * @param paginationOptions pagination options
+   * @returns summary for location
+   */
+  async getPaginatedProgramsForAEST(
+    institutionId: number,
+    offeringTypes: OfferingTypes[],
+    paginationOptions: PaginationOptions,
+  ): Promise<ProgramsSummaryPaginated> {
+    // default data table sort field
+    const sortByColumn = "programs.createdAt"; //Default sort column
```
andrewsignori-aot, line 257: I see other places declaring the same as "DEFAULT_SORT_FIELD"

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -154,24 +174,175 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+    const paginatedProgramQuery = this.repo
+      .createQueryBuilder("programs")
+      .select([
+        "programs.id as programId",
```
andrewsignori-aot, line 261: If the select/addSelect is used here (e.g. addSelect("programs.name", "programName")) it could avoid using the all lower case properties at line 328 on. Here there will be maybe a confusion in the future on why the alias is declared with the correct case but the properties have all lower case letters.

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -154,24 +174,175 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+
+    // total raw count
+    const totalCount = await getRawCount(
+      this.repo,
+      paginatedProgramQuery.getSql(),
+      queryParams,
+    );
+
+    if (paginationOptions?.pageLimit) {
+      paginatedProgramQuery.limit(paginationOptions.pageLimit);
```
andrewsignori-aot, line 304: Can you elaborate on why skip/take was not used for this query?

sources/packages/api/src/database/entities/education-program.model.ts
```diff
@@ -157,7 +158,7 @@ export class EducationProgram extends RecordDataModel {
   @Column({
     name: "approval_status",
   })
-  approvalStatus: string;
+  approvalStatus: ApprovalStatus;
```
dheepak-aot, line 161: Nice work that we have used enum for Approval Status. Do we have a plan or backlog ticket to change the database column to use enum respectively?

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -274,4 +431,58 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+        "programs.isWILApproved",
+        "programs.wilProgramEligibility",
+        "programs.hasTravel",
+        "programs.travelProgramEligibility",
+        "programs.hasIntlExchange",
+        "programs.intlExchangeProgramEligibility",
+        "programs.programDeclaration",
+        "institution.id",
+      ])
+      .innerJoin("programs.institution", "institution")
```
andrewsignori-aot, line 484: If the _institutionId_ is all that we need maybe we can avoid adding the _innerJoin_ if we add a @RelationId to _institutionId_ in _EducationProgram_ entity model.

sources/packages/web/src/assets/css/base.scss
```diff
@@ -102,3 +108,34 @@ ul.no-bullets {
...
+  line-height:$line-height-header-md;
+  /* Black */
+  color: $brand-black;
+  opacity: 0.5;
+  margin: 0px;
+}
+
+
+/** large heading **/
+.heading-x-large {
```
andrewsignori-aot, line 126: Is it too different from the category-header-large that we have for form.io?
If not, I believe that we should use the same styles, moving the form.io styles into the base.scss

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -81,32 +85,43 @@ export class EducationProgramOfferingController {
...
+      locationId,
+      programId,
+      {
+        searchName: searchName,
+        sortField: sortField,
+        sortOrder: sortOrder,
+        page: page,
+        pageLimit: pageLimit,
+      },
+      [OfferingTypes.public],
```
dheepak-aot, line 123: Nice 👍

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -81,32 +85,43 @@ export class EducationProgramOfferingController {
     return createdProgramOffering.id;
   }
 
+  /**
+   * Offering Summary for an institution location
```
dheepak-aot, line 89: LocationId missing in the comment @param

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -301,53 +293,64 @@ export class EducationProgramOfferingController {
   }
 
   /**
-   * Get programs for a particular institution with pagination.
```
dheepak-aot, line 304: locationId missing in the comment.

sources/packages/web/src/components/common/InstitutionUserSummary.vue
```diff
@@ -1,57 +1,55 @@
...
-          @keyup.enter="searchUserTable()"
-        />
-        <v-btn
-          @click="searchUserTable()"
-          tile
-          class="ml-2 primary-btn-background"
-        >
-          <font-awesome-icon :icon="['fas', 'search']" />
-        </v-btn>
+  <div class="mb-4">
```
andrewsignori-aot, line 3: IMO this component is not a good one to be shared because the Ministry one is going to be way too simple and the Institution will be more complex.
This could be noticed in the amount of times that there is a "if" to check the client type consuming it. Even the API call are different.
Just would like to share these thoughts, if there is a Team decision to move on as it is it is fine 😉

sources/packages/api/src/route-controllers/education-program-offering/models/education-program-offering.dto.ts
```diff
@@ -52,3 +52,36 @@ export interface ProgramOfferingDto extends ProgramOfferingBaseDTO {
...
+    studyStartDate: offering.studyStartDate,
+    studyEndDate: offering.studyEndDate,
+    actualTuitionCosts: offering.actualTuitionCosts,
+    programRelatedCosts: offering.programRelatedCosts,
+    mandatoryFees: offering.mandatoryFees,
+    exceptionalExpenses: offering.exceptionalExpenses,
+    tuitionRemittanceRequestedAmount: offering.tuitionRemittanceRequestedAmount,
+    offeringDelivered: offering.offeringDelivered,
+    lacksStudyDates: offering.lacksStudyDates,
+    lacksStudyBreaks: offering.lacksStudyBreaks,
```
henryoforeh-aot, line 76: is lacksStudyBreaks a mandatory field that will always have a value returned ? If not please can you add a check for null ?

sources/packages/web/src/types/contracts/GeneralStatusForBadge.ts
```diff
@@ -3,6 +3,9 @@ export enum GeneralStatusForBadge {
   InActive = "inactive",
   ActiveRestriction = "active restriction",
   ResolvedRestriction = "resolved",
+  Approved = "approved",
```
andrewsignori-aot, line 6: Are we putting every single status for a badge in one single Enum?

sources/packages/api/src/route-controllers/education-program-offering/models/education-program-offering.dto.ts
```diff
@@ -52,3 +52,36 @@ export interface ProgramOfferingDto extends ProgramOfferingBaseDTO {
...
+    lacksFixedCosts: offering.lacksFixedCosts,
+    tuitionRemittanceRequested: offering.tuitionRemittanceRequested,
+    offeringIntensity: offering.offeringIntensity,
+    yearOfStudy: offering.yearOfStudy,
+    showYearOfStudy: offering.showYearOfStudy,
+    hasOfferingWILComponent: offering.hasOfferingWILComponent,
+    offeringWILType: offering.offeringWILType,
+    studyBreaks: offering.studyBreaks,
```
henryoforeh-aot, line 84: for studyBreaks, please see previous comments I made on lacksStudyBreaks

sources/packages/web/src/views/aest/institution/Programs.vue
```diff
@@ -40,8 +31,11 @@
           @sort="pageSortEvent($event)"
           :loading="loading"
         >
+          <template #empty>
+            <p class="text-center font-weight-bold">No records found.</p>
+          </template>
           <Column
-            field="submittedDate"
+            :field="AESTProgramFields.SubmittedDate"
```
andrewsignori-aot, line 38: These program fields are going to be reused all the times that we are doing a sort on the programs, right? Why does it includes the prefix "AEST"?

sources/packages/api/src/route-controllers/education-program/education-program.controller.ts
```diff
@@ -35,27 +47,42 @@ export class EducationProgramController {
     private readonly formService: FormService,
   ) {}
 
+  /**
+   * Get programs for a particular institution with pagination.
+   * @param locationId id of the location.
+   * @param pageSize is the number of rows shown in the table
```
dheepak-aot, line 53: The param name is pageLimit.

sources/packages/api/src/route-controllers/education-program/education-program.controller.ts
```diff
@@ -125,13 +152,13 @@ export class EducationProgramController {
...
   /**
-   * This retuns only the subset of the EducationProgram to get
+   * This returns only the subset of the EducationProgram to get
    * the complete EducationProgram DTO use the @Get(":id") method
    * @param programId
    * @returns
    */
   @AllowAuthorizedParty(AuthorizedParties.institution)
-  @Get(":programId/summary")
+  @Get(":programId/details")
```
dheepak-aot, line 161: What was the reason behind the URL value change?

sources/packages/api/src/route-controllers/education-program/education-program.controller.ts
```diff
@@ -280,4 +307,62 @@ export class EducationProgramController {
...
+   * @param page is the number of rows that is skipped/offset from the total list.
+   * For example page 2 the skip would be 10 when we select 10 rows per page.
+   * @param sortColumn the sorting column.
+   * @param sortOrder sorting order.
+   * @param searchProgramName Search the program name in the query
+   * @returns ProgramsSummaryPaginated.
+   */
+  @AllowAuthorizedParty(AuthorizedParties.aest)
+  @Groups(UserGroups.AESTUser)
+  @Get("institution/:institutionId")
```
dheepak-aot, line 347: API URL can have aest in it specifying this is for Ministry. Also it will help to avoid potential router conflicts.

sources/packages/api/src/route-controllers/education-program/models/save-education-program.dto.ts
```diff
@@ -36,6 +37,13 @@ export interface EducationProgramDto {
   programDeclaration: boolean;
 }
 
+export interface EducationProgramData extends EducationProgramDto {
```
henryoforeh-aot, line 40: Please can you keep the prefix/suffix of naming convention here in sync? **EducationProgramData** is a **Dto** which extends a another base **EducationProgramDto**  **Dto** interface that is specific to education program with respect to common attributes. Example, something like  **export interface EducationProgramDataDto extends BaseEducationProgramDto**  will suffice.

sources/packages/api/src/utilities/database.ts
```diff
@@ -18,3 +18,23 @@ export async function configureIdleTransactionSessionTimeout(
...
+ * pagination logic (i.e, without .skip(), .take(), .limit() or .offset())
+ * @param parameter pass the parameter of the sql query.
+ * @return the total count
+ */
+export async function getRawCount(
+  repo: Repository<any>,
+  sqlQuery: string,
+  parameter: any[],
+): Promise<[{ count: number }]> {
+  return repo.query(`SELECT COUNT(*) FROM (${sqlQuery}) AS count`, parameter);
```
guru-aot, line 39: Instead of using the sql query, we can use .getRawMany() in the called query and can get the .length of it. So that we will have a single transaction.

sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
dheepak-aot: I agree. In this scenario, we can avoid the if on line 219 and initialization at line 218.

sources/packages/web/src/components/common/OfferingSummary.vue
```diff
@@ -0,0 +1,249 @@
...
+    },
+  },
+  setup(props: any) {
+    const router = useRouter();
+    const loading = ref(false);
+    const searchBox = ref("");
+    const currentPage = ref();
+    const currentPageLimit = ref();
+    const { formatDate } = useFormatters();
+    const DATE_FORMAT = "MMMM, D YYYY";
```
andrewsignori-aot, line 118: We already this format defined in the _useFormatters_ on the method _dateOnlyLongString_. Can the conde be changed to use it, please?

sources/packages/web/src/composables/useFormatStatuses.ts
```diff
@@ -0,0 +1,20 @@
+import { ApprovalStatus, GeneralStatusForBadge } from "@/types";
+
+export function useFormatStatuses() {
+  const getProgramStatusToGeneralStatus = (
```
andrewsignori-aot, line 4: As we mentioned in the last call we should get rid of the _GeneralStatusForBadge_.
Just check if makes sense change this code now or when we get rid of it.

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -118,12 +118,26 @@ export function useFormatters() {
     };
   };
 
+  /**
+   * format a date to given format
+   * @param date date
+   * @param dateFormat format
+   * @returns extended date format like March, 22 2021
+   */
+  const formatDate = (date: Date | string, dateFormat: string): string => {
```
andrewsignori-aot, line 127: I believe that the idea is concentrate in this _formater_ all the formats needed, so for the case using this method I believe that we already have the method _dateOnlyLongString_.

sources/packages/web/src/constants/routes/RouteConstants.ts
```diff
@@ -82,6 +82,9 @@ export const AESTRoutesConst = {
   APPLICATION_DETAILS: Symbol(),
   LOGIN: Symbol(),
   LOGIN_WITH_NOT_ALLOWED_USER: Symbol(),
+  VIEW_PROGRAM: Symbol(),
+  VIEW_OFFERING: Symbol(),
+
```
andrewsignori-aot, line 87: Please remove extra space.

sources/packages/web/src/services/http/EducationProgramApi.ts
```diff
@@ -49,15 +54,41 @@ export class EducationProgramApi extends HttpBaseClient {
...
-        this.addAuthHeader(),
-      );
-      return response.data as SummaryEducationProgramDto[];
+      let URL = `institution/education-program/location/${locationId}/summary?page=${page}&pageLimit=${pageCount}`;
+      if (searchProgramName) {
+        URL = `${URL}&searchProgramName=${searchProgramName}`;
+      }
+      if (sortField && sortOrder) {
+        const sortDBOrder =
+          sortOrder === DataTableSortOrder.DESC
```
andrewsignori-aot, line 85: Do we need a _DataTableSortOrder_ and a _FieldSortOrder_? If both are just ASC/DESC can we use only the FieldSortOrder that seems to have a more generic name?

sources/packages/web/src/services/http/EducationProgramOfferingApi.ts
```diff
@@ -33,16 +38,42 @@ export class EducationProgramOfferingApi extends HttpBaseClient {
...
+      let URL = `institution/offering/location/${locationId}/education-program/${programId}?page=${page}&pageLimit=${pageCount}`;
+      if (searchName) {
+        URL = `${URL}&searchName=${searchName}`;
+      }
+      if (sortField && sortOrder) {
+        const sortDBOrder =
+          sortOrder === DataTableSortOrder.DESC
+            ? FieldSortOrder.DESC
+            : FieldSortOrder.ASC;
+        URL = `${URL}&sortField=${sortField}&sortOrder=${sortDBOrder}`;
```
andrewsignori-aot, line 73: This logic is going to be repeated many times, can we have a small helper to build the URL with the sort parameters always in the same way?

sources/packages/web/src/services/http/EducationProgramOfferingApi.ts
```diff
@@ -33,16 +38,42 @@ export class EducationProgramOfferingApi extends HttpBaseClient {
...
+    sortField?: OfferingSummaryFields,
+    sortOrder?: DataTableSortOrder,
+  ): Promise<PaginatedOffering> {
     try {
-      const response = await this.apiClient.get(
-        `institution/offering/location/${locationId}/education-program/${programId}`,
-        this.addAuthHeader(),
-      );
-      return response.data as EducationProgramOfferingDto[];
+      let URL = `institution/offering/location/${locationId}/education-program/${programId}?page=${page}&pageLimit=${pageCount}`;
```
andrewsignori-aot, line 64: My understating is that when a variable starts with an acronym it must be lower case, in this case it would be "url" IMO.

andrewsignori-aot (commented):
I went through the first round of the review.
Great work enabling the pagination and sort and sharing the components between Institution and Ministry. Please take a look at the comments 😉

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -70,17 +78,38 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
-        "offerings.offeringType in (:...offeringTypes)",
-        {
-          offeringTypes,
-        },
+      offeringsQuery.andWhere("offerings.offeringType in (:...offeringTypes)", {
+        offeringTypes,
+      });
+    }
+    // search offering name
+    if (paginationOptions?.searchName) {
```
dheepak-aot, line 86: Is paginationOptions an optional field? if so it can be declared at line 62 as optional

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -154,24 +174,175 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+      summaryItem.credentialTypeToDisplay = credentialTypeToDisplay(
+        summary.credentialtype,
+      );
       summaryItem.approvalStatus = summary.approvalstatus;
       summaryItem.totalOfferings = summary.totalofferings;
       return summaryItem;
     });
+    return {
+      programsSummary: programSummary,
+      totalProgram: totalCount[0].count,
```
dheepak-aot, line 240: Is there a way that we can get const[totalCount] =  await getRawCount() so that you can skip the index reference.

sources/packages/api/src/services/education-program-offering/education-program-offering.service.models.ts
ann-aot: updated as discussed

sources/packages/api/src/services/education-program/education-program.service.models.ts
ann-aot: updated as discussed

sources/packages/api/src/utilities/datatable-config.ts
dheepak-aot: I believe this is the text coming from search input text. IMO.  searchText would be more self-descriptive.

sources/packages/web/src/assets/css/base.scss
```diff
@@ -1,7 +1,7 @@
 @import "./global-style-variables.scss";
 html * {
   font-family: "BCSans", "Noto Sans", Verdana, Arial, sans-serif, "primeicons";
-  font-style: normal;
+  font-style: $font-style-normal;
```
dheepak-aot, line 4: Effort taken to improve the existing code is much appreciated 👏

sources/packages/web/src/assets/css/global-style-variables.scss
```diff
@@ -50,6 +52,14 @@ $border-color-info: #6554c0;
 
 /** Status badge color variables **/
 $badge-border-warning: rgb(255, 122, 0, 0.2);
-$badge-border-error: rgb(228, 34, 46, 0.2);
+$badge-border-error: rgb(228, 34, 46);
```
dheepak-aot, line 55: As per our discussion, please restore the opacity.

sources/packages/web/src/components/common/InstitutionUserSummary.vue
```diff
@@ -234,6 +233,7 @@ export default {
           );
           break;
       }
+      loading.value = false;
```
dheepak-aot, line 236: Just to check, are we planning to use loader for all our data-tables?

sources/packages/web/src/components/common/ProgramDetails.vue
```diff
@@ -0,0 +1,168 @@
...
+    </span>
+    <status-badge
+      class="ml-2"
+      :status="getProgramStatusToGeneralStatus(educationProgram.approvalStatus)"
+    ></status-badge>
+    <v-btn
+      class="float-right"
+      outlined
+      @click="programButtonAction()"
+      color="#2965C5"
```
dheepak-aot, line 14: can use color from constants file.

sources/packages/web/src/components/common/ProgramDetails.vue
```diff
@@ -0,0 +1,168 @@
...
+          props.programId,
+        );
+      } else if (isAESTUser.value) {
+        educationProgram.value = await EducationProgramService.shared.getEducationProgramForAEST(
+          props.programId,
+        );
+      }
+    };
+
+    onMounted(getEducationProgramAndOffering);
```
dheepak-aot, line 155: await ?

sources/packages/web/src/components/generic/StatusBadge.vue
```diff
@@ -16,6 +16,7 @@ status Badge
 <script lang="ts">
 import { ref, onMounted, watch } from "vue";
 import { GeneralStatusForBadge } from "@/types";
+import "@/assets/css/global-style-variables.scss";
```
dheepak-aot, line 19: why did we import global-style-variables here? I do not see an usage of sass variables

sources/packages/web/src/composables/useFormatters.ts
dheepak-aot: Yeah. But we have 2 date formatters for different formats. This function will simplify that letting us to transform dates flexibly.

sources/packages/web/src/services/http/EducationProgramApi.ts
dheepak-aot: Datatable sort order is 1,-1 but in typeorm it is "ASC","DESC"

sources/packages/web/src/services/http/EducationProgramApi.ts
```diff
@@ -133,4 +164,24 @@ export class EducationProgramApi extends HttpBaseClient {
...
+  /**
+   * Education Program Details for ministry users
+   * @param programId program id
+   * @returns Education Program Details
+   */
+  public async getEducationProgramForAEST(
+    programId: number,
+  ): Promise<EducationProgramData> {
+    try {
+      const response = await this.apiClient.get(
```
dheepak-aot, line 177: Does API return any error that needs to be shown on toast? if not we can use getCall

sources/packages/web/src/views/institution/locations/programs/LocationProgramAddEdit.vue
```diff
@@ -1,30 +1,40 @@
...
-    </v-container>
-  </v-sheet>
+  <p class="muted-heading-text">
+    <a @click="goBack()">
+      <v-icon left> mdi-arrow-left </v-icon> Program detail</a
+    >
+  </p>
+  <span class="heading-x-large">
+    <span v-if="isReadonly">View Program</span>
+    <span v-if="programId && !isReadonly">Edit Program</span>
```
dheepak-aot, line 9: why is programId checked here? can it be null for this component?

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
```diff
@@ -47,45 +68,114 @@
...
+    };
+    const loadProgramDetails = async () => {
+      locationDetails.value = await InstitutionService.shared.getInstitutionLocation(
+        props.locationId,
+      );
+    };
+    // search program table
+    const searchProgramTable = async () => {
+      await loadSummary(
+        currentPage.value ?? DEFAULT_PAGE_NUMBER,
```
dheepak-aot, line 155: we can use || instead for the advantage of compatible browsers

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
```diff
@@ -47,45 +68,114 @@
...
+        currentPageLimit.value ?? DEFAULT_PAGE_LIMIT,
+      );
+    };
+    onMounted(async () => {
+      await loadSummary();
```
dheepak-aot, line 160: We can user Promise.all for better efficiency.

dheepak-aot:
Great work @ann-aot . Much appreciated effort on taking the extra mile to fix existing issues and refactoring code. 👏

sources/packages/api/src/utilities/datatable-config.ts
andrewsignori-aot: I believe that both are _self-descriptive_ 😉

sources/packages/api/src/utilities/database.ts
andrewsignori-aot: @guru-aot do you mean execute the _.getRawMany()_, retrieve all the records and then do the ".length"?
The parallel could happen either way, there is no need to change it.

guru-aot: Sure, @andrewsignori-aot . As discussed, we can handle just the parallel call.

sources/packages/web/src/composables/useFormatters.ts
andrewsignori-aot: Using this function the consumer need use this function plus the date format. We need to have the date format centralized, right? 
So why just don't give the format a nick name and consume the method directly?

sources/packages/web/src/services/http/EducationProgramApi.ts
andrewsignori-aot: Thanks @dheepak-aot I did not noticed it 😉

sources/packages/web/src/components/common/InstitutionUserSummary.vue
andrewsignori-aot: If it that is a PrimeVue specific feature I would avoid spending time on it.

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
andrewsignori-aot: In my understanding this is a concern while doing form.io but not for Vue or any modern UI framework. Please let me know if we need further explanations here.

dheepak-aot: @andrewsignori-aot once we had a discussion on NOA story, we noticed that Nullish coalescing operator is not supported in IE and Opera. While || is supported for all browsers.

image

sources/packages/api/src/services/education-program/education-program.service.ts
ann-aot: as discussed, I was not able to access the `institution id` by specifying the name 
image
it is working only when we select whole entity like 
image

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
ann-aot: thanks for the catch 👍

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
ann-aot: thanks for the catch 👍

sources/packages/api/src/route-controllers/education-program-offering/models/education-program-offering.dto.ts
ann-aot: I had a look into what is discussed yesterday, we are actually checking if `offerings` is there or not before the transforming to dto (`line 360`),  so, we need not need an offering null, 
image

sources/packages/api/src/route-controllers/education-program-offering/models/education-program-offering.dto.ts
ann-aot: added comment above

sources/packages/api/src/route-controllers/education-program/education-program.controller.ts
ann-aot: yes, this controller actually returns a program details, not a program summary

sources/packages/api/src/services/education-program-offering/education-program-offering.service.ts
ann-aot: `paginationOptions ` is not optional, but don't need null check, removed

sources/packages/api/src/services/education-program/education-program.service.ts
ann-aot: this code was changed

sources/packages/web/src/components/common/InstitutionUserSummary.vue
ann-aot: `loading` is primevue dataTable property

sources/packages/web/src/views/institution/locations/programs/LocationProgramAddEdit.vue
ann-aot: yes, as in the label, if `programId` is not there, then, it is in create mode

sources/packages/web/src/composables/useFormatters.ts
ann-aot: this function is no longer required, removed

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.controller.ts
```diff
@@ -81,32 +86,48 @@ export class EducationProgramOfferingController {
...
+   * @returns paginated offering summary
+   */
   @AllowAuthorizedParty(AuthorizedParties.institution)
   @HasLocationAccess("locationId")
   @Get("location/:locationId/education-program/:programId")
   async getAllEducationProgramOffering(
     @Param("locationId") locationId: number,
     @Param("programId") programId: number,
-  ): Promise<EducationProgramOfferingDto[]> {
+    @Query("searchName") searchName: string,
```
andrewsignori-aot, line 107: Please rename the parameter to _searchCriteria_ also.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -578,12 +578,9 @@ export class ApplicationService extends RecordDataModelService<Application> {
     if (
       sortField &&
       sortField !== "status" &&
-      databaseFieldOfApplicationDataTable(sortField)
+      sortApplicationsSummary(sortField)
```
andrewsignori-aot, line 581: It seems that it is still targeting the UI. Is it not supposed to be something like _sortApplicationsColumnMap_?

sources/packages/api/src/services/education-program/education-program.service.ts
```diff
@@ -133,45 +144,210 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+      .addSelect("programs.cipCode", "cipCode")
+      .addSelect("programs.credentialType", "credentialType")
+      .addSelect("programs.approvalStatus", "approvalStatus")
       .addSelect(
         (query) =>
           query
             .select("COUNT(*)")
             .from(EducationProgramOffering, "offerings")
             .where("offerings.educationProgram.id = programs.id")
+
```
andrewsignori-aot, line 171: Please remove the extra blank line.

sources/packages/api/src/services/institution/institution.service.ts
```diff
@@ -37,7 +37,7 @@ import {
 import { AccountDetails } from "../bceid/account-details.model";
 import { InstitutionUserAuthDto } from "../../route-controllers/institution/models/institution-user-auth.dto";
 import {
-  databaseFieldOfUserDataTable,
+  sortUsersSummary,
```
andrewsignori-aot, line 40: I would say the "Summary" seems to be UI specific. I would recommend something like _sortUsersColumnMap_.

sources/packages/api/src/utilities/datatable-config.ts
```diff
@@ -10,3 +10,12 @@ export enum FieldSortOrder {
  */
 export const DEFAULT_PAGE_LIMIT = 10;
 export const DEFAULT_PAGE_NUMBER = 0;
+
+// pagination option
+export class PaginationOptions {
```
andrewsignori-aot, line 15: Can you please covert it to an interface if there is no reason to it be a class?

sources/packages/api/src/utilities/datatable-database-mapping-utils.ts
```diff
@@ -6,32 +6,66 @@
  * @param fieldName
  * @returns fieldName
  */
-export const databaseFieldOfUserDataTable = (fieldName: string): [] => {
-  const databaseFieldOfUserDataTableMap = {
+export const sortUsersSummary = (fieldName: string): [] => {
+  const userSortOptions = {
```
andrewsignori-aot, line 10: Cool solution for multiple sorting using a single identifier (displayName) 👍

sources/packages/web/src/components/common/ProgramDetails.vue
andrewsignori-aot: There is no need in this context 😉

sources/packages/web/src/types/contracts/GeneralStatusForBadge.ts
ann-aot: updated

sources/packages/web/src/components/common/OfferingSummary.vue
ann-aot: removed. as it no longer needed

sources/packages/web/src/components/generic/StatusBadge.vue
ann-aot: removed

andrewsignori-aot (commented):
Thanks again for doing the adjustments. It is getting better and better 😉

sources/packages/web/src/helpers/helpers/uri.ts
```diff
@@ -0,0 +1,23 @@
+import { DataTableSortOrder, FieldSortOrder } from "@/types";
+
+/**
+ * helper to append pagination sort and order to the url
+ * @param url api url
+ * @param sortField sortField
+ * @param sortOrder sortOrder
+ * @returns url with append sort options to url
+ */
+export function addSortOptions(
```
andrewsignori-aot, line 10: Please add the return type
```javascript
export function addSortOptions: string(
```

andrewsignori-aot (approved):
Looks good to me 👍 
Again, great work and if we missed anything we still have time to adjust and improve 😄

sources/packages/web/src/components/common/ProgramDetails.vue
dheepak-aot: Ah Yes. await is inside getEducationProgramAndOffering.

dheepak-aot (approved):
Great work @ann-aot . 👏 Thanks for making the changes.

sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
ann-aot: resolving as discussed

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/api/src/database/entities/education-program.model.ts
ann-aot: added in story, https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/578

henryoforeh-aot (approved):

Issue #422, opened 2021-05-06
Implement FT/PT Application Overlap


mzmichellesmith:
**Additional Context**
- A student cannot have multiple applications (fulltime or parttime) with overlapping dates

**Examples of NOT allowed**
- FullTime Application: Sept 01 - Dec 01
-FullTime Application: Dec 01- March 01
(Allowed: Parttime||fulltime Application: Dec 02- March 01)

- FullTime Application: Sept 01 - Dec 01
-Parttime Application: Dec 01- March 01
(Allowed: Parttime||fulltime Application: Dec 02- March 01)

- Parttime Application: Sept 01 - Dec 01
-FullTime Application: Dec 01- March 01
(Allowed: Parttime||fulltime Application: Dec 02- March 01)

- Parttime Application: Sept 01 - Dec 01
-Parttime Application: Dec 01- March 01
(Allowed: Parttime||fulltime Application: Dec 02- March 01)

Acceptance Criteria:
- [x] Implement vacation check where a student cannot have 2 applications submitted with the same dates.
- [x] If an application has a PIR waiting (Or any case where offering_id is NULL) do not allow student to submit another application
- [x] Check must include the sfas.applications table as well
- [x] Applications that are cancelled or overwritten are NOT considered in this validation
- [x] Show toast saying "Your application had overlapping dates with another application of the same type. Please review your application dates"

PR #970, opened 2022-01-22
[Api][Web][SIMS #422] Application overlap validation

dheepak-aot:
- [x] Implement validation check where a student cannot have 2 applications submitted with the same dates.
- [x] If an application has a PIR waiting (Or any case where offering_id is NULL) do not allow student to submit another application
- [x] Check must include the sfas_applications and sfas_part_time_applications table and  as well
- [x] Applications that are draft, cancelled or overwritten are NOT considered in this validation
- [x] Show toast saying "Your application had overlapping dates with another application of the same type. Please review your application dates"
image


sources/packages/api/src/database/entities/base-sfas-application.model.ts
```diff
@@ -0,0 +1,24 @@
+import { JoinColumn, ManyToOne, Column } from "typeorm";
```
dheepak-aot, line 1: Moved individual related properties to base model to avoid code duplication issue

ann-aot (commented):
As discussed, please add the 2 logic to check the overlap in all queries @dheepak-aot . rests looks good to me @dheepak-aot

dheepak-aot:
@ann-aot That was an excellent catch! Thanks. I added the missing or condition in query

.github/workflows/deploy.yml
```diff
@@ -74,6 +74,7 @@ jobs:
       ATBC_LOGIN_ENDPOINT: "https://auth.at-bc.ca/api/login"
       ATBC_ENDPOINT: "https://sabc.at-bc.ca/api"
       BYPASS_CRA_INCOME_VERIFICATION: true
+      BYPASS_VALIDATIONS_AT_LOCAL: true
```
andrewsignori-aot, line 77: I believe the "AT_LOCAL" lost the meaning since we are using it also on DEV. Can we rename it to something more expressive like:
BYPASS_APPLICATION_SUBMIT_VALIDATIONS

ann-aot (approved):
LGTM @dheepak-aot . Thanks for adding the logic. Good Work 👍

sources/packages/api/src/database/entities/base-sfas-application.model.ts
andrewsignori-aot: We are doing segregating the entity models at this level so far. I believe that it was done due to SonarCloud, but IMO we should keep in the way that we have been doing and ask @abschwenker to ignore the issue on SonarCloud if it is a false positive.

henryoforeh-aot (approved):
Nice

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -161,6 +175,43 @@ export class ApplicationController extends BaseController {
...
     );
+    /** Validation for application overlapping dates or Pending PIR.
+     * This validation can be disabled by setting BYPASS_VALIDATIONS_AT_LOCAL to true in .env file.
+     * */
+
+    if (!this.config.bypassValidationsAtLocal) {
+      const existingOverlapApplication =
+        await this.applicationService.validatePIRAndDateOverlap(
+          userToken.userId,
+          subtractDays(studyStartDate, 1),
```
andrewsignori-aot, line 186: As we talked, the BETWEEN in Postgres is inclusive and maybe there is something else to be checked. So, the subtractDays and addDays would not be necessary here.

sources/packages/api/src/utilities/application-utils.ts
```diff
@@ -47,3 +55,16 @@ export const transformToApplicationSummaryDTO = (
     status: application.applicationStatus,
   } as ApplicationSummaryDTO;
 };
+
+/**
+ * Throw validation exception on Application date overlap or PIR Pending.
+ */
+export const throwExceptionForPIRDateOverlap = (
+  application: Application | SFASApplication | SFASPartTimeApplications,
```
andrewsignori-aot, line 63: It could be simplified as a Boolean variable also.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
```diff
@@ -46,6 +46,51 @@ export class SFASPartTimeApplicationsService
...
+    birthDate: Date,
+    lastName: string,
+    studyStartDate: Date,
+    studyEndDate: Date,
+  ): Promise<SFASPartTimeApplications> {
+    return this.repo
+      .createQueryBuilder("sfasPTApplication")
+      .select(["sfasPTApplication.id"])
+      .innerJoin("sfasPTApplication.individual", "sfasPTstudent")
+      .where("sfasPTstudent.lastName = :lastName", { lastName })
```
andrewsignori-aot, line 69: Please do the comparison in a not case sensitive way. If I am not wrong the user last name already has a index for lower(lastName).

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
```diff
@@ -46,6 +46,51 @@ export class SFASPartTimeApplicationsService
...
+          qb.where(
+            "sfasPTApplication.startDate BETWEEN :startDatePT AND :endDatePT",
+            { startDatePT: studyStartDate, endDatePT: studyEndDate },
+          )
+            .orWhere(
+              "sfasPTApplication.endDate BETWEEN :startDatePT AND :endDatePT",
+              { startDatePT: studyStartDate, endDatePT: studyEndDate },
+            )
+            .orWhere(
+              "startDatePT: BETWEEN sfasPTApplication.startDate AND sfasPTApplication.endDate",
```
andrewsignori-aot, line 83: Please check the parameter syntax for "startDatePT:" on line 83. The ":" seems to be in the wrong place, same for line 87.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
```diff
@@ -46,6 +46,51 @@ export class SFASPartTimeApplicationsService
...
+   * @param studyEndDate
+   * @returns SFAS part-time application.
+   */
+  async validateDateOverlap(
+    sin: string,
+    birthDate: Date,
+    lastName: string,
+    studyStartDate: Date,
+    studyEndDate: Date,
+  ): Promise<SFASPartTimeApplications> {
```
andrewsignori-aot, line 64: The return type can be the object or null, right? If yes we can include the null as a possible return type.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
```diff
@@ -46,6 +46,51 @@ export class SFASPartTimeApplicationsService
...
+      .select(["sfasPTApplication.id"])
+      .innerJoin("sfasPTApplication.individual", "sfasPTstudent")
+      .where("sfasPTstudent.lastName = :lastName", { lastName })
+      .andWhere("sfasPTstudent.sin = :sin", { sin })
+      .andWhere("sfasPTstudent.birthDate = :birthDate", { birthDate })
+      .andWhere(
+        new Brackets((qb) => {
+          qb.where(
+            "sfasPTApplication.startDate BETWEEN :startDatePT AND :endDatePT",
+            { startDatePT: studyStartDate, endDatePT: studyEndDate },
```
andrewsignori-aot, line 76: Just as a alternative, when the same parameter must be passed over and over again, like startDatePT, we could be provided once to the query builder using the setParameter method in the query builder.

sources/packages/api/src/services/sfas/sfas-application.service.ts
```diff
@@ -50,6 +50,51 @@ export class SFASApplicationService
...
+    birthDate: Date,
+    lastName: string,
+    studyStartDate: Date,
+    studyEndDate: Date,
+  ): Promise<SFASApplication> {
+    return this.repo
+      .createQueryBuilder("sfasApplication")
+      .select(["sfasApplication.id"])
+      .innerJoin("sfasApplication.individual", "sfasFTstudent")
+      .where("sfasFTstudent.lastName = :lastName", { lastName })
```
andrewsignori-aot, line 73: Please do the comparison in a not case sensitive way. If I am not wrong the user last name already has a index for lower(lastName).

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1482,4 +1482,51 @@ export class ApplicationService extends RecordDataModelService<Application> {
       })
       .getOne();
   }
+
+  async validatePIRAndDateOverlap(
```
andrewsignori-aot, line 1486: It would be good to some comments explaining why the PIR is being validated and why it matters.
I believe that the idea that we talked in the past was to try to comment as much as possible and goes beyond the technical and explain a bit more the business part.
For instance, I do know what the query is doing just looking it, but maybe someone not used to the code will not know why we are looking for an offering id that is null, that it means that there is PIR and that means that we do not have sure about the rea start date and end date of the application.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
dheepak-aot: Query is returning undefined on a happy path.

sources/packages/api/src/route-controllers/application/application.controller.ts
```diff
@@ -161,6 +175,43 @@ export class ApplicationController extends BaseController {
     const student = await this.studentService.getStudentByUserId(
       userToken.userId,
     );
+    /** Validation for application overlapping dates or Pending PIR.
+     * This validation can be disabled by setting BYPASS_VALIDATIONS_AT_LOCAL to true in .env file.
+     * */
+
+    if (!this.config.bypassValidationsAtLocal) {
+      const existingOverlapApplication =
+        await this.applicationService.validatePIRAndDateOverlap(
```
andrewsignori-aot, line 184: I would say that, usually the overlap will not happen and the three queries will be executed, so I believe that the three queries can be executed in parallel, what also eliminates the need of creating the method _throwExceptionForPIRDateOverlap_ because after the parallel execution returns we can check if any query return something we can display the error.

andrewsignori-aot (commented):
Good work 😉 

abschwenker (approved):

sources/packages/api/src/database/entities/base-sfas-application.model.ts
dheepak-aot: Will get the sonar rule in place and moving forward get rid of such situtation.

.github/workflows/deploy.yml
dheepak-aot: Env variable renamed.

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: As discussed, the between function did include the date when we match with exact start date or end date. It was an issue with Postgres and Node running in different time zones. Thank you for helping to sort the issue out.

sources/packages/api/src/utilities/application-utils.ts
dheepak-aot: With updated code the util became obsolete and removed.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
dheepak-aot: lower function added.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
dheepak-aot: fixed.

sources/packages/api/src/services/sfas/sfas-application.service.ts
dheepak-aot: lower added.

sources/packages/api/src/services/application/application.service.ts
dheepak-aot: Comment added.

sources/packages/api/src/route-controllers/application/application.controller.ts
dheepak-aot: All the validation resolved as single promise as discussed.

andrewsignori-aot (approved):
Great work, thanks for the adjustments 👍

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
```diff
@@ -46,6 +46,51 @@ export class SFASPartTimeApplicationsService
...
+    birthDate: Date,
+    lastName: string,
+    studyStartDate: Date,
+    studyEndDate: Date,
+  ): Promise<SFASPartTimeApplications> {
+    return this.repo
+      .createQueryBuilder("sfasPTApplication")
+      .select(["sfasPTApplication.id"])
+      .innerJoin("sfasPTApplication.individual", "sfasPTstudent")
+      .where("lower(sfasPTstudent.lastName) = lower(:lastName)", { lastName })
```
guru-aot, line 69: Why are we searching for an application to match, only with the last name of the student?

dheepak-aot: Lastname + DOB + SIN is the formula to match a student in SFAS individual table.

sources/packages/api/src/services/sfas/sfas-part-time-application.service.ts
guru-aot: The reason for the undefined, is due to the getOne() at the end, I feel it would be better to fetch the count of any application that are overlapping, if we are not using the application related data in the front end to show it.

guru-aot (approved):
LGTM, I have a minor comment on the query to check for overlapping application. Nice Work @dheepak-aot

PR #972, opened 2022-01-24
[Api][SIMS#443]Document Number

henryoforeh-aot:
- I created a a new function called **updateDisbursementSchedule** in disbursement-schedule-service. This function invokes the logic generate a new document number.
- I removed the implementation that generates a new document number and assigns it to the document number in the **createDisbursementSchedules** function  when a new disbursement is created.
- **updateDisbursementSchedule** is invoked in the confirmation-of-enrollment controller after the application status is **completed**


henryoforeh-aot:
Wrong PR, I'm closing

PR #973, opened 2022-01-24
[Api][Web][SIMS#443]Document Number

henryoforeh-aot:
- I created a a new function called updateDisbursementSchedule in disbursement-schedule-service. This function invokes the logic generate a new document number.

- I removed the implementation that generates a new document number and assigns it to the document number in the createDisbursementSchedules function when a new disbursement is created.

- updateDisbursementSchedule is invoked in the confirmation-of-enrollment controller after the application status is completed.

- I implemented functionality in the front end web application to disable students from creating or editing/updating applications when their SIN status is not valid.

### DIFF
--- a/sources/packages/api/src/auth/auth.module.ts
+++ b/sources/packages/api/src/auth/auth.module.ts
@@ -10,6 +10,8 @@ import {
   KeycloakService,
   TokensService,
   UserService,
+  StudentService,
+  SFASIndividualService,
 } from "../services";
 import { JwtAuthGuard } from "./guards/jwt-auth.guard";
 import { JwtStrategy } from "./jwt.strategy";
@@ -21,6 +23,7 @@ import {
   ActiveUserGuard,
   GroupsGuard,
   RestrictionsGuard,
+  SINValidationGuard,
 } from "./guards";
 import { RolesGuard } from "./guards/roles.guard";
 
@@ -39,6 +42,8 @@ const jwtModule = JwtModule.register({
     KeycloakService,
     ConfigService,
     JwtStrategy,
+    StudentService,
+    SFASIndividualService,
     {
       provide: APP_GUARD,
       useClass: JwtAuthGuard,
@@ -70,6 +75,10 @@ const jwtModule = JwtModule.register({
     {
       provide: APP_GUARD,
       useClass: RestrictionsGuard,
+    },
+    {
+      provide: APP_GUARD,
+      useClass: SINValidationGuard,
     },
   ],
   exports: [jwtModule, TokensService, KeycloakService],
--- a/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
+++ b/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
@@ -208,6 +208,11 @@ export class ConfirmationOfEnrollmentController {
       );
     }
 
+    //TODO: Document Number should be Updated in the same transaction when COE status changes
+    await this.disbursementScheduleService.updateDisbursementScheduleDocumentNumber(
+      applicationId,
+    );
+
     const updatedCOEStatus =
       await this.applicationService.updateApplicationCOEStatus(
         applicationId,
@@ -221,9 +226,6 @@ export class ConfirmationOfEnrollmentController {
       );
     }
 
-    await this.disbursementScheduleService.updateDisbursementScheduleDocumentNumber(
-      applicationId,
-    );
     // Send a message to allow the workflow to proceed.
     await this.workflow.sendConfirmCOEMessage(application.assessmentWorkflowId);
   }
--- a/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
+++ b/sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
@@ -17,6 +17,7 @@ import {
   APPLICATION_NOT_FOUND,
   WorkflowActionsService,
   COEDeniedReasonService,
+  DisbursementScheduleService,
 } from "../../services";
 import {
   Application,
@@ -37,6 +38,7 @@ export const COE_REQUEST_NOT_FOUND_ERROR = "COE_REQUEST_NOT_FOUND_ERROR";
 @Controller("institution/location")
 export class ConfirmationOfEnrollmentController {
   constructor(
+    private readonly disbursementScheduleService: DisbursementScheduleService,
     private readonly applicationService: ApplicationService,
     private readonly workflow: WorkflowActionsService,
     private readonly deniedCOEReasonService: COEDeniedReasonService,
@@ -218,6 +220,10 @@ export class ConfirmationOfEnrollmentController {
         "Confirmation of Enrollment and application status update to completed is failed",
       );
     }
+
+    await this.disbursementScheduleService.updateDisbursementScheduleDocumentNumber(
+      applicationId,
+    );
     // Send a message to allow the workflow to proceed.
     await this.workflow.sendConfirmCOEMessage(application.assessmentWorkflowId);
   }
--- a/sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
+++ b/sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
@@ -93,7 +93,6 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
       const newDisbursement = new DisbursementSchedule();
       newDisbursement.disbursementDate = disbursement.disbursementDate;
       newDisbursement.negotiatedExpiryDate = disbursement.negotiatedExpiryDate;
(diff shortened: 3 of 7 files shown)


sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -236,4 +235,20 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
...
+  /**
+   * Update DisbursementSchedule with documentNumber
+   * @param documentNumber document Number
+   * @returns the result of update.
+   */
+  async updateDisbursementScheduleDocumentNumber(
+    applicationId: number,
+  ): Promise<DisbursementSchedule> {
+    const disbursementScheduleToUpdate = await this.repo.findOne({
+      application: { id: applicationId },
```
ann-aot, line 248: mm, actually Typeform has this `.update()` feature, you can use that here

sources/packages/web/src/components/aest/StudentApplications.vue
```diff
@@ -139,6 +151,12 @@ export default {
       required: true,
     },
   },
+  computed: {
+    sinValidStatus() {
```
ann-aot, line 155: actually, you can write the computed filed in the `setup()`

sources/packages/web/src/views/student/financial-aid-application/Applications.vue
```diff
@@ -15,13 +15,21 @@
...
 
 export default {
   props: {
     hasRestriction: {
       type: Boolean,
     },
   },
+  computed: {
+    sinValidStatus() {
+      const store = useStore();
```
ann-aot, line 29: same here, computed can be written inside `setup`

ann-aot (commented):
Added some comments @henryoforeh-aot , also I think you missed to remove the first `document number` saving logic from the disbursement schedule controller, I didn't find the removed file here,  please let me know if I am missing anything

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -236,4 +235,20 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
...
+
+  /**
+   * Update DisbursementSchedule with documentNumber
+   * @param documentNumber document Number
+   * @returns the result of update.
+   */
+  async updateDisbursementScheduleDocumentNumber(
+    applicationId: number,
+  ): Promise<DisbursementSchedule> {
+    const disbursementScheduleToUpdate = await this.repo.findOne({
```
andrewsignori-aot, line 247: If I am not wrong, an application can have one or two disbursement schedules associate with it. How do we know that we are updating the expected one here?

sources/packages/web/src/components/aest/StudentApplications.vue
```diff
@@ -80,7 +85,12 @@
                 "
               />
             </v-btn>
-            <v-btn :disabled="hasRestriction" plain>
+            <v-btn
+              :disabled="
+                hasRestriction || sinValidStatus !== SINStatusEnum.VALID
```
andrewsignori-aot, line 90: Is this validation happening also in the API?

sources/packages/web/src/components/aest/StudentApplications.vue
```diff
@@ -126,6 +136,8 @@ import { useRouter } from "vue-router";
 import { StudentRoutesConst } from "@/constants/routes/RouteConstants";
 import ConfirmEditApplication from "@/components/students/modals/ConfirmEditApplication.vue";
 import CancelApplication from "@/components/students/modals/CancelApplicationModal.vue";
+import { useStore } from "vuex";
+import { SINStatusEnum } from "@/types";
```
andrewsignori-aot, line 140: Please check the code smell related to the duplicated declaration of "@/types".

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -218,6 +220,10 @@ export class ConfirmationOfEnrollmentController {
         "Confirmation of Enrollment and application status update to completed is failed",
       );
     }
+
+    await this.disbursementScheduleService.updateDisbursementScheduleDocumentNumber(
```
andrewsignori-aot, line 224: It is nice to have the number being generated in a separated function, but should it not be saved altogether with the status change of the COE on the 212? What happens if the COE is updated but the execution of the line 224 fails?
I believe that it should be all together in the same "transaction" to avoid DB inconsistences.

andrewsignori-aot (commented):
Please take a look at some comments 😉

sources/packages/web/src/components/aest/StudentApplications.vue
henryoforeh-aot: Yes validation is happening in the API. I created a decorator called **@CheckSinValidation()** in the previous PR I worked on.

andrewsignori-aot: Nice work with the decorator, please just include it in _auth.module.ts_ as we talked 😉

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
```diff
@@ -206,6 +208,11 @@ export class ConfirmationOfEnrollmentController {
       );
     }
 
+    //TODO: Document Number should be Updated in the same transaction when COE status changes
```
dheepak-aot, line 211: Sequentially COE confirmation should happen first and then document number should be generated. The order doesn't matter if it is under same transaction.
I see a todo here. Until the todo is implemented, this sequence should be followed IMO.

dheepak-aot:
I just have one comment and one clarification needed. Rest all good.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.controller.ts
henryoforeh-aot: I had a chat with @andrewsignori-aot before moving the line of code up. There will be more work coming down the line   which will also involve some refactoring.

sources/packages/api/src/auth/auth.module.ts
```diff
@@ -71,6 +76,10 @@ const jwtModule = JwtModule.register({
       provide: APP_GUARD,
       useClass: RestrictionsGuard,
     },
+    {
```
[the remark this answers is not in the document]
henryoforeh-aot: Yes, I actually forgot to test the decorator downstream.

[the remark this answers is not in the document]
dheepak-aot: Okay got it.

ann-aot (approved):
LGTM 👍 Good work @henryoforeh-aot . Thanks for doing the chnages

dheepak-aot (approved):
LGTM @henryoforeh-aot  👍

guru-aot (approved):
LGTM, nice work @henryoforeh-aot

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -185,7 +185,7 @@ export class StudentService extends RecordDataModelService<Student> {
     return await this.repo
       .createQueryBuilder("student")
       .innerJoin("student.user", "user")
-      .where("student.userId = :userIdParam", { userIdParam: userId })
+      .where("student.user_Id = :userIdParam", { userIdParam: userId })
```
ann-aot, line 188: we can use `      .where("user.id= :userId ", { userId  })`, since the query already has a user `innerjoin`

sources/packages/web/src/components/aest/StudentApplications.vue
```diff
@@ -195,6 +209,9 @@ export default {
     };
 
     onMounted(async () => {
+      sinValidStatus.value = computed(
```
andrewsignori-aot, line 212: The computed property does not need another ref variable. Can you please check other places where the computed was created and used?

sources/packages/web/src/views/student/financial-aid-application/Applications.vue
```diff
@@ -25,6 +28,15 @@ export default {
   setup() {
     const router = useRouter();
     const toast = useToastMessage();
+    const sinValidStatus = ref(1);
+    const store = useStore();
+
+    onMounted(async () => {
+      sinValidStatus.value = computed(
```
andrewsignori-aot, line 35: The computed property does not need another ref variable. Can you please check other places where the computed was created and used?

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
andrewsignori-aot: The way that the query was changed does not change the fact that the update can happen in two records because these two records will be associated with the same application.
I will not consider this as a blocker because there will be a second feature to deal with the second COE.

andrewsignori-aot (approved):
Nice work 😉 Looks good!

abschwenker (approved):

Issue #406, opened 2021-04-27
Institution User Administrator Determines which locations they want to be designated on the Locations Page


mzmichellesmith:
This is ticket 1 of 2 for the designation feature

UX Ticket #892, Ticket 2 of 2 #467

As discussed:
Ticket 1:
Institution summary (common component) (list)
DB changes
request form
submit
Load form

Ticket 2:
Read-only form
ministry approve
ministry decline
ministry pending summary page
student application drop down

**Additional Context**
- Designation is the process of an institution signing an agreement with SABC in order to to be eligible for students to apply for funding
- A - https://studentaidbc.ca/sites/all/files/school-officials/policy_manual.pdf
- B - https://studentaidbc.ca/sites/all/files/school-officials/admin_manual.pdf
- D - https://studentaidbc.ca/sites/all/files/school-officials/Schedule_D_Designation_Agreement.pdf

**Acceptance Criteria**
- [x] Create DB migration for the designation feature

Institution Portal
- [x] On click of "manage designation" display designation summary page. This component will also be displayed on the ministry portal
- Submitted On
- Start date
- Expiry date
- status (of the agreement)
- View button (always read-only)
- Request designation (only clickable by legal signing authority)
- [x] On click of Request designation (only clickable by legal signing authority) display the designation agreement form (content and layout see figma)
- [x] Only Institution name, Institution type, Status, designated locations, submit button exist for All institutions. The remaining fields exist for ONLY BC Privates (hide form fields)
- Institution name (top level, not the location). The institution will be retrieve from the institution "Legal Operating Name", same from BCeID, that is not currently editable.
- Institution type
- Status (of the request)
- Designated locations (Name, address, and checkbox for "request for designation)
- Signing officers (minimum 1 required but can enter multiple, Name, Position (title), email, phone
- Enrolment officers (minimum 1 required but can enter multiple, Name, Position (title), email, phone
- Checkbox's for Schedule A, B D
- Autofill Legal Signing Authority (name and email)
- Agreement text
- Checkbox to agree
- submit button
- [x] On submit, redirect back to summary screen and show the Designation request in pending
- [x] Note: Only 1 request can be in pending at a time

Tech Notes
- Designated means: a location with an approved designation agreement that covers todays date
- If ministry needs to remove Designation for 1 location
Re-approve a DA with different locations
- If ministry needs to remove Designation for all locations
Decline an approved DA
- If institutions needs to add Designation for 1 location
Re-create a DA with different locations
[...]

PR #987, opened 2022-02-04
[web][API][SIMS #406]Institution Submits Designation Agreement - Part 1

andrewsignori-aot:
- This PR includes only the basic part of a designation agreement form submission by the institution. Manage designations page has just a button to open the designation agreement form for now.
- Form.io definition will b added in a upcoming PR. The work in progress is available on DEV.
- Create migrations to sabe designations and designations locations
- Created API to create designation and retrieve one designation by id
- Created a reusable designation agreement form that will b later used for visualization for Ministry/Institution.
- Added institution information to Vue store to use in the designation header form. Below information is coming from Vue store that is updated after login or when the Institution Profile is changed.

image

- Created a method to allow the use of the hyperlink in the form.io definition that can cause redirects in the Vue application. We had some technical challenges in the past to achieve the same and I believe that this is a good way to deliver Figma design.

image

**BC Private Designation Request**

image

**Non BC Private Designation**

image


sources/packages/api/src-sql/DesignationAgreements/Create-designation-agreements-table.sql
```diff
@@ -0,0 +1,56 @@
+CREATE TABLE IF NOT EXISTS sims.designation_agreements (
+  id SERIAL PRIMARY KEY,
+  institution_id INT REFERENCES sims.institutions(id) ON DELETE CASCADE NOT NULL,
```
dheepak-aot, line 3: ON DELETE CASCADE 👍

sources/packages/api/src/database/entities/designation-agreement-location.model.ts
```diff
@@ -0,0 +1,71 @@
...
+      designationAgreementLocation.designationAgreement,
+  )
+  designationAgreementId: number;
+  /**
+   * Designation agreement.
+   */
+  @ManyToOne(
+    () => DesignationAgreement,
+    (designation) => designation.designationAgreementLocations,
+    { eager: false, orphanedRowAction: "delete" },
```
ann-aot, line 39: i would like to check whether `orphanedRowAction: "delete"` is different from `onDelete:CASCADE`

sources/packages/api/src/database/entities/designation-agreement.model.ts
```diff
@@ -0,0 +1,134 @@
...
+    nullable: true,
+  })
+  endDate?: Date;
+  /**
+   * Ministry user that approved or declined the designation agreement.
+   */
+  @ManyToOne(() => User, { eager: false, cascade: false })
+  @JoinColumn({
+    name: "assessed_by",
+    referencedColumnName: ColumnNames.ID,
```
ann-aot, line 108: `nullable: true`?

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+  constructor(
+    private readonly designationAgreementService: DesignationAgreementService,
+    private readonly formService: FormService,
+  ) {}
+
+  /**
+   * Initiates a new designation agreement request. This action
+   * is meant to me initiated by the institution signing officer
+   * for further assessment of the Ministry.
+   * @param institutionId institution id requesting the designation.
```
ann-aot, line 39: i think `institutionId` is not a parameter here

sources/packages/api/src-sql/Types/Create-designation-agreement-status.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TYPE sims.designation_agreement_status
+  AS ENUM ('Pending', 'Approved', 'Declined');
```
dheepak-aot, line 2: What does pending status represent? does it mean the designation requested? or not even requested?

sources/packages/api/src/database/entities/designation-agreement-location.model.ts
andrewsignori-aot: Apparently yes @ann-aot. For this relationship when child is removed from the parent (Designation in this case), the default behavior was try to set "null" to the designation_agreement_id on the child (designations_locations in this case), so it was "removing" the relationship but not the record. Checking the documentation, the orphanedRowAction was added in a version above the one that we were using (that why there is a typeorm version upgrade in this PR).

sources/packages/api/src/database/entities/designation-agreement.model.ts
andrewsignori-aot: Apparently we do not have this option for @JoinColumn decorator.
Were you able to do it in any place so far in this solution?

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
andrewsignori-aot: Right, I will remove it!

sources/packages/api/src-sql/Types/Create-designation-agreement-status.sql
andrewsignori-aot: It means that it was requested and it is pending from Ministry review.
In the enum there are further explanation.
image

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Retrieve the designation agreement information and
+   * the associated locations approvals.
+   * @param designationId designation id.
+   * @param institutionId institution id.
+   * @returns designation agreement information.
+   */
+  async getDesignationById(
```
ann-aot, line 73: IMO,  here designation results are dependent upon both `designationId` and `institutionId`, I think it will be to rename the function. because when I see `getDesignationById`, I think it should search only with `designationId`

sources/packages/api/src/database/entities/designation-agreement.model.ts
```diff
@@ -0,0 +1,134 @@
...
+  institution: Institution;
+  /**
+   * Dynamic data that represents the designation agreement requested by the Institution to be approved by the Ministry.
+   */
+  @Column({
+    name: "submitted_data",
+    type: "jsonb",
+    nullable: false,
+  })
+  submittedData: any;
```
dheepak-aot, line 51: Can we define an interface to avoid any ?

sources/packages/web/src/components/common/DesignationAgreement/DesignationAgreementForm.vue
```diff
@@ -0,0 +1,58 @@
...
+  },
+  setup(props: any, context: SetupContext) {
+    const router = useRouter();
+    // Regular hyperlinks ids presents inside the form.io definition that
+    // needs execute a redirect in the vue application.
+    const MANAGE_LOCATIONS_LINK = "goToManageLocations";
+    const MANAGE_USERS_LINK = "goToManageUsers";
+
+    const formLoaded = async () => {
+      RouteHelper.AssociateHyperlinkClick(MANAGE_LOCATIONS_LINK, () =>
```
ann-aot, line 39: (y)

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
dheepak-aot, line 39: There is no param institutionId here. Instead payload is there.

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+   * @returns the new designation agreement id created.
+   */
+  @IsInstitutionAdmin()
+  @Post()
+  async submitDesignationAgreement(
+    @UserToken() userToken: IInstitutionUserToken,
+    @Body() payload: SubmitDesignationAgreementDto,
+  ) {
+    // Validates if the user has the right role.
+    const isLegalSigningAuthority = userToken.authorizations.hasAdminRole(
```
dheepak-aot, line 49: isLegalSigningAuthority is check only on this API alone. Otherwise a decorator would be nice here IMO.

sources/packages/web/src/services/http/DesignationAgreementApi.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Http API client for Designation agreements.
+ */
+export class DesignationAgreementApi extends HttpBaseClient {
+  async submitDesignationAgreement(
+    designationAgreement: SubmitDesignationAgreementDto,
+  ): Promise<void> {
+    try {
+      await this.apiClient.post(
```
ann-aot, line 15: we have `.postCall()`
image

sources/packages/web/src/views/institution/designations/DesignationRequest.vue
```diff
@@ -0,0 +1,107 @@
...
+    const toastMessage = useToastMessage();
+    const formatter = useFormatters();
+    const { institutionState } = useInstitutionState();
+    const {
+      userFullName,
+      userEmail,
+      isLegalSigningAuthority,
+    } = useInstitutionAuth();
+
+    const designationModel = reactive({} as DesignationModel);
```
ann-aot, line 52: 👍

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: The institution id is included for authorization purpose only. The method is really to get the designation by its Id.
I would avoid renaming it to something like getDesignationByIdAndInstitution.
Maybe we can call it getInstitutionDesignationById? Any thoughts?

sources/packages/web/src/views/institution/designations/DesignationRequest.vue
```diff
@@ -0,0 +1,107 @@
...
+      try {
+        await DesignationAgreementService.shared.submitDesignationAgreement({
+          dynamicData: model.dynamicData,
+          locations: model.locations.map(location => ({
+            locationId: location.locationId,
+            requestForDesignation: location.requestForDesignation,
+          })),
+        } as SubmitDesignationAgreementDto);
+        toastMessage.success("Submitted", "Designation agreement submitted.");
+        await router.push({ name: InstitutionRoutesConst.MANAGE_DESIGNATION });
```
ann-aot, line 92: do we need to await here?

ann-aot (commented):
LGTM 👍 Good Work, added some minor commnets

sources/packages/api/src/database/entities/designation-agreement.model.ts
andrewsignori-aot: The data is going to be the result of an area in the form.io that is not required to have any constraints and could be freely changed by the Ministry. The data under this particular section is also not supposed to be used anywhere in the API, only to populate the same form again for visualization, so, I do not see a reason for this particular case.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
ann-aot: ya `getInstitutionDesignationById` make sense to me

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
andrewsignori-aot: I will remove it 😉

sources/packages/api/src/route-controllers/designation-agreement/models/designation-agreement.model.ts
```diff
@@ -0,0 +1,34 @@
+import { DesignationAgreementStatus } from "../../../database/entities";
+
+/**
+ * This DTO contains dynamic data that must
+ * be validated by the form.io dryrun validation.
+ */
+export interface SubmitDesignationAgreementDto {
+  dynamicData: any;
```
dheepak-aot, line 8: Is it the form.io data? we can have an interface here?

sources/packages/api/src/database/entities/designation-agreement.model.ts
ann-aot: maybe we can do something like 
image

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
andrewsignori-aot: I thought about the decorator and decided to use the "inline" check because apparently there will be no other endpoint that requires the same check. The only purpose of the LegalSigningAuthority is to execute this submission, all other access for visualization, for instance, are for any Admin. If there is any scenario in the future that would required the same check maybe we can create one.
Every time that we introduce a new decorator globally it will be tested for every single request for every single endpoint.
So, I can still create one if that is a Team decision 😉

sources/packages/api/src/database/entities/designation-agreement-location.model.ts
ann-aot: okay got it 👍 Thanks for the explaination

sources/packages/web/src/views/institution/designations/DesignationRequest.vue
andrewsignori-aot: Because it is the last instruction in the method we do not, It will not change the behavior.

sources/packages/api/src/route-controllers/designation-agreement/models/designation-agreement.model.ts
andrewsignori-aot: The data is going to be the result of an area in the form.io that is not required to have any constraints and could be freely changed by the Ministry. The data under this particular section is also not supposed to be used anywhere in the API, only to populate the same form again for visualization, so, I do not see a reason for this particular case.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @param institutionId institution id requesting the designation.
+   * @param submittedData dynamic data that represents the designation requested.
+   * @param submittedByUserId institution user submitting the designation.
+   * @param submittedDate date that the institution user submitted the designation.
+   * @param requestedLocationsIds ids of the locations requested to be designated.
+   * @returns the new designation agreement created.
+   */
+  async submitDesignationAgreement(
+    institutionId: number,
+    submittedData: any,
```
dheepak-aot, line 39: when I see any, I assume it is because this is a part of PR. Please correct me If I am wrong.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
dheepak-aot: I agree. We are using the same way in many place like applicationId and studentId|userId .

sources/packages/web/src/components/generic/formio.vue
```diff
@@ -46,12 +46,16 @@ export default {
...
       let cachedFormDefinition: string | null = null;
-      try {
-        // Try to load the definition from the session storage.
-        cachedFormDefinition = sessionStorage.getItem(props.formName);
-      } catch {
-        // No action needed. In case of failure it will load the form from the server
-        // in the same way as it is the first time load.
+      // Avoid caching during development to allow that the changes
+      // on form.io definitions have effect immediately.
+      if (process.env.NODE_ENV !== "development") {
```
dheepak-aot, line 51: Good thinking! 👍

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: This is not because it is part of the PR. Part of the reason that the Ministry is using form.io is because there are some forms that are flexible in a way that they could be changed in many ways without impacting the application or process at all. The area of the form here defined as "any" does not have a fixed structure.

The data is going to be the result of an area in the form.io that is not required to have any constraints and could be freely changed by the Ministry. The data under this particular section is also not supposed to be used anywhere in the API, only to populate the same form again for visualization, so, I do not see a reason for this particular case.

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
dheepak-aot: My point is we need decorator only if LegalSigningAuthority is needed for any other API in future which can be done if the requirement comes. At present what you have done is good.

sources/packages/web/src/components/common/DesignationAgreement/DesignationAgreementForm.models.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Used do define conditions to change the form.io, for instance,
+ * some components visibility.
+ */
+export enum DesignationFormViewModes {
+  submission = "submission",
+  viewOnly = "viewOnly",
```
dheepak-aot, line 7: does it mean ReadOnly?

dheepak-aot:
Good work @andrewsignori-aot . Just few minor comments.

sources/packages/web/src/components/common/DesignationAgreement/DesignationAgreementForm.models.ts
andrewsignori-aot: Kind off 😄 
I did that on purpose because when the form goes to readOnly/viewOnly it receives new items in the UI, like the status on the header and a new column for approval status. So, since it it was not just about making the components readOnly I try to use a different name.

sources/packages/api/src/database/entities/designation-agreement.model.ts
andrewsignori-aot: Sure, I will change it, I was looking in the wrong decorator 😃

sources/packages/api/src/database/entities/designation-agreement.model.ts
andrewsignori-aot: @dheepak-aot, I added more comments to the property because developers in the future can have the same quetion. Hope that helps 😉 
image

ann-aot (approved):
LGTM 👍 Thanks for addressing the comments 👍

sources/packages/web/src/components/common/DesignationAgreement/DesignationAgreementForm.models.ts
dheepak-aot: Makes sense.

dheepak-aot (approved):
LGTM @andrewsignori-aot 👍
Thanks for making the changes.

sources/packages/api/src/database/entities/designation-agreement.model.ts
```diff
@@ -0,0 +1,137 @@
...
+  Institution,
+  User,
+} from ".";
+import { ColumnNames, TableNames } from "../constant";
+import { dateOnlyTransformer } from "../transformers/date-only.transformer";
+import { RecordDataModel } from "./record.model";
+
+/**
+ * Designation agreement created by the institution for the Ministry assessment.
+ * The Ministry will check the data provide and approve of deny the request,
```
guru-aot, line 22: please change it to "provided and approve or deny the request"

guru-aot (approved):

andrewsignori-aot:
Because this one is a dependency to start the next part and I got the approvals from @ann-aot, @guru-aot and @dheepak-aot I will merge it. @henryoforeh-aot, please feel free to add comments here. We can adjust in the upcoming PR 😉

PR #997, opened 2022-02-09
[web][api][SIMS #406]Designation Agreement Summary and View

andrewsignori-aot:
image

image


sources/packages/web/src/components/generic/BodyHeader.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <div class="mb-4">
+    <span class="category-header-large color-blue">
+      {{ fullTitleMessage }}
+    </span>
+    <div class="float-right">
```
ann-aot, line 6: in summary screens. we do have a search box too right? so, are we planning to create a slot for the search box or it will be inside the `button` slot?

sources/packages/web/src/components/generic/BodyHeader.vue
```diff
@@ -0,0 +1,42 @@
...
+    },
+    recordsCount: {
+      type: Number,
+      required: false,
+    },
+  },
+  setup(props: any) {
+    const fullTitleMessage = computed(() => {
+      if (props.recordsCount) {
+        return `${props.title} (${props.recordsCount})`;
```
ann-aot, line 35: I see, `recordsCount` is not required (line 29), so if `recordsCount` is not passed then this will return `title (undefined)` right?

ann-aot: also. for body headers like this 
image
there is no count, so we don't want the brackets themselves. right?

sources/packages/web/src/components/generic/StatusChip.vue
```diff
@@ -0,0 +1,78 @@
...
+        class="mr-1"
+        :color="iconColor"
+      />
+      <span class="text-uppercase">{{ label ?? status }}</span>
+    </template>
+  </v-badge>
+</template>
+<script lang="ts">
+import { computed } from "vue";
+import { StatusShipTypes } from "@/components/generic/StatusChip.models";
```
ann-aot, line 15: Ship? did u mean chip?

sources/packages/web/src/components/generic/BodyHeader.vue
andrewsignori-aot: Good question. If the search is present frequently if can have it even here.
For now the idea was just create something to start. Maybe we can rename it to a more generic name. What do you think?

sources/packages/web/src/components/generic/StatusChip.vue
andrewsignori-aot: It is supposed to be StatusChipTypes. I will rename it.

sources/packages/web/src/composables/useDesignationAgreement.ts
```diff
@@ -0,0 +1,21 @@
+import { StatusShipTypes } from "@/components/generic/StatusChip.models";
+import { DesignationAgreementStatus } from "@/types/contracts/DesignationAgreementContract";
+
+export function useDesignationAgreement() {
+  const mapDesignationChipStatus = (
+    status: DesignationAgreementStatus,
```
ann-aot, line 6: mm, I do not understand the reason for this mapping. can we sync up?

sources/packages/web/src/components/generic/StatusChip.vue
andrewsignori-aot: Fixed!

sources/packages/web/src/composables/useDesignationAgreement.ts
andrewsignori-aot: The idea is that the StatusChipTypes is generic in the way that for different statuses (Designation, Restriction, COE, etc.) green status (StatusChipTypes.success) could be mapped to whatever is needed for the particular Designation, Restriction, COE, etc, status. We can sync up on this.

sources/packages/web/src/components/generic/EmptyTableTemplate.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div v-if="isEmpty">
```
ann-aot, line 2: since you have created this generic component, just a minor question, are we excepting more future requirements using this component?

sources/packages/web/src/components/generic/BodyHeader.vue
andrewsignori-aot: Slot renamed to _actions_, as per our conversation 😉

sources/packages/web/src/components/generic/EmptyTableTemplate.vue
andrewsignori-aot: As per my conversation with Lynn the idea is to have an empty template to display with DataTables.
Using this component it will be just a matter to copy the image from Figma and consume it.

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
```diff
@@ -114,4 +127,32 @@ export class DesignationAgreementController {
       ),
     } as GetDesignationAgreementDto;
   }
+
+  /**
+   * Get the list of all the designations that belongs to
+   * the institution user currently authenticated.
+   * @param userToken user authentication token.
+   * @returns the list of all the designations that
+   * belongs to one the institution.
```
henryoforeh-aot, line 136: Did you mean to write "belongs to one of the institutions" here ?

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -96,4 +96,49 @@ export class DesignationAgreementService extends RecordDataModelService<Designat
...
+        "designation.endDate",
+      ])
+      .where("designation.institution.id = :institutionId", { institutionId })
+      .orderBy(
+        `CASE designation.designationStatus
+            WHEN '${DesignationAgreementStatus.Pending}' THEN 1
+            WHEN '${DesignationAgreementStatus.Approved}' THEN 2
+            WHEN '${DesignationAgreementStatus.Declined}' THEN 3
+            ELSE 4
+          END`,
```
henryoforeh-aot, line 124: Is it possible for 1, 2, 3 and 4 to be Enums so that it will a bit more meaningful ?

sources/packages/web/src/components/partial-view/DesignationAgreement/DesignationAgreementSummary.vue
```diff
@@ -0,0 +1,100 @@
...
+      emptyMessage="You don’t have any agreements yet"
+    >
+      <template #image>
+        <v-img
+          height="200"
+          alt="You don’t have any agreements yet"
+          src="@/assets/images/designation_summary.svg"
+        />
+      </template>
+      <DataTable
```
ann-aot, line 14: should this `<datatable>`  be inside `<empty-table-template></empty-table-template>`?

sources/packages/web/src/services/http/common/HttpBaseClient.ts
```diff
@@ -39,6 +39,19 @@ export default abstract class HttpBaseClient {
     }
   }
 
+  protected async getCallTyped<T>(url: string, authHeader?: any): Promise<T> {
```
ann-aot, line 42: 👍

sources/packages/web/src/views/institution/designations/DesignationView.vue
```diff
@@ -0,0 +1,81 @@
...
+    },
+  },
+  setup(props: any) {
+    const { institutionState } = useInstitutionState();
+    const formatter = useFormatters();
+    const { mapDesignationChipStatus } = useDesignationAgreement();
+
+    const designationModel = reactive({} as DesignationModel);
+    designationModel.institutionName = institutionState.value.operatingName;
+    designationModel.institutionType = institutionState.value.institutionType;
```
ann-aot, line 50: if any of the store values like `operatingName` is changed, then will it reflect in `designationModel`. how about using a computed field?

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
andrewsignori-aot: It is supposed to be "the list of all the designations that belongs to one institution."
I updated it 👍

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: In this case the numbers are supposed to be really a number that will be used to imposed the proper order to the query results. What the enum would look like? Something like below? @guru-aot, @ann-aot, @dheepak-aot, do you believe that it would be readable?
```javascript
enum SortPriority {
    Priority1 = 1,
    Priority2 = 2,
    Priority3 = 3,
    Priority4 = 4,

WHEN '${DesignationAgreementStatus.Declined}' THEN ${SortPriority.Priority1}
}
```

sources/packages/web/src/components/partial-view/DesignationAgreement/DesignationAgreementSummary.vue
andrewsignori-aot: It is actually inside _empty-table-template_ in the default slot. I am not sure it I am answering the question properly, please let me know.

sources/packages/web/src/assets/css/formio-shared.scss
```diff
@@ -303,6 +304,51 @@
...
+/* form.io status chip */
+.formio-status-chip {
+  border-width: 1px;
+  padding-left: 15px;
+  padding-right: 10px;
+  position: absolute;
+  border-radius: 30px;
+  border-style: solid;
+  text-transform: uppercase;
+  font-weight: 700 !important;
```
henryoforeh-aot, line 316: There is a variable called $formio-header-wt  in the global-style-variable.scss file which can be used here.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -96,4 +96,49 @@ export class DesignationAgreementService extends RecordDataModelService<Designat
...
+      )
+      .addOrderBy("designation.submittedDate")
```
guru-aot, line 126: Dont we need to add the submittedDate before the DesignationAgreementStatus? It sorts the designation list by the submittedDate first and then by the DesignationAgreementStatus. So in a way the latest list of pending designation agreement will be shown and not a mix of DesignationAgreementStatus in submitted date order.

sources/packages/api/src/utilities/date-utils.ts
```diff
@@ -115,6 +115,18 @@ export function getDateOnlyFromFormat(
   return getDateOnly(isoDate);
 }
 
+/**
+ * Get the date only part of a date/time object.
+ * @param date date/time to have the date extracted.
+ * @returns date only string in ISO format YYYY-MM-DD.
+ */
+export function getDateOnlyString(date?: Date): string | null {
```
guru-aot, line 123: getDateOnlyFormat method returns the format YYYY MM DD, can we please rename the method name to something that defines the ISO format? like getISODateOnlyString? or something?

ann-aot (commented):
LGTM 👍  added some comments

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
ann-aot: this looks fine to me

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: The requirement is to have the list ordered by Status and for the group of status order by the submission date.
I did not follow the comment entirely but the idea is to show all Pending (only one always), then all the Approved ones and last all the Declined ones. Am I missing something here?

sources/packages/web/src/services/http/common/HttpBaseClient.ts
```diff
@@ -39,6 +39,19 @@ export default abstract class HttpBaseClient {
     }
   }
 
+  protected async getCallTyped<T>(url: string, authHeader?: any): Promise<T> {
+    try {
+      const response = await this.apiClient.get(
+        url,
+        authHeader ?? this.addAuthHeader(),
+      );
```
henryoforeh-aot, line 47: Is it possible to explore the possibility of implementing an interceptor to inject the required auth header(s) instead of passing it as an argument in this generic function?

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
guru-aot: Thanks for the demo looks good, please share the sql if you can

andrewsignori-aot: Sharing the SQL as we talked. Just added the "DESC" in the end that was missing.
image

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: I updated the code per the enum that I shared above. Please let me know if it is ok @ann-aot, @henryoforeh-aot, @dheepak-aot, @guru-aot.

sources/packages/web/src/components/partial-view/DesignationAgreement/DesignationAgreementSummary.vue
andrewsignori-aot: Component renamed to _toggle-content_ and adjusted. Please check and lets discuss as needed.

sources/packages/api/src/utilities/date-utils.ts
andrewsignori-aot: Method renamed.

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
```diff
@@ -16,9 +17,10 @@ import {
   UserToken,
 } from "../../auth/decorators";
 import { DesignationAgreementService, FormService } from "../../services";
-import { getUTCNow } from "../../utilities";
+import { getDateOnlyString, getUTCNow } from "../../utilities";
 import {
   GetDesignationAgreementDto,
+  GetDesignationAgreementsDto,
```
dheepak-aot, line 23: IMO, while naming the DTOs we generally avoid Get and use them commonly for method names.

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
```diff
@@ -63,6 +65,17 @@ export class DesignationAgreementController {
         "Not able to create a designation agreement due to an invalid request.",
       );
     }
+
+    if (
+      await this.designationAgreementService.hasPendingDesignation(
```
dheepak-aot, line 70: Just a suggestion. instead of making service call inside if, it can be assigned to a const and the const can be used in if statement.

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
andrewsignori-aot: Actually the original idea was express in the DTOs the purpose input/output. Usually the input ones requires some class validators and will require to be a class instead of an interface, hence having the same name will be tricky.
We have quite a few starting with Get. Thoughts?

sources/packages/api/src/route-controllers/designation-agreement/designation-agreement.controller.ts
andrewsignori-aot: We can if we decided to go in this way. That was my first choice and I avoided because I though that there will a comment against it 😄 
I will create the const 👍

andrewsignori-aot: Changed 👍

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
dheepak-aot: Sorry for being late to reply on this. 
IMO as per enum, when we declare enum as 

enum SortPriority {
    Priority1,
    Priority2,
    Priority3,
    Priority4,
}

 the value is automatically set as 1,2,3,4 for Priority1 .. Priority4 respectively. 

Also as we know the enum is for Priority we can declare like

enum SortPriority {
    ONE,
    TWO,
    THREE,
    FOUR,
}

Here comes another thought. If the order of enum items declared in DB is same as their priority  then we can avoid this WHEN and THEN. I know it won't be possible always with all the enums. So in exceptional cases SortPriority enum can be used,

sources/packages/web/src/assets/css/formio-shared.scss
andrewsignori-aot: Renamed the variable $formio-header-wt to $strong-font-weight and consumed everywhere.

henryoforeh-aot (approved):
Looks good, I will approve.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: I got the idea about ONE, TWO, THREE, FOUR but to me it is going to far 😄 
To avoid using numbers we are converting numbers to its written format. It does not make much sense to me.
Yes, we should avoid "magic numbers" in the code, but is it the case for this specific one?

dheepak-aot: ONE, TWO, THREE, FOUR is just a convention which came to my mind. Naming is up to the dev to decide. But my main opinion is not assigning numbers. to enum and leverage it's own benefit.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -96,4 +97,49 @@ export class DesignationAgreementService extends RecordDataModelService<Designat
...
+
+  /**
+   * Verify when the institution already have a pending designation
+   * agreement. Institutions are not supposed to have more than one
+   * pending designation at the same time.
+   * @param institutionId institution to be verified.
+   * @returns true, if there is already a pending designation agreement.
+   */
+  async hasPendingDesignation(institutionId: number): Promise<boolean> {
+    const pendingDesignationsCount = await this.repo.count({
```
dheepak-aot, line 139: As per our discussion, select 1 can be used.

sources/packages/web/src/components/generic/StatusChip.vue
```diff
@@ -0,0 +1,78 @@
...
+    const iconColor = computed(() => {
+      switch (props.status) {
+        case StatusChipTypes.Success:
+          return COLOR_BANNER_SUCCESS;
+        case StatusChipTypes.Warning:
+          return COLOR_BANNER_WARNING;
+        case StatusChipTypes.Error:
+          return COLOR_BANNER_ERROR;
+        default:
+          return COLOR_BANNER_ERROR;
```
dheepak-aot, line 45: I see at line 58 that default status is inactive, hence for inactive status icon color is black.

sources/packages/api/src/services/designation-agreement/designation-agreement.service.ts
andrewsignori-aot: Query adjusted to check if there is one record at least, avoid count the records.

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -12,8 +12,9 @@ import ActiveApplicationEdit from "@/views/institution/locations/active-applicat
 import LocationCOESummary from "@/views/institution/locations/confirmation-of-enrollment/LocationCOESummary.vue";
 import AddInstitutionLocation from "@/views/institution/AddInstitutionLocation.vue";
 import EditInstitutionLocation from "@/views/institution/EditInstitutionLocation.vue";
-import ManageDesignation from "../views/institution/designations/ManageDesignation.vue";
+import ManageDesignations from "../views/institution/designations/ManageDesignations.vue";
 import DesignationRequest from "../views/institution/designations/DesignationRequest.vue";
+import DesignationView from "../views/institution/designations/DesignationView.vue";
```
dheepak-aot, line 17: I see @/ and ../ imports, we can make it consistent.

dheepak-aot:
Nice and Clean work @andrewsignori-aot . Appreciate the additional efforts taken to make the components 👏
Some minor comments.

dheepak-aot (approved):
Good work @andrewsignori-aot 👍Thanks for doing the changes.

Issue #870, opened 2021-11-23
Institution Views Welcome Page Content


michesmi:
UX/UI Ticket #888
**Acceptance Criteria**
- [x] Build the welcome page, static content only
Hyperlink URL: https://studentaidbc.ca/help-centre

PR #1016, opened 2022-02-14
[web][SIMS #870] Institution views welcome page

henryoforeh-aot:
I created the new welcome page according to figma design for institution using form io

institution_welcome1
institution_welcome2
institution_welcome3


sources/packages/web/src/assets/css/global-style-variables.scss
```diff
@@ -41,6 +41,8 @@ $line-height-header-sm: 22px;
 $formio-content: 16px;
 $formio-content-lh: 24px;
 $formio-header-margin: 0.5rem;
+$formio-header-wt: 700;
```
andrewsignori-aot, line 44: Please use the variable $font-weight-bold, it is equivalent for 700.

sources/packages/web/src/assets/css/global-style-variables.scss
```diff
@@ -41,6 +41,8 @@ $line-height-header-sm: 22px;
...
+$formio-category-header-wt: 800;
```
andrewsignori-aot, line 45: Is it really needed? It it affecting the final UI?

sources/packages/web/src/assets/css/formio-shared.scss
```diff
@@ -352,3 +360,14 @@
 .form-check-input {
   position: inherit;
 }
+
+.need-help-image {
```
andrewsignori-aot, line 364: This content page is a more generic page that could be completely changed by the Ministry.
I would avoid adding so specific styles dedicated to a so specific purpose here.
Maybe we can create just a relative class here and deal with the spacing directly in the form.io using some bootstrap classes already available.
Same for welcome-image style.

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
```diff
@@ -0,0 +1,3175 @@
...
+                            "tag": "div",
+                            "attrs": [
+                                {
+                                    "attr": "",
+                                    "value": ""
+                                }
+                            ],
+                            "content": "Welcome to your institution account!",
+                            "refreshOnChange": false,
+                            "customClass": "category-header-large primary-color mt-5 mb-5",
```
andrewsignori-aot: Maybe you the html is changed to a "<p>" there is no need for the "mt-5 mb-5".
I believe that other forms are using more <p> in general.

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
```diff
@@ -0,0 +1,3175 @@
...
...cgMzIzLjI2MiA4Mi41ODQ2IDMyMy4zOTggODAuNzkzMkMzMjMuNDYxIDc5LjgxOTcgMzIzLjQ2MSA3OC44NDM2IDMyMy4zOTggNzcuODcwMUMzMjMuMzUgNzcuMzE3NiAzMjMuMjY5IDc2Ljc2ODEgMzIzLjE1MyA3Ni4yMjQzQzMyMy4wNzMgNzUuOTIzNCAzMjMuMDE0IDc1LjY0MDkgMzIyLjg4OCA3NS4yOTdMMzIyLjcxNiA3NC44MThMMzIyLjU3NyA3NC40OTg3QzMyMS4xMzkgNzEuMTQ3MyAzMTkuMzIyIDY3Ljk0NjMgMzE3LjE1OCA2NC45NDk0QzMxNi4xMTEgNjMuNDc1NiAzMTUuMDExIDYyLjA1MDkgMzEzLjg0NSA2MC42NTA3QzMxMy4yNjkgNTkuOTU2OCAzMTIuNjc5IDU5LjI3NTIgMzEyLjA3NiA1OC41OTM1QzMxMS40NzQgNTcuOTExOSAzMTAuODc3IDU3LjI2NyAzMTAuMTg4IDU2LjU1NDdMMzA0LjMxOSA2MS40MDYxWiIgZmlsbD0iI0ZGQjU3MyIvPgo8cGF0aCBkPSJNMzA5LjIwMiA1NC4wNDRDMzE0LjEzIDU0LjY5NDkgMzE5LjQ3NyA2NS40NTM5IDMxOS40NzcgNjUuNDUzOUwzMTAuMTQ5IDczLjYzMzhDMzEwLjE0OSA3My42MzM4IDMwNi4yOCA2OC4yMjM1IDMwMy4wOTQgNjIuOTczQzI5OS42MjkgNTcuMjU1NyAzMDMuMjM5IDUzLjI1NzkgMzA5LjIwMiA1NC4wNDRaIiBmaWxsPSIjNDA3QkZGIi8+CjxwYXRoIGQ9Ik0zMTYuMzU2IDk0LjMyMDNMMzEzLjIzNiA5OC4xNDYyTDMxOS43MjEgOTkuNjAxNkMzMTkuNzIxIDk5LjYwMTYgMzIxLjU3NiA5NS43NzU3IDMxOS40NTcgOTQuNTY1OUwzMTYuMzU2IDk0LjMyMDNaIiBmaWxsPSIjRkZCNTczIi8+CjxwYXRoIGQ9Ik0zMTMuMzQzIDEwMi4xNTlMMzE4Ljc1NSAxMDMuMTI5TDMxOS43MjMgOTkuNjAzOUwzMTMuMjM3IDk4LjE0ODRMMzEzLjM0MyAxMDIuMTU5WiIgZmlsbD0iI0ZGQjU3MyIvPgo8L3N2Zz4K\"></figure>",
```
andrewsignori-aot: I would say that this will increase the content of the json payload to load the form definition and this is cached only daily. I would say that this is a wonderful solution for the future if some change is needed, but now I would prefer having this image in Vue source in the same way the student application is doing.
@ann-aot @guru-aot @dheepak-aot thoughts?

andrewsignori-aot (commented):
The page looks really great, awesome work 👍 

sources/packages/web/src/assets/css/formio-shared.scss
```diff
@@ -45,6 +45,14 @@
   text-transform: uppercase;
 }
 
+.category-header-small-no-transform {
+  font-style: $font-style-normal;
+  font-weight: $formio-category-header-wt;
+  font-size: $formio-content;
+  line-height: $line-height-header-sm;
+  letter-spacing: 0.5px;
+}
```
dheepak-aot, line 54: If getting rid of  text-transform: uppercase from category header small is the only intention here then you can use @extend from scss.

e.g 
_.category-header-small-no-transform {
@extend .category-header-small;
text-transform: none;
}_

sources/packages/web/src/assets/css/global-style-variables.scss
dheepak-aot: I see why @henryoforeh-aot used weight 800. Figma used 800 as weight for bold in the design system. 
image
 @henryoforeh-aot Can you please check with Lynn to see if we use font-weight 800 for bold and not 700.

dheepak-aot:
LGTM. Just 2 comments @henryoforeh-aot

sources/packages/web/src/assets/css/global-style-variables.scss
andrewsignori-aot: @dheepak-aot I believe that the BCSans used in the project does not even support 800, that is why I asked if setting 800 was causing some impact. If BC Sans does not support it and if it is not causing an impact in the UI we will not be using it, right?

henryoforeh-aot: I have checked with Lynn, we will not use 800 for font-weight. Thanks for the clarification.

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
```diff
@@ -0,0 +1,3175 @@
+{
+    "forms": [
+        {
```
guru-aot: Can we please remove the forms tag and copy-paste only the array inside it? For example, please check the other formio json.

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
```diff
@@ -0,0 +1,3175 @@
...
+                            "unique": false,
+                            "persistent": false,
+                            "clearOnHide": true,
+                            "refreshOn": "",
+                            "redrawOn": "",
+                            "dataGridLabel": false,
+                            "labelPosition": "top",
+                            "description": "",
+                            "errorLabel": "",
+                            "toolt
```
guru-aot: I believe we are not using this hidden content "applicationStatus" in our form, can we please remove it?

sources/packages/web/src/assets/css/formio-shared.scss
```diff
@@ -27,7 +27,7 @@
 }
 
 .secondary-color-light {
-  color: $secondary-header-color-lt;
+  color: $secondary-header-color-lt !important;
```
ann-aot, line 30: do we really need `!important`?

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
```diff
@@ -0,0 +1,3175 @@
...
+                                {
+                                    "components": [
+                                        {
+                                            "title": "",
+                                            "labelWidth": "",
+                                            "labelMargin": "",
+                                            "theme": "default",
+                                            "tooltip": "",
+                                            "customClass": "formio-institution-get-started-panel",
+                                            "collapsible": false,
```
ann-aot: mm..maybe we can give some meaning full, generic name for the custom class, most probably we  will need to reuse this class in part of application like a welcome page of aest or any other user

ann-aot: also, Iam not seeing this custom class definition, did u forget to add it?

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
ann-aot: actually, I kind of like this , as you said, it gives the user flexibility to change without code change.

ann-aot (commented):
just 2 minor comments @henryoforeh-aot . Good Work 👍

sources/packages/web/src/assets/css/global-style-variables.scss
dheepak-aot: @andrewsignori-aot Yes. I didn't have a look at BCSans while making the comment. @henryoforeh-aot thanks for checking with Lynn.

sources/packages/web/src/assets/css/formio-shared.scss
henryoforeh-aot: Yes, the font color does not reflect on the formio side without the !important property

sources/packages/forms/src/form-definitions/institutionwelcomepage.json
andrewsignori-aot: So, is it better to have the images being loaded every time as a part of a big request from the server every time?
I would say that this should be used as a second option, IMO.

dheepak-aot: Considering that the image loads only once daily and the image is considerably light, I would say benefits outweigh the risk. So option one IMO.

guru-aot: I feel like having this image in Vue source, like the previous screens, to be consistent.

ann-aot: maybe we can have a quick up and finalize this after the demo, maybe I am missing something

andrewsignori-aot: @dheepak-aot, when I said that the page loads once a day I was just being too right level.
The form definition (right now with 310Kb) will load every time that a user loads the Institution Portal in a new Tab The form size will grow more if more images are embedded.
From the user perspective there will no much impact. From our server perspective it will be downloading every time the images because they are embedded. The form without the images is something around 126KB, 40% of the size.

Lets say that this is not a Institution User portal, are we going to use embedded images for the welcome page on Student Portal, for instance? @dheepak-aot @ann-aot 

The ideal scenario, IMO, would be the Ministry offer a place that we can dump the images and just add the reference to the image in the form.io definition. Since we do not have it right now, we were adding to Vue source.

sources/packages/web/src/views/institution/InstitutionDashboard.vue
```diff
@@ -11,4 +11,22 @@ export default {
 };
 </script>
 
-<style></style>
+<style>
+.img-welcome-background {
```
ann-aot, line 15: mm, can we move the styles to .scss file?

andrewsignori-aot (approved):

ann-aot (approved):
LGTM 👍 Good Work @henryoforeh-aot  👏  Thanks for doing the changes

guru-aot (approved):
LGTM, thanks for doing the changes @henryoforeh-aot

dheepak-aot (approved):
LGTM. Good work @henryoforeh-aot

Issue #894, opened 2021-11-29
Ministry user Approve / Deny Education Program


michesmi:
**Additional Context**
- Database changes are required

**Acceptance Criteria**
- As a ministry user when viewing PENDING program:
- [x] Should display "Approve" "Decline" buttons
- [x] Submitted date should populate
- [x] Approved date should populate
- [x] Declined date should populate
- [x] Effective End date should populate
- [x] We need to include approve by/declined by/submitted by
- As a ministry user when clicking APPROVE PENDING program:
- [x] Display Approve program" Modal
- [x] include Effective end date (this is for data collection purposes only, does not have any associated business rules)
- [x] Include Notes, free text up to 500 characters (confirm same length as other note components)
- [x] Notes should display under notes page with program filter working
- As a ministry user when clicking DENY PENDING program:
- [x] Include Notes, free text up to 500 characters (confirm same length as other note components)
- [x] Notes should display under notes page with program filter working

IrenaMilosevic1:
Approved.

PR #1021, opened 2022-02-16
[api][SIMS #894]  logic update and user null check for getUserFullName function

ann-aot:
- added null check 
image
- corrected create program logic


henryoforeh-aot (approved):

dheepak-aot (approved):
LGTM @ann-aot

PR #1025, opened 2022-02-17
[web][SIMS #406]Fix for Designation Agreement Links

andrewsignori-aot:
While executing the association of the hyperlinks with the form.io during the load event, after the form call a redraw due to a binding changing, the association with the hyperlinks was lost.
Exposed the event from form.io called render that will happen after a redraw, allowing us to attach the hyperlink events every time that the hyperlink is recreated due to a redraw.

### DIFF
--- a/sources/packages/web/src/components/partial-view/DesignationAgreement/DesignationAgreementForm.vue
+++ b/sources/packages/web/src/components/partial-view/DesignationAgreement/DesignationAgreementForm.vue
@@ -35,7 +35,7 @@ export default {
     const MANAGE_LOCATIONS_LINK = "goToManageLocations";
     const MANAGE_USERS_LINK = "goToManageUsers";
 
-    const formRender = async () => {
+    const formRender = () => {
       RouteHelper.AssociateHyperlinkClick(MANAGE_LOCATIONS_LINK, () =>
         router.push({ name: InstitutionRoutesConst.MANAGE_LOCATIONS }),
       );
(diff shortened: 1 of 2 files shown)


sources/packages/web/src/components/partial-view/DesignationAgreement/DesignationAgreementForm.vue
```diff
@@ -35,7 +35,7 @@ export default {
     const MANAGE_LOCATIONS_LINK = "goToManageLocations";
     const MANAGE_USERS_LINK = "goToManageUsers";
 
-    const formLoaded = async () => {
+    const formRender = async () => {
       RouteHelper.AssociateHyperlinkClick(MANAGE_LOCATIONS_LINK, () =>
```
henryoforeh-aot, line 39: Is this really an async function ? I can see that the implementation is non asynchronous.

andrewsignori-aot: No, it is not. I will change it.

andrewsignori-aot: Changed, good catch 😉

henryoforeh-aot (approved):

ann-aot (approved):
👍

dheepak-aot (approved):
LGTM @andrewsignori-aot . Nice catch 👍

Issue #451, opened 2021-05-21
Upgrade NestJS


hellolynn-dsib:
**Additional Context**
- Refer to #761 for work started
- POC Branch: https://github.com/bcgov/SIMS/tree/poc/updating-nestjs

**Acceptance Criteria**
- [x] Create PR to allow the Nestjs upgrade;

abschwenker:
Openshift still appears to be on node-12. IST-main-1023

I believe the issue is in the makefile still pulling a node12 version, causing the docker16 build to revert to 12

abschwenker:
nest is version 8.2.1 deployed. nest --version

PR #1030, opened 2022-02-17
[api][SIMS #451] Source code update for NEST upgrade

dheepak-aot:
Following changes are made to have the code ready for NEST upgrade and fix the decorators which are identified to be possibly used between more the one client types(Student | Institution | AEST | Supporting users)

- [x] Update setupDB.ts removing the line which deletes the config property schema and uses local variable to pass to the setup
- [x] Adding the property enum to all entities which has an enum column.
- [x] Removing @Inject("Connection") from all the service constructors.
- [x] Updating InstitutionLocation, InstitutionAdmin and Groups guard to validate only their respective client type users.

### DIFF
--- a/sources/packages/api/scripts/setupDB.ts
+++ b/sources/packages/api/scripts/setupDB.ts
@@ -9,7 +9,6 @@ const config = require("../ormconfig");
   try {
     // Create Connection
     console.log("**** Running setupDB ****");
-    delete config.entities;
     const connection = await createConnection({
       ...config,
       logging: ["error", "warn", "info"],
--- a/sources/packages/api/scripts/setupDB.ts
+++ b/sources/packages/api/scripts/setupDB.ts
@@ -9,16 +9,14 @@ const config = require("../ormconfig");
   try {
     // Create Connection
     console.log("**** Running setupDB ****");
-    delete config.schema;
     delete config.entities;
     const connection = await createConnection({
       ...config,
       logging: ["error", "warn", "info"],
     });
-    const schema = process.env.DB_SCHEMA || "sims";
-    await connection.query(`CREATE SCHEMA IF NOT EXISTS ${schema};`);
-    await connection.query(`SET search_path TO ${schema}, public;`);
-    await connection.query(`SET SCHEMA '${schema}';`);
+    await connection.query(`CREATE SCHEMA IF NOT EXISTS ${config.schema};`);
+    await connection.query(`SET search_path TO ${config.schema}, public;`);
+    await connection.query(`SET SCHEMA '${config.schema}';`);
     console.log(`**** Running migration ****`);
     await connection.runMigrations();
     await connection.close();
--- a/sources/packages/api/src/auth/guards/groups.guard.ts
+++ b/sources/packages/api/src/auth/guards/groups.guard.ts
@@ -1,5 +1,6 @@
 import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common";
 import { Reflector } from "@nestjs/core";
+import { AuthorizedParties } from "../authorized-parties.enum";
 import { GROUPS_KEY } from "../decorators";
 import { UserGroups } from "../user-groups.enum";
 import { IUserToken } from "../userToken.interface";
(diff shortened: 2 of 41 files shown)


ann-aot:
> Following changes are made to have the code ready for NEST upgrade and fix the decorators which are identified to be possibly used between more the one client types(Student | Institution | AEST)

what about supporting users?

sources/packages/api/scripts/setupDB.ts
```diff
@@ -9,16 +9,14 @@ const config = require("../ormconfig");
   try {
     // Create Connection
     console.log("**** Running setupDB ****");
-    delete config.schema;
```
ann-aot, line 12: so, now we no longer need to delete this, just asking for my info

sources/packages/api/src/auth/guards/groups.guard.ts
```diff
@@ -22,6 +23,14 @@ export class GroupsGuard implements CanActivate {
       return true;
     }
     const { user } = context.switchToHttp().getRequest();
+    /**
+     * When a controller is shared between more then one client type(e.g Institution and Ministry)
+     * Following logic ensures that group check is done only for ministry(AEST) user.
+     * For client types except AEST, this decorator will not validate anything.
```
ann-aot, line 29: 👍 thanks for the comment

sources/packages/api/src/auth/guards/institution-location.guard.spec.ts
```diff
@@ -21,6 +22,7 @@ const activateGuard = async (
   const guard = new InstitutionLocationGuard(reflector);
   const user = {} as IInstitutionUserToken;
   user.authorizations = authorizations;
+  user.authorizedParty = AuthorizedParties.institution;
```
ann-aot, line 25: why so we need this `authorizedParty`, for my info

ann-aot (commented):
added some questions.

sources/packages/api/src/auth/guards/institution-location.guard.spec.ts
dheepak-aot: In test cases where we expect the guard to block the execution, it escapes the validation because no clientType is assigned to the user when it is initialized for test cases. So initialized with right client type for unit test.

image

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -60,7 +60,7 @@ export class ApplicationService extends RecordDataModelService<Application> {
   logger: LoggerService;
 
   constructor(
-    @Inject("Connection") connection: Connection,
+    connection: Connection,
```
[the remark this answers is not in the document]
dheepak-aot: When the connection is just required to be passed to the super class constructor, then we do not use access specifier i.e private readonly. In this case connection is considered as parameter not a property.
image

image
 But when any of the member functions, require connection object, then we use private readonly connection.
In this case connection is considered as property.
image

image
image

andrewsignori-aot (approved):
It looks good 👍 
As we talked, please ensure that it will run on a clean database from scratch and a second time when the database is already present 😉

sources/packages/api/src/services/application/application.service.ts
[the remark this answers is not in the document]
henryoforeh-aot: it might be better to change the scope to public in case we start writing unit tests soon.

sources/packages/api/scripts/setupDB.ts
andrewsignori-aot: It is not clear the reason on why this line was introduced in the first place. This is basically ensuring that the schema property from _Typeorm_ config will be ignored, but we need it.
After Nestjs was upgraded during POC there was a issue with migrations where the check performed by Typeorm to verify if the Migrations table was already present was apparently being executed on the _public_ schema instead of the _sims_ schema. 
Allowing this config to be present solves the issue.

ann-aot (approved):
LGTM 👍

sources/packages/api/src/services/application/application.service.ts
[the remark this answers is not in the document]
ann-aot: i think my proper qn is what is the difference between sources/packages/api/src/services/institution-type/institution-type.service.ts and sources/packages/api/src/services/institution-type/institution-type.service.ts, w.r.t connection

dheepak-aot:
@andrewsignori-aot as discussed to test the removal of delete config.schema and delete config.entities, I ran the migration from a clean database(aest_test in my case) and I saw that it ran without any issue. Also I ran the migration again with existing tables and it worked fine. Additionally I created an institution and a location from the clean DB. 
image

image

sources/packages/api/src/services/application/application.service.ts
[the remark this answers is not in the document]
dheepak-aot: @henryoforeh-aot I did not get your statement. You meant to we should use public readonly for connection and other global variables? If so IMO they should be private only, as they are not suppose to be accessed outside the class. 
Any discussion regarding this we can take it separately outside the story.

PR #1036, opened 2022-02-23
[Api][SIMS #451] NEST Upgrade

dheepak-aot:
Updated the following as per #761  

- [x] Updating the packages of rxjs, nestjs and typescript. (As a result, there are a list of packages dependent to it are updated)
- [x] Adding enumName to all properties of entities which reference a DB enum object.
- [x] Logger service extending consoleLogger instead of logger as per POC. setContext belongs to consoleLogger.
- [x] Remove set schema sql from the database service as schema is already present in config.

### DIFF
--- a/sources/packages/api/package-lock.json
+++ b/sources/packages/api/package-lock.json
@@ -1,46 +1,11629 @@
 {
   "name": "api",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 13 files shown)


andrewsignori-aot (approved):
looks good 👍

guru-aot (approved):
LGTM

ann-aot (approved):
LGTM 👍

henryoforeh-aot (approved):

Issue #467, opened 2021-05-27
Ministry Manages Designation


mzmichellesmith:
This is ticket 2 of 2 for the designation feature

UX Ticket #892, Ticket 1 of 2 #406

As discussed:
Ticket 1:
Institution summary (common component)
DB changes
request form
submit

Ticket 2:
Read-only form (with approved)
ministry approve
ministry decline
ministry pending summary page
student application drop down

**Additional Context**
- Designation is the process of an institution signing an agreement with SABC in order to to be eligible for students to apply for funding

**Institution Portal**
- [x] When clicking view on a designation request display the read-only form

**Ministry Portal**
- [x] Under the institution Portal, designation tab, display institution designation summary page (same used on institution page)
- [x] While viewing a pending Designation Request. display "decline" and "approve designation" buttons
- [x] On click of "Approve designation" display the approve designation modal.
- Start date
- Expiry date
- Selectable locations to designate (a ministry user can "overwrite" the institutions choices)
- Notes (these notes should show up under the notes section in the profile)
- [x] On click of "Decline designation" display the decline designation modal
- [x] On click of "view" display read-only designation request
- [x] Notes (these notes should show up under the notes section in the account)

**Both**
- [x] While viewing an approved designation agreement display the approved designation agreement
NOTE: A ministry user can Re-approve a designation agreement with different dates/locations
- [x] While viewing a declined designation agreement display the declined designation agreement
- [x] A ministry user can also decline an approved designation agreement

Tech Notes
- Designated means: a location with an approved designation agreement that covers todays date
- If ministry needs to remove Designation for 1 location
Re-approve a DA with different locations
- If ministry needs to remove Designation for all locations
Decline an approved DA
- If institutions needs to add Designation for 1 location
Re-create a DA with different locations
- If institutions needs to remove Designation for 1 location
Re-create a DA with different locations
- If institutions needs to remove Designation for all location
Re-create a DA with no locations

dheepak-aot:
Lazy pagination has not been considered for the pending designations data table as it is assumed that number of designations that are pending to be approved at given time will be considerably less to ignore lazy pagination. If this changes anytime in future we can add the pagination and sort layer to the API.

PR #1072, opened 2022-03-08
[Web][SIMS #467] Work around to fix the non-reactive nature of form.io readOnly property.

dheepak-aot:
Form.io is not reactively binding the property readOnly. If the readOnly value change after DOM(form) is mounted, form does not respond to it.

Hence, the modelLoaded variable ensures that API is called and data is initialized before form is loaded.


andrewsignori-aot (approved):

ann-aot (approved):
👍

guru-aot (approved):
LGTM

PR #1085, opened 2022-03-10
Updated the student schema for the migrations

guru-aot:
Bug during the migration revert. Students does not contain schema appended to it.


andrewsignori-aot (approved):
Looks good, thanks for fixing it 👍

Issue #28, opened 2020-12-15
DB Changes for Reassessments


mzmichellesmith:
Preparation for reassessment (new DB structure)
- [x] 1. Create new DB structure to accommodate central assessment table
- [ ] 2. Adjust the API/Camunda to use the new fields/stop using the old one
- [ ] 3. Cleanup the not used fields/relationships (after point 2 changes)

PR #1090, opened 2022-03-15
[api][workflow][SIMS #28]API Changes for Assessments/Reassessments - Part 1

andrewsignori-aot:
- Created a new endpoint for assessments/reassessments (system-access/assessment);
- Moved and adapted some endpoints from system-access/applications to system-access/assessment;

- Started to adapt the assessment-gateway workflow to deal with the reassessments
image

- Changed the below API calls to use the new endpoints
image

- New endpoints moved and adapted to the new assessment controller
image

- More cleanups to happen in upcoming PRs;
- SonarCloud issues related to TODOs;
- SonarCloud code duplication not identified;


andrewsignori-aot:
Sonar Cloud is wrongly considering a duplication on the code below code.
image
image

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -36,7 +38,7 @@ export class Application extends RecordDataModel {
     type: "jsonb",
     nullable: false,
   })
-  data: any;
+  data: ApplicationData;
```
ann-aot, line 41: my story also has this point. But it's partial right? shouldn't it be `Partial<ApplicationData>`

sources/packages/api/src/services/application/application.models.ts
```diff
@@ -1,6 +1,12 @@
...
 export interface ApplicationOverriddenResult {
   overriddenApplication: Application;
   createdApplication: Application;
+  createdAssessment: StudentAssessment;
+}
+
+export interface ApplicationSubmissionResult {
+  application: Application;
+  assessment: StudentAssessment;
 }
```
henryoforeh-aot, line 12: Are we also creating database entities as interfaces ?

andrewsignori-aot: Sorry, I am not following. This is a service model used to return a result from a service method. Can we please elaborate more?

henryoforeh-aot: Oh ! never mind, I thought it was in database entities. Resolving comment.

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
```diff
@@ -0,0 +1,173 @@
...
+  @IsOptional()
+  @IsInt()
+  @Min(1)
+  offeringId?: number;
+  @IsNotEmpty()
+  @IsInt()
+  @Min(1)
+  locationId: number;
+  @IsEnum(ProgramInfoStatus)
+  status: ProgramInfoStatus;
```
ann-aot, line 112: shouldn't we need to add `@IsNotEmpty()` for enum?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -119,6 +128,15 @@ export class ApplicationService extends RecordDataModelService<Application> {
         APPLICATION_NOT_VALID,
       );
     }
+
+    const now = new Date();
+    const auditUser = { id: auditUserId } as User;
+    const originalAssessment = new StudentAssessment();
+    originalAssessment.triggerType = AssessmentTriggerType.OriginalAssessment;
+    originalAssessment.submittedBy = auditUser;
+    originalAssessment.submittedDate = now;
```
ann-aot, line 137: since the date is a timestamp, its better to fed directly right? same in line 153, my thought

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  /**
+   * Updates the assessment workflowId.
+   * The workflow id on the assessment table must be null otherwise
+   * the update will not be performed. Avoiding the update prevents
+   * that two different workflow instances try to process the same
+   * assessment.
+   * @param assessmentId assessment id to be updated.
+   * @param assessmentWorkflowId workflowId to be updated.
```
ann-aot, line 134: return comment

sources/packages/api/src/services/workflow/workflow-actions.service.ts
```diff
@@ -12,11 +12,13 @@ export class WorkflowActionsService {
...
+   * @param assessmentId student assessment that need to be processed.
    * @returns result of the application start.
    */
   async startApplicationAssessment(
     workflowName: string,
     applicationId: number,
+    assessmentId: number,
   ): Promise<WorkflowStartResult> {
     try {
       return await this.workflowService.start(workflowName, {
```
ann-aot, line 24: do we need await ?

sources/packages/api/src/database/entities/application.model.ts
ann-aot: mm, I am not clear, may be we can syncup

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
```diff
@@ -0,0 +1,173 @@
...
+ */
+export class ApplicationAssessmentDTO {
+  /**
+   * Origin of the assessment.
+   */
+  triggerType: AssessmentTriggerType;
+  /**
+   * Application dynamic data.
+   */
+  data: ApplicationData;
```
ann-aot, line 71: partial?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -92,11 +100,12 @@ export class ApplicationService extends RecordDataModelService<Application> {
    */
   async submitApplication(
     applicationId: number,
+    auditUserId: number,
     studentId: number,
     programYearId: number,
     applicationData: any,
```
ann-aot, line 106: we can change the type any to ApplicationData

sources/packages/api/src/route-controllers/assessment/assessment.system-access.controller.ts
```diff
@@ -0,0 +1,309 @@
...
+
+  /**
+   * Get the assessment and the related application information.
+   * This method is used by the the assessment workflow as a main source
+   * of information for the assessment/reassessment and the application.
+   * @param assessmentId assessment id .
+   * @returns assessment and the related application information.
+   */
+  @Get(":id")
+  @ApiResponse({
```
guru-aot, line 53: Are we planning to have this in our newly created methods? or as discussed can remove and do it in a common place?

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
```diff
@@ -0,0 +1,173 @@
...
+export class UpdateProgramInfoDTO {
+  @IsOptional()
+  @IsInt()
+  @Min(1)
+  programId?: number;
+  @IsOptional()
+  @IsInt()
+  @Min(1)
+  offeringId?: number;
+  @IsNotEmpty()
```
guru-aot, line 107: Instead of IsNotEmpty we can use IsDefined for non string objects, as 
image
https://github.com/typestack/class-validator#validation-decorators

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Values to be associated with a disbursement.
+ */
+export class DisbursementValueDTO {
+  @IsNotEmpty()
+  valueCode: string;
+  @IsEnum(DisbursementValueType)
+  valueType: DisbursementValueType;
+  @IsNotEmpty()
+  @IsNumber()
+  @Min(0)
```
guru-aot, line 140: Instead of using @IsNumber() && @Min(0) we can use IsPositive()
Just a suggestion, we can discuss, if we want to.
[https://github.com/typestack/class-validator#validation-decorator](https://github.com/typestack/class-validator#validation-decorators)

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Why being a timestamp should be a justification to use the new Date()? Do you mean if between the time of one new Date() and another there will be some milliseconds of difference, right?
From DB's perspective, it seems better to me if the date is expressed as the same. We are capturing the time that the record was created and not the time that it was generated in the server, does it make sense?

sources/packages/api/src/services/workflow/workflow-actions.service.ts
andrewsignori-aot: Yes, because of the try/catch.

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
andrewsignori-aot: Why?

sources/packages/api/src/route-controllers/assessment/assessment.system-access.controller.ts
andrewsignori-aot: During the discussion, my understanding was that whatever is part of the body of the controller is suitable to be added as API response and also the OK/CREATED when we would provide some description.

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
guru-aot: I think @IsEnum implicitly checks whether the value is not empty and will through validation error if null or empty value is passed.

sources/packages/api/src/services/application/application.service.ts
guru-aot: I guess, the use of constant here is to match the milliseconds between the assessment submitted date and applicationStatusUpdatedDate.

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -0,0 +1,237 @@
...
+      ])
+      .innerJoin("assessment.application", "application")
+      .innerJoin("application.programYear", "programYear")
+      .innerJoin("application.student", "student")
+      .leftJoin("assessment.offering", "offering")
+      .leftJoin("offering.educationProgram", "educationProgram")
+      .leftJoin("application.location", "location")
+      .leftJoin("location.institution", "institution")
+      .leftJoin("institution.institutionType", "institutionType")
+      .andWhere("assessment.id = :assessmentId", {
```
guru-aot, line 74: We can use .where instead of .andWhere

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -0,0 +1,237 @@
...
+        "application.data",
+      ])
+      .innerJoin("assessment.application", "application")
+      .where("assessment.id = :assessmentId", { assessmentId })
+      .getOne();
+
+    if (assessment.assessmentWorkflowId) {
+      throw new CustomNamedError(
+        `Student assessment was already started and has a workflow associated with. Assessment id ${assessmentId}`,
+        INVALID_OPERATION_IN_THE_CURRENT_STATUS,
```
guru-aot, line 168: The custom error is appropriate to the actual error defined, but it would be better to have a specific named error like "ASSESSMENT_ALREADY_ASSOCIATED" or "ASSESSMENT_FOR_THE_APPLICATION_ASSOCIATED"

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: Talked with @ann-aot and I believe that we are good about it.
Yes @guru-aot, that is the idea 😉

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: @ann-aot I changed the service layer to use the ApplicationData as we talked but when I do it on the DTO it impact too many places outside the PR, so I will not do the change on DTO right now.

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
andrewsignori-aot: Good catch @guru-aot. We can use IsPositive() when we have @min(1), but when we have min(0) it will not be applicable because zero is not positive 😉

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
andrewsignori-aot: We can be more specific, yes, but there will be no difference for the consumer of the constants because they all say the same: "It was not possible to execute the process". If we need a specific reason to differentiate them I would besupporting the creation of the specific ones. Please, let me know your thoughts.
Actually, the _INVALID_OPERATION_IN_THE_CURRENT_STATUS_ was supposed to be _INVALID_OPERATION_IN_THE_CURRENT_STATE_.

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -0,0 +1,237 @@
...
+        "student.studentPDVerified",
+      ])
+      .innerJoin("assessment.application", "application")
+      .innerJoin("application.programYear", "programYear")
+      .innerJoin("application.student", "student")
+      .leftJoin("assessment.offering", "offering")
+      .leftJoin("offering.educationProgram", "educationProgram")
+      .leftJoin("application.location", "location")
+      .leftJoin("location.institution", "institution")
+      .leftJoin("institution.institutionType", "institutionType")
```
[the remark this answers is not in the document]
andrewsignori-aot: I am not sure if I am following but, the inner join "scope' is not about the immediate previous join, right?
If we changed the institution type to be INNER JOIN and there is no location id present, this will make that the entire row be not returned. Does it make sense?

sources/packages/api/src/database/entities/application.model.ts
andrewsignori-aot: As we talked, partial does not apply to this context. Please let me know if you find something that indicates otherwise.

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
andrewsignori-aot: Changed to use @IsOptional() and @IsPositive().

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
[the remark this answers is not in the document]
guru-aot: Yes you are right. If the leftjoin in the previous joins does not return any rows the institutionType will not fetch any rows, so if the usecase is to return even if the rows are not returned, then we are good.

guru-aot (approved):
LGTM, thanks for doing the changes.

ann-aot (approved):
Good Work @andrewsignori-aot 👍

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -18,23 +18,19 @@ export class SaveApplicationDto {
   /**
    * Application dynamic data.
    */
-
-  @IsNotEmpty()
+  @IsObject()
   data: any;
```
dheepak-aot, line 22: Can we use the interface created for Application here?

andrewsignori-aot: Yes, we can try and see if Nestjs will keep all dynamic properties. I believe that yes because it is an interface.
When I did it, it generated too many changes that would impact too many files outside this PR so I reverted it.

sources/packages/api/src/route-controllers/assessment/assessment.system-access.controller.ts
```diff
@@ -0,0 +1,309 @@
...
+  ): Promise<void> {
+    if (payload.offeringId) {
+      const offering = await this.offeringService.getProgramOffering(
+        payload.locationId,
+        payload.programId,
+        payload.offeringId,
+      );
+      if (!offering) {
+        throw new UnprocessableEntityException(
+          "Not able to find the offering associate with the program and location.",
```
dheepak-aot, line 144: associated ?

sources/packages/api/src/route-controllers/assessment/assessment.system-access.controller.ts
```diff
@@ -0,0 +1,309 @@
...
+  })
+  @ApiResponse({
+    status: HttpStatus.UNPROCESSABLE_ENTITY,
+    description:
+      "The disbursement information is already present or either the assessment or the application is not in the correct state.",
+  })
+  async createDisbursement(
+    @Param("assessmentId") assessmentId: number,
+    @Body() payload: CreateDisbursementsDTO,
+  ): Promise<number[]> {
```
dheepak-aot, line 260: Why are we returning the ids of the created disbursements. Asking for my understanding.

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
```diff
@@ -0,0 +1,168 @@
...
+  valueAmount: number;
+}
+
+/**
+ * Disbursement to be created altogether with its values.
+ */
+export class DisbursementScheduleDTO {
+  @IsNotEmpty()
+  disbursementDate: Date;
```
dheepak-aot, line 144: for date fields we can use @IsDate() this allows datestring as well.

sources/packages/api/src/route-controllers/assessment/assessment.system-access.controller.ts
andrewsignori-aot: So far when we are doing a POST, we are returning the id/ids of the created resources.
We started this a long time ago to try to make it a bit more "rest" in the sense that POST usually returns the 201 status and indicates in some way the resource created, in the location header usually.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -132,13 +151,17 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
         studentId,
         application.studentFiles,
         associatedFiles,
       );
-      return this.repo.save(application);
+      application.modifier = auditUser;
+      application.studentAssessment = [originalAssessment];
+
+      await this.repo.save(application);
+      return { application, assessment: originalAssessment };
```
dheepak-aot, line 164: when I saw the code where submitApplication is called, 

I saw const {assessment} = this.service.submitApplication() which made me think that the property assessment from Application(Entity) is being parsed here.

Hence in line: 164 can we use a property name which is not existing in application entity. like studentAssessment.?

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -0,0 +1,240 @@
...
+} from "./student-assessment.constants";
+
+/**
+ * Manages the student assessment related operations.
+ */
+@Injectable()
+export class StudentAssessmentService extends RecordDataModelService<StudentAssessment> {
+  constructor(
+    connection: Connection,
+    private readonly workflow: WorkflowActionsService,
```
dheepak-aot, line 28: Can we have the variable name ending with service?

sources/packages/api/src/route-controllers/assessment/models/assessment.system-access.dto.ts
andrewsignori-aot: This is a date only field 😉 
I tested and it worked. Change it!

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: I believe that the below code syntax is pretty clear about destructuring the result of the method.
I would like a second option on this: @ann-aot @guru-aot @henryoforeh-aot 
```
const { assessment } = this.service.submitApplication();
```

andrewsignori-aot: Renamed _assessment_ to _createdAssessment_ as we talked.

dheepak-aot (approved):
Good work @andrewsignori-aot. 👍
Thanks for doing the changes.

Issue #998, opened 2022-02-09
Revalidate SIN on student account change


abschwenker:
**Acceptance Criteria**
- [x] This will handle the way we handle the relationship between valid_SIN and student (remove when change name and add new SIN record for example)
- [x] When we receive a student change we must re-validate a students SIN number with CRA
- Given Names
- Last Name
- DOB change
 - [x] Keep last validated record (do not overwrite)
- [x] Ensure records and relationship functioning on account creation
- [x] Ensure banner and application start behavior reflects this status
- [x] Ensure disbursements query will not be sent while not valid

PR #1095, opened 2022-03-18
[SIMS-#998][API]Revalidate SIN

guru-aot:
- [x] This will handle the way we handle the relationship between valid_SIN and student (remove when change name and add new SIN record for example)
- [x] When we receive a student change we must re-validate a students SIN number with CRA
- Given Names
- Last Name
- DOB change
 - [x] Keep last validated record (do not overwrite)
- [x] Ensure banner and application start behavior reflects this status

### DIFF
--- a/sources/packages/api/src/route-controllers/student/student.controller.ts
+++ b/sources/packages/api/src/route-controllers/student/student.controller.ts
@@ -121,7 +121,7 @@ export class StudentController extends BaseController {
         phone: existingStudent.contactInfo.phone,
       },
       pdVerified: existingStudent.studentPDVerified,
-      validSin: existingStudent.sinValidation.isValidSIN,
+      validSin: existingStudent.sinValidation?.isValidSIN,
       pdSentDate: existingStudent.studentPDSentAt,
       pdUpdatedDate: existingStudent.studentPDUpdateAt,
       pdStatus: determinePDStatus(existingStudent),
--- a/sources/packages/api/src/route-controllers/student/student.controller.ts
+++ b/sources/packages/api/src/route-controllers/student/student.controller.ts
@@ -121,7 +121,7 @@ export class StudentController extends BaseController {
         phone: existingStudent.contactInfo.phone,
       },
       pdVerified: existingStudent.studentPDVerified,
-      validSin: existingStudent.sinValidation?.isValidSIN,
+      validSin: existingStudent.sinValidation.isValidSIN,
       pdSentDate: existingStudent.studentPDSentAt,
       pdUpdatedDate: existingStudent.studentPDUpdateAt,
       pdStatus: determinePDStatus(existingStudent),
--- a/sources/packages/api/src/services/student/student.service.ts
+++ b/sources/packages/api/src/services/student/student.service.ts
@@ -146,14 +146,18 @@ export class StudentService extends RecordDataModelService<Student> {
     let mustSave = false;
     const sinValidation = new SINValidation();
     const userTokenBirthdate = getDateOnly(userToken.birthdate);
+    if (userToken.givenNames === undefined) {
+      userToken.givenNames = null;
+    }
     if (
       userTokenBirthdate !== studentToSync.birthDate ||
       userToken.lastName !== studentToSync.user.lastName ||
-      userToken.givenNames != studentToSync.user.firstName
+      userToken.givenNames !== studentToSync.user.firstName
     ) {
       studentToSync.birthDate = userTokenBirthdate;
       studentToSync.user.lastName = userToken.lastName;
       studentToSync.user.firstName = userToken.givenNames;
+      sinValidation.user = studentToSync.user;
       studentToSync.sinValidation = sinValidation;
       mustSave = true;
     }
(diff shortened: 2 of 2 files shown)


sources/packages/api/src/services/student/student.service.ts
```diff
@@ -143,27 +148,29 @@ export class StudentService extends RecordDataModelService<Student> {
         "Not able to find a student using the username (bcsc name)",
       );
     }
-
     let mustSave = false;
-
+    const userTokenBirthdate = getDateOnly(userToken.birthdate);
+    if (userToken.givenNames === undefined) {
+      userToken.givenNames = null;
+    }
```
henryoforeh-aot, line 155: if(!userToken.givenNames) checks for undefined, null and empty. I'm not sure if lines 153 to 155 are necessary.

guru-aot: Nice question, To all @ann-aot @andrewsignori-aot  @dheepak-aot @henryoforeh-aot @abschwenker Yes, as per the .save function in typeorm, if the value passed has undefined, the save fails for the particular column, So when the sync happens and the BCSC return the givenname as undefined, its set in our DB as null, to further creating records in the sinvalidation table. Note: The other columns are mandatory in BCSC form, so the problem is givenNames.

guru-aot: Nice question, To all @ann-aot @andrewsignori-aot  @dheepak-aot @henryoforeh-aot @abschwenker Yes, as per the .save function in typeorm, if the value passed has undefined, the save fails for the particular column, So when the sync happens and the BCSC return the givenname as undefined, its set in our DB as null, to further creating records in the sinvalidation table. Note: The other columns are mandatory in BCSC form, so the problem is givenNames.

sources/packages/api/src/route-controllers/student/student.controller.ts
```diff
@@ -121,7 +121,7 @@ export class StudentController extends BaseController {
         phone: existingStudent.contactInfo.phone,
       },
       pdVerified: existingStudent.studentPDVerified,
-      validSin: existingStudent.sinValidation.isValidSIN,
+      validSin: existingStudent.sinValidation?.isValidSIN,
```
andrewsignori-aot, line 124: What is the scenario when _sinValidation_ is null?

henryoforeh-aot (approved):

sources/packages/api/src/services/student/student.service.ts
andrewsignori-aot: @guru-aot, fail to save does not seems fair 😄 
It is the way that we tell Typeorm to ignore that particular column during the _update/insert_.

sources/packages/api/src/route-controllers/student/student.controller.ts
guru-aot: You are right, I added it as part of the blue box testing. Updated now.

sources/packages/api/src/services/student/student.service.ts
guru-aot: He he, yeah its not doing its job properly 💯

andrewsignori-aot: Not sure if we are on the same page here 😅 
Typeorm is right and it is doing what it is supposed to do.

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -143,27 +148,29 @@ export class StudentService extends RecordDataModelService<Student> {
...
     if (
-      userToken.email !== studentToSync.user.email ||
+      userTokenBirthdate.getTime() !== studentToSync.birthDate.getTime() ||
```
andrewsignori-aot, line 157: Is there a reason to add the _.getTime()_ to do the comparison here?

guru-aot: image
Yes the new Date does not work with !==  properly without the .getTime()

ann-aot: image
i didn't get the point

guru-aot: updated to dayjs as discussed

andrewsignori-aot (approved):
Looks good 👍 Great work!

sources/packages/api/src/services/student/student.service.ts
```diff
@@ -143,27 +149,28 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      sinValidation.user = studentToSync.user;
+      studentToSync.sinValidation = sinValidation;
       mustSave = true;
-      //TODO update the SIN validation here
     }
-
-    const userTokenBirthdate = getDateOnly(userToken.birthdate);
     if (
-      userTokenBirthdate !== studentToSync.birthDate ||
+      userToken.email !== studentToSync.user.email ||
```
dheepak-aot, line 170: These conditions can be added to the previous if itself. why has it been added here separately?

dheepak-aot: Is it because, change in gender and email will not trigger SIN validation again right?

abschwenker (approved):

ann-aot (approved):
👍

PR #1097, opened 2022-03-18
Create common component for student application - designation query enhancement

ann-aot:
Designation query enhancement, added case, which will return true if location is designated and return false for else case

### DIFF
--- a/sources/packages/api/src/services/institution-location/institution-location.service.ts
+++ b/sources/packages/api/src/services/institution-location/institution-location.service.ts
@@ -3,11 +3,9 @@ import { RecordDataModelService } from "../../database/data.model.service";
 import { InstitutionLocation } from "../../database/entities/institution-location.model";
 import { Connection, UpdateResult } from "typeorm";
 import { ValidatedInstitutionLocation } from "../../types";
-import {
-  DesignatedAndNotDesignatedLocations,
-  InstitutionLocationTypeDto,
-} from "../../route-controllers/institution-locations/models/institution-location.dto";
+import { InstitutionLocationTypeDto } from "../../route-controllers/institution-locations/models/institution-location.dto";
 import { DesignationAgreementLocationService } from "../designation-agreement/designation-agreement-locations.service";
+import { LocationWithDesigationStatus } from "./institution-location.models";
 @Injectable()
 export class InstitutionLocationService extends RecordDataModelService<InstitutionLocation> {
   constructor(
@@ -203,9 +201,7 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
    * @param locationId location id
    * @returns InstitutionLocation
    */
-  async getLocation(
-    locationId: number,
-  ): Promise<DesignatedAndNotDesignatedLocations> {
+  async getLocation(locationId: number): Promise<LocationWithDesigationStatus> {
     return this.repo
       .createQueryBuilder("location")
       .select("location.name", "locationName")
@@ -215,7 +211,8 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
             .getExistsDesignatedLocation()
             .getSql()}) THEN true
           ELSE false
-        END "isDesignated"`,
+        END`,
+        "isDesignated",
       )
       .where("location.id = :locationId", { locationId })
       .getRawOne();
--- a/sources/packages/api/src/services/institution-location/institution-location.service.ts
+++ b/sources/packages/api/src/services/institution-location/institution-location.service.ts
@@ -3,7 +3,10 @@ import { RecordDataModelService } from "../../database/data.model.service";
 import { InstitutionLocation } from "../../database/entities/institution-location.model";
 import { Connection, UpdateResult } from "typeorm";
 import { ValidatedInstitutionLocation } from "../../types";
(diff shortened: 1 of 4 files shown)


sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -200,31 +203,21 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
-  /**
-   * Get institution location by location id.
-   * @param locationId location id
-   * @returns InstitutionLocation
-   */
-  async getDesignatedLocationById(
+  async getLocation(
     locationId: number,
-  ): Promise<InstitutionLocation> {
+  ): Promise<DesignatedAndNotDesignatedLocations> {
```
andrewsignori-aot, line 208: Please move _DesignatedAndNotDesignatedLocations_ class to the service models.

andrewsignori-aot, line 208: Can we rename the object to `LocationDesignation` or `LocationWithDesigationStatus` maybe 🤔

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -200,31 +203,21 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
     return this.repo
       .createQueryBuilder("location")
-      .select(["location.name"])
-      .where("location.id = :locationId", { locationId })
-      .andWhere(
-        `EXISTS(${this.designationAgreementLocationService
-          .getExistsDesignatedLocation()
-          .getSql()})`,
+      .select("location.name", "locationName")
+      .addSelect(
```
andrewsignori-aot, line 212: Because we are using the addSelect I would prefer to use the overload, as below:
```typescript
.addSelect(
        `CASE
          WHEN EXISTS(${this.designationAgreementLocationService
            .getExistsDesignatedLocation()
            .getSql()}) THEN true
          ELSE false
        END`, "isDesignated"
      )
```

andrewsignori-aot (commented):
Thanks, @ann-aot for taking this forward. The method looks good, just take a look in the minor comments 😉

henryoforeh-aot (approved):

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -3,7 +3,10 @@ import { RecordDataModelService } from "../../database/data.model.service";
 import { InstitutionLocation } from "../../database/entities/institution-location.model";
 import { Connection, UpdateResult } from "typeorm";
 import { ValidatedInstitutionLocation } from "../../types";
-import { InstitutionLocationTypeDto } from "../../route-controllers/institution-locations/models/institution-location.dto";
+import {
+  DesignatedAndNotDesignatedLocations,
```
dheepak-aot, line 7: Not a part of this PR. The dto name DesignatedAndNotDesignatedLocations denotes locations apparently. So dto name could have been simple.

dheepak-aot: Comment above updated. I thought initially that dto was not defined in this PR.

sources/packages/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -200,31 +203,21 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+        `CASE
+          WHEN EXISTS(${this.designationAgreementLocationService
+            .getExistsDesignatedLocation()
+            .getSql()}) THEN true
+          ELSE false
+        END "isDesignated"`,
       )
-      .getOne();
+      .where("location.id = :locationId", { locationId })
+      .getRawOne();
```
dheepak-aot, line 221: We could add a transient property to the entity(isDesignated) and use getOne() instead of getRawOne(). 
If the query results deviate a lot from Entity model, then service model will be more optimal. may be in this case transient property can help IMO.

dheepak-aot:
Nice work @ann-aot . 👍 2 comments.

sources/packages/api/src/services/institution-location/institution-location.service.ts
ann-aot: DesignatedAndNotDesignatedLocations  is not a DTO here, also interface name updated

sources/packages/api/src/services/institution-location/institution-location.service.ts
ann-aot: i didn't understand? could you explain

sources/packages/api/src/services/institution-location/institution-location.service.ts
ann-aot: ya... 👍

andrewsignori-aot (approved):
Noce work @ann-aot and thanks for doing the changes, looks good to me 👍

ann-aot:
> Noce work @ann-aot and thanks for doing the changes, looks good to me 👍

Thanks for helping me with the query 😄

dheepak-aot (approved):
LGTM @ann-aot. Thanks for testing with transient property 👍

Issue #1003, opened 2022-02-09
Student Reports a change


abschwenker:
Additional Context
- A student reporting a change is the student requesting a re-assessment when something on their application (or supporting user) should use a differnt value

**Acceptance Criteria**
**Students**  - Not PY specific
- [x] Add "Request a Change" Button on top nav bar (always present)
- [x] On click of "Request a Change" navigate to new page with content on figma
- [x] A student must enter an application number
- [x] Validate on "next" application # is a completed application 
- [x] A student can select multiple "types" of changes
- [x] On click of "next" display single UI with multiple forms (1 form per "type" selected)
- [x] All fields are always required
- [x] On "submit for review" save appeal to "sims.appeals" and create relationship in sims.appeal_requests
- [x] A student can only submit 1 "Request a Change" at a time (regardless of application)

**Forms Definition**
- [x] Create appeal form "Student Income Appeal"
- [x] Single component form with dollar text box "My new income is:"
- [x] Create appeal form "Partner Income Appeal"
- [x] Single component form with dollar text box "My Partner's new income is:"

IrenaMilosevic1:
Approved

PR #1101, opened 2022-03-22
[Api][Web][SIMS #1003] Student request a change

dheepak-aot:
- [x] Add "Request a Change" Button on top nav bar
image
- [x] On click of "Request a Change" navigate to new page with content on Figma
image
- [x] A student must enter an application number
- [x] Validate on "next" application # is a completed application 
image
- [x] A student can select multiple "types" of changes
image
- [x] On click of "next" display single UI with multiple forms (1 form per "type" selected)
image
- [x] All fields are always required
image


sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+      );
+    }
+    const submissionResults = payload.studentAppealRequests.map(
+      (studentAppeal) =>
+        this.formService.dryRunSubmission(
+          studentAppeal.formName,
+          studentAppeal.formData,
+        ),
+    );
+    await Promise.all(submissionResults);
```
henryoforeh-aot, line 87: Is Promise.all needed here since you are not trying to resolve an array of results ?

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -0,0 +1,74 @@
...
+      .innerJoin("appeal.application", "application")
+      .innerJoin("application.student", "student")
+      .innerJoin("student.user", "user")
+      .where("user.id = :userId", { userId })
+      .andWhere("appealRequests.appealStatus = :pending", {
+        pending: StudentAppealStatus.Pending,
+      })
+      .limit(1)
+      .getRawOne();
+
```
henryoforeh-aot, line 71: Is it possible to use getOne() here since what you are checking is that a student actually has appeal requests and not using any type of aggregate function?

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -618,4 +620,43 @@ export class ApplicationStudentsController extends BaseController {
       }
     }
   }
+
+  /**
+   * Get application to request appeal.
+   ** Application eligible to be requested for
```
andrewsignori-aot, line 626: Are the special comments with "*" ignored when the swagger is generated?
If not, I would suggest to avoid them 😉

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -618,4 +620,43 @@ export class ApplicationStudentsController extends BaseController {
...
+        new ApiProcessError(
+          "Given application either does not exist or is not complete to request change.",
+          INVALID_APPLICATION_NUMBER,
+        ),
+      );
+    }
+    return {
+      id: application.id,
+      applicationNumber: application.applicationNumber,
+    };
```
andrewsignori-aot, line 660: Can we add the cast to the return type `as ApplicationPrimaryDTO`?

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -618,4 +620,43 @@ export class ApplicationStudentsController extends BaseController {
...
+  })
+  @ApiNotFoundResponse({
+    description:
+      "Application either not found or not eligible to request for change.",
+  })
+  @Get(":applicationNumber/appeal")
+  async getApplicationToRequestChange(
+    @Param("applicationNumber") applicationNumber: string,
+    @UserToken() userToken: IUserToken,
+  ): Promise<ApplicationPrimaryDTO> {
```
andrewsignori-aot, line 643: What does "Primary" mean in this context? Could the DTO be only ApplicationAppealDTO?

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -618,4 +620,43 @@ export class ApplicationStudentsController extends BaseController {
...
+    const application =
+      await this.applicationService.getApplicationToRequestChange(
```
andrewsignori-aot, line 645: I would say that internally we are calling "Request a Change" appeal hence I would suggest renaming the method to `getApplicationToRequestAppeal`, or something similar.

sources/packages/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+/**
+ * DTO for student appeal.
+ */
+export class StudentAppealDTO {
+  applicationId?: number;
+  @IsArray()
+  @ArrayMinSize(1)
+  studentAppealRequests: StudentAppealRequestDTO[];
```
andrewsignori-aot, line 17: Please add the validation to the child objects. Can you try adding the below?
```
@IsArray()
@ArrayMinSize(1)
@ValidateNested({ each: true })
@Type(() => <child type>)
```

sources/packages/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
```diff
@@ -0,0 +1,18 @@
+import { ArrayMinSize, IsArray } from "class-validator";
+/**
+ * DTO for student appeal request.
+ */
+export class StudentAppealRequestDTO {
```
andrewsignori-aot, line 5: Please add the validators accordingly.

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+    super();
+  }
+
+  /**
+   * Submit student appeal.
+   * @param applicationId application for which the appeal is submitted.
+   * @param payload student appeal with appeal requests.
+   * @param userToken
+   */
+  @ApiCreatedResponse({ description: "Student appeal created successfully" })
```
andrewsignori-aot, line 46: Please add the period.

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  @ApiNotFoundResponse({
+    description:
+      "Application either not found or not eligible to request for change.",
```
andrewsignori-aot, line 49: Can we change to call it an appeal?
`Application either not found or not eligible to request an appeal.`

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  })
+  @ApiUnprocessableEntityResponse({
+    description: "There is an existing appeal for this student.",
+  })
+  @Post("application/:applicationId")
+  async submitStudentAppeal(
+    @Param("applicationId") applicationId: number,
+    @Body() payload: StudentAppealDTO,
+    @UserToken() userToken: IUserToken,
+  ): Promise<void> {
```
andrewsignori-aot, line 59: In general, the POST methods are returning the id of the generated resource.

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
dheepak-aot: Actually we are trying resolve the results but we don't parse the response. While submitting student appeal request, each appeal is a form and we are doing dryRunSubmission for array of forms and waiting for the results of all the forms. Until dryrun is successfully completed for all the forms, we do not move further. 
So await of promise.all is required here.

andrewsignori-aot: I like the way that it is executed in parallel, great work, but we do need to inspect the results, like the below, right?
image

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+    const application = this.applicationService.getApplicationToRequestChange(
```
andrewsignori-aot, line 60: Can we replace RequestChange for appeal?

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
dheepak-aot: @andrewsignori-aot Yes. that is missed. will add that part.

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+      undefined,
```
andrewsignori-aot, line 61: If this parameter is optional it should be marked as optional, right?

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
dheepak-aot: As select 1 instead of any column(which is part of entity), getOne will always return undefined irrespective of the results. hence getRawOne is used here.

image

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1164,4 +1164,43 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .andWhere("application.applicationStatus = :completed", {
+        completed: ApplicationStatus.completed,
+      });
+    if (applicationId) {
+      applicationQuery.andWhere("application.id = :applicationId", {
+        applicationId,
+      });
+    } else {
+      applicationQuery.andWhere(
+        "application.applicationNumber = :applicationNumber",
```
andrewsignori-aot, line 1198: It seems that the `applicationNumber` is optional sometimes, please check if it is defined before use it.

sources/packages/api/src/services/student-appeal/student-appeal.model.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Service model for student appeal.
+ */
+export class StudentAppealRequestModel {
```
andrewsignori-aot, line 4: Please change to an interface unless there is a reason to keep it as a class.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Service layer for Student appeals.
+ */
+@Injectable()
+export class StudentAppealService extends RecordDataModelService<StudentAppeal> {
+  constructor(connection: Connection) {
+    super(connection.getRepository(StudentAppeal));
+  }
+
+  /**
+   * Save student appeals that are requested by the student
```
andrewsignori-aot, line 24: Please add the missing period.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @param applicationId
+   * @param userId
```
andrewsignori-aot, line 26: Can we add some commnets? For instance, for user one we did the below in the past.
image

sources/packages/web/src/components/common/StudentAppealForm.vue
```diff
@@ -0,0 +1,23 @@
...
+export default {
+  components: { formio },
+  emits: ["appealFormLoaded"],
+  props: {
+    formName: {
+      type: String,
+      required: true,
+    },
+  },
+  setup(props: any, context: any) {
```
andrewsignori-aot, line 16: Please declare context with the proper `SetupContext` type.

sources/packages/web/src/components/common/StudentAppealForm.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
andrewsignori-aot, line 1: We talked briefly about this component receiving an array of _formNames_ and doing the "for" to render all of them internally. You did not pursue this approach for some specific reason?

sources/packages/web/src/constants/routes/RouteConstants.ts
```diff
@@ -22,6 +22,7 @@ export const StudentRoutesConst = {
   NOTIFICATIONS_SETTINGS: Symbol(),
   STUDENT_APPLICATION_SUMMARY: Symbol(),
   STUDENT_APPLICATION_DETAILS: Symbol(),
+  STUDENT_REQUEST_CHANGE: Symbol(),
```
andrewsignori-aot, line 25: We need to decide if internally we would call this a "Request a Change" or Appeal.
From the DB perspective, we called everything an Appeal, so I believe that we can stick with it.
But I do understand that the UI is calling it "Student Request a Change", so on the Vue application maybe we can it as it is. Open for thoughts 😉

sources/packages/web/src/types/contracts/student/ApplicationContract.ts
```diff
@@ -76,14 +76,20 @@ export interface ApplicationStatusToBeUpdatedDto {
...
+export interface ApplicationPrimaryDTO {
+  id: number;
+  applicationNumber: string;
+}
+
 /**
  * Base DTO for application dynamic data
  */
-export interface GetApplicationBaseDTO {
+export interface GetApplicationBaseDTO extends ApplicationPrimaryDTO {
```
andrewsignori-aot, line 90: I am missing why `GetApplicationBaseDTO` is now extending the `ApplicationPrimaryDTO`. Can you please clarify it?

sources/packages/web/src/types/contracts/student/ApplicationContract.ts
```diff
@@ -138,11 +144,9 @@ export interface StudentApplicationDetailsForTracking {
 /**
  * Interface for application summary
  */
-export interface ApplicationSummaryDTO {
-  applicationNumber: string;
+export interface ApplicationSummaryDTO extends ApplicationPrimaryDTO {
```
andrewsignori-aot, line 147: I am missing why GetApplicationBaseDTO is now extending the ApplicationPrimaryDTO. Can you please clarify it?

sources/packages/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
dheepak-aot: For the StudentAppealRequestDTO the dryRunSubmission is happening and it is happening in loop, that is the reason why I did not add class-validators to it's properties. But the dry run does not check if at least one object is present in array for StudentAppealDTO. Hence I added class validator only there.

sources/packages/web/src/views/student/StudentRequestChange.vue
```diff
@@ -0,0 +1,166 @@
...
+    };
+
+    const appealFormLoaded = (form: any) => {
+      appealForms.push(form);
+    };
+
+    const submitAppeal = async () => {
+      const studentAppealRequests = [] as StudentAppealRequestDTO[];
+      appealForms.forEach((form: any) => {
+        form.submit();
```
andrewsignori-aot, line 125: What does this call do?

sources/packages/forms/src/form-definitions/partnerincomeappeal.json
```diff
@@ -0,0 +1,407 @@
+{
+    "title": "Partner Income Appeal",
+    "name": "partnerIncomeAppeal",
```
andrewsignori-aot, line 3: As a suggestion, can we start all appeal forms with some suffix like _appeal_ to have them all grouped-ish by name? For instance:
```
"title": "Appeals - Partner Income",
"name": "appealsPartnerIncome",
"path": "appealspartnerincome"
```

sources/packages/web/src/components/common/StudentAppealForm.vue
dheepak-aot: May be there is a gap in communication. I was in a understanding that we are using a component to represent an appeal form and iterate this component with the list of formnames we get. We can sync up to get it cleared out.

sources/packages/forms/src/form-definitions/studentrequestchange.json
```diff
@@ -0,0 +1,696 @@
+{
```
andrewsignori-aot, line 1: Please ensure that the below option is checked.
image

@ann-aot @henryoforeh-aot @guru-aot just a reminder, always that we have a "list of values" like dropdowns, checkboxes, radio button lists, please ensure that the above option is checked.

andrewsignori-aot (commented):

sources/packages/web/src/constants/routes/RouteConstants.ts
dheepak-aot: Yes on the UI perspective I have kept the context as request for change. And in API perspective it is been kept as appeal. That seems to be ok IMO.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: The special comments are fine if the swagger can deal with them, as per your screenshot, it does nicely 😉 
I was afraid that the additional "*" would be in the middle of the comments.
Let's keep using them as much as needed 👍

sources/packages/web/src/views/student/StudentRequestChange.vue
dheepak-aot: When submit button is clicked, this submits the array of forms and shows the pristine status in UI. 
image

sources/packages/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: We had the same conversation in the past and, if I am not wrong, it was actually between you and @ann-aot when we talked about adding the cast in the return of the controller. Does it ring any bell?
I do not get where this fits into my comment: "Remember when we spoke once about assigning a dto instance to service method where service expects a service model it automatically casted"

sources/packages/forms/src/form-definitions/partnerincomeappeal.json
dheepak-aot: Currently all appeal forms(part of this PR) are suffixed with appeal. You meant prefix?

sources/packages/web/src/views/student/StudentRequestChange.vue
```diff
@@ -0,0 +1,166 @@
...
+        }
+        toast.error("Unexpected error", errorMessage);
+      }
+    };
+
+    const submitStudentRequest = async () => {
+      return requestFormData.submit();
+    };
+
+    const backToRequest = () => {
```
ann-aot, line 113: the name and logic inside does make sense to me

sources/packages/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: I got it, can we then rename it to `ApplicationIdentifiersDTO`?

sources/packages/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
andrewsignori-aot: Not a proper reason to skip the validation here. The dry run is validating part of the DTO that, for instance, needs the _formName_ that must be present. Adding the validators here will avoid, for instance, a useless call for a _dryRun_.

sources/packages/web/src/components/common/StudentAppealForm.vue
andrewsignori-aot: We talked about two possible ways:
1. Do the v-for where it is right now
2. Try to do it inside the component.
The idea was to prepare the component to be reused for the Ministry, right?

sources/packages/web/src/views/student/StudentRequestChange.vue
andrewsignori-aot: So, when it fails it interrupts the forEach and the `StudentAppealService.shared.submitStudentAppeal` will not happen?

sources/packages/forms/src/form-definitions/partnerincomeappeal.json
andrewsignori-aot: Yes, right, prefix as per the example posted.

sources/packages/web/src/views/student/StudentRequestChange.vue
dheepak-aot: Doesn't make sense? yes, I get your feeling. This is a 2 stage request cycle where you enter the application number and select appeals in stage 1 and then move to stage 2 where you enter appeal specific details. Without v-stepper we are showing this with hide and view. So basically going back to previous view needs to re-initialize the values.

sources/packages/web/src/views/student/StudentRequestChange.vue
```diff
@@ -0,0 +1,166 @@
...
+      } catch (error) {
+        let errorMessage = "An error happened while requesting a change.";
+        if (error.response.data?.errorType === INVALID_APPLICATION_NUMBER) {
+          errorMessage = error.response.data.message;
+        }
+        toast.error("Unexpected error", errorMessage);
+      }
+    };
+
+    const submitStudentRequest = async () => {
```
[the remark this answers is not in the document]
andrewsignori-aot: I would be willing to hear a further explanation on this.

sources/packages/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
dheepak-aot: IMO dry definitely required because the formdata can be anything and is dynamic so we cannot do a perfect validation with class validators for formData. On the formname, why I did not do validation because formname is also a participant of dryrun and i believe unless it is perfect, dryrun will not be successful. Still ok to add class validator for formName but felt it can be skipped as well.

guru-aot (approved):
LGTM, @dheepak-aot  nice work on the changes done as discussed

sources/packages/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
andrewsignori-aot: Thanks for doing the change as we talked 😉

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,119 @@
...
+      throw new UnprocessableEntityException(
+        "There is already a pending appeal for this student.",
+      );
+    }
+    try {
+      const dryRunPromise = payload.studentAppealRequests.map((appeal) =>
+        this.formService.dryRunSubmission(appeal.formName, appeal.formData),
+      );
+      const submissionResults = await Promise.all(dryRunPromise);
+      submissionResults.forEach((result) => {
```
andrewsignori-aot, line 93: I would suggest using `.some` for this, it seems to be more suitable since we just need to know if there is at least one.

sources/packages/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (!result.valid) {
+          throw new CustomNamedError(
```
andrewsignori-aot, line 95: Using the `CustomNamedError` seems to be an overread in this case, IMO. We can use 2 try/catch and handles the 2 possible different exceptions.

sources/packages/web/src/components/common/StudentAppealForm.vue
andrewsignori-aot: As per the conversation, let's try to have the code centralized as much as possible when the upcoming Ministry view came. The effort so far was already pretty good.

ann-aot (approved):
Good Work with student reports a change@dheepak-aot 👍 Thanks for doing the changes 😄

andrewsignori-aot (approved):
Great work, the UI is looking great and we have a lot in place to move on to the Ministry side of it 😉

PR #1139, opened 2022-04-04
End of line to LF and ESLint formatting

dheepak-aot:
### _This PR does not have any code change_
The modified files are a result of converting the end of line character to LF from CRLF.
**Note:** Some of the source code files still were using CRLF and end of line format.

### Steps for DEVS to ensure that your GIT settings and VS Code settings are configured to use LF for end of line

- Create a new file in VS code and check the format. It must be LF as shown here. 

image

If otherwise please make sure your VS code file editor settings are configured to use LF.
image

- Check for core.autocrlf settings in GIT.
Run the following command and it must return false.
`git config --get core.autocrlf`

If the above command returns true, please run the following command
Ref: https://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf
`git config --local core.autocrlf false`

Now validate again with `git config --get core.autocrlf` to ensure your command ran successfully. The command must return false.


andrewsignori-aot (approved):
Thanks for doing the changes globally. Looks good 👍

henryoforeh-aot (approved):

guru-aot (approved):
LGTM, Nice work

PR #1140, opened 2022-04-05
[Api][SIMS #925] Program Information Requestion application overlap

henryoforeh-aot:
I refactored **application.students.controller** by moving the method **validateOverlappingDatesAndPIR** to application service to make it accessible to both **application.students.controller** and **program-info-request.controller**.


sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1288,4 +1299,71 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+   * Validation for application overlapping dates or Pending PIR.
+   * This validation can be disabled by setting BYPASS_APPLICATION_SUBMIT_VALIDATIONS to true in .env file.
+   * @param applicationId
+   * @param lastName
+   * @param userId
+   * @param sin
+   * @param birthDate
+   * @param studyStartDate
+   * @param studyEndDate
+   * @param bypassApplicationSubmitValidations
```
ann-aot, line 1313: looks like `bypassApplicationSubmitValidations` is not defined

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -75,6 +82,10 @@ export class ApplicationService extends RecordDataModelService<Application> {
     this.logger.log("[Created]");
   }
 
+  setBypassApplicationSubmitValidations(payload: boolean): void {
```
ann-aot, line 85: do we need this? we can directly use `this.config.bypassApplicationSubmitValidations` right? or am i missing anything

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1288,4 +1299,71 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    userId: number,
+    sin: string,
+    birthDate: Date,
+    studyStartDate: Date,
+    studyEndDate: Date,
+  ): Promise<void> {
+    if (!this.bypassApplicationSubmitValidations) {
+      const existingOverlapApplication = this.validatePIRAndDateOverlap(
+        userId,
+        new Date(studyStartDate),
```
ann-aot, line 1327: do we need `new Date` transformation? the type is already date right? line 1321 or 1222

sources/packages/api/src/services/application/application.service.ts
henryoforeh-aot: I used that before but I had code smells caused by more than 7 parameters in the validateOverlappingDatesAndPIR method. I also tried injecting ConfigService in constructor and still had sonar code smell in because we already have more than 7 injections in some of the constructors of the controller where it is required.

andrewsignori-aot: @henryoforeh-aot  If the code smell is only because of the number of injected parameters we can override Sonarcloud validations before doing the merge. If not necessary, we can avoid doing changes just to make the Sonarcloud pass 😉

ann-aot: maybe to resolve the code smell you can pass student details as a single object, something like `async validateOverlappingDatesAndPIR(
    applicationId: number,
    {
         lastName: string,
         userId: number,
         sin: string,
         birthDate: Date,
    },
         studyStartDate: Date,
         studyEndDate: Date,
  )`

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -214,13 +207,19 @@ export class ApplicationStudentsController extends BaseController {
       userToken.userId,
     );
 
-    await this.validateOverlappingDatesAndPIR(
+    this.applicationService.setBypassApplicationSubmitValidations(
+      this.config.bypassApplicationSubmitValidations,
```
andrewsignori-aot, line 211: I would recommend moving this inside the constructor and setting the variable as read-only since it will not change.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -214,13 +207,19 @@ export class ApplicationStudentsController extends BaseController {
       userToken.userId,
     );
 
-    await this.validateOverlappingDatesAndPIR(
+    this.applicationService.setBypassApplicationSubmitValidations(
```
andrewsignori-aot, line 210: I would recommend moving this logic inside the `ApplicationService`.
I do not see a reason to create a method for it, while executing the method `validateOverlappingDatesAndPIR` the configuration can be directly checked. This is a very specific setting that is supposed to be used only for development purposes.

andrewsignori-aot (commented):
Overall looks good, nice work. Please take a look at the comments.
It seems that there is a need to change the way that the bypass is being consumed.
IMO it should be handled inside the `ÀpplicationService` only.

sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
```diff
@@ -59,19 +60,23 @@ import {
...
 import BaseController from "../BaseController";
+import { IConfig } from "../../types";
 
 @AllowAuthorizedParty(AuthorizedParties.institution)
 @Controller("institution/location")
 @ApiTags("institution")
 export class ProgramInfoRequestController extends BaseController {
+  private readonly config: IConfig;
   constructor(
+    private readonly configService: ConfigService,
```
dheepak-aot, line 71: I believe configService doesn't need to be a global variable here. We just pull the config from config service.

image

guru-aot (approved):
LGTM, @henryoforeh-aot Nice work

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1288,4 +1299,71 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        existingOverlapApplication,
+        existingSFASFTApplication,
+        existingSFASPTApplication,
+      ]);
+      if (
+        !!applicationResponse ||
+        !!sfasFTApplicationResponse ||
+        !!sfasPTApplicationResponse
+      ) {
+        throw new UnprocessableEntityException(
```
dheepak-aot, line 1363: This is service layer, hence we must throw error/custom error with the highlighted name so that is it captured and translated appropriately at the controller. 
image

dheepak-aot:
Added some comments @henryoforeh-aot. Nice work.

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: @ann-aot my understanding is that the Sonarcloud is complaining about the constructor, not the method. Am I missing something?

dheepak-aot: You're right. Sonar is complaining only about the constructor.

ann-aot (approved):
Good Work @henryoforeh-aot . Thanks for doing the changes 👍

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -214,13 +203,16 @@ export class ApplicationStudentsController extends BaseController {
       userToken.userId,
     );
 
-    await this.validateOverlappingDatesAndPIR(
+    await this.applicationService.validateOverlappingDatesAndPIR(
```
andrewsignori-aot, line 206: Please move this method call should inside the try/catch to generate the proper HTTP exception.

sources/packages/api/src/route-controllers/program-info-request/program-info-request.controller.ts
```diff
@@ -324,6 +324,16 @@ export class ProgramInfoRequestController extends BaseController {
         );
       }
 
+      await this.applicationService.validateOverlappingDatesAndPIR(
```
andrewsignori-aot, line 327: Please add the `INVALID_OPERATION_IN_THE_CURRENT_STATUS` to the array of possible errors that will generate the `UnprocessableEntityException`.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1288,4 +1297,71 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        existingSFASPTApplication,
+      ]);
+      if (
+        !!applicationResponse ||
+        !!sfasFTApplicationResponse ||
+        !!sfasPTApplicationResponse
+      ) {
+        throw new CustomNamedError(
+          PIR_OR_DATE_OVERLAP_ERROR_MESSAGE,
+          INVALID_OPERATION_IN_THE_CURRENT_STATUS,
```
andrewsignori-aot, line 1362: I believe that PIR_OR_DATE_OVERLAP_ERROR would be better than INVALID_OPERATION_IN_THE_CURRENT_STATUS, right?

andrewsignori-aot: Actually, even `PIR_OR_DATE_OVERLAP_ERROR ` does not seem right 😄 
This would be simple APPLICATION_DATE_OVERLAP_ERROR, IMO.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -67,10 +69,10 @@ export const INSUFFICIENT_APPLICATION_SEARCH_PARAMS =
 export class ApplicationService extends RecordDataModelService<Application> {
   @InjectLogger()
   logger: LoggerService;
-  private bypassApplicationSubmitValidations: boolean;
-
+  private readonly config: IConfig;
   constructor(
     connection: Connection,
+    private readonly configService: ConfigService,
```
dheepak-aot, line 74: As per my previous comment, configService instance does not need to be global variable.

dheepak-aot (approved):
LGTM @henryoforeh-aot . Added one more comment. It is not a blocker, but please have a look.

andrewsignori-aot (approved):
Thanks for doing the changes. Please take a look at the last comment from @dheepak-aot.

PR #1173, opened 2022-04-22
[Web][Api][SIMS#1143] Implement Country Provinces

ann-aot:
- tried to make an address a common object.
- added a hidden country and postal code, which is used inside API.
- added new address format to
1. institutionprofilecreation
2. institutionlocation
3. institutionprofile
4. studentInformation

**_screenshots_**
image
image


sources/packages/web/src/services/http/dto/Common.dto.ts
```diff
@@ -65,3 +53,18 @@ export interface InstitutionPrimaryContactAPIOutDTO {
   primaryContactLastName: string;
   primaryContactPhone: string;
 }
+
+/**
+ * Common DTO for address details.
+ */ export interface AddressDetailsFormAPIDTO {
```
henryoforeh-aot, line 59: Can you please move this to a new line ?

ann-aot: mm, actually that is a line between 2 interfaces

sources/packages/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
```diff
@@ -223,6 +222,10 @@ export class InstitutionLocationInstitutionsController extends BaseController {
       primaryContactLastName: institutionLocation.primaryContact.lastName,
       primaryContactEmail: institutionLocation.primaryContact.email,
       primaryContactPhone: institutionLocation.primaryContact.phoneNumber,
+      canadaPostalCode: institutionLocation.data.address.canadaPostalCode,
```
andrewsignori-aot, line 225: I got the point that UI needs a second field to deal with the _Canadian/not Canadian_ postal code, but I do not believe that it should be reflected in the database.
IMO, DB should keep only `postalCode` and it is valid also for the `country`.

sources/packages/api/src/route-controllers/models/common.dto.ts
```diff
@@ -15,14 +15,30 @@ export class AddressAPIInDTO {
...
   provinceState: string;
   @IsNotEmpty()
   country: string;
   @IsNotEmpty()
   postalCode: string;
+  @IsOptional()
+  // This property will have canada postal code.
+  // This property is for formIO purpose.
+  @IsOptional()
+  canadaPostalCode?: string;
```
andrewsignori-aot, line 28: Can we check the possibility to use the conditional validators here: https://github.com/typestack/class-validator#conditional-validation.

sources/packages/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
ann-aot: when the submitted form is viewed again we need these values also need to be populated. in the case of Pincode, canadaPostalCode and otherPostalCode are different. Canada has a mask and the other is free-text. are you suggesting to calculate those in API or in vue side?

sources/packages/api/src/route-controllers/models/common.dto.ts
ann-aot: I like this conditional validator. actually, I next story is the continuation of this, in the story, I will be validating the form using dry run and I may replace this validators with @Allow()

sources/packages/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
andrewsignori-aot: It could be on API at the controller level, IMO.
Do you think that it is valid to save these additional fields on DB?
For instance, when generating the e-Cert files, I would like to avoid checking where the country is saved for instance. I would say that it would be better to have a single property on DB that tell us the country.

sources/packages/api/src/utilities/address-utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    province: addressDetails.provinceState,
+    country: addressDetails.country,
+    city: addressDetails.city,
+    postalCode: addressDetails.postalCode,
+    canadaPostalCode: addressDetails.canadaPostalCode,
+    otherPostalCode: addressDetails.otherPostalCode,
+    selectedCountry: addressDetails.selectedCountry,
+    otherCountry: addressDetails.otherCountry,
+  };
+}
```
henryoforeh-aot, line 24: A \re all the attributes optional here ?

sources/packages/api/src/utilities/address-utils.ts
```diff
@@ -0,0 +1,24 @@
+import { AddressDetailsModel } from "../services";
+import { AddressInfo } from "../types";
+
+/**
+ * Util to transform address details.
+ * @param addressDetails is a partial AddressDetailsModel.
+ * @returns address info as AddressInfo.
+ */
+export function transformAddressDetails(
+  addressDetails: Partial<AddressDetailsModel>,
```
henryoforeh-aot, line 10: Will all the attributes coming from addressDetails be optional ?

sources/packages/web/src/services/http/dto/Common.dto.ts
henryoforeh-aot: Is the DTO for in or out ?

sources/packages/api/src/route-controllers/models/common.dto.ts
```diff
@@ -15,14 +15,30 @@ export class AddressAPIInDTO {
   addressLine2: string;
   @IsNotEmpty()
   city: string;
-  @IsNotEmpty()
+  @IsOptional()
```
guru-aot, line 18: can we please add provinceState? as its optional?

sources/packages/web/src/services/http/dto/Common.dto.ts
ann-aot: it's used in both IN and OUT, that's why I didn't specify IN or OUT. let me know devs thoughts

sources/packages/api/src/route-controllers/models/common.dto.ts
ann-aot: mm, I didn't understand @guru-aot

sources/packages/api/src/utilities/address-utils.ts
ann-aot: good point. will fix this

henryoforeh-aot (approved):

guru-aot (approved):
LGTM, Nice work @ann-aot

dheepak-aot (approved):
LGTM 👍

sources/packages/api/src/services/address/address.models.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Service model for address details.
+ */
+
```
andrewsignori-aot, line 4: Please remove extra line.

sources/packages/web/src/services/http/dto/Common.dto.ts
andrewsignori-aot: This is supposed to reflect the API classes. If the API has only one we can keep only one and vice-versa.

sources/packages/api/src/utilities/address-utils.ts
```diff
@@ -0,0 +1,84 @@
...
+    postalCode: addressDetails.postalCode,
+    selectedCountry: addressDetails.selectedCountry,
+  };
+}
+/**
+ * Util to transform address details for formIO.
+ * @param addressDetails is a partial AddressDetailsModel.
+ * @returns address info as AddressDetailsAPIOutDTO.
+ */
+export function transformAddressDetailsForForm(
```
andrewsignori-aot, line 28: It seems that this class belongs to the DTO scope only.
The utils are supposed to be something that falls under a helper that will be common and can be used in any layer.

ann-aot: mm, are you suggesting to move it to the dto file?

andrewsignori-aot (approved):
Nice work and thanks for doing the changes. There are just 2 minor comments that can be handled in the upcoming refactor.

sources/packages/api/src/utilities/address-utils.ts
andrewsignori-aot: I would say yes because it is specifically doing something for DTOs.

ann-aot: okay. I will update this in my future PR

Issue #1179, opened 2022-04-27
MSFAA and ECert File Issues


abschwenker:
**Acceptance Criteria**
- Resolve the following issues for MSFAA and ecert send files
- [x] Change sequence number to ensure it is starting at 1 for each day (subsequent generations on the same day should be 2,3,4,etc)
- [x] Ensure the header in the file matched above change
- [x] user FELIX was not working as his firstname was NULL. Ensure students can be sent with NULL givenName. Fill with spaces
- [x] REMOVE .DAT at end of file, not actually used in send files of ESDC files(MSFAA and ECERT)

PR #1180, opened 2022-04-27
[Bug][SIMS-#1179]MSFAA and ECerts File sent bugs

guru-aot:
- Resolve the following issues for MSFAA and ecert send files
- [x] Change sequence number to ensure it is starting at 1 for each day (subsequent generations on the same day should be 2,3,4,etc)
- [x] Ensure the header in the file matched above change
- [x] user FELIX was not working as his firstname was NULL. Ensure students can be sent with NULL givenName. Fill with spaces
- [x] REMOVE .DAT at end of file, not actually used in send files of ESDC files(MSFAA and ECERT)

### DIFF
--- a/sources/packages/api/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
+++ b/sources/packages/api/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
@@ -17,6 +17,7 @@ import {
   ECERT_PART_TIME_FEEDBACK_FILE_CODE,
   getDayOfTheYear,
   getFieldOfStudyFromCIPCode,
+  getISODateOnlyString,
 } from "../../utilities";
 import { EntityManager } from "typeorm";
 import { ESDCFileHandler } from "../esdc-file-handler";
@@ -148,7 +149,7 @@ export class ECertFileHandler extends ESDCFileHandler {
     const now = new Date();
     const dayOfTheYear = getDayOfTheYear(now);
     await this.sequenceService.consumeNextSequence(
-      sequenceGroup,
+      `${sequenceGroup}_${getISODateOnlyString(new Date())}`,
       async (nextSequenceNumber: number, entityManager: EntityManager) => {
         try {
           this.logger.log(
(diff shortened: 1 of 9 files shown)


sources/packages/api/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
```diff
@@ -179,11 +180,11 @@ export class ECertFileHandler extends ESDCFileHandler {
           this.logger.log(`Uploading ${offeringIntensity} content...`);
           await eCertIntegrationService.uploadContent(
             fileContent,
-            fileInfo.filePath,
+            fileInfo.filePath.slice(0, -4),
```
andrewsignori-aot, line 183: Is this code removing the file extension? If it is there are specific libraries to do it, please check: https://nodejs.org/api/path.html#path_path_parse_path

Why do we need to remove the extension here?

sources/packages/api/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-detail.ts
```diff
@@ -42,7 +42,7 @@ export class MSFAAFileDetail implements MSFAARequestFileLine {
     record.appendDate(this.birthDate, DATE_FORMAT);
     record.appendDate(this.processDate, DATE_FORMAT);
     record.appendWithEndFiller(this.surname, 25, SPACE_FILLER);
-    record.appendWithEndFiller(this.givenName, 15, SPACE_FILLER);
+    record.appendWithEndFiller(this.givenName || "", 15, SPACE_FILLER);
```
andrewsignori-aot, line 45: Please use the "??" that will ensure that "" will be used specifically for `null` or `undefined`.

sources/packages/api/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
ann-aot: i also has the same question

guru-aot: Yeah, we can do it. Thanks

andrewsignori-aot (approved):
Thanks for doing the changes nd fixes, looks good 👍

ann-aot (approved):
👍

dheepak-aot (approved):
LGTM.

abschwenker (approved):

Issue #612, opened 2021-08-03
Student Authorization Decorator


andrewsignori-aot:
- [ ] Create a decorator for Students that allows the validation to check if a valid user is authenticated and also includes the Student ID on a new @StudentToken
- [ ] Add forms-flow-bpm user id to DB and user token. To later be saved in auditing column

andrewsignori-aot:
Create a decorator for Students that allows the validation to check if a valid user is authenticated and also includes the Student ID on a new @StudentToken.

PR #1210, opened 2022-05-10
[api][SIMS #612]New decorator and new service account user

andrewsignori-aot:
- Created a new user (when not already created) for the `forms-flow-bpm` that will allow saving the audit user also during workflow operations on the SIMS API.
- Added validation for the IDP used for authorization based on the authorized party. It will basically enforce that a user logged in to the Institution was authenticated using BCeID, a student was authenticated using BCSC, and so on. **THE SAME VALIDATION ALREADY EXISTS ON THE VUE APPLICATION BUT IS WAS MISSING ON THE API.**
- Created a new decorator `@RequiresStudentAccount()` that will ensure that there is a student account created and associated with the current user. The decorator can also be used as `@RequiresStudentAccount(false)` in methods that the student must access but the student account is not created yet, for instance, the POST to create the student account itself.
 
image

- The endpoints that need to validate if a current student exists can now skip the validation. For instance, the below one

image

can be refactored as below

image

- Included the audience (`audience: "sims-api"`) also in the token validation and added a new scope to necessary clients on DEV/TEST

image

image

### DIFF
--- a/sources/packages/api/src/auth/guards/authorized-parties.guard.ts
+++ b/sources/packages/api/src/auth/guards/authorized-parties.guard.ts
@@ -68,6 +68,8 @@ export class AuthorizedPartiesGuard implements CanActivate {
         return idp === IdentityProviders.BCeID;
       case AuthorizedParties.aest:
         return idp === IdentityProviders.IDIR;
+      case AuthorizedParties.formsFlowBPM:
+        return true;
       default:
         return false;
     }
--- a/sources/packages/api/src/auth/guards/authorized-parties.guard.ts
+++ b/sources/packages/api/src/auth/guards/authorized-parties.guard.ts
@@ -1,7 +1,14 @@
-import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common";
+import {
+  Injectable,
+  CanActivate,
+  ExecutionContext,
+  ForbiddenException,
+} from "@nestjs/common";
 import { Reflector } from "@nestjs/core";
 import { AuthorizedParties } from "../authorized-parties.enum";
 import { AUTHORIZED_PARTY_KEY } from "../decorators/authorized-party.decorator";
+import { IdentityProviders } from "../identity-providers.enum";
+import { IUserToken } from "../userToken.interface";
 
 /**
  * Inspect the token to check if the correct authorized party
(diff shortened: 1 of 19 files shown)


sources/packages/api/src/services/user/user.model.ts
```diff
@@ -0,0 +1,5 @@
+export class UserLoginInfo {
+  id: number;
+  isActive: boolean;
+  studentId?: number;
+}
```
henryoforeh-aot, line 5: Will it be better to make this an interface since it is not a DTO that needs to be rendered in swagger and also it looks like this model was created for type-checking?

sources/packages/api/src/services/user/user.service.ts
```diff
@@ -20,25 +22,27 @@ export class UserService extends DataModelService<User> {
...
+      .where("user.userName = :userName", { userName })
       .getRawOne();
 
     if (!user) {
-      // When Students and Institutions users logins for the first time
-      // there will no users records until the Institution Profile or
-      // Student Profile is finalized, and this is expected.
+      // When users logins for the first time there will be no users records.
+      // For instance, Students, Institutions Users and Supporting Users, must complete
+      // their profiles in oder to have the user persisted to the database.
```
ann-aot, line 38: typo `oder`

ann-aot (commented):
LGTM 👍  a minor comment and henrys point. We can have a walk-through tomorrow

sources/packages/api/src/services/user/user.model.ts
andrewsignori-aot: That was by mistake. I will change it.

sources/packages/api/src/services/user/user.service.ts
andrewsignori-aot: Fixed!

sources/packages/api/src/services/user/user.model.ts
andrewsignori-aot: Fixed!

ann-aot (approved):
LGTM 👍

henryoforeh-aot (approved):

sources/packages/api/src/auth/guards/authorized-parties.guard.ts
```diff
@@ -19,9 +26,52 @@ export class AuthorizedPartiesGuard implements CanActivate {
...
+   * Determines if the client is authorized through the expected IDP.
+   * @param authorizedParty authorized party type to be checked.
+   * @param idp identity provider used for authentication on Keycloak.
+   * @returns true if IDP is allowed, otherwise, false.
+   */
+  private isAllowedIDP(
+    authorizedParty: AuthorizedParties,
+    idp: IdentityProviders,
+  ): boolean {
+    switch (authorizedParty) {
```
dheepak-aot, line 63: Why are we validating the IDP here? I also did not see the IDP validation in the user story.

guru-aot (approved):
LGTM, thanks for the decorator

sources/packages/api/src/auth/guards/authorized-parties.guard.ts
andrewsignori-aot: If multiple are allowed if just need to change the switch to accommodate it. I changed the switch to show how it looks like. Please take a look.
```typescript
switch (authorizedParty) {
      case AuthorizedParties.student:
      case AuthorizedParties.supportingUsers:
        return [IdentityProviders.BCSC, IdentityProviders.BCeID].includes(idp);
      case AuthorizedParties.institution:
        return idp === IdentityProviders.BCeID;
      case AuthorizedParties.aest:
        return idp === IdentityProviders.IDIR;
      case AuthorizedParties.formsFlowBPM:
        return true;
      default:
        return false;
    }
```

Issue #1205, opened 2022-05-06
FT File Bug


abschwenker:
Acceptance Criteria
**Feedback File**
- [ ] Filename should be TEDU.PBC.FTECERTSFB.20220502
- [ ] Header for the file should have record_type as 100

**FTEcert File**
- [ ] Province and postal code are not being sent and must be
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/d10cf301-cb71-4638-9d4d-629915c31fc4)

PR #1215, opened 2022-05-11
[api][SIMS #1205]provinceState/postalCode optionals

andrewsignori-aot:
- provinceState and postalCode as nullable as per e-Cert file requirements.


ann-aot (approved):
👍 Thanks for doing the changes @andrewsignori-aot

PR #1230, opened 2022-05-18
[vue]Full page container improvements

andrewsignori-aot:
The main idea is to change the `full-page-container` to have a header slot and also allow it to have pre-defined "templates" that are supposed to be common between multiple pages. The idea in continue to improve the component. For now there are some examples of what would be possible.

- We can configure the container to fill the entire page width, as shown below.
image

- We can configure the container to have centralized content without a default `v-card`, as shown below.
image
image

- We can configure the container to have centralized content with a default `v-card`, as shown below.

### DIFF
--- a/sources/packages/web/src/components/aest/StudentApplications.vue
+++ b/sources/packages/web/src/components/aest/StudentApplications.vue
@@ -1,6 +1,5 @@
 <template>
   <!-- This component is shared between ministry and student users -->
-  <p class="category-header-large color-blue">Applications</p>
   <DataTable
     :value="applicationAndCount.applications"
     :lazy="true"
--- a/sources/packages/web/src/components/generic/HeaderNavigator.vue
+++ b/sources/packages/web/src/components/generic/HeaderNavigator.vue
@@ -12,11 +12,16 @@
   </slot>
   <v-row>
     <v-col>
-      <slot name="subTitle">
-        <div class="header-sub-title">
-          {{ subTitle }}
-        </div>
-      </slot>
+      <span class="float-left">
+        <slot name="subTitle">
+          <div class="header-sub-title">
+            {{ subTitle }}
+          </div>
+        </slot>
+      </span>
+      <span class="float-left">
+        <slot name="subTitleDetails"></slot>
+      </span>
     </v-col>
     <v-col>
       <div class="float-right ml-2 header-button">
--- a/sources/packages/web/src/components/generic/HeaderNavigator.vue
+++ b/sources/packages/web/src/components/generic/HeaderNavigator.vue
@@ -1,17 +1,17 @@
 <template>
-  <slot name="title">
-    <div v-if="routeLocation" class="header-title">
-      <a @click="goBack()">
-        <font-awesome-icon :icon="['fas', 'arrow-left']" class="mr-2" />
-        {{ title }}</a
-      >
-    </div>
-    <div v-else class="header-title">
-      {{ title }}
-    </div>
-  </slot>
-  <v-row>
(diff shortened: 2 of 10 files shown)


sources/packages/web/src/components/aest/StudentApplications.vue
```diff
@@ -1,116 +1,121 @@
...
-  >
-    <template #empty>
-      <p class="text-center font-weight-bold">No records found.</p>
-    </template>
-    <Column
-      :field="StudentApplicationFields.ApplicationNumber"
-      sortable="true"
-      header="Application #"
+  <body-header title="Applications" class="m-1"> </body-header>
+  <content-group>
```
ann-aot, line 4: is this for the border?

guru-aot (approved):

sources/packages/web/src/components/aest/StudentApplications.vue
andrewsignori-aot: Yes, it is. This one was missing.

sources/packages/web/src/components/generic/HeaderNavigator.vue
```diff
@@ -1,25 +1,30 @@
...
-        <font-awesome-icon :icon="['fas', 'arrow-left']" class="mr-2" />
-        {{ title }}</a
-      >
-    </div>
-    <div v-else class="header-title">
-      {{ title }}
-    </div>
-  </slot>
-  <v-row>
+  <v-row align="end" class="mb-8">
```
ann-aot, line 2: does `align="end"` matters here? I see both `align-end` and `mb-8` is contradicting. Am i missing anything

sources/packages/web/src/components/layouts/FullPageContainer.vue
```diff
@@ -1,9 +1,53 @@
...
+      </v-row>
+    </template>
   </v-container>
 </template>
+
+<script lang="ts">
+import { computed } from "vue";
+
+enum LayoutTemplates {
+  Centered = "Centered",
```
ann-aot, line 28: I know for enum we discided to use PascalCase for enum, but since these values are related to the layout API property, can we use hyphen-case like `centered and centered-card`. just a thought to know the opinion of the dev. 
Generally in frameworks like vuetify, we use hyphen-case 
image

ann-aot (commented):
Good work @andrewsignori-aot  with the full-page layout :) added a thought

henryoforeh-aot (approved):

sources/packages/web/src/components/generic/HeaderNavigator.vue
andrewsignori-aot: `align="end"` if forcing the button to be aligned to the bottom and better aligned with the text, as shown below.

**With `align="end"`**
image

**Without `align="end"`**
image

`mb-8` is adding space between the header and the below component.

sources/packages/web/src/components/layouts/FullPageContainer.vue
andrewsignori-aot: I believe that would be better, I will change it.

andrewsignori-aot: Enum values updated.

andrewsignori-aot:
> Good work @andrewsignori-aot with the full-page layout :) added a thought and a comment. Maybe we can also add the syntax to the wiki

I can do it once the PR is merged 😉

ann-aot (approved):
👍 Thanks for doing the changes

Issue #1124, opened 2022-03-29
Reorganize API Student Users


abschwenker:
**Additional Context**
- Keep the PRs small
- Just an example. Move in 1 PR, add decorators in another. etc 
- No functionality is changing as part of this ticket

**Acceptance Criteria**
- [ ]  Reorganize API for Student Users to the new structure
- [x] This controller has the file upload, when searching for this controller will affect the form.io definition. 
- [ ] Add API tags for swagger
- [ ] Convert all input DTO to classes
- [ ] Ensure that all input DTO properties have a decorator (@IsOptional, @IsRequired, AND OTHERS).

Some decorators already used in the API
@IsPositive()
@IsNotEmpty()
@IsArray()
@ArrayMinSize(1)
@ValidateNested({ each: true })
@Type(() => DisbursementValueDTO)
@IsEnum(...)
@IsObject()
More: https://docs.nestjs.com/techniques/validation

PR #1232, opened 2022-05-19
[web][api][SIMS #1124]Changing student profile - Part 1

andrewsignori-aot:
- Changes related to getting the student profile. Method and DTO changed to be shared between AEST and Student;
- Some DTOs were renamed and moved to the DTO file on Vue;
- Removal of some files that are not used anymore;
- The country name is being displayed in lower case in the UI. Added a lower case comparison to method `transformAddressDetailsForAddressBlockForm` to allow us to change the country dropdown **code** in form.io later to have Canada in the proper case.


sources/packages/api/src/route-controllers/utils/address-utils.ts
```diff
@@ -16,6 +16,7 @@ export const COUNTRY_CANADA = "canada";
 export function transformAddressDetailsForAddressBlockForm(
   addressDetails: AddressInfo,
 ): AddressDetailsAPIOutDTO {
+  const selectedCountry = addressDetails.selectedCountry.toLowerCase();
```
ann-aot, line 19: may I know the reason to use `toLowerCase();` ?

andrewsignori-aot: The country name is being displayed in lower case in the UI. Added a lower case comparison to method transformAddressDetailsForAddressBlockForm to allow us to change the country dropdown code in form.io later to have Canada in the proper case.

sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -126,17 +112,15 @@ export class StudentFileMetadataAPIOutDTO {
...
 
-export class StudentInfo {
+export class StudentProfileAPIOutDTO {
   firstName: string;
   lastName: string;
+  fullName: string;
   email: string;
   gender: string;
-  dateOfBirth: Date;
+  dateOfBirth: string;
```
henryoforeh-aot, line 121: Any particular reason for changing date of birth to string ?

andrewsignori-aot: This is a date-only value and there is no time associated with it. 
For these cases, mapping to a Date object, that has Date+Time, causes issues and is not accurate.
Always that we have a date-only we should map it to a string and have it parsed as below:
```
dateOfBirth: getISODateOnlyString(student.birthDate)
```

sources/packages/api/src/route-controllers/student/student.aest.controller.ts
```diff
@@ -177,4 +178,23 @@ export class StudentAESTController extends BaseController {
...
+  @Get(":studentId")
+  @ApiNotFoundResponse({ description: "Not able to find the student." })
+  async getStudentProfile(
+    @Param("studentId") studentId: number,
+  ): Promise<StudentProfileAPIOutDTO> {
+    const student = await this.studentControllerService.getStudentProfile(
+      studentId,
+    );
+    if (!student) {
+      throw new NotFoundException("Not able to find the student.");
```
henryoforeh-aot, line 196: "Student profile not found" sounds more like it.

sources/packages/api/src/route-controllers/student/student.aest.controller.ts
```diff
@@ -177,4 +178,23 @@ export class StudentAESTController extends BaseController {
...
     );
   }
+
+  /**
+   * Get the student information that represents the profile.
+   * @param studentId student id to retrieve the data.
+   * @returns student profile details.
+   */
+  @Get(":studentId")
+  @ApiNotFoundResponse({ description: "Not able to find the student." })
```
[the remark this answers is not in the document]
andrewsignori-aot: It seems to me that both are valid ways, is it a suggestion?

sources/packages/api/src/route-controllers/student/student.aest.controller.ts
andrewsignori-aot: It seems to me that both are valid ways, is it a suggestion?

sources/packages/api/src/route-controllers/utils/address-utils.ts
ann-aot: mm, got it... so how about changing the dropdown value itself to `Canada`, so that we can avoid these case transformations? just a thought.

sources/packages/api/src/route-controllers/student/student.aest.controller.ts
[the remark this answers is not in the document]
henryoforeh-aot: Yes, just a suggestion.

sources/packages/api/src/route-controllers/student/student.aest.controller.ts
henryoforeh-aot: Yes

sources/packages/api/src/route-controllers/student/student.students.controller.ts
```diff
@@ -199,4 +200,15 @@ export class StudentStudentsController extends BaseController {
...
+  /**
+   * Get the student information that represents the profile.
+   * @returns student profile information.
+   */
+  @Get()
+  async getStudentProfile(
+    @UserToken() userToken: StudentUserToken,
+  ): Promise<StudentProfileAPIOutDTO> {
+    return this.studentControllerService.getStudentProfile(userToken.studentId);
+  }
```
henryoforeh-aot, line 213: I think you need to throw a "not found" exception when a student's profile is not found.

andrewsignori-aot: For the student, if this code is reached the student was already validated by the guards and validators and we are sure that the student account is present.

henryoforeh-aot: Yeah, makes sense.

henryoforeh-aot (approved):

sources/packages/api/src/route-controllers/student/student.students.controller.ts
```diff
@@ -199,4 +200,15 @@ export class StudentStudentsController extends BaseController {
...
+
+  /**
+   * Get the student information that represents the profile.
+   * @returns student profile information.
+   */
+  @Get()
+  async getStudentProfile(
+    @UserToken() userToken: StudentUserToken,
+  ): Promise<StudentProfileAPIOutDTO> {
+    return this.studentControllerService.getStudentProfile(userToken.studentId);
```
dheepak-aot, line 212: The studentId in user token has made things so easy ❤️

andrewsignori-aot: Agree, it is coming in handy in many controllers 👍

sources/packages/api/src/route-controllers/utils/address-utils.ts
andrewsignori-aot: As we talked about, the form.io definitions will be changed in upcoming PRs and the idea to have the lowerCase is to avoid future issues if someone inadvertently change it.

sources/packages/web/src/services/StudentService.ts
```diff
@@ -119,8 +119,10 @@ export class StudentService {
    * @param studentId
    * @returns StudentDetail
    */
-  async getStudentDetail(studentId: number): Promise<StudentDetail> {
-    return ApiClient.Students.getStudentDetail(studentId);
+  async getAESTStudentProfile(
```
dheepak-aot, line 122: The get method can be named irrespective of client type(getStudentProfile) with studentId as optional parameter.

sources/packages/web/src/services/StudentService.ts
```diff
@@ -142,14 +144,11 @@ export class StudentService {
    * @param studentId student to have the file saved.
    * @param payload list of files to be saved.
    */
-  async saveMinistryUploadedFilesToStudent(
+  async saveAESTUploadedFilesToStudent(
```
dheepak-aot, line 147: same here.

ann-aot (approved):
LGTM 👍Good Work @andrewsignori-aot . Thanks for the clean up 👍

dheepak-aot:
Nice work @andrewsignori-aot  👍
One minor comment on naming.

sources/packages/web/src/services/StudentService.ts
andrewsignori-aot: Yes, it can, but should we do it?
The reasons that I kept two are:
1. We need to adapt the URL for one or another if the student id is present or not;
2. The AEST one can possibly return more data in the future, like SIN number, which would make the method return two DTOs.
Considering the above points, do you believe it would still be worth keeping only one?
I am ok to change it if it is a Team understanding. @ann-aot @guru-aot @henryoforeh-aot - FYI

guru-aot: I believe, I remember a similar comment on the renaming of the method name in the API service side, I changed it. But here this is in the web side and I think it would be better to have it irrespective of the client, until we have a requirement for it. Just my thoughts 😄

dheepak-aot: Agree with @guru-aot . If the difference between both both API calls are growing wider than just studentId/InstitutionId as optional param, then we can start using different client service methods.

guru-aot (approved):

sources/packages/web/src/services/StudentService.ts
andrewsignori-aot: Method is centralized.

sources/packages/web/src/services/StudentService.ts
andrewsignori-aot: Method is centralized.

dheepak-aot (approved):
LGTM @andrewsignori-aot . Thanks for doing the change. 👍

Issue #1145, opened 2022-04-08
Remove Scholastic Standing records for assessment requests


abschwenker:
- [x] In the requests table on the assessment page of the ministry portal (top top table) scholastic standing records is not possible
- [x] In the DB, approved_data, assessed_by, assessed_date no longer used
- [x] ENUM labels pending and declined no longer needed

PR #1250, opened 2022-05-30
 [web][api][SIMS #1145] Remove Scholastic Standing records for assessment requests

ann-aot:
-  In the requests table on the assessment page of the ministry portal (top top table) scholastic standing records is not possible
-  In the DB, approved_data, assessed_by, assessed_date no longer used
-  ENUM labels pending and declined no longer needed
- **screenshots**
image


sources/packages/api/src-sql/StudentScholasticStandings/Remove-scholastic-standing-approval-fields.sql
```diff
@@ -0,0 +1,12 @@
+-- Drop approval realted fields.
```
andrewsignori-aot, line 1: Do we still need `the scholastic_standing_status`? Since we are removing the other ones this one can also be removed.

sources/packages/api/src-sql/Types/Revert-scholastic-standing-status-enum.sql
```diff
@@ -0,0 +1,16 @@
+-- Remove pending and declined from scholastic_standing_status.
```
andrewsignori-aot, line 1: I believe that we can remove the entire type.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -140,6 +140,15 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
           );
         }),
       )
+      .orderBy(
+        `CASE 
+          WHEN EXISTS(${this.studentAppealRequestsService
```
andrewsignori-aot, line 145: Did we try to do this based on the `status` directly?

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -140,6 +140,15 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
...
+            .appealsByStatusQueryObject(StudentAppealStatus.Pending)
+            .getSql()}) THEN 1
```
andrewsignori-aot, line 147: We do have an enum for the sort order:
image

andrewsignori-aot (commented):
Nice work with the refactor, good amount of code simplified. Please take a look at the comments.

sources/packages/api/src-sql/StudentScholasticStandings/Revert-scholastic-standing-approval-fields.sql
```diff
@@ -0,0 +1,24 @@
+-- Revert scholastic standing approval fields.
```
dheepak-aot, line 1: As a suggestion, I would recommend to rename the file as Revert-remove-scholastic-standing-approval-fields.sql to help understand adding columns. Otherwise migration.ts is the place that we need to check for the relevance.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
ann-aot: ya I tried with `status` it was not working

sources/packages/api/src-sql/StudentScholasticStandings/Revert-scholastic-standing-approval-fields.sql
```diff
@@ -0,0 +1,24 @@
...
+ADD
+  COLUMN IF NOT EXISTS approved_data JSONB;
+
+COMMENT ON COLUMN sims.student_scholastic_standings.approved_data IS 'Dynamic form data that represents the final data revised by the Ministry.';
+
+-- Create assessed_by.
+ALTER TABLE
+  sims.student_scholastic_standings
+ADD
+  COLUMN IF NOT EXISTS assessed_by INT REFERENCES sims.users(id) ON DELETE CASCADE;
```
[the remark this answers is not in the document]
ann-aot: `DELETE ` is there for the initial create migration

sources/packages/api/src-sql/Types/Revert-scholastic-standing-status-enum.sql
ann-aot: mm, so are you suggesting to update the below field to a  string 
image

sources/packages/api/src-sql/StudentScholasticStandings/Remove-scholastic-standing-approval-fields.sql
dheepak-aot: I believe in UI data-table(History) we can hardcode the status based on the type.

sources/packages/api/src-sql/Types/Revert-scholastic-standing-status-enum.sql
ann-aot: yup make sense, removing the status column itself

sources/packages/api/src-sql/StudentScholasticStandings/Revert-scholastic-standing-approval-fields.sql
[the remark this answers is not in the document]
dheepak-aot: It shouldn't be a concern on a rollback script IMO.

sources/packages/api/src-sql/Types/Revert-scholastic-standing-status-enum.sql
dheepak-aot: @ann-aot I believe @andrewsignori-aot  is suggesting to remove the column itself scholastic_standing_status which eventually leads to removal of type sims.scholastic_standing_status.

sources/packages/api/src-sql/StudentScholasticStandings/Revert-scholastic-standing-approval-fields.sql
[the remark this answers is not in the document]
dheepak-aot: I was mentioning about ON DELETE CASCADE which could be tricky sometimes. But not in this case.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -140,6 +140,15 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
...
+              ELSE 2
+            END`,
+      )
+      .addOrderBy("studentAppeal.submittedDate", "DESC")
```
dheepak-aot, line 151: Do we still need to use the getRawMany() because it is not a union anymore?

dheepak-aot:
Nice work @ann-aot . Added some comments.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
ann-aot: we were not using union. `getRawMany()` is used because of subqueries

sources/packages/api/src-sql/StudentScholasticStandings/Revert-scholastic-standing-approval-fields.sql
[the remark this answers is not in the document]
andrewsignori-aot: The rollback need put back in place the behavior that was before, if the CASCADE was there then it must be present on the rollback script to put the DB back to its previous state.
That said, we should use `ON DELETE SET NULL` to avoid the unintentional removal of absolutely everything that is related to the user like every single record that the user is present as some sort of audit user.
That said, I asked @abschwenker to check with DBAs if we can get rid of all CASCADING options. It is still pending an answer.

sources/packages/api/src/services/student-appeal/student-appeal.service.ts
andrewsignori-aot: We can use the getRawAndEntities also now as an alternative with the helper to populate the object model. I would say that this would be the preferred way.

sources/packages/api/src-sql/StudentScholasticStandings/Revert-remove-scholastic-standing-approval-fields.sql
```diff
@@ -0,0 +1,24 @@
+-- Revert scholastic standing approval fields.
+-- Create approved_data.
+ALTER TABLE
+  sims.student_scholastic_standings
+ADD
+  COLUMN IF NOT EXISTS approved_data JSONB;
```
andrewsignori-aot, line 6: Nice catch, I forgot about this one 👍

sources/packages/api/src-sql/Types/Remove-scholastic-standing-status-enum-and-col.sql
```diff
@@ -0,0 +1,6 @@
+-- Remove column scholastic_standing_status for student_scholastic_standings.
+ALTER TABLE
```
andrewsignori-aot, line 2: Is there a reason to create a second migration file to remove more columns from the same `sims.student_scholastic_standings` table?

andrewsignori-aot (commented):
Thanks for doing the changes, please take a look at the comments, just minor ones.

henryoforeh-aot (approved):

sources/packages/api/src-sql/Types/Remove-scholastic-standing-status-enum-and-col.sql
ann-aot: I did it to group the enum and dependent column, ya... i can them all in a single file

andrewsignori-aot (approved):

Issue #1198, opened 2022-05-05
Restriction for Scholastic Standing


abschwenker:
**Additional Context**
- When an institution reports a change related to scholastic standing we add certain restrictions

**Additional Context**
- [ ] Institution report Withdrawal for a FT course application, add WTHD restriction to student
- [ ] Institution report Withdrawal for a FT course on a student WITH a WTHD restriction ADD SSR restriction
- [ ] Institution report Withdrawal OR unsuccessful for a PT course application, add PTSSD restriction to student

dheepak-aot:
The account activity button in the menu of the top nav bar does not have icon unlike other buttons in it.  This is because eventually other menu item buttons will also be removed as per discussion with @hellolynn-tbtb 
![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/74263b91-d142-4e69-b4cf-a20835f1bf2f)

PR #1253, opened 2022-06-01
[SIMS #1198] Minor form update

dheepak-aot:
Updated following in studentaccountactivity form

- Added missing text to panel

image

Enabled the property initialize empty array to data grid(otherwise data grid always initializes empty array as [{}] instead of []).

image

- Added a period to text.

image


andrewsignori-aot (approved):
Looks good 👍

Issue #1200, opened 2022-05-05
End Application (soft archive)


abschwenker:
**- Additional Context**
- We need to stop actions form occurring at the end of the application lifecycle

**- Acceptance Criteria**
- [x] Add archive column in application table, boolean
- [x] When an institution user submits a scholastic standing event, change archive flag to true
- [x] Create Scheduler - When 43 days have passed the end of study period, change flag to true
- [x] When archive flag is true, students can no longer request a change to their application (change application number field validation to make archived applications not eligible. If this fails display toast message "This application is no longer eligible to request changes"
- [x] When archive flag is true, Institutions can no longer request a change

PR #1254, opened 2022-06-01
[workflow][api][SIMS #1200] End Application soft archive

henryoforeh-aot:
- Created an environment variable to store number of days passed.
- Created new boolean column in applications table to flag archived applications.
- Implemented functionality to set archived column to true when an institution user submits a scholastic standing event.
- Implemented functionality to throw an error / fail gracefully when either an institution user or student user tries to submit a scholastic standing event / appeal on an archived application.
- Created a new end point to set archived column to true in applications table when 43 days have passed end of study period.
- Created a new scheduler workflow to consume the new end point above.
newscreen1
newscreen2


devops/Makefile
```diff
@@ -20,6 +20,7 @@ export BACKUP_CONFIGMAP_NAME := $(or $(BACKUP_CONFIGMAP_NAME), backup-conf)
 export BYPASS_CRA_INCOME_VERIFICATION := $(or $(BYPASS_CRA_INCOME_VERIFICATION), false)
 export BYPASS_APPLICATION_SUBMIT_VALIDATIONS := $(or $(BYPASS_APPLICATION_SUBMIT_VALIDATIONS), false)
 export SWAGGER_ENABLED := $(or ${SWAGGER_ENABLED}, true)
+export APPLICATION_ARCHIVE_DAYS := $(or ${APPLICATION_ARCHIVE_DAYS}, 43)
```
andrewsignori-aot, line 23: Please search by the `SWAGGER_ENABLED` and check all the places that need to be changed.
There are a few places missing:
1. workflows\deploy.yml
2. workflows\deploytest.yml
3. env-example

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -293,6 +293,15 @@ export class Application extends RecordDataModel {
     { eager: false, cascade: false, nullable: true },
   )
   craIncomeVerifications?: CRAIncomeVerification[];
+
+  /**
+   * Flag to represent when an application is archived
+   */
+  @Column({
+    name: "is_archived",
+    nullable: true,
```
andrewsignori-aot, line 302: Please change to nullable false.

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -293,6 +293,15 @@ export class Application extends RecordDataModel {
     { eager: false, cascade: false, nullable: true },
   )
   craIncomeVerifications?: CRAIncomeVerification[];
+
+  /**
+   * Flag to represent when an application is archived
```
andrewsignori-aot, line 298: Please add a period to the sentence.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -226,4 +227,20 @@ export class ApplicationSystemController extends BaseController {
...
+   */
+  @ApiOkResponse({
+    description: "All application(s) were successfully archived.",
+  })
+  @Patch("archive")
+  async archiveApplication(): Promise<void> {
+    try {
+      await this.applicationService.archiveApplication();
+    } catch (error) {
+      throw new InternalServerErrorException("Internal server error");
```
andrewsignori-aot, line 243: This will be automatically handled by the Nestjs.

sources/packages/api/src/route-controllers/application/application.system.controller.ts
```diff
@@ -226,4 +227,20 @@ export class ApplicationSystemController extends BaseController {
...
+   * Archives one or more applications when 43 days
+   * have passed the end of the study period.
+   */
+  @ApiOkResponse({
+    description: "All application(s) were successfully archived.",
+  })
+  @Patch("archive")
+  async archiveApplication(): Promise<void> {
+    try {
+      await this.applicationService.archiveApplication();
```
andrewsignori-aot, line 241: Please capture the user token and add the user id to the modifier column while updating the applications.

sources/packages/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
```diff
@@ -93,6 +94,13 @@ export class StudentScholasticStandingsService extends RecordDataModelService<St
       );
     }
 
+    if (application.isArchived) {
```
ann-aot, line 97: mm, you can add this logic in the query where condition and adjust line 92 message

sources/packages/web/src/services/http/dto/ScholasticStanding.dto.ts
```diff
@@ -1,6 +1,9 @@
 export const APPLICATION_NOT_FOUND = "APPLICATION_NOT_FOUND";
 export const INVALID_OPERATION_IN_THE_CURRENT_STATUS =
   "INVALID_OPERATION_IN_THE_CURRENT_STATUS";
+export const APPLICATION_CHANGE_NOT_ELIGIBLE =
```
andrewsignori-aot, line 4: There is nothing wrong with this particular code but we have been creating these error constants in different places.
We do have a file called `error-code.constants.ts` on both API and WEB project, can we have an agreement to use these files from now on and move the constants whatever is possible?
@guru-aot @ann-aot @dheepak-aot?

ann-aot: yup, i too agree

sources/packages/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
dheepak-aot: If the logic is added to the query it will throw not found exception instead of unprocessibleEntity.

dheepak-aot:
Nice and clean work @henryoforeh-aot . Had the same comments what other devs mentioned. Please have a look.

sources/packages/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
dheepak-aot: Like-wise the logic inside inside the function is meant only for completed applications but we throw error104 to return Http422(UnProcessible entity). 

Btw. I checked the controller. It is throwing  422 on all the cases. At least the error description is more precise.
image

sources/packages/api/src-sql/Applications/Add-col-archive.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    sims.applications
+ADD
+    COLUMN IF NOT EXISTS is_archived BOOLEAN;
```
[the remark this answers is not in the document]
andrewsignori-aot: Is there a reason for the column not to be defined as NOT NULL as asked in the comment?

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1356,4 +1360,36 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+   * @param auditUserId user making changes to table.
+   */
+  async archiveApplications(auditUserId: number): Promise<void> {
+    const auditUser = { id: auditUserId } as User;
+
+    // Build sql statement to get all application ids to archive
+    const applicationsToArchive = this.repo
+      .createQueryBuilder("application")
+      .select("application.id")
+      .leftJoin("application.currentAssessment", "currentAssessment")
```
andrewsignori-aot, line 1376: Please change both `leftJoin` to `inner`.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1356,4 +1360,36 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .leftJoin("currentAssessment.offering", "offering")
+      .where("application.applicationStatus = :completed")
+      .andWhere(
+        "(CURRENT_DATE - offering.studyEndDate) > :applicationArchiveDays",
+      )
+      .andWhere("application.isArchived <> :isApplicationArchived")
+      .getSql();
+
+    this.repo
```
andrewsignori-aot, line 1385: Thanks for changing the query. 
Please add the `await`.

sources/packages/api/src-sql/Applications/Add-col-archive.sql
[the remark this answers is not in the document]
henryoforeh-aot: Sorry, I forgot. I will update.

sources/packages/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
andrewsignori-aot: @ann-aot @dheepak-aot I personally had done on both ways and received PR comments on both occasions 😄 
I believe that both are valid, as explained below.
For the way that Dheepaks is saying unless there is a performance hit, we can do it when we need to be more specific and provide better individual messages for the errors.
For the way that Ann is describing we can compensate by including in the error message further explanation.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1356,4 +1360,36 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .createQueryBuilder()
+      .update(Application)
+      .set({ isArchived: true, modifier: auditUser })
+      .where(`applications.id IN (${applicationsToArchive})`)
```
dheepak-aot, line 1389: Please test this scenario with an empty array. I believe it won't fail. I saw one of these old issues mentioning it. 
https://github.com/typeorm/typeorm/issues/3630

henryoforeh-aot: Thanks for this, I will test.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1356,4 +1360,36 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .setParameter("completed", ApplicationStatus.completed)
+      .setParameter("applicationArchiveDays", 43)
```
andrewsignori-aot, line 1391: Missed that before, please use the env variable here that is already created.

henryoforeh-aot: Yeah got it.

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1356,4 +1360,36 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        "(CURRENT_DATE - offering.studyEndDate) > :applicationArchiveDays",
+      )
+      .andWhere("application.isArchived <> :isApplicationArchived")
+      .getSql();
+
+    this.repo
+      .createQueryBuilder()
+      .update(Application)
+      .set({ isArchived: true, modifier: auditUser })
+      .where(`applications.id IN (${applicationsToArchive})`)
```
ann-aot, line 1389: may I know the reason why the variable was not set something like this , 
image

dheepak-aot: Nice catch @ann-aot

sources/packages/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
ann-aot: if it's a requirement in the story, then Iam Good with it. My point is  in a normal scenario, as per my understanding, I think this scenario won't happen, because the active applications page will only have the `isArchived=false` applications, and only these applications will have the `report a change ` option unless someone tries to override the  `isArchived=true` application number from the URL (In that case error in line 92 or adjusting msg should be fine, IMO)

sources/packages/api/src/services/application/application.service.ts
henryoforeh-aot: Yeah I will add.

andrewsignori-aot: Sorry, I am not following this, the variable is already set as a parameter `.setParameter("completed", ApplicationStatus.completed)`. What is the point here?

ann-aot: ignore my comments, i missunderstood

andrewsignori-aot: @henryoforeh-aot apparently there was a misunderstanding. There is nothing to be changed related to this.

henryoforeh-aot: Oh I thought your comment was on filtering for applications either in completed or enrollment status.

henryoforeh-aot: Okay thanks for the clarification.

sources/packages/api/src/services/application/application.service.ts
andrewsignori-aot: No need to test it. It will work in the subquery, like below:
```
select * from sims.applications a where a.id in (select id from sims.applications a2 where 1 = 2)
```

sources/packages/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
dheepak-aot: @guru-aot @henryoforeh-aot  please share your thoughts  as well

ann-aot: @dheepak-aot as I mentioned above, this point is part of the story, so, don't have any problem
image, the comment was my opinion 😄 . if you want to talk about it we can have a different call😄

henryoforeh-aot: In my opinion I would leave it where it is.

dheepak-aot: I am good here @ann-aot  😊

configs/env-example
```diff
@@ -6,6 +6,7 @@ BUILD_ID=0
 API_PORT=3000
 APP_PORT=8080
 SWAGGER_ENABLED=true
+APPLICATION_ARCHIVE_DAYS = 43
```
andrewsignori-aot, line 9: Please remove the extra spaces between the `43` to look like the other ones in the same file.

devops/openshift/api-deploy.yml
```diff
@@ -453,3 +455,4 @@ parameters:
     value: gc-notify-to-address
   - name: SWAGGER_ENABLED
     value: "true"
+  - name: APPLICATION_ARCHIVE_DAYS
```
andrewsignori-aot, line 458: If no value will be provided please add a `required: true`

andrewsignori-aot (commented):
Two minor comments, please take a look.

andrewsignori-aot (approved):

guru-aot (approved):
LGTM, Nice work @henryoforeh-aot

ann-aot (approved):
Good Work @henryoforeh-aot 👍

dheepak-aot (approved):
Good work @henryoforeh-aot . 👍

PR #1266, opened 2022-06-09
[workflow]Fix Application Exception

andrewsignori-aot:
- Fix exception duplication removal before submitting the API payload;
- Add checkpoints `assessment-gateway` to make the Ministry approval faster and also provide better troubleshooting.


ann-aot (approved):
👍

Issue #1216, opened 2022-05-12
API test cases


TeresaBia:
**- Additional Context**
- Conduct a POC and Create an API test for the below endpoint

**- Acceptance Criteria**
- [ ] Put testing code in below folder locations
- Source code is located
testing\src 
- Dev Test data is located
testing\src\data\dev
- Test Test data is located
testing\src\data\test
- Prod Test data is located
testing\src\data\prod

- [X] Create Cypress test case for PATCH for /institutions/institution

bhavikccits:
https://github.com/bcgov/SIMS/pull/1237

TeresaBia:
@andrewsignori-aot @abschwenker Can we close this ticket?

andrewsignori-aot:
> @andrewsignori-aot @abschwenker Can we close this ticket?

Fine by me.

PR #1267, opened 2022-06-10
[cypress][SIMS #1216]Updated code & moved testing folder & separated data for dev.

bhavikccits:
Organized the branch as follows

Source code is located
testing\src

Dev Test data is located
testing\src\data\dev

Test Test data is located
testing\src\data\test

Prod Test data is located
testing\src\data\prod


testing/src/cypress/page-objects/Institution-objects/InstitutionProfileObject.ts
```diff
@@ -88,18 +88,22 @@ export default class InstitutionProfileObject {
   }
 
   countryInputText() {
-    return cy.get(".form-control").eq(16);
+    return cy.xpath("//div[@id='e96525o']//div[@role='combobox']");
```
andrewsignori-aot, line 91: Just double checking, what is the reason why `data-cy` is not used here?
Is there a technical issue or the below is not the right one?
image

testing/src/cypress/page-objects/ministry-objects/InstitutionNoteRestrictionsObject.ts
```diff
@@ -134,4 +126,28 @@ export default class InstitutionNoteRestrictionsObject {
...
+  noteDropdown() {
+    return cy.xpath("//div[@role='combobox']");
+  }
+
+  noteValue() {
+    return cy.xpath("//input[@role='textbox']");
+  }
+
+  restrictionsDropdownReason() {
+    return cy.xpath("//div[@id='etm8wyi']//div[@role='combobox']");
```
andrewsignori-aot, line 139: Should I assume that these are not part of the new development and are shown here because of the files reorganization?

bhavikccits: Yes.

andrewsignori-aot (commented):
Folders looks better. Just left minor comments, please take a look.

testing/src/cypress/page-objects/Institution-objects/InstitutionProfileObject.ts
andrewsignori-aot: Why is that? Can you elaborate more, please? The `data-cy` is right on the html `select`, right?

bhavikccits: Reason for "Why data-cy is not possible here in dropdown".

When data-cy is used in a "country" dropdown, elements are not highlighted because data-cy is not added at the proper place to identify those dropdown values.
note-proper


Data-cy is not getting used here, so I am using xpath instead, & you can see that the dropdown is getting highlighted when using xpath selector.
image


Example for where data-cy is used for - In textbox elements, data-cy identifies and highlighted elements when searching with its selector (data-cy).
proper-data-cy

bhavikccits: > Folders looks better. Just left minor comments, please take a look.

Replied. Thanks for quick review.

andrewsignori-aot: As we talked it seems that it is technically possible. I will resolve this assuming that the upcoming PR will explore a better solution.

andrewsignori-aot (approved):
Approving this assuming that the upcoming PR will explore a better solution to deal with the formio dropdowns selectors that are not right now using the `data-cy` as expected.

Issue #298, opened 2021-03-11
Generate disbursement report


abschwenker:
**Additional Context**
- SIMS generate individual types of grant disbursement reports for AEST accounting purposes
- SABC can record accurate expenditures for each program being disbursed through NSLSC
- Same as above - reports must include total expenditure amount and also broken down by individual student, same as report: MRPT_003: JV Report.  This report will allow SABC finance to record expenditures for grants being disbursed via NSLSC
- For this report, SIMS must replicate existing SFAS report MRPT_003: JV Report (see Appendix of analysis doc) with potential UI improvements 
- Use ticket #559 for guidelines

**Acceptance Criteria**
**- Create Report**
- User must have the ability to generate “Individual Grant Disbursement Report” in SIMS
- [x] Separate reports for Full Time and Part Time
- [x] On click of "Export CSV File" (from above UI) generate following CSV
- [x] Report must display each Loan and Grant for each student for the day of disbursement (date range will depend on Ministry user selection)
- Report DETAILS:
- [x] Report Name (Disbursement Report)
- [x] Date and time of report generation (Part of the file name)

- Columns in CSV: 
- [x] Date of Disbursement
- [x] SIN
- [x] Application Number
- [x] Certificate Number
- [x] Funding Code (CSLF, BCAG, BCSG, etc)
- [x] Disbursement amount
- Criteria: Check for received amounts

CSV Example,
Disbursement Forecast Report - YYYYMMDD TT:TT
Date of Disbursement, SIN, Application Number, Certificate Number, Funding Code,Disbursement amount,
2020-AUG-01,123456789,2020000001,000000001,BCAG,16073.00,12
2020-AUG-01,123456789,2020000001,000000001,BGPD,5073.00,15
2020-AUG-01,123456789,2020000002,000000001,CSLF,1236073.00,206
2020-AUG-04,123456789,2020000003,000000001,BGPD,77073.00,88
2020-AUG-04,123456789,2020000003,000000001,BCAG,36073.00,71
2020-AUG-05,123456789,2020000004,000000001,BCSG,10415.00,88

PR #1273, opened 2022-06-15
[FORMS][DB][SIMS-#298]-Generate disbursement reports

guru-aot:
**- Create Report**
- User must have the ability to generate “Individual Grant Disbursement Report” in SIMS
- [x] Separate reports for Full Time and Part Time
- [x] On click of "Export CSV File" (from above UI) generate following CSV
- [x] Report must display each Loan and Grant for each student for the day of disbursement (date range will depend on Ministry user selection)
- Report DETAILS:
- [x] Report Name (Disbursement Report)
- [x] Date and time of report generation (Part of the file name)

- Columns in CSV: 
- [x] Date of Disbursement
- [x] SIN
- [x] Application Number
- [x] Certificate Number
- [x] Funding Code (CSLF, BCAG, BCSG, etc)
- [x] Disbursement amount
- Criteria: Check for received amounts

CSV Example,
Disbursement Forecast Report - YYYYMMDD TT:TT
Date of Disbursement, SIN, Application Number, Certificate Number, Funding Code,Disbursement amount,
2020-AUG-01,123456789,2020000001,000000001,BCAG,16073.00,12
2020-AUG-01,123456789,2020000001,000000001,BGPD,5073.00,15
2020-AUG-01,123456789,2020000002,000000001,CSLF,1236073.00,206
2020-AUG-04,123456789,2020000003,000000001,BGPD,77073.00,88
2020-AUG-04,123456789,2020000003,000000001,BCAG,36073.00,71
2020-AUG-05,123456789,2020000004,000000001,BCSG,10415.00,88

image

### DIFF
--- a/sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
+++ b/sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
@@ -1,44 +1,49 @@
---Insert config for disbursement report.
 INSERT INTO
-    sims.report_configs (report_name, report_sql)
+  sims.report_configs (report_name, report_sql)
 VALUES
-    (
-        'Disbursement_Report',
-        'SELECT "Date of Disbursement",
-            "SIN",
-            "Application Number",
-            "Certificate Number",
-            "Funding Code",
-            "Disbursement amount"
-          FROM
-            (SELECT TO_CHAR(DR.DISBURSE_DATE, ''YYYY-MM-DD'') AS "Date of Disbursement",
-                DR.STUDENT_SIN AS "SIN",
-                SA.APPLICATION_ID AS "Application Number",
-                DS.DOCUMENT_NUMBER AS "Certificate Number",
-                DRV.GRANT_TYPE AS "Funding Code",
-                DRV.GRANT_AMOUNT AS "Disbursement amount",
-                EPO.OFFERING_INTENSITY AS "Offering intensity"
-              FROM SIMS.DISBURSEMENT_RECEIPTS DR
-              INNER JOIN SIMS.DISBURSEMENT_RECEIPT_VALUES DRV ON DRV.DISBURSEMENT_RECEIPT_ID = DR.ID
-              INNER JOIN SIMS.DISBURSEMENT_SCHEDULES DS ON DS.ID = DR.DISBURSEMENT_SCHEDULE_ID
-              INNER JOIN SIMS.STUDENT_ASSESSMENTS SA ON SA.ID = DS.STUDENT_ASSESSMENT_ID
-              INNER JOIN SIMS.EDUCATION_PROGRAMS_OFFERINGS EPO ON EPO.ID = SA.OFFERING_ID
-              UNION ALL SELECT TO_CHAR(DR.DISBURSE_DATE, ''YYYY-MM-DD'') AS "Date of Disbursement",
-                DR.STUDENT_SIN AS "SIN",
-                SA.APPLICATION_ID AS "Application Number",
-                DS.DOCUMENT_NUMBER AS "Certificate Number",
-                CASE
-                        WHEN DR.FUNDING_TYPE = ''BC'' THEN ''BCSL''
-                        WHEN DR.FUNDING_TYPE = ''FE'' THEN ''CSL''
-                END AS "Funding Code",
-                DR.TOTAL_DISBURSED_AMOUNT AS "Disbursement amount",
-                EPO.OFFERING_INTENSITY AS "Offering intensity"
-              FROM SIMS.DISBURSEMENT_RECEIPTS DR
-              INNER JOIN SIMS.DISBURSEMENT_SCHEDULES DS ON DS.ID = DR.DISBURSEMENT_SCHEDULE_ID
-              INNER JOIN SIMS.STUDENT_ASSESSMENTS SA ON SA.ID = DS.STUDENT_ASSESSMENT_ID
-              INNER JOIN SIMS.EDUCATION_PROGRAMS_OFFERINGS EPO ON EPO.ID = SA.OFFERING_ID) AS DATA
-          WHERE "Offering intensity" = ANY(:offeringIntensity)
-            AND "Date of Disbursement" BETWEEN :startDate AND :endDate
-          ORDER BY "Date of Disbursement",
-            "Certificate Number"'
-    );
+  (
+    'Disbursement_Report',
+    '(
+      select
+        to_char(dr.disburse_date, ''YYYY-MM-DD'') as "Date of Disbursement",
+        dr.student_sin as "SIN",
+        sa.application_id as "Application Number",
+        ds.document_number as "Certificate Number",
+        drv.grant_type as "Funding Code",
+        drv.grant_amount as "Disbursement Amount"
+      from
+        sims.disbursement_receipts dr
+        inner join sims.disbursement_receipt_values drv on drv.disbursement_receipt_id = dr.id
+        inner join sims.disbursement_schedules ds on ds.id = dr.disbursement_schedule_id
+        inner join sims.student_assessments sa on sa.id = ds.student_assessment_id
+        inner join sims.education_programs_offerings epo on epo.id = sa.offering_id
+      where
+        epo.offering_intensity = any(:offeringIntensity)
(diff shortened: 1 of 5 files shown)


sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
```diff
@@ -0,0 +1,44 @@
...
+    (
+        'Disbursement_Report',
+        'SELECT "Date of Disbursement",
+            "SIN",
+            "Application Number",
+            "Certificate Number",
+            "Funding Code",
+            "Disbursement amount"
+          FROM
+            (SELECT TO_CHAR(DR.DISBURSE_DATE, ''YYYY-MM-DD'') AS "Date of Disbursement",
```
andrewsignori-aot, line 14: I missed it in the previous PR, it is the double quotes on  `''YYYY-MM-DD''` needed because of Typeorm?
@dheepak-aot

guru-aot: Yes @andrewsignori-aot it translates to single quotes, when the query is inserted during migration.

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
```diff
@@ -0,0 +1,44 @@
...
+                        WHEN DR.FUNDING_TYPE = ''BC'' THEN ''BCSL''
+                        WHEN DR.FUNDING_TYPE = ''FE'' THEN ''CSL''
+                END AS "Funding Code",
+                DR.TOTAL_DISBURSED_AMOUNT AS "Disbursement amount",
+                EPO.OFFERING_INTENSITY AS "Offering intensity"
+              FROM SIMS.DISBURSEMENT_RECEIPTS DR
+              INNER JOIN SIMS.DISBURSEMENT_SCHEDULES DS ON DS.ID = DR.DISBURSEMENT_SCHEDULE_ID
+              INNER JOIN SIMS.STUDENT_ASSESSMENTS SA ON SA.ID = DS.STUDENT_ASSESSMENT_ID
+              INNER JOIN SIMS.EDUCATION_PROGRAMS_OFFERINGS EPO ON EPO.ID = SA.OFFERING_ID) AS DATA
+          WHERE "Offering intensity" = ANY(:offeringIntensity)
```
andrewsignori-aot, line 40: How DB is resolving this with the `where` clause outside the queries?
Is it going to resolve both queries and then apply the result to the final combination of both?
If this is the case, the execution plan would not be better for executing the where on each one of the queries?
Was it checked?

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
```diff
@@ -0,0 +1,44 @@
...
+        'SELECT "Date of Disbursement",
+            "SIN",
+            "Application Number",
+            "Certificate Number",
+            "Funding Code",
+            "Disbursement amount"
+          FROM
+            (SELECT TO_CHAR(DR.DISBURSE_DATE, ''YYYY-MM-DD'') AS "Date of Disbursement",
+                DR.STUDENT_SIN AS "SIN",
+                SA.APPLICATION_ID AS "Application Number",
```
andrewsignori-aot, line 16: The query is not following the same format as the previous one already in place. For instance, this one is using the DB columns/aliases in all upper cases while the previous ones are not.

sources/packages/api/src-sql/Reports/Drop-disbursement-report.sql
```diff
@@ -0,0 +1,2 @@
+--Insert config for disbursement report.
```
andrewsignori-aot, line 1: The comment seems wrong. I personally will not request a comment to be added to a so self-explanatory line 😉 
IMO it can be removed.

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
```diff
@@ -0,0 +1,44 @@
+--Insert config for disbursement report.
+INSERT INTO
+    sims.report_configs (report_name, report_sql)
+VALUES
+    (
+        'Disbursement_Report',
+        'SELECT "Date of Disbursement",
```
andrewsignori-aot, line 7: Is there a reason to add this select on the top of the 2 queries?

andrewsignori-aot, line 7: Was this query formatted with the same formater that we use in VSCode. If I format it on my VS Code it looks different.

andrewsignori-aot (commented):
Good work, please take a look at the comments.

sources/packages/forms/src/form-definitions/exportfinancialreports.json
```diff
@@ -115,6 +110,10 @@
                       {
                         "label": "Forecast disbursements",
                         "value": "Disbursement_Forecast_Report"
+                      },
+                      {
+                        "label": "Disbursements",
+                        "value": "Disbursement_Report"
```
ann-aot, line 116: I know the `"Disbursement_Forecast_Report"` is always not following camelCase.  I believe we decided to use camelCase in both camunda and forms too. but may I know the reason, not to use camelCase?

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
guru-aot: Yes its resolving, both the queries and apply the where clause to the result. I checked, it is the same as executing them as separate queries.

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
guru-aot: Yes, as we are following the report generation using 3 params -> offering intensity, start date and end date of disbursement. When we create 2 where clause inside the union queries, it is expecting 6 params, 3 for the first set and 3 for the second. So I tested the scenarios, with the where clause inside and outside and the results are the same. So created the overall select at the top.

guru-aot: We can connect, it is not working for me

sources/packages/forms/src/form-definitions/exportfinancialreports.json
guru-aot: The reason it would have happened is, formio auto creats the id when the value is created in the data tab. We can override it, so I followed it as it was before. Lets discuss and change. @andrewsignori-aot @henryoforeh-aot @dheepak-aot what you guys feel to change the id here?

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
guru-aot: As discussed changed

sources/packages/api/src-sql/Reports/Create-disbursement-report.sql
guru-aot: As discussed changed

guru-aot: As discussed changed

sources/packages/forms/src/form-definitions/exportfinancialreports.json
andrewsignori-aot: This one is targeting a DB column "unique identifier" (which could also be defined as camelCase). Maybe because it is not targeting a variable name we can keep it as it is. I believe that it is also used to name the downloaded file.
That said, I missed that in the previous PR and I am not sure if there is a strong reason to have this format, even using the "_" that is not so common in the solution while naming whatever. Maybe the file name is the only possible explanation.

andrewsignori-aot (approved):
Thanks for changing the query and working to fix the parameters code issue. Looks good 👍

ann-aot (approved):
👍 Good Work @guru-aot

henryoforeh-aot (approved):

abschwenker (approved):

sources/packages/forms/src/form-definitions/exportfinancialreports.json
dheepak-aot: Yes it is @andrewsignori-aot . The file name is derived from this value dynamically.

dheepak-aot (approved):

Issue #1064, opened 2022-03-03
Process ESDC SIN Validation File


TeresaBia:
**Additional Context**
- SIN must be validated against ESDC
- additional documentation including draft UI under devs/esdc documentation/sin validation
- Ministry need to be able to
- Change a SIN
- Revalidate a SIN with ESDC
- Manually change and validate a SIN
- Add an expiry date for SINS that start with a 9
- Validate that the study period end date is within the SIN expiry date
- File shuffle job will be required to be changed to accommodate file name (D,T,P naming format)

**Acceptance Criteria**

**- File Processing**
- [x] Mark CRA SIN Validation processes as obsolete so they cannot be called but keep the code
- [x] Create send file for ESDC SIN Validation
- REFERENCE_IDX = sims.sin_validation.id
- First Name = Given Names
- File Name has a generic sequence number
- [ ] Create scheduler so once a day we call this new endpoint to create the file
- [ ] Create receive file for ESDC SIN Validation
- [ ] Create scheduler so once a day we call this new endpoint to receive the file
- [x] Change DB to accommodate new file structure 
- Ensure columns are added to accommodate SIN expiry date, and audit columns

PR #1276, opened 2022-06-17
[api][SIMS #1064]Process ESDC SIN Validation File

andrewsignori-aot:
- DB Migrations to removed CRA columns and add ESDC columns on sims.sin_validations;
- Remove CRA SIN validation-related specific source code. CRA is still used to do the income verification that also contains a SIN verification, this code remains;
- Created a new controller for sin-validation under ESDC;
- SIN moved from sims.students to sims.sin_validaions (**this caused some small changes across many files**);
- Add a column sims.sin_validations.temporary_sin as a generated("calculated") column to determine if the SIN is temporary or not.
- Logged operations

image

image

- Schedulers changes to be added in upcoming PR.


sources/packages/api/src-sql/Student/Rollback-remove-sin-col.sql
```diff
@@ -0,0 +1,22 @@
+ALTER TABLE
+    sims.students
+ADD
+    COLUMN IF NOT EXISTS sin VARCHAR(10);
+
+-- Copy SIN from sin_validations to students.
+UPDATE
+    sims.students
+SET
+    sin = CAST(sims.sin_validations.sin AS VARCHAR(10))
```
ann-aot, line 10: if the student has multiple sin rows in the `sin_validations`, when which one it will take, just for my info

andrewsignori-aot: I was not so worried about picking a specific one. It was more about having a value to allow setting the `NOT NULL` constraint back. Maybe it would be good to add a where as below at least:
```sql
WHERE
    sims.sin_validations.user_id = students.user_id;
    AND sims.sin_validations.sin IS NOT NULL
```
Thoughts?

ann-aot: mm...ya.. need not worry about it. but just wanted to know, if it's the first one or the last one. I think the current one should be fine since it will only happen when migration is reverted and which is not a scenario in PROD.

sources/packages/api/src/esdc-integration/sin-validation/models/sin-validation-models.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Possible known types of the SIN check status.
+ */
+export enum SINCheckStatus {
+  /**
+   * Passed electronic or manual validation.
+   */
+  Passed = "1",
```
ann-aot, line 50: just curios, why the values are `string`

sources/packages/api/src/esdc-integration/sin-validation/sin-validation-files/sin-validation-file-header.ts
```diff
@@ -0,0 +1,59 @@
...
+   * Get the information as a fixed line format to be
+   * added to the file uploaded to the SFTP.
+   * @returns fixed line formatted.
+   */
+  getFixedFormat(): string {
+    const record = new StringBuilder();
+    record.append(this.recordTypeCode);
+    record.appendWithStartFiller(this.batchNumber, 8, NUMBER_FILLER);
+    record.appendDate(this.processDate, DATE_FORMAT);
+    record.append("BC");
```
ann-aot, line 39: now we have the Province code in our DB right? can we use it from there? instead of hard coding.

sources/packages/api/src/services/sin-validation/sin-validation.service.ts
```diff
@@ -1,80 +1,195 @@
...
+        record: updatedRecord,
+      };
+    }
+
+    if (sinStatusChanged) {
+      // The SIN validation is already present and the received status is different from
+      // the existing one. It can happen when the status received is 2 (under review) and later
+      // a new status will be received once the review is done, what means that one request can
+      // result in more than one response. In this case, if the received record is the most updated
+      // for the student, a "cloned" record will be inserted.
```
ann-aot, line 153: just wanted to know why to clone the record. why can't we update the same row status?May be we can have sync up for details

sources/packages/api/src/esdc-integration/sin-validation/models/sin-validation-models.ts
andrewsignori-aot: It is how it is defined in the ESDC file documentation. We are just reading and saving as it is.

sources/packages/api/src/esdc-integration/sin-validation/sin-validation-files/sin-validation-file-header.ts
andrewsignori-aot: These are fixed province codes. I will create a const only.

sources/packages/api/src/services/sin-validation/sin-validation.service.ts
andrewsignori-aot: For auditing history, when we receive a response for a record already updated, we create a new record (only if the new record is the most updated one for the student).

sources/packages/api/src/services/sin-validation/sin-validation.service.ts
```diff
@@ -1,80 +1,195 @@
...
+        .select("MAX(sinValidation.dateReceived)", "maxDate")
+        .innerJoin("sinValidation.user", "user")
+        .where("user.id = :userId", {
+          userId: existingValidation.user.id,
+        })
+        .getRawOne();
+      if (processDate > mostUpdatedRecordDate.maxDate) {
+        // This will force the record to be inserted with all the values selected
+        // during the initial SQL query.
+        delete existingValidation.id;
```
ann-aot, line 165: should we also need to delete the previous or old audit fields and add new audit fields for the new cloned row?

andrewsignori-aot: The only columns retrieved in the SQL select are the ones meant to be cloned. I added comments about it on the query and in line 163 also "to be inserted with all the values selected during the initial SQL query."

sources/packages/api/src-sql/Student/Rollback-remove-sin-col.sql
andrewsignori-aot: The `WHERE` was not accurate. I changed it to use the `sin_validation_id ` already present on  `students.sin_validation_id = sims.sin_validations.id;`.

sources/packages/api/src/esdc-integration/sin-validation/sin-validation-files/sin-validation-file-header.ts
andrewsignori-aot: Constant add for ESDC SIN validation scope only.

ann-aot (approved):
Good work wth ESDC SIN Validation @andrewsignori-aot  Thanks for doing the changes 👍

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
```diff
@@ -0,0 +1,128 @@
...
+  sims.sin_validations
+SET
+  sin = CAST(sims.students.sin AS VARCHAR(9))
+FROM
+  sims.students
+WHERE
+  students.sin_validation_id = sims.sin_validations.id;
+
+-- Remove sin_validations that were not update from students.
+DELETE FROM
```
guru-aot, line 84: This is done to clean existing records?

andrewsignori-aot: Yes, to avoid having records that are not related to any student. The previous update `Copy SIN from student to the new sin column on sims.sin_validations` will ensure that every sims.sin_validations record will have a SIN value. In case there is any garage it will be removed.

sources/packages/api/src/cra-integration/cra-personal-verification.service.ts
```diff
@@ -23,14 +22,12 @@ import {
...
 import * as path from "path";
-import { SINValidationService } from "../services/sin-validation/sin-validation.service";
 
-const SIN_VALIDATION_TAG = "SIN_VALIDATION";
 const INCOME_VERIFICATION_TAG = "VERIFICATION_ID";
 
 /**
  * Manages the retrieval of students or parents or another
- * individual that needs to have his data verified
+ * individual that needs to have their data verified
```
henryoforeh-aot, line 30: Please change to "individuals that need to have their data verified".

henryoforeh-aot (approved):

guru-aot (approved):
LGTM, Nice work @andrewsignori-aot on the SIN column movement.

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
```diff
@@ -0,0 +1,128 @@
...
+ADD
+  COLUMN IF NOT EXISTS valid_gender_check CHAR(1),
+ADD
+  COLUMN IF NOT EXISTS sin_expire_date DATE;
+
+-- New columns added to allow audit of manual SIN manipulation.
+ALTER TABLE
+  sims.sin_validations
+ADD
+  COLUMN IF NOT EXISTS sin_edited_by INT REFERENCES sims.users (id) ON DELETE
```
dheepak-aot, line 53: IMO we can use modified/updated instead of edited as edit is a client level action which will result in modification.

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
```diff
@@ -0,0 +1,128 @@
...
+  NULL,
+ADD
+  COLUMN IF NOT EXISTS expired_date_edited_date TIMESTAMP WITH TIME ZONE,
+ADD
+  COLUMN IF NOT EXISTS expired_date_edited_note_id INT REFERENCES sims.notes (id) ON DELETE
+SET
+  NULL;
+
+-- Copy SIN from student to the new sin column on sims.sin_validations.
+UPDATE
```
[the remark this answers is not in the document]
andrewsignori-aot: Unless there is a strong reason I would like to avoid adding more files to the migration. Also, the DDL and DML are happening in a specific order to make it work.

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
```diff
@@ -0,0 +1,128 @@
...
+  DROP COLUMN IF EXISTS given_name_match_status_code,
+  DROP COLUMN IF EXISTS dob_match_status_code;
+
+-- Adding SIN column, previously it was just on the student table and from now on
+-- it will be the only this the sims.sin_validations.
+ALTER TABLE
+  sims.sin_validations
+ADD
+  -- This will be changed to NOT NULL in the end of the script.
+  COLUMN IF NOT EXISTS sin VARCHAR(9),
```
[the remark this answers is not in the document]
andrewsignori-aot: Yes, true. I was trying to follow what we have and we do have different ways right now.
I will change it to CHAR(9) which makes more sense.

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
andrewsignori-aot: The same `modified/updated` will also be updated by the system during ESDC SIN validation response process. So I would not believe that we should use the same column for both update audits.

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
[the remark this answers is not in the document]
andrewsignori-aot: Changed!

dheepak-aot (approved):
Good work @andrewsignori-aot 👍

sources/packages/api/src-sql/SINValidations/Alter-columns-for-esdc-sin-validations.sql
```diff
@@ -0,0 +1,128 @@
...
+ADD
+  COLUMN IF NOT EXISTS temporary_sin BOOLEAN NOT NULL GENERATED ALWAYS AS (
```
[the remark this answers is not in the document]
andrewsignori-aot: As discussed lest avoid using the sufix.

Issue #1249, opened 2022-05-30
Process disbursement file


abschwenker:
**Additional Context**
- This ticket is to save the information to SIMSDB from the daily disbursement file, ie. read file and convert to DB structure
- Future tickets will be created to action that information (eg. #139)
- Archiving of original file is not necessary
- If error/bad record, keep reading file - cannot predict errors at this stage
- Error tracking is a separate issue
- Do not process document numbers that do not exist in our DB

**Acceptance Criteria**
- [x] Create scheduler to read and process the file daily
- [x] Read the file ("the receipt file" for disbursement) and save the disbursement information
- [x] If a document number/certificate number exists on the disbursement receipt table for the particular FE/BC record, just ignore it

**Technical details**

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/19e2cda4-b028-4cff-a722-be464f09fee7
)

1. For each disbursement number there will be 2 records, one federal (FE) and one provincial (BC) (item 1 on the image above);
2. The records will be identified on our DB mainly by the document number (item 2 on the image);
3. When the documentation referer to EI it means the Institution (item 3 on the image);
4. The grant/loan code will be saved as it is received in a free text column on DB (item 4 on the image). The blank ones Codes with 0's represent no grants (they 0 fill)
5. The money values should be read as 5 digits only (item 5 on the image);
6. The end of each record contains an array of code+values (as shown on the image below) that will be saved into a one-to-many table related to the federal/provincial master record (mentioned on item 1).

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/75466a2e-76c0-4db1-a3c0-dc21814985b7)

**Please see below a draft ER**
- Consider adding disbursement schedule ID and do not add the document number itself

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/1f8fa56e-49bc-4632-9936-0cf26a7444ac)

PR #1278, opened 2022-06-20
[API][Workflow][SIMS #1249] - Disbursement Receipt - Part 2

dheepak-aot:
Draft

### DIFF
--- a/sources/packages/api/src/database/entities/disbursement-receipt-values.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-receipt-values.model.ts
@@ -1,3 +1,4 @@
+import { IsNotEmpty } from "class-validator";
 import {
   Column,
   Entity,
@@ -34,6 +35,7 @@ export class DisbursementReceiptValue extends RecordDataModel {
     name: "grant_type",
     nullable: false,
   })
+  @IsNotEmpty()
   grantType: string;
   /**
    * Grant amount.
@@ -44,5 +46,6 @@ export class DisbursementReceiptValue extends RecordDataModel {
     type: "numeric",
     nullable: false,
   })
+  @IsNotEmpty()
   grantAmount: string;
 }
--- a/sources/packages/api/src/database/entities/disbursement-receipt-values.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-receipt-values.model.ts
@@ -1,0 +1,48 @@
+import {
+  Column,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+} from "typeorm";
+import { DisbursementReceipt } from ".";
+import { ColumnNames, TableNames } from "../constant";
+import { RecordDataModel } from "./record.model";
+
+/**
+ * Grants(Both federal and provincial) which belong to a disbursement receipt.
+ */
+@Entity({ name: TableNames.DisbursementReceiptValues })
+export class DisbursementReceiptValue extends RecordDataModel {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  /**
+   * Related disbursement receipt entry.
+   */
+  @ManyToOne(() => DisbursementReceipt, { eager: false, nullable: false })
+  @JoinColumn({
+    name: "disbursement_receipt_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  disbursementReceipt: DisbursementReceipt;
+
+  /**
+   * Federal or Provincial grant type.
+   */
+  @Column({
+    name: "grant_type",
+    nullable: false,
+  })
+  grantType: string;
+  /**
+   * Grant amount.
+   * !Decimal values are retrieved by Typeorm as string from Postgres.
+   */
+  @Column({
+    name: "grant_amount",
+    type: "numeric",
+    nullable: false,
+  })
+  grantAmount: string;
+}
--- a/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
@@ -1,4 +1,11 @@
-import { IsDate, Length } from "class-validator";
+import { Type } from "class-transformer";
+import {
+  ArrayMinSize,
+  IsDate,
+  IsNotEmpty,
+  Length,
+  ValidateNested,
+} from "class-validator";
 import {
   Column,
   Entity,
@@ -79,6 +86,7 @@ export class DisbursementReceipt extends RecordDataModel {
     type: "numeric",
     nullable: false,
   })
+  @IsNotEmpty()
   totalEntitledDisbursedAmount: string;
 
   /**
@@ -89,6 +97,7 @@ export class DisbursementReceipt extends RecordDataModel {
     type: "numeric",
     nullable: false,
   })
+  @IsNotEmpty()
   totalDisbursedAmount: string;
 
   /**
@@ -111,6 +120,7 @@ export class DisbursementReceipt extends RecordDataModel {
     type: "numeric",
     nullable: false,
   })
+  @IsNotEmpty()
   disburseAmountStudent: string;
 
   /**
@@ -121,6 +131,7 @@ export class DisbursementReceipt extends RecordDataModel {
     type: "numeric",
     nullable: false,
   })
+  @IsNotEmpty()
   disburseAmountInstitution: string;
 
   /**
@@ -142,6 +153,7 @@ export class DisbursementReceipt extends RecordDataModel {
     name: "institution_code",
     nullable: false,
   })
+  @Length(4)
   institutionCode: string;
 
   /**
@@ -151,6 +163,7 @@ export class DisbursementReceipt extends RecordDataModel {
     name: "disburse_method_student",
     nullable: false,
   })
+  @Length(1)
   disburseMethodStudent: string;
 
   /**
@@ -160,12 +173,15 @@ export class DisbursementReceipt extends RecordDataModel {
     () => DisbursementReceiptValue,
     (disbursementReceiptValue) => disbursementReceiptValue.disbursementReceipt,
     {
-      eager: false,
+      eager: true,
       cascade: true,
-      nullable: false,
+      nullable: true,
     },
   )
-  disbursementReceiptValues: DisbursementReceiptValue[];
+  @ArrayMinSize(0)
+  @ValidateNested({ each: true })
+  @Type(() => DisbursementReceiptValue)
+  disbursementReceiptValues?: DisbursementReceiptValue[];
 
   /**
    * Study period end date.
--- a/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
@@ -167,6 +167,50 @@ export class DisbursementReceipt extends RecordDataModel {
   disburseMethodStudent: string;
 
   /**
+   * Total Federal or BC grant entitled amount.
+   */
+  @Column({
+    name: "total_entitled_grant_amount",
+    type: "numeric",
+    nullable: false,
+  })
+  @IsNotEmpty()
+  totalEntitledGrantAmount: string;
+
+  /**
+   * Total Federal or BC grant disbursed amount.
+   */
+  @Column({
+    name: "total_disbursed_grant_amount",
+    type: "numeric",
+    nullable: false,
+  })
+  @IsNotEmpty()
+  totalDisbursedGrantAmount: string;
+
+  /**
+   * Total Federal or BC grant disbursed amount to student.
+   */
+  @Column({
+    name: "total_disbursed_grant_amount_student",
+    type: "numeric",
+    nullable: false,
+  })
+  @IsNotEmpty()
+  totalDisbursedGrantAmountStudent: string;
+
+  /**
+   * Total Federal or BC grant disbursed amount to institution.
+   */
+  @Column({
+    name: "total_disbursed_grant_amount_institution",
+    type: "numeric",
+    nullable: false,
+  })
+  @IsNotEmpty()
+  totalDisbursedGrantAmountInstitution: string;
+
+  /**
    * Values for this disbursement.
    */
   @OneToMany(
--- a/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
@@ -1,3 +1,4 @@
+import { IsDate, Length } from "class-validator";
 import {
   Column,
   Entity,
@@ -28,6 +29,7 @@ export class DisbursementReceipt extends RecordDataModel {
     transformer: dateOnlyTransformer,
     nullable: false,
   })
+  @IsDate()
   batchRunDate: Date;
 
   /**
@@ -37,6 +39,7 @@ export class DisbursementReceipt extends RecordDataModel {
     name: "student_sin",
     nullable: false,
   })
+  @Length(9)
   studentSIN: string;
 
   /**
@@ -65,6 +68,7 @@ export class DisbursementReceipt extends RecordDataModel {
     name: "funding_type",
     nullable: false,
   })
+  @Length(2)
   fundingType: string;
 
   /**
@@ -96,6 +100,7 @@ export class DisbursementReceipt extends RecordDataModel {
     transformer: dateOnlyTransformer,
     nullable: false,
   })
+  @IsDate()
   disburseDate: Date;
 
   /**
@@ -127,6 +132,7 @@ export class DisbursementReceipt extends RecordDataModel {
     transformer: dateOnlyTransformer,
     nullable: false,
   })
+  @IsDate()
   dateSignedInstitution: Date;
 
   /**
@@ -170,5 +176,6 @@ export class DisbursementReceipt extends RecordDataModel {
     transformer: dateOnlyTransformer,
     nullable: false,
   })
+  @IsDate()
   studyPeriodEndDate: Date;
 }
--- a/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
+++ b/sources/packages/api/src/database/entities/disbursement-receipts.model.ts
@@ -1,0 +1,174 @@
+import {
+  Column,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  OneToMany,
+  PrimaryGeneratedColumn,
+  RelationId,
+} from "typeorm";
+import { DisbursementSchedule, DisbursementReceiptValue } from ".";
+import { ColumnNames, TableNames } from "../constant";
+import { dateOnlyTransformer } from "../transformers/date-only.transformer";
+import { RecordDataModel } from "./record.model";
+
+/**
+ * Disbursement receipts received from ESDC for the disbursements sent.
+ */
+@Entity({ name: TableNames.DisbursementReceipts })
+export class DisbursementReceipt extends RecordDataModel {
+  @PrimaryGeneratedColumn()
+  id: number;
+  /**
+   * Effective date of data on the disbursement receipt file.
+   */
+  @Column({
+    name: "batch_run_date",
+    type: "date",
+    transformer: dateOnlyTransformer,
+    nullable: false,
+  })
+  batchRunDate: Date;
+
+  /**
+   * SIN number of the student who received the disbursement.
+   */
+  @Column({
+    name: "student_sin",
+    nullable: false,
+  })
+  studentSIN: string;
+
+  /**
+   * Disbursement id to which the document number from the file belongs to..
+   */
+  @ManyToOne(() => DisbursementSchedule, { eager: false, nullable: false })
+  @JoinColumn({
+    name: "disbursement_schedule_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  disbursementSchedule: DisbursementSchedule;
+
+  /**
+   * disbursement schedule id.
+   */
+  @RelationId(
+    (disbursementReceipt: DisbursementReceipt) =>
+      disbursementReceipt.disbursementSchedule,
+  )
+  disbursementScheduleId: number;
+
+  /**
+   * Indicates the funding type of the disbursement receipt item. Values can have FE(Federal) or BC(Provincial)..
+   */
+  @Column({
+    name: "funding_type",
+    nullable: false,
+  })
+  fundingType: string;
+
+  /**
+   * Total entitled disbursed amount for either FE(Federal) or BC(Provincial)..
+   */
+  @Column({
+    name: "total_entitled_disbursed_amount",
+    type: "numeric",
+    nullable: false,
+  })
+  totalEntitledDisbursedAmount: string;
+
+  /**
+   * Total disbursed amount for either FE(Federal) or BC(Provincial)..
+   */
+  @Column({
+    name: "total_disbursed_amount",
+    type: "numeric",
+    nullable: false,
+  })
+  totalDisbursedAmount: string;
+
+  /**
+   * Financial posting date of NSLSC.
+   */
+  @Column({
+    name: "disburse_date",
+    type: "date",
+    transformer: dateOnlyTransformer,
+    nullable: false,
+  })
+  disburseDate: Date;
+
+  /**
+   * Amount disbursed to the student.
+   */
+  @Column({
+    name: "disburse_amount_student",
+    type: "numeric",
+    nullable: false,
+  })
+  disburseAmountStudent: string;
+
+  /**
+   * Amount disbursed to the institution.
+   */
+  @Column({
+    name: "disburse_amount_institution",
+    type: "numeric",
+    nullable: false,
+  })
+  disburseAmountInstitution: string;
+
+  /**
+   * The date signed by the institution of the student at the time they receive the loan certificate.
+   */
+  @Column({
+    name: "date_signed_institution",
+    type: "date",
+    transformer: dateOnlyTransformer,
+    nullable: false,
+  })
+  dateSignedInstitution: Date;
+
+  /**
+   * Institution code.
+   */
+  @Column({
+    name: "institution_code",
+    nullable: false,
+  })
+  institutionCode: string;
+
+  /**
+   * Method of disbursement to student. C(By cheque),E(By EFT), <space> disbursement only to institution.
+   */
+  @Column({
+    name: "disburse_method_student",
+    nullable: false,
+  })
+  disburseMethodStudent: string;
+
+  /**
+   * Values for this disbursement.
+   */
+  @OneToMany(
+    () => DisbursementReceiptValue,
+    (disbursementReceiptValue) => disbursementReceiptValue.disbursementReceipt,
+    {
+      eager: false,
+      cascade: true,
+      nullable: false,
+    },
+  )
+  disbursementReceiptValues: DisbursementReceiptValue[];
+
+  /**
+   * Study period end date.
+   */
+  @Column({
+    name: "study_period_end_date",
+    type: "date",
+    transformer: dateOnlyTransformer,
+    nullable: false,
+  })
+  studyPeriodEndDate: Date;
+}
--- a/sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
+++ b/sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
@@ -56,6 +56,22 @@ export class DisbursementReceiptDetail extends DisbursementReceiptRecord {
     return this.convertToDateRecord(this.line.substring(86, 94));
   }
 
+  public get totalEntitledGrantAmount() {
+    return this.convertToAmountString(this.line.substring(94, 101));
+  }
+
+  public get totalDisbursedGrantAmount() {
+    return this.convertToAmountString(this.line.substring(101, 108));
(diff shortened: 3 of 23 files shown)


sources/packages/api/src/route-controllers/models/esdc.shared.dto.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Shared DTO for EDSC file processing API response.
+ * TODO: Other esdc file processing services must eventually use this DTO.
+ */
+export class ESDCFileResponseAPIOutDTO {
```
andrewsignori-aot, line 5: There is already a shared DTO for this one `ESDCFileResponseDTO`.
The name is not following the conventions because it is old, but it is a shared one. Can we use it?

sources/packages/api/src/database/entities/disbursement-receipts.model.ts
```diff
@@ -0,0 +1,214 @@
...
+   * SIN number of the student who received the disbursement.
+   */
+  @Column({
+    name: "student_sin",
+    nullable: false,
+  })
+  studentSIN: string;
+
+  /**
+   * Disbursement id to which the document number from the file belongs to..
```
andrewsignori-aot, line 43: There is a double period in the end.

sources/packages/api/src/database/entities/disbursement-receipts.model.ts
```diff
@@ -0,0 +1,214 @@
...
+   */
+  @ManyToOne(() => DisbursementSchedule, { eager: false, nullable: false })
+  @JoinColumn({
+    name: "disbursement_schedule_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  disbursementSchedule: DisbursementSchedule;
+
+  /**
+   * disbursement schedule id.
```
andrewsignori-aot, line 53: Please start with an upper letter.

sources/packages/api/src/database/entities/disbursement-receipts.model.ts
```diff
@@ -0,0 +1,214 @@
...
+  totalDisbursedGrantAmountInstitution: string;
+
+  /**
+   * Values for this disbursement.
+   */
+  @OneToMany(
+    () => DisbursementReceiptValue,
+    (disbursementReceiptValue) => disbursementReceiptValue.disbursementReceipt,
+    {
+      eager: true,
```
andrewsignori-aot, line 197: I would not recommend leaving this as eager `true`. Is there any strong reason to have it as `true`?

sources/packages/api/src/database/entities/disbursement-receipts.model.ts
```diff
@@ -0,0 +1,214 @@
...
+      cascade: true,
```
andrewsignori-aot, line 198: I would recommend allowing the cascade only for `insert` here. The records are never going to be updated or removed as far as we know so far, right?

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Disbursement receipt detail record which has the receipt details for disbursements sent.
+ * The document number which is present in each of the receipt detail record connects the
+ * disbursement sent with the receipt received.
+ */
+export class DisbursementReceiptDetail extends DisbursementReceiptRecord {
+  constructor(line: string, lineNumber: number) {
+    super(line, lineNumber);
+  }
+
+  //Number of occurrences of the grant field.
```
andrewsignori-aot, line 23: The comment is good but I would recommend adding it in a way that VS Code will be able to recognize it.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
```diff
@@ -0,0 +1,123 @@
...
+  private GRANT_OCCURRENCE = 10;
+
+  public get studentSIN() {
```
andrewsignori-aot, line 26: We no longer are using the public explicit as per the wiki.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  public get totalDisbursedGrantAmountInstitution() {
+    return this.convertToAmountString(this.line.substring(115, 122));
+  }
+
+  public get grants() {
+    const grants: DisbursementReceiptGrant[] = [];
+    let grantIndex = 122;
+    for (let i = 1; i <= this.GRANT_OCCURRENCE; i++) {
+      const grantType = this.line.substring(grantIndex, grantIndex + 6).trim();
```
andrewsignori-aot, line 94: The number 6 here can be added as a const like `GRANT_TYPE_LENGTH`, it would increase the core readability.
Same for the amount.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+    return grants;
+  }
+
+  /**
+   * Validate the record detail data.
+   * @returns validation error message if validation fails.
+   */
+  public getInvalidDataMessage(): string | null {
```
andrewsignori-aot, line 113: Nice way of checking for possible bad data in the file 👍

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  public get grants() {
+    const grants: DisbursementReceiptGrant[] = [];
+    let grantIndex = 122;
+    for (let i = 1; i <= this.GRANT_OCCURRENCE; i++) {
+      const grantType = this.line.substring(grantIndex, grantIndex + 6).trim();
+      grantIndex = grantIndex + 6;
+
+      const grantAmountText = this.line.substring(grantIndex, grantIndex + 7);
+      grantIndex = grantIndex + 7;
```
andrewsignori-aot, line 98: It could be simplified to `grantIndex += 7;`

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
```diff
@@ -0,0 +1,47 @@
+import { getDateOnlyFromFormat } from "../../../utilities";
+import { DATE_FORMAT } from "../../models/esdc-integration.model";
+import { DisbursementReceiptRecordType } from "../models/disbursement-receipt-integration.model";
+
+/**
+ * Base class for disbursement receipt record.
+ * All the records of disbursement receipt must be a sub class of this.
+ */
+export class DisbursementReceiptRecord {
```
andrewsignori-aot, line 9: Can this class be marked as abstracted?

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
```diff
@@ -0,0 +1,47 @@
...
+    return getDateOnlyFromFormat(dateText, DATE_FORMAT);
+  }
+
+  /**
+   * Parses any amount field to numeric string by ignoring any junk character is present in data.
+   ** parseFloat is used to convert string like 00520.00 to 520.00.
+   * @param amountText
+   * @returns amount value as numeric string.
+   */
+  protected convertToAmountString(amountText: string) {
```
andrewsignori-aot, line 38: I can understand the point about removing the possible garbage from the decimals (already a known issue) but I would not recommend bypassing a possible NaN in the main part (0, 5). If that value comes to a NaN it should be considered an invalid record and it should be part of the log for the investigation.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Any date record as string is parsed to date object.
+   * @param dateText
+   * @returns Date object.
+   */
+  protected convertToDateRecord(dateText: string) {
+    return getDateOnlyFromFormat(dateText, DATE_FORMAT);
+  }
+
+  /**
+   * Parses any amount field to numeric string by ignoring any junk character is present in data.
```
andrewsignori-aot, line 33: I got the idea of a workaround for the current garbage characters received in the file. Just keep in mind that we follow the ticket orientation and read only the main part (0, 5) this method will be not necessary at all.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
```diff
@@ -0,0 +1,47 @@
...
+import { DisbursementReceiptRecordType } from "../models/disbursement-receipt-integration.model";
+
+/**
+ * Base class for disbursement receipt record.
+ * All the records of disbursement receipt must be a sub class of this.
+ */
+export class DisbursementReceiptRecord {
+  constructor(
+    protected readonly line: string,
+    protected readonly _lineNumber = 0,
```
andrewsignori-aot, line 12: This can be private if there is an associated public property `get lineNumber()`.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-integration.service.ts
```diff
@@ -0,0 +1,62 @@
...
+        `The Disbursement receipt file ${remoteFilePath} has an invalid transaction code on trailer: ${trailer.recordType}`,
+      );
+      // If the trailer is not the expected one.
+      throw new Error("Invalid file trailer.");
+    }
+    let totalSINHashCalculated = 0;
+    const records: DisbursementReceiptDetail[] = [];
+    fileLines.forEach((line, index) => {
+      const record = new DisbursementReceiptDetail(line, index + 1);
+      totalSINHashCalculated += parseInt(record.studentSIN);
```
andrewsignori-aot, line 51: `parseInt` can be changed by the shortcut `+record.studentSIN`, just a suggestion.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    private readonly disbursementScheduleService: DisbursementScheduleService,
+    private readonly disbursementReceiptService: DisbursementReceiptService,
+  ) {
+    this.esdcConfig = config.getConfig().ESDCIntegration;
+  }
+
+  /**
+   * Process all the available disbursement receipt files in SFTP location.
+   * Once the file is processed, it gets deleted.
+   * @param auditUserId
```
andrewsignori-aot, line 39: Please add a comment. There are many already with a common one, like:
```
user that should be considered the one that is causing the changes.
```

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+   * @returns Summary details of the processing.
+   */
+  async process(auditUserId: number): Promise<ProcessSFTPResponseResult> {
```
andrewsignori-aot, line 42: I would say that usually, this method would return one `ProcessSFTPResponseResult` for each processed file. Does keeping all the logs about all the files in one single log makes it cleaner?

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+        result.errorsSummary.push(
+          `Error while deleting disbursement receipt file: ${filePath}`,
+        );
+        result.errorsSummary.push(error);
+      }
+    }
+    return result;
+  }
+  /**
+   *
```
andrewsignori-aot, line 76: Please add a method comment.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      );
+    } catch (error: unknown) {
+      this.logger.error(error);
+      result.errorsSummary.push(
+        `Error downloading file ${remoteFilePath}. Error: ${error}`,
+      );
+      return;
+    }
+
+    const documentNumbers = responseData.records.map(
```
andrewsignori-aot, line 101: Can we add a type here like below?
```typescript
interface Foo {
   [key: string]: number;
}
```

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      (record) => record.documentNumber,
+    );
+    const disbursementSchedules =
+      await this.disbursementScheduleService.getDisbursementsByDocumentNumbers(
+        documentNumbers,
+      );
+    const disbursementScheduleMap = {};
```
andrewsignori-aot, line 108: Can we add a type here like below?
```typescript
interface Foo {
   [key: string]: number;
}
```

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const createdAt = new Date();
+    for (const response of responseData.records) {
+      const disbursementScheduleId =
+        disbursementScheduleMap[response.documentNumber];
+      if (disbursementScheduleId) {
+        const disbursementReceipt = new DisbursementReceipt();
+        disbursementReceipt.batchRunDate = responseData.header.batchRunDate;
+        disbursementReceipt.studentSIN = response.studentSIN;
+        disbursementReceipt.disbursementSchedule = {
```
andrewsignori-aot, line 127: This object can be created while processing the `disbursementScheduleMap`.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+        disbursementReceipt.studyPeriodEndDate = response.studyPeriodEndDate;
+        disbursementReceipt.totalEntitledGrantAmount =
+          response.totalEntitledGrantAmount;
+        disbursementReceipt.totalDisbursedGrantAmount =
+          response.totalDisbursedGrantAmount;
+        disbursementReceipt.totalDisbursedGrantAmountStudent =
+          response.totalDisbursedGrantAmountStudent;
+        disbursementReceipt.totalDisbursedGrantAmountInstitution =
+          response.totalDisbursedGrantAmountInstitution;
+        disbursementReceipt.creator = { id: auditUserId } as User;
```
andrewsignori-aot, line 154: The object can be created once and used multiple times.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+            const disbursementReceiptValue = new DisbursementReceiptValue();
+            disbursementReceiptValue.grantType = grant.grantType;
+            disbursementReceiptValue.grantAmount = grant.grantAmount;
+            disbursementReceiptValue.creator = { id: auditUserId } as User;
+            disbursementReceiptValue.createdAt = createdAt;
+            return disbursementReceiptValue;
+          },
+        );
+        try {
+          const invalidDataMessage = response.getInvalidDataMessage();
```
andrewsignori-aot, line 167: Can this be done before creating the object to be saved to DB?

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    this.logger.log(
+      "Processing all the records in file with valid data and document number that belongs to SIMS.",
+    );
+
+    const createdAt = new Date();
+    for (const response of responseData.records) {
+      const disbursementScheduleId =
+        disbursementScheduleMap[response.documentNumber];
+      if (disbursementScheduleId) {
+        const disbursementReceipt = new DisbursementReceipt();
```
andrewsignori-aot, line 124: I would prefer to have this logic to create the entity model and save it moved to inside the entity model service itself.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+          (grant) => {
+            const disbursementReceiptValue = new DisbursementReceiptValue();
+            disbursementReceiptValue.grantType = grant.grantType;
+            disbursementReceiptValue.grantAmount = grant.grantAmount;
+            disbursementReceiptValue.creator = { id: auditUserId } as User;
+            disbursementReceiptValue.createdAt = createdAt;
+            return disbursementReceiptValue;
+          },
+        );
+        try {
```
andrewsignori-aot, line 166: IMO, the `try` should involve more code to ensure that even small issues will not stop the processing.

sources/packages/api/src/services/disbursement-receipt/disbursement-receipt.service.ts
```diff
@@ -0,0 +1,51 @@
...
+      const result = await transactionalEntityManager
+        .createQueryBuilder()
+        .insert()
+        .into(DisbursementReceipt)
+        .values(disbursementReceipt)
+        .orIgnore(
+          "ON CONSTRAINT disbursement_schedule_id_funding_type_unique DO NOTHING",
+        )
+        .execute();
+      if (
```
andrewsignori-aot, line 37: Can we refactor the below code to remove the `[0]`? Something as below:
image

sources/packages/api/src/route-controllers/disbursement-receipt/disbursement-receipt.system-access.controller.ts
```diff
@@ -0,0 +1,38 @@
...
+import { AuthorizedParties } from "../../auth/authorized-parties.enum";
+import { AllowAuthorizedParty, UserToken } from "../../auth/decorators";
+import { IUserToken } from "../../auth/userToken.interface";
+import { DisbursementReceiptProcessingService } from "../../esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service";
+import { ClientTypeBaseRoute } from "../../types";
+import { ESDCFileResponseAPIOutDTO } from "../models/esdc.shared.dto";
+import BaseController from "../BaseController";
+
+@AllowAuthorizedParty(AuthorizedParties.formsFlowBPM)
+@Controller("disbursement-receipt")
```
andrewsignori-aot, line 12: This is an eCert-related operation, do we need a new controller for this one?

sources/packages/api/src/database/entities/disbursement-receipt-values.model.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Federal or Provincial grant type.
+   */
+  @Column({
+    name: "grant_type",
+    nullable: false,
+  })
+  grantType: string;
+  /**
+   * Grant amount.
+   * !Decimal values are retrieved by Typeorm as string from Postgres.
```
guru-aot, line 40: 👍

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    this.logger.log(
+      "Processing all the records in file with valid data and document number that belongs to SIMS.",
+    );
+
+    const createdAt = new Date();
+    for (const response of responseData.records) {
+      const disbursementScheduleId =
+        disbursementScheduleMap[response.documentNumber];
+      if (disbursementScheduleId) {
```
andrewsignori-aot, line 123: When the record is skipped because the document number is not present in our DB maybe it would be worth registering a log about it.

sources/packages/api/src/route-controllers/models/esdc.shared.dto.ts
dheepak-aot: Ah, I thought it was not shared. I believe I need to move the controller class as well inside esdc-integration folder.

sources/packages/api/src/route-controllers/disbursement-receipt/disbursement-receipt.system-access.controller.ts
andrewsignori-aot, line 12: There is a controllers folder specific for `esdc-integration`.

sources/packages/api/src/database/entities/disbursement-receipts.model.ts
dheepak-aot: I was experimenting insert() if there is a way to casade insert relationships. It is a left-over of the experiment

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/models/disbursement-receipt-integration.model.ts
```diff
@@ -0,0 +1,19 @@
+import { DisbursementReceiptDetail } from "../disbursement-receipt-files/disbursement-receipt-file-detail";
+import { DisbursementReceiptHeader } from "../disbursement-receipt-files/disbursement-receipt-file-header";
+
+/**
+ * Record type of disbursement receipt records.
+ */
+export enum DisbursementReceiptRecordType {
+  Header = "F",
+  Detail = "D",
+  Trailer = "T",
```
andrewsignori-aot, line 10: The file that handles this record is called `DisbursementReceiptFooter`. Can we call both footers or both trailers? I personally would prefer the footer because I believe that is what we used mostly but both are accurate.

andrewsignori-aot (commented):
Great and extensive work. Please take a look at the comments.

sources/packages/api/src/route-controllers/models/esdc.shared.dto.ts
andrewsignori-aot: I would suggest not creating a new controller. IMO the operation is related to the e-Cert files and it can stay in the existing e-Cert controller.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
```diff
@@ -0,0 +1,47 @@
...
+   * @returns amount value as numeric string.
+   */
+  protected convertToAmountString(amountText: string) {
+    const amountWholeValue = amountText.substring(0, 5);
+    const amountDecimalValue = amountText.substring(5, 7);
+    const wholeValue = isNaN(+amountWholeValue) ? "0" : amountWholeValue;
+    const amount = isNaN(+amountDecimalValue)
+      ? `${wholeValue}.00`
+      : `${wholeValue}.${amountDecimalValue}`;
+    return parseFloat(amount).toFixed(2);
```
ann-aot, line 45: there was some issue with the `.toFixed` rounding. that's the reason we moved all `.toFixed` used in camunda to `Math.round`

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
```diff
@@ -0,0 +1,188 @@
...
+   * Once the file is processed, it gets deleted.
+   * @param auditUserId
+   * @returns Summary details of the processing.
+   */
+  async process(auditUserId: number): Promise<ProcessSFTPResponseResult> {
+    const result = new ProcessSFTPResponseResult();
+
+    // Get the list of all files from SFTP ordered by file name.
+    const fileSearch = new RegExp(
+      `^${this.esdcConfig.environmentCode}EDU.PBC.DIS.D[\w]*\.[0-9]*`,
```
guru-aot, line 47: can you please confirm whether the filename that we read from the server has a file extension line .001 or 2 or .300 etc?. If there is a specific type of filename like .txt or .dat, please append them at the end.

ann-aot (commented):
Good Work @dheepak-aot . most of the comments are captured by @andrewsignori-aot

guru-aot (commented):
Nice work @dheepak-aot

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
dheepak-aot: I get it. 👍

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
dheepak-aot: The constructor is meant to be used by the sub classes. Hence if it is private, subclass will not be able to use super() and pass the value.

sources/packages/api/src/route-controllers/disbursement-receipt/disbursement-receipt.system-access.controller.ts
dheepak-aot: I can move this to esdc-controller.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-record.ts
andrewsignori-aot: The method as it is right now can be just as below, what usually we have been doing in the method itself.
```typescript
+amountText.substring(0, 5)
```
I believe that typeorm will not need the `.toString()` also.
IMO the entire method can be removed.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
andrewsignori-aot: So we have a comment that is used in many other places as mentioned in the comment. It also seems to be more descriptive than `user id of API user.`. That said, is there any reason to prefer `user id of API user.` instead of `user that should be considered the one that is causing the changes.`?

sources/packages/api/src/services/disbursement-receipt/disbursement-receipt.service.ts
andrewsignori-aot: The second was not changed, can it be changed, please?
image

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
dheepak-aot: No specific reason. There was one other method using the API comment. I can change

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-files/disbursement-receipt-file-detail.ts
```diff
@@ -0,0 +1,140 @@
...
+ */
+export class DisbursementReceiptDetail extends DisbursementReceiptRecord {
+  constructor(line: string, lineNumber: number) {
+    super(line, lineNumber);
+  }
+
+  /**
+   * Number of occurrences of the grant field.
+   */
+  private GRANT_OCCURRENCE = 10;
```
andrewsignori-aot, line 26: These are supposed to be declared as a `const`, right?

andrewsignori-aot (commented):
Thanks for doing the changes, minor comments only, please take a look.

sources/packages/api/src/esdc-integration/disbursement-receipt-integration/disbursement-receipt-processing.service.ts
andrewsignori-aot: Not a big deal or blocker, just curious. I will resolve this comment wither way 😉

henryoforeh-aot (approved):

andrewsignori-aot (approved):

guru-aot (approved):

Issue #1063, opened 2022-03-03
SIN validation management


TeresaBia:
### Create Ministry View SIN UI
- [x]  Add new tab for SIN under student profile
- [x] On click of SIN tab, display a table with the following columns and the following buttons
- [x] Columns
- SIN 
- SIN Validated
- Response code
- SIN accepted
- First name
- Last name
- Date of birth
- Gender
- Expiry Date
- Default to most recent on the top, the 1 on the top will ALWAYS be the active SIN validation record for the student
- Note: Ministry user cares about SIN validated column - other column values are coming from ESDC
- [x] Button "Add new sin" (opens modal)
-  On click of button "Add new sin" -- display a modal that:
- Add new SIN
- Check box for "Skip the validations"
- When user clicks on check box, red danger banner
- Note box - content will be displayed under "general" in Notes tab
- This will be a blank record that will get validated with ESDC (when skip validation is not checked)
- This will be a complete record with the user that created it as the validating_user and validated_date (when skip validation is checked)
- [x] For temporary SIN numbers, on click of "Add expiry date" allow the user to enter a date with a note. Note will be displayed under "general" in Notes tab
- [x] When adding new SIN, update the student SIN relation

### Student Create Account UI changes
- [x] When a student enters a SIN number that starts with a 9, display the content "You have a temporary Social Insurance Number. You may still create your account but your applications will not be processed until we verify your SIN information. Please refer here (https://studentaidbc.ca/apply/eligibility#basic-eligibility) for the documentation required to prepare for the documentation required to submit and support your citizenship status’."

### System Add Restriction
- [x] Create restriction SINF. With the following
- restriction type = provincial
- restriction category = verification
- restriction code = SINF
- ministry description = Student submitted application with incorrect SIN expiry date
- notification_type = error
- action_type = stop full time disbursement, stop part time disbursement,
- [x] When a student whos SIN starts with temporary_sin submits an application. If the expiry date of their SIN Expiry is NULL or BEFORE the study end date, add restriction SINF
- [x] When an institution submits a PIR for a student whos SIN is temporary_sin. If the expiry date of their SIN Expiry is NULL or BEFORE the study end date, add restriction SINF
- [x] If a student submits an applicant with this restriction, do not add another
- [x] If an institution submits a PIR with this restriction, do not add another

### Add student account activity page content
- [x] Add panel for Code SINF
- [x] Content:
[...]

PR #1287, opened 2022-06-24
[web][api][SIMS #1063]AEST SIN Management UI - Part 2

andrewsignori-aot:
- Allow the Ministry to create the new SIN.
- Allow the Ministry to add an expiry date to a temporary SIN.
- Directly edited more DB columns that were wrongly named as expire instead of expiry.


sources/packages/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -181,3 +189,31 @@ export class SINValidationsAPIOutDTO {
   temporarySIN: boolean;
   sinExpiryDate?: string;
 }
+
+/**
+ * DTO to allow manually creation of SIN validations.
+ */
+export class CreateSINValidationAPIInDTO {
+  @IsValidSIN()
```
guru-aot, line 197: Nice validator

andrewsignori-aot: @dheepak-aot @henryoforeh-aot just to share a conversation that I had with @guru-aot, this code is also used directly on form.io definitions that contain a SIN to be validated. It was kept simple and mostly the same as the one used in form.io. Please keep in mind also that the code does not use any of the newest javascript features to keep it as compatible as possible.

sources/packages/api/src/route-controllers/utils/custom-validators/sin-validator.ts
```diff
@@ -0,0 +1,58 @@
...
+  ValidationArguments,
+} from "class-validator";
+import { SIN_MAX_LENGTH } from "../../../database/entities";
+
+/**
+ * SIN validation algorithm (Luhn algorithm).
+ * ! This logic is also present on form.io client side validations.
+ */
+@ValidatorConstraint()
+export class IsValidSINConstraint implements ValidatorConstraintInterface {
```
henryoforeh-aot, line 15: Just wondering if this class needs to be exported since it is only used during registration of the decorator which is in the same file.

andrewsignori-aot: Good catch, no it is not needed. I just removed it.

henryoforeh-aot (approved):

guru-aot (approved):

dheepak-aot (approved):

PR #1288, opened 2022-06-24
[api][SIMS #1063]AEST SIN Management - Restrictions - Part 3

andrewsignori-aot:
- Created the new SIN restriction code SINF.
- Verify if the SIN restriction should be created when an application is submitted. Added the check when the workflow is setting the offering on the sims.student_assessments table.
- Verify during the PIR complete process if the SIN restriction must be created.
- New message was added to the student profile during the creation process to indicate that the SIN is temporary.

image

- New content was added to the student activity page.

image


sources/packages/api/src/services/application/application.service.ts
```diff
@@ -653,36 +655,49 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+          "Not able to find an application that requires a PIR to be completed.",
+          PIR_REQUEST_NOT_FOUND_ERROR,
+        );
+      }
+      const auditUser = { id: auditUserId } as User;
+      application.currentAssessment.offering = offering;
+      application.currentAssessment.modifier = auditUser;
+      application.pirStatus = ProgramInfoStatus.completed;
+      application.modifier = auditUser;
+      await this.studentRestrictionService.assessSINRestrictionForOfferingId(
```
andrewsignori-aot, line 692: This is the new method, the others changes are due to the addition of the transaction only.

sources/packages/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -193,33 +198,54 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+      assessment.application.modifier = auditUser;
+      assessment.modifier = auditUser;
+      assessment.offering = {
+        id: offeringId,
+      } as EducationProgramOffering;
 
-    return this.repo.save(assessment);
+      // If the offering will be set in the assessment check for possible SIN restrictions.
+      if (offeringId) {
+        await this.studentRestrictionService.assessSINRestrictionForOfferingId(
```
andrewsignori-aot, line 239: This is the new method, the others changes are due to the addition of the transaction only.

sources/packages/forms/src/form-definitions/studentaccountactivity.json
```diff
@@ -4,7 +4,9 @@
   "type": "form",
   "name": "studentAccountActivity",
   "path": "studentaccountactivity",
-  "tags": ["common"],
+  "tags": [
+    "common"
```
ann-aot, line 8: is this format correct @andrewsignori-aot ?

sources/packages/forms/src/form-definitions/studentinformation.json
```diff
@@ -1097,6 +1011,333 @@
...
+              "validateOn": "change",
+              "validate": {
+                "required": true,
+                "minLength": 9,
+                "maxLength": 11,
+                "minWords": "",
+                "maxWords": "",
+                "pattern": "",
+                "customMessage": "Invalid Social Insurance Number",
+                "custom": "// Common SIN validation algorithm (Luhn algorithm).\r\nvalid = false;\r\nif (input) {\r\n  input = input.replace(/\\s/g, \"\");\r\n  if (input.length === 9) {\r\n    var checksum = 0;\r\n    for (var i = 0; i < input.length; i++) {\r\n      const currentDigit = parseInt(input.charAt(i));\r\n      if ((i + 1) % 2 === 0) {\r\n        const digitTimes2 = currentDigit * 2;\r\n        checksum += digitTimes2 < 10 ? digitTimes2 : digitTimes2 - 9;\r\n      } else {\r\n        checksum += parseInt(input.charAt(i));\r\n      }\r\n    } \r\n    if (checksum % 10 === 0) {\r\n      valid = true;\r\n    }\r\n  }\r\n}",
```
ann-aot, line 1180: 👍

sources/packages/forms/src/form-definitions/studentinformation.json
```diff
@@ -1295,7 +1536,7 @@
           "case": "",
           "truncateMultipleSpaces": false,
           "encrypted": false,
-          "redrawOn": "",
+          "redrawOn": "data",
```
dheepak-aot, line 1539: is redraw on data going to watch every property of data?

sources/packages/api/src/services/restriction/student-restriction.service.ts
```diff
@@ -326,4 +327,76 @@ export class StudentRestrictionService extends RecordDataModelService<StudentRes
...
+    if (!student.sinValidation.temporarySIN) {
+      // SIN is not temporary so no restriction should be created.
+      return;
+    }
+    // By default assume that the temporary SIN does not have an expiry
+    // date defined and the restriction must be created.
+    let mustCreateSINException = true;
+    if (student.sinValidation.sinExpiryDate) {
+      // Temporary SIN has an expiry date and it must be validated with the end date of the offering.
+      const offering = await entityManager
```
dheepak-aot, line 384: We can pass the offering entity model to this method instead of offering id and avoid this additional query.

sources/packages/forms/src/form-definitions/studentinformation.json
andrewsignori-aot: Probably yes. This belongs to the address area that is not part of this PR. This type of `redrawOn` is also present across many other form.io definitions.
image

sources/packages/api/src/services/restriction/student-restriction.service.ts
andrewsignori-aot: It will force the consumers (workflow setting offering and PIR complete process) to know what properties of the offering we need for this specific part. This method is selecting exactly what is needed and I would prefer that the method's signature be specific about what is needed instead of receiving an object with a lot of properties that are not used at all.
For instance, while setting the offering id from the workflow process we just have the offering id so, so this selection will be needed always.
Selecting the offering end date is only needed when the SIN is temporary, which means that for all the cases where the SIN is not temporary this query will never be executed.

guru-aot (approved):

PR #1296, opened 2022-06-29
[api]Removed IsInstitutionAdmin decorator

andrewsignori-aot:
- The API endpoint to get the application details was restricted only for admin users but it is also needed during the login process for non-admin-users which leads to a wrong _Forbidden User_ error.

image


guru-aot (approved):
LGTM

TeresaBia:
@andrewsignori-aot Thanks!

henryoforeh-aot (approved):

ann-aot (approved):
👍

Issue #1300, opened 2022-07-05
Reorganize API controllers


TeresaBia:
Reorganize the below API controllers:
- [ ] During COE - Remove edit application (application rollback) for PIR Required application
- [ ] educationprogramoffering controller moved to new ticket #1338

![edited.jpg](https://images.zenhubusercontent.com/61c390f5d34132a9b12ca101/fe664562-1dab-4342-a91e-fd43dbaf5f3e)

PR #1323, opened 2022-07-21
[web][api][SIMS #1300]Institution COE controller reorganization

andrewsignori-aot:
- Moved confirmation of enrollment methods to a specific controller under `Institutions` API client.
- Removed extra services from `app.module` that were not used anymore due to other removed controllers.
- Renamed method `getFirstCOEOfApplication` to `getFirstDisbursementSchedule` and allow search by application id and disbursement it. Apparently the Vue application was using the disbursement id to send a value that was consumed in the API as an application id.
- Added pipe validators(`ParseIntPipe`, `ParseEnumPipe`) to API parameters.
- Renamed methods on API and Vue to be the same and share the same comments.


sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.institutions.controller.ts
```diff
@@ -73,9 +80,11 @@ const INVALID_TUITION_REMITTANCE_AMOUNT_MESSAGE =
   "Tuition amount provided should be lesser than both (Actual tuition + Program related costs) and (Canada grants + Canada Loan + BC Loan).";
 
 @AllowAuthorizedParty(AuthorizedParties.institution)
-@Controller("institution/location")
-@ApiTags("institution")
-export class ConfirmationOfEnrollmentController extends BaseController {
+@Controller("location")
+@ApiTags(
+  `${ClientTypeBaseRoute.Institution}-location[Confirmation Of Enrollment - CoE]`,
```
ann-aot, line 85: 👍

henryoforeh-aot (approved):

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -556,75 +557,79 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
...
-          coeDeniedReasonId === COE_DENIED_REASON_OTHER_ID
-            ? otherReasonDesc
-            : null,
-      })
-      .where("application.id = :applicationId", { applicationId })
-      .andWhere("coeStatus = :required", { required: COEStatus.required })
-      .execute();
+  ): Promise<DisbursementSchedule> {
+    const coeToUpdate = new DisbursementSchedule();
+    coeToUpdate.id = disbursementScheduleId;
```
ann-aot, line 574: is this method safe to update a row? if by any chance the id is not present, it will create a new column with the id right? will affect the default id iteration of the db?. 

doesn't the below where condition required for updating?  `.where("application.id = :applicationId", { applicationId })
      .andWhere("coeStatus = :required", { required: COEStatus.required })`

let me know if I am missing anything

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.institutions.controller.ts
andrewsignori-aot: Thanks @ann-aot. I was not so sure about it but I believe that it looks better on Swagger.

sources/packages/api/src/utilities/date-utils.ts
```diff
@@ -50,19 +50,16 @@ export const dateString = (date: string | Date): string => {
...
- * @param lastDate Last date.
+ * Difference in days between endDate and startDate (endDate-startDate).
+ * @param endDate end date.
+ * @param startDate start date.
  * @returns the date difference in days.
  */
 export const dateDifference = (
-  firstDate: string | Date,
-  lastDate: string | Date,
+  endDate: string | Date,
```
ann-aot, line 59: I think the order should be reversed, right?  `firstDate` is `startDate` and `lastDate` is `endDate`

sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
```diff
@@ -1,115 +1,128 @@
...
   }
 
-  public async rollbackCOE(
+  /**
+   * Creates a new Student Application to maintain history,
+   * overriding the current one in order to rollback the
+   * process and start the assessment all over again.
+   * @param locationId location id executing the COE rollback.
+   * @param disbursementScheduleId application to be rolled back.
+   * @returns the id of the newly created Student Application.
```
ann-aot, line 84: the comment looks wrong. no need of `returns` comment

sources/packages/api/src/utilities/date-utils.ts
andrewsignori-aot: I would say that both are valid.
If you would be doing a subtraction of 2 dates, I would say that the most natural way would be doing BiggerDate-SmallerDate, which would be EndDate-StartDate, right?
For instance, it is the same that `dayjs` follows.
image

Does it make sense? We can talk about it 😉

sources/packages/web/src/services/http/ConfirmationOfEnrollmentApi.ts
andrewsignori-aot: I will remove from both methods Api/Service.

sources/packages/api/src/utilities/date-utils.ts
ann-aot: okay..that good

ann-aot (commented):
Looks good to me. just a few comments

sources/packages/api/src/route-controllers/confirmation-of-enrollment/confirmation-of-enrollment.institutions.controller.ts
ann-aot: ya yes

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
andrewsignori-aot: A couple of notes.

About "will affect the default id iteration of the db?", typeorm prevents it. If you try to force it the column is decorated with the `PrimaryGeneratedColumn()` that will prevent it.

About "if by any chance the id is not present", the id is not received from the client and used without validation. At this point, the id is already validated and the method is called considering it. The service method can have more safeguards to ensure its own integrity, so I can change from a save to an update, for instance.

About "doesn't the below where condition required for updating?", I will add some of the conditions back but not in the same way. The update MUST NOT HAPPEN using the application id. This code is probably from the time when the COE was on the application.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -99,15 +99,20 @@ export class ApplicationStudentsController extends BaseController {
...
       );
-
-    const firstCOE =
-      await this.disbursementScheduleService.getFirstCOEOfApplication(
+    const firstCOEPromise =
+      this.disbursementScheduleService.getFirstDisbursementSchedule({
         applicationId,
-      );
+      });
+    const [applicationData, firstCOE] = await Promise.all([
```
guru-aot, line 110: Cool 👍

sources/packages/api/src/route-controllers/confirmation-of-enrollment/models/confirmation-of-enrollment.dto.ts
```diff
@@ -49,7 +49,18 @@ export class ConfirmationOfEnrollmentAPIInDTO {
 /**
  * Read only Interface for study break item.
  */
-export interface StudyBreak {
+export class StudyBreak {
```
guru-aot, line 52: Please update comment to Class from "interface"

sources/packages/api/src/services/application/application.service.ts
```diff
@@ -1012,16 +1010,25 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        transactionalEntityManager.getRepository(Application);
+      // Application must be saved to have the id properly generated and associated
+      // to the assessment. This is a special case where the application is associated
+      // in the assessment record and the assessment record is associated as a
+      // currentAssessment also in the application.
+      await applicationRepository.save(newApplication);
+      newApplication.creator = auditUser;
+      newApplication.studentAssessments = [originalAssessment];
+      newApplication.currentAssessment = originalAssessment;
+      await applicationRepository.save([appToOverride, newApplication]);
```
guru-aot, line 1025: 👍

sources/packages/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
andrewsignori-aot: Method changed.

sources/packages/api/src/route-controllers/confirmation-of-enrollment/models/confirmation-of-enrollment.dto.ts
andrewsignori-aot: Changed!

guru-aot (approved):
LGTM. Nice work @andrewsignori-aot

sources/packages/api/src/route-controllers/confirmation-of-enrollment/models/confirmation-of-enrollment.dto.ts
```diff
@@ -29,12 +29,12 @@ export interface ApplicationDetailsForCOEDTO {
   disbursementDate: string;
 }
 
-export interface COEDeniedReasonDto {
+export class COEDeniedReasonAPIOutDTO {
```
dheepak-aot, line 32: Not a part of the PR. I assume this class is for loading the dropdown values in a prime vue component. 
We can add a todo to move it to common vuetify dto when we change to vuetify.

ann-aot (approved):
Good work @andrewsignori-aot  👍 Thanks for doing the changes

dheepak-aot (approved):
Nice work @andrewsignori-aot . Approving as I have only a clarification and a minor comment. 👍

sources/packages/api/src/route-controllers/confirmation-of-enrollment/models/confirmation-of-enrollment.dto.ts
andrewsignori-aot: We already have a generic version called `OptionItemAPIOutDTO`. I had some back and forth about it and end up deciding no change it because maybe this one could be extended with more info.
If I was building it I would use the generic one  `OptionItemAPIOutDTO`, since it is already there, for this case, I just left it as it is right now.

Issue #1312, opened 2022-07-13
Remove cascade deletes


abschwenker:
**- Acceptance Criteria**
- [ ] Create 1 migration script for remove delete cascade rules in DB
- [ ] Ensure the delete cascade between assessment and application is working correctly (not deleting) #1162

PR #1340, opened 2022-07-29
[API][SIMS #1312] Remove delete cascade

henryoforeh-aot:
- Created migration script to remove all existing delete cascade rules.
- Created script to roll back by adding delete cascade rules.
- Tested delete cascade between application and assessment after removing delete cascade rules.


sources/packages/api/src-sql/DeleteCascades/Add-cascade-rules.sql
```diff
@@ -0,0 +1,203 @@
+-- Create reference between application_exception_requests and application_exceptions --
```
andrewsignori-aot, line 1: When the rollback script looks like something that we are adding we usually add some prefixes like `Rollback-drop-cascade-rules.sql`.

henryoforeh-aot: I will update

ann-aot: periods for all comments

henryoforeh-aot:
Creating a new PR

Issue #66, opened 2021-01-08
Define Ministry User Access and Permissions (per IDIR)


mzmichellesmith:
**Additional Context**
- Roles will be defined in keycloak groups
- TBTB will make idir users
- Spreadsheet of permissions located General -> Files -> Analysis -> SIMS User Profiles and Access Permissions
- Read roles from token, parse token (doesnt need to be in store because it is there in object) and search for access
- We do not need to restrict the URL, just the button
- API needs to be changed to read the roles

**Acceptance Criteria**
- [ ] Use spreadsheet for keycloak group names (already created in dev)
- [ ] If user does not have permissions, disable buttons 
- [ ] Ensure api validations block user
- [ ] Map all roles (descriptors in left hand column of user profile matrix - "Create Institution", "Search Students" etc) to permissions

Suggestion for Vue checks to be executed in a generic way.
![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/1f15339c-b494-4d59-8b36-69255bbfa90b)

PR #1357, opened 2022-08-08
[api][sims-#66] - Define Ministry User Access and Permissions (per IDIR)#66

ann-aot:
- updates role enum and role decorator
 - added AEST roles to the `post` and `patch` endpoints
 - FIXED button alignment and variant AEST student file upload modal

### DIFF
--- a/sources/packages/api/src/auth/guards/groups.guard.ts
+++ b/sources/packages/api/src/auth/guards/groups.guard.ts
@@ -33,7 +33,8 @@ export class GroupsGuard implements CanActivate {
     }
     const userToken = user as IUserToken;
     // Check if the user has any of the groups required to have access to the resource.
-    // UserGroups 'aest/user/x' and 'aest/user' are valid, we are here looking for matches, not the exact string.
+    // UserGroups 'aest/user/x' and 'aest/user' are valid, as they start with 'aest/user'
+    // we are here looking for matches, not the exact string.
     return requiredGroups.some((group: string) =>
       userToken.groups?.some((userGroup) => userGroup.startsWith(group)),
     );
--- a/sources/packages/api/src/auth/guards/groups.guard.ts
+++ b/sources/packages/api/src/auth/guards/groups.guard.ts
@@ -33,8 +33,9 @@ export class GroupsGuard implements CanActivate {
     }
     const userToken = user as IUserToken;
     // Check if the user has any of the groups required to have access to the resource.
+    // UserGroups 'aest/user/x' and 'aest/user' are valid, we are here looking for matches, not the exact string.
     return requiredGroups.some((group: string) =>
-      userToken.groups?.includes(group),
+      userToken.groups?.some((userGroup) => userGroup.startsWith(group)),
     );
   }
 }
--- a/sources/packages/api/src/auth/guards/roles.guard.ts
+++ b/sources/packages/api/src/auth/guards/roles.guard.ts
@@ -21,6 +21,10 @@ export class RolesGuard implements CanActivate {
       return true;
     }
     const { user } = context.switchToHttp().getRequest();
-    return requiredRoles.some((role) => user.roles?.includes(role));
+    let userRoles;
+    if (user.resource_access && user.azp) {
+      userRoles = user.resource_access[user.azp].roles;
+    }
+    return requiredRoles.some((role) => userRoles?.includes(role));
   }
 }
--- a/sources/packages/api/src/auth/guards/roles.guard.ts
+++ b/sources/packages/api/src/auth/guards/roles.guard.ts
@@ -2,6 +2,7 @@ import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common";
 import { Reflector } from "@nestjs/core";
 import { ROLES_KEY } from "../decorators/roles.decorator";
 import { Role } from "../roles.enum";
+import { IUserToken } from "../userToken.interface";
 
 /**
  * Allow the authorization based on roles making use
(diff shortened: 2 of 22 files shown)


sources/packages/api/src/auth/jwt.strategy.ts
```diff
@@ -41,11 +41,11 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
     // also as needed to be validate using a decorator/annotation.
     userToken.roles = [];
     if (payload.resource_access) {
-      Object.keys(payload.resource_access).forEach((value) => {
-        payload.resource_access[value].roles.forEach((roleValue: string) => {
+      payload.resource_access[payload.azp].roles.forEach(
```
andrewsignori-aot, line 44: Do we need this processing to happen? I believe that the token is already in the format that we expected, right?

sources/packages/api/src/auth/roles.enum.ts
```diff
@@ -2,5 +2,34 @@
  * Application roles to be used in alongside with RolesGuard.
  */
 export enum Role {
+  /**
+   * Student roles.
```
andrewsignori-aot, line 6: Can we check if it is used, I believe that it can be removed.

sources/packages/api/src/route-controllers/education-program-offering/education-program-offering.aest.controller.ts
```diff
@@ -30,6 +35,7 @@ import {
   OfferingChangeAssessmentAPIInDTO,
 } from "./models/education-program-offering.dto";
 import { CustomNamedError } from "../../utilities";
+import { Role } from "src/auth/roles.enum";
```
ann-aot, line 38: i will fix this

sources/packages/api/src/auth/guards/groups.guard.ts
```diff
@@ -33,8 +33,9 @@ export class GroupsGuard implements CanActivate {
     }
     const userToken = user as IUserToken;
     // Check if the user has any of the groups required to have access to the resource.
+    // UserGroups 'aest/user/x' and 'aest/user' are valid, we are here looking for matches, not the exact string.
```
dheepak-aot, line 36: Comment can also mention that both the groups are valid as they start with 'aest'.

sources/packages/api/src/auth/roles.enum.ts
```diff
@@ -2,5 +2,34 @@
  * Application roles to be used in alongside with RolesGuard.
  */
 export enum Role {
```
dheepak-aot, line 4: Are these roles AEST specific? If so we can rename the enum to AESTRoles?

sources/packages/api/src/auth/guards/groups.guard.ts
```diff
@@ -33,8 +33,9 @@ export class GroupsGuard implements CanActivate {
...
     return requiredGroups.some((group: string) =>
-      userToken.groups?.includes(group),
+      userToken.groups?.some((userGroup) => userGroup.startsWith(group)),
```
guru-aot, line 38: The requiredGroups is a list and the userGroup is also a list, using some will return true if one of the required userGroup is available in the group list. It will not check for all the required userGroups, Is the scenario right?

sources/packages/api/src/auth/jwt.strategy.ts
ann-aot: yup.i will update

sources/packages/api/src/auth/roles.enum.ts
ann-aot: actually, I was about to remove this. then I realized that its use in 
image

ann-aot: its a test controller, i think it can be removed

henryoforeh-aot (approved):

sources/packages/api/src/auth/roles.enum.ts
ann-aot: As discussed, I believe this is a generic Role.

sources/packages/api/src/auth/guards/groups.guard.ts
ann-aot: yes, `.some()` will check at least one matches the logic. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

sources/packages/api/src/auth/roles.enum.ts
dheepak-aot: May be the enum when it was created is to represent generic roles.. But all the aest roles in this are aest-specific roles.

sources/packages/api/src/testHelpers/controllers/auth-test/auth-test.controller.ts
```diff
@@ -42,7 +42,7 @@ export class AuthTestController {
   /**
    * Only authenticated users with specific role will have access to this endpoint.
    */
-  @Roles(Role.Student)
+  @Roles(Role.AESTReports)
```
dheepak-aot, line 45: I know that this is test controller. But why did we use AESTReports role to authorize this?

sources/packages/api/src/auth/guards/roles.guard.ts
```diff
@@ -21,6 +21,10 @@ export class RolesGuard implements CanActivate {
       return true;
     }
     const { user } = context.switchToHttp().getRequest();
-    return requiredRoles.some((role) => user.roles?.includes(role));
+    let userRoles;
```
andrewsignori-aot, line 24: This can be changed as below:
```typescript

if (user.resource_access && user.azp) {
  const userRoles = user.resource_access[user.azp].roles;
  return requiredRoles.some((role) => userRoles?.includes(role));
}
return true;
```

andrewsignori-aot (approved):

sources/packages/api/src/auth/guards/groups.guard.ts
guru-aot: 👍

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/api/src/testHelpers/controllers/auth-test/auth-test.controller.ts
ann-aot: as you said. its test controller. we are testing a role for a aest user

dheepak-aot (approved):
Nice work @ann-aot 👍

Issue #1217, opened 2022-05-12
Test cases management POC in TestQuality


TeresaBia:
**- Additional Context**
- This is a test management task not execution of test cases
- Automation of test cases execution on TestQuality will come at a later date
- Conduct a POC for TestQuality to implement an easy way to organize test scenarios and results and for our Team or Ministry to find and follow

**- Requirements**
High priority
- View of test cases
- View of test results with run date/time
- Organization of test scenarios (higher level)

Lower priority
- Notifications of run results
- Automatic or easy integration for creation of bugs
- Organization of test steps (lower level)

**- Acceptance Criteria**
- [x] Determine based off the above criteria if TestQuality will fit the needs of SIMS, include notes in comments
- [x] Create POC for test management which includes sample test cases and pass/fail scenarios
- [x] Upload test cases to TestQuality
- [x] Upload test results to TestQuality
- [x] Write upload functions/scripts to upload to TestQuality
- [x] Use access token to represent using github actions secrets

PR #1378, opened 2022-08-17
 [cypress][SIMS #1217] TestCase Management Tool Support

velang01:
Adding support to upload the test case results to TestCase Management Tool - Test Quality


testing/src/cypress/page-objects/student-objects/WelcomeObject.ts
```diff
@@ -16,6 +16,10 @@ export default class WelcomeObject {
   signUpWithBCSCButton() {
     return cy.contains("Sign Up with BCSC");
   }
+  
+  loginSignUpButton(){
+    return cy.contains(" Login / Sign up with BCSC ")
```
henryoforeh-aot, line 21: Any particular reason for the spacing in the text ?

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
+var request = require('request');
+var fs = require('fs');
+import * as dotenv from "dotenv";
+
+
+
+token = undefined
```
henryoforeh-aot, line 7: Please use let to declare this variable. let token = undefined;

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
...
+const tq_username = process.env.TQ_USER_NAME;
```
henryoforeh-aot, line 8: Constants need to be in upper case.

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
...
+        'Content-Type': 'application/json'
+  },
+  body: JSON.stringify({
+    "grant_type": "password",
+    "client_id": "2",
+    "client_secret": `"${TQ_SECRET}"`,
+    "username": `"${TQ_USER_NAME}"`,
+    "password": `"${TQ_PASSWORD}"`
+  })
+};
```
henryoforeh-aot, line 24: Please create  new line here.

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
...
+function loginToTQ (callback){
+    request(options, function (error, response) {
+        if (error) throw new Error(error);
+        const obj =  JSON.parse(response.body)
+        token = obj["access_token"]
+        // console.log("Fetching the TQ Token")
```
henryoforeh-aot, line 30: You can delete this commented code if you are no longer using it.

testing/src/cypress.config.ts
```diff
@@ -34,5 +30,8 @@ export default defineConfig({
     token_url:
       "https://dev.oidc.gov.bc.ca/auth/realms/jxoe2o46/protocol/openid-connect/token",
     update_institution_url: "/api/institutions/institution",
+    studentURL: "https://test-aest-sims.apps.silver.devops.gov.bc.ca/student/login",
```
ann-aot, line 33: token URL is `dev` and student URL is `test`

andrewsignori-aot (approved):
Wrongly approved here, please disregard.

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
+var request = require('request');
```
andrewsignori-aot, line 1: The file must be converted to typescript.
We do not have plain javascript in the project unless it is really required.

testing/src/uploadResultTQ.js
andrewsignori-aot, line 8: `tq_username`, `tq_password` and `tq_secret` are declared and never used.

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
...
+var fs = require('fs');
+import * as dotenv from "dotenv";
+
+
```
andrewsignori-aot, line 5: Please ensure Prettier is installed and formatting the code when saving (configurable on VSCode). These blank lines would be automatically removed, for instance, if it was enabled.

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
...
+
+token = undefined
+const tq_username = process.env.TQ_USER_NAME;
+const tq_password = process.env.TQ_PASSWORD;
+const tq_secret = process.env.TQ_SECRET;
+var options = {
```
andrewsignori-aot, line 11: The other projects are using Axios to execute the HTTP requests, maybe it would be good to follow the same.

testing/src/uploadResultTQ.js
```diff
@@ -0,0 +1,58 @@
...
+    "password": `"${TQ_PASSWORD}"`
+  })
+};
+function loginToTQ (callback){
+    request(options, function (error, response) {
+        if (error) throw new Error(error);
+        const obj =  JSON.parse(response.body)
+        token = obj["access_token"]
+        // console.log("Fetching the TQ Token")
+        callback()
```
andrewsignori-aot, line 31: It would be good to use async/await pattern instead callback to be aligned with the rest of the solution.

testing/src/cypress.config.ts
```diff
@@ -34,5 +30,8 @@ export default defineConfig({
...
+    username: "",
```
andrewsignori-aot, line 34: These variables seem to be too generic, are they supposed to be used for student login?
If this is not the focus of the PR we can handle this later.

testing/src/cypress.config.ts
```diff
@@ -34,5 +30,8 @@ export default defineConfig({
     token_url:
```
andrewsignori-aot, line 30: Are these env variables in the correct place? 
If this is not the focus of the PR we can handle this later, but I believe that we are missing the proper env file or equivalent for cypress (https://docs.cypress.io/api/cypress-api/env).

andrewsignori-aot (commented):
Nice one to have as a first PR 😉 

testing/src/cypress.config.ts
velang01: I believe so - its the right place. But yeah we can discuss if we need to move it to .env file altogether.

testing/src/cypress.config.ts
velang01: The original was set against to dev. But we want to run this against test env for the most part. I ll update all all the endpoints to point the test

testing/src/cypress.config.ts
velang01: Yes they are for login. we just need to set the specific  set test users for automation. But we can handle this into our env like one above

testing/src/cypress/page-objects/student-objects/WelcomeObject.ts
velang01: Wanted to keep it consistent with the DOM element. But will removed the trailing white spaces

testing/src/cypress.config.ts
velang01: I ll change them to a dev env file in a separate PR

testing/src/cypress.config.ts
velang01: updated. but all these urls will be moved to separate env file.

testing/src/cypress/page-objects/student-objects/WelcomeObject.ts
```diff
@@ -17,6 +17,10 @@ export default class WelcomeObject {
     return cy.contains("Sign Up with BCSC");
   }
 
+  loginSignUpButton() {
+    return cy.contains("Login / Sign up with BCSC");
```
andrewsignori-aot, line 21: Not part of the PR, but we will not be using any sort of selectors other than `data-cy` to get the elements from the UI. As needed we can keep adding the `data-cy` on Vue and form.io definitions.

testing/src/package.json
```diff
@@ -15,7 +15,8 @@
     "generate-report": "marge cypress/reports/mochareports/report.json -f report -o cypress/reports/mochareports",
     "posttest": "npm run combine-reports && npm run generate-report",
     "test": "npm run scripts || npm run posttest",
-    "Run headless for all modes": "npm run test"
+    "Run headless for all modes": "npm run test",
+    "uploadResultTQ" : "npx ts-code uploadResult.ts"
```
andrewsignori-aot, line 19: I believe that we do not need the npx here. It can be executed in the same way that we do for the _setup:db_: 
```
setup:db": "ts-node scripts/setupDB
```

henryoforeh-aot (approved):

testing/src/uploadResult.ts
```diff
@@ -0,0 +1,59 @@
...
+import fs from "fs";
+
+const TQ_USERNAME = process.env.TQ_USER_NAME;
+const TQ_PASSWORD = process.env.TQ_PASSWORD;
+const TQ_SECRET = process.env.TQ_SECRET;
+const TQ_RUN_ID = process.env.TQ_RUN_ID; //( default id is 26606. Will need to change once we have integrated TQ account for the team)
+
+const getAuthToken = async (): Promise<AxiosResponse<any>> => {
+  const url = "https://api.testquality.com/api/oauth/access_token";
+  const body = JSON.stringify({
```
andrewsignori-aot, line 12: No need to use the `JSON.stringify` while proving the payload to axios.

testing/src/uploadResult.ts
```diff
@@ -0,0 +1,59 @@
+import axios, { AxiosResponse } from "axios";
+import FormData from "form-data";
+import fs from "fs";
+
+const TQ_USERNAME = process.env.TQ_USER_NAME;
+const TQ_PASSWORD = process.env.TQ_PASSWORD;
+const TQ_SECRET = process.env.TQ_SECRET;
+const TQ_RUN_ID = process.env.TQ_RUN_ID; //( default id is 26606. Will need to change once we have integrated TQ account for the team)
+
+const getAuthToken = async (): Promise<AxiosResponse<any>> => {
```
andrewsignori-aot, line 10: The return object type is probably a known type like `string`. We should use `any` only under exceptional circumstances.
The result should probably be typed as `Promise<string>`.

andrewsignori-aot, line 10: Please keep the method declaration consistent. Please use here the same style used for `async function uploadResultFile()`.

testing/src/package.json
```diff
@@ -27,6 +28,8 @@
     "mochawesome-report-generator": "^6.2.0",
     "ts-loader": "^9.2.7",
     "typescript": "^4.6.2",
-    "webpack": "^5.70.0"
+    "webpack": "^5.70.0",
```
andrewsignori-aot, line 31: Is `axios` missing on the `dependencies` llist?

testing/src/uploadResult.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+async function uploadResultFile() {
+  const auth = await getAuthToken();
+  const url = `https://api.testquality.com/api/plan/${TQ_RUN_ID}/junit_xml`;
+  var formData = new FormData();
+  var resultFile = fs.createReadStream("test-results.xml");
+  formData.append("file", resultFile, "test-results.xml");
+  try {
+    if (String(auth).length > 1) {
```
andrewsignori-aot, line 42: No need to do a check here, IMO. If the API call to get the access token is 200, the token will be there, is it fail any different HTTP code Axios will throw an exception.

testing/src/uploadResult.ts
```diff
@@ -0,0 +1,59 @@
...
+import FormData from "form-data";
+import fs from "fs";
+
+const TQ_USERNAME = process.env.TQ_USER_NAME;
+const TQ_PASSWORD = process.env.TQ_PASSWORD;
+const TQ_SECRET = process.env.TQ_SECRET;
+const TQ_RUN_ID = process.env.TQ_RUN_ID; //( default id is 26606. Will need to change once we have integrated TQ account for the team)
+
+const getAuthToken = async (): Promise<AxiosResponse<any>> => {
+  const url = "https://api.testquality.com/api/oauth/access_token";
```
andrewsignori-aot, line 11: Please move the const `url` in both methods to the top and/or create a const to the base URL that both constants are sharing.

testing/src/uploadResult.ts
```diff
@@ -0,0 +1,59 @@
...
+  formData.append("file", resultFile, "test-results.xml");
+  try {
+    if (String(auth).length > 1) {
+      const settings = {
+        headers: {
+          ...formData.getHeaders(),
+          Authorization: `Bearer ${auth}`,
+        },
+      };
+      const response = await axios.post(url, formData, settings);
```
andrewsignori-aot, line 49: The `const response` is not used, please remove it.

andrewsignori-aot (commented):
Thanks for doing the changes. Please take a look at the comments, I believe that only minor ones now. The conversation to the ts file was also well done, thanks for taking care of it.

dheepak-aot (approved):
LGTM @velang01 👍

Issue #419, opened 2021-05-05
Analysis - Enable CORS on SIMS API (allow API call from formsflow.ai)


andrewsignori-aot:
**Additional Context**
- The CORS was enabled on SIMS Api in a permissive way to allow the formsflow.ai to have access to SIMS API.
We need to configure CORS to be more restrictive as possible or consider deploying the formsflow.ai applications under the same domain.

As an output of the analysis it has been found that, in API CORS must be restricted to all web domains. 
As camunda to SIMS API call does not happen  through browser, CORS does not have any impact on Camunda API calls.
As SIMS Web and API are sharing the same origin, they API calls from SIMS web to API are not considered as Cross Origin Resource Sharing.

**Acceptance Criteria**
- [x] Enable CORS restriction in all non developement/openshift environments
- [ ] Determine method to resolve CORS issue
- [ ] Put api+web under same domain and disable CORS
OR
- [ ] explicitly define domain to enable CORS for a particular domain

dheepak-aot:
Tested in DEV by submitting application no: 2023000016 (Application id: 1427)

Camunda was able to call SIMS API and successfully update the application. 

![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/de8bca2a-17cb-465a-9922-e2c31f4933c3)

Camunda workflow details: 
![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/323b7c79-a24c-47bf-94e2-51107734f5a8)

PR #1396, opened 2022-08-31
[API][SIMS #419] - Restrict CORS on all non-local-development environment

dheepak-aot:
Defined a variable to allow cors request based on environment variable NODE_ENV.

- [x] CORS is restricted on all openshift environments.
- [x] CORS is allowed on local environment as our API(http://localhost:3000) and Web(http://localhost:8080) run on different domain.
- [x] The value of MAX_UPLOAD_PARTS has been updated to 4 from 3. Since with file interceptor is rejecting file with 3 parts if the maximum parts is set to 3.

### DIFF
--- a/sources/packages/api/src/utilities/system-configurations-constants.ts
+++ b/sources/packages/api/src/utilities/system-configurations-constants.ts
@@ -99,7 +99,9 @@ export const MAX_UPLOAD_FILES = 1;
  * For multipart forms, the max number of parts (fields + files).
  * 3 means 'the file' + uniqueFileName + group.
  */
-export const MAX_UPLOAD_PARTS = 3;
+// TODO: The value has been set to 4 from 3 to fix the ongoing file upload issue.
+// TODO: On further investigation re-evaluate accordingly.
+export const MAX_UPLOAD_PARTS = 4;
 
 /**
  * Group name associated with the files uploaded by the Ministry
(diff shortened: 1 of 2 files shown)


guru-aot (approved):
LGTM

andrewsignori-aot (approved):
Looks good and thanks for fixing the issue with the file upload.

sources/packages/api/src/utilities/system-configurations-constants.ts
```diff
@@ -99,7 +99,9 @@ export const MAX_UPLOAD_FILES = 1;
  * For multipart forms, the max number of parts (fields + files).
  * 3 means 'the file' + uniqueFileName + group.
  */
-export const MAX_UPLOAD_PARTS = 3;
+// TODO: The value has been set to 4 from 3 to fix the ongoing file upload issue.
```
ann-aot, line 102: okay

sources/packages/api/src/utilities/system-configurations-constants.ts
```diff
@@ -99,7 +99,9 @@ export const MAX_UPLOAD_FILES = 1;
...
+// TODO: On further investigation re-evaluate accordingly.
```
ann-aot, line 103: 👍

ann-aot (approved):
Good investigation and work @dheepak-aot  👍

Issue #1115, opened 2022-03-25
UI updates for Ministry Part 1 of 3


TeresaBia:
**Additional Context:**
Related UX ticket [#1076](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/1076)
- This ticket is to cover the Search students and subpages

Related UX ticket (Ministry user): [#1111](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/1111)
Related UX ticket (Ministry views): [#1112](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/1112)

**Acceptance Criteria:**
- Ensure above pages matches figma and below criteria are complete
- [ ] Apply common components to all applicable areas
- [ ] Remove remaining Primvue components Change modals to Vuetify (except for file uploaders)
- [ ] Buttons outside of the forms - component already created (ministry create institution for example)
- [ ] Buttons always enabled so validations show up (taken care of when buttons are outside forms)
- [ ] For Vue based forms (eg. search) always enable button and display validations
- [ ] Use full-page container
- [ ] Use the formio submit call instead of check validity method so the error summary shows at the top - follow how student application works
- [ ] Remove from Dev deprecated modals
- [ ] Include image alt-tags

PR #1398, opened 2022-09-02
[Web][SIMS-#1115]UI updates ministry - Part 1 (Login and Sidebar)

guru-aot:
image

image


sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
```diff
@@ -56,103 +73,85 @@
...
+import { ref } from "vue";
 import { AESTRoutesConst } from "@/constants/routes/RouteConstants";
 import { MenuModel, Role } from "@/types";
 import CheckPermissionRole from "@/components/generic/CheckPermissionRole.vue";
 
 export default {
   components: { CheckPermissionRole },
   setup() {
-    const router = useRouter();
+    const drawer = ref("drawer");
```
andrewsignori-aot, line 84: Is there a reason to have this? If not please remove it. If it was copied from another code and there is no reason to keep it, please remove it.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
```diff
@@ -1,51 +1,68 @@
...
     </v-list>
-    <v-list density="compact" nav>
-      <v-list-subheader>Student requests</v-list-subheader>
+    <v-list
+      active-class="active-sidebar-item"
+      density="compact"
+      bg-color="background"
+      active-color="primary"
+    >
+      <v-list-subheader class="nav-subtitle">Student requests</v-list-subheader>
```
andrewsignori-aot, line 23: The spacings in the left menu seem different from Figma, I am not sure if it is part of the PR or not, but can we have at least the proper case adjusted for the sub-headers, as shown below?

image

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -1,41 +1,58 @@
...
-      ><div>
-        Please use your IDIR to authenticate. You must be previously authorized
-        by the system administrator in order to successfully login.
-      </div>
-      <banner
-        class="mt-2"
-        :type="BannerTypes.Error"
-        v-if="showNotAllowedUser"
-        summary="The user was validated successfully but is not currently allowed to
+  <full-page-container layout-template="centered-card" :enable-max-width="true">
```
andrewsignori-aot, line 2: What is the idea behind using `:enable-max-width="true"`?
Is it to apply a fixed `730px` max-width? Is it something common for every `full-page-container` that will need a max-width defined?
For the login pages, do we need to use the `full-page-container`?
@ann-aot @dheepak-aot

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -1,41 +1,58 @@
...
-            prepend-icon="fa:fa fa-user"
-          >
-            Login with IDIR
-          </v-btn>
-        </v-row>
-      </div>
-    </v-card-text>
-  </v-card>
+            />
+            <!-- TODO: v-btn inside v-card-actions is not working properly. try in vuetify 3 stable version -->
```
andrewsignori-aot, line 32: I do not believe that this is still necessary.

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -1,41 +1,58 @@
...
+        </v-card-header>
+        <content-group>
+          <h3 class="category-header-medium primary-color">
+            Login for ministry staff
+          </h3>
+          <p class="sign-in-description">
+            With authorization from the system administrator, login here using
+            your IDIR to access the AEST portal.
+          </p>
+
```
andrewsignori-aot, line 22: Please remove the extra spaces unless they are meaningful.

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -1,41 +1,58 @@
...
-            Login with IDIR
-          </v-btn>
-        </v-row>
-      </div>
-    </v-card-text>
-  </v-card>
+            />
+            <!-- TODO: v-btn inside v-card-actions is not working properly. try in vuetify 3 stable version -->
+          </p>
+
```
andrewsignori-aot, line 34: Please remove the extra spaces unless they are meaningful.

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -1,41 +1,58 @@
...
+                Login with IDIR
+              </v-btn>
+            </v-row>
+          </div>
+        </content-group></v-col
+      ><v-col cols="3"
+        ><v-img
+          height="260"
+          class="mt-8"
+          alt="A person entering their login information into a digital screen.."
```
andrewsignori-aot, line 52: Please remove the double periods at the end of the sentence.

andrewsignori-aot (commented):
Nice work, the UI is looking better. Please take a look at the comments.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
dheepak-aot: Also the Search Student and Search Institution are singular in figma but plural here. 
image

guru-aot: Yes these changes has been verified by @hellolynn-tbtb and the figma was not current. @hellolynn-tbtb Please verify.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
```diff
@@ -56,103 +73,85 @@
...
         label: "Home",
         icon: "mdi-home-outline",
-        command: () => {
-          router.push({
-            name: AESTRoutesConst.AEST_DASHBOARD,
-          });
-        },
+        command: () => ({
+          name: AESTRoutesConst.AEST_DASHBOARD,
+        }),
```
dheepak-aot, line 89: The existing code doesn't hurt. But title and icon can be set in the JSON itself like this. So that we don't have to re-assign them as attribute value in the component.

image

dheepak-aot: @andrewsignori-aot @ann-aot @guru-aot what is your opinion?

sources/packages/web/src/views/aest/Login.vue
dheepak-aot: I believe the idea is to ensure a max-width(730 px) for full-page-container with fluid(full width) false. 
But I think the max-width class can be set based on fullWidth as false.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
dheepak-aot: There is a purpose of v-model for the drawer as per vuetify. I think we should keep the v-model.

image

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
andrewsignori-aot: Sorry, I am not following it, can you please give an example of how it would improve the current code?

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
andrewsignori-aot: There is a purpose for every single property that exists in Vuetify 😄 
This one here does not seem to be used in the right way or causing any difference. Can you please provide a single benefit to keeping it? Maybe I am missing something.

sources/packages/web/src/views/aest/Login.vue
andrewsignori-aot: The `full-page-container` already has a max-width class that is used when it needs to be centered (`full-page-container-size`).
This one is creating a new max-width defined as 730px which I guess was used only for login pages so far.
The question is: Is `enable-max-width` something generic with the potential to be used anywhere else? Do we have other possible places that will need to use the 730px?

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
andrewsignori-aot: Got it now after our chat. Looks better to me also 👍

dheepak-aot: As discussed 😊

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
dheepak-aot: Understood after our discussion @andrewsignori-aot. @guru-aot the v-model is not useful for a permanent type drawer.

sources/packages/web/src/views/aest/Login.vue
dheepak-aot: I went further and saw other pages. It seems like enable-max-width is only specific to login page.(And must not be used on other pages where fluid is false as it is breaking them). @ann-aot can surely give a better answer.

ann-aot: yes the idea of `enable-max-width` is to apply `730px` max-width for login cards as discussed with @hellolynn-tbtb. As discussed with @andrewsignori-aot, we can avoid using `full-page-container` for login pages @guru-aot, instead, we can use `<v-container` with `v-cards` with `max-width = "730px"`.

sources/packages/web/src/views/aest/AESTDashboard.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-container>
+  <full-page-container>
     <formio formName="aestdashboard"></formio>
```
ann-aot, line 3: please use `formio-container`

ann-aot: is the dashborad changes part of the next PR?

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
guru-aot: Drawer is used to show the sidebar, without the property side bar is not showing.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
guru-aot: For now not implementing as the :to property cannot be assigned to the items. Navigation is not supported.

sources/packages/web/src/views/aest/AESTDashboard.vue
guru-aot: yes

sources/packages/web/src/views/aest/Login.vue
dheepak-aot: In that case for we create a common component(e.g. login-container) to use in all login pages. 
If not in this PR, you can also create the common component in next PR @guru-aot .

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
dheepak-aot: We can give it a try later in upcoming PR/UI stories.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
ann-aot: institution sidebar subtitle font is different in Figma, maybe we can check with @hellolynn-tbtb 
image

dheepak-aot (approved):
Nice work @guru-aot 👍
We can discuss about the common component for login container in nerx PR.

ann-aot (approved):
Good work @guru-aot. Thanks for doing the changes

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
```diff
@@ -1,51 +1,68 @@
...
+        :to="offerings.command()"
       />
     </v-list>
     <template #append>
       <v-list density="compact" nav>
         <check-permission-role :role="Role.AESTReports">
           <template #="{ notAllowed }">
             <v-list-item
-              @click="reports.command"
+              :to="reports.command()"
```
andrewsignori-aot, line 65: I believe that there is no reason to call a method here. The value can be directly set in the same way as the other ones.

andrewsignori-aot (approved):
Thanks for doing the changes. Please lets take care of the below comment in the next PR.
https://github.com/bcgov/SIMS/pull/1398/files#r962022862

hellolynn-tbtb:
@guru-aot Hi guru, looking good. Just missed this one (removing the uppercase formatting). Thanks.
image

Issue #1364, opened 2022-08-11
Change sin validations table


abschwenker:
**Additional Context**
- Pulled from #1057
- Need to review the sims.sin_validations table that is relying on user_id to group the related records for a student, as shown below. If a student will be able to be associated with more than one user the below structure must be reviewed.
- This will create a 2 way relationship to specify which record is active and provide the history of the sin validations

**Acceptance Criteria**
- [x] Change the user_id from sin_validations table to use the student_id
- [x] adjust the related methods on the sinValidation service 
- [x] do not create new sin record when a student upgrades or downgrades their account (if upgrading/downgrading, don't create sin validation record)

PR #1400, opened 2022-09-07
[API][SIMS #1364] SIN Validation Changes to use Student Id

dheepak-aot:
## SIN Validation table changes

- [x] Change the user_id from sin_validations table to use the student_id
- [x] adjust the related methods on the sinValidation service 
- [x] do not create new sin record when a student upgrades or downgrades their account (if upgrading/downgrading, don't create sin validation record)

**_Following user upgraded to BCSC from basic BCeID._**

image

**_Following user downgraded to basic BCeID from BCSC._**

image


sources/packages/api/src-sql/SINValidations/Update-column-student-id.sql
```diff
@@ -0,0 +1,20 @@
...
+        SELECT
+            id
+        FROM
+            sims.students
+        WHERE
+            user_id = sv.user_id
+    );
+
+-- Once the student_id column is populated then make the column not null.
+ALTER TABLE
```
andrewsignori-aot, line 15: I believe that the DEV DB will have users that are associated with more than one student, due to the tests with BCeID upgrade/downgrade, which will make some records on the `sims.sin_validations` not have the student_id associated.

sources/packages/api/src-sql/Student/Alter-sin-validation-id.sql
```diff
@@ -0,0 +1,7 @@
+-- Dropping the not null constraint as sims.students and sims.sin_validations are dependant on each other and
```
ann-aot, line 1: 👍

sources/packages/api/src/database/__tests__/student.model.spec.ts
```diff
@@ -39,12 +39,16 @@ describe("Test student model", () => {
     user.firstName = faker.name.firstName();
     user.lastName = faker.name.lastName();
     sub.user = user;
+
+    // Save student
```
ann-aot, line 43: full stop :), line 51 too

sources/packages/api/src/services/sin-validation/sin-validation.service.ts
```diff
@@ -55,13 +57,13 @@ export class SINValidationService extends RecordDataModelService<SINValidation>
         "sinValidation.temporarySIN",
         "sinValidation.sinExpiryDate",
       ])
-      .where("sinValidation.user.id = :userId", { userId })
+      .where("sinValidation.student.id = :studentId", { studentId })
```
andrewsignori-aot, line 60: As possible, we can take these opportunities to convert string-based queries to the new type-safe Typeorm model.

sources/packages/api/src/database/migrations/1662507874777-ChangeSINValidationReferenceToStudent.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class ChangeSINValidationReferenceToStudent1662507874777
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Add-column-student-id.sql", "SINValidations"),
+    );
+
+    await queryRunner.query(
```
ann-aot, line 12: I was thinking can't we merge the migration to a single file?

sources/packages/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
```diff
@@ -68,13 +68,16 @@ describe("Test ATBC Controller", () => {
     simsUser.firstName = faker.name.firstName();
     simsUser.lastName = faker.name.lastName();
     fakeStudent.user = simsUser;
+
+    // Save the student in SIMS
```
ann-aot, line 72: full stop

sources/packages/api/src/services/sin-validation/sin-validation.service.ts
```diff
@@ -98,7 +100,7 @@ export class SINValidationService extends RecordDataModelService<SINValidation>
       newSINValidation.sinEditedNote = savedNote;
       newSINValidation.creator = auditUser;
       newSINValidation.createdAt = now;
-      newSINValidation.user = student.user;
+      newSINValidation.student = student;
```
ann-aot, line 103: ` newSINValidation.student = {id:student.id} as student;`

sources/packages/api/src/database/migrations/1662507874777-ChangeSINValidationReferenceToStudent.ts
dheepak-aot: I wanted to keep the migrations separate for sin-validations and student table changes. Also wanted to reduce the complexity of that migration which already has 3 sql files for up and down each.

sources/packages/api/src/services/sin-validation/sin-validation.service.ts
dheepak-aot: as the student instance was already present, used the same instead of creating a new instance. Also there is no cascade from sinvalidation to student so any other property other than id will be disregarded.

sources/packages/api/src-sql/SINValidations/Add-column-student-id.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    sims.sin_validations
+ADD
+    COLUMN IF NOT EXISTS student_id INT REFERENCES sims.students(id);
+
+COMMENT ON COLUMN sims.sin_validations.student_id IS 'Student to whom the SIN validation record belongs to.';
```
henryoforeh-aot, line 6: Can you please indicate in the comment that this is a foreign key reference to students table just like you have done in the comment for user_id.

andrewsignori-aot (commented):
Pretty nice work, please take a look in a minor comment.

sources/packages/api/src-sql/SINValidations/Update-column-student-id.sql
dheepak-aot: As discussed, using the student_users table to populate the student_id column.

sources/packages/api/src-sql/SINValidations/Add-column-student-id.sql
dheepak-aot: IMO when I mention foreign key in the comment it would be just rephrasing the verbose. My intention here was to provide meta-data/business context to the column. @andrewsignori-aot @ann-aot please let me know your thoughts as well.

sources/packages/api/src/database/migrations/1662507874777-ChangeSINValidationReferenceToStudent.ts
andrewsignori-aot: Just to share some thoughts, I would also prefer fewer SQL migration files but I do not see an issue with the way as it is right now. Just mentioning to ensure that both ways are ok (IMO) and it should not become a pattern to be followed on every PR review 😉

andrewsignori-aot (approved):
Thank for doing the changes, looks good 👍

sources/packages/api/src-sql/SINValidations/Add-column-student-id.sql
henryoforeh-aot: Sure, resolving.

henryoforeh-aot (approved):

Issue #140, opened 2021-02-02
Change common auditing fields use the same date type


fwpushan:
**Task 1**
- [x] Verify and change common auditing fields use the same date type "timestamp with time zone" if they are not
- Create alter table migration for those AUDITING FIELDS ONLY that are not "timestamp with time zone"

PR #1401, opened 2022-09-07
[API][SIMS #140] change common auditing fields

henryoforeh-aot:
- Created migration script to alter all common auditing fields (created_at and updated_at) to use the same date type "timestamp with time zone". 

- The script above applies only to tables where the auditing fields were created using timestamp without time zone.

### DIFF
--- a/sources/packages/api/src-sql/AlterAuditColumns/Add-with-time-zone.sql
+++ b/sources/packages/api/src-sql/AlterAuditColumns/Add-with-time-zone.sql
@@ -1,0 +1,281 @@
+-- Alter created_at column in applications to include timestamp with time zone. --
+ALTER TABLE
+    sims.applications
+ALTER COLUMN
+    created_at TYPE TIMESTAMP WITH TIME ZONE;
+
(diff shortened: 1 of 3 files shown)


sources/packages/api/src-sql/AlterAuditColumns/Add-with-time-zone.sql
```diff
@@ -0,0 +1,281 @@
+-- Alter created_at column in applications to include timestamp with time zone. --
```
dheepak-aot, line 1: Comments doesn't need to end with --

henryoforeh-aot: I will remove them.

andrewsignori-aot (approved):
Nice work, looks good 👍

dheepak-aot (approved):
Nice work @henryoforeh-aot 👍

ann-aot (approved):
Good Work @henryoforeh-aot

Issue #880, opened 2021-11-24
UI Updates for Application Details 


michesmi:
**Additional Context**
- Pages not included: Assessment, Enrolment, Completed 
- This page will be completely vuetify
- Figma: https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=7809%3A108528

**Acceptance Crtieria**
- [x] While in draft, display new card
- [x] While in submitted, display new card
- [x] While in Cancelled, display new card Cancelled, cancelled date comes from application status updated on
- While in progress the following applies
- [x] The gray cards are displayed when something is flagged as required (PIR/Parent information/ see figma for states) 
- [x] The green cards are displayed when a previous required attribute now compelted(PIR/Parent information/ see figma for states) 
- [x] The red cards are displayed when a previous required attribute now declined (PIR see figma for states) 
- [x] No cards are displayed when a process is not required

**Application details**
- [x] Capture submitted date and display in footer
- [x] Populate study dates field in header when a student selects "offering not found" and display the date when an offering is found
- [x] Update tracker UI

**tech note**
- build 1 component to represent each application status based on applicationID

**Acceptance Criteria**
- [ ] Ensure above pages match figma and below criteria are complete
- [ ] Apply common components to all applicable areas
- [ ] Remove remaining Primvue components Change modals to Vuetify (except for file uploaders)
- [ ] Buttons outside of the forms - component already created (ministry create institution for example)
- [ ] Buttons always enabled so validations show up (taken care of when buttons are outside forms)
- [ ] For Vue based forms (eg. search) always enable button and display validations
- [ ] Use full-page container
- [ ] Use the formio submit call instead of check validity method so the error summary shows at the top - follow how student application works
- [ ] Remove from Dev deprecated modals
- [ ] Include image alt-tags

PR #1414, opened 2022-09-19
[API][Web][SIMS-#880] UI Updates for Application Details #880

ann-aot:
- Application details page below status
> Draft
> Submitted
> In Progress
> Cancelled
- Each status is a separate component.
- Created a common component for the alert banner
- Tracker is achebied by `v-slider`
- Migration for the application submitted date
- Added  ` .resetValidation();` which i missed in my previous PR.
- For the status that is in the tracker (` 0: ApplicationStatus.submitted,
      1: ApplicationStatus.inProgress,
      2: ApplicationStatus.assessment,
      3: ApplicationStatus.enrollment,
      4: ApplicationStatus.completed,`)
image
, the check is done like `applicationStatusTracker[trackerApplicationStatus] === ApplicationStatus.submitted`, so that in the future if we need to see the previous, it can be easily attained just by removing the read-only param from the v-slider or by adding a simple logic for read-only.
-Created an event that can be emitted, whenever there is a denial card logic eg, 
image
- Fixed search box alignment issues.
- Some CSS changes and adjustments
- Code smells - TODO
- Code duplication is due to API and Web interfaces
- 
- Screenshots
image
image
image
image


sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -320,6 +320,16 @@ export class Application extends RecordDataModel {
...
+
+  /**
+   * Application submitted date.
+   */
+  @Column({
+    name: "submitted_date",
+    type: "timestamptz",
+    nullable: true,
+  })
+  submittedDate: Date;
```
andrewsignori-aot, line 332: Please change to a nullable property `submittedDate?`.

sources/packages/api/src/database/entities/application.model.ts
```diff
@@ -320,6 +320,16 @@ export class Application extends RecordDataModel {
     referencedColumnName: ColumnNames.ID,
   })
   applicationException?: ApplicationException;
+
```
andrewsignori-aot, line 323: Please keep the same spacing from the rest of the file.

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,78 @@ export class ApplicationControllerService {
...
+    const supportingUserIncomeVerification = CRAIncomeVerification.filter(
+      (incomeVerification) => incomeVerification.supportingUser,
+    );
+    studentIncomeVerification.forEach((incomeVerification) => {
+      // Student income verification details.
+      incomeVerificationDetails.studentIncomeVerificationStatus = {
+        waiting: !incomeVerification.dateReceived,
+        success: !!incomeVerification.dateReceived,
+      };
+    });
```
henryoforeh-aot, line 235: Trying to understand what is happening here. Is the intention to create an array ?

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,78 @@ export class ApplicationControllerService {
       );
     }
   }
+
+  /**
+   * Process application income verification details.
+   * @param CRAIncomeVerification CRA income verification data.
+   * @return processed object.
+   */
+
```
andrewsignori-aot, line 218: Please remove the extra line.

sources/packages/api/src/route-controllers/application/application.controller.service.ts
ann-aot: i will look something like 
image

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,78 @@ export class ApplicationControllerService {
...
+    CRAIncomeVerification: CRAIncomeVerification[],
+  ): ApplicationIncomeVerification {
+    const incomeVerificationDetails = {} as ApplicationIncomeVerification;
+    const studentIncomeVerification = CRAIncomeVerification.filter(
+      (incomeVerification) => !incomeVerification.supportingUser,
+    );
+    const supportingUserIncomeVerification = CRAIncomeVerification.filter(
+      (incomeVerification) => incomeVerification.supportingUser,
+    );
+    studentIncomeVerification.forEach((incomeVerification) => {
```
andrewsignori-aot, line 229: I believe that there is no need here to do a for each when we expected only one record. We can access the first item in the array only.
```ts
const [studentIncomeVerification] = studentIncomeVerifications;
```

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,76 @@ export class ApplicationControllerService {
...
     }
   }
+
+  /**
+   * Process application income verification details.
+   * @param CRAIncomeVerification CRA income verification data.
+   * @return processed object.
+   */
+  processApplicationIncomeVerificationDetails(
+    CRAIncomeVerification: CRAIncomeVerification[],
```
dheepak-aot, line 219: camel case.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -0,0 +1,152 @@
...
+    // trackFillColor will vary when more status are added.
+    const trackFillColor = ref("warning");
+
+    const applicationStatusOrder = (status: ApplicationStatus) => {
+      const [key] =
+        Object.entries(applicationStatusTracker.value).find(
+          ([, value]) => value === status,
+        ) ?? [];
+
+      if (key !== undefined) {
```
henryoforeh-aot, line 104: You can also do this if(key).

ann-aot: there is a key = 0, so, if(0) is false

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,76 @@ export class ApplicationControllerService {
...
+    };
+
+    supportingUserIncomeVerification.forEach((incomeVerification, index) => {
+      // Supporting user income verification details.
+      if (
+        incomeVerification.supportingUser.supportingUserType ===
+        SupportingUserType.Parent
+      ) {
+        // If supporting user type is parent, then there can be 1 or 2 parent.
+        incomeVerificationDetails[
```
andrewsignori-aot, line 242: Because the object is not dynamic, I would try to ensure the type safety and do something like below:
```ts
const supportingUserParents = CRAIncomeVerification.filter(
      (incomeVerification) =>
        incomeVerification.supportingUser?.supportingUserType ===
        SupportingUserType.Parent,
    );
    const [parent1, parent2] = supportingUserParents;
    if (parent1) {
      incomeVerificationDetails.parent1IncomeVerificationStatus = {
        waiting: !parent1.dateReceived,
        success: !!parent1.dateReceived,
      } as SuccessWaitingStatus;
    }

    if (parent2) {
      incomeVerificationDetails.parent2IncomeVerificationStatus = {
        waiting: !parent1.dateReceived,
        success: !!parent1.dateReceived,
      } as SuccessWaitingStatus;
    }
```

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
```diff
@@ -0,0 +1,296 @@
...
+      type: Number,
+      required: true,
+    },
+  },
+  setup(props, { emit }) {
+    const applicationDetails = ref<InProgressApplicationDetailsAPIOutDTO>();
+    const router = useRouter();
+
+    const goToStudentApplication = async () => {
+      router.push({
```
henryoforeh-aot, line 265: await ?

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
+  studentIncomeVerificationStatus: SuccessWaitingStatus;
+}
+
+export interface ApplicationSupportingUserDetails {
+  parent1Info: SuccessWaitingStatus;
+  parent2Info: SuccessWaitingStatus;
+  partnerInfo: SuccessWaitingStatus;
+}
+
+export interface InProgressApplicationDetailsAPIOutDTO
```
andrewsignori-aot, line 128: Please change it to a class to the right swagger schema generation.

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
   active: boolean;
 }
+
+export interface SuccessWaitingStatus {
+  success: boolean;
+  waiting: boolean;
+}
+
+export interface ApplicationIncomeVerification {
+  parent1IncomeVerificationStatus: SuccessWaitingStatus;
```
andrewsignori-aot, line 116: Besides students, are not all other properties nullable?

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
+  parent2IncomeVerificationStatus: SuccessWaitingStatus;
+  partnerIncomeVerificationStatus: SuccessWaitingStatus;
+  studentIncomeVerificationStatus: SuccessWaitingStatus;
+}
+
+export interface ApplicationSupportingUserDetails {
+  parent1Info: SuccessWaitingStatus;
```
andrewsignori-aot, line 123: I believe that all these properties must be defined as nullable, right?

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
+  exceptionStatus: ApplicationExceptionStatus;
+}
+
+export interface CancelledApplicationDetailsAPIOutDTO {
+  statusUpdatedOn: Date;
+}
+
+export interface ApplicationDetailsAPIOutDTO {
+  applicationNumber: string;
+  applicationInstitutionName: string;
```
andrewsignori-aot, line 145: For an object named `ApplicationDetails`, do we need to define all properties with the `application` prefix?

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
+  applicationStartDate: Date | string;
```
andrewsignori-aot, line 146: I would prefer to have it defined as a `Date`. When needed, convert the string to a Date. The API must ensure that the proper object is returned, independently from specific typed DB column of JSONB.

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
+  PIRDeniedReason: string;
+  offeringStatus: OfferingStatus;
+  exceptionStatus: ApplicationExceptionStatus;
+}
+
+export interface CancelledApplicationDetailsAPIOutDTO {
+  statusUpdatedOn: Date;
+}
+
+export interface ApplicationDetailsAPIOutDTO {
```
andrewsignori-aot, line 143: Please covert it to a class and check other DTOs also.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
...
+   * @returns application details.
+   */
+  @Get(":id/in-progress")
+  @ApiNotFoundResponse({
+    description: "Application id not found.",
+  })
+  async getInProgressApplicationDetails(
+    @Param("id", ParseIntPipe) applicationId: number,
+  ): Promise<InProgressApplicationDetailsAPIOutDTO> {
+    const application = await this.applicationService.getApplicationByIdAndUser(
```
andrewsignori-aot, line 472: We must check if the student has access to the application and we should not rely on the userId since the same student can have multiple ones.

sources/packages/web/src/types/contracts/student/ApplicationContract.ts
```diff
@@ -209,3 +216,46 @@ export interface PartTimeAssessment extends BaseAssessment {
...
+  waiting: boolean;
+}
+
+export interface InProgressApplicationDetailsAPIOutDTO {
+  id: number;
+  applicationStatus: ApplicationStatus;
+  pirStatus: ProgramInfoStatus;
+  PIRDeniedReason: string;
+  offeringStatus: OfferingStatus;
+  exceptionStatus: ApplicationExceptionStatus;
```
andrewsignori-aot, line 239: Please review the possible nullable properties in this object. probably all but `id`, `applicationStatus` and `pirStatus` must be nullable.

sources/packages/web/src/types/contracts/student/ApplicationContract.ts
```diff
@@ -209,3 +216,46 @@ export interface PartTimeAssessment extends BaseAssessment {
...
+  applicationStartDate: Date | string;
+  applicationEndDate: Date | string;
+}
+
+export interface SuccessWaitingStatus {
+  success: boolean;
+  waiting: boolean;
+}
+
+export interface InProgressApplicationDetailsAPIOutDTO {
```
andrewsignori-aot, line 233: Please move DTOs to the DTO files.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -0,0 +1,152 @@
...
+      </template>
+    </v-slider>
+
+    <draft
+      @editApplication="$emit('editApplication')"
+      v-if="applicationStatus === ApplicationStatus.draft"
+    />
+    <!-- The below components are checked with applicationStatusTracker[trackerApplicationStatus], so that in future if we need to see the previous, it can be easily attained just by removing readonly param from the v-slider or by adding a simple logic. -->
+    <submitted
+      v-if="
```
andrewsignori-aot, line 42: This is a scenario to apply a v-else-if, IMO.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
...
+      await this.craIncomeVerificationService.allIncomeVerificationsForAnApplication(
+        applicationId,
+      );
+    const incomeVerification =
+      this.applicationControllerService.processApplicationIncomeVerificationDetails(
+        incomeVerificationDetails,
+      );
+
+    const supportingUserDetails =
+      await this.supportingUserService.getSupportingUsersByApplicationId(
```
andrewsignori-aot, line 491: Please keep it consistent between `process` or `get` as prefixes.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
...
+   * @returns application details.
+   */
+  @Get(":id/cancelled")
+  @ApiNotFoundResponse({
+    description: "Application id not found.",
+  })
+  async getCancelledApplicationDetails(
+    @Param("id", ParseIntPipe) applicationId: number,
+  ): Promise<CancelledApplicationDetailsAPIOutDTO> {
+    const application = await this.applicationService.getApplicationByIdAndUser(
```
andrewsignori-aot, line 524: We must check if the student has access to the application and we should not rely on the userId since the same student can have multiple ones.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
...
+  /**
+   * Get cancelled details of an application by application id.
+   * @param applicationId application id.
+   * @returns application details.
+   */
+  @Get(":id/cancelled")
+  @ApiNotFoundResponse({
+    description: "Application id not found.",
+  })
+  async getCancelledApplicationDetails(
```
andrewsignori-aot, line 521: Do we need a new method to get just this info?

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
...
+  /**
+   * Get details of an application by application id.
+   * @param applicationId application id.
+   * @returns application details.
+   */
+  @Get(":id/details")
+  @ApiNotFoundResponse({
+    description: "Application id not found.",
+  })
+  async getApplicationStatusDetails(
```
andrewsignori-aot, line 546: Do we need a new method to get this info? Can we use the `getByApplicationId`?

sources/packages/api/src/utilities/date-utils.ts
```diff
@@ -159,7 +159,9 @@ export function getExtendedDateFormat(date: Date): string {
  * @returns extended date format like Mar 22 2021
  */
 export function getDateOnlyFormat(date?: string | Date): string {
-  return dayjs(date).format(DATE_ONLY_FORMAT);
+  if (date) {
```
andrewsignori-aot, line 162: It would be good to always have something to be returned.

sources/packages/web/src/components/institutions/modals/confirmationOfEnrollment/ApproveCOE.vue
```diff
@@ -54,7 +54,6 @@
           primaryLabel="Continue to confirmation"
           @primaryClick="submit"
           @secondaryClick="cancel"
-          :disablePrimaryButton="notAllowed"
```
andrewsignori-aot, line 57: Can you please elaborate on why was this removed?

sources/packages/web/src/components/institutions/modals/confirmationOfEnrollment/DenyCOE.vue
```diff
@@ -54,7 +54,6 @@
           primaryLabel="Decline enrolment now"
           @primaryClick="submit"
           @secondaryClick="cancel"
-          :disablePrimaryButton="notAllowed"
```
andrewsignori-aot, line 57: Can you please elaborate on why was this removed?

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -0,0 +1,152 @@
+<template>
+  <template v-if="applicationStatus !== ApplicationStatus.cancelled">
+    <body-header title="Track your application" />
+    <v-slider
```
andrewsignori-aot, line 4: I would preferer to have only the `v-slider` in the `ApplicationProgressBar` component.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -0,0 +1,152 @@
...
+            icon="fa:fas fa-exclamation-circle"
+            :size="20"
+            color="danger"
+            class="pl-4"
+        /></span>
+        <span class="label-value default-color" v-else>{{ tick.label }} </span>
+      </template>
+    </v-slider>
+
+    <draft
```
andrewsignori-aot, line 36: I would prefer that the show/hide of all these components would be based on a simple status of the `ApplicationStatus` in a way that is is not dependent on `applicationStatusTracker[trackerApplicationStatus]`.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -0,0 +1,152 @@
...
+      required: true,
+    },
+    applicationStatus: {
+      type: Object as PropType<ApplicationStatus>,
+      required: true,
+    },
+  },
+  setup(props) {
+    const hasDeclinedCard = ref(false);
+    const applicationStatusTracker = ref<Record<number, ApplicationStatus>>({
```
andrewsignori-aot, line 88: This could be achieved with a simple enum. It would be clear, IMO.

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,76 @@ export class ApplicationControllerService {
...
+  /**
+   * Process application income verification details.
+   * @param CRAIncomeVerification CRA income verification data.
+   * @return processed object.
+   */
+  processApplicationIncomeVerificationDetails(
+    CRAIncomeVerification: CRAIncomeVerification[],
+  ): ApplicationIncomeVerification {
+    const incomeVerificationDetails = {} as ApplicationIncomeVerification;
+    const studentIncomeVerifications = CRAIncomeVerification.filter(
```
dheepak-aot, line 222: studentIncomeVerification can be assigned directly like this.

const [studentIncomeVerification] = CRAIncomeVerification.filter(
      (incomeVerification) => !incomeVerification.supportingUser,
    );

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -0,0 +1,152 @@
...
+      "
+    />
+    <in-progress
+      v-if="
+        trackerApplicationStatus !== undefined &&
+        applicationStatusTracker[trackerApplicationStatus] ===
+          ApplicationStatus.inProgress
+      "
+      :applicationId="applicationId"
+      @declinedEvent="declinedEvent"
```
andrewsignori-aot, line 55: I am struggling to follow the idea behind this event, It looks like more as property than an event. Can we have a sync up about it?

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,76 @@ export class ApplicationControllerService {
...
+      (incomeVerification) => !incomeVerification.supportingUser,
+    );
+    const supportingUserIncomeVerification = CRAIncomeVerification.filter(
+      (incomeVerification) => incomeVerification.supportingUser,
+    );
+    const [studentIncomeVerification] = studentIncomeVerifications;
+    // Student income verification details.
+    incomeVerificationDetails.studentIncomeVerificationStatus = {
+      waiting: !studentIncomeVerification.dateReceived,
```
dheepak-aot, line 231: I did not get this. for status why are we using waiting and success as key with value? 
```
Why is it not like this incomeVerificationDetails.studentIncomeVerificationStatus = !!studentIncomeVerification.dateReceived ?  SomeEnum.success : SomeEnum.waiting
```

sources/packages/web/src/components/students/applicationTracker/Cancelled.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <application-status-tracker-banner
+    label="You've cancelled your application"
+    icon="fa:fas fa-exclamation-circle"
+    icon-color="danger"
+    background-color="error_bg"
```
andrewsignori-aot, line 6: Why did we go for the snake pattern on these ones?

sources/packages/web/src/components/students/applicationTracker/Cancelled.vue
```diff
@@ -0,0 +1,50 @@
...
+    ><template #content
+      ><span
```
andrewsignori-aot, line 8: Is the span needed here?

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
```diff
@@ -0,0 +1,296 @@
...
+    icon-color="secondary"
+    content="StudentAid BC is currently reviewing your application and the 
+    additional information you provided for 1 or more of the questions."
+    v-if="
+      applicationDetails?.exceptionStatus === ApplicationExceptionStatus.Pending
+    "
+  />
+
+  <!-- Success cards. -->
+  <application-status-tracker-banner
```
andrewsignori-aot, line 111: We can have a component that will hold the `waiting` and `success` card states and decide which one to show based on a condition. `icon-color` and `icon` would be defined internally by the component and the success/wait/error could be grouped.

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +209,76 @@ export class ApplicationControllerService {
...
+      success: !!studentIncomeVerification.dateReceived,
```
dheepak-aot, line 232: Also as a status which is derived from if dateReceived is present, completed is more suitable than success IMO.

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
```diff
@@ -0,0 +1,296 @@
...
+        name: StudentRoutesConst.STUDENT_APPLICATION_FORM,
+      });
+    };
+
+    onMounted(async () => {
+      applicationDetails.value =
+        await ApplicationService.shared.getInProgressApplicationDetails(
+          props.applicationId,
+        );
+      // Any declined cards will call declined event.
```
andrewsignori-aot, line 275: It does not seem to be something to be handled with an event. We probably need a sync-up in order to me better understand the reasons why it was necessary.

sources/packages/web/src/components/students/applicationTracker/generic/ApplicationStatusTrackerBanner.vue
```diff
@@ -0,0 +1,74 @@
...
+    variant="tonal"
+    class="application-tracker-alert mt-3"
+    :class="backgroundColor"
+    :color="backgroundColor"
+  >
+    <template #prepend>
+      <v-icon class="mt-1" :icon="icon" :size="20" :color="iconColor" />
+    </template>
+    <template #title>
+      <div class="category-header-medium mb-3 label">
```
andrewsignori-aot, line 12: Can we change this to a `<p>` or some `<h?>`?

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
...
+      applicationId,
+    );
+    if (!application) {
+      throw new NotFoundException(
+        `Application id ${applicationId} was not found.`,
+      );
+    }
+
+    const incomeVerificationDetails =
+      await this.craIncomeVerificationService.allIncomeVerificationsForAnApplication(
```
dheepak-aot, line 482: to keep the naming consistent, the method name can be getAllIncomeVerifications

sources/packages/web/src/components/students/applicationTracker/generic/ApplicationStatusTrackerBanner.vue
```diff
@@ -0,0 +1,74 @@
...
+      default: "white",
+    },
+    iconColor: {
+      type: String,
+      required: false,
+      default: "default",
+    },
+  },
+  setup(_props, { slots }) {
+    const hasImage = ref(false);
```
andrewsignori-aot, line 62: I believe that it can be converted to a computed property like below.
```ts
const hasImage = computed(() => !!slots.image);
````

File: sources/packages/api/src/route-controllers/application/application.students.controller.ts (hunk 16 above)
Comment (dheepak-aot) [reply sources/packages/api/src/route-controllers/application/application.students.controller.ts:472]: Where are we using the userId here @andrewsignori-aot ?

File: sources/packages/web/src/components/students/applicationTracker/generic/ApplicationStatusTrackerBanner.vue [hunk 41]
```diff
@@ -0,0 +1,74 @@
...
+    // Check if the slot exists by name and has content.
+    // It returns an empty array if it's empty.
+    if (slots.image && slots.image().length) {
+      hasImage.value = true;
+    }
+
+    return {
+      hasImage,
```
andrewsignori-aot, line 70: Maybe it can also be done from vue template as `v-if="$slots.image"`.

sources/packages/web/src/composables/useApplication.ts
```diff
@@ -1,5 +1,10 @@
 import { StatusChipTypes } from "@/components/generic/StatusChip.models";
-import { ApplicationStatus } from "@/types";
+import {
+  ApplicationDetailHeader,
+  ApplicationStatus,
+  OfferingIntensity,
+} from "@/types";
+import { useFormatters } from "./useFormatters";
```
andrewsignori-aot, line 7: Please use @.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -448,4 +456,120 @@ export class ApplicationStudentsController extends BaseController {
       applicationNumber: application.applicationNumber,
     };
   }
+
+  /**
+   * Get in progress details of an application by application id.
+   * @param applicationId application id.
+   * @returns application details.
+   */
+  @Get(":id/in-progress")
```
dheepak-aot, line 465: Minor one. we can use  applicationId here. 

`@Get(":applicationId/in-progress")`

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -165,4 +170,46 @@ export class ApplicationApi extends HttpBaseClient {
...
+
+  /**
+   * Get in progress details of an application by application id.
+   * @param applicationId application id.
+   * @returns application details.
+   */
+  async getInProgressApplicationDetails(
+    applicationId: number,
+  ): Promise<InProgressApplicationDetailsAPIOutDTO> {
+    const response = await this.getCall(
```
andrewsignori-aot, line 182: Please use `getCallTyped`, same for the other ones.

andrewsignori-aot (commented):
Nice work, the application details is looking great. Please take a look at the comments.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
dheepak-aot: Wanted to ask this. Currently we do not have a get API for applications on student side which returns a summary of application details(excluding the json data in it). It would be nice to build one isn't it?

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
   programYearId: number;
   active: boolean;
 }
+
+export interface SuccessWaitingStatus {
+  success: boolean;
+  waiting: boolean;
+}
+
+export interface ApplicationIncomeVerification {
```
dheepak-aot, line 115: Besides these ones must also be class to be visible in swagger and also all the classes must be present in a file suffixed with .dto.ts for swagger to recognize them as dto.

sources/packages/api/src/route-controllers/application/models/application.model.ts
```diff
@@ -103,3 +106,46 @@ export interface ApplicationWithProgramYearDto {
...
+  partnerInfo: SuccessWaitingStatus;
+}
+
+export interface InProgressApplicationDetailsAPIOutDTO
+  extends ApplicationIncomeVerification,
+    ApplicationSupportingUserDetails {
+  id: number;
+  applicationStatus: ApplicationStatus;
+  pirStatus: ProgramInfoStatus;
+  PIRDeniedReason: string;
```
dheepak-aot, line 134: camel case

sources/packages/api/src/services/cra-income-verification/cra-income-verification.service.ts
```diff
@@ -249,4 +249,32 @@ export class CRAIncomeVerificationService extends RecordDataModelService<CRAInco
...
     }
   }
+
+  /**
+   * Get the student and supporting users (if any) income verification details for an applications.
+   * @param applicationId application id.
+   * @returns income verification details for an applications.
+   */
+  async allIncomeVerificationsForAnApplication(
+    applicationId: number,
```
dheepak-aot, line 259: Looking at it as an independent service, should we consider including studentId in the where?

sources/packages/web/src/assets/css/base.scss
```diff
@@ -269,19 +269,65 @@ html * {
...
+}
+
+.application-tracker-alert.v-alert:not(.white) .v-alert__underlay {
+  opacity: unset;
+  z-index: -1;
+}
+
+// col img css.
+.col-img {
+  margin-top: -3%;
```
dheepak-aot, line 321: why are we having the margin in % here?

sources/packages/web/src/assets/css/base.scss
```diff
@@ -269,19 +269,65 @@ html * {
...
+  margin-right: -21%;
+}
+
+.application-tracker-alert.v-alert {
+  /* UI/Borders */
+  border: 1px solid $common-border-color;
```
dheepak-aot, line 327: we can use the single line comment.

sources/packages/web/src/components/common/OfferingSummary.vue
```diff
@@ -6,7 +6,6 @@
     <template #actions>
       <v-row class="m-0 p-0">
         <v-text-field
-          v-if="!!offeringsAndCount.count"
```
dheepak-aot, line 9: why did we remove the v-if here? is that a requirement to have search text irrespective of data returned?

sources/packages/web/src/components/institutions/active-application/ActiveApplicationSummaryData.vue
```diff
@@ -7,7 +7,6 @@
       >
         <template #actions>
           <v-text-field
-            v-if="!!applications.results?.length"
```
dheepak-aot, line 10: same question here.

sources/packages/web/src/components/institutions/modals/confirmationOfEnrollment/ApproveCOE.vue
dheepak-aot: I believe that assignment of this attribute is a dead code as notAllowed is not a property that is defined in the component.

sources/packages/web/src/components/students/applicationTracker/Submitted.vue
```diff
@@ -0,0 +1,23 @@
...
+  <application-status-tracker-banner
+    label="Your application has been submitted!"
+    icon="fa:fas fa-check-circle"
+    icon-color="success"
+    content="Keep up-to-date with your application by tracking it here."
+  >
+    <template #image>
+      <v-img
+        height="60"
+        alt="You don't have any agreements yet."
```
dheepak-aot, line 11: alt value doesn't seem to be right.

sources/packages/web/src/components/students/applicationTracker/Cancelled.vue
```diff
@@ -0,0 +1,50 @@
...
+        >You cancelled your application on
+        {{ dateOnlyLongString(applicationDetails?.statusUpdatedOn) }}. However
```
dheepak-aot, line 10: We can get rid of this null check by initializing the the ref like this. 
const applicationDetails = ref<CancelledApplicationDetailsAPIOutDTO>({});

sources/packages/web/src/components/students/applicationTracker/Draft.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <application-status-tracker-banner
+    label="Draft application"
+    content="Your application has not been submitted to StudentAid BC yet. Please complete and submit your application when you are ready."
+    icon="fa:fas fa-pencil-alt"
+    ><template #actions>
```
dheepak-aot, line 6: The button action can be implemented in Draft.vue instead of emitting IMO.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
dheepak-aot, line 88: I believe the logic of this method is to return the tracker value based on applicationStatus. 

if so the key and value of how this object is built can be swapped to make this logic simple.

```
const applicationStatusTracker = ref<Record<ApplicationStatus,number>>({
      Submitted: 0,
      "In Progress": 1,
      Assessment: 2,
      Enrollment: 3,
      Completed: 4,
    });
const trackerApplicationStatus = computed(() =>
      applicationStatusTracker[props.applicationStatus] 
    );
```

sources/packages/web/src/services/http/ApplicationApi.ts
dheepak-aot, line 182: Why are we assigning it to response? generally for get call we just return this.getCall() right?

Also we can use getcallTyped.

Same for other get calls

sources/packages/web/src/services/ApplicationService.ts
```diff
@@ -127,4 +130,37 @@ export class ApplicationService {
...
+  ): Promise<CancelledApplicationDetailsAPIOutDTO> {
+    return ApiClient.Application.getCancelledApplicationDetails(applicationId);
+  }
+
+  /**
+   * Get details of an application by application id.
+   * @param applicationId application id.
+   * @returns application details.
+   */
+  public async getApplicationStatusDetails(
```
dheepak-aot, line 161: we can remove public from all these methods.

sources/packages/web/src/composables/useApplication.ts
```diff
@@ -22,5 +27,23 @@ export function useApplication() {
...
+
+    return {
+      "Application #": application.applicationNumber,
+      Institution: application.applicationInstitutionName,
+      "Study dates":
+        application.applicationStartDate && application.applicationEndDate
+          ? `${dateOnlyLongString(
+              application.applicationStartDate,
+            )} - ${dateOnlyLongString(application.applicationEndDate)}`
+          : "-",
```
dheepak-aot, line 43: Is that a requirement that when there is no data for the header component we must show - or instead not populate the keyValue itself?

sources/packages/web/src/components/students/applicationTracker/generic/ApplicationStatusTrackerBanner.vue
```diff
@@ -0,0 +1,74 @@
...
+    </template>
+    <template #title>
+      <div class="category-header-medium mb-3 label">
+        {{ label }}
+      </div>
+    </template>
+    <span class="label-value-normal">
+      <slot name="content">
+        <v-row>
+          <v-col :md="hasImage ? 10 : 12">
```
dheepak-aot, line 19: Just for clarification, without setting md will it not adjust based on the next column if it is available.?

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
```diff
@@ -0,0 +1,296 @@
...
+  >
+  <application-status-tracker-banner
+    v-if="
+      applicationDetails?.exceptionStatus ===
+      ApplicationExceptionStatus.Declined
+    "
+    label="StudentAid BC has not accepted your application"
+    icon="fa:fas fa-exclamation-circle"
+    icon-color="danger"
+    content="StudentAid BC has not accepted 1 or more of the additional information that you provided in your
```
dheepak-aot, line 230: Such a long content can be returned though variable from script.

dheepak-aot:
Nice work @ann-aot. Please take a look at the comments.

hellolynn-tbtb (approved):
Thanks for the work @ann-aot ! UI looks great. In the header, just missing a dash for offering intensity (e.g. Part-time, Full-time) as per Figma. Is it easy to add now? It's okay to make this change in future tickets too since we will come back to this area.

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
ann-aot: removing async

sources/packages/api/src/route-controllers/application/models/application.model.ts
ann-aot: PIR waiting case, dates are taken from the data filed, where it is in string format, 
image

ann-aot: did you mean to convert the string to the Date format and send it?

sources/packages/api/src/route-controllers/application/models/application.model.ts
ann-aot: now I realize how much I was working on UI tickets for past tickets 😉

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
henryoforeh-aot: Sure, got it. You are correct. Thanks for the clarification

sources/packages/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: @ann-aot please note that there is an error in the above sample where the paren1 was consumed in the parent2 logic.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: @dheepak-aot we are not using `userId` here and this is the first problem, there is no check if the student has access to the application.
@ann-aot this PR comment can be fixed by adding the userId which will ensure the authorization check in the same way as the rest of the controller.
Ideally, we will have a tech-debit to change the authorization check to be based on the student.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: Please referer to comment https://github.com/bcgov/SIMS/pull/1414#discussion_r982751253

sources/packages/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: @dheepak-aot I would be ok with creating one like what you said or maybe just providing a query string to the existing one to avoid retrieving all the dynamic data. Either way, I would just avoid creating too specific a method unless there is a performance constraint.

andrewsignori-aot:
> Thanks for the work @ann-aot ! UI looks great. In the header, just missing a dash for offering intensity (e.g. Part-time, Full-time) as per Figma. Is it easy to add now? It's okay to make this change in future tickets too since we will come back to this area.

@hellolynn-tbtb it can be done in individual places quickly but it will need to be done everywhere. Since apparently, we have the final decision on the correct wording I would prefer to change it globally. @abschwenker I would suggest a ticket to us to change the DB enum that will reflect everywhere. We need also to look for possible _form.io_ definitions and other possible impacts, for instance, the offering bulk upload.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
ann-aot: so, I used `get` to get data from DB and `process` to transform the data to a dto

sources/packages/api/src/route-controllers/application/application.students.controller.ts
ann-aot: as @andrewsignori-aot  said we have `getByApplicationId` for students. I was planning to use it first. then I noticed it is returning a lot of details including COE details, which are not needed for a canceled application. 
I will add this `application.applicationStatusUpdatedOn`  to `getApplicationStatusDetails`, which is consumed by the parent of the component and let the parent pass this details to the child

abschwenker:
> > Thanks for the work @ann-aot ! UI looks great. In the header, just missing a dash for offering intensity (e.g. Part-time, Full-time) as per Figma. Is it easy to add now? It's okay to make this change in future tickets too since we will come back to this area.
> 
> @hellolynn-tbtb it can be done in individual places quickly but it will need to be done everywhere. Since apparently, we have the final decision on the correct wording I would prefer to change it globally. @abschwenker I would suggest a ticket to us to change the DB enum that will reflect everywhere. We need also to look for possible _form.io_ definitions and other possible impacts, for instance, the offering bulk upload. The below enum will also need to be adjusted. image

No need to change anything part of this pr/story. If a change it required we will create a separate task.

sources/packages/api/src/route-controllers/application/application.students.controller.ts
ann-aot: `getByApplicationId` It is not take care of PIR waiting scenario (when some properties get their value from application form data), 
image

sources/packages/web/src/components/institutions/modals/confirmationOfEnrollment/ApproveCOE.vue
ann-aot: it was a spillover. this is an institution component

sources/packages/web/src/components/institutions/modals/confirmationOfEnrollment/DenyCOE.vue
ann-aot: same here

sources/packages/web/src/components/students/applicationTracker/Cancelled.vue
ann-aot: mm, span is generally used as a content container. that's  the reason i always use span for contents

sources/packages/web/src/components/students/applicationTracker/generic/ApplicationStatusTrackerBanner.vue
ann-aot: this logic is used multiple tags, that's why i added in setup, 
image

sources/packages/web/src/assets/css/base.scss
ann-aot: spill over

sources/packages/web/src/components/common/OfferingSummary.vue
ann-aot: we have an alignment issue in some data tables (empty table). this condition is actually not needed in the search box.

sources/packages/web/src/components/students/applicationTracker/Draft.vue
ann-aot: edit program route needs properties like 
image
.and the function is used used in the parent (already an API is called in the parent to revive these details). emit is used by a child to communicate with their parent in vue.

sources/packages/web/src/composables/useApplication.ts
ann-aot: ` instead not populate the keyValue itself` i didn't understand. what do you mean?

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
ann-aot: mm, why. Vue is an HTML-based template, and it's pretty straight forward right? what is the befit of doing so?

henryoforeh-aot (approved):

sources/packages/web/src/composables/useApplication.ts
ann-aot: checked with @hellolynn-tbtb, she asked to keep "-" for now

sources/packages/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -203,4 +211,98 @@ export class ApplicationControllerService {
...
+      (incomeVerification) =>
+        incomeVerification.supportingUserType === SupportingUserType.Parent,
+    );
+    if (parent1) {
+      supportingUserDetails.parent1Info = !parent1.supportingData
+        ? SuccessWaitingStatus.Waiting
+        : SuccessWaitingStatus.Success;
+    }
+    if (parent2) {
+      supportingUserDetails.parent1Info = !parent2.supportingData
```
andrewsignori-aot, line 290: Is it supposed to be `supportingUserDetails.parent2Info`, right?

sources/packages/web/src/components/students/applicationTracker/generic/ApplicationStatusTrackerBanner.vue
```diff
@@ -0,0 +1,62 @@
...
+        {{ label }}
+      </h3>
+    </template>
+    <span class="label-value-normal">
+      <slot name="content">
+        <v-row>
+          <v-col md="auto">
+            <span>{{ content }}</span>
+          </v-col>
+          <v-col offset-md="5" class="mt-n8">
```
dheepak-aot, line 22: Why do we have a negative margin here?

sources/packages/web/src/components/students/applicationTracker/InProgress.vue
andrewsignori-aot: As we talked, the order in which the components are displayed would be better following the states (error, waiting, success). As also talked with @hellolynn-tbtb we will be keeping the component as it is right now.

sources/packages/api/src/route-controllers/application/application.controller.service.ts
ann-aot: yes, thanks for that

andrewsignori-aot (approved):
Great work @ann-aot, looks good 👍

dheepak-aot (approved):
Thanks for making the changes @ann-aot 👍

guru-aot (approved):
LGTM, nice work @ann-aot . Thanks for doing the changes.

Issue #1248, opened 2022-05-30
Ensure API client name convention


andrewsignori-aot:
Acceptance Criteria
- [ ] Record standards in WIKI
- Check and change if required, the API naming convention for API controllers and client names:
- [ ] Students (student seems to be incorrectly named as students/students)
- [ ] Institutions
- [ ] SupportingUsers
- [ ] SystemActions

PR #1429, opened 2022-09-29
[API][SIMS-#1248] API Client Naming

henryoforeh-aot:
- Changed API sub resource(s) from plural to singular.
- Updated affected files due to updated subresources in vue web application.
- Updated wiki documentation.

**Wiki Update**
https://github.com/bcgov/SIMS/wiki/Development-Standards-and-Decisions#controller-api-naming-convention
image

### DIFF
--- a/sources/packages/web/src/services/http/StudentApi.ts
+++ b/sources/packages/web/src/services/http/StudentApi.ts
@@ -24,7 +24,7 @@ export class StudentApi extends HttpBaseClient {
    * @param studentProfile information needed to create the user.
    */
   async createStudent(studentProfile: CreateStudentAPIInDTO): Promise<void> {
-    await this.postCall(this.addClientRoot("students"), studentProfile);
+    await this.postCall(this.addClientRoot("student"), studentProfile);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


ann-aot (approved):
👍

sources/packages/web/src/services/http/StudentApi.ts
```diff
@@ -24,7 +24,7 @@ export class StudentApi extends HttpBaseClient {
    * @param studentProfile information needed to create the user.
    */
   async createStudent(studentProfile: CreateStudentAPIInDTO): Promise<void> {
-    await this.postCall(this.addClientRoot("students"), studentProfile);
+    await this.postCall(this.addClientRoot("student"), studentProfile);
```
andrewsignori-aot, line 27: Just to every body be on the same page, this was what @dheepak-aot mention about having the const.
We would be creating a const like below and reuing among all app calls. Dheepak, correct me if I am wrong.
```ts
const API_CONTROLLER =  "student";
```

```ts
await this.postCall(this.addClientRoot(API_CONTROLLER), studentProfile);
return this.postCall(this.addClientRoot(`${API_CONTROLLER}/search`), payload);
````

andrewsignori-aot (commented):
The changes look good 👍 
I believe that we are missing also the `aest/reports`, right?

sources/packages/web/src/services/http/StudentApi.ts
henryoforeh-aot: maybe I can just make the change in this PR so that we don't forget about it.

guru-aot (approved):
LGTM

andrewsignori-aot (approved):
Looks good, thanks for doing this changes also 👍

Issue #1397, opened 2022-09-01
Camunda 8 analysis


abschwenker:
Additional Context:
- This analysis is focused on how our application stack will work with Camunda 8 and worker processes
- previous POC branch camunda-8-poc
- https://docs.nestjs.com/standalone-applications
- https://docs.nestjs.com/cli/monorepo#workspaces

Acceptance Criteria
- [ ] Determine the pathway for development based on 1 of the below options.
- [ ] Provide an analysis of each option with pro's con's
- [ ] Analysis to include: Estimated work effort, level of segregation to achieve, code splitting, deployment practices, maintenance, knowledge onboarding, etc

**Option 1**
- Single API with multiple NEST applications (stand-alone application dedicated for workers).
- Investigate separate pods (eg. api-start, workers-start, etc)

**Option 2**
- Segregated API for Camunda workers (new package, segregate db, etc).

PR #1444, opened 2022-10-05
[SIMS #1397]Fix workers deploy

andrewsignori-aot:


andrewsignori-aot:
Auto approving as talked with @abschwenker.

Issue #1442, opened 2022-10-05
Camunda 8 assessment gateway initial workflow


abschwenker:
Additional Context
- Continuation of #1367

Acceptance Criteria
- [ ] Convert assessment-gateway to a Camunda 8 workflow using workers and not api calls from Camunda to SIMS-API
- [ ] Create output technical debt and user story tickets are required for remaining work

PR #1485, opened 2022-10-25
[CI/CD][SIMS #1442]Call reusable workflows

andrewsignori-aot:


andrewsignori-aot:
Auto-approving as talked with @abschwenker.

PR #1492, opened 2022-10-26
[CI/CD][SIMS #1442]Rename SIMS-Creds to be generic

andrewsignori-aot:
- Renamed `sims-api-creds` shared between `API` and `Workers` to the more generic name `sims-creds`.
- Enabled the `sims-creds` to be deployed to the same environment multiple times using different names, for instance, `sims-qa-creds` to allow the use of the secrets by the QA sub-environment. This is needed to allow the different secret configurations, for instance, the Zeebe configurations.


andrewsignori-aot:

PR #1497, opened 2022-10-26
[Web][SIMS-#1110] UI updates for Supporting-Users - Search refactoring

andrepestana-aot:
- used common rules for validations; 
- changed supportingUserApi to use postCall and handleAPICustomError;
- removed snackBar for field validation; 
- refactored catches for request reponses;

image
image

### DIFF
--- a/sources/packages/web/src/services/http/SupportingUserApi.ts
+++ b/sources/packages/web/src/services/http/SupportingUserApi.ts
@@ -13,16 +13,15 @@ export class SupportingUserApi extends HttpBaseClient {
     supportingUserType: SupportingUserType,
     payload: ApplicationIdentifierApiInDTO,
   ): Promise<ApplicationApiOutDTO> {
-    let response;
     try {
-      response = await this.postCall(
+      return await this.postCall(
         this.addClientRoot(`supporting-user/${supportingUserType}/application`),
         payload,
       );
     } catch (error: unknown) {
       this.handleAPICustomError(error);
+      throw error;
     }
-    return response;
   }
 
   public async updateSupportingInformation(
--- a/sources/packages/web/src/services/http/SupportingUserApi.ts
+++ b/sources/packages/web/src/services/http/SupportingUserApi.ts
@@ -13,21 +13,16 @@ export class SupportingUserApi extends HttpBaseClient {
     supportingUserType: SupportingUserType,
     payload: ApplicationIdentifierApiInDTO,
   ): Promise<ApplicationApiOutDTO> {
+    let response;
     try {
-      const response = await this.apiClient.post(
+      response = await this.postCall(
         this.addClientRoot(`supporting-user/${supportingUserType}/application`),
         payload,
-        this.addAuthHeader(),
       );
-      return response.data;
-    } catch (error) {
-      if (!error.response.data?.errorType) {
-        // If it is an not expected error,
-        // handle it the default way.
-        this.handleRequestError(error);
-      }
-      throw error;
+    } catch (error: unknown) {
+      this.handleAPICustomError(error);
     }
+    return response;
   }
 
   public async updateSupportingInformation(
--- a/sources/packages/web/src/views/supporting-user/SupportingInformation.vue
+++ b/sources/packages/web/src/views/supporting-user/SupportingInformation.vue
@@ -104,6 +104,7 @@ import {
   useFormatters,
   useSnackBar,
   useFormioUtils,
+  useRules,
 } from "@/composables";
 import { SupportingUsersService } from "@/services/SupportingUserService";
 import { SupportingUserRoutesConst } from "@/constants/routes/RouteConstants";
@@ -119,7 +120,6 @@ import {
   VForm,
   ApiProcessError,
 } from "@/types";
-import { useRules } from "@/composables";
 
 export default defineComponent({
   props: {
--- a/sources/packages/web/src/views/supporting-user/SupportingInformation.vue
+++ b/sources/packages/web/src/views/supporting-user/SupportingInformation.vue
@@ -14,13 +14,13 @@
           <v-col>
             <v-text-field
               density="compact"
-              :label="applicationNumberLabel"
+              label="Application number"
               variant="outlined"
               v-model="applicationNumber"
               data-cy="applicationNumber"
               :rules="[
-                (v) => checkNullOrEmptyRule(v, applicationNumberLabel),
-                (v) => checkOnlyDigitsRule(v, applicationNumberLabel),
+                (v) => checkNullOrEmptyRule(v, 'Number'),
+                (v) => checkOnlyDigitsRule(v, 'Number'),
               ]"
               hide-details="auto"
             />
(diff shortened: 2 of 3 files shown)


sources/packages/web/src/services/http/SupportingUserApi.ts
```diff
@@ -13,21 +13,16 @@ export class SupportingUserApi extends HttpBaseClient {
...
   ): Promise<ApplicationApiOutDTO> {
+    let response;
     try {
-      const response = await this.apiClient.post(
+      response = await this.postCall(
         this.addClientRoot(`supporting-user/${supportingUserType}/application`),
         payload,
-        this.addAuthHeader(),
       );
-      return response.data;
```
ann-aot, line 22: mm...what was the issue here?

sources/packages/web/src/views/supporting-user/SupportingInformation.vue
```diff
@@ -8,47 +8,53 @@
...
+    <v-form ref="searchApplicationsForm">
+      <content-group class="mb-4">
+        <v-row>
+          <v-col>
+            <v-text-field
+              density="compact"
+              :label="applicationNumberLabel"
+              variant="outlined"
+              v-model="applicationNumber"
+              data-cy="applicationNumber"
```
ann-aot, line 20: you can use `hide-details=auto`. it will save the bottom space. it was use for all input fields

sources/packages/web/src/services/http/SupportingUserApi.ts
```diff
@@ -13,21 +13,16 @@ export class SupportingUserApi extends HttpBaseClient {
     supportingUserType: SupportingUserType,
     payload: ApplicationIdentifierApiInDTO,
   ): Promise<ApplicationApiOutDTO> {
+    let response;
```
andrewsignori-aot, line 16: The external variable to the `try` does not seems to be needed.
The code can be simplified as below:
```ts
try {
    return await this.postCall(
        this.addClientRoot(`supporting-user/${supportingUserType}/application`),
        payload,
     );
 } catch (error: unknown){
    this.handleAPICustomError(error);
 }
```

sources/packages/web/src/services/http/SupportingUserApi.ts
```diff
@@ -13,21 +13,16 @@ export class SupportingUserApi extends HttpBaseClient {
...
-      if (!error.response.data?.errorType) {
-        // If it is an not expected error,
-        // handle it the default way.
-        this.handleRequestError(error);
-      }
-      throw error;
+    } catch (error: unknown) {
+      this.handleAPICustomError(error);
     }
+    return response;
```
dheepak-aot, line 25: Do we really need a response here? are we doing something with it? If not required not need to return anything.

Also on error the undefined response it returned here which doesn't seem to be right.

sources/packages/web/src/views/supporting-user/SupportingInformation.vue
```diff
@@ -135,6 +144,11 @@ export default defineComponent({
     const isLastPage = ref(false);
     const showNav = ref(false);
     let formInstance: FormIOForm;
+    const searchApplicationsForm = ref({} as VForm);
+    const { checkOnlyDigitsRule, checkNullOrEmptyRule } = useRules();
+    const applicationNumberLabel = "Application number";
```
dheepak-aot, line 149: What is the benefit of having const for the label? as they are statis IMO we do not require const.

sources/packages/web/src/views/supporting-user/SupportingInformation.vue
```diff
@@ -8,47 +8,53 @@
...
+              ]"
+            />
+          </v-col>
+          <v-col>
+            <v-text-field
+              density="compact"
+              :label="studentsLastNameLabel"
+              variant="outlined"
+              v-model="studentsLastName"
+              data-cy="studentsLastName"
```
dheepak-aot, line 33: Thanks for taking care of data-cy 👍

sources/packages/web/src/services/http/SupportingUserApi.ts
andrepestana-aot: The issue here is that if the return is inside the try-catch, TS complains because the return can be `undefined`:
image

sources/packages/web/src/services/http/SupportingUserApi.ts
andrepestana-aot: The response is the application details used in this use case. What should we return in case of error?

sources/packages/web/src/views/supporting-user/SupportingInformation.vue
andrepestana-aot: It's used for the label and validations:
image

sources/packages/web/src/views/supporting-user/SupportingInformation.vue
andrepestana-aot: ok. thanks!

sources/packages/web/src/services/http/SupportingUserApi.ts
andrepestana-aot: I changed it as a @andrewsignori-aot suggestion as the other use cases are using postCall and handleAPICustomError.

sources/packages/web/src/services/http/SupportingUserApi.ts
andrewsignori-aot: I can see that. All other methods where we use the `handleAPICustomError` are not returning anything, this is the only one that does. The `handleAPICustomError` will throw an exception in one way or another but the compiler is not able to see it.
There is a plan to move the `handleAPICustomError` into the base HTTP class, so, IMO, it is ok to keep it as it is.
Thanks for the explanation.

ann-aot: > The issue here is that if the return is inside the try-catch, TS complains because the return can be `undefined`: image

I think, that because you need to throw the error as in line line 117 of the screenshot
image

andrewsignori-aot: Yes @ann-aot the code would be like below. We do not need to bu the compiler is not able to follow.
image

sources/packages/web/src/services/http/SupportingUserApi.ts
andrewsignori-aot: This is a not-so-usual method that does a POST to execute a "GET" because the information sent can be considered sensitive and would be open if we use the GET instead.

sources/packages/web/src/services/http/SupportingUserApi.ts
ann-aot: I realize that `hanleAPICustomError`, throws an error, maybe we can add `: never` and it will fix it. devs thoughts? 
image

sources/packages/web/src/services/http/SupportingUserApi.ts
dheepak-aot: I also recommend using Get instead.

sources/packages/web/src/views/supporting-user/SupportingInformation.vue
andrepestana-aot: Ok. As per our conversation, I removed the consts and let the messages more generic.

ann-aot (approved):
Good work @andrepestana-aot  👍 thanks for doing the changes

guru-aot (approved):

dheepak-aot (approved):
Thanks for doing the changes @andrepestana-aot . LGTM 👍

andrewsignori-aot (approved):

PR #1503, opened 2022-10-27
[Web][SIMS-#1435]-Removed max-width and min-width as it is in common component

guru-aot:
Removed max-width and min-width as it is in common modal-dailog-base component


andrewsignori-aot (approved):

ann-aot (commented):
@guru-aot I see its removed from the customer modal. but I didn't see it added to the custom component

guru-aot:
> @guru-aot I see its removed from the customer modal. but I didn't see it added to the modal-dailog-base component

Its added in the previous PR @ann-aot

ann-aot (approved):
👍

Issue #1451, opened 2022-10-12
UI Content for Institution - Create account, welcome, profile 


hellolynn-tbtb:
### **Describe the task**
- Lynn to implement content updates in DEV environments.
- Content updates include business feedback, accessibility (AA+), and B.C. government content standards.

### **Acceptance Criteria**
- [x] Update create account (part 1)
- [x] Update welcome (part 2)
- [x] Update profile (part 3)
- [x] Banner fix (text underlapping button)
- [x] Check off on forms (Allow only available values) https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/176

### **Additional context**
- Update in forms
- To familiarize push to GIT/PR process review

PR #1537, opened 2022-11-10
[UX][Web][SIMS #1451] UI Content for Institution - Profile page (banner fix)

hellolynn-tbtb:
### Profile page (banner fix)
- Fixed overlapping button on text
- Changed col to accommodate longer button text
- Checked UI for other banners (e.g. restrictions and CRA)

### Additional information
- [Figma mockups (for reference only)](https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=3651%3A95172)

### Screenshot

Larger resolution
image

Smaller resolution
image

Without button
image

### DIFF
--- a/sources/packages/web/src/components/generic/Banner.vue
+++ b/sources/packages/web/src/components/generic/Banner.vue
@@ -9,14 +9,14 @@
       <div class="label-bold-normal">{{ header }}</div>
     </template>
     <v-row>
-      <v-col md="10">
+      <v-col md="8">
         <div class="label-value-normal">
           <slot name="content">{{ summary }}</slot>
         </div>
       </v-col>
-      <v-col md="2">
-        <div class="float-right"><slot name="actions"></slot></div
-      ></v-col>
+      <v-col md="4">
+        <div class="float-right"><slot name="actions"></slot></div>
+      </v-col>
     </v-row>
   </v-alert>
 </template>
--- a/sources/packages/web/src/components/generic/Banner.vue
+++ b/sources/packages/web/src/components/generic/Banner.vue
@@ -9,14 +9,12 @@
       <div class="label-bold-normal">{{ header }}</div>
     </template>
     <v-row>
-      <v-col md="8">
+      <v-col>
         <div class="label-value-normal">
           <slot name="content">{{ summary }}</slot>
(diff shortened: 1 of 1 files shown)


ann-aot (commented):
👍

sources/packages/web/src/components/generic/Banner.vue
```diff
@@ -9,14 +9,14 @@
       <div class="label-bold-normal">{{ header }}</div>
     </template>
     <v-row>
-      <v-col md="10">
+      <v-col md="8">
```
ann-aot, line 12: I think we can add "auto" for this col. so that, it will auto-adjust for the banners with no btns

hellolynn-tbtb: Thanks @ann-aot auto works without buttons but it will look like this with buttons. So I don't think it will work.
image

ann-aot: try to remove "md" from the 2nd "v-col" and add "auto" to first "v-col" and see

ann-aot: okay... can you do something like this and see 
image

ann-aot: image
image
**zoom:150%**
image

hellolynn-tbtb: Thanks @ann-aot Hmm I like auto. But should consider the banner incase we have more text. The shot below shows a banner (button) with more text.
image

andrewsignori-aot: Just a suggestion.
```html
   <v-row>
      <v-col>
        <div class="label-value-normal">
          <slot name="content">{{ summary }}</slot>
        </div>
      </v-col>
      <v-col cols="auto"> <slot name="actions"></slot></v-col>
    </v-row>
```

hellolynn-tbtb: Thanks @andrewsignori-aot that works.

With button
image

Without button
image

andrewsignori-aot (approved):

Issue #1453, opened 2022-10-12
send student restriction notifications


abschwenker:
Additional Context
- Followup to #1423
- Create single PR for first 3 bullets
- Scheduler will need to be created to send pending messages from notifications table (specifically the case below is for the federal restrictions)

Acceptance Criteria
- [x] Move services into shared library for API and Workers
- [x] Create transaction for saving DB inserts (during file upload for example). Notification and whatever event is causing the notification.
- [x] Change notifications.user_id to be nullable
- [x] Analyze, determine, and implement path for restrictions inset and notification messages
- Federal restrictions load the temporary table and then execute sql queries based off what to insert. Perhaps an additional query is needed
- Analysis require on efficiency of federal restrictions adding 1 more sql query
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/64e3d73a-c9db-4dfd-b8df-fef4ee26958d)

- Send the following emails at the specific event, each notification will have a common set of variables (subject to change in future)

- [x] Scenario - Ministry places restriction on student account WHERE notification type != no effect
- Template name: Student - Restriction Added WHERE notification type != no effect
- givenNames, lastName, email address, date
- message sentence: A restriction has placed on your account. Please visit your account activity page to view additional details.

- [x] Scenario - System-generated restrictions on student account (example: Application submit (SINF), institution report change (WTHD), etc) WHERE notification type != no effect
- Template name: Student - Restriction Added
- givenNames, lastName, email address, date
- message sentence: A restriction has placed on your account. Please visit your account activity page to view additional details.

- [x] Scenario - Federal restriction placed on student account WHERE notification type != no effect
- Template name: Student - Restriction Added WHERE notification type != no effect
- givenNames, lastName, email address, date
- message sentence: A restriction has placed on your account. Please visit your account activity page to view additional details.

PR #1538, opened 2022-11-14
[api][SIMS #1453]Created global Config and Logger modules.

andrewsignori-aot:
- Moving LoggerService and ConfigService to the shared library.
- Created a module ConfigModule and declared it as `@Global` (global modules do not need to be imported in every single module).
- Logger module declared as `@Global`.
- ConfigService broke into smaller properties instead of a single method.


sources/packages/backend/apps/api/src/esdc-integration/sin-validation/sin-validation-processing.service.ts
```diff
@@ -35,7 +30,7 @@ export class SINValidationProcessingService {
     private readonly sequenceService: SequenceControlService,
     private readonly sinValidationIntegrationService: SINValidationIntegrationService,
   ) {
-    this.esdcConfig = config.getConfig().ESDCIntegration;
+    this.esdcConfig = config.esdcIntegration;
```
ann-aot, line 33: so we no longer have this `getConfig` in config service?

andrewsignori-aot: I don't think that we need it. The problem with the getConfig is that it parses all the env variables, which is not needed for other applications like `Workers` or the queue consumers that will be introduced. That means that not all env variables will be present on all PODs, so, depending on how they are read, the missing variables can cause an issue just for not being present.

sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -0,0 +1,196 @@
+import { Injectable } from "@nestjs/common";
+import {
```
ann-aot, line 2: 👍

sources/packages/backend/libs/utilities/src/logger/logger.service.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable, ConsoleLogger, Scope } from "@nestjs/common";
+
+/**
+ * Common log across entire solution.
+ */
+@Injectable({ scope: Scope.TRANSIENT })
```
ann-aot, line 6: for my info what is `scope: Scope.TRANSIENT`

sources/packages/backend/libs/utilities/src/config/config.module.ts
```diff
@@ -0,0 +1,9 @@
+import { Global, Module } from "@nestjs/common";
+import { ConfigService } from "./config.service";
+
+@Global()
```
ann-aot, line 4: for my info what does this decorator do?

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -59,25 +42,52 @@ async function bootstrap() {
...
+        req.headers.origin ?? "unknown"
+      } | ${req.headers["user-agent"] ?? "unknown"}`,
+    );
+    next();
+  });
 
   // pipes
   setGlobalPipes(app);
 
+  // Configure Swagger
```
ann-aot, line 58: fullstop

ann-aot (approved):
👍 Good work

sources/packages/backend/libs/utilities/src/logger/logger.service.ts
andrewsignori-aot: From Nestjs docs: "Transient providers are not shared across consumers. Each consumer that injects a transient provider will receive a new, dedicated instance."
Basically, a new instance is created every time that it is injected.

sources/packages/backend/apps/api/src/main.ts
andrewsignori-aot: Not part of the PR.

guru-aot (approved):

Issue #1541, opened 2022-11-15
Test Case - Institution Add data-cy Ids - Part 1


velang01:
**Update the existing frameworks to use data-cy ids instead of text where ever possible**
Forms/Views to be be updated 
- [x] Home/landing page
- [x] Manage institution page
- [x] Manage profile
- [x] Manage locations
- [x] Manage designation
- [x] Manage users

PR #1542, opened 2022-11-15
[cypress][SIMS #1541] Adding Institution data-cy ids - I 

velang01:
Trying to get away from using the text components as identifier for HTML elements


sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -57,7 +61,9 @@
             <template v-for="(item, index) in menuItems" :key="index">
               <v-list-item :value="index" @click="item.command">
                 <v-list-item-title>
-                  <span class="label-bold">{{ item.label }}</span>
+                  <span class="label-bold" data-cy="settingsButton1">{{
```
andrewsignori-aot, line 64: Why the `settingsButton1` was used here?
There will be many items rendered by the `v-for`, right?

sources/packages/web/src/views/institution/InstitutionProfile.vue
```diff
@@ -1,7 +1,11 @@
 <template>
   <full-page-container>
     <template #header>
-      <header-navigator title="Manage institution" subTitle="Manage Profile" />
+      <header-navigator
+        title="Manage institution"
+        data-cy="manageProfileLandingPage"
```
andrewsignori-aot, line 6: I saw sometimes this type of `data-cy` added to the `header-navigator` and sometimes to the `full-page-container`. What is the idea here?

sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/LocationCOESummary.vue
```diff
@@ -4,21 +4,26 @@
       <header-navigator
         :title="locationName"
         subTitle="Confirmation Of Enrolment"
+        data-cy="confirmationOfEnrolment"
```
andrewsignori-aot, line 7: Are those not too generic (data-cy="confirmationOfEnrolment")?
Is it not better to have something more specific like `data-cy="confirmationOfEnrolmentHeader"`?
Maybe I am not following the idea. Can you please elaborate on that?

velang01: I thought of keeping it more specific, but wanted to keep it consistent with the value in the `API` value in the formIO

sources/packages/web/src/views/institution/InstitutionProfile.vue
velang01: The id's in the header navigator are specifically for the landing page - just to make sure that the user is now navigated to the correct page. The id's in the full page containers are to retrieve the elements where the individual cy-ids cannot be assigned for the dynamic elements

sources/packages/web/src/views/institution/AppInstitution.vue
velang01: I was not completely sure if there will only be one CTA at this point. But you are right - i ll probably change it to just settings button. Will update it to be more generic and handle using index in the tests

sources/packages/web/src/views/institution/InstitutionProfile.vue
andrewsignori-aot: I did not follow this part "cannot be assigned for the dynamic elements". Is it about the form.io forms?
For the header, I would still suggest some sort of naming pattern like `manageProfileHeader`.

ann-aot: same qn

sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/LocationCOESummary.vue
velang01: updated it

sources/packages/web/src/views/institution/InstitutionProfile.vue
velang01: updated it

andrewsignori-aot (approved):
About the possible workaround to add classes to the columns of the database, was it removed from this PR?

velang01:
> Thanks for doing the changes, looks good 👍 About the possible workaround to add classes to the columns of the database, was it removed from this PR?

yeah i think i need some help from devs to do that.

ann-aot (approved):
👍

Issue #577, opened 2021-07-23
Assessment Gaps


roopfw:
**Acceptance Criteria**
- [x] Configure assessment gateway to process based off offering intensity
- [x] Migrate assessment and disbursement workflows for fulltime to camunda 8
- [x] Review assessment 2021-2022 fulltime workflow for business gaps
- [x] Migrate assessment and disbursement workflows for parttime to camunda 8
- [x] Review assessment 2021-2022 parttime workflow for business gaps

PR #1557, opened 2022-11-18
[workers][api][SIMS #577]Workers ad Student Login Sync Fixes

andrewsignori-aot:
- Changed worker filter object to ensure that when a value is `undefined` it will be converted to `null` to allow the variable to always be created on Camunda.
- Found an issue in the Student User Sync that would prevent the Student to see the top menu navigation and also from having his data revalidated when his name or DOB is changed on BCSC.


andrepestana-aot (approved):
LGTM

abschwenker (approved):
I will remove is defined check for PR#1555 as this will resolve

ann-aot (approved):
Nice catch 👍

dheepak-aot (approved):
LGTM 👍

Issue #1543, opened 2022-11-15
UI Content for Institution - Manage users, designation


hellolynn-tbtb:
### **Describe the task**
- Lynn to implement content updates in DEV environment.
- Content updates include business feedback, accessibility (AA+), and B.C. government content standards.

### **Acceptance Criteria**
- [x] Update Manage Users (including modals)
- [x] Update Manage Designation Summary
- [ ] Update Designation Details (FORM)

### **Additional context**
- This work is in Content Release 2. Noting for documentation purposes only. See[ MOD FO Channel](https://teams.microsoft.com/l/channel/19%3acaeefbc5d52f4af38678d3bb757aa3eb%40thread.tacv2/General?groupId=5b5d0688-c01d-4c16-8fb8-bf00fdddeac7&tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc) for release details.
- In addition, this work has been split up in two tickets (#1737) for Content Release 2

PR #1567, opened 2022-11-23
[UX][Web][SIMS #1543] Manage Users Page

hellolynn-tbtb:
### Manage Users Page
- New content in header (role definitions)
- Enhanced header styles for better readability on PC (rendering BCSans Bold in addition to font-weight as fallback)
- New badge border colours for higher contrast
- Table enhancements and alignments, formatting

### Additional information
- Left BCSans Bold out for smaller font sizes
- Badge border colours are derived from the main semantic colours with transparency but in HEX
- [Figma mockups (reference only)](https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=9474%3A97285&t=61Xy9Bq4brnHXkB2-0)

### Screenshot
image


sources/packages/web/src/assets/css/base.scss
```diff
@@ -330,3 +349,19 @@ html * {
   background-color: $default-background;
   border: 2px solid $common-border-color !important;
 }
+
+// List style - apply to lists used in content
+ul {
+  margin-left: 20px;
+  margin-bottom: 0;
+
```
andrewsignori-aot, line 357: Can we remove the extra blank line?

sources/packages/web/src/plugins/vuetify.ts
```diff
@@ -22,10 +22,10 @@ export default createVuetify({
           warning: "#ff7a00",
           success: "#16c92e",
           error: "#e4222e",
-          "info-bg": "#f0eef9",
-          "warning-bg": "#fff9ed",
-          "success-bg": "#eeffef",
-          "error-bg": "#fdf4f4",
+          "info-bg": "#a99fda",
+          "warning-bg": "#FFC38C",
```
ann-aot, line 26: we will also need to change here, 
image

sources/packages/web/src/assets/css/base.scss
```diff
@@ -79,15 +80,31 @@ html * {
...
 }
 
-/** Style to align records per page with arrows in data table. **/
+// Data table styles
+// Style to align records per page with arrows in data table.
 .p-paginator .p-dropdown .p-dropdown-label {
   margin-top: 8px;
 }
 
+.p-datatable.p-paginator-bottom {
```
ann-aot, line 89: mm.. these are primevue data table classes. maybe we can add a todo, to remove these styles, when we move to vuetify data table

sources/packages/web/src/assets/css/base.scss
```diff
@@ -79,15 +80,31 @@ html * {
   border-color: $badge-border-error !important;
 }
 
-/** Style to align records per page with arrows in data table. **/
+// Data table styles
```
ann-aot, line 83: mm, we add a full stop after each sentence in our comments. we review other comments too

andrewsignori-aot (approved):
Nice work, looks good 👍

sources/packages/web/src/assets/css/base.scss
hellolynn-tbtb: @ann-aot I agree. I was thinking the same. You'll see I put all the p-datatable styles stuff under a comment, which we should be able to remove easily (and not think too much about it) when we change tables.

sources/packages/web/src/plugins/vuetify.ts
hellolynn-tbtb: @ann-aot Thanks ann. The ones in Vuetify.ts change the badge border colours which are different from banner background colours. So I believe we should leave the $banner-background-info as is.

sources/packages/web/src/assets/css/base.scss
hellolynn-tbtb: @ann-aot Thanks Ann - can you give me an example what you mean? Are you referring to the commenting style?

sources/packages/web/src/assets/css/base.scss
```diff
@@ -79,15 +80,31 @@ html * {
...
+// Style to align records per page with arrows in data table.
```
hellolynn-tbtb, line 84: @ann-aot does this need to be put back using //** **// ?

sources/packages/web/src/assets/css/base.scss
hellolynn-tbtb: @ann-aot Ah okay I see we have a todo style comment now. Thank you. Done!

sources/packages/web/src/plugins/vuetify.ts
ann-aot: vuetify theme is not only for badges border, it is for all the components with property `color` and `background color` like vuetify banners, buttons,snack bar, alerts etc. these styles variable are used wherever we can't use color or background-color property and sometimes even to override the vuetify default styles

sources/packages/web/src/assets/css/base.scss
ann-aot: example, 
image
image
period `(.)` at the end of a sentance

sources/packages/web/src/assets/css/base.scss
ann-aot: for styles, we use both.  https://github.com/bcgov/SIMS/wiki/Development---Standards,-Decisions,-and-Troubleshooting#sample-class-example

hellolynn-tbtb: Thanks @ann-aot !

PR #1570, opened 2022-11-23
[camunda][SIMS#577] Fix For Disbursement Date

abschwenker:
1. Removed event that was always overwritting disbursement 1 date with today
2. Changed decimal to ceiling
3. Renamed friendly name to include fulltime and partime for readability in console


andrewsignori-aot (approved):
👍

Issue #1346, opened 2022-08-03
Determine and calculate Overawards


abschwenker:
Additional Context
- Further analysis docs are located in the analysis folder.
- This ticket is the fundamental changes for overawards, remaining UI-related tickets will be done in #1355.
- Overawards are generated only for loans.
- Overawards are generated when the student received more money than he is supposed to receive.
- Grants are not paid twice but they do not generate overawards in case more money than the student is entitled was already paid.
- All possible deductions for loans and grants are always per award code in the same way that the overaward balance is also per award code.
- When the e-Cert is generated the money is considered disbursed/paid/sent.

**Acceptance Criteria**

**Overawards generation (loans only)**

- [x] Once a reassessment happens on an application that already had money disbursed (paid), the amount already paid must be subtracted from the new assessment generated. For instance, if the student already received $2000 for CSLF and a reassessment happens where the new values indicate that the student is entitled to receive only $1500 the CSLF will have its value defined as 0 (zero) and an overaward of $500 will be added to the student account.
- [x] While calculating the total amount already paid to the student, the "overaward amount" (subtracted during e-Cert generation) is added to the "effective amount" because it should be considered part of what the student received. The "overaward amount subtracted' is money that the student is being paid but before e-Cert generation, it is used to deduct student debt, and the deduction is made with "student money".
- [x] The deductions must happen as much as possible from the sooner disbursement. For instance, if the student already received $2000 and there will be two disbursement schedules for CSLF where schedule 1 is defined as $1500 and schedule 2 is defined as $1500 the disbursements will look as below (no overawards generated for this case but it servers to illustrate how the deduction happen).
    - Disbursement Schedule 1: Award = $1500 | Deduction $1500 = 0 (**ducuted as much as possible from the first schedule**)
    - Disbursement Schedule 2: Award = $1500 | Deduction $500 = $1000

**Grants reassessments**

- [x] Grants already paid should not be paid again but if the value already paid is greater than the amount that the student is entitled to after the reassessment no overaward is generated. For instance, assuming that the student already received $2000 for CSPT and after the reassessment, the new CSPT was defined as $1500. In this situation, the student will not receive any money in the new assessment for CSPT but the $500 extra already received will not generate an overaward.

**Overall criteria**

- [x] The overawards are calculated during the execution of the `save-disbursement-schedules` worker when the disbursement schedules are created.
- [x] The overaward balance is only applied (deducted) upon e-Cert generation.
[...]

PR #1572, opened 2022-11-24
[workers][SIMS #1346]Overaward Calculation

andrewsignori-aot:
- Disbursement overaward included in the disbursement schedule calculation.
- New table and new columns created to support the overaward calculations.
image


andrewsignori-aot:
Gave on test API error. Trying to create a new PR.

Issue #1454, opened 2022-10-13
Persist Audit Info for Workers/Queue-consumers


andrewsignori-aot:
Define a solution to update the audit DB columns when the records are affected by the workers or any other operation that does not have an authenticated user.
- [x] Create 1 user
- [x] Create service to create audit user if not there, and if there get id and pass to db for audit column

PR #1573, opened 2022-11-25
[backend][sims-#1454] Persist Audit Info for Workers/Queue-consumers#1454

ann-aot:
- created a system user for auditing on demand, if the user already exists then gets the user id else create and returns the id


sources/packages/backend/libs/utilities/src/system-users/system-users.models.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * System user enum.
+ */
+export enum SystemUser {
```
andrewsignori-aot, line 4: I believe that this is not enum information.

sources/packages/backend/libs/utilities/src/system-users/system-users.models.ts
```diff
@@ -0,0 +1,14 @@
...
+  LastName = "system-user",
+  UserName = "8fb44f70-6ce6-11ed-b307-8743a2da47ef@system",
+}
+
+/**
+ *  System user interface,
```
andrewsignori-aot, line 10: Please change the comma by a period.

sources/packages/backend/libs/utilities/src/system-users/system-users.service.ts
```diff
@@ -0,0 +1,62 @@
+import { Injectable } from "@nestjs/common";
```
andrewsignori-aot, line 1: I believe that this service should be moved to the `@sims/services`.
The `@sims/utilities` is supposed to be consumed by any app/lib. If we create this here, the `@sims/utilities` will have a dependency on the `@sims/sims-db` then we no longer can consume the  `@sims/utilities` on `@sims/sims-db` otherwise we would be creating a circular dependency, don't we? Does it make sense?

sources/packages/backend/libs/utilities/src/system-users/system-users.service.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Get system user details from the cache.
+   * If not found will create a new system user
+   * and return the user details.
+   * @return system user details.
+   */
+  async systemUser(): Promise<SystemUserDetails> {
+    const systemUserKey = "systemUserDetails";
```
andrewsignori-aot, line 50: Can we please change this to private property as we talked about over the call?

sources/packages/backend/libs/utilities/src/system-users/system-users.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    return this.userRepo.save(user);
+  }
+
+  /**
+   * Get system user details from the cache.
+   * If not found will create a new system user
+   * and return the user details.
+   * @return system user details.
+   */
+  async systemUser(): Promise<SystemUserDetails> {
```
andrewsignori-aot, line 49: I do not see an issue in return the User entity model here. Thoughts?

andrewsignori-aot (commented):

sources/packages/backend/libs/utilities/src/system-users/system-users.service.ts
ann-aot: mm, I was thinking that as per future demand we can add the property in the SystemUserDetails

sources/packages/backend/libs/utilities/src/system-users/system-users.service.ts
ann-aot: mm..ya make sense

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    if (existingUser) {
+      return existingUser;
+    }
+
+    // Create new system user if not exists.
+    const user = new User();
+    user.userName = SYSTEM_USER_USER_NAME;
+    user.firstName = null;
```
guru-aot, line 42: i think firstName is nullable and no need to explicitly assign.

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/backend/libs/utilities/src/system-users/system-users.service.ts
andrewsignori-aot: I belive that either way is ok. The suggestion was to try to avoid any transformation while consuming it.
For instance, when assigning the audit user as below.
```ts
const myDatabaseModel = new MyDatabaseModel();
myDatabaseModel.creator = await this.systemUsersService.systemUser();
```
instead of
```ts
const myDatabaseModel = new MyDatabaseModel();
myDatabaseModel.creator = { id: await this.systemUsersService.systemUser().id } as User;
```

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
```diff
@@ -0,0 +1,61 @@
+import { Injectable } from "@nestjs/common";
+import { InjectRepository } from "@nestjs/typeorm";
+import { User } from "@sims/sims-db";
+import { Repository } from "typeorm";
+import { SERVICE_ACCOUNT_DEFAULT_USER_EMAIL } from "../../../utilities/src/system-configurations-constants";
```
andrewsignori-aot, line 5: Please review the `import` to use the shortcut `@sims/utilities`.

sources/packages/backend/apps/api/src/services/user/user.service.ts
```diff
@@ -1,8 +1,8 @@
 import { Injectable } from "@nestjs/common";
-import { SERVICE_ACCOUNT_DEFAULT_USER_EMAIL } from "../../utilities";
 import { DataSource, UpdateResult } from "typeorm";
 import { DataModelService, Student, User } from "@sims/sims-db";
 import { UserLoginInfo } from "./user.model";
+import { SERVICE_ACCOUNT_DEFAULT_USER_EMAIL } from "@sims/utilities";
```
dheepak-aot, line 5: Isn't `SYSTEM_USER_DEFAULT_EMAIL` be more suitable?

andrewsignori-aot (commented):
Thanks for doing the changes, minor comments only.

dheepak-aot (approved):
Looks good @ann-aot . Just one minor comment.

andrepestana-aot (approved):
👍

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
ann-aot: oh..this automatically happens when we rename a folder

sources/packages/backend/apps/api/src/services/user/user.service.ts
ann-aot: it was already there, i just moved to libs

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
andrewsignori-aot: Yes, I got it. When I was moving the files lately I was avoiding the automatic import fix to force the compile errors and adjust them manually.

sources/packages/backend/apps/api/src/services/user/user.service.ts
andrewsignori-aot: I do not have a strong reason to change it.

andrewsignori-aot (approved):

PR #1575, opened 2022-11-25
[Devops] Removing readiness and liveness probe for queue consumers

dheepak-aot:
Removing readiness and liveness probe for queue consumers as it does not have a non-authenticated end point to respond.


andrewsignori-aot (approved):
Looks good 👍

Issue #1476, opened 2022-10-20
Queues Analysis and Implementation


abschwenker:
**Analysis**
- [x] investigate nest usage of bull and redis for queue usage
- [x] investigate persistent vs memory storage
- [x] investigate backup of queue
- [x] determine work effort of of fully enabling queue in all environments with backups
- [x] determine work effort to create pods and deployment scripts
- [x] determine local setup
- [x] investigate alternatives (postgres queue?)
- [x] investigate queue monitoring ui

**Implementation**
- [x] Define and review above implementation
- [x] Implement defined above solution

**Purposed Work Strategy**
- Create new application
- Install nest package
- create consumer
- make local work
- dev review
- create and deploy 1 queue working
- dev review
- create deployment scripts
- create backups

dheepak-aot:
Created technical debt #1580 for liveness and readiness probe.

dheepak-aot:
Back up of queues are subjective on how queues are going to be used further. Created tech debt #1583 if backup is required for redis.

PR #1577, opened 2022-11-29
[API][QUEUES][SIMS #1476] Adding queue prefix and retry options

dheepak-aot:


.github/workflows/build-deploy-dev.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         working-directory: "./devops"
         run: |
           make oc-build-workers
-      - name: Build Queues Image
+      - name: Build Queue Consumers Image
```
andrewsignori-aot, line 57: There are descriptions like `Build Web, SIMS-Api and Workers` and `Deploy Web, SIMS-Api and Workers` that can be adapted. I am not sure if we added the `queue-consumers` or just stop mentioning them individually.

.github/workflows/test.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       REDIS_PORT: 6379
       REDIS_PASSWORD: redis
       REDIS_STANDALONE_MODE: "true"
+      QUEUE_PREFIX: "{sims-api-test}"
```
andrewsignori-aot, line 33: I was under the impression that the prefix would be just `DEV/TEST/QA/PROD`. Is it just for testing? If it is I would say `TEST` would be enough, IMO.

devops/Makefile
```diff
@@ -446,6 +449,7 @@ deploy-queue-consumers:
 		-p DATABASE_NAME_KEY=$(DATABASE_NAME_KEY) \
 		-p PORT=$(QUEUE_CONSUMERS_PORT) \
 		-p HOST_NAME=$(HOST) \
+		-p HOST_NAME=$(QUEUE_PREFIX) \
```
guru-aot, line 452: Can we please remove the HOST_NAME parameter in 451, i think the idea here is to use the QUEUE_PREFIX as hostname.

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -6,3 +6,13 @@ import { Duration } from "zeebe-node";
...
 export const ZEEBE_PUBLISH_MESSAGE_DEFAULT_TIME_TO_LEAVE =
   Duration.seconds.of(10);
+
+/**
+ * Number of attempts that a queue will re-try a failed job.
+ */
+export const QUEUE_RETRY_ATTEMPTS = 3;
+
+/**
+ * Delay at which every retry operation is performed on a failed job.
```
andrewsignori-aot, line 16: Please specify the unit in the comments.

sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts
```diff
@@ -32,7 +32,7 @@ export class WorkflowClientService {
         `Error while starting application assessment workflow: ${workflowName}`,
       );
       this.logger.error(error);
-      // The error is not thrown here, as we are failing silently.
+      throw error;
```
guru-aot, line 35: Can you please brief, why we are moving out from the existing fail silent scenario?

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -6,3 +6,13 @@ import { Duration } from "zeebe-node";
  */
 export const ZEEBE_PUBLISH_MESSAGE_DEFAULT_TIME_TO_LEAVE =
   Duration.seconds.of(10);
+
+/**
+ * Number of attempts that a queue will re-try a failed job.
+ */
+export const QUEUE_RETRY_ATTEMPTS = 3;
```
andrewsignori-aot, line 13: Maybe those variables can be renamed to include `default` like `QUEUE_DEFAUL_RETRY_ATTEMPTS`.
Whatever we do not know what to use we go for these ones.
Also, it can be the individual properties or the object as below:
```ts
export const QUEUE_RETRY_DEFAULT_CONFIG = { attempts: QUEUE_RETRY_ATTEMPTS, backoff: QUEUE_RETRY_DELAY };
```

sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -223,10 +223,17 @@ export class ConfigService {
   get queueConsumersPort(): number {
     return this.getCachedConfig(
       "queueConsumersPortConfig",
-      +process.env.QUEUE_CONSUMERS_PORT || 3001,
+      +process.env.QUEUE_CONSUMERS_PORT || 3002,
```
guru-aot, line 226: Thanks for doing the change, can we please change it to 3010, i know it doesn't matter, but usually we allocate 10 ports for a container. So for example in our scenario, formio as 3001 is used we take from 3000 to 3009.

sources/packages/backend/libs/services/src/queue/queue.module.ts
```diff
@@ -67,9 +68,15 @@ async function getConnectionFactory(
  * Builds the bull module options to register queues in a dynamic way.
  * @returns bull module options with all existing queues.
  */
-function getQueueModules(): BullModuleOptions[] {
-  return Queues.map<BullModuleOptions>((queue) => ({
+function getQueueModules(): BullModuleAsyncOptions[] {
```
andrewsignori-aot, line 71: What was the motivation to change it to async? Is it in preparation for a possible future change?

sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts
andrewsignori-aot, line 35: Is it the final decision to throw the error here or are we still pending a conversation with @abschwenker about it?

andrewsignori-aot (commented):
Nice research and solution using the `QUEUE_PREFIX` to enable the env segregations. Please take a look at the minor comments.

sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts
ann-aot: same qn

sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -223,10 +223,17 @@ export class ConfigService {
...
     );
   }
 
+  /**
+   * Queue prefix.
+   */
+  get queuePrefix(): string {
+    return this.getCachedConfig("queuePrefixConfig", process.env.QUEUE_PREFIX);
```
ann-aot, line 234: actauly, this.getCachedConfig is not working as intented. I think the plan is to remove it for now. right @andrewsignori-aot

sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts
dheepak-aot: The fail silent was in place because the workflow actions service was invoked in the same thread/process of submitting the applications so that if camunda is down, still the application submission happens without any interruption.

Now with queues in place, that is not the case anymore. When application is submitted, we just send a message to the queue. And workflow actions service is called when the queue messages are processed, and now we need the queue to fail if camunda is not available to that it can be re-processed when camunda is up.

sources/packages/backend/libs/services/src/queue/queue.module.ts
dheepak-aot: We need async to include the config service from where the queue prefix come from.

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -6,3 +6,13 @@ import { Duration } from "zeebe-node";
...
+
+/**
+ * Delay at which every retry operation is performed on a failed job.
+ */
+export const QUEUE_RETRY_DELAY = 180000;
```
ann-aot, line 18: our plan is to move it to env right

dheepak-aot: Yes, eventually.

guru-aot (approved):
LGTM, nice work @dheepak-aot

andrepestana-aot (approved):
LGTM!

sources/packages/backend/apps/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -162,10 +163,16 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
       );
     }
 
-    await this.startAssessmentQueue.add({
-      workflowName: assessment.application.data.workflowName,
-      assessmentId: assessment.id,
-    });
+    await this.startAssessmentQueue.add(
```
andrewsignori-aot, line 166: The idea of creating the object was to provide it as below. Did you find any issues following the below?
```
await this.startAssessmentQueue.add(
      {
        workflowName: assessment.application.data.workflowName,
        assessmentId: assessment.id,
      },
      QUEUE_RETRY_DEFAULT_CONFIG
    );
```

ann-aot (approved):
Thanks for doing the changes 👍 Good work @dheepak-aot

andrewsignori-aot (approved):

Issue #1550, opened 2022-11-16
Move Modules to Shared Library Part 3 of 3


abschwenker:
Acceptance Criteria
- Move modules used by schedulers to shared library (IER/Federal Restrictions/etc)
- [ ] Institution (IER, ECE)

- Create scheduler
- [ ] Institution (IER, ECE)

ann-aot:
https://optimalbits.github.io/bull/
![image.png](https://images.zenhubusercontent.com/608872f7d0d25cdfe8dab75d/fc5b2c20-6c06-48f2-bf2f-8cf39cefcc64)

ann-aot:
https://github.com/felixmosh/bull-board/issues/266#issuecomment-832372178,
https://github.com/s-r-x/bull-monitor

PR #1582, opened 2022-11-30
[backend][sims-#1550] Move Modules to Shared Library Part 1 of 3 - scheduler

ann-aot:
- Created an abstract base scheduler, which is extended by the integration schedulers. the responsibility of the base class is to delete old jobs if the cron options are changed and add the new job to the scheduler. (Note: if there is old retrying job, then it won't be deleted )
- Bull is smart enough not to add the same repeatable job if the repeat options are the same.
- If there are no workers running, repeatable jobs will not accumulate the next time a worker is online.
- created an IER scheduler that runs getRepeatableJobs
- Note: Passing data as a parameter is not done in this ticket, as bull board won't have this option, as per my research, I found a package called https://github.com/s-r-x/bull-monitor, details added in my ticket, https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/bcgov/sims/1550
- 
image
image
image

Reference: https://optimalbits.github.io/bull/ scroll to https://optimalbits.github.io/bull/


.github/workflows/build-deploy-dev.yml
```diff
@@ -93,6 +93,7 @@ jobs:
       WORKERS_SECRET_NAME: ${{ secrets.OPENSHIFT_SIMS_CREDS_NAME }}
       QUEUE_CONSUMERS_SECRET_NAME: ${{ secrets.OPENSHIFT_SIMS_CREDS_NAME }}
       QUEUE_PREFIX: ${{ secrets.QUEUE_PREFIX }}
+      IER_CRON: ${{ secrets.IER_CRON }}
```
andrewsignori-aot, line 96: I guess that we would have many of those. Maybe we can start with some prefixes like `SCHEDULRS_IER_CRON`.

sources/Makefile
```diff
@@ -20,6 +20,7 @@ export REDIS_STANDALONE_MODE := $(or $(REDIS_STANDALONE_MODE), true)
 export QUEUE_DASHBOARD_PASSWORD := $(or $(QUEUE_DASHBOARD_PASSWORD), queueboardpass)
 export QUEUE_DASHBOARD_USER := $(or $(QUEUE_DASHBOARD_USER), queueBoardUser)
 export QUEUE_PREFIX := $(or $(QUEUE_PREFIX), {sims-local})
+export IER_CRON := $(or $(IER_CRON),)
```
andrewsignori-aot, line 23: We add some default values like once a day.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -0,0 +1,40 @@
...
+  abstract cronOptions: Bull.JobOptions;
+  abstract schedulerQueue: Queue<T>;
+
+  /**
+   * Once all modules have been initialized
+   * it will check, if there is any old cron job
+   * delete it and add the new job to the
+   * queue.
+   */
+  async onApplicationBootstrap() {
```
andrewsignori-aot, line 15: Can we change it to `protected`?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Check if there is any old cron job  (i.e whenever there is a
+   * change in cron option, then a new job is created the old job
+   * will be still there in the queue) and delete it and add the
+   * new job to the queue.
+   * Note: If there is an old retrying job, it won't be deleted,
+   * as "getRepeatableJobs" will not fetch retrying jobs.
+   */
+  async deleteOldRepeatableJobs(): Promise<void> {
```
andrewsignori-aot, line 28: Can we change it to `private`?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
ann-aot: image

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -0,0 +1,40 @@
+import { OnApplicationBootstrap } from "@nestjs/common";
+import Bull, { CronRepeatOptions, Queue } from "bull";
+
+export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
+  abstract initializeScheduler(): void;
```
andrewsignori-aot, line 5: I believe that it should return a Promise<void>.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
andrewsignori-aot, line 15: Please make the return type explicit as `Promise<void>`.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
andrewsignori-aot, line 5: Please add comments also here.
It can start with something like `When overridden in a derived class....`.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/models/ier.model.ts
```diff
@@ -0,0 +1,8 @@
+export class IER12ResultQueueOutDTO {
+  generatedFile: string;
+  uploadedRecords: number;
+}
+
+export class GeneratedDateQueueIInDTO {
```
andrewsignori-aot, line 6: There is a double "II" in the `GeneratedDateQueueIInDTO`.

sources/packages/backend/libs/services/src/queue/constants/queue.constant.ts
```diff
@@ -2,6 +2,7 @@ import { QueueModel } from "../model/queue.model";
 
 export enum QueueNames {
   StartApplicationAssessment = "start-application-assessment",
+  StartIERIntegration = "start-ier-integration",
```
andrewsignori-aot, line 5: I would say this one can be only `ier-integration`.

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -31,3 +31,5 @@ export const QUEUE_RETRY_DEFAULT_CONFIG = {
    */
   backoff: QUEUE_RETRY_DELAY,
 };
+
+export const IER_SCHEDULER_JOB_ID = "IER_SCHEDULER_JOB_ID";
```
andrewsignori-aot, line 35: Maybe we can use the queue name here also. The one defined in the enum `QueueNames`.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
```diff
@@ -0,0 +1,75 @@
...
+    @InjectQueue(QueueNames.IERIntegration)
+    readonly schedulerQueue: Queue<GeneratedDateQueueInDTO>,
+    private readonly ierRequest: IER12FileService,
+    config: ConfigService,
+  ) {
+    super();
+    // cron options.
+    this.cronOptions = {
+      ...QUEUE_RETRY_DEFAULT_CONFIG,
+      jobId: IER_SCHEDULER_JOB_ID,
```
dheepak-aot, line 32: Is there any specific reason for passing the job id here?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
```diff
@@ -0,0 +1,75 @@
...
+  IER12ResultQueueOutDTO,
+} from "./models/ier.model";
+
+@Processor(QueueNames.IERIntegration)
+export class IERIntegrationScheduler extends BaseScheduler<GeneratedDateQueueInDTO> {
+  cronOptions: Bull.JobOptions = undefined;
+
+  constructor(
+    @InjectQueue(QueueNames.IERIntegration)
+    readonly schedulerQueue: Queue<GeneratedDateQueueInDTO>,
```
dheepak-aot, line 24: protected readonly?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
```diff
@@ -0,0 +1,75 @@
...
+import Bull, { CronRepeatOptions, Job, Queue } from "bull";
+import { BaseScheduler } from "../base-scheduler";
+import {
+  GeneratedDateQueueInDTO,
+  IER12ResultQueueOutDTO,
+} from "./models/ier.model";
+
+@Processor(QueueNames.IERIntegration)
+export class IERIntegrationScheduler extends BaseScheduler<GeneratedDateQueueInDTO> {
+  cronOptions: Bull.JobOptions = undefined;
```
dheepak-aot, line 20: protected?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
```diff
@@ -0,0 +1,75 @@
...
+  ): Promise<IER12ResultQueueOutDTO[]> {
+    this.logger.log("Executing IER 12 file generation ...");
+    const uploadResult = await this.ierRequest.processIER12File(
+      job.data.generatedDate,
+    );
+    this.logger.log("IER 12 file generation completed.");
+    return uploadResult;
+  }
+
+  @OnQueueActive()
```
dheepak-aot, line 66: This event is basically to inform other process which needs to know if the job is started. 
for e.g if we want to send an email to operations team whenever this job starts, we use these events. 
If it is just logging, then line 58 is as good as line 68.

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -31,3 +32,5 @@ export const QUEUE_RETRY_DEFAULT_CONFIG = {
    */
   backoff: QUEUE_RETRY_DELAY,
 };
+
+export const IER_SCHEDULER_JOB_ID = QueueNames.IERIntegration;
```
dheepak-aot, line 36: Unless it is a blocker, I would suggest not to assign the job id.

andrewsignori-aot: My understanding is that the `job id` is different from the `job key` which is unique and includes even the corn expression.
What is the impact of having it associated with the job?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor(
+    @InjectQueue(QueueNames.IERIntegration)
+    readonly schedulerQueue: Queue<GeneratedDateQueueInDTO>,
+    private readonly ierRequest: IER12FileService,
+    config: ConfigService,
+  ) {
+    super();
+    // cron options.
+    this.cronOptions = {
```
andrewsignori-aot, line 30: I believe that we need to review how to handle the `this.cronOptions` and the `schedulerQueue`.

For the queue, maybe the injected queue can be passed as a parameter to the constructor of the base class using the `super(...)` and the base class can have it as a protected property.

For the `cronOptions` maybe it must be handled inside the `initializeScheduler`.
This can be impacted by the upcoming changes to have the queues configured from DB. Maybe we can try to settle this as simply as possible e work on changes in the upcoming ticket.

andrewsignori-aot (commented):
Nice start for schedulers. Please take a look at the comments.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
ann-aot: mm, I don't think it only informs another process. This is a local event handler, which is invoked whenever the local event is emitted. I don't think it wrong to add logs and these are logs are job info. 
I believe to trigger another process, I think we need to use global event listers.
Anyways, I don't know if it's necessary to add job logs here, I added them as I was exploring this area, and I think I found the logging in the nest doc too. 
If the devs think that it's not required/useless here, I can remove it. thoughts?
https://docs.nestjs.com/techniques/queues#event-listeners

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
ann-aot: yes, the job id and key are different.
this job id is internally the id of the repeatable job,
image
Please refer https://github.com/bcgov/SIMS/pull/1582#discussion_r1038261065

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
dheepak-aot: Yes I know that It is a local event handler. My point it what is the significance using this event to log alone? when listening to any event, IMO there must be a valid justification.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
ann-aot: we can add multiple jobs to the same queue, for that we will need jobid, since here are going with separate queues, removing it

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
ann-aot: as I mentioned above, I added them as I was exploring this area.  moved the job to the process

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/ier-Integrations/ier-integration.scheduler.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ConfigService } from "@sims/utilities/config";
+import { InjectLogger, LoggerService } from "@sims/utilities/logger";
+import { Job, Queue } from "bull";
+import { BaseScheduler } from "../base-scheduler";
+import {
+  GeneratedDateQueueInDTO,
+  IER12ResultQueueOutDTO,
+} from "./models/ier.model";
+
+@Processor(QueueNames.IERIntegration)
```
guru-aot, line 13: Can we please change the name from IERIntegration to IER12Integration? and also the filenames should append 12 to it. 

Place them in a folder sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ier12-integration.scheduler.ts? So that similar files can be grouped and schedulers can be created.

dheepak-aot (approved):
Great piece of work @ann-aot 👍

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -0,0 +1,47 @@
+import { OnApplicationBootstrap } from "@nestjs/common";
+import { QUEUE_RETRY_DEFAULT_CONFIG } from "@sims/services/constants";
+import Bull, { CronRepeatOptions, Queue } from "bull";
+
+export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
+  // When overridden in a derived class, it hold the repeatable cron expression.
```
andrewsignori-aot, line 6: This comment is not in the VSCode format.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -0,0 +1,47 @@
...
+  protected abstract cronExpression: string;
+
+  constructor(protected schedulerQueue: Queue<T>) {}
+
+  protected get cronOptions(): Bull.JobOptions {
```
andrewsignori-aot, line 11: Add comments.

Issue #1449, opened 2022-10-12
send student notifications


abschwenker:
Additional Context
- Followup to #1423

Acceptance Criteria
- Send the following emails at the specific event, each notification will have a common set of variables (subject to change in future)

- [x] Scenario - Ministry completes or denies an exception
- Template name: Student - Student - Ministry Completes Exception
- givenNames, lastName, email address, date
- message sentence: Your exception request has been completed. Please visit your application details page to view additional information.

- [x] Scenario - Ministry completes or denies a change request
- Template name: Student - Student - Ministry Completes Requested Change
- givenNames, lastName, email address, date
- message sentence: Your requested change has been completed. Please visit your application details page to view additional information.

- [x] Scenario - School reports a change
- Template name: Student - Student - Institution Reports A Change
- givenNames, lastName, email address, date
- message sentence: Your institution has reported a change to your application. Please visit your application details page to view additional information.

- [x] Scenario - School completes a PIR
- Template name: Student - Student - Institution Completes Program Information Request
- givenNames, lastName, email address, date
- message sentence: Your institution has completed your program information request. Please visit your application details page to view additional information.

- [x] Scenario - School completes Enrollment (including tuition remittance status)
- Template name: Student - Student - Institution Confirms Enrolment
- givenNames, lastName, email address, date
- message sentence: Your institution has completed your confirmation of enrolment. Please visit your application details page to view additional information.

dheepak-aot:
Technical debt SIMS #1618 to create index for notifications table has been created.

PR #1608, opened 2022-12-07
[Queues][SIMS #1449] Send student notifications - Scheduler Part

dheepak-aot:
## Create scheduler

- [x]  Added column in notification table to save the error
- [x] Create process notification scheduler
- [x] Addressed few bugs on creating notification to db in few scenarios
- [x] Addressed a bug in student appeal


sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1268,12 +1273,18 @@ export class ApplicationService extends RecordDataModelService<Application> {
         applicationStatus: ApplicationStatus.completed,
       },
     };
-    if (applicationId) {
+    if (applicationNumber) {
```
dheepak-aot, line 1276: Existing bug which was fixed.

andrewsignori-aot: Just as a suggestion, I believe that the below would also work if `applicationNumber` and `applicationId` are undefined in the method parameter. This seems to be one of the first that we did.
```ts
const findQuery: FindOneOptions<Application> = {
      select: {
        id: true,
        applicationNumber: true,
        isArchived: true,
      },
      where: {
        student: { user: { id: userId } },
        applicationStatus: ApplicationStatus.completed,
        applicationNumber,
        id: applicationId,
      },
    };
```

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/models/notification.dto.ts
```diff
@@ -0,0 +1,9 @@
+export interface ProcessNotificationsQueueInDTO {
+  /** Maximum number of notifications to process per schedule */
```
andrewsignori-aot, line 2: Please VSCode proper for properties and IntelliSense.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/models/notification.dto.ts
```diff
@@ -0,0 +1,9 @@
...
+  pollingLimit: number;
```
andrewsignori-aot, line 3: I was under the impression that this type of parameter (non-user inputted) would be acquired and used only by the consumer.
The scheduler would be sending a parameter to the queue in a few limited scenarios like IER12 when the default date would not be the one required. But this works too.

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: what was the bug here?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/models/notification.dto.ts
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export interface ProcessNotificationsResponseQueueOutDTO {
+  notificationsProcessed: number;
+  notificationsSuccessfullyProcessed: number;
```
andrewsignori-aot, line 8: Can we add the error list here also? Even if the queue will be persisted only for 24h it can be helpful for a quick troubleshoot, if it will not add an overhead to the code.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
-      where: {
-        id: notificationId,
-      },
-    });
-    // Call GC Notify send email method.
-    const gcNotifyResult = await this.gcNotifyService.sendEmailNotification(
-      notification.messagePayload,
-    );
+  async processUnsentNotifications(
+    pollingLimit: number,
```
andrewsignori-aot, line 110: `pollingLimit` to me seems misleading. It is not a limit for the pooling but a limit to the max records to be retrieved.
I would suggest `pollingRecordsLimit` or similar.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+    notificationsProcessed = 0,
+    notificationsSuccessfullyProcessed = 0,
+  ): Promise<NotificationProcessingStatus> {
+    const notificationsToProcess = await this.repo
+      .createQueryBuilder("notification")
+      .select(["notification.id", "notification.messagePayload"])
+      .where("notification.dateSent IS NULL")
+      .limit(pollingLimit)
+      .getMany();
```
andrewsignori-aot, line 119: Please add an `order by` date ASC.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
-    const gcNotifyResult = await this.gcNotifyService.sendEmailNotification(
-      notification.messagePayload,
-    );
+  async processUnsentNotifications(
+    pollingLimit: number,
+    notificationsProcessed = 0,
+    notificationsSuccessfullyProcessed = 0,
+  ): Promise<NotificationProcessingStatus> {
+    const notificationsToProcess = await this.repo
+      .createQueryBuilder("notification")
```
andrewsignori-aot, line 115: Please use the Typeorm object query unless there is a limitation.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/process-notifications.scheduler.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Job, Queue } from "bull";
+import { NotificationService } from "@sims/services/notifications";
+import {
+  ProcessNotificationsQueueInDTO,
+  ProcessNotificationsResponseQueueOutDTO,
+} from "./models/notification.dto";
+import { BaseScheduler } from "../base-scheduler";
+import {
+  PROCESS_NOTIFICATIONS_POLLING_LIMIT,
+  PROCESS_NOTIFICATION_CLEANUP_PERIOD,
```
ann-aot, line 11: mm...actually, we are saving these values in the qeueue_configuration table, I was under the impression that we will be taking the polling limit value from the table.
or are planning to add it to my PR? that also works, I added a method to retrieve the cleanup period

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+      .getMany();
+
+    if (notificationsToProcess.length) {
+      this.logger.log(
+        `Processing ${notificationsToProcess.length} notifications`,
+      );
+
+      // Used to limit the number of asynchronous operations
+      // that will start at the same time.
+      const maxPromisesAllowed = os.cpus().length;
```
andrewsignori-aot, line 128: Because we have the `pollingLimit` I was not expecting to have this as well.
As it is already done I am ok with it, if it was not, the pooling would be enough, IMO.
Not part of this PR but we used this technique so many times that it is passed the time to build a helper method to reuse this logic 😉

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/process-notifications.scheduler.ts
```diff
@@ -0,0 +1,60 @@
...
+   * Process all the unsent notifications and return
+   * summary of processing.
+   * @param job process notification job.
+   * @returns processing summary.
+   */
+  @Process()
+  async processNotifications(
+    job: Job<ProcessNotificationsQueueInDTO>,
+  ): Promise<ProcessNotificationsResponseQueueOutDTO> {
+    const processNotificationResponse =
```
ann-aot, line 50: are we not adding any logs as we did in other scheduler or integrations?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+      const promises: Promise<boolean>[] = [];
+
+      for (const notification of notificationsToProcess) {
+        promises.push(this.sendEmailNotification(notification));
 
-    // Update date sent column in notification table after sending email notification successfully.
-    await this.updateNotification(notification.id, entityManager);
+        if (promises.length >= maxPromisesAllowed) {
+          // Waits for all be executed.
+          const processingResult = await Promise.all(promises);
```
andrewsignori-aot, line 138: We talked about avoiding using the `Promise.all` and using the `Promise.allSettled()`, does anything changed from the last conversation?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
-    return gcNotifyResult;
+      if (promises.length > 0) {
+        // Waits for methods, if any outside the loop.
+        const processingResult = await Promise.all(promises);
+        notificationsSuccessfullyProcessed =
+          notificationsSuccessfullyProcessed +
+          processingResult.filter((result) => result).length;
+      }
+      //Assign the value for total notifications processed.
+      notificationsProcessed =
```
andrewsignori-aot, line 155: This can be simplified as below:
```ts
notificationsProcessed += notificationsToProcess.length;
```

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
-      },
-      where: {
-        id: notificationId,
-      },
-    });
-    // Call GC Notify send email method.
-    const gcNotifyResult = await this.gcNotifyService.sendEmailNotification(
-      notification.messagePayload,
-    );
+  async processUnsentNotifications(
```
andrewsignori-aot, line 109: When a method needs to recurse I would prefer having a pair of methods where one of them will be public and expose the parameters needed and the other private would be myRecursiveMethod which will then expose the parameter needed to be accumulated during the recursive operation.

sources/packages/backend/libs/services/src/notifications/notification/notification.model.ts
```diff
@@ -35,3 +35,8 @@ export interface StudentNotification {
   toAddress: string;
   userId: number;
 }
+
+export interface NotificationProcessingStatus {
```
ann-aot, line 39: `NotificationProcessingSummary`?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
-   * @param entityManager optional repository that can be provided, for instance,
-   * to execute the command as part of an existing transaction. If not provided
-   * the local repository will be used instead.
+   * @param permanentFailureError Error details of a permanent failure if occurred while processing the notification.
    * @returns result of the record updated.
    */
   async updateNotification(
     notificationId: number,
-    entityManager?: EntityManager,
+    permanentFailureError?: PermanentFailureError[],
```
ann-aot, line 88: `permanentFailureErrors`?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
-   * @param notificationId notification id used to retrieve the payload that will be passed
-   * as a parameter to the sendMailNotification method.
-   * @param entityManager optional repository that can be provided, for instance,
-   * to execute the command as part of an existing transaction. If not provided
-   * the local repository will be used instead.
-   * @returns GC Notify API call response.
+   * Process all the unsent notifications with a polling limit.
+   * Processing continues recursively until all the records are processed.
+   * Call GCNotify to send email notification.
+   * @param pollingLimit
```
ann-aot, line 102: pls brief the comment

sources/packages/backend/libs/services/src/notifications/notification/gc-notify.service.ts
```diff
@@ -29,13 +33,24 @@ export class GCNotifyService {
       });
       return response.data as GCNotifyResult;
     } catch (error: unknown) {
-      const axiosError = error as AxiosError;
-      if (axiosError.isAxiosError && axiosError.response?.data) {
+      const axiosError = error as AxiosError<GCNotifyErrorResponse>;
+      if (
+        axiosError.isAxiosError &&
+        axiosError.response?.data &&
```
andrewsignori-aot, line 39: Is line 39 needed?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+            processingResult.filter((result) => result).length;
+          // Clear the array.
+          promises.splice(0, promises.length);
+        }
+      }
 
-    return gcNotifyResult;
+      if (promises.length > 0) {
+        // Waits for methods, if any outside the loop.
+        const processingResult = await Promise.all(promises);
```
ann-aot, line 149: `Promise.allSettled`?

sources/packages/backend/libs/services/src/notifications/notification/gc-notify.service.ts
```diff
@@ -1,8 +1,12 @@
-import { Injectable } from "@nestjs/common";
+import { HttpStatus, Injectable } from "@nestjs/common";
 import axios, { AxiosError } from "axios";
-import { GCNotifyResult } from "./gc-notify.model";
+import { GCNotifyErrorResponse, GCNotifyResult } from "./gc-notify.model";
 import { LoggerService, InjectLogger } from "@sims/utilities/logger";
 import { ConfigService, GCNotify } from "@sims/utilities/config";
+import { CustomNamedError } from "@sims/utilities";
+
+export const GC_NOTIFY_PERMANENT_FAILURE_ERROR =
```
andrewsignori-aot, line 8: We have been adding this type of const to `error-code-constants`.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+          notificationsSuccessfullyProcessed +
+          processingResult.filter((result) => result).length;
+      }
+      //Assign the value for total notifications processed.
+      notificationsProcessed =
+        notificationsProcessed + notificationsToProcess.length;
+
+      // Calling process notification in recursion until all the notifications
+      // are processed.
+      const response = await this.processUnsentNotifications(
```
ann-aot, line 160: I am wondering, so will there be an overlap between 2 consecutive notification queue jobs

ann-aot: do we really need this recursion? anyway, jobs are running every minute?

sources/packages/backend/libs/sims-db/src/entities/notification.model.ts
```diff
@@ -64,6 +65,16 @@ export class Notification extends RecordDataModel {
...
+
+  /**
+   * Error details of a permanent failure if occurred while processing the notification.
+   */
+  @Column({
+    name: "permanent_failure_error",
+    type: "jsonb",
+    nullable: true,
+  })
+  permanentFailureError: PermanentFailureError[];
```
ann-aot, line 77: why is it an array? is that the format?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+    notificationsSuccessfullyProcessed = 0,
+  ): Promise<NotificationProcessingStatus> {
+    const notificationsToProcess = await this.repo
+      .createQueryBuilder("notification")
+      .select(["notification.id", "notification.messagePayload"])
+      .where("notification.dateSent IS NULL")
+      .limit(pollingLimit)
+      .getMany();
+
+    if (notificationsToProcess.length) {
```
andrewsignori-aot, line 121: The big `if` "body" can be avoided if we return earlier here.
```ts
if (notificationsToProcess.length) {
    return; // Or something like this.
}
```

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +80,130 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+      if (promises.length > 0) {
+        // Waits for methods, if any outside the loop.
+        const processingResult = await Promise.all(promises);
+        notificationsSuccessfullyProcessed =
+          notificationsSuccessfullyProcessed +
+          processingResult.filter((result) => result).length;
+      }
+      //Assign the value for total notifications processed.
+      notificationsProcessed =
+        notificationsProcessed + notificationsToProcess.length;
```
andrewsignori-aot, line 156: Do we need to accumulate the values at this point or just after the recursive call on line 160?

andrewsignori-aot (commented):
Good work, nice to see the notifications and schedulers progress. Please take a look at the comments.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: @ann-aot  the bug here is, we check for applicationId and put applicationNumber in the where clause.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
dheepak-aot: I couldn't use limit(PollingRecordsLimit) by typeorm query.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/process-notifications.scheduler.ts
dheepak-aot: Planning to leave it to your PR @ann-aot as the intention of this PR is to go hard-coded.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/notification/process-notifications.scheduler.ts
dheepak-aot: As it runs every minute, I wanted add log only if a notification is present in db to be processed. Otherwise it will create log every minute.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
dheepak-aot: Yes, now that we are not throwing error at all and failing safe all the times, went for Promise.all instead of all settled.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
dheepak-aot: The array object was created to match the error response type of GC Notify. But it is always treated as one permanent failure error.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
dheepak-aot: To the first question - The scheduler jobs in bull does not overlap, if job 1 takes more than one minute job 2 will not start until job 1 is finished.

to second question - As per our sync up discussion, most of the times 100 records if present is expected to be finished in few seconds and we thought it would be a good idea not to wait until the next minute to pick the next 100 and process recursively in same job.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
andrewsignori-aot: True, I guess that we talked about it and I forgot.
Just to keep in mind, yes, there is a try/catch inside the `sendEmailNotification` where another error can happen and force the process to be aborted since the catch also do processing that it is not in a `try/catch`.
My point is, `Promise.allSettled()` would be the best choice for this type of process.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
andrewsignori-aot: @ann-aot wonder that the GC notify was out of service for 24h and we have 1000 messages to be sent.
With the recursion, we can send the 1000 messages in a few seconds, without the recursion it will be 10min, right assuming a 100 pooling.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: Thanks for doing the refactor 😉

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -6,8 +6,18 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
   // When overridden in a derived class, it hold the repeatable cron expression.
   protected abstract cronExpression: string;
 
+  /**
+   * Payload data which could be overridden if required by the implementing subclass.
+   */
+  protected get payload(): T {
+    return undefined;
```
guru-aot, line 13: 👍

sources/packages/backend/apps/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -501,32 +501,44 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
...
-          coeDeniedReasonId === COE_DENIED_REASON_OTHER_ID
-            ? otherReasonDesc
-            : null,
-        modifier: auditUser,
-        updatedAt: now,
-      })
-      .where("id = :disbursementScheduleId", { disbursementScheduleId })
-      .andWhere("coeStatus = :required", { required: COEStatus.required })
-      .execute();
+    return await this.dataSource.transaction(
```
ann-aot, line 504: await?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +84,151 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+   * Call GC Notify to process the given notification.
+   * @param notification notification to be processed.
+   * @returns status of notification processing.
+   */
+  private async sendEmailNotification(
+    notification: Notification,
+  ): Promise<boolean> {
+    // Call GC Notify send email method.
+    try {
+      await this.gcNotifyService.sendEmailNotification(
```
guru-aot, line 126: Can you please explain, why we do not have a promise.all for this and the updateNotification? I can see if the first call fails, the processNotification is updated with errors in the table. Is it because of that call, its not synced?

guru-aot (approved):
Just a minor doubt, nice work @dheepak-aot

sources/packages/backend/apps/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
andrewsignori-aot: @ann-aot is the suggestion to remove the `await` here?

sources/packages/backend/apps/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
```diff
@@ -501,32 +501,44 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
...
+          .execute();
 
-    if (updateResult.affected !== 1) {
-      throw new Error(
-        `While updating COE status to '${COEStatus.declined}' the number of affected row was bigger than the expected one. Expected 1 received ${updateResult.affected}`,
-      );
-    }
+        if (updateResult.affected !== 1) {
+          throw new Error(
+            `While updating COE status to '${COEStatus.declined}' the number of affected row was bigger than the expected one. Expected 1 received ${updateResult.affected}`,
```
andrewsignori-aot, line 528: I would change from `row was bigger` to something that just states that it is different, in this way, if zero records were affected the message would still fit.

sources/packages/backend/apps/db-migrations/src/sql/NotificationMessages/Insert-student-notification-messages.sql
```diff
@@ -39,6 +39,6 @@ INSERT INTO
 VALUES
     (
         8,
-        'Institution confirms enrolment for an application.',
+        'Institution completes enrolment for an application.',
```
andrewsignori-aot, line 42: DEV/TEST/QA is going to be manually updated?

sources/packages/backend/libs/services/src/notifications/notification/gc-notify.service.ts
```diff
@@ -29,13 +31,23 @@ export class GCNotifyService {
...
           )}`,
         );
+        // When the error is identified to be GC Notify error
+        // throw a custom error with all GC Notify error details.
+        throw new CustomNamedError(
+          axiosError.message,
+          GC_NOTIFY_PERMANENT_FAILURE_ERROR,
+          axiosError.response.data,
+        );
       } else {
```
andrewsignori-aot, line 51: Is the `else` needed here?

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -294,18 +294,18 @@ export class NotificationActionsService {
    * @param auditUserId user who confirms COE.
    * @param entityManager entity manager to execute in transaction.
    */
-  async saveInstitutionConfirmCOENotification(
+  async saveInstitutionCompletesCOENotification(
```
andrewsignori-aot, line 297: Nice catch 😉

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +84,151 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+    notificationsSuccessfullyProcessed = 0,
+  ): Promise<NotificationProcessingSummary> {
+    const notificationsToProcess = await this.repo.find({
       select: {
         id: true,
         messagePayload: true,
       },
       where: {
-        id: notificationId,
+        dateSent: IsNull(),
```
andrewsignori-aot, line 171: Can we create a tech debt t have an index added to the table that includes "dateSent/createdAt"?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +84,151 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+        promises.push(this.sendEmailNotification(notification));
+
+        if (promises.length >= maxPromisesAllowed) {
+          // Waits for all be executed.
+          const processingResult = await Promise.all(promises);
+          notificationsSuccessfullyProcessed += processingResult.filter(
+            (result) => result,
+          ).length;
+          // Clear the array.
+          promises.splice(0, promises.length);
```
andrewsignori-aot, line 201: In another opportunity, I also used `promises.length = 0` to clear the array.

andrewsignori-aot (approved):
Thanks for doing the changes. I left a few comments but they are not blockers, hence I am approving it.

sources/packages/backend/libs/sims-db/src/entities/notification-permanent-failure-error.type.ts
```diff
@@ -0,0 +1,4 @@
+export interface PermanentFailureError {
```
ann-aot, line 1: maybe u can add a comment that its the format from GC notify, that's the reason we are not following our enum standards here

sources/packages/backend/apps/api/src/services/disbursement-schedule-service/disbursement-schedule-service.ts
ann-aot: do we await here?

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -70,57 +84,151 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
         messagePayload: true,
       },
       where: {
-        id: notificationId,
+        dateSent: IsNull(),
       },
+      order: {
+        createdAt: "ASC",
+      },
+      take: pollingRecordsLimit,
```
ann-aot, line 176: 👍

ann-aot (approved):
Great work @dheepak-aot . Thanks for doing the changes 👍

Issue #990, opened 2022-02-04
Logout timer bug


ann-aot:
Actual Behavior
- if there are 2 tabs open in a same browser, and one tab is inactive more than timer period and other tab is active -> then its getting logout). Suggestion, explore ability to share inactive timeout. For instance - session storage.

- If you hibernate your computer inactivate your tab, the counter doesn't countdown. Possible solution - save date and time to session storage and compare
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/1cf4a891-76e1-4d18-985b-a0bfd7aba98b)

Expected Behavior
- User is logged out after hitting timer threshold regardless of the state of the window
- Review and implement modal UI design

PR #1611, opened 2022-12-08
[web][SIMS #990] Logout timer bug

andrepestana-aot:
If the computer hibernates or the tab/window is inactive, the setInterval does not work but when it gets back to "active" state it should count the time the user session is idle and then:
- if it is already expired, logout;
- it is not expired but there's less than COUNT_DOWN_TIMER_FOR_LOGOUT (in seconds) left, it should show the correct seconds that are left in the modal;
- otherwise the timer should count the same time independent when it got in the active state.

Also, every tab should check if the user extended their time or if they logged out. 

Some of the main changes:
- removed logic from modal; 
- created logic to use the right remaining time in IdleTimeChecker.vue; 
- logout when the other tab has logged out; 
- close modal when requested more time in other tab/window; 
- shared activity time for all tabs/windows;

image
image

### DIFF
--- a/sources/packages/web/src/components/common/IdleTimeChecker.vue
+++ b/sources/packages/web/src/components/common/IdleTimeChecker.vue
@@ -79,21 +79,21 @@ export default {
         resetIdleCheckerTimer();
       }
     };
-    let n = 0;
     const checkIdle = () => {
-      console.log("Seconds: ", ++n);
-      const idleTimeInMinutes = getDatesDiff(
+      const idleTimeInSeconds = getDatesDiff(
         lastActivityLogin.value,
         new Date(),
         "second",
-        true,
+        false,
       );
-      if (idleTimeInMinutes > minimumIdleTime.value) {
-        logoff();
-      } else if (
-        idleTimeInMinutes >=
-        minimumIdleTime.value - COUNT_DOWN_TIMER_FOR_LOGOUT
+      console.log("idleTimeInSeconds", idleTimeInSeconds);
+      if (
+        idleTimeInSeconds >
+        minimumIdleTime.value + COUNT_DOWN_TIMER_FOR_LOGOUT
       ) {
+        // Logoff immediately in case session is expired.
+        logoff();
+      } else if (idleTimeInSeconds >= minimumIdleTime.value) {
         confirmExtendTimeModal();
       }
     };
--- a/sources/packages/web/src/components/common/IdleTimeChecker.vue
+++ b/sources/packages/web/src/components/common/IdleTimeChecker.vue
@@ -29,7 +29,6 @@ import {
   MAXIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION,
   MAXIMUM_IDLE_TIME_FOR_WARNING_AEST,
   COUNT_DOWN_TIMER_FOR_LOGOUT,
-  LOGGED_OUT_LOCAL_STORAGE_ITEM,
 } from "@/constants/system-constants";
 import ConfirmExtendTime from "@/components/common/modals/ConfirmExtendTime.vue";
 
@@ -47,7 +46,7 @@ export default {
     const extendTimeModal = ref({} as ModalDialog<boolean>);
     const { isAuthenticated } = useInstitutionAuth();
     const { getDatesDiff } = useFormatters();
-    const { executeLogout } = useAuth();
+    const { executeLogout, isLoggedOut, resetLoggedOut } = useAuth();
     const countdown = ref(COUNT_DOWN_TIMER_FOR_LOGOUT);
     const IDLE_CHECKER_TIMER_INTERVAL = 1000;
     const COUNTDOWN_INTERVAL = 1000;
@@ -58,7 +57,7 @@ export default {
     onMounted(async () => {
       setLastActivityTime();
       resetIdleCheckerTimer();
-      localStorage.removeItem(LOGGED_OUT_LOCAL_STORAGE_ITEM);
+      resetLoggedOut();
     });
 
     const logoff = async () => {
@@ -117,15 +116,17 @@ export default {
       if (await extendTimeModal.value.showModal()) {
         extendUserSessionTime();
       } else {
-        setLoggedOut();
+        logoff();
       }
     };
 
     const checkIdle = () => {
-      if (getLoggedOut()) {
+      // Check if it was logged out from another tab/window.
+      if (isLoggedOut()) {
         logoff();
         return;
       }
+
       idleTimeInSeconds = getDatesDiff(
         getLastActivityTime(),
         new Date(),
@@ -141,7 +142,11 @@ export default {
         idleTimeInSeconds >=
         maximumIdleTime.value - COUNT_DOWN_TIMER_FOR_LOGOUT
       ) {
+        // Open modal.
         confirmExtendTimeModal();
+      } else {
+        // Close modal.
+        extendTimeModal.value.showDialog = false;
       }
     };
 
@@ -162,20 +167,6 @@ export default {
         return new Date(+epochLastActivityTimeStringValue);
       }
       return new Date();
-    };
-
-    const setLoggedOut = () => {
-      localStorage.setItem(LOGGED_OUT_LOCAL_STORAGE_ITEM, "true");
-    };
-
-    const getLoggedOut = () => {
-      const loggedOutStringValue = localStorage.getItem(
-        LOGGED_OUT_LOCAL_STORAGE_ITEM,
-      );
-      if (loggedOutStringValue === "true") {
-        return true;
-      }
-      return false;
     };
 
     return {
--- a/sources/packages/web/src/components/common/IdleTimeChecker.vue
+++ b/sources/packages/web/src/components/common/IdleTimeChecker.vue
@@ -113,7 +113,6 @@ export default {
     };
 
     const confirmExtendTimeModal = async () => {
-      console.log("confirmExtendTimeModal");
       modalOpen.value = true;
       if (await extendTimeModal.value.showModal()) {
         extendUserSessionTime();
@@ -132,7 +131,6 @@ export default {
         "second",
         false,
       );
-      console.log("idleTimeInSeconds", idleTimeInSeconds.value);
 
       // Exceeded user session time.
       if (idleTimeInSeconds.value > maximumIdleTime.value) {
--- a/sources/packages/web/src/components/common/IdleTimeChecker.vue
+++ b/sources/packages/web/src/components/common/IdleTimeChecker.vue
@@ -50,7 +50,6 @@ export default {
     const { getDatesDiff } = useFormatters();
     const { executeLogout } = useAuth();
     const countdown = ref(COUNT_DOWN_TIMER_FOR_LOGOUT);
-    const modalOpen = ref(false);
     const IDLE_CHECKER_TIMER_INTERVAL = 1000;
     const COUNTDOWN_INTERVAL = 1000;
     const LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM = "lastActivityTime";
@@ -185,7 +184,6 @@ export default {
       extendTimeModal,
       isAuthenticated,
       countdown,
-      modalOpen,
       extendUserSessionTime,
     };
   },
--- a/sources/packages/web/src/components/common/IdleTimeChecker.vue
+++ b/sources/packages/web/src/components/common/IdleTimeChecker.vue
@@ -13,12 +13,18 @@
 <script lang="ts">
 import { onMounted, ref, computed } from "vue";
 import { ClientIdType } from "@/types";
-import { ModalDialog, useInstitutionAuth, useFormatters } from "@/composables";
+import {
+  ModalDialog,
+  useInstitutionAuth,
+  useAuth,
+  useFormatters,
+} from "@/composables";
 import {
   MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER,
   MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT,
   MINIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION,
   MINIMUM_IDLE_TIME_FOR_WARNING_AEST,
+  COUNT_DOWN_TIMER_FOR_LOGOUT,
 } from "@/constants/system-constants";
 import ConfirmExtendTime from "@/components/common/modals/ConfirmExtendTime.vue";
 
@@ -37,7 +43,11 @@ export default {
     const extendTimeModal = ref({} as ModalDialog<boolean>);
     const { isAuthenticated } = useInstitutionAuth();
     const { getDatesDiff } = useFormatters();
+    const { executeLogout } = useAuth();
 
+    const logoff = async () => {
+      await executeLogout(props.clientIdType);
+    };
     const minimumIdleTime = computed(() => {
       switch (props.clientIdType) {
         case ClientIdType.Student:
@@ -58,7 +68,7 @@ export default {
       clearInterval(interval.value);
       if (isAuthenticated.value) {
         /* eslint-disable */
-        interval.value = setInterval(checkIdle, 30000);
+        interval.value = setInterval(checkIdle, 1000);
         /* eslint-enable */
       }
     };
@@ -69,15 +79,21 @@ export default {
         resetIdleCheckerTimer();
       }
     };
-
+    let n = 0;
     const checkIdle = () => {
-      const idleTimeInMintutes = getDatesDiff(
+      console.log("Seconds: ", ++n);
+      const idleTimeInMinutes = getDatesDiff(
         lastActivityLogin.value,
         new Date(),
         "second",
         true,
       );
-      if (idleTimeInMintutes >= minimumIdleTime.value) {
+      if (idleTimeInMinutes > minimumIdleTime.value) {
+        logoff();
+      } else if (
+        idleTimeInMinutes >=
+        minimumIdleTime.value - COUNT_DOWN_TIMER_FOR_LOGOUT
+      ) {
         confirmExtendTimeModal();
       }
     };
--- a/sources/packages/web/src/components/common/IdleTimeChecker.vue
+++ b/sources/packages/web/src/components/common/IdleTimeChecker.vue
@@ -26,11 +26,13 @@ import {
   useFormatters,
 } from "@/composables";
 import {
-  MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER,
-  MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT,
-  MINIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION,
-  MINIMUM_IDLE_TIME_FOR_WARNING_AEST,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_AEST,
   COUNT_DOWN_TIMER_FOR_LOGOUT,
+  LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM,
+  LOGGED_OUT_LOCAL_STORAGE_ITEM,
 } from "@/constants/system-constants";
 import ConfirmExtendTime from "@/components/common/modals/ConfirmExtendTime.vue";
 
@@ -57,22 +59,22 @@ export default {
     onMounted(async () => {
       setLastActivityTime();
       resetIdleCheckerTimer();
-      localStorage.removeItem("loggedOut");
+      localStorage.removeItem(LOGGED_OUT_LOCAL_STORAGE_ITEM);
     });
 
     const logoff = async () => {
       await executeLogout(props.clientIdType);
     };
-    const minimumIdleTime = computed(() => {
+    const maximumIdleTime = computed(() => {
       switch (props.clientIdType) {
         case ClientIdType.Student:
-          return MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT;
+          return MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT;
         case ClientIdType.Institution:
-          return MINIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION;
+          return MAXIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION;
         case ClientIdType.SupportingUsers:
-          return MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER;
+          return MAXIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER;
         case ClientIdType.AEST:
-          return MINIMUM_IDLE_TIME_FOR_WARNING_AEST;
+          return MAXIMUM_IDLE_TIME_FOR_WARNING_AEST;
         default:
           console.error("Invalid Client type");
           return 0;
@@ -92,7 +94,7 @@ export default {
         if (newValue) {
           countdownInterval.value = setInterval(() => {
             if (countdown.value > 0) {
-              countdown.value = minimumIdleTime.value - idleTimeInSeconds.value;
+              countdown.value = maximumIdleTime.value - idleTimeInSeconds.value;
             }
           }, 1000);
         } else {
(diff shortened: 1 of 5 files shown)


sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+      countdown.value = COUNT_DOWN_TIMER_FOR_LOGOUT;
+    };
+
     const confirmExtendTimeModal = async () => {
+      modalOpen.value = true;
       if (await extendTimeModal.value.showModal()) {
-        lastActivityLogin.value = new Date();
-        resetIdleCheckerTimer();
+        extendUserSessionTime();
+      } else {
```
guru-aot, line 119: Can we please remove the else here?

andrepestana-aot: No, we can't. If the condition returns false, it means that the user clicked on "No" to extend their user session time. In other words, the user wants to be logged out.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+        LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM,
+      );
+      if (localStorageValue) {
+        return new Date(+localStorageValue);
+      }
+      return new Date();
+    };
+
+    const setLoggedOut = () => {
+      localStorage.setItem(LOGGED_OUT_LOCAL_STORAGE_ITEM, "true");
```
andrewsignori-aot, line 175: If the user is logged off and login again, does it need to be reset?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+      const localStorageValue = localStorage.getItem(
+        LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM,
+      );
+      if (localStorageValue) {
+        return new Date(+localStorageValue);
+      }
+      return new Date();
+    };
+
+    const setLoggedOut = () => {
```
andrewsignori-aot, line 174: I am not able to follow this exactly. If the time is shared in local storage and both tabs/sessions would be monitoring the same local storage, why do we need this extra variable in the storage?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -6,19 +6,33 @@
...
+  useFormatters,
+} from "@/composables";
+import {
+  MAXIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION,
+  MAXIMUM_IDLE_TIME_FOR_WARNING_AEST,
+  COUNT_DOWN_TIMER_FOR_LOGOUT,
+  LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM,
+  LOGGED_OUT_LOCAL_STORAGE_ITEM,
```
andrewsignori-aot, line 35: If these constants are used only in this scope they can be declared only in this file.
The exception to this is usually error constants that will be mostly of the time declared to be shared.
I would recommend declaring these ones in the file scope only.
To be clear, I am talking about `LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM` and `LOGGED_OUT_LOCAL_STORAGE_ITEM`

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrewsignori-aot: I got the answer to this. It is reset during the `onMount` and since every time that we do a new login, we leave the portal and reload the entire portal the component will also be reloaded.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -6,19 +6,33 @@
     @keyup="setLastActivityTime"
   >
     <slot></slot>
-    <ConfirmExtendTime ref="extendTimeModal" :clientIdType="clientIdType" />
+    <ConfirmExtendTime
```
andrewsignori-aot, line 9: During a local test, once the modal is visible if the user clicked anywhere on the screen the modal is closed and the timer is reset. Is it the expected behavior?
I was expecting that the user must answer the modal Yes/No question. Can you try to replicate this?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+        // Logoff immediately in case session is expired.
+        logoff();
+      } else if (
+        idleTimeInSeconds.value >=
+        maximumIdleTime.value - COUNT_DOWN_TIMER_FOR_LOGOUT
+      ) {
         confirmExtendTimeModal();
+      } else {
+        // Close the modal in case idleTimeInSeconds has changed
+        // because lastAcitivityTime changed from activity in other tab/window.
```
andrewsignori-aot, line 150: Small typo on `lastActivityTime`.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
     const resetIdleCheckerTimer = () => {
       clearInterval(interval.value);
       if (isAuthenticated.value) {
-        /* eslint-disable */
-        interval.value = setInterval(checkIdle, 30000);
-        /* eslint-enable */
+        interval.value = setInterval(checkIdle, 1000);
```
andrewsignori-aot, line 87: Please move the `1000` on lines 87 and 99 to a CONST.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -32,22 +46,35 @@ export default {
...
   },
   setup(props: any) {
-    const lastActivityLogin = ref(new Date());
     const interval = ref();
     const extendTimeModal = ref({} as ModalDialog<boolean>);
     const { isAuthenticated } = useInstitutionAuth();
     const { getDatesDiff } = useFormatters();
+    const { executeLogout } = useAuth();
+    const countdown = ref(COUNT_DOWN_TIMER_FOR_LOGOUT);
+    const countdownInterval = ref(0);
```
andrewsignori-aot, line 55: I do not see a reason for the `countdownInterval` be declared with `red`. Am I missing something?

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Yes, I thought that was the wished behavior because that is how the ModalDialogBase.vue works. That requirement is not clear. I can change that to wait for a yes/no answer.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: I think we should use LOGGED_OUT_LOCAL_STORAGE_ITEM for when the user clicks on the logout in the main menu. Then other tabs would detect it and log themselves out too.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+      }
+      return new Date();
+    };
+
+    const setLoggedOut = () => {
+      localStorage.setItem(LOGGED_OUT_LOCAL_STORAGE_ITEM, "true");
+    };
+
+    const getLoggedOut = () => {
+      const localStorageValue = localStorage.getItem(
```
ann-aot, line 179: what value is this? can you pls give a meaning variable name

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: last activity time is share among the tabs but how to signal to the tabs that the user clicked on "NO" for the time extension?   "No" in this case means "Log me out now". Then the other will detect the user was logged out and  log themselves out too.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrewsignori-aot: That did not happen at least in my local test. The second tab was still logged.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrewsignori-aot: I see the point. For cross-tab communication, it is possible also to use storage events, as below. I never used it but seems straightforward. For the mentioned scenario maybe an event would be better.
```ts
window.addEventListener('storage', (event) => {
 if (event.storageArea != localStorage) return;
 if (event.key === 'loggedIn') {
   // Do something with event.newValue
 }
});
```

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrewsignori-aot: True, but it is also configurable. Checking the other modals in the solution they do not follow this behavior 😉 
I guess that there was a decision to force the user to pick one option.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+      } else {
+        setLoggedOut();
       }
     };
 
     const checkIdle = () => {
-      const idleTimeInMintutes = getDatesDiff(
-        lastActivityLogin.value,
+      if (getLoggedOut()) {
+        logoff();
```
andrewsignori-aot, line 126: This is an async method, should it wait here to wait till the redirect happen?

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: ok. I'll change that. Thanks!

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: You probably clicked on the logout in the main menu. That's what I mean. I didn't change that but I would ask if we should do that for the main logout option too (I think we should). 😄

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: ok. I see. I'm just not sure how compatible it is across the modern browsers.

andrewsignori-aot (commented):

sources/packages/web/src/components/common/IdleTimeChecker.vue
ann-aot, line 9: u can update it to `<confirm-extend-time>`

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -32,22 +46,35 @@ export default {
...
+    const countdown = ref(COUNT_DOWN_TIMER_FOR_LOGOUT);
+    const countdownInterval = ref(0);
+    const modalOpen = ref(false);
+    const idleTimeInSeconds = ref(0);
 
-    const minimumIdleTime = computed(() => {
+    onMounted(async () => {
+      setLastActivityTime();
+      resetIdleCheckerTimer();
+      localStorage.removeItem(LOGGED_OUT_LOCAL_STORAGE_ITEM);
```
ann-aot, line 62: is the purpose of this localStorage `LOGGED_OUT_LOCAL_STORAGE_ITEM` to communicate with the other tabs?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -32,22 +46,35 @@ export default {
...
+    });
+
+    const logoff = async () => {
+      await executeLogout(props.clientIdType);
+    };
+    const maximumIdleTime = computed(() => {
```
ann-aot, line 68: pls add a extra line after the function definition

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -32,22 +46,35 @@ export default {
...
   setup(props: any) {
-    const lastActivityLogin = ref(new Date());
     const interval = ref();
     const extendTimeModal = ref({} as ModalDialog<boolean>);
     const { isAuthenticated } = useInstitutionAuth();
     const { getDatesDiff } = useFormatters();
+    const { executeLogout } = useAuth();
+    const countdown = ref(COUNT_DOWN_TIMER_FOR_LOGOUT);
+    const countdownInterval = ref(0);
+    const modalOpen = ref(false);
```
ann-aot, line 56: is this name correct? is it the answer/rply from the modal?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
-        true,
+        false,
       );
-      if (idleTimeInMintutes >= minimumIdleTime.value) {
+
+      // Exceeded user session time.
+      if (idleTimeInSeconds.value > maximumIdleTime.value) {
+        if (modalOpen.value === true) {
+          modalOpen.value = false;
+          clearInterval(countdownInterval.value);
```
ann-aot, line 139: do we need this line, bcz line 138 will trigger the watch right? and watch alreday has these logic.

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -43,65 +41,30 @@ export default {
...
-    const interval = ref();
-    const { executeLogout } = useAuth();
-
-    const logoff = async () => {
-      await executeLogout(props.clientIdType);
-    };
+  setup() {
+    const { resolvePromise, showModal } = useModalDialog<boolean>();
 
     const dialogClosed = () => {
```
dheepak-aot, line 56: This method and extendTime at line 60 are doing exactly the same except the argument. THey can be unified.

sources/packages/web/src/constants/system-constants.ts
```diff
@@ -1,15 +1,19 @@
...
 // TODO : for development  added all IDLE timeout as 1hour, revert it back to 4.5 before merging to PROD
-// * MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT in seconds, i,e 270 seconds is 4.5 minutes
-export const MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT = 3600;
-// * MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT in seconds i,e 270 seconds is 4.5 minutes
-export const MINIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION = 3600;
-// * MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT in seconds i,e 270 seconds is 4.5 minutes
-export const MINIMUM_IDLE_TIME_FOR_WARNING_SUPPORTING_USER = 3600;
-// * MINIMUM_IDLE_TIME_FOR_WARNING_STUDENT in seconds i,e 270 seconds is 4.5 minutes
-export const MINIMUM_IDLE_TIME_FOR_WARNING_AEST = 3600;
+// * MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT in seconds, i,e 270 seconds is 4.5 minutes
```
dheepak-aot, line 4: May be not a part of PR, but these values can come from environment.

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
```diff
@@ -43,65 +41,30 @@ export default {
       required: true,
       default: "" as ClientIdType,
     },
+    countdown: {
+      type: Number,
+      required: true,
+    },
+    showDialog: {
```
ann-aot, line 48: bit confused.  What is happing here? is it overriding `showDialog` from `useModalDialog<boolean>();`

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
       if (isAuthenticated.value) {
-        /* eslint-disable */
-        interval.value = setInterval(checkIdle, 30000);
-        /* eslint-enable */
+        interval.value = setInterval(checkIdle, 1000);
       }
     };
 
+    watch(
+      () => modalOpen.value,
```
ann-aot, line 92: mm, what exactly is this value? `modalOpen`,

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: no, it's a variable to control/check whether the modal is open or not.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Do you mean line 139, right?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,114 @@ export default {
...
+    const extendUserSessionTime = () => {
+      setLastActivityTime();
+      resetIdleCheckerTimer();
+      modalOpen.value = false;
+      clearInterval(countdownInterval.value);
+      countdown.value = COUNT_DOWN_TIMER_FOR_LOGOUT;
+    };
+
     const confirmExtendTimeModal = async () => {
+      modalOpen.value = true;
```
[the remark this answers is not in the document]
andrepestana-aot: it needs to be set before it call `await extendTimeModal.value.showModal()`

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: yes, all the browser instances check if the variable is true and logout.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: The value is true if the modal is open for the user.

sources/packages/web/src/components/common/modals/ConfirmExtendTime.vue
andrepestana-aot: ok. I changed it to use `showDialog`. Thanks!

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Anyways, I changed that part a bit.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrewsignori-aot: Jus adding here for history, as per this page it is pretty compatible: https://www.w3schools.com/jsref/event_storage_url.asp
image

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrewsignori-aot: That is a nice catch 😉

sources/packages/web/src/constants/system-constants.ts
dheepak-aot: Are we going to have a technical debt for this. ?

dheepak-aot (approved):
Great work @andrepestana-aot 👍. 
Also did we not use the local storage events approach ?

sources/packages/web/src/constants/system-constants.ts
andrepestana-aot: @andrewsignori-aot @ann-aot @guru-aot Should we create a ticket for this?

andrepestana-aot:
> Great work @andrepestana-aot 👍. Also did we not use the local storage events approach ?

Thanks! No, we're not using the events approach.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -6,19 +6,31 @@
     @keyup="setLastActivityTime"
   >
     <slot></slot>
-    <ConfirmExtendTime ref="extendTimeModal" :clientIdType="clientIdType" />
+    <confirm-extend-time
+      ref="extendTimeModal"
+      :clientIdType="clientIdType"
```
ann-aot, line 11: i think we can remove `:clientIdType="clientIdType"` as its no longer used in `confirm-extend-time`

ann-aot (approved):
Awesome work @andrepestana-aot . Thanks for doing the changes 👍 . Add one minor comment

sources/packages/web/src/constants/system-constants.ts
ann-aot: ya. I belive our plan was to create env variable in the future

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,106 @@ export default {
...
+        LAST_ACTIVITY_TIME_LOCAL_STORAGE_ITEM,
+      );
+      if (epochLastActivityTimeStringValue) {
+        return new Date(+epochLastActivityTimeStringValue);
       }
+      return new Date();
     };
+
+    const setLoggedOut = () => {
+      localStorage.setItem(LOGGED_OUT_LOCAL_STORAGE_ITEM, "true");
```
andrewsignori-aot, line 168: It seems that a method was created in the `useAuth` to do the same. Is there a reason for this method not to be used here?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,106 @@ export default {
...
+    };
+
+    const getLoggedOut = () => {
+      const loggedOutStringValue = localStorage.getItem(
```
andrewsignori-aot, line 172: In the same way that the method `setLoggedOut` was added to the `useAuth` can we add this also? In this way the only way to manipulate the `LOGGED_OUT_LOCAL_STORAGE_ITEM` would be from the  `useAuth` and event the const LOGGED_OUT_LOCAL_STORAGE_ITEM can be set only there.

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -57,44 +84,106 @@ export default {
...
+        LOGGED_OUT_LOCAL_STORAGE_ITEM,
+      );
+      if (loggedOutStringValue === "true") {
+        return true;
+      }
+      return false;
```
andrewsignori-aot, line 178: This logic can be simplified to 
```ts
return loggedOutStringValue === "true";
```

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Yes, you're right. I'll change that. Thanks!

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Yes, you're right. I'll change that. Thanks!

sources/packages/web/src/constants/system-constants.ts
andrewsignori-aot: I do not see it as a priority right now. Can be useful for testing purposes only, IMO.

sources/packages/web/src/constants/system-constants.ts
```diff
@@ -1,15 +1,18 @@
...
+export const MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT = 3600;
+// * MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT in seconds i,e 270 seconds is 4.5 minutes
```
andrewsignori-aot, line 6: Not a part of this PR, but just a note about the "*" in front of every comment, this is not how it is supposed to be used 😉

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Done.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Done.

sources/packages/web/src/components/common/IdleTimeChecker.vue
andrepestana-aot: Done.

andrewsignori-aot (approved):
Thanks for doing the changes, it looks good 👍

PR #1613, opened 2022-12-09
[cypress][SIMS #1541] Fixing the tests 1

velang01:
- Fixing the tests

### DIFF
--- a/testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
+++ b/testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
@@ -2,58 +2,102 @@ import DashboardInstitutionObject from "../../page-objects/Institution-objects/D
 import InstitutionProfileObject from "../../page-objects/Institution-objects/InstitutionProfileObject";
 import { profileData } from "../../../data/dev/institution-data/institutionProfileData";
 import InstitutionHelperActions from "../../custom-command/institution/common-helper-functions.cy";
+import ManageInstitutionObject, {
+  SideBarMenuItems,
+} from "../../page-objects/Institution-objects/ManageInstitutionObject";
+import ManageLocationObject, {
+  Provinces,
+} from "../../page-objects/Institution-objects/ManageLocationObject";
+import { indexOf } from "cypress/types/lodash";
 
 const dashboardInstitutionObject = new DashboardInstitutionObject();
 const institutionObject = new InstitutionProfileObject();
 const institutionHelperActions = new InstitutionHelperActions();
+const manageInstitutionObject = new ManageInstitutionObject();
+const manageLocationObject = new ManageLocationObject();
 
-describe("Institution Profile", () => {
-  beforeEach(() => {
+describe("Institution Manage Profile", () => {
+  before(() => {
     institutionHelperActions.loginIntoInstitutionSingleLocation();
+    dashboardInstitutionObject.manageInstitutionButton().click();
+    manageInstitutionObject.clickOnSideBar(SideBarMenuItems.ManageProfile);
   });
 
-  it("Verify that user redirect to institution profile page", () => {
-    dashboardInstitutionObject.manageInstitutionButton().click();
-    dashboardInstitutionObject.allLocationsText().should("be.visible");
+  it("Verify that clicking on Manage Profile should navigate “My Profile Page”", () => {
     institutionObject.manageProfileButton().click();
+    dashboardInstitutionObject
+      .manageInstitutionProfileHeader()
+      .should("be.visible");
   });
 
-  it("Check that legal operating name text field should be disable or not", () => {
-    dashboardInstitutionObject.manageInstitutionButton().click();
-    dashboardInstitutionObject.allLocationsText().should("be.visible");
-    institutionObject.manageProfileButton().click();
-    institutionObject.legalOperatingNameInputText().should("be.disabled");
+  it("Verify “Manage Institution Profile” should contain institution profile details", () => {
+    institutionObject
+      .legalOperatingNameInputText()
+      .should("be.visible")
+      .should("be.disabled");
+    institutionObject
+      .operatingName()
+      .should("be.visible")
+      .should("be.disabled");
+    institutionObject
+      .institutionType()
+      .should("be.visible")
+      .should("be.disabled");
+    // institutionObject
+    //   .institutionRegulatingBody()
+    //   .should("be.visible")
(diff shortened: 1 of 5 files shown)


testing/src/cypress/page-objects/Institution-objects/BaseMethods.ts
```diff
@@ -1,5 +1,104 @@
...
+  /**
+   *
+   * @param fieldName Field Name that needs to be in the error message
+   * @param errorType What kind of error message to be constructed
+   * @returns The complete error message
+   */
+  errorMessageConstructor(fieldName: string, errorType: ErrorTypes) {
+    const fieldNameTrimmed = fieldName.trim();
+    switch (errorType) {
+      case 1:
```
andrewsignori-aot, line 27: This can be replaced with the enum type as `case ErrorTypes.fieldIsMandatory`, right?

testing/src/cypress/page-objects/Institution-objects/BaseMethods.ts
```diff
@@ -1,5 +1,104 @@
...
+    if (checkForErrorMessage) {
+      const errorMessage = this.errorMessageConstructor(
+        fieldName,
+        ErrorTypes.fieldIsMandatory
+      );
+      cy.contains(`${errorMessage}`);
+    }
+  }
+
+  /**
```
andrewsignori-aot, line 93: Please either add the comments or remove them.

testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
```diff
@@ -2,71 +2,165 @@ import DashboardInstitutionObject from "../../page-objects/Institution-objects/D
...
+    });
+  });
+
+  it("Verify that selecting Canada from country dropdown makes the pre-populate the provinces available", () => {
+    institutionObject.countryInputText().click({ force: true });
+    institutionObject.countrySearchInputText(profileData.countryInput);
+    institutionObject.provinceInputText().click();
+    const provinceIds = Object.values(Provinces);
+    for (let i = 0; i < provinceIds.length; i++) {
+      institutionObject.provinceSearchInputText(Object.keys(Provinces)[i]);
```
ann-aot, line 145: mm.. what is happing here?

testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
```diff
@@ -2,71 +2,165 @@ import DashboardInstitutionObject from "../../page-objects/Institution-objects/D
...
+      institutionObject.verifyTheDropDownOptionsAreVisible(
+        institutionObject.provinceDropdownMenu(),
+        provinceIds[i]
```
ann-aot, line 148: just a suggestion. we can use `.map() or .forEach()`, we can avoid `provinceIds[i]`.

testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
```diff
@@ -2,71 +2,165 @@ import DashboardInstitutionObject from "../../page-objects/Institution-objects/D
...
+    countryOptions.forEach((country) => {
+      institutionObject.verifyTheDropDownOptionsAreVisible(
+        institutionObject.countryDropdownMenu(),
+        country
+      );
+    });
+  });
+
+  it("Verify that selecting Canada from country dropdown makes the pre-populate the provinces available", () => {
+    institutionObject.countryInputText().click({ force: true });
```
ann-aot, line 140: just curios, what does `force: true`?

testing/src/cypress/page-objects/Institution-objects/BaseMethods.ts
```diff
@@ -1,5 +1,104 @@
...
+enum ErrorTypes {
+  fieldIsMandatory = 1,
+  fieldContainsMoreThanAllowedChars = 2,
+}
 export default class BaseMethods {
+  /**
+   *
+   * @param cyId data-cy id for the Element
+   * @returns Cypress.Chainable<JQuery<HTMLElement>>
+   */
```
ann-aot, line 12: maybe u can define return type when defining a function

andrewsignori-aot (approved):
I left a few minor comments, not blockers, hence approving it.

testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
velang01: The dropdown for provinces is not scrollable, and this makes the validation of the elements in the dropdown difficult. 

So the approach used here 
Line - 143 - Fetch all the values from Provinces enum 
Line - 145 - Search for the province by typing in the value from above, and making sure that the dropdown item exists. So typing in the dropdown search box will filter the items and the dropdown selection will be visible to cypress

testing/src/cypress/e2e/02-institution/03-institution-profile.cy.ts
velang01: It does forces the action without any errors like if there is any other element floating on the subject or even an element is disabled. 

But in this context - it clicks on the dropdown menu even if the dropdown is already expanded.

testing/src/cypress/page-objects/Institution-objects/InstitutionProfileObject.ts
```diff
@@ -8,27 +10,43 @@ export default class InstitutionProfileObject {
   }
 
   legalOperatingNameInputText() {
-    return cy.get("[data-cy='legalOperatingName']");
+    return this.getElementByCyId("legalOperatingName");
+  }
+
+  operatingName() {
+    return this.getElementByCyId("operatingName");
+  }
```
andrepestana-aot, line 18: Is it worth to create a method just to get an element?

andrepestana-aot (commented):
Good job!

testing/src/cypress/page-objects/Institution-objects/InstitutionProfileObject.ts
velang01: Yeah - the point is to be more structured and consistent in keeping all the individual page elements in a single file. This is also to avoid putting in the element identifiers directly in the test case, so if this element is being validated in more than one case, we can reduce code redundancy

andrepestana-aot (approved):

ann-aot (approved):
Good work @velang01 . Thanks for doing the changes 👍

Issue #1606, opened 2022-12-06
Move Remaining Modules to Shared Library


abschwenker:
Acceptance Criteria
- [x] Move modules for SFAS integration
- [x] Move modules for ATBC integration
- [x] Create scheduler for SFAS integration
- [x] Create scheduler for ATBC (send daily disbursement file to FIN)

dheepak-aot:
The apply for PD status call to ATBC API is retained in API for the purpose that when a student is applying for PD from application, and once the ATBC API is called the Apply PD button needs to be disabled right away. Hence removed the last AC which mentions about requesting for PD from Queues.

dheepak-aot:
Tested with SFAS File in Dev environment. 
![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/6d9aa023-08fa-4d54-9f27-d44fe1ec1398)

PR #1629, opened 2022-12-21
#1606 - Remove sfas env variable from api

dheepak-aot:
Remove sfas env variable SFAS_RECEIVE_FOLDER from api deploy yml


andrewsignori-aot (approved):
👍

Issue #1548, opened 2022-11-16
Move Modules to Shared Library Part 3 of 3


abschwenker:
Acceptance Criteria
- [x] Move modules for EDSC (SIN validation, pt/ft msfaa, pt/ft ecert, pt/ft feedback file, daily disbursement file (receipt), DailyDisbursementReport (send daily disbursement file to FIN), federal restrictions)
- [x] Create scheduler for EDSC (SIN validation, pt/ft msfaa, pt/ft ecert, pt/ft feedback file, daily disbursement file (receipt), DailyDisbursementReport (send daily disbursement file to FIN), federal restrictions)
- Ensure naming conventions use part time and full time

**done**
- process SIN
- response SIN
- PT MSFAA request
- FT MSFAA request
- PT MSFAA response
- FT MSFAA response
- PT E-Cert
- FT E-Cert
- PT Feedback file
- FT Feedback file
- FT disbursement receipt
- FIN
- Fed restriction

dheepak-aot:
Added DailyDisbursementReport (send daily disbursement file to FIN) to the AC as  it is tightly coupled with disbursement receipt processing and needs to be moved together.

ann-aot:
All integration were tested without files, are all were working as before.
there was a bug in ECert integration, it was fixed, ![image.png](https://images.zenhubusercontent.com/608872f7d0d25cdfe8dab75d/73758287-55b6-4594-a0bd-5c2573434571)

PR #1642, opened 2023-01-04
#1548 - bug fix in MSFSS and E-Cert

ann-aot:
1. #1548 - bug fix in MSFSS and E-Cert
image
image

### DIFF
--- a/sources/packages/backend/libs/integrations/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
+++ b/sources/packages/backend/libs/integrations/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
@@ -121,7 +121,7 @@ export class ECertFileHandler extends ESDCFileHandler {
       await this.disbursementScheduleService.getECertInformationToBeSent(
         offeringIntensity,
       );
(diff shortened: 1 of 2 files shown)


sources/packages/backend/libs/integrations/src/esdc-integration/e-cert-integration/e-cert-file-handler.ts
```diff
@@ -121,7 +121,7 @@ export class ECertFileHandler extends ESDCFileHandler {
       await this.disbursementScheduleService.getECertInformationToBeSent(
         offeringIntensity,
       );
-    if (!disbursements.length) {
+    if (!disbursements) {
```
andrewsignori-aot, line 124: Is this wrong because of the behavior of `mapFromRawAndEntities` that is returning `null` when there are no entities?
image

ann-aot: yes, its `mapFromRawAndEntities` returns null

andrewsignori-aot (approved):
👍

PR #1654, opened 2023-01-11
Bug in importing module for workers

guru-aot:
Bug in importing module for workers


dheepak-aot (approved):
Thanks for taking care @guru-aot 👍

Issue #1649, opened 2023-01-10
Request Basic BCeID account creation limited to one at a time


JasonCTang:
**Describe the Bug**
Staff indicated that they can only request one Basic BCeID account creation at a time.  If there is already one in progress, nobody is able to request another one. 

**Expected Behaviour**
Multiple requests to create a Basic BCeID account should be able to exist at the same time.  In other words, a user should be able to submit a Basic BCeID account creation request regardless of whether there are other existing requests..

**Actual Behaviour**
Currently, if there is already a request for a Basic BCeID account creation and someone tries to start a second one, they receive a message that their account creation request is already in progress. 

** Steps To Reproduce**
Steps to reproduce the behaviour:
1. Submit request to create basic BCeID account creation request
2. Without approving the 1st request, navigate back to the page for submitting a Basic BCeID account creation request.
3. You will not be able to load the form and instead will receive a message along the lines that your account creation request is already in progress.

PR #1671, opened 2023-01-18
#1649 - Request Basic BCeID account creation limited to one at a time

andrepestana-aot:
TypeOrm has a [bug](https://github.com/typeorm/typeorm/issues/9316) that returns a record when the param is `null` or `undefined`! 😲 
The issue happened because userId is `undefined` in `studentUserToken` in the request to check if the student has pending account application.  Then used `userName` instead of `userId`.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student-account-applications/student-account-application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student-account-applications/student-account-application.students.controller.ts
@@ -71,21 +71,14 @@ export class StudentAccountApplicationStudentsController extends BaseController 
     @UserToken() userToken: IUserToken,
     @Body() payload: CreateStudentAccountApplicationAPIInDTO,
   ): Promise<PrimaryIdentifierAPIOutDTO> {
-    if (userToken.userId) {
-      // If a userId is present in the token it means the user is already present
-      // on DB so check for possible pending student account applications.
-      const hasPendingStudentAccountApplication =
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/api/src/route-controllers/student-account-applications/student-account-application.students.controller.ts
```diff
@@ -76,7 +76,7 @@ export class StudentAccountApplicationStudentsController extends BaseController
       // on DB so check for possible pending student account applications.
       const hasPendingStudentAccountApplication =
         await this.studentAccountApplicationsService.hasPendingStudentAccountApplication(
```
guru-aot, line 78: IMO, can we do a small check if userToken.userId is undefined or null before this call is happening ? in a way we will be querying through the id rather than the username. Just a thought.

andrewsignori-aot: The user is created once the user submits the form. At this point, there is no user in our DB (because we shouldn't have one) and that is why checking by user name makes sense.
The Typeorm bug mentioned by @andrepestana-aot  masked the error but the DEV testing was really shallow also and probably there was no business test during the approval either.

andrewsignori-aot (approved):
Nice fix, looks good 👍

dheepak-aot (approved):
Good work @andrepestana-aot 👍

guru-aot (approved):
LGTM, thanks

Issue #1587, opened 2022-12-01
Dependant student with no Parents stuck waiting for data


abschwenker:
**Describe the Bug**
When a student is dependant and answers the below question with either C or D the workflow will not complete as it is waiting for parent data
![image](https://user-images.githubusercontent.com/74625947/205117704-719c2588-49d4-4501-954f-10d3b5c2bdae.png)
![image](https://user-images.githubusercontent.com/74625947/205117787-03f867b8-a0e5-44bc-8674-d7ee5f2cf485.png)

**Expected Behaviour**
- [x] Workflow should complete without the need to collect student parent data. The student is determined to be independant if answering the above question C or D

**The Fix**
- Change Form.io to use 2 values for calculating dependent status/parent panel
- [x] Change parent panel to be shown based off the current logic for determining if you are a dependent (This matters only for formio and is not intended to be used during the workflow)
- [x] Change dependant_status value to be calculated based off the panel validation (point above) AND the logic from this question

 **Steps To Reproduce**
Steps to reproduce the behaviour:
1. Create application as a dependant student
2. Answer question above with C or D

**Validations that are failing**
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/e1466279-c1df-4ab0-80ba-fa4a260c81b2)
![image.png](https://images.zenhubusercontent.com/5fce9c77aab275719112dc74/1a5b9bf5-9f70-40a2-a5fa-ac02f3b5de8f)

PR #1675, opened 2023-01-19
#1587 Get dependent status excluding parents if both parents estranged or passed away

dheepak-aot:
## Modify the logic of dependent status in form.io

- [x] Change parent panel to be shown based off the current logic for determining if you are a dependent.
Created a new hidden variable `showParentInformation` which uses the current logic of determining  dependant status.
image

- [x] Change dependant_status value to be calculated based off the panel validation (point above) AND the logic from this question
 Form.io Javascript in VS Code
image

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2021-22.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2021-22.json
@@ -2822,8 +2822,7 @@
                       },
                       "properties": {},
                       "customClass": "banner-warning",
-                      "customConditional": "",
-                      "isNew": false
+                      "customConditional": ""
                     }
                   ],
                   "allowPrevious": false,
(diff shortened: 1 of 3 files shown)


sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -31605,6 +31717,26 @@
...
+          "type": "hidden",
+          "tags": [],
+          "conditional": {
+            "show": "",
+            "when": null,
+            "eq": ""
+          },
+          "properties": {},
+          "calculateServer": true,
+          "calculateValue": "/**\r\n * When parent information is entered if student selects\r\n * both the parents passed away \r\n * or estranged from both of my parents then consider the\r\n * student as not dependant.\r\n */\r\nconst isDependant =\r\n  data.showParentInformation &&\r\n  (data.parentbcResident === \"yes\" || data.parentbcResident === \"no\");\r\n\r\nvalue = isDependant ? \"dependant\" : \"independant\";",
```
andrepestana-aot, line 31738: I think variables with the prefix `is` should be "boolean" accepting `true` and `false` or at most "yes" and "no".

dheepak-aot: I didn't get the question.

andrepestana-aot (approved):
LGTM.

guru-aot (approved):
LGTM, nice work @dheepak-aot

andrewsignori-aot (approved):
Nice and precise work, looks good 👍

PR #1681, opened 2023-01-24
#1146 - Ensure Dry Run Result Right Usage

andrepestana-aot:
Checked and did the following to ensure the right usage of the dryrun process:

- application.students.controller
	- submitApplication - changed to use values from the dryrun submission 
- designation-agreement.institutions.controller
	- submitDesignationAgreement - changed to use values from the dryrun submission 
- education-program.controller.service
	- saveProgram - ok
- institution-location.institutions.controller
	- create - ok
- report.aest.controller
	- exportReport - didn't changed validation to be done on DTO to allow dynamic data
- student.students.controller
	- create - ok
	- update - ok
- student-account-application.aest.controller
	- approveStudentAccountApplication - ok
- student-account-application.students.controller
	- create - ok
- student-appeal.students.controller
	- submitStudentAppeal - didnt changed validation to be done on DTO to allow dynamic data
- student-scholastic-standings.institutions.controller
	- saveScholasticStanding - ok
- supporting-user.supporting-users.controller
	- updateSupportingInformation - changed to use values from the dryrun submission

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -199,7 +199,7 @@ export class ApplicationStudentsController extends BaseController {
     }
 
     const submissionResult: DryRunSubmissionResult<ApplicationData> =
-      await this.formService.dryRunSubmission(
+      await this.formService.dryRunSubmission<ApplicationData>(
         programYear.formName,
         payload.data,
       );
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -158,11 +158,6 @@ export class ApplicationStudentsController extends BaseController {
     @Param("applicationId", ParseIntPipe) applicationId: number,
     @UserToken() studentToken: StudentUserToken,
   ): Promise<void> {
-    await this.applicationControllerService.offeringIntensityRestrictionCheck(
-      studentToken.studentId,
-      payload.data.howWillYouBeAttendingTheProgram,
-    );
-
     const programYear = await this.programYearService.getActiveProgramYear(
       payload.programYearId,
     );
@@ -201,6 +196,11 @@ export class ApplicationStudentsController extends BaseController {
         "Not able to create an application due to an invalid request.",
       );
     }
+
+    await this.applicationControllerService.offeringIntensityRestrictionCheck(
+      studentToken.studentId,
+      submissionResult.data.data.howWillYouBeAttendingTheProgram,
+    );
 
     const student = await this.studentService.getStudentById(
       studentToken.studentId,
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -68,6 +68,7 @@ import {
 import { ApplicationControllerService } from "./application.controller.service";
 import { CustomNamedError } from "@sims/utilities";
 import { PrimaryIdentifierAPIOutDTO } from "../models/primary.identifier.dto";
+import { ApplicationData } from "@sims/sims-db/entities/application.model";
 
 @AllowAuthorizedParty(AuthorizedParties.student)
 @RequiresStudentAccount()
@@ -197,7 +198,7 @@ export class ApplicationStudentsController extends BaseController {
       payload.data.selectedOfferingEndDate = studyEndDate;
     }
 
-    const submissionResult: DryRunSubmissionResult =
+    const submissionResult: DryRunSubmissionResult<ApplicationData> =
       await this.formService.dryRunSubmission(
         programYear.formName,
         payload.data,
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -48,11 +48,7 @@ import {
   RequiresStudentAccount,
 } from "../../auth/decorators";
 import { AuthorizedParties } from "../../auth/authorized-parties.enum";
-import {
-  ApiProcessError,
-  ClientTypeBaseRoute,
-  DryRunSubmissionResult,
-} from "../../types";
+import { ApiProcessError, ClientTypeBaseRoute } from "../../types";
 import { getPIRDeniedReason, PIR_OR_DATE_OVERLAP_ERROR } from "../../utilities";
 import {
   INVALID_APPLICATION_NUMBER,
@@ -198,7 +194,7 @@ export class ApplicationStudentsController extends BaseController {
       payload.data.selectedOfferingEndDate = studyEndDate;
     }
 
-    const submissionResult: DryRunSubmissionResult<ApplicationData> =
+    const submissionResult =
       await this.formService.dryRunSubmission<ApplicationData>(
         programYear.formName,
         payload.data,
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -48,7 +48,11 @@ import {
   RequiresStudentAccount,
 } from "../../auth/decorators";
 import { AuthorizedParties } from "../../auth/authorized-parties.enum";
-import { ApiProcessError, ClientTypeBaseRoute } from "../../types";
+import {
+  ApiProcessError,
+  ClientTypeBaseRoute,
+  DryRunSubmissionResult,
+} from "../../types";
 import { getPIRDeniedReason, PIR_OR_DATE_OVERLAP_ERROR } from "../../utilities";
 import {
   INVALID_APPLICATION_NUMBER,
@@ -144,6 +148,7 @@ export class ApplicationStudentsController extends BaseController {
   @ApiUnprocessableEntityResponse({
     description:
       "Program Year is not active or " +
+      "Selected offering id is invalid or " +
       "invalid study dates or selected study start date is not within the program year" +
       "or APPLICATION_NOT_VALID or INVALID_OPERATION_IN_THE_CURRENT_STATUS or ASSESSMENT_INVALID_OPERATION_IN_THE_CURRENT_STATE " +
       "or OFFERING_NOT_VALID.",
@@ -163,7 +168,7 @@ export class ApplicationStudentsController extends BaseController {
     );
     if (!programYear) {
       throw new UnprocessableEntityException(
-        "Program Year is not active. Not able to create an application invalid request",
+        "Program Year is not active. Not able to create an application invalid request.",
       );
     }
 
@@ -174,6 +179,11 @@ export class ApplicationStudentsController extends BaseController {
       const offering = await this.offeringService.getOfferingById(
         payload.data.selectedOffering,
       );
+      if (!offering) {
+        throw new UnprocessableEntityException(
+          "Selected offering id is invalid.",
+        );
+      }
       // if  studyStartDate is not in payload
       // then selectedOffering will be there in payload,
       // then study start date taken from offering
@@ -187,10 +197,11 @@ export class ApplicationStudentsController extends BaseController {
       payload.data.selectedOfferingEndDate = studyEndDate;
     }
 
-    const submissionResult = await this.formService.dryRunSubmission(
-      programYear.formName,
-      payload.data,
-    );
+    const submissionResult: DryRunSubmissionResult =
+      await this.formService.dryRunSubmission(
+        programYear.formName,
+        payload.data,
+      );
     if (!submissionResult.valid) {
       throw new BadRequestException(
         "Not able to create an application due to an invalid request.",
--- a/sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
@@ -82,11 +82,10 @@ export class DesignationAgreementInstitutionsController extends BaseController {
       );
     }
     // Validate the dynamic data submission.
-    const submissionResult: DryRunSubmissionResult =
-      await this.formService.dryRunSubmission(
-        FormNames.DesignationAgreementDetails,
-        payload,
-      );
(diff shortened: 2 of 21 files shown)


sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -202,6 +197,11 @@ export class ApplicationStudentsController extends BaseController {
       );
     }
 
+    await this.applicationControllerService.offeringIntensityRestrictionCheck(
```
dheepak-aot, line 200: ❤️

sources/packages/backend/apps/api/src/route-controllers/report/models/report.dto.ts
```diff
@@ -1,17 +1,25 @@
-import { Allow } from "class-validator";
+import { IsNotEmpty } from "class-validator";
 /**
  * Filter param dynamic json.
  */
 export class ReportFilterParamAPIInDTO {
-  [columnName: string]: unknown;
```
dheepak-aot, line 6: This DTO must be using dynamic key name and must not use static property keys. 

The reason is that the reports that ministry view from this page is generated in a dynamic fashion and so is the report filters.

Also this is an exceptional case because of this dynamic behavior that we must do a dry run based validation instead of class validators. 

For more context on these reports please see here https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/559

sources/packages/backend/apps/api/src/route-controllers/report/report.aest.controller.ts
```diff
@@ -66,21 +64,8 @@ export class ReportAESTController extends BaseController {
     @Body() payload: ReportsFilterAPIInDTO,
     @Res() response: Response,
   ): Promise<void> {
-    const submissionResult = await this.formService.dryRunSubmission(
-      FormNames.ExportFinancialReports,
```
dheepak-aot, line 70: Please refer to the previous comment. https://github.com/bcgov/SIMS/pull/1681/files#r1087210193

I believe we must revert this.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
```diff
@@ -4,18 +4,32 @@ import {
...
+
 /**
  * DTO for student appeal request.
  */
 export class StudentAppealRequestAPIInDTO {
   @IsNotEmpty()
   formName: string;
   @IsDefined()
-  formData: unknown;
+  formData: StudentIncomeAppealRequest | PartnerIncomeAppealRequest;
```
dheepak-aot, line 20: Same here on the context of dynamic behavior. 
Also I believe there are more appeal requests to come, so IMO we must just leave it the way it was. 

Correct me if wrong @andrewsignori-aot @guru-aot

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -111,40 +104,11 @@ export class StudentAppealStudentsController extends BaseController {
         "There is already a pending appeal for this student.",
       );
     }
-    let dryRunSubmissionResults: DryRunSubmissionResult[] = [];
-    try {
-      const dryRunPromise = payload.studentAppealRequests.map((appeal) =>
```
dheepak-aot, line 116: Please refer to the comment https://github.com/bcgov/SIMS/pull/1681/files#r1087213717

Student appeal can still use dry run.

sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
```diff
@@ -141,7 +141,7 @@ export class SupportingUserSupportingUsersController extends BaseController {
   ): Promise<void> {
     // Regardless of the API call is successful or not, create/update
     // the user being used to execute the request.
-    const userQuery = this.userService.syncUser(
+    const userQuery = await this.userService.syncUser(
```
dheepak-aot, line 144: Line 144 and 154 are expected to be promise which is resolved in line 161 (Promise.all)

dheepak-aot:
Good work @andrepestana-aot . Please take a look at the comments, most importantly on reports controller and student appeal where it we need to retain dry run validation approach.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
guru-aot: Yes you are right, currently we know these 2.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot, line 200: Not part of the PR but a bit critical, if at line 174 the `getOfferingById` does not return an offering, we can throw as below:
```ts
if(!offering){
    throw new BadRequestException(
        "Selected offering id is invalid.",
    );
}
```

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
```diff
@@ -105,9 +105,9 @@ export class DesignationAgreementInstitutionsController extends BaseController {
     const createdDesignation =
       await this.designationAgreementService.submitDesignationAgreement(
         userToken.authorizations.institutionId,
-        payload.dynamicData,
+        submissionResult.data.dynamicData,
```
andrewsignori-aot, line 108: These values are being consumed as `any`. We need to type them in some way.
Maybe we can change the `DryRunSubmissionResult` as below:
```ts
export interface DryRunSubmissionResult<T = unknown> {
  valid: boolean;
  data?: T;
  formName: string;
}
```
The `filter((location)` and the `map((location)` are treating the `location` and `any` also due to this.
If not in this PR I believe we should do one more PR right after this and maybe also change the `data` property to `resultData` to avoid the `data.data`.

@guru-aot @dheepak-aot FYI

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
andrewsignori-aot: In terms of dynamic data, as a general rule, the API must define types for whatever the API need to interact with. So far for these appeals (and hopefully all others to come), the API's responsibility is just to receive and save them (also validate with the `dryRun`) as they are to later also provide them as they were received to the workflow.
Short answer, please leave it as it was following what @dheepak-aot mentioned.

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -59,7 +59,7 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
   async saveStudentAppeals(
     applicationId: number,
     userId: number,
-    studentAppealRequests: StudentAppealRequestModel[],
+    studentAppealRequests: StudentAppealRequestAPIInDTO[],
```
andrewsignori-aot, line 62: The idea here was to avoid the service relying on a DTO class. Is there a reason to have it changed?

andrewsignori-aot (commented):
Good work chasing the different points to be changed. Please take a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
andrepestana-aot: Yes, you're right. Thanks. I've trying something and forgot to remove that.

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
andrepestana-aot: My idea was to remove duplicate code as we have `StudentAppealRequestModel` and `StudentAppealRequestAPIInDTO`. As I reverted student appeal changes, I'm reverting this one too. Thanks!

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
andrepestana-aot: For now I just typed the dryrun submission result. We need to talk about it. Thanks!

andrepestana-aot: I made the `DryRunSubmissionResult` optional generic. Then when needed/possible we can type the `.data.data` object.

dheepak-aot (approved):
LGTM @andrepestana-aot 👍

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -192,16 +198,22 @@ export class ApplicationStudentsController extends BaseController {
       payload.data.selectedOfferingEndDate = studyEndDate;
     }
 
-    const submissionResult = await this.formService.dryRunSubmission(
-      programYear.formName,
-      payload.data,
-    );
+    const submissionResult: DryRunSubmissionResult<ApplicationData> =
```
andrewsignori-aot, line 201: Thank you for doing this change.
Can the `dryRunSubmission` be also adapted as below to complete the change?
```ts
async dryRunSubmission<T = any>(
    formName: string,
    data: unknown,
  ): Promise<DryRunSubmissionResult<T>>
```

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
```diff
@@ -81,10 +82,11 @@ export class DesignationAgreementInstitutionsController extends BaseController {
       );
     }
     // Validate the dynamic data submission.
-    const submissionResult = await this.formService.dryRunSubmission(
-      FormNames.DesignationAgreementDetails,
-      payload,
-    );
+    const submissionResult: DryRunSubmissionResult =
```
andrewsignori-aot, line 85: The explicit type here is optional. `submissionResult` is already typed.

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
```diff
@@ -72,9 +78,10 @@ export class SubmitDesignationAgreementAPIInDTO {
  * be validated by the form.io dryrun validation.
  */
 export class SubmittedLocationsAPIInDTO {
-  @Allow()
+  @IsNumber()
```
andrewsignori-aot, line 81: My understanding is that `@IsNumber()` is already validated by the `@IsPositive()`.
The other places along the API using `@IsPositive()` are also not used in combination with the `@IsNumber()`.

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
andrewsignori-aot: Thanks for reverting it. The code should not be considered duplicated in this case. The DTO and the service model belong to different layers and they can evolve differently.
Now that we have the API service layer more dedicated to the API it is not as a big deal as it used to be but ideally, we try to avoid using API DTOs at the service level.
@dheepak-aot @guru-aot please correct me if I am not saying something accurate here.

sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
```diff
@@ -95,10 +96,11 @@ export class EducationProgramControllerService {
     auditUserId: number,
     programId?: number,
   ): Promise<EducationProgram> {
-    const submissionResult = await this.formService.dryRunSubmission(
-      FormNames.EducationProgram,
-      payload,
-    );
+    const submissionResult: DryRunSubmissionResult<SaveEducationProgram> =
```
andrewsignori-aot, line 99: Please see https://github.com/bcgov/SIMS/pull/1681#discussion_r1089270994

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
andrepestana-aot: Yes. Thanks!

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
```diff
@@ -63,7 +66,13 @@ export class UpdateDesignationAPIInDTO {
 export class SubmitDesignationAgreementAPIInDTO {
   @Allow()
   dynamicData: unknown;
-  @Allow()
+  /**
```
andrewsignori-aot, line 69: I would recommend changing as below. I like the idea of using the form "`locations`" but I believe that would be more for the inline object types, for instance, when we create inline `options` types.
```ts
/**
   * Locations being designated.
   * Must be validated because it is not part of the dynamic data.
   */
```

sources/packages/backend/apps/api/src/services/form/form.service.ts
```diff
@@ -88,41 +88,6 @@ export class FormService {
...
   }
 
-  /**
-   * Validate and save the data to the From IO environment generating also
-   * the absolute path for the formio generated submission.
-   * @param formName Name of the form to be saved.
-   * @param data Data to be saved.
-   * @returns Submission related information from formio submission result.
-   */
-  async submission(formName: string, data: any): Promise<SubmissionResult> {
```
andrewsignori-aot, line 98: Thanks for removing the no longer needed code ❤️

sources/packages/backend/apps/api/src/types/form.ts
```diff
@@ -1,14 +1,5 @@
-export interface DryRunSubmissionResult {
+export interface DryRunSubmissionResult<T = any> {
```
andrewsignori-aot, line 1: In the future, in a separate effort, we can change the `any` to `unknown`. Right now this is which will bring some improvement and cause less impact.
Thanks for doing this change also.

andrewsignori-aot (approved):
Thanks for doing the changes, refactors, and adding the generic type to the dry run. Looks good 👍

Issue #1666, opened 2023-01-13
Review Dependabot recommendations


andrewsignori-aot:
Review the current Dependabot recommendations and clear the PRs.

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/9ddd41e7-e240-40d2-b51f-4c1404d8d276)

PR #1682, opened 2023-01-24
#1666-Dependabot dependency error

guru-aot:
image
Resolved the dependency which the @nestjs/mapped-types version that did not update, once they implement the class-validator in their version, the package-lock will be updated automatically.

FYI: Similar reference of error in build in Stackoverflow
https://stackoverflow.com/questions/75139457/error-or-breaking-changes-in-nest-class-validator0-13-2


andrewsignori-aot (approved):
👍

PR #1688, opened 2023-01-26
Formio dockerfile update

guru-aot:
As forms image was deleted and the formio was not supporting the latest version of node artifactory, changed the supported version of node and build the dockerfile.

### DIFF
--- a/sources/packages/forms/Dockerfile
+++ b/sources/packages/forms/Dockerfile
@@ -45,6 +45,9 @@ RUN apk del git
 #          See docker-compose.yml for instructions
 RUN ln -sf $NPM_PACKAGES/node_modules node_modules
 
+RUN ls
+RUN PWD
+
 # Set this to inspect more from the application. Examples:
 #   DEBUG=formio:db (see index.js for more)
 #   DEBUG=formio:*
--- a/sources/packages/forms/Dockerfile
+++ b/sources/packages/forms/Dockerfile
@@ -12,16 +12,11 @@ WORKDIR /app
 RUN apk update && \
   apk upgrade && \
   apk add make=4.2.1-r2 && \
-  apk add g++=8.3.0-r0
-
-# At least one buried package dependency is using a `git` path.
-# Hence we need to haul in git.
-RUN apk --update add git
-# Use https to avoid requiring ssh keys for public repos.
-RUN git config --global url."https://github.com/".insteadOf "ssh://git@github.com/"
+  apk add g++=8.3.0-r0 \
+  apk add git
 
 # Clone code
-RUN git clone ${FORMIO_SOURCE_REPO_URL} -b ${FORMIO_SOURCE_REPO_TAG} --single-branch /app
+RUN git clone ${FORMIO_SOURCE_REPO_URL} -b ${FORMIO_SOURCE_REPO_TAG} --single-branch /app/
 
 # Using an alternative package install location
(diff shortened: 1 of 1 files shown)


sources/packages/forms/Dockerfile
```diff
@@ -32,13 +30,12 @@ ENV NPM_PACKAGES=/.npm-packages \
...
-
-# Use "Continuous Integration" to install as-is from package-lock.json
-RUN npm ci --prefix=$NPM_PACKAGES
+RUN mkdir -p $NPM_PACKAGES/
+# Include details of the required dependencies
+RUN cp package.json $NPM_PACKAGES/
 
-RUN apk del git
+# Use install as-is from package.json
+RUN npm i --prefix=$NPM_PACKAGES
```
dheepak-aot, line 38: What is the benefit of npm i over npm ci here. 

Formio docker file uses npm ci, thats why asked.

sources/packages/forms/Dockerfile
```diff
@@ -32,13 +30,12 @@ ENV NPM_PACKAGES=/.npm-packages \
...
-COPY ./package.json $NPM_PACKAGES/
-COPY ./package-lock.json $NPM_PACKAGES/
-
-# Use "Continuous Integration" to install as-is from package-lock.json
-RUN npm ci --prefix=$NPM_PACKAGES
+RUN mkdir -p $NPM_PACKAGES/
+# Include details of the required dependencies
+RUN cp package.json $NPM_PACKAGES/
 
-RUN apk del git
```
dheepak-aot, line 41: Why apk del git was removed.?

sources/packages/forms/Dockerfile
```diff
@@ -32,13 +30,12 @@ ENV NPM_PACKAGES=/.npm-packages \
...
 
 # Include details of the required dependencies
-COPY ./package.json $NPM_PACKAGES/
-COPY ./package-lock.json $NPM_PACKAGES/
-
-# Use "Continuous Integration" to install as-is from package-lock.json
-RUN npm ci --prefix=$NPM_PACKAGES
+RUN mkdir -p $NPM_PACKAGES/
+# Include details of the required dependencies
+RUN cp package.json $NPM_PACKAGES/
```
dheepak-aot, line 35: why package-lock is not copied?

sources/packages/forms/Dockerfile
guru-aot: the package-lock and package.json has mismatched version of libraries, so when the build happens, the installations was not happening due to the conflict.

sources/packages/forms/Dockerfile
guru-aot: It was unwanted, followed the formio dockerfile

sources/packages/forms/Dockerfile
guru-aot: when we are installing via the package.json using npm i the package-lock.json is implicitly unused. So removed it and it may cause libraries, version mismatch conflict.

andrewsignori-aot (approved):
👍

andrewsignori-aot (commented):
I am still not confident that this should be the PROD model. Can we create a tech debt to review the form.io installation?
@guru-aot

dheepak-aot (approved):
Thanks for all the troubleshooting efforts.👍

Issue #1665, opened 2023-01-13
Implement Banners for Enrolment stage on Student Application Tracker


JasonCTang:
### Context
Enhancements to display relevant information for the **Enrolment** stage of the student application tracker. These may need to be broken up into smaller tickets.
______________________
### Reference mockups
https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=7809%3A108528&t=WNRHHCnCKfJXwzsH-0
______________________
#### Enrolment stage 
- [x] If COE, display "waiting" banner and orange bar
- [x] If COE DECLINED, display red banner and red bar (w/ alert icon)
![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/69a26848-1db4-447c-a7ad-ee655de52cc9)
- [x] If 2 COE, display 2 "waiting" banners and orange bar
- [x] In "Application actions" menu, replace "Edit applications" with "Request a change" link

______________________

#### Completed (for original assessments) 
**HAPPY PATH 😸**
- [x] If COE confirmed, display green banner and green bar (w/ check icon)
- [x] If Payment SENT, display green banner and green bar (w/ check icon)
- [x] If 1 of 2 COE confirmed, display 1 green banner, green bar (w/ check icon), 3 waiting banners
- [x] If 1 of 2 Payment SENT, display 2 green banners, green bar (w/ check icon), 2 waiting banners
- [x] If 2 of 2 COE confirmed, display 3 green banners, green bar (w/ check icon), 1 waiting banner
- [x] If 2 of 2 Payment SENT, display 4 green banners, green bar (w/ check icon)
- [x] NICE TO HAVE (background green highlight of last card) - DEV investigate what we have done for other stages
![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/bea8dc21-88de-481b-ba91-0938f0ac4f1b)

 **UNHAPPY PATH 😿**

- [x] If 1 of 2 COE (DECLINED), display red banner and red bar (in enrolment)
- [x] If 2 of 2 COE (DECLINED), display red bar (in completed), 2 green banners, 1 red banner, and hide waiting banner
![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/e4338e66-50ec-4e59-98d0-5422f22c34dd)

______________________

#### Completed (for reassessments)
 **UNHAPPY PATH 😿**
[...]

PR #1689, opened 2023-01-26
#1665 - Application progress tracker for enrolment

dheepak-aot:
## Application progress tracker updated for enrolment status 

** Progress tracker for completed status will be done in the upcoming PR **

- [x] If COE, display "waiting" banner and orange bar
image

- [x] If COE DECLINED, display red banner and red bar (w/ alert icon)
image

- [x] If 2 COE, display 2 "waiting" banners and orange bar
image

- [x] Created API to get the enrolment details of a given student application to load all enrolment related

### Additional refactoring work done

- Created API to get all the requests and confirmation status of student application like Application exception status, PIR status, and COE status(both the COE status). This API will guide the tracker to identify if an application has reached it's terminal state where the application cannot proceed further.

e.g PIR declined by institution, application exception declined, COE declined or application completed with all the COEs confirmed.

- Refactored the Progress Bar to be a separate component(which is tightly coupled with application progress now) which could be re-used if required anywhere.

- Modified the theme colors warning-bg, success-bg, error-bg and info-bg to match with figma.

This eventually fixed the background color of error and warning banners.
image

image

- Added theme colors info-chip-outline, warning-chip-outline, success-chip-outline and error-chip-outline which is also updated in figma.

- Updated the content of cancelled application banner to add the context of applications being cancelled by system when there is an approved offering change requested by institution and if any of the application which belongs to the offering has not reached completed status is being cancelled.

image

- Height of the progress bar has been adjusted as per figma.

Current progress bar

image

Adjusted progress bar

image


sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -482,10 +488,95 @@ export class ApplicationStudentsController extends BaseController {
...
+    const application = await this.applicationService.getApplicationDetails(
+      applicationId,
+      userToken.studentId,
+    );
+    if (!application) {
+      throw new NotFoundException(
+        `Application id ${applicationId} was not found.`,
+      );
+    }
+    const disbursements = application.currentAssessment?.disbursementSchedules
```
andrewsignori-aot, line 519: It could be simplified as below.
```ts
const disbursements = application.currentAssessment?.disbursementSchedules ?? [];
```

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1420,16 +1428,66 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+            coeStatus: true,
+            disbursementDate: true,
+            disbursementScheduleStatus: true,
+            coeDeniedReason: { id: true, reason: true },
+            coeDeniedOtherDesc: true,
+          },
+        },
+      },
+      relations: {
+        currentAssessment: { disbursementSchedules: { coeDeniedReason: true } },
```
andrewsignori-aot, line 1473: Do we have an idea if this creates an inner or a left join?

sources/packages/web/src/assets/css/vuetify.scss
```diff
@@ -131,6 +131,12 @@
   border-bottom: 2px solid $common-border-color;
 }
 
+// CSS to match the height of progress bar as per figma.
+.v-slider.v-input--horizontal .v-slider-track__fill,
```
andrewsignori-aot, line 135: Is this assuming that every progress bar across the AEST/student/etc. will be the same?

sources/packages/web/src/components/common/ProgressBar.vue
```diff
@@ -0,0 +1,85 @@
+<template>
```
andrewsignori-aot, line 1: `ProgressBar` seems to be too generic in this case. Should it not be more specific and mention the "steps", for instance? For instance `SteppedProgressBar`.

sources/packages/web/src/components/common/ProgressBar.vue
```diff
@@ -0,0 +1,85 @@
...
+    </template>
+  </v-slider>
+</template>
+<script lang="ts">
+import { PropType, defineComponent, computed } from "vue";
+
+export default defineComponent({
+  props: {
+    progressStepLabels: {
+      type: Object as PropType<Record<number, string>>,
```
andrewsignori-aot, line 38: Why do we need the number here? The array index would not be enough?

sources/packages/web/src/components/common/ProgressBar.vue
```diff
@@ -0,0 +1,85 @@
...
+      type: String,
+      required: false,
+    },
+    progressLabelIconColor: {
+      type: String,
+      required: false,
+    },
+  },
+  setup(props) {
+    const currentStatus = computed(() => props.progressBarValue);
```
andrewsignori-aot, line 74: Is there a reason for the `props.progressBarValue` not being consumed directly?

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -45,36 +22,54 @@
...
 import Submitted from "@/components/students/applicationTracker/Submitted.vue";
 import InProgress from "@/components/students/applicationTracker/InProgress.vue";
 import Cancelled from "@/components/students/applicationTracker/Cancelled.vue";
 import Assessment from "@/components/students/applicationTracker/Assessment.vue";
+import Enrolment from "@/components/students/applicationTracker/Enrolment.vue";
 
+interface ApplicationEndStatusIconDetails {
+  endStatusType?: "success" | "error";
+  endStatusIcon?: string;
+}
```
andrewsignori-aot, line 62: Please add a blank line below.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: Type orm find* methods use left join. 
image

sources/packages/web/src/assets/css/vuetify.scss
dheepak-aot: It was done assuming the stepper progress bar height is same through the application. But can confirm the same with @hellolynn-tbtb

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -85,26 +80,68 @@ export default defineComponent({
...
+          applicationProgressDetails.value.secondCOEStatus ===
+            COEStatus.completed)
+      ) {
+        applicationEndStatus.value = {
+          endStatusType: "success",
+          endStatusIcon: "fa:fas fa-check-circle",
+        };
+      }
+      // One of the requests or confirmation is declined.
+      else if (
```
andrewsignori-aot, line 126: IMO we can get rid of the first `if` and the second `if` can be the first one where the `else` would be the 
```t
applicationEndStatus.value = {
      endStatusType: "success",
      endStatusIcon: "fa:fas fa-check-circle",
   };
```

sources/packages/web/src/components/common/ProgressBar.vue
dheepak-aot: Makes sense. What about StepperProgessBar?

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -85,26 +80,68 @@ export default defineComponent({
...
+      }
+      if (applicationEndStatus.value.endStatusType === "error") {
+        return "error";
+      }
+      return "warning";
+    });
+
+    onMounted(async () => {
+      applicationProgressDetails.value =
+        await ApplicationService.shared.getApplicationProgressDetails(
```
andrewsignori-aot, line 109: Since we have a specific API method to get the application progress (`getApplicationProgressDetails`) I would not mind having the API returning a status for it that would make all the logic from line 112 till 140 more simple.

sources/packages/web/src/components/common/ProgressBar.vue
dheepak-aot: Because the props are immutable and cannot be assigned to a v-model(2 way binded) directly.

image

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
```diff
@@ -130,27 +167,21 @@ export default defineComponent({
       [ApplicationStatus.draft, ApplicationStatus.submitted].includes(
         props.applicationStatus,
       )
-        ? 20
+        ? 14
```
andrewsignori-aot, line 170: Please convert it to a `const` in this file.

sources/packages/web/src/components/students/applicationTracker/Cancelled.vue
```diff
@@ -1,13 +1,18 @@
...
     icon="fa:fas fa-exclamation-circle"
     icon-color="danger"
     background-color="error-bg"
     ><template #content
-      >You cancelled your application on
+      >Your application was cancelled on
       {{ dateOnlyLongString(cancelledDate) }}. However you can still view your
-      application by clicking on the “Application actions” button.
+      application by clicking on the "Application actions" button. If you did
+      not cancel your application, it means that your application has been
```
andrewsignori-aot, line 11: @hellolynn-tbtb I am not sure how accurate this is or how close to production need it will be.
```
it means that your application has been canceled due to changes in the study period made by your institution.
```
In my opinion, it could be more generic and not point to the institution as the only possible point of responsibility for the change.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
dheepak-aot: By using the second if we are trying to ecape the default path where all the requests for an application are in Required Status where Icon and Icon color doesn't needs to be set.

sources/packages/web/src/components/students/applicationTracker/EnrolmentBanner.vue
```diff
@@ -0,0 +1,73 @@
...
+      required: true,
+    },
+  },
+  setup(props) {
+    const enrolmentBannerDetails = computed<EnrolmentBannerModel>(() => {
+      switch (props.coeStatus) {
+        case COEStatus.completed:
+          return {
+            bannerLabel:
+              "Your enrolment has been confirmed by your institution",
```
andrewsignori-aot, line 42: I can see that other statuses with the "heavy text" defined part at the template level. Is this one following a different approach or am I missing something?

dheepak-aot: Are you talking about label? I see the label defined as string argument in all the other status as well.

image

dheepak-aot: > I can see that other statuses with the "heavy text" defined part at the template level. Is this one following a different approach or am I missing something? For instance: image

About content?

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
dheepak-aot: I was initially thinking the same, then with the feeling that the logic is a client specific one, moved the logic to client side keeping API simple.

sources/packages/web/src/components/students/applicationTracker/MultipleEnrolmentBanner.vue
```diff
@@ -0,0 +1,132 @@
...
+    const getBannerDetails = (
+      coeStatus: COEStatus,
+      coeDenialReason: string,
+    ): EnrolmentBannerModel => {
+      switch (coeStatus) {
+        case COEStatus.completed:
+          return {
+            bannerIcon: "fa:fas fa-check-circle",
+            iconColor: "success",
+            content:
```
andrewsignori-aot, line 72: Same from https://github.com/bcgov/SIMS/pull/1689/files#r1091124809

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: Are you sure that it is not based on the entity model? I was expecting a sample from the generated query.

sources/packages/web/src/assets/css/vuetify.scss
andrewsignori-aot: "stepper progress bar height" can maybe be the same but here we are setting the height for any v-slider, right?
My point was that maybe we should do it by adding one more class to the CSS filter.

sources/packages/web/src/components/common/ProgressBar.vue
andrewsignori-aot: That works too.

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
andrewsignori-aot: "client specific one" is a debatable point. Unless the API is returning some property like `statusForTheProgressBar` or `colorToThatComponent`, having the API returning something aligned with the `ProgressDetails` will not be a problem IMO.

sources/packages/web/src/components/students/applicationTracker/EnrolmentBanner.vue
andrewsignori-aot: I was talking about all of them
image

sources/packages/web/src/components/students/applicationTracker/ApplicationProgressBar.vue
dheepak-aot: As discussed leaving the logic to get the end status stay at the client side and when we get the requirement to show the application end status we can either have the API return some status or move the existing client side code to composable.

andrewsignori-aot (approved):

sources/packages/web/src/components/students/applicationTracker/Cancelled.vue
hellolynn-tbtb: Thanks @andrewsignori-aot. Yeah we may need it to be more generic if it contains other scenarios. But I am okay to leave it as is since we still have the "content review" track with business.

hellolynn-tbtb (commented):
LGTM! Just in the application tracker we spell "Enrollment" but should be "enrolment" with 1 L if that is a simple label change we can do now. If it is a bigger change, no problem. Just leave it for now. Thanks @dheepak-aot. Good work on UI!

Issue #380, opened 2021-04-20
Add server side from based validation of profile (Institution and student) pages.


fwpushan:
**Describe the task**
Add server side from based validation of profile (Institution and student) pages.

**Acceptance Criteria**
- [ ] Server side validation based on from dry run submission 
- [ ] A valid submission must accepted by server
- [ ] Create e2e tests (api) for profiles form validation.

**Additional context**

PR #1698, opened 2023-02-01
#380 - Constraint to save draft

andrepestana-aot:
- Added max size validation to dynamic content that is saved as a json to DB. Created `@JsonMaxSize()` used as a property validator for objects that must have a maximum length as a json.
- Fixed `AppAllExceptionsFilter` to show exception details as below:
image
- Confirmed that the Nest's default configuration restricts all body lengths to be a maximum of 100kb. Example of log for Nest's default restriction: 
Screenshot 2023-01-30 171534

Created a ticket [Add class-validators to dryRun input DTOs](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/1700) as technical debt found while resolving the ticket.


sources/packages/backend/apps/api/src/app.exception.filter.ts
```diff
@@ -6,14 +6,17 @@ import { LoggerService, InjectLogger } from "@sims/utilities/logger";
...
 
     // Logging Additional info
-    this.logger.error("Un-handle exception");
-    this.logger.error(`Request Path [${request.path}]`);
-    this.logger.error(`Exception Details: \n ***** \n\t${exception} \n *****`);
+    this.logger.error("Unhandled exception");
+    this.logger.error(`Request path [${request.path}]`);
+    this.logger.error(
+      `Exception details: \n ***** \n\t${
+        exception["response"] ? exception["response"]["message"] : exception
```
dheepak-aot, line 17: Here we are assuming that if response is present, then `response.message` must be available. 

Possibly extend the condition 
```
exception["response"] &&  exception["response"]["message"]  ? 
exception["response"]["message"] : error;

```

sources/packages/backend/apps/api/src/auth/decorators/payload-max-size.decorator.ts
```diff
@@ -0,0 +1,8 @@
+import { SetMetadata } from "@nestjs/common";
+
+/**
+ * This decorator checks if the payload is greater than the maximum size in bytes.
+ */
+export const PAYLOAD_MAX_SIZE_KEY = "payload-max-size";
+export const PayloadMaxSize = (maxSize: number) =>
```
dheepak-aot, line 7: We can have a comment here mentioning the unit of maxSize(kb | mb).

sources/packages/backend/apps/api/src/utilities/class-validation/custom-validators/json-max-size.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Max size validation for the json resulted from the annotated object.
+ */
+@ValidatorConstraint()
+class JsonMaxSizeConstraint implements ValidatorConstraintInterface {
+  validate(json: object, args: ValidationArguments): boolean {
+    const [maxSize] = args.constraints;
+    const stringifiedJson = JSON.stringify(json);
+    if (stringifiedJson.length > maxSize) {
```
andrewsignori-aot, line 16: The code can be simplified to 
```ts
return stringifiedJson.length <= maxSize;
```

sources/packages/backend/apps/api/src/utilities/class-validation/custom-validators/json-max-size.ts
```diff
@@ -0,0 +1,41 @@
...
+  registerDecorator,
+  ValidationArguments,
+} from "class-validator";
+
+/**
+ * Max size validation for the json resulted from the annotated object.
+ */
+@ValidatorConstraint()
+class JsonMaxSizeConstraint implements ValidatorConstraintInterface {
+  validate(json: object, args: ValidationArguments): boolean {
```
andrewsignori-aot, line 13: Please change the `object` to `unknown`.

sources/packages/backend/apps/api/src/auth/guards/payload-max-size.guard.ts
```diff
@@ -0,0 +1,37 @@
...
+export class PayloadMaxSizeGuard implements CanActivate {
+  constructor(private reflector: Reflector) {}
+  /**
+   * Implementation of canActivate method in CanActivate.
+   * Checks if the payload is not greater than max size.
+   * @param context Execution context.
+   * @returns true if payload length is not greater than the max size.
+   */
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const payloadMaxSize = this.reflector.getAllAndOverride<any>(
```
dheepak-aot, line 23: `this.reflector.getAllAndOverride<number>` ?

sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
```diff
@@ -90,6 +91,7 @@ export class EducationProgramControllerService {
       "Not able to a save the program due to an invalid request or " +
       "SABC code is duplicated.",
   })
+  @PayloadMaxSize(20480)
```
dheepak-aot, line 94: Correct me if wrong, we can use this in controller instead of controller.service right?

sources/packages/backend/apps/api/src/app.exception.filter.ts
andrewsignori-aot: Using `exception["response"]["message"]` is not recommended. It is bypassing exactly what the `unknown` is trying to restrict and enforce.
Since this was not part of the PR I would recommend just doing a simple `JSON.stringify(exception)` (if possible).

sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
andrewsignori-aot: I am under the impression that they would not work if not in the controller, please correct me if I am wrong. Either way, the guards are to be added to the controllers, unless I am missing something.

sources/packages/backend/apps/api/src/auth/auth.module.ts
```diff
@@ -81,6 +82,10 @@ const jwtModule = JwtModule.register({
       provide: APP_GUARD,
       useClass: SINValidationGuard,
     },
+    {
+      provide: APP_GUARD,
+      useClass: PayloadMaxSizeGuard,
```
andrewsignori-aot, line 87: Maybe I got lost in the middle of some conversation but I was under the impression that we should be having one or another way to have the dynamic payloads restricted.
Is the `PayloadMaxSize` to be used when we have a `dryRun` executed on DTOs with the `@Allow()`?I If that is the case, that was not the intention of the ticket because these types of dynamic-ish DTOs can also receive a `@MaxLength`.
Can we please have a quick dev sync-up on this?
@dheepak-aot thoughts?

sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
```diff
@@ -11,12 +11,14 @@ import {
...
   DisbursementScheduleStatus,
 } from "@sims/sims-db";
+import { JsonMaxSize } from "../../../utilities/class-validation";
 
 export class SaveApplicationAPIInDTO {
   /**
    * Application dynamic data.
    */
   @IsObject()
+  @JsonMaxSize(20480)
```
andrewsignori-aot, line 21: We try to avoid using the numbers directly and the recommended way would be to create a const.
It could be something trivial as `PAYLOAD_2MB` or something similar.

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
```diff
@@ -3,6 +3,7 @@ import {
   DesignationAgreementStatus,
   NOTE_DESCRIPTION_MAX_LENGTH,
 } from "@sims/sims-db";
+import { JsonMaxSize } from "../../../utilities/class-validation";
```
andrewsignori-aot, line 6: Can you please add this to the index file and import it from there?

sources/packages/backend/apps/api/src/auth/auth.module.ts
dheepak-aot: Looking at `PayloadMaxSize`  I understood that it was for payloads which were absolutely dynamic like student appeals, but looking at education-program controller using it, I didn't get the idea. 
Quick sync up is required here.

sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
andrepestana-aot: Yes, you're right. It should be only in the controllers. Thanks!

sources/packages/backend/apps/api/src/auth/auth.module.ts
andrepestana-aot: This `PayloadMaxSizeGuard` is to check the whole payload. There are some requests we use the whole body to save it as a json and there are others that we have a property of the payload object that will be saved as a json. So for those properties we can use the class validator and for the payload we can use the guard. We can discuss that.

sources/packages/backend/apps/api/src/app.exception.filter.ts
andrepestana-aot: Do you prefer this way, @andrewsignori-aot ?
image

andrewsignori-aot (commented):
Nice research and good findings about Nestjs out-of-box constraints also. Please take a look at the comments and let's have a quick dev sync-up to review some comments.

dheepak-aot:
Great work on implementing the json constraint. Please have a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
andrepestana-aot: It's added to class-validation's index. Do you mean import from the utilities` index?

sources/packages/backend/apps/api/src/app.exception.filter.ts
andrewsignori-aot: Short answer, yes 😉
This would be better because we would have access to the entire information and any piece of it can be valid for troubleshooting.
A middle-term solution would be figuring out if the `exception` has a type defined on Nestjs (or creating a type for it) that allows access(typed) to the most important information to avoid something like, what seems to be, repeated information shown in the above printscreen shared.
Since it was not initially part of this ticket, to keep it simple, I would go with the `JSON.stringify(exception)`.
@dheepak-aot thoughts?

sources/packages/backend/apps/api/src/auth/auth.module.ts
andrepestana-aot: ok, removed `@PayloadMaxSize` as discussed.

sources/packages/backend/apps/api/src/app.exception.filter.ts
dheepak-aot: I agree.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
```diff
@@ -1,20 +1,29 @@
...
   IsEnum,
   IsNotEmpty,
   IsPositive,
+  MaxLength,
   ValidateNested,
 } from "class-validator";
 import { StudentAppealStatus } from "@sims/sims-db";
+import { JsonMaxSize } from "../../../utilities/class-validation";
+
+const JSON_10KB = 10240;
```
dheepak-aot, line 15: The constants FORM_NAME_MAX_LENGTH and JSON_10KB can be global and not just for the scope of student appeal DTO.

andrepestana-aot: Where to put it?

sources/packages/backend/apps/api/src/route-controllers/student-account-applications/models/student-account-application.dto.ts
```diff
@@ -1,8 +1,12 @@
+import { JsonMaxSize } from "../../../utilities/class-validation";
 import { Allow, IsNotEmptyObject } from "class-validator";
 import { CreateStudentAPIInDTO } from "../../../route-controllers/student/models/student.dto";
 
+const JSON_10KB = 10240;
```
dheepak-aot, line 5: Must be a constant shared by multiple DTOs.  We can declare them in a constant file created in `route-controllers/utils`

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
dheepak-aot: Sorry I mentioned that in the next comment. 
We can declare them in a constant file created in `route-controllers/utils`

sources/packages/backend/apps/api/src/route-controllers/student-account-applications/models/student-account-application.dto.ts
andrewsignori-aot: We also have the `src/constants`.

sources/packages/backend/apps/api/src/utilities/class-validation/custom-validators/json-max-size.ts
```diff
@@ -0,0 +1,38 @@
...
+} from "class-validator";
+
+/**
+ * Max size validation for the json resulted from the annotated object.
+ */
+@ValidatorConstraint()
+class JsonMaxSizeConstraint implements ValidatorConstraintInterface {
+  validate(json: unknown, args: ValidationArguments): boolean {
+    const [maxSize] = args.constraints;
+    const stringifiedJson = JSON.stringify(json);
```
andrewsignori-aot, line 15: One last-minute question, if the `json` parameter is `undefined`, does it not cause an exception when `stringifiedJson.length` is checked?

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
andrewsignori-aot: Usually, we create `constants` folders as needed with specific constant files.
We already have one under `api/constants`. We can add specific files there and just add it to the index.

andrepestana-aot: ok. Done. Thanks!

sources/packages/backend/apps/api/src/route-controllers/student-account-applications/models/student-account-application.dto.ts
andrepestana-aot: ok. Done. Thanks!

sources/packages/backend/apps/api/src/utilities/class-validation/custom-validators/json-max-size.ts
andrepestana-aot: Yes,  good catch. Added a check. Thanks!

sources/packages/backend/apps/api/src/constants/class-validator-constants.ts
```diff
@@ -0,0 +1,11 @@
+export const JSON_10KB = 10240;
+export const JSON_20KB = 2048;
```
andrewsignori-aot, line 2: I believe that there is a missing "0" on 2048.

dheepak-aot (approved):
Thanks for doing the changes. LGTM 👍

andrewsignori-aot (approved):

sources/packages/backend/apps/api/src/constants/class-validator-constants.ts
andrepestana-aot: Thanks!

Issue #1676, opened 2023-01-20
PIR Data not showing up in institutions on TEST/DEV


hellolynn-tbtb:
**Describe the Bug**
Institution doesn't display PIR data 
It looks like this happens when PIR created without program or offering selected
This was discovered due to student applications stuck "submitted" stage (did not go forward to "in progress")

**Expected Behaviour**
Display data

**Actual Behaviour**

![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/37e88535-0d7c-4e09-8c6d-40cb68e50485)

Network console message
![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/f2ee3203-50ff-475e-86c7-9cabee8b61c8)

** Steps To Reproduce**
Steps to reproduce the behaviour:
1. Student submits application with PIR (without program or offering selected)
2. Institution clicks on student PIR
3. Institution sees student PIR data (see above screenshot)

PR #1708, opened 2023-02-04
#1676 - PIR Data not showing up in institutions

andrepestana-aot:
- added optional chaining to avoid exceptions when undefined;
- tested other all the situations  where the user can't find the program, period;


andrewsignori-aot (approved):

dheepak-aot (approved):
LGTM. Nice catch 👍

PR #1711, opened 2023-02-06
#1676 - Condition to display course only with Part Time

andrepestana-aot:
- added condition to display course details when offering equals to Part Time.
image

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/programinformationrequest.json
+++ b/sources/packages/forms/src/form-definitions/programinformationrequest.json
@@ -1049,12 +1049,12 @@
           "tags": [],
           "properties": {},
           "conditional": {
-            "show": "",
-            "when": null,
-            "eq": "",
+            "show": "true",
+            "when": "isSelectedOfferingPartTime",
+            "eq": "true",
             "json": ""
           },
-          "customConditional": "show = data.offeringIntensitySelectedByStudent === 'Part Time'",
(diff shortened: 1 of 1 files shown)


sources/packages/forms/src/form-definitions/programinformationrequest.json
```diff
@@ -1049,12 +1049,12 @@
...
           "properties": {},
           "conditional": {
-            "show": null,
+            "show": "",
             "when": null,
             "eq": "",
             "json": ""
           },
-          "customConditional": "",
+          "customConditional": "show = data.offeringIntensitySelectedByStudent === 'Part Time'",
```
dheepak-aot, line 1057: Just a suggestion. We can have a hidden variable(isSelectedOfferingPartTime) which is calculated as `value = data.offeringIntensitySelectedByStudent === 'Part Time'` and use that.  It is just a suggestion not a blocker.

dheepak-aot (approved):
LGTM. 👍Add a suggestion. Not a blocker.

sources/packages/forms/src/form-definitions/programinformationrequest.json
andrepestana-aot: Ok. Done!

andrewsignori-aot (approved):
Looks good 👍

Issue #1693, opened 2023-01-27
Resume E2E/Unis Tests Efforts


andrewsignori-aot:
Decide the first effort to have E2E/Units tests added to the solution.

#### Some meetings outputs
- Review current test strategies from Nestjs documentation.
- Focus on E2E tests to allow more coverage with less effort at the beginning.
- Enable the code coverage and create a practice for new code changes do not drop the current coverage.
- Create unit tests as needed unless there is already some coverage from E2E tests.

#### Suggested next steps
- Review E2E/Unit tests for API/Workers/Queue Consumers.
- Enable all tests to be executed on CI/CD.
- Start to add E2E to API as a starting point.
- Enable the code coverage on CI/CD.
- Investigate how to enable Camunda tests.

PR #1716, opened 2023-02-08
#1693-E2E tests cleanup and local ad-hoc dev fixes

andrewsignori-aot:
- Fixed `make forms`;
- Created `local-all` make command to give an idea of what needed to be executed to have a complete environment up and running;
- Fixed form.io address inside `docker-compose` that was causing the API to not reach the `formio` service;
- Fixed swagger API URL that was including `api` twice in the address;
- Form.io user and pass for deploying definitions is now using the **main .env** file.


sources/packages/backend/apps/api/src/main.ts
```diff
@@ -69,7 +69,6 @@ async function bootstrap() {
         },
         "access-token",
       )
-      .addServer("/api")
```
dheepak-aot, line 72: Adding screenshot to explain if anyone having doubts on this. 

image

dheepak-aot (approved):
Thanks for taking care of form.io as well. Nice work. 👍

sources/packages/backend/apps/api/src/main.ts
andrewsignori-aot: Thanks @dheepak-aot, please see below also the double `api` in the URL that was causing all calls to fail.
image

sources/Makefile
```diff
@@ -62,12 +62,16 @@ export CRA_ENVIRONMENT_CODE := $(or $(CRA_ENVIRONMENT_CODE), )
 #Institution Integration
 export INSTITUTION_REQUEST_FOLDER := $(or $(INSTITUTION_REQUEST_FOLDER), Institution-Request)
 
+# Starts all applications: SIMS application (Web Portals, SIMS Api, Workers, Queue Consumers), 
+# and also Form.io server, and Camunda releated services.
+# Use make deploy-camunda-definitions to deploy all workflow definitions.
+# Use make deploy-form-definitions to deploy all form.io definitions.
+local-all: | print-status forms camunda local-build-run
```
andrepestana-aot, line 69: 👍

andrepestana-aot (approved):
Looks good!

Issue #1714, opened 2023-02-07
COE and Tuition Remittance - UI/Validations Changes


andrewsignori-aot:
The tuition remittance calculated at the time of ECERT generation must be based on the effective value of the disbursement. In other words, the dollar figure should have any overaward deduction, maximum limits hit, or restrictions applied.

However, at the time of tuition remittance request, the validation is calculated based on the potential award amounts only, WITHOUT considering any overaward deductions, maximum limits hit, or restrictions applied.

We will need to display their maximum allowed remittance value to the institution in the modal (if they select "Yes" to tuition remittance) and if the student has an overaward balance, we will need to display a banner warning the institution that there may be potential deductions at the time of disbursement that will reduce their remittance amount.

The final UI will be defined. Right now the API must return the maximum tuition remittance value allowed and also if the student has any overaward value in his balance greater than 0(zero).

**Acceptance Criteria**

- [ ] Change the current tuition remittance validation to use the estimated disbursement for its calculations. The maximum will be the lesser of the below calculations:
    - Tuition + program costs (calculation already in place);
    - (`Canada Loan` - `Canada Loan already paid`) + (`Canada Grant` - `Canada Grant already paid`) + (`BC Loan` - `BC Loan already paid`). "already paid" will be present only for reassessments (current implementation is not considering the "already paid" value, this is the main change for this calculation).
- [ ] Only when the COE is being confirmed and the institution is requesting a tuition remittance, display an info banner with the maximum remittance allowed for the COE: calculated using the same formula that is used in the field validation of the remittance request modal. Use the temporary content as "Maximum estimated tuition remittance {value}.".
- [ ] Only when the COE is being confirmed and the institution is requesting a tuition remittance, If the student has an overaward balance, display a banner warning institution that their actual disbursed remittance may be lower than their request. Use the temporary content as "The student has some overaward balance that can affect the tuition remittance requested during the disbursement. If this happens, the tuition remittance will be adjusted accordingly.".

PR #1731, opened 2023-02-11
#1714 - COE Tuition Remittance Request Changes

andrewsignori-aot:
- Created a centralized method to calculate the tuition remittance to be later used during e-Cert generation.
- Added the `max tuition remittance` and `has overawards` information to the COE confirmation payload.
- Changed the modal for COE confirmation to display the max tuition and a warning banner in case the student has an overaward.
- Added the max tuition as a max validation constraint on the UI.
- The UI and content was reviewed and adjusted by @hellolynn-tbtb.

image

image

image


sources/packages/backend/libs/services/src/confirmation-of-enrollment/confirmation-of-enrollment.service.ts
```diff
@@ -0,0 +1,114 @@
...
+      );
+    } else {
+      // Estimate the max tuition base in what is known before disbursement (e-Cert generation).
+      totalAwards = tuitionAwards.reduce(
+        (total, award) =>
+          total + award.valueAmount - (award.disbursedAmountSubtracted ?? 0),
+        0,
+      );
+    }
+    const OfferingTotalCosts =
```
dheepak-aot, line 110: variable name starting with uppercase.

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -1,35 +1,55 @@
...
+   * @param studentId student to be checked.
+   * @returns true if the student has any positive overaward balance value
+   * for any award, otherwise, false.
+   */
+  async hasOverawardBalance(studentId: number): Promise<boolean> {
+    const overawards = await this.getOverawardBalance([studentId]);
+    if (!overawards[studentId]) {
+      return false;
+    }
+    return Object.values(overawards[studentId]).some(
```
dheepak-aot, line 31: This is an effective balance right. Consider this scenario where student has an overaward and they pay the overaward back.

```
{
CSLF:100  //Re-assessment overaward
CSLF:-100 //Manual entry
}

```
The logic here will return true I believe. We must use reduce here isn't it?

Let me know If I am missing something.

sources/packages/backend/libs/services/src/confirmation-of-enrollment/confirmation-of-enrollment.service.ts
```diff
@@ -0,0 +1,114 @@
...
+export class ConfirmationOfEnrollmentService {
+  constructor(
+    @InjectRepository(DisbursementSchedule)
+    private readonly disbursementScheduleRepo: Repository<DisbursementSchedule>,
+  ) {}
+
+  /**
+   * Get the maximum estimated tuition remittance for the provided disbursement.
+   * @param disbursementId disbursement id.
+   * @returns maximum estimated tuition remittance if any, otherwise
```
dheepak-aot, line 30: I guess the comment for the returns is not complete.

sources/packages/backend/libs/services/src/confirmation-of-enrollment/_tests_/unit/confirmation-of-enrollment.getMaxTuitionRemittance.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+describe("ConfirmationOfEnrollmentService-getMaxTuitionRemittance", () => {
+  let service: ConfirmationOfEnrollmentService;
+
+  beforeAll(() => {
+    // disbursementScheduleRepo is not a dependency for getMaxTuitionRemittance.
+    const disbursementScheduleRepo = {} as Repository<DisbursementSchedule>;
+    service = new ConfirmationOfEnrollmentService(disbursementScheduleRepo);
+  });
+
+  it("Should calculate total estimated value from awards when awards are lesser than offering costs", () => {
```
dheepak-aot, line 15: Great job on covering all the scenrios.

dheepak-aot:
Nice work on centralizing the maximum tuition remittance and adding the banners and rules. 👍

sources/packages/web/src/composables/useRules.ts
```diff
@@ -38,18 +38,70 @@ export function useRules() {
...
+   * Check if a value is truthy.
+   * All values are truthy except false, 0, -0, 0n,
+   * "", null, undefined, and NaN.
+   * @param value value to be checked.
+   * @param fieldName friendly field name to be added
+   * to the validation message.
+   * @returns true if the value is truthy,
+   * otherwise a validation message.
+   */
+  const truthyRule = (value: string, fieldName: string) => {
```
dheepak-aot, line 51: We have an existing rule already

image

sources/packages/web/src/composables/useRules.ts
```diff
@@ -38,18 +38,70 @@ export function useRules() {
...
+    const numberValue = parseFloat(value);
+    if (numberValue < min || numberValue > max) {
+      let minText = min.toString();
+      let maxText = max.toString();
+      if (numberFormatter) {
+        minText = numberFormatter(minText);
+        maxText = numberFormatter(maxText);
+      }
+      return `${
+        fieldName ?? "Value"
```
dheepak-aot, line 92: Why are we performing null check on fieldname? I see that it is required in the method.

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrewsignori-aot: The method `getOverawardBalance` returns the SUM for every award. The "reduce", as you mentioned, happens on DB.
The format of the result will be something as below. Please let me know if further clarification is needed.
```ts
/**
 * Total overawards balances per awards indexed per student id.
 * @example considering 123 and 456 as students ids.
 * 123: {
 *    CSLF: 659,
 *    BCSL: 6565,
 * },
 * 456: {
 *    CSLF: 9898,
 * },
 */
```

sources/packages/web/src/composables/useRules.ts
andrewsignori-aot: My intention was to leave it optional but I forgot that the same message goes to the top of the form, so not including the field name would be confusing to the user. I will be making it required.

sources/packages/backend/libs/services/src/confirmation-of-enrollment/models/confirmation-of-enrollment.models.ts
```diff
@@ -0,0 +1,35 @@
+import { DisbursementValue, EducationProgramOffering } from "@sims/sims-db";
+
+/**
+ * Award information needed for the max tuition remittance calculation.
+ */
+export type Award = Pick<
```
andrepestana-aot, line 6: Why using Pick?

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrepestana-aot: @andrewsignori-aot do you mean that the value doesn't come to JS?

andrepestana-aot (commented):
Great job. Left some comments.

sources/packages/backend/libs/services/src/confirmation-of-enrollment/models/confirmation-of-enrollment.models.ts
andrewsignori-aot: It is to create a subset from the DB entity model to express precisely what this specific method requires to be executed.
Besides that, if the entity model ever changes this will also follow whatever type is changed on the Pick type.
For instance, the offering entity model has way more properties but we are using only two, for now. We can create a new object with only two properties or just reuse the entity model one.
I am not always looking forward to using it (Pick) but this calculation, as far as we know, will always be receiving the DB entities models as parameters.

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrewsignori-aot: Yes, for instance, for Part-Time Canada Loan (CSLF) there will always none to one records and never more than on for the particular student due to the `.groupBy("student.id").addGroupBy("disbursementOveraward.disbursementValueCode")` on `getOverawardBalance`

sources/packages/backend/libs/services/src/confirmation-of-enrollment/models/confirmation-of-enrollment.models.ts
andrepestana-aot: ok. Thanks for the explanation!

sources/packages/web/src/composables/useRules.ts
andrewsignori-aot: Adpted the current one to accept also `number` as the value that is the type that came from the input defined as a number and added some comments.

dheepak-aot: thanks for the modification to current method

dheepak-aot (approved):
Thanks for making the changes @andrewsignori-aot 👍

andrepestana-aot (approved):
👍

PR #1733, opened 2023-02-13
#1693 - Removing env from tests workflow

andrewsignori-aot:
When GitHub `dependabot` is creating the PR it does not have access to the env and an error happens "Error: Resource not accessible by integration". For instance: https://github.com/bcgov/SIMS/actions/runs/4130521677/jobs/7137296569
Using the closed ticket #1693 for reference only.

This workflow is executed during pull requests triggered by a user or by the GitHub `dependabot`.
When created by the `dependabot` the access to the env variables is restricted, please see https://github.com/dependabot/dependabot-core/issues/3253 and
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions
To allow access to the secrets in both cases, the test users' passwords (or any other secret) are defined at the repository
level and also at the `dependabot` secrets for now.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,7 +5,6 @@ on: [pull_request]
 jobs:
   api-e2e-tests:
     name: SIMS API E2E Tests
-    environment: DEV
     runs-on: ubuntu-latest
     env:
       PROJECT_NAME: aest-sims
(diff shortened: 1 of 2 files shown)


andrepestana-aot (approved):

dheepak-aot (approved):
👍

Issue #1677, opened 2023-01-23
Manage overaward values imported from SFAS


JasonCTang:
When a new student account is created, the system must check if the individual has some overawards values on the "SFAS imported data" to then have the same values inserted in the SIMS overaward balance.
Also, due to the fact the SFAS and SIMS will possibly be generating overawards in parallel (during some point in time), any new overaward received during the SFAS data import ([SFAS to SIMS bridge](https://bcgov.sharepoint.com/:w:/r/teams/03177/Shared%20Documents/DEVS/SFAS%20Integration/SFAS%20to%20SIMS%20Bridge.docx?d=w091d09fd474a4fbbbd2e84e6b22590d1&csf=1&web=1&e=4tr5iT)) must also be inserted or updated on SIM.

Two scenarios: 
1. If a student account does not exist yet in SIMS: when a student creates an account in SIMS, needs to check for pre-existing overaward values from SFAS data and create records for that student in SIMS overaward balance tables. 
2. If a student account already exists in SIMS: if there is a new overaward value imported from SFAS, update in SIMS overaward balance for that student.

SFAS imported data
Students(individuals) must always be searched by the combination of SIN, last name, and DOB (reference [sfas-individual.service.ts](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts#L24))
Table: `sims.sfas_individuals`
Canada Loan **Full-Time** - CSLF: column `csl_overaward`
BC Loan **Full-Time** - BCSL - BCSL: column `bcsl_overaward`
_There is no overaward data available for **Part-Time** applications._

**Acceptance Criteria**
- [ ] During student account creation check if SFAS imported data have any value greater than 0(zero) and create the new entries on `sims.disbursement_overawards` using the  "Legacy" origin_type and the CSLF or BCSL.
- [ ] During the SFAS import data, check if the imported data has any overaward value greater than 0(zero) and insert the records on the `sims.disbursement_overawards` or update the records in case they are already there. The checks for insert/update must happen independently for CSLF/BCSL since one award can be present or maybe neither could be present (reference file [sfas-individual.service.ts](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts#L26)).

PR #1739, opened 2023-02-14
#1677 - Manage overaward values imported from SFAS

andrepestana-aot:
- Disbursement overaward update when creating student profile.
- Disbursement overaward update when SFAS Integration process happens.
  - Added student_id to sims.sfas_individuals table.


sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -76,6 +78,32 @@ export class DisbursementOverawardService {
...
+  async addLegacyOveraward(
+    studentId: number,
+    overawardValue: number,
+    disbursementValueCode: string,
+    entityManager?: EntityManager,
+  ): Promise<void> {
+    const repo =
+      entityManager?.getRepository(DisbursementOveraward) ??
+      this.disbursementOverawardRepo;
+    const auditUser = await this.systemUsersService.systemUser();
```
andrewsignori-aot, line 97: The `auditUser` in this situation should probably be received as a parameter.
For instance, this method is used by the API to update the `LegacyOveraward` and the student user is the one to be saved as a `creator`.

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -76,6 +78,32 @@ export class DisbursementOverawardService {
...
+    await repo.insert({
+      student: { id: studentId } as Student,
+      disbursementValueCode,
+      overawardValue,
+      originType: DisbursementOverawardOriginType.LegacyOveraward,
+      creator: auditUser,
+    } as DisbursementOveraward);
```
andrewsignori-aot, line 104: IMO, the cast (`as DisbursementOveraward`) is not bringing any benefit in this scenario.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +572,40 @@ export class StudentService extends RecordDataModelService<Student> {
...
+    entityManager: EntityManager,
+  ) {
+    const sfasIndividual = await this.sfasIndividualService.getSFASOverawards(
+      student.user.lastName,
+      student.birthDate,
+      sinNumber,
+    );
+
+    if (sfasIndividual?.bcslOveraward > 0) {
+      this.disbursementOverawardService.addLegacyOveraward(
```
andrewsignori-aot, line 595: Missing `await`.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +572,40 @@ export class StudentService extends RecordDataModelService<Student> {
...
+    sinNumber: string,
+    entityManager: EntityManager,
+  ) {
+    const sfasIndividual = await this.sfasIndividualService.getSFASOverawards(
+      student.user.lastName,
+      student.birthDate,
+      sinNumber,
+    );
+
+    if (sfasIndividual?.bcslOveraward > 0) {
```
andrewsignori-aot, line 594: May be not an issue here because it will always be 0 to 2 records to be added, but just keep in mind that the objects records can be created and provided to Typeorm as an array also 😉

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +572,40 @@ export class StudentService extends RecordDataModelService<Student> {
...
+  /**
+   * Search for the student's SFAS data and update overaward values in
+   * disbursement overawards table if any overaward values for BCSL or CSLF exist.
+   * @param student student object.
+   * @param sinNumber student's sin number.
+   * @param entityManager entityManager to be used to perform the query.
+   */
+  async updateSFASOveraward(
+    student: Student,
+    sinNumber: string,
```
andrewsignori-aot, line 585: Really minor, but naming `sinNumber` is the same as naming it socialInsuranceNumberNumber 😄

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +572,40 @@ export class StudentService extends RecordDataModelService<Student> {
...
   logger: LoggerService;
+
+  /**
+   * Search for the student's SFAS data and update overaward values in
+   * disbursement overawards table if any overaward values for BCSL or CSLF exist.
+   * @param student student object.
+   * @param sinNumber student's sin number.
+   * @param entityManager entityManager to be used to perform the query.
+   */
+  async updateSFASOveraward(
```
andrewsignori-aot, line 583: The name gives the impression, IMO, that we are updating something on SFAS. I would recommend something like `importSFASOverawards`.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -28,12 +28,18 @@ import {
   STUDENT_ACCOUNT_CREATION_MULTIPLES_SIN_FOUND,
   STUDENT_SIN_CONSENT_NOT_CHECKED,
 } from "../../constants";
+import { DisbursementOverawardService } from "@sims/services";
+import {
+  BC_STUDENT_LOAN_AWARD_CODE,
+  CANADA_STUDENT_LOAN_FULL_TIME_AWARD_CODE,
+} from "@sims/services/constants/disbursements.constants";
```
andrewsignori-aot, line 35: `disbursements.constants` is part of the index file also.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: Sorry but I didn't get what you mean by "objects records can be created and provided to Typeorm as an array".

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: Yes, I know that but SIN Number how it is known. Maybe people call it "sin number" to distinguish from the religious sin. 😄

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: For instance:
```ts
await applicationRepository.insert([application, newApplication]);
```
or
```ts
await applicationRepository.save([application, newApplication]);
```

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -180,6 +188,9 @@ export class StudentService extends RecordDataModelService<Student> {
         student.sinValidation = sinValidation;
         await entityManager.getRepository(Student).save(student);
       }
+
+      await this.importSFASOverawards(student, studentSIN, entityManager);
```
andrewsignori-aot, line 192: For this scenario, I would prefer that the method parameters are restricted to what is really needed, in this case, only `lastName` and `birthDate` instead of the entire student object.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE sims.sfas_individuals
+    ADD COLUMN IF NOT EXISTS student_id INTEGER NULL;
```
andrewsignori-aot, line 2: Please add a DB comment to the column also.
I am assuming that there is no integrity check for performance reasons and because it is supposed to be used only during the import part.
If it is true, it can be mentioned in the DB comment also.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Remove-student-id-sfas-individuals.sql
```diff
@@ -0,0 +1,3 @@
+
```
andrewsignori-aot, line 1: Please remove the blank line.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
```diff
@@ -55,4 +83,20 @@ export class SFASIndividualService
     individual.extractedAt = getUTC(extractedDate);
     await this.repo.save(individual, { reload: false, transaction: false });
   }
+
+  async updateStudentId(): Promise<void> {
```
andrewsignori-aot, line 87: Please add a blank line between the methods and comments to all of them.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
```diff
@@ -0,0 +1,30 @@
+/*
```
andrewsignori-aot, line 1: Please install the `SQL Formatter` mentioned on the wiki and have all the SQL files formatted. It will not always be the format that everyone likes but at least will be the same.
https://github.com/bcgov/SIMS/wiki/Development---Standards,-Decisions,-and-Troubleshooting#plugins-to-be-installed

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
```diff
@@ -0,0 +1,30 @@
...
+ * the overard is different than 0. 
+ */
+INSERT INTO sims.disbursement_overawards(
+		student_id,
+		disbursement_value_code,
+		overaward_value,
+		origin_type,
+		creator) 
+SELECT	sfas_individuals.student_id, 
+	   	E'BCSL',
```
andrewsignori-aot, line 12: Can you please elaborate on why we need the `E` on lines 12 and 14?

andrewsignori-aot, line 12: Is there a possibility to provide the query `BCSL` as a parameter and reuse the same query for both awards?
If so, maybe `system-user` and `Legacy overaward` can also be provided as parameters.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
```diff
@@ -0,0 +1,30 @@
+/*
+ * Insert BCSL overawards from SFAS integration to disbursement overawards table where 
+ * the overard is different than 0.
```
andrewsignori-aot, line 3: Typo on "overard".

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-update-bcsl-disbursement-overaward.sql
```diff
@@ -0,0 +1,12 @@
...
+ */
+UPDATE	sims.disbursement_overawards disbursement_overawards
+SET		overaward_value = sfas_individuals.bcsl_overaward, 
+		updated_at = NOW(),
+		modifier = users.id
+FROM 	sims.sfas_individuals sfas_individuals
+		INNER JOIN sims.users users ON users.last_name = 'system-user'
+WHERE 	disbursement_overawards.student_id = sfas_individuals.student_id
+AND		disbursement_overawards.origin_type = 'Legacy overaward'
+AND		disbursement_overawards.disbursement_value_code = 'BCSL';
```
andrewsignori-aot, line 12: We can add also add a clause to avoid updating the value if the value is the same.
The sfas_individuals can be updated multiple times due to other columns changes.

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
```diff
@@ -38,4 +38,34 @@ export class SFASIndividualService extends DataModelService<SFASIndividual> {
...
+    sin: string,
+  ): Promise<SFASIndividual> {
+    const individual = await this.repo.findOne({
+      select: {
+        id: true,
+        cslOveraward: true,
+        bcslOveraward: true,
+      },
+      where: {
+        lastName: Raw((alias) => `LOWER(${alias}) = LOWER(:lastName)`, {
```
andrewsignori-aot, line 61: I liked the solution using the `Raw` to workaround the object query limitations ❤️

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
```diff
@@ -38,4 +38,34 @@ export class SFASIndividualService extends DataModelService<SFASIndividual> {
...
+   * @param birthDate student's birth date.
+   * @param sin student's sin.
+   * @returns SFAS individual details.
+   */
+  async getSFASOverawards(
+    lastName: string,
+    birthDate: string,
+    sin: string,
+  ): Promise<SFASIndividual> {
+    const individual = await this.repo.findOne({
```
andrewsignori-aot, line 54: It can be simplified to
```ts
return this.repo.findOne({
```

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -76,6 +76,32 @@ export class DisbursementOverawardService {
...
+   * @param studentId id of the student.
+   * @param overawardValue overaward value.
+   * @param disbursementValueCode disbursement value code.
+   * @param entityManager entity manager used to perform the query.
+   */
+  async addLegacyOveraward(
+    studentId: number,
+    overawardValue: number,
+    disbursementValueCode: string,
+    userId: number,
```
andrewsignori-aot, line 90: As a recommendation, searching the code, most of the time the audit user is declared as below. It helps to differentiate from other moments where the `userId` is not provided with the audit intention.
```ts
* @param auditUserId user that should be considered the one that is causing the changes.
```

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -142,6 +142,11 @@ export class SFASIntegrationProcessingService {
       // Delete the file only if it was processed with success.
       if (result.success) {
         try {
+          await this.sfasIndividualService.updateStudentId();
+          await this.sfasIndividualService.updateBCSLDisbursementOveraward();
```
andrewsignori-aot, line 146: I believe that at least the updates and inserts can be grouped and executed in parallel. Thoughts?

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -142,6 +142,11 @@ export class SFASIntegrationProcessingService {
       // Delete the file only if it was processed with success.
       if (result.success) {
         try {
+          await this.sfasIndividualService.updateStudentId();
```
andrewsignori-aot, line 145: This operation must happen outside the try/catch for the delete file.
For instance, if there is an error, the log message will say `Error while deleting SFAS integration file`, unless I am missing something.

andrewsignori-aot, line 145: These methods can be grouped inside the `sfas-individual.service.ts`.

andrewsignori-aot (commented):
Just double checking, were you able to confirm that the student creation using the basic BCeID will end up calling the same student creation method and updating the overawards, right?

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +574,43 @@ export class StudentService extends RecordDataModelService<Student> {
...
+   * disbursement overawards table if any overaward values for BCSL or CSLF exist.
+   * @param student student object.
+   * @param sinNumber student's sin number.
+   * @param entityManager entityManager to be used to perform the query.
+   */
+  async importSFASOverawards(
+    student: Student,
+    sinNumber: string,
+    entityManager: EntityManager,
+  ) {
```
ann-aot, line 589: we can add `): Promise<void> {`

sources/packages/backend/apps/api/src/services/student/student.service.ts
ann-aot: I didn't understand the context

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -76,6 +76,32 @@ export class DisbursementOverawardService {
     return result;
   }
 
+  /**
+   * Adds a legacy overaward value to disbursement overawards table.
+   * @param studentId id of the student.
+   * @param overawardValue overaward value.
+   * @param disbursementValueCode disbursement value code.
+   * @param entityManager entity manager used to perform the query.
```
ann-aot, line 84: comment doesn't match with the param

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-update-bcsl-disbursement-overaward.sql
```diff
@@ -0,0 +1,12 @@
+/*
+ * Update BCSL overawards from SFAS integration to disbursement overawards table.
+ */
+UPDATE	sims.disbursement_overawards disbursement_overawards
+SET		overaward_value = sfas_individuals.bcsl_overaward, 
+		updated_at = NOW(),
```
ann-aot, line 6: `updated_at` is the inbuilt audit filed right? I think the framework is taking care of it

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +574,43 @@ export class StudentService extends RecordDataModelService<Student> {
 
   @InjectLogger()
   logger: LoggerService;
+
+  /**
+   * Search for the student's SFAS data and update overaward values in
+   * disbursement overawards table if any overaward values for BCSL or CSLF exist.
+   * @param student student object.
```
dheepak-aot, line 581: We can try to avoid words like object in the comment description.

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
```diff
@@ -38,4 +38,34 @@ export class SFASIndividualService extends DataModelService<SFASIndividual> {
...
+   * @param lastName student's last name.
+   * @param birthDate student's birth date.
+   * @param sin student's sin.
+   * @returns SFAS individual details.
+   */
+  async getSFASOverawards(
+    lastName: string,
+    birthDate: string,
+    sin: string,
+  ): Promise<SFASIndividual> {
```
dheepak-aot, line 53: Just for my understanding. I assume there will be only one record per student in sfas_individuals right?

sources/packages/backend/apps/api/src/services/student/student.service.ts
dheepak-aot: I agree @andrewsignori-aot

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
dheepak-aot: So the idea of not having the foreign key is for performance right? If that is the case, how costly is the difference?

Not a part of this PR. I believe that nullable columns doesn't need to be explicitly mentioned with NULL(I guess we are doing that in our application). We can ignore the explicit mentioning @ann-aot  @andrewsignori-aot ?

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
andrepestana-aot: Sorry. No need to include the "E" here. I was having an issue and the "E" worked. It's used to escape in Postgres.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
```diff
@@ -14,8 +17,33 @@ export class SFASIndividualService
   extends DataModelService<SFASIndividual>
   implements SFASDataImporter
 {
+  private readonly bulkUpdateStudentIdSQL: string;
+  private readonly bulkUpdateBCSLDisbursementOverawardSQL: string;
+  private readonly bulkUpdateCSLFDisbursementOverawardSQL: string;
+  private readonly bulkInsertBCSLDisbursementOverawardSQL: string;
+  private readonly bulkInsertCSLFDisbursementOverawardSQL: string;
```
dheepak-aot, line 24: I know CSLP is a part time overaward and current sfas integration may not have it. 
But looking at the column(in sfas_individuals) it seems to be very generic (csl_overaward). Is there a possibility of part time overaward coming through this integration?

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
andrewsignori-aot: @dheepak-aot, yes, the `NULL` does not need to be explicit and we have been just omitting it usually.
About adding the foreign, we can add it.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
andrepestana-aot: It's possible if we create the SQL file as a template:
`AND disbursement_overawards.disbursement_value_code = ${disbursementValueCode};`
Is that what you mean?

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-update-bcsl-disbursement-overaward.sql
andrewsignori-aot: @ann-aot, this query is not using an entity model, which means that TypeOrm has no way to take care of it 😉

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
andrewsignori-aot: Yes @dheepak-aot, that is the assumption. At least the combination of sin/last name/DOB should be unique.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-cslf-disbursement-overaward.sql
```diff
@@ -0,0 +1,31 @@
...
+		INNER JOIN sims.students students ON sfas_individuals.student_id = students.id
+		INNER JOIN sims.users users ON users.last_name = 'system-user'
+WHERE 	NOT EXISTS (
+		SELECT 	1 
+		FROM 	sims.disbursement_overawards disbursement_overawards
+		WHERE 	disbursement_overawards.student_id = students.id
+		AND 	disbursement_overawards.disbursement_value_code = 'CSLF'
+		AND 	disbursement_overawards.origin_type = 'Legacy overaward'
+)
+AND		sfas_individuals.csl_overaward <> 0;;
```
dheepak-aot, line 26: semi-colon repeated.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
andrewsignori-aot: No, there is no possibility of part-time coming in the integration because there is no concept of a part-time overaward on SAIL. Part-time overaward exists only on SIMS.

sources/packages/backend/libs/src-sql/sfas-individuals/Bulk-update-students-foreign-key.sql
```diff
@@ -0,0 +1,18 @@
...
+SET
+    student_id = students.id
+FROM
+    sims.students students
+    INNER JOIN sims.users users ON students.user_id = users.id
+    INNER JOIN sims.sin_validations sin_validations ON sin_validations.id = students.sin_validation_id
+WHERE
+    students.birth_date = sims.sfas_individuals.birth_date
+    AND sin_validations.sin = sims.sfas_individuals.sin
+    AND lower(users.last_name) = lower(sims.sfas_individuals.last_name)
```
dheepak-aot, line 18: semi-colon missing at the end.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
dheepak-aot: That would be perfect.

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
dheepak-aot: Also I would suggest to use the name as auditUserId instead of userId.

dheepak-aot:
Great work @andrepestana-aot. Please have a look at the comments.

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrepestana-aot: in the other [comment](https://github.com/bcgov/SIMS/pull/1739#discussion_r1116358148) I understood it should be 
any userId here because this method may be used to add legacy overaward using other users, not only audit users. Did I get it wrong?

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
andrepestana-aot: Yes, the 2 updates and 2 inserts to be executed in parallel works.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
andrewsignori-aot: Not really, if I am not wrong we can use real DB parameters.
```ts
repository.query(`select * from users where LOWER(first_name) = $1 OR LOWER(last_name) = $2`, [ first_name.toLocaleLowerCase(), last_name.toLocaleLowerCase() ])
```

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrewsignori-aot: @andrepestana-aot the mentioned comment states to receive the user as a parameter and not use the "system user". On top of that name this user as `auditUserId` because it is a user and it is used for the sole intention of auditing.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-cslf-disbursement-overaward.sql
```diff
@@ -0,0 +1,31 @@
...
+		origin_type,
+		creator) 
+SELECT	sfas_individuals.student_id, 
+	   	E'CSLF', 
+	   	sfas_individuals.csl_overaward, 
+		E'Legacy overaward',
+		users.id
+FROM 	sims.sfas_individuals sfas_individuals
+		INNER JOIN sims.students students ON sfas_individuals.student_id = students.id
+		INNER JOIN sims.users users ON users.last_name = 'system-user'
```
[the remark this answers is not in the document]
andrewsignori-aot: We can use the user_name instead which is a const `8fb44f70-6ce6-11ed-b307-8743a2da47ef@system`.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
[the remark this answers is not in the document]
andrepestana-aot: As this is used only for this import of the overawards and updated after the SFAS import I think it is ok not have the check to improve the performance in this process.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-update-bcsl-disbursement-overaward.sql
andrepestana-aot: Not in this case. I did some tests and for raw query that doesn't work.

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
andrepestana-aot: Yes, that is what we expect from SFAS.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
andrepestana-aot: As far as I know, that is not expected to happen for now.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
andrepestana-aot: I think I can create the foreign key and if we have any issue with performance we can remove it then.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: ok. Saving the array now.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
andrepestana-aot: ok. Removed the NULL and added the foreign key. Thanks!

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-bcsl-disbursement-overaward.sql
andrepestana-aot: ok. Added parameters. The SQL Formatter doesn't work well with parameters but it's just turn it off when changing this file.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -139,8 +139,18 @@ export class SFASIntegrationProcessingService {
       await Promise.all(promises);
       this.logger.log("Records imported.");
 
-      // Delete the file only if it was processed with success.
       if (result.success) {
+        try {
+          await this.sfasIndividualService.updateDisbursementOverawards();
+        } catch (error) {
+          const logMessage = `Error while updating SFAS overawards.`;
```
ann-aot, line 146: pls use double quotes

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrewsignori-aot: I believe that we are not on the same page on this one or I am missing something.
While saving the overawards I can see two scenarios:
1 - The Student is creating its account, then the student is the one causing the action, then the Student must be the user saved as `creator`.
2 - The overawards are imported from SFAS, and there is no user authenticated, then we consider this a system operation, then the `creator` will be defined as the `system-user`.
```ts
/**
   * Search for the student's SFAS data and update overaward values in
   * disbursement overawards table if any overaward values for BCSL or CSLF exist.
   * @param studentId student id for the disbursement overaward record.
   * @param lastName last name for sfas individuals.
   * @param sinNumber student's sin number.
   * @param entityManager entityManager to be used to perform the query.
   * @param auditUserId user that should be considered the one that is causing the changes.
   */
  async importSFASOverawards(
    studentId: number,
    lastName: string,
    birthDate: string,
    sinNumber: string,
    entityManager: EntityManager,
    auditUserId: number,
  ): Promise<void> {
    const sfasIndividual = await this.sfasIndividualService.getSFASOverawards(
      lastName,
      birthDate,
      sinNumber,
    );
    const overawards: DisbursementOveraward[] = [];
    if (sfasIndividual?.bcslOveraward > 0) {
      overawards.push({
        student: { id: studentId } as Student,
        disbursementValueCode: BC_STUDENT_LOAN_AWARD_CODE,
        overawardValue: sfasIndividual.bcslOveraward,
        creator: { id: auditUserId } as User,
      } as DisbursementOveraward);
    }
    if (sfasIndividual?.cslOveraward > 0) {
      overawards.push({
        student: { id: studentId } as Student,
        disbursementValueCode: CANADA_STUDENT_LOAN_FULL_TIME_AWARD_CODE,
        overawardValue: sfasIndividual.cslOveraward,
        creator: { id: auditUserId },
      } as DisbursementOveraward);
    }
    await this.disbursementOverawardService.addLegacyOverawards(
      overawards,
      entityManager,
    );
  }
```

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -76,6 +76,24 @@ export class DisbursementOverawardService {
     return result;
   }
 
+  /**
+   * Adds a legacy overaward value to disbursement overawards table.
+   * @param studentId id of the student.
```
andrewsignori-aot, line 81: Please review the parameters in the comments.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE
+    sims.sfas_individuals
+ADD
+    COLUMN IF NOT EXISTS student_id INTEGER;
+
+COMMENT ON COLUMN sims.sfas_individuals.student_id IS 'Student id used to import data to sims.disbursement_overawards table.';
+
+ALTER TABLE
```
andrewsignori-aot, line 8: Please take a look at existing relationships, usually, we just add them alongside the column declaration. This will also enforce the proper naming conventions. The `fk_student_id_id` is not aligned with the existing ones.
```sql
ALTER TABLE
    sims.sfas_individuals
ADD
    COLUMN IF NOT EXISTS student_id INTEGER REFERENCES sims.students(id);
```

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
```diff
@@ -55,4 +85,79 @@ export class SFASIndividualService
...
+    auditUserName: string,
+  ): Promise<void> {
+    await this.repo.manager.query(this.bulkInsertDisbursementOverawardSQL, [
+      disbursementValueCode,
+      originType,
+      auditUserName,
+    ]);
+  }
+
+  async updateDisbursementOverawards(): Promise<void> {
```
andrewsignori-aot, line 126: Please add some comments.
If the other methods do not need to be called externally to this class they can be made `private`.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -139,8 +139,18 @@ export class SFASIntegrationProcessingService {
       await Promise.all(promises);
       this.logger.log("Records imported.");
 
-      // Delete the file only if it was processed with success.
       if (result.success) {
+        try {
+          await this.sfasIndividualService.updateDisbursementOverawards();
+        } catch (error) {
+          const logMessage = "Error while updating SFAS overawards.";
```
andrewsignori-aot, line 146: The message does not seem accurate. Can we change it to something like:
```
Error while importing SFAS overawards balances.
```
or
```
Error while updating overawards balances imported from SFAS.
```

sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
andrepestana-aot: ok. Sorry. Got it now.

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-disbursement-overaward.sql
```diff
@@ -0,0 +1,39 @@
...
+WHERE
+	NOT EXISTS (
+		SELECT
+			1
+		FROM
+			sims.disbursement_overawards disbursement_overawards
+		WHERE
+			disbursement_overawards.student_id = students.id
+			AND disbursement_overawards.disbursement_value_code = $1::text
+			AND disbursement_overawards.origin_type::text = $2::text
```
andrewsignori-aot, line 37: Can we please avoid the cast on the left side of the operation and convert only the right side?

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-insert-disbursement-overaward.sql
```diff
@@ -0,0 +1,39 @@
...
+	)
+	AND sfas_individuals.bcsl_overaward <> 0;
```
andrewsignori-aot, line 39: Does the `sfas_individuals.bcsl_overaward <> 0` needs to be adjusted also, right?

sources/packages/backend/libs/src-sql/disbursement-overawards/Bulk-update-disbursement-overaward.sql
```diff
@@ -0,0 +1,26 @@
...
+	END,
+	updated_at = NOW(),
+	modifier = users.id
+FROM
+	sims.sfas_individuals sfas_individuals
+	INNER JOIN sims.users users ON users.user_name = $1::text
+	AND users.first_name IS NULL
+WHERE
+	disbursement_overawards.student_id = sfas_individuals.student_id
+	AND disbursement_overawards.origin_type::text = $2::text
```
andrewsignori-aot, line 20: Same comment from https://github.com/bcgov/SIMS/pull/1739/files#r1120820189

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -180,6 +189,16 @@ export class StudentService extends RecordDataModelService<Student> {
         student.sinValidation = sinValidation;
         await entityManager.getRepository(Student).save(student);
       }
+
+      await this.importSFASOverawards(
+        student.id,
+        student.user.lastName,
+        student.birthDate,
+        studentSIN,
+        student.user.id,
```
andrewsignori-aot, line 198: The method has an `auditUserId`, please use it.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Add-student-id-sfas-individuals.sql
```diff
@@ -0,0 +1,8 @@
...
+    sims.sfas_individuals
+ADD
+    COLUMN IF NOT EXISTS student_id INTEGER REFERENCES sims.students(id);
+
+COMMENT ON COLUMN sims.sfas_individuals.student_id IS 'Student id used to import data to sims.disbursement_overawards table.';
```
andrewsignori-aot, line 6: The column can state what it is instead of the purpose because it was added.
Sugestion
```
Existing student id identified by the combination of SIN, last name, and date of birth.
```

andrewsignori-aot (commented):
Thanks for doing the changes, please take a look at the comments.

dheepak-aot (approved):
Thanks for doing the changes. 👍LGTM.

Please look at the comments from other devs.

andrepestana-aot:
> Great work, please take a look at the comments. Just double checking, were you able to confirm that the student creation using the basic BCeID will end up calling the same student creation method and updating the overawards, right?

Yes, I double checked that. It also end up calling `internalCreateStudent` method.

ann-aot (approved):
Good Work @andrepestana-aot 👍 Thanks for doing the changes

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: I am not able to see this change. Is it updated on the repo?

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -563,4 +583,51 @@ export class StudentService extends RecordDataModelService<Student> {
...
+    auditUserId: number,
+    entityManager: EntityManager,
+  ): Promise<void> {
+    const sfasIndividual = await this.sfasIndividualService.getSFASOverawards(
+      lastName,
+      birthDate,
+      sinNumber,
+    );
+    const overawards: DisbursementOveraward[] = [];
+    if (sfasIndividual?.bcslOveraward > 0) {
```
andrewsignori-aot, line 610: As per the conversation with @JasonCTang, let's allow the import to accept negative numbers also.
```ts
if (sfasIndividual?.bcslOveraward !== 0)
```
Sorry for the last-minute change and thanks for including it on this PR.

andrewsignori-aot (approved):
Thanks for doing the changes and for the great work, looks good 👍

Issue #1730, opened 2023-02-10
Upgrade Vuetify to Release Version


andrewsignori-aot:
- Upgrade Vuetify to most recent release version ([currently 3.1](https://github.com/vuetifyjs/vuetify/releases/tag/v3.1.0)).
- Check if the [v-data-table](https://next.vuetifyjs.com/en/api/v-data-table/) is fully functional and can replace the current PrimeVue component.

ann-aot:
1. implemented `v-datatable-server` for location program. 
2. search option is not available as of now, so we are using the existing implementation.
4. we are restricting `All` option in the pagination in the API.  -> planning to work on it in the feature tickets if the new release does not any custom pagination option.

PR #1754, opened 2023-02-23
#1730 Upgrade Vuetify to Release Version#1730

ann-aot:
- Upgrade Vuetify to 3.1.5.
- Imported `VDataTableServer` from "vuetify/labs/VDataTable".
- Replaced Location program datatate with <v-data-table-server
- Implemented <v-data-table-server default sort and pagination
- Search feature is not in the <v-data-table-server 
- There was a bug in the existing API, for page 0, it was returning the entire values,- fixed

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
@@ -24,9 +24,10 @@ abstract class PaginationOptionsAPIInDTO {
   page: number;
   /**
    * Page size or records per page.
+   * -1 means all records.
    */
-  @Min(1)
-  @Max(50)
+  @Min(-1)
+  @Max(100)
   pageLimit: number;
   /**
    * Criteria to be used to filter the records.
--- a/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
@@ -25,7 +25,7 @@ abstract class PaginationOptionsAPIInDTO {
   /**
    * Page size or records per page.
    */
-  @Min(0)
+  @Min(1)
   @Max(100)
   pageLimit: number;
   /**
--- a/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
@@ -24,9 +24,8 @@ abstract class PaginationOptionsAPIInDTO {
   page: number;
   /**
    * Page size or records per page.
-   * -1 means all records.
    */
-  @Min(-1)
+  @Min(0)
   @Max(100)
   pageLimit: number;
   /**
--- a/sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
+++ b/sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
@@ -268,7 +268,7 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
     // For getting total raw count before pagination.
     const sqlQuery = paginatedProgramQuery.getSql();
 
-    if (paginationOptions.pageLimit > 0) {
+    if (paginationOptions.pageLimit) {
       paginatedProgramQuery.limit(paginationOptions.pageLimit);
     }
 
--- a/sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
+++ b/sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
@@ -268,16 +268,14 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
     // For getting total raw count before pagination.
     const sqlQuery = paginatedProgramQuery.getSql();
 
-    if (paginationOptions.pageLimit) {
+    if (paginationOptions.pageLimit > 0) {
       paginatedProgramQuery.limit(paginationOptions.pageLimit);
     }
-    if (paginationOptions.page) {
-      paginatedProgramQuery.offset(
-        paginationOptions.page * paginationOptions.pageLimit,
-      );
-    } else {
-      paginatedProgramQuery.limit(0);
-    }
+
+    paginatedProgramQuery.offset(
+      paginationOptions.page * paginationOptions.pageLimit,
+    );
+
     // sort
     if (paginationOptions.sortField && paginationOptions.sortOrder) {
       paginatedProgramQuery.orderBy(
--- a/sources/packages/web/package.json
+++ b/sources/packages/web/package.json
@@ -32,7 +32,7 @@
     "vue": "^3.2.31",
     "vue-class-component": "^8.0.0-0",
     "vue-router": "^4.0.14",
-    "vuetify": "^3.0.0-beta.13",
+    "vuetify": "^3.1.5",
     "vuex": "^4.0.2",
     "yup": "^0.32.11"
   },
--- a/sources/packages/web/src/assets/css/base.scss
+++ b/sources/packages/web/src/assets/css/base.scss
@@ -378,3 +378,8 @@ b,
 strong {
   font-weight: $font-weight-bold;
 }
+
+// Added FontAwesome as fallback.
+.mdi:before {
+  font-family: "Material Design Icons", "FontAwesome" !important;
+}
--- a/sources/packages/web/src/assets/css/base.scss
+++ b/sources/packages/web/src/assets/css/base.scss
@@ -380,6 +380,6 @@ strong {
 }
 
 // Added FontAwesome as fallback.
-.mdi:before {
-  font-family: "Material Design Icons", "FontAwesome", "BCSans", sans-serif !important;
+.fas:before {
+  font-family: "FontAwesome", $font-bold !important;
 }
--- a/sources/packages/web/src/assets/css/base.scss
+++ b/sources/packages/web/src/assets/css/base.scss
@@ -381,5 +381,5 @@ strong {
 
 // Added FontAwesome as fallback.
 .mdi:before {
-  font-family: "Material Design Icons", "FontAwesome" !important;
+  font-family: "Material Design Icons", "FontAwesome", "BCSans", sans-serif !important;
 }
--- a/sources/packages/web/src/helpers/helpers/uri.ts
+++ b/sources/packages/web/src/helpers/helpers/uri.ts
@@ -7,7 +7,7 @@ import {
 } from "@/types";
 
 /**
- * todo: remove sortOrder: DataTableSortOrder when all primevue datatables are removed.
+ * @deprecated use getPaginationQueryString
  * Helper to append pagination sort and order to the url
  * @param url api url
  * @param sortField sortField
@@ -30,6 +30,7 @@ export function addSortOptions(
 }
 
 /**
+ * @deprecated use getPaginationQueryString
  * Utility to build the pagination query parameters.
  * @param url
  * @param page
@@ -56,7 +57,7 @@ export const getPaginationQueryString = (
   paginationOptions: PaginationOptions,
 ) => {
   const parameters: string[] = [];
-  // Paginations parameters.
+  // Pagination parameters.
   parameters.push(`${PaginationParams.Page}=${paginationOptions.page}`);
   parameters.push(
     `${PaginationParams.PageLimit}=${paginationOptions.pageLimit}`,
--- a/sources/packages/web/src/helpers/helpers/uri.ts
+++ b/sources/packages/web/src/helpers/helpers/uri.ts
@@ -55,7 +55,12 @@ export const addPaginationOptions = (
  */
 export const getPaginationQueryString = (
   paginationOptions: PaginationOptions,
+  enableZeroPage = false,
 ) => {
+  if (enableZeroPage) {
+    paginationOptions.page = paginationOptions.page - 1;
+  }
+
   const parameters: string[] = [];
   // Pagination parameters.
   parameters.push(`${PaginationParams.Page}=${paginationOptions.page}`);
--- a/sources/packages/web/src/helpers/helpers/uri.ts
+++ b/sources/packages/web/src/helpers/helpers/uri.ts
@@ -1,4 +1,5 @@
 import {
+  DataTableSortByOrder,
   DataTableSortOrder,
   FieldSortOrder,
   PaginationOptions,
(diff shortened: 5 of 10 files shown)


sources/packages/web/package.json
```diff
@@ -32,7 +32,7 @@
     "vue": "^3.2.31",
     "vue-class-component": "^8.0.0-0",
     "vue-router": "^4.0.14",
-    "vuetify": "^3.0.0-beta.13",
+    "vuetify": "^3.1.5",
```
andrewsignori-aot, line 35: First time ever in the project that we are using a non-(alpha/beta) version 🥳

sources/packages/web/src/assets/css/base.scss
```diff
@@ -378,3 +378,8 @@ b,
 strong {
   font-weight: $font-weight-bold;
 }
+
+// Added FontAwesome as fallback.
+.mdi:before {
+  font-family: "Material Design Icons", "FontAwesome", "BCSans", sans-serif !important;
```
andrewsignori-aot, line 384: Can you please elaborate on this for my own knowledge only 😉

sources/packages/web/src/helpers/helpers/uri.ts
```diff
@@ -1,12 +1,14 @@
...
   DataTableSortOrder,
   FieldSortOrder,
   PaginationOptions,
   PaginationParams,
 } from "@/types";
 
 /**
- * helper to append pagination sort and order to the url
+ * todo: remove sortOrder: DataTableSortOrder when all primevue datatables are removed.
+ * Helper to append pagination sort and order to the url
```
andrewsignori-aot, line 11: If I am not wrong, this method can be marked as `deprecated` because the `getPaginationQueryString` is supposed to do the same that `addSortOptions` and `addPaginationOptions` in a cleaner way. Does it make sense?

sources/packages/web/src/types/contracts/DataTableContract.ts
```diff
@@ -101,3 +135,18 @@ export interface PageAndSortEvent {
...
+
+/**
+ * Vuetify dataTable sort order.
+ */
+export enum DataTableSortByOrder {
+  DESC = "desc",
+  ASC = "asc",
+}
+
+// Datatable options.
```
andrewsignori-aot, line 147: Can we keep the comment following the same from the others on the file?
```ts
/**
 * Vuetify dataTable sort order.
 */
```

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
```diff
@@ -37,66 +37,28 @@
...
+          :items="programAndCount.results"
+          :items-length="programAndCount.count"
           :loading="loading"
+          :items-per-page="DEFAULT_PAGE_LIMIT"
+          @update:options="paginationAndSortEvent"
         >
-          <Column
-            :field="ProgramSummaryFields.CipCode"
-            header="CIP"
-            data-cy="programCIP"
```
andrewsignori-aot, line 54: Is the `data-cy` added somewhere?

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
```diff
@@ -173,22 +138,25 @@ export default defineComponent({
...
+    const paginationAndSortEvent = async (event: DataTableOptions) => {
+      currentPage.value = event.page - 1;
+      currentPageLimit.value = event.itemsPerPage;
+      const [sortByOptions] = event.sortBy;
       await loadSummary(
-        event.page,
-        event.rows,
-        event.sortField,
-        event.sortOrder,
+        event.page - 1,
```
andrewsignori-aot, line 146: Can you please elaborate on why we cannot use the `event.page` as it is and we must add the `-1` on line 142 and 146?

andrewsignori-aot (commented):
Nice work. I assume the upgrade did not cause any significant layout break, right?
Was there any direction that indicates what is the plan to move the `v-data-table-server` component from labs to a final release? The real question would be, can we start to replace the other components right now or hold till the component is part of the release?

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
ann-aot: when we were using primevue data table, the value of the page was 0 for the first page and in the vuetify data table the first page value is 1.
eg, this is first page in prime vue 
image
and in vuetify 3
image

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
```diff
@@ -24,9 +24,10 @@ abstract class PaginationOptionsAPIInDTO {
   page: number;
   /**
    * Page size or records per page.
+   * -1 means all records.
    */
-  @Min(1)
-  @Max(50)
+  @Min(-1)
+  @Max(100)
```
dheepak-aot, line 30: Why did we change this to 100?

sources/packages/web/src/assets/css/base.scss
ann-aot: ya... 
In our application, we have set icons default set as "mdi", 
image
so, vuetify component using the font awesome class is having class "mdi", which is the `Material Design` class and it's overriding the "FontAwesome" font family with "`Material Design` ". 
image

because of this, those icons are not visible,
eg, 
image
image
image

I tried to update the  `defaultSet`  with `fa`, 
image, doing so, `mdi` icons were having issues.
image


I have updated the CSS, and I have reassigned the font awesome class with its family 
image

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
```diff
@@ -24,9 +24,10 @@ abstract class PaginationOptionsAPIInDTO {
   page: number;
   /**
    * Page size or records per page.
+   * -1 means all records.
```
dheepak-aot, line 27: Min could be also 0 here as 0 is returning all records.

image

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
andrewsignori-aot: We can change the API to work with the pagination starting in 1 (because 0 based or 1 would be both accaptable) or we can adjust the pagination as needed at the Vue API service layer only, for instance, while calling the method `getPaginationQueryString`.
What I would like to avoid is doing "-1" everywhere.

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
andrewsignori-aot: For the data tables loading data from the server we should not allow the request for all the data. Is it possible in the component? I am asking because the "-1 means all records.".

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -268,16 +268,14 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
-    if (paginationOptions.page) {
-      paginatedProgramQuery.offset(
-        paginationOptions.page * paginationOptions.pageLimit,
-      );
-    } else {
-      paginatedProgramQuery.limit(0);
-    }
+
+    paginatedProgramQuery.offset(
+      paginationOptions.page * paginationOptions.pageLimit,
```
dheepak-aot, line 276: what happens here when `paginationOptions.page` is not set (no pagination used)?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -268,16 +268,14 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
     // For getting total raw count before pagination.
     const sqlQuery = paginatedProgramQuery.getSql();
 
-    if (paginationOptions.pageLimit) {
+    if (paginationOptions.pageLimit > 0) {
```
dheepak-aot, line 271: if  `pageLimit` min is set to 0 then we can use the old condition itself right?

ann-aot:
> Nice work. I assume the upgrade did not cause any significant layout break, right? Was there any direction that indicates what is the plan to move the `v-data-table-server` component from labs to a final release? The real question would be, can we start to replace the other components right now or hold till the component is part of the release?

yes, @andrewsignori-aot I did a smoke test, and I didn't find any significant layout break. I think it's pretty stable. And regarding the component release, I was searching for it and I couldn't find any information regarding them. I think it is pretty stable and they are also adding features and doing bug fixes for the component, https://github.com/vuetifyjs/vuetify/releases?q=VDataTableServer&expanded=true. IMO, the component has what we needed,  except the search functionality (and I couldn't find anything saying there will introduce it in v-datatable-server, though the other  2 data table has that functionality)

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
ann-aot: In the example here you are showing the `v-pagination` not the default one. below is the default one 
image

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
ann-aot: because the default pagination has `100`

sources/packages/web/src/types/contracts/DataTableContract.ts
```diff
@@ -101,3 +136,25 @@ export interface PageAndSortEvent {
...
+  ASC = "asc",
+}
+
+/**
+ * DataTable options.
+ */
+export interface DataTableOptions {
+  itemsPerPage: number;
+  page: number;
+  sortBy: { key: string; order: DataTableSortByOrder }[] | [];
```
dheepak-aot, line 154: Correct me if wrong, the `| []` is to accommodate the current sort order in place.

If so having a to-do here will be helpful to remove that.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
ann-aot: what do you mean by no pagination? pagination will be always there, right?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
ann-aot: -1 is for all records. i have added in the code comment

sources/packages/web/src/types/contracts/DataTableContract.ts
ann-aot: mm, I didn't understand. when there is no sort (sort icon not clicked) then sortBy: will be empty array

sources/packages/web/src/types/contracts/DataTableContract.ts
```diff
@@ -68,6 +69,39 @@ export enum ProgramSummaryFields {
...
 
+export const ProgramSummaryHeaders = [
+  {
+    title: "CIP",
+    sortable: false,
+    key: "cipCode",
+  },
+  {
+    title: "Program Name",
+    sortable: true,
```
dheepak-aot, line 80: I guess the sortable true is by default.

sources/packages/web/src/views/institution/locations/programs/LocationPrograms.vue
```diff
@@ -37,66 +37,40 @@
...
-              <v-btn
-                color="primary"
-                @click="goToViewProgram(slotProps.data.programId)"
-                data-cy="viewProgram"
-                >View</v-btn
-              >
-            </template>
-          </Column>
-        </DataTable>
+          <template v-slot:item="{ item }">
```
dheepak-aot, line 48: using shortcut #item ?

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
dheepak-aot: According to what @andrewsignori-aot which is right min value must be 1 right?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
dheepak-aot: Because all the pagination properties are query parameters and technically the API can be used without pagination and sort.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
dheepak-aot: I saw the comment but made this comment with respect to whether should use Min as -1 for the pagelimit.

sources/packages/web/src/types/contracts/DataTableContract.ts
dheepak-aot: I got it now.

sources/packages/web/src/assets/css/base.scss
dheepak-aot: Ah, I see may be that is why the icons are not visible in v-select

image

dheepak-aot:
Great work @ann-aot . I also had the same question that @andrewsignori-aot  had  as is it the right time to change other components to vuetify data table?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
ann-aot: i think pagination will be always there, i can't see a scenario, where there is no pagination

sources/packages/web/src/assets/css/base.scss
andrewsignori-aot: Thanks @ann-aot for providing the details.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
andrewsignori-aot: On `PaginationOptionsAPIInDTO`, `page` and `pageLimit` are required. I believe the methods that are built for pagination must provide the pagination parameters.

andrepestana-aot (commented):
Great job!

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
ann-aot: agree with @andrewsignori-aot. min is 0 now, so the API will restrict when all records are fetched. As discussed I could find a way to disable `All` or to pass pagination options as primevue, the only option currently is to extend the footer and add a custom one. Leaving it like this and When we decide to change all data tables, if there isn`t any option available we can invest more time on it as discussed

ann-aot:
> Great work @ann-aot . I also had the same question that @andrewsignori-aot had as is it the right time to change other components to vuetify data table or wait till it move from labs to vuetify/lib?

https://github.com/bcgov/SIMS/pull/1754#issuecomment-1444550164

andrewsignori-aot (approved):

dheepak-aot (approved):
LGTM 👍. Thanks for making the changes.

Issue #1692, opened 2023-01-26
Implement Ministry Overaward Interface


JasonCTang:
## Context
Implement an overawards section on the student profile. Display the students overawards and allow ministry staff to manually deduct from their overawards balance in the event the student pays back their debt through the FEDS (student needs to call in to tell SABC). Otherwise system will trigger a deduction from future loan amounts (through reassessment)

## Figma
[Ministry mockups > Student Profile section > Overawards tab](https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=10649%3A132523&t=To0Ob0zgHEqyeuXL-0)

## Acceptance Criteria
- [x] In student profile, display a new tab called "Overawards"
- [x] Display a students overaward balance by CSLF, CSLP, BCSL - include data from SIMS and SFAS (legacy overaward)
- [x] Display in a table a students overawards generated (sorted by date added - most recent)
- [x] Display in a table overawards deducted (sorted by date added) - either by system deductions (award deducted) OR when the student pays back on NSLSC (manual record)
- [x] Display a button called "Add manual record" triggering a modal to open with the type of award and dollar amount that the student paid back
- [x] The value entered into the "Add manual record" modal, must be inverted when inserting into the table.
- [x] Validation for value entered into the "Add manual record" - Between 0 - 99999
- [x] Rename database ENUMs to match Figma.
- [x] Checking if "pending awards cancel" can be removed.
- [x] All displays to be created as common components (will be eventually reused in Student Portal)
- [x] Client side pagination only.
- [ ] Only "student-add-overaward-manual" can add manual record.
- [ ] Role "student-add-overaward-manual" must be part of aest-business-administrator

dheepak-aot:
Ticker #1753 created to add notes during manual overaward deduction.

PR #1755, opened 2023-02-24
#1692 - Fix for the toggle content with right constant

dheepak-aot:
- [x] Assign the right constant to the toggle content
- [x] Made the v-container fluid


andrepestana-aot (approved):
👍

PR #1770, opened 2023-03-02
#1677 - Bug when there is no sfasIndividual

andrepestana-aot:
Just fixing an issue for the last PR. It was not checking if sfasIndividual was null.


andrewsignori-aot (approved):
Nice catch, looks good 👍

dheepak-aot (commented):
👍

Issue #1769, opened 2023-03-01
E2E/Unit - API Controllers


JasonCTang:
- program-year.students.controller.ts
- application-exception.aest.controller.ts

PR #1803, opened 2023-03-15
#1769 - E2E/Unit - API Controllers - Part 1 - Program Years

andrepestana-aot:
- Created DB seeding CreateStudentUsers;
- Refactored students controller to use student user persisted by the DB seeding;
- Created e2e tests for program years;


sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      });
+
+    // Revert
+    programYear.active = true;
+    await programYearRepo.save(programYear);
+  });
+
+  it("Should get an active program year by id.", async () => {
+    // Arrange
+    const endpoint = "/students/program-year/1/active";
```
andrewsignori-aot, line 101: I would recommend that the test create the program year as required and executed the test on that record.
We may rely on some DB seeding in some cases, but I believe that program year would not be one of them.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  let appDataSource: DataSource;
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    appDataSource = dataSource;
+    app = nestApplication;
+  });
+
+  it("Should get the list of the available program years.", async () => {
+    // Arrange
+    const endpoint = "/students/program-year/options-list";
```
andrewsignori-aot, line 24: Same for https://github.com/bcgov/SIMS/pull/1803/files#r1137781213

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+describe("ProgramYearStudentsController(e2e)-getProgramYears", () => {
+  let app: INestApplication;
+  let appDataSource: DataSource;
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    appDataSource = dataSource;
+    app = nestApplication;
+  });
+
+  it("Should get the list of the available program years.", async () => {
```
ann-aot, line 22: we are following `should` and `when` approach

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .then((response) => {
+        const programYearList = response.body;
+        expect(programYearList).toHaveLength(3);
+        expect(programYearList).toEqual(
```
andrewsignori-aot, line 37: I would recommend doing the check following the `expected` sntax as below. IMO it is simpler and cleaner. It will eliminate the need to check the 36 also.
image

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+            }),
+          ]),
+        );
+      });
+  });
+
+  it("Should not include inactive program years in the list.", async () => {
+    // Arrange
+    const programYearRepo = appDataSource.getRepository(ProgramYear);
+    const programYear = await programYearRepo.findOneBy({ id: 2 });
```
andrewsignori-aot, line 62: Changing the current program years causes side effects in other tests.
This would be the same case of https://github.com/bcgov/SIMS/pull/1803/files#r1137781213

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .then((response) => {
+        const programYearList = response.body;
+        expect(programYearList).toHaveLength(2);
+        expect(programYearList).toEqual(
+          expect.arrayContaining([
```
andrewsignori-aot, line 79: Same from https://github.com/bcgov/SIMS/pull/1803/files#r1137786075

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+          id: 1,
+          formName: "SFAA2021-22",
+          programYear: "2021-2022",
+          programYearDesc:
+            "Study starting between August 01, 2021 and July 31, 2022",
+        });
+      });
+  });
+
+  it("Should return 'not found' status for an inexistent program year.", async () => {
```
andrewsignori-aot, line 123: Please try to use `Should/when` when possible.
For instance: "Should return 'not found' status when the program year id does not exist."

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    // Arrange
+    const endpoint = "/students/program-year/4/active";
```
andrewsignori-aot, line 125: Same from https://github.com/bcgov/SIMS/pull/1803/files#r1137786075

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    const studentToken = await getStudentToken(
+      FakeStudentUsersTypes.FakeStudentUserType1,
+    );
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.NOT_FOUND);
```
andrewsignori-aot, line 134: Please be precise about the message, as shown below.
image

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    const endpoint = "/students/program-year/3/active";
+    const studentToken = await getStudentToken(
+      FakeStudentUsersTypes.FakeStudentUserType1,
+    );
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.NOT_FOUND);
```
andrewsignori-aot, line 151: Same from https://github.com/bcgov/SIMS/pull/1803/files#r1137789924

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  });
+  it("Should return 'not found' status for an inactive program year.", async () => {
```
ann-aot, line 136: pls add a space before the `it`

ann-aot: https://github.com/bcgov/SIMS/pull/1803/files#r1137789037

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrepestana-aot: What should I test in this test if "ignore" the records that are persisted by the DB seeding? Just check if it contains the records I added?

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
```diff
@@ -1,113 +1,63 @@
...
-    simsUser.email = faker.internet.email();
-    simsUser.firstName = faker.name.firstName();
-    simsUser.lastName = faker.name.lastName();
-    fakeStudent.user = simsUser;
-
-    // Save the student in SIMS.
-    await studentService.save(fakeStudent);
-
-    // creating mockup for ATBCCreateClient, this function actually calls the ATBC server to create the student profile
+  it("Should return an HTTP 200 status when applying for PD and student is valid", async () => {
```
ann-aot, line 39: full stop?

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrepestana-aot: Ok. I can setup them on beforeAll and remove them on afterAll, right? But also it may interfere in other tests that might be running at the same time, no?

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @returns a student.
+ */
+async function getStudentByUsername(
+  userName: string,
+  dataSource: DataSource,
+): Promise<Student> {
+  const userRepo = dataSource.getRepository(User);
+  const user = await userRepo.findOneBy({ userName });
+  const studentRepo = dataSource.getRepository(Student);
+  return await studentRepo.findOneBy({ user: { id: user.id } });
```
ann-aot, line 66: await not needed

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrewsignori-aot: Yes, test if yours belongs to the list or select the active ones from DB and check if the list matches with what you got from the API.
The 3 program years records right now were added at the beginning of the project, they are subject to be changed. I would recommend that the test have control of the data tested.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrewsignori-aot: First, you do not need to remove them, unless there is a need to. They will be created on the test DB only.
They can be created `beforeAll` to be shared as needed but if the tests will be changing them, in a way that could interfere with other tests in the same file, the specific test can create some program years records also as needed.

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
```diff
@@ -1,113 +1,63 @@
...
-
+    // Act/Assert
+    await request(app.getHttpServer())
+      .patch("/students/atbc/apply-pd-status")
+      .auth(accessToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK);
+  });
+  afterAll(async () => {
+    // Putting the student in the same state as before the test.
+    const fakeStudent = await getStudentByFakeStudentUserType(
```
andrewsignori-aot, line 55: I am not following this code here. Is the `studentPDSentAt` changed by these tests?

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+      process.env.E2E_TEST_STUDENT_USERNAME,
+      process.env.E2E_TEST_STUDENT_PASSWORD,
+      studentClientId,
+    );
+  }
+  return studentToken.access_token;
+}
+
+/**
+ * Get a student by one of the {@link FakeStudentUsersTypes}.
```
andrewsignori-aot, line 36: `{@link FakeStudentUsersTypes}` ❤️

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
andrewsignori-aot, line 66: Can the student be found, using the user name, using only the `studentRepo`?
For instance `repo.findOne({ where: { user: { userName } } })`?

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+import { DataSource } from "typeorm";
+
+const studentClientId = "student";
+
+export enum FakeStudentUsersTypes {
+  /**
+   * Fake student user type used for simple tests.
+   * This student has a valid SIN and SIN consent.
+   */
+  FakeStudentUserType1,
```
andrewsignori-aot, line 13: I believe that we should decide on a pattern to have the ENV variable, user name, and FakeStudentUsersTypes in sync.
Thoughts? 
@andrepestana-aot @ann-aot @dheepak-aot @guru-aot @sh16011993

sources/packages/backend/apps/test-db-seeding/src/db-seeding/student/basic-seeding/create-student-users.ts
```diff
@@ -0,0 +1,103 @@
...
+import { createFakeStudent, createFakeUser } from "@sims/test-utils";
+import {
+  DataSeed,
+  DataSeedMethod,
+  SeedPriorityOrder,
+} from "../../../seed-executors";
+import { Repository } from "typeorm";
+import * as faker from "faker";
+
+const fakeStudents = [
```
andrewsignori-aot, line 13: Can we have this moved to a model following what was done for the institutions?

sources/packages/backend/apps/test-db-seeding/src/db-seeding/student/basic-seeding/create-student-users.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (options?.sin) {
+      const sinValidation = new SINValidation();
+      sinValidation.student = fakeStudent;
+      fakeStudent.sinValidation = sinValidation;
+      sinValidation.sin = options.sin;
+      sinValidation.isValidSIN = options.isValidSIN ?? false;
+    }
+
+    await this.studentRepo.save(fakeStudent);
+    return fakeUser;
```
andrewsignori-aot, line 101: Apparently, there is no usage for the user returned here. I would recommend avoiding it.
If something needs to be returned maybe the student would be more meaningful, since the method name is `createStudentUser` and the student object also contains the user.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    programYear.active = false;
+    await programYearRepo.save(programYear);
+    const endpoint = "/students/program-year/options-list";
+    const studentToken = await getStudentToken(
+      FakeStudentUsersTypes.FakeStudentUserType1,
+    );
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
```
dheepak-aot, line 72: This failure is not in test scenario.

image

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
```diff
@@ -1,113 +1,63 @@
...
-        AppStudentsModule,
-        NotificationsModule,
-        createMockedZeebeModule(),
-        MockedQueueModule,
-      ],
-    }).compile();
-    userService = await moduleFixture.get(UserService);
+    const { nestApplication, module, dataSource } =
+      await createTestingAppModule();
+    const moduleFixture = module;
```
dheepak-aot, line 32: What is the need for this re-assignment?

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
dheepak-aot: Yes. I agree. Also IMO type1 could be named more aligned with the type.

for e.g if we differentiate student users between bceid and bcsc authorized users, then FakeStudentUserTypeBceid and FakeStudentUserBCSC 

or if we don't have such thing, then just FakeStudentUser would be better than FakeStudentUserTypeSomething.

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
andrepestana-aot: yes, when the student apply for PD status it stamps the datetime in `pd_date_sent` column and if that is not reverted, next time the test is run it will get a 422 "Unprocessable Entity".
image

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .then((response) => {
+        const programYearList = response.body;
+        expect(programYearList).toHaveLength(3);
+        expect(programYearList).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              id: 1,
```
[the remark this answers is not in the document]
guru-aot: I guess, as its fetched from the DB and when the initial seeding happened with db migration, this is okay to check.

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
dheepak-aot: Again this method might be useful as helper if there is going to be more than one keycloak user.

dheepak-aot:
Nice work @andrepestana-aot . Please have a look at the comments.

sources/packages/backend/apps/test-db-seeding/src/db-seeding/student/basic-seeding/create-student-users.ts
guru-aot: Yes i think, just assign the faker user to the student object and leave it as it is.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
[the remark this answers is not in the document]
andrepestana-aot: Why not?

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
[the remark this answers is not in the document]
andrepestana-aot: ok. I'm gonna split it then.

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
andrepestana-aot: yes,  you're right. Not need for that. Thanks!

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * Get a student by one of the {@link FakeStudentUsersTypes}.
+ * @param fakeStudentUserType a fake student user type.
+ * @param dataSource the application data source.
+ * @returns a student.
+ */
+export async function getStudentByFakeStudentUserType(
+  fakeStudentUserType: FakeStudentUsersTypes,
```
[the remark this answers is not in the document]
andrepestana-aot: We're going to have more in the future. We're just not sure what are they going to be for now. Please feel free to suggest on https://github.com/bcgov/SIMS/pull/1803#discussion_r1137847465

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
andrepestana-aot: @dheepak-aot we'll need different student users to test other data like with PD Status or temporary SIN as test those conditions.

ann-aot: how about following the same format as that of the env variable?

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrepestana-aot: As those program years are subject to be changed, I think they shouldn't be used by other tests either, right?

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
andrepestana-aot: > how about following the same format as that of the env variable?

Do you mean same name as the env variable?

I think we could just give them different numbers and explain what data or what they are used for in the comments. Otherwise they have too long names like `FakeUserStudentWithPdStatusXAndValidSinAndApplicationUsedToTestY`.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYearById.e2e-spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  let appDataSource: DataSource;
+  let programYear2001: ProgramYear;
+  let programYear2026: ProgramYear;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    appDataSource = dataSource;
+    const programYearRepo = appDataSource.getRepository(ProgramYear);
+
+    programYear2001 = {
```
andrewsignori-aot, line 23: For E2E tests, the idea is to avoid the pain of individually creating objects and setting all their properties, that is why we have, for instance, the `createFakeSomething` factories. The idea is to generate dummy values and set only the ones needed for testing.
We do have a `createFakeProgramYear` that can be, for instance, adapted to receive a year and generate all other fake properties based on that parameter.

For instance, lines 23 to 46 can be converted to:
```ts
const programYear2001Promise = programYearRepo.save(createFakeProgramYear(2001));
const programYear2026Promise = programYearRepo.save(createFakeProgramYear(2026));
```

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYearById.e2e-spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      partnerFormName: "SFAA2026-01-partner",
+      programYearPrefix: "2026",
+    } as ProgramYear;
+
+    const programYear2001Promise = programYearRepo.save(programYear2001);
+    const programYear2026Promise = programYearRepo.save(programYear2026);
+
+    [programYear2001, programYear2026] = await Promise.all([
+      programYear2001Promise,
+      programYear2026Promise,
```
andrewsignori-aot, line 50: Can we use a date in the past instead of 2026? I liked the idea of using 2001, for instance.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYearById.e2e-spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      .expect(HttpStatus.NOT_FOUND)
+      .expect({
+        statusCode: 404,
+        message: `Program Year Id ${programYear2001.id} was not found.`,
+        error: "Not Found",
+      });
+  });
+
+  afterAll(async () => {
+    const programYearRepo = appDataSource.getRepository(ProgramYear);
```
andrewsignori-aot, line 115: As a general rule, we have not been cleaning the data after the tests are executed.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrewsignori-aot, line 1: Same comments from `program-year.students.controller.getProgramYearById.e2e-spec.ts` can be applied here.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      programYear2024Promise,
+      programYear2025Promise,
+    ]);
+
+    app = nestApplication;
+  });
+
+  it("Should get the list of the available program years when available.", async () => {
+    // Arrange
+    const endpoint = "/students/program-year/options-list";
```
andrewsignori-aot, line 71: We have some tests (in general, not talking about this PR) using the `endpoint` and `token` directly at the `Act/Assert`.
I believe that it looks better here at the `Arrange` level as it is right here.
Can we agree to always have the target `endpoint` and the `token` declared at this level?
@ann-aot @guru-aot @dheepak-aot @sh16011993?

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
andrewsignori-aot: A see the point but:
1. It happens only on the DEV machine while DEV is reusing the same DB over and over again. The tests are intended to be executed on a clean DB.
2. This is the type of test where we are changing the student data in a way that can potentially cause side effects in other tests. Should we maybe not be using the `FakeStudentUserType1` then?

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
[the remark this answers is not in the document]
dheepak-aot: My point was id is something which is created by DB sequence and why do we need to check it's value. 

Having said that, If we are creating the data (e.g. program year) at test case level, then we can check for all the properties including id as we have the persistent object.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
```diff
@@ -0,0 +1,114 @@
...
+  let appDataSource: DataSource;
+  let programYear2000: ProgramYear;
+  let programYear2024: ProgramYear;
+  let programYear2025: ProgramYear;
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    appDataSource = dataSource;
+    const programYearRepo = appDataSource.getRepository(ProgramYear);
+
+    programYear2000 = {
```
dheepak-aot, line 23: We can leverage the factory to create program year as per our input.

sources/packages/backend/apps/api/src/testHelpers/auth/student-token-helpers.ts
andrewsignori-aot: To be honest, because we are talking about test users I would not mind being pretty expressive with the naming.
We do not need to express in the name absolute every single detail but some key points would be good, IMO, and then we can extend the intended scenario in the comments.
`FakeUserStudentWithPdStatusXAndValidSinAndApplicationUsedToTestY` can be, for instance, `FakeStudentusers.BCeIDWithPDAndTemporarySIN`
or even
`FakeStudentusers.BCSCWithActiveRestrictionToPreventApplicationSubmission`

We always have the possibility to mock the token authentication and have just a few real students execute the Keycloak authentication, after the real authentication is done we can then mock the validation token. This would allow us to create students as needed. It would also break a bit the idea of the E2E to validate the authentication part, something to maybe have a DEV sync up to talk about it.

sources/packages/backend/apps/test-db-seeding/src/db-seeding/student/basic-seeding/create-student-users.model.ts
```diff
@@ -0,0 +1,24 @@
...
+  username: string;
+  options?: {
+    sin?: string;
+    isValidSIN?: boolean;
+    sinConsent?: boolean;
+    birthDate?: string;
+    gender?: string;
+  };
+}
+export const STUDENTS_INITIAL_DATA = [
```
andrewsignori-aot, line 13: Add a blank line above, please.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
dheepak-aot: I didn't see a test scenario for this. I believe it is because of the program year created in data seed right? We cannot achieve a not found scenario

andrewsignori-aot (commented):
Thanks for doing the changes, please take a look at the comments we can have a dev sync up to expedite the review also.

dheepak-aot:
Added 2 more comments @andrepestana-aot . Please have a look.

sources/packages/backend/apps/api/src/route-controllers/student/student.students.controller.e2e-spec.ts
andrepestana-aot: > 1. It happens only on the DEV machine while DEV is reusing the same DB over and over again. The tests are intended to be executed on a clean DB.
I thought that the idea was to run the tests as many times as we needed with the same result but ok, I got what you say. 

> 2. This is the type of test where we are changing the student data in a way that can potentially cause side effects in other tests. Should we maybe not be using the `FakeStudentUserType1` then?
Yes, but that other user should be used specifically for this test then, right?

andrepestana-aot: Ok, we discussed and decided that I'm going to use the same user for both tests as they don't affect one another. Thanks!

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
andrepestana-aot: How can I test that scenario if the DB migration already includes 3 program years? As discussed earlier we should not remove those program years (https://github.com/bcgov/SIMS/pull/1803#discussion_r1137839064). Also those program years are used for some tests and some applications use them.

sources/packages/backend/apps/api/src/testHelpers/fake-entities/program-year-fake.ts
```diff
@@ -1,12 +1,21 @@
 import { ProgramYear } from "@sims/sims-db";
 
-export function createFakeProgramYear(): ProgramYear {
+export function createFakeProgramYear(
```
andrewsignori-aot, line 3: The `createFakeProgramYear` is created by default under the lib because they are potentially used everywhere. There is already a `createFakeProgramYear` at the libs that can be replaced entirely by this code (that looks way better 😉 ).

sources/packages/backend/apps/api/src/testHelpers/fake-entities/program-year-fake.ts
```diff
@@ -1,12 +1,21 @@
...
+  programYearPrefix?: number,
+  active?: boolean,
```
andrewsignori-aot, line 5: In general, we have not been passing parameters just for the sake of direct assignment, like `active`.
The idea is that we can invoke the `createFakeProgramYear` method and set the property there.
For instance, the `programYearPrefix` is a great example of a parameter that, when provided, is useful to set many other properties.

sources/packages/backend/apps/api/src/route-controllers/program-year/_tests_/e2e/program-year.students.controller.getProgramYears.e2e-spec.ts
dheepak-aot: Yes that is what exactly I mentioned that the program years are already there.

dheepak-aot (approved):
Thanks for doing the changes @andrepestana-aot 👍LGTM.

sources/packages/backend/libs/test-utils/src/factories/application.ts
```diff
@@ -19,6 +19,7 @@ export function createFakeApplication(relations?: {
   const application = new Application();
   application.data = {} as ApplicationData;
   application.programYear = relations?.programYear ?? createFakeProgramYear();
+  application.programYear.id = 2;
```
andrewsignori-aot, line 22: Can you elaborate on why we need to have it hard-coded here always to use id 2?
Was it here causing some issues?

andrepestana-aot: yes, the previous createFakeProgramYear() used that id as the programYear is already in the DB.

andrewsignori-aot (approved):
Nice work and thanks for the discussions along the PR that will help us decide the way to have the E2E authentication working for Students 👍

ann-aot (approved):
Great work @andrepestana-aot  nd nice discussions 👍

sh16011993 (approved):

guru-aot (approved):

Issue #1780, opened 2023-03-07
Move MSFAA association from Application to Disbursement


JasonCTang:
Currently MSFAA is associated to the application, but we need to associate it at the disbursement level because the number can change in the middle of an application.

Currently, when an application is submitted and there is an active MSFAA in place, SIMS associates the exisitng MSFAA to the application. This needs to change to being associated with the disbursement.

When a new MSFAA is issued, it needs to be reassociated with all pending disbursements.

**Additional Context**

No change: an ECERT does not get generated when the student does not have a valid MSFAA.
No change: an application is submitted and the student does not have an active MSFAA (including never had a MSFAA before), a new MSFAA number is generated.
No change: Separate MSFAAs for PT and FT
No change: MSFAA is sent with every ECERT

**Acceptance Criteria**
- [x] Create new field to store the MSFAA number in the disbursement table.
- [x] Database migration to copy the MSFAA number to the new column in the corresponding disbursement record.
- [x] Drop the MSFAA number column from the application table.
- [x] Adjusting the workflow to work based on the new table structure. (Read and write MSFAA number to/from the disbursement table instead of application table)
- [x] Ensure that MSFAA is still generated during assessment workflow execution and the NOA is still displaying the MSFAA number.

dheepak-aot:
The association of MSFAA to an application was based on (it is assumed that offering Intensity is matched for all the given points)

1. Check if the student has either a _**signed MSFAA number where the date_signed is not older than 2 years(730 days) and with cancelled_date as null**_ or
an _**MSFAA number which is pending to be signed with cancelled_date as null**_
2. If the above conditions does not return an MSFAA number, then check for the most recent study period for which the student submitted an application and the application is in completed status( atleast one enrolment is confirmed) and the study period end date is not older than 2 years and if the condition is satisfied, get the MSFAA number of that application.
3. Even if point 2 did not return an MSFAA number, then create a new MSFAA number.

With the implementation of this story, where the association of MSFAA is with the disbursement and not with the application, the point 2 was slightly modified. @JasonCTang that is what we discussed over the call.

Modified point 2.
**_If the point 1 does not return any MSFAA number, check for the most recent study period who's disbursement is sent and and the study period end date is not older than 2 years and if the condition is satisfied, get the MSFAA number of that disbursement._**

Issue #1777, opened 2023-03-07
Process MSFAA cancellation and reactivation from federal file.


JasonCTang:
`We` are already reading the MSFAA cancellation from the federal file, and updating two fields in the database: the cancellation date column and the new issuing province column.

We will need to inform the student that their MSFAA has been cancelled in the interface through notification.

Federal government can also reactivate an existing cancelled MSFAA number.  A "reactivated" MSFAA record just looks like the response to a new "received" MSFAA request from Finastra.   In other words, this will look like a regular MSFAA record (cancelled date and new issuing province will be null).  

When reactivating an MSFAA, update the signed date and null the "cancellation date" column and "new issuing province" column.  Then reassociate MSFAA with any disbursements pending ECERT generation.  Lastly, cancel any outstanding UNSIGNED MSFAA numbers for the student for the same offering intensity as the reactivated MSFAA (Full vs Part time) by updating the "cancellation date" and leave the "new issuing province" column blank.

As per the conversation with @JasonCTang, MSFAA reactivation template is not required as the federal government will send them notification. Received the following notification template for MSFAA Cancellation:

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Template Name: Student - MSFAA cancelled

Subject line of the email: Your MSFAA number has been cancelled – StudentAid BC

Email message:

Hey ((givenNames)) ((lastName)),

Your Master Student Financial Assistance Agreement (MSFAA) has been cancelled by National Student Loan Service Centre (NSLSC).
This will not receive any further funds until your MSFAA is reinstated.

To resolve this issue, contact StudentAid BC by phone at 1-800-561-1818.
___

**This is an automated email. Please do not reply.** StudentAid BC is committed to protecting your privacy and personal information. For security reasons we do not include account details in automated notifications.

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------

**Additional Context**
- We are already reading the file from Linda.
- We are already updating the fields.
- Cancellation message: "Your Master Student Financial Aid Agreement has been cancelled." 
- Reactivation message: "Your previously signed Master Student Financial Aid Agreement has been reactivated."
- Confirmed that reactivation record just looks like a signed record response.

**Acceptance Criteria**
- [ ] Confirm that the cancellation date is being checked in the appropriate places to determine if the MSFAA is "cancelled." Modify if necessary.
- [ ] When MSFAA is cancelled and reactivated send notification to student.
[...]

PR #1847, opened 2023-03-28
#1780 - Associate MSFAA with disbursement

dheepak-aot:
## Associate MSFAA with disbursement

 - [x]  Migration created to remove `msfaa_number_id` from sims.applications and add to sims.disbursement_schedules
 - [x]  Moved the worker `associate-msfaa` to `DisbursementController` from `ApplicationController` in workers as the associate msfaa is related to the disbursement now.
 - [x] E2E test for disbursement controller in worker is adjusted accordingly.
 - [x] Assessment gateway workflow modified to have the service task to associate msfaa for both original assessment and re-assessments.
image

 - [x] E2E test for assessment-gateway workflow is adjusted accordingly.
image

 - [x] As DisbursementScheduleService is already available in shared library and it has not been named as *sharedService, the import is using `DisbursementScheduleService as DisbursementScheduleSharedService` as we have one more DisbursementScheduleService at worker scope. Modifying DisbursementScheduleService to DisbursementScheduleSharedService is not considered to scope of PR or story.

 - [x] The MSFAA number is now at disbursement level, but the NOA still has only one placeholder for MSFAA(not at disbursement schedule level). Hence to get the value appearing in NOA correctly, it uses the msfaa of the first disbursement. 
image


sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
+import { Injectable } from "@nestjs/common";
+import {
+  ApplicationStatus,
+  DisbursementSchedule,
+  MSFAANumber,
+  OfferingIntensity,
+  RecordDataModelService,
+} from "@sims/sims-db";
+import { CustomNamedError } from "@sims/utilities";
+
```
ann-aot, line 10: pls remove the space

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
...
+        msfaaNumber: true,
+      },
+      where: {
+        studentAssessment: { id: assessmentId },
+      },
+    });
+
+    if (!firstDisbursement) {
+      throw new CustomNamedError(
+        "Disbursement not found or one of its associations is missing.",
```
ann-aot, line 57: what do you mean by `one of its associations is missing` ??  I think only `Disbursement not found` is needed right? am I missing anything

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
...
+        DISBURSEMENT_NOT_FOUND,
+      );
+    }
+
+    if (firstDisbursement.msfaaNumber?.id) {
+      throw new CustomNamedError(
```
ann-aot, line 63: I have 2 commnets, 
1. In this PR, we are not considering the case of `reactivating` the msfaa by the federal right? 
2. I assume this is the method that is called by the camunda to associate the msfaa number for the first time, so, I would like to know what is the possible scenario, where msfaa number will already be associated with the disbursement.

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
...
+      await this.msfaaNumberService.getCurrentValidMSFAANumber(
+        studentId,
+        offeringIntensity,
+      );
+    if (existingValidMSFAANumber) {
+      // Reuse the MSFAA that is still valid and avoid creating a new one.
+      msfaaNumberId = existingValidMSFAANumber.id;
+    } else {
+      // Get previously completed and signed disbursement of an application for the student
+      // to determine if an existing MSFAA is still valid.
```
ann-aot, line 88: I didn't follow the comment, if it is a valid MSAFAA, I  believe it should be part of line 78.

ann-aot: I got the answer, I checked the function, `getCurrentValidMSFAANumber` it was looking ` qb.where("msfaaNumber.dateSigned is null") OR min valid date;`
image
image

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
...
+      const previousSignedDisbursement =
+        await this.getPreviouslySignedDisbursement(
+          studentId,
+          offeringIntensity,
+        );
+
+      let hasValidMSFAANumber = false;
+      if (previousSignedDisbursement) {
+        // checks if the MSFAA number is still valid.
```
ann-aot, line 97: `Checks`

sources/packages/backend/apps/workers/src/controllers/disbursement/disbursement.controller.ts
```diff
@@ -64,4 +73,38 @@ export class DisbursementController {
...
+   * whatever is needed to create a new MSFAA or use an
+   * existing one instead.
+   */
+  @ZeebeWorker(Workers.AssociateMSFAA, {
+    fetchVariable: [ASSESSMENT_ID],
+    maxJobsToActivate: MaxJobsToActivate.Low,
+  })
+  async associateMSFAA(
+    job: Readonly<
+      ZeebeJob<AssignMSFAAJobInDTO, ICustomHeaders, IOutputVariables>
```
andrewsignori-aot, line 88: Since the worker was moved, is it not the case to move `AssignMSFAAJobInDTO` also?

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
...
+        studentAssessment: {
+          offering: { offeringIntensity },
+          application: {
+            student: { id: studentId },
+            applicationStatus: ApplicationStatus.Completed,
+          },
+        },
+        msfaaNumber: {
+          dateSigned: Not(IsNull()),
+          cancelledDate: IsNull(),
```
andrewsignori-aot, line 173: Thanks for doing the conversion to the object query and adding the missing `cancelledDate` ❤️

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,179 @@
...
+          offeringIntensity,
+        );
+        msfaaNumberId = newMSFAANumber.id;
+      }
+    }
+    const msfaaNumber = { id: msfaaNumberId } as MSFAANumber;
+    firstDisbursement.msfaaNumber = msfaaNumber;
+    const disbursementsToUpdate = [firstDisbursement];
+
+    // If the assessment has second disbursement, associate MSFAA number
```
andrewsignori-aot, line 127: As a sugestion to make the code from line 124 to 134 simple, the beginning of the method can be declared as:
```ts
const disbursements = await this.repo.find({.....
const [firstDisbursement, secondDisbursement] = disbursements ;
```
then the last part of the method wold be
```ts
disbursements.foreach(disbursement => disbursement.msfaaNumber = msfaaNumber);
this.repo.save(disbursements);
```

sources/packages/forms/src/form-definitions/noticeofassessment.json
```diff
@@ -5006,7 +5015,8 @@
                             "properties": {},
                             "allowMultipleMasks": false,
                             "addons": [],
-                            "tag": "p"
+                            "tag": "p",
+                            "className": ""
```
ann-aot, line 5019: just curious, why this `className` is getting added to all p tags.

ann-aot (commented):
Added some comments @dheepak-aot . Good work

sources/packages/backend/apps/workers/src/services/application/application.service.ts
```diff
@@ -109,149 +96,4 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
-      .select(["applications.id", "msfaaNumbers.id", "offerings.studyEndDate"])
-      .innerJoin("applications.studentAssessments", "assessment")
-      .innerJoin("assessment.offering", "offerings")
-      .innerJoin("applications.msfaaNumber", "msfaaNumbers")
-      .innerJoin("applications.student", "students")
-      .where("applications.applicationStatus = :completedStatus", {
-        completedStatus: ApplicationStatus.Completed,
-      })
-      .andWhere("assessment.triggerType = :triggerType", {
-        triggerType: AssessmentTriggerType.OriginalAssessment,
```
andrewsignori-aot, line 246: I believe that we need to mention/explain/validate this change to @JasonCTang that we are no longer relying on the original assessment offering data to check MSFAA validity.
For instance, with the change, a reassessment of an application that was reusing an existing MSFAA, may en up creating a new MSFAA if the offering start date changes.
Assuming that the application's original assessment had the offering start date of Jan 1st, 2023 and it had an MSFAA valid till Jan 15th, 2023.
If the application is reassessed and the offering change to Jan 31st, 2023, a new MSFAA will be required, right?
I believe that this is the right change, we just need to ensure that it is captured in the ticket and that Jason is aware.

andrewsignori-aot (commented):

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
andrewsignori-aot: I would be ok with the short message as @ann-aot mentioned also.

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
andrewsignori-aot: More information about workers best practices: https://docs.camunda.io/docs/components/best-practices/development/writing-good-workers/

dheepak-aot: @ann-aot 

1. Yes we are not considering reactivating as just groomed yesterday and importantly is part of #1777 
2. As @andrewsignori-aot mentioned it is to ensure the idempotency. 

I was also thinking about checking socendDisbursement.msfaaNumber?.id but realized that the association either happens to both or none. So didn't consider it.

sources/packages/backend/apps/workers/src/controllers/disbursement/disbursement.controller.ts
dheepak-aot: Yes. Nice catch.

sources/packages/forms/src/form-definitions/noticeofassessment.json
dheepak-aot: This is the first time we are extracting form defintions for noticeofassessment through form.io.

May be that is the difference between form.io and formsflow.ai

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
```diff
@@ -0,0 +1,175 @@
...
+        studentAssessment: { id: assessmentId },
+      },
+    });
+
+    const [firstDisbursement] = disbursements;
+
+    if (!firstDisbursement) {
+      throw new CustomNamedError(
+        "Disbursement not found.",
+        DISBURSEMENT_NOT_FOUND,
```
andrepestana-aot, line 59: Just for my understanding: if the disbursement is not found here, it will try to associate again in the next scheduled time, right? How many times the queue tries to associate it?

andrepestana-aot (approved):
Great job! I just left a comment for my knowledge.

andrewsignori-aot (approved):

guru-aot (approved):
LGTM, nice work @dheepak-aot

sources/packages/backend/apps/workers/src/services/disbursement-schedule/disbursement-schedule.service.ts
dheepak-aot: Associate disbursement here is the state of workflow which happens after disbursements are saved for an assessment.

image

So the expectation at this state is there must be disbursement(s) available for the assessment to make the association happen. 

There must never be a situation when trying to associate msfaa, the disbursements are not found unless there is hidden bug in workflow(for e.g. Save disbursement having an error but not failing and moving to next stage).

Issue #1837, opened 2023-03-24
"Designation Request" returns missing details (blank text fields) for Signing Officers, Agreement Documentation and Designation Acknowledgements for both Institution and Ministry User


sh16011993:
**Describe the Bug**
Designation Request returns missing details (blank text fields) for Signing Officers, Agreement Documentation and Designation Acknowledgements

**Expected Behaviour**
All the text fields and the check boxes for Signing Officers, Agreement Documentation and Designation Acknowledgements must be filled with the values that were entered while submitting the Designation Request.

**Actual Behaviour**
On submitting a "Designation Request", it returns missing details (blank text fields) for Signing Officers, Agreement Documentation and Designation Acknowledgements

![image.png](https://images.zenhubusercontent.com/63e1969c3f013ebfc60703f3/bb218623-1329-46cf-a4a1-9431c09dcac1)

![image.png](https://images.zenhubusercontent.com/63e1969c3f013ebfc60703f3/edb758b4-b7b9-4cfd-9a6c-3de72c7fee9f)

** Steps To Reproduce**
Steps to reproduce the behaviour:

A) Institution User
1) Login as an Institution User and click on Manage Institution tab
2) Then click on Manage Designation option
3) Click on Request Designation
4) Fill out all the details and submit it
5) From the Designation Requests section, click on the pending designation request that was just submitted
6) The details of the Signing Officers, Agreement Documentation and Designation Acknowledgements are missing

B) Ministry User
1) Login as a ministry user and click on Designations
2) Select the recently submitted designation and click on View under the actions column
3) The details of the Signing Officers, Agreement Documentation and Designation Acknowledgements are missing

PR #1848, opened 2023-03-29
#1837 - Fixed the Designation Request bug

sh16011993:
The following was fixed as a part of this PR:

- Private Institution User

When the private institution user submits a Designation Request as below:

image

The Pending and the Approved Designation Request can be viewed by the institution user with all the details properly populated.

image

When the ministry user logs in to approve this Designation Request, they can see all the details related to the Designation Request being properly populated.

image

- Public Institution User

For the public institution user, verified that their Designation Request View remains as before.

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
@@ -25,6 +25,7 @@ import {
 import { InstitutionUserRoles } from "../../auth/user-types.enum";
 import { FormNames } from "../../services/form/constants";
 import { DesignationAgreementControllerService } from "./designation-agreement.controller.service";
+import { InstitutionControllerService } from "../institution/institution.controller.service";
 import {
   ApiBadRequestResponse,
   ApiForbiddenResponse,
@@ -46,6 +47,7 @@ export class DesignationAgreementInstitutionsController extends BaseController {
     private readonly designationAgreementService: DesignationAgreementService,
     private readonly formService: FormService,
     private readonly designationAgreementControllerService: DesignationAgreementControllerService,
+    private readonly institutionControllerService: InstitutionControllerService,
   ) {
     super();
   }
@@ -81,6 +83,15 @@ export class DesignationAgreementInstitutionsController extends BaseController {
         "User does not have the rights to create a designation agreement.",
       );
     }
+
+    // Check if institution is private or public and append it to the payload
+    const isInstitutionPrivate =
+      await this.institutionControllerService.checkIfInstitutionIsPrivate(
+        userToken.authorizations.institutionId,
+      );
+
+    payload.isBCPrivate = isInstitutionPrivate;
+
     // Validate the dynamic data submission.
     const submissionResult = await this.formService.dryRunSubmission(
       FormNames.DesignationAgreementDetails,
@@ -106,7 +117,9 @@ export class DesignationAgreementInstitutionsController extends BaseController {
     const createdDesignation =
       await this.designationAgreementService.submitDesignationAgreement(
         userToken.authorizations.institutionId,
-        submissionResult.data.data.dynamicData,
+        submissionResult.data.data.dynamicData
+          ? submissionResult.data.data.dynamicData
+          : {},
         userToken.userId,
         payload.locations
           .filter((location) => location.requestForDesignation)
--- a/sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
@@ -75,6 +75,11 @@ export class SubmitDesignationAgreementAPIInDTO {
   @ValidateNested({ each: true })
   @Type(() => SubmittedLocationsAPIInDTO)
   locations: SubmittedLocationsAPIInDTO[];
+
+  /**
+   * Indicates if Institution is private or public
+   */
+  isBCPrivate: boolean;
 }
 
 /**
--- a/sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
@@ -80,4 +80,15 @@ export class InstitutionControllerService {
       description: institutionType.name,
     }));
   }
+
+  /**
+   * Check if institution is private or public.
+   * @param institutionId is the institution id.
+   * @returns boolean true if institution is private, else false.
+   */
+  async checkIfInstitutionIsPrivate(institutionId: number): Promise<boolean> {
+    const institutionType =
+      await this.institutionService.getInstitutionTypeById(institutionId);
+    return INSTITUTION_TYPE_BC_PRIVATE === institutionType.institutionType.id;
+  }
 }
--- a/sources/packages/backend/apps/api/src/services/institution/institution.service.ts
+++ b/sources/packages/backend/apps/api/src/services/institution/institution.service.ts
@@ -695,6 +695,27 @@ export class InstitutionService extends RecordDataModelService<Institution> {
   }
 
   /**
+   * Get the institutionType by institution id.
+   * @param institutionId Institution id.
+   * @returns Institution retrieved, if found, otherwise returns null.
+   */
+  async getInstitutionTypeById(institutionId: number): Promise<Institution> {
+    return this.repo.findOne({
+      select: {
+        institutionType: {
+          id: true,
+        },
+      },
+      where: {
+        id: institutionId,
+      },
+      relations: {
+        institutionType: true,
+      },
+    });
+  }
+
+  /**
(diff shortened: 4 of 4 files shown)


sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
```diff
@@ -81,6 +83,15 @@ export class DesignationAgreementInstitutionsController extends BaseController {
...
       );
     }
+
+    // Check if institution is private or public and append it to the payload
+    const isInstitutionPrivate =
+      await this.institutionControllerService.checkIfInstitutionIsPrivate(
+        userToken.authorizations.institutionId,
+      );
+
+    payload.isBCPrivate = isInstitutionPrivate;
```
andrewsignori-aot, line 93: Just a suggestion, creating a separate variable sometimes makes the code better to read, for this case I would not mind having it directly associated.
```ts
payload.isBCPrivate = await this.institutionControllerService.checkIfInstitutionIsPrivate(
        userToken.authorizations.institutionId,
      );
```

sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
```diff
@@ -80,4 +80,15 @@ export class InstitutionControllerService {
       description: institutionType.name,
     }));
   }
+
+  /**
+   * Check if institution is private or public.
+   * @param institutionId is the institution id.
+   * @returns boolean true if institution is private, else false.
+   */
+  async checkIfInstitutionIsPrivate(institutionId: number): Promise<boolean> {
```
andrewsignori-aot, line 89: I believe that this method can be moved to `api\src\services\institution\institution.service.ts`.

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
dheepak-aot, line 93: Are there no form definition changes? is isBCPrivate already in definition?

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
```diff
@@ -75,6 +75,11 @@ export class SubmitDesignationAgreementAPIInDTO {
   @ValidateNested({ each: true })
   @Type(() => SubmittedLocationsAPIInDTO)
   locations: SubmittedLocationsAPIInDTO[];
+
+  /**
+   * Indicates if Institution is private or public
```
andrewsignori-aot, line 80: We can adjust the comment as below.
```
isBCPrivate is part of the form and defines if the dynamic area of the form.io definition will be visible or not, which also will impact the validation using the dryrun.
Since this value has the source of truth on the institution, it must be populated by the API prior to the dryrun validation and it is part of DTO to make explicit its place in the form payload submitted but it will also be ignored by Nestjs because it does not have a decorator.
```

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/models/designation-agreement.dto.ts
```diff
@@ -75,6 +75,11 @@ export class SubmitDesignationAgreementAPIInDTO {
   @ValidateNested({ each: true })
   @Type(() => SubmittedLocationsAPIInDTO)
   locations: SubmittedLocationsAPIInDTO[];
+
```
andrewsignori-aot, line 78: Please remove the extra blank line to be aligned with the other property in the same DTO.

dheepak-aot:
Nice work @sh16011993 . As @andrewsignori-aot mentioned, making submitted_data as nullable is better than {}.

andrewsignori-aot (commented):
Nice work, please take a look in the comments.

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -694,6 +694,27 @@ export class InstitutionService extends RecordDataModelService<Institution> {
       .getOne();
   }
 
+  /**
+   * Get the institutionType by institution id.
+   * @param institutionId Institution id.
+   * @returns Institution retrieved, if found, otherwise returns null.
+   */
+  async getInstitutionTypeById(institutionId: number): Promise<Institution> {
+    return this.repo.findOne({
```
guru-aot, line 703: if the where is just an id, we can use findonebyId instead of findOne, just a suggestion

sh16011993: In that case, I will leave it as it is.

ann-aot (commented):
Good work @sh16011993 . same comments as that @andrewsignori-aot . Curious to know, how it worked before. do you have any idea? did we remove the `IsBcprivate` from dry run payload or is it something else

sh16011993:
> Good work @sh16011993 . same comments as that @andrewsignori-aot . Curious to know, how it worked before. do you have any idea? did we remove the `IsBcprivate` from dry run payload or is it something else

We have added isBCPrivate to the payload for all the designation requests now. It worked just once when we were discussing during our call.

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
```diff
@@ -81,6 +86,13 @@ export class DesignationAgreementInstitutionsController extends BaseController {
         "User does not have the rights to create a designation agreement.",
       );
     }
+
+    // Check if institution is private or public and append it to the payload
+    payload.isBCPrivate =
```
andrepestana-aot, line 91: I think that the comment sounds like there are only those 2 types: public and private but we're checking if it is "BC Private", right?  It can be a private "International" institution.

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
andrepestana-aot: The findOneBy here wouldn't bring the institutionType. AFAIK findOneBy doesn't allow you to specify the relations.

andrepestana-aot (commented):
Great job! I left a comment.

sources/packages/backend/apps/api/src/route-controllers/designation-agreement/designation-agreement.institutions.controller.ts
sh16011993: AFAIK, it is private vs "other types". Updated my comment line in the code

sources/packages/backend/apps/db-migrations/src/migrations/1680120153698-DesignationAgreementsTableAlterSubmittedData.ts
```diff
@@ -0,0 +1,24 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class DesignationAgreementsTableAlterSubmittedData1680120153698
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData(
+        "Update-col-submitted_data-drop-not-null.sql",
```
dheepak-aot, line 10: Minor thing. We never use `_` for sql file names. (I saw there is only one file which is exceptional).

dheepak-aot (approved):
Looks good @sh16011993 . Just one minor comment.

sources/packages/backend/apps/db-migrations/src/migrations/1680120153698-DesignationAgreementsTableAlterSubmittedData.ts
sh16011993: I will fix it.

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -694,6 +695,37 @@ export class InstitutionService extends RecordDataModelService<Institution> {
       .getOne();
   }
 
+  /**
+   * Check if institution is private
+   * @param institutionId is the institution id.
+   * @returns boolean true if institution is private, else false.
+   */
+  async checkIfInstitutionIsPrivate(institutionId: number): Promise<boolean> {
```
andrewsignori-aot, line 703: Just a minor change that I missed in the previous review, a method like this usually will be named as `isPrivateInstitution`, IMO.

sources/packages/backend/apps/db-migrations/src/sql/DesignationAgreements/Update-col-submitted-data-drop-not-null.sql
andrewsignori-aot, line 1: Please adjust also the entity model to reflect the DB change.

image

guru-aot (approved):
LGTM, nice work @sh16011993

ann-aot (approved):
Good work @sh16011993 👍

andrewsignori-aot (approved):
Looks good 👍

andrepestana-aot (approved):

Issue #1786, opened 2023-03-08
Enable Search Student for Institutions


JasonCTang:
**As a** * PUBLIC Institution user* **I want** *the ability to search for a student* **So That** *I can help the student understand their awards, assessments and disbursements.*

**Additional Context**
https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=11337-140247&t=K5C151j6g3EQtqRi-0

**To be answered**
- [ ] **When does this link to the institution become valid for search? (i.e. is it right after the application are submitted?**
- [ ] Do we include students with applications that are cancelled? Drafts can become cancelled status without being submitted. **POSSIBLE SOLUTION:  Only include applications with an assessment (created at the moment it is submitted) and is not overwritten.

**Acceptance Criteria**
- [ ] Enable authorization paths to the Search Student for PUBLIC institution users.
- [ ] All institution users can search students from **ANY** location.
- [ ] Add Search Student to the navigation pane for Institutions. (NOT location specific)
- [ ] Results must be limited to students who have at least one application for the institution of the user performing the search.  **Suggestion:** If student has an application with an associated assessment record and the application is not in overwritten state, **DO NOT** count that application as a link to the instittion for the student.  
- [ ] Fix the search fields overlapping when window size is adjusted.  
- [ ] Build E2E tests.

Issue #1830, opened 2023-03-22
Authorization for public vs non-public institutions


JasonCTang:
**Additional Context**

Public institutions require access to student data in a way that the system does not allow yet. Once the student starts to be related to a certain public institution, this institution should have access to the student data in a similar way to the Ministry.
Besides the new UIs that will be provided to the public institutions, the SIMS API should ensure that only public institutions will have access to the student data and also that only the students related to that student will be accessible. The proper criteria to consider a student related to an institution is defined on #1786.

**Acceptance Criteria**

- [ ] Enable access to the institution's endpoints to return some student data to only public institutions.
    - Create the decorator `@IsBCPublicInstitution` to be applied to endpoints that do not explicitly have the student id, for instance, student search. The institution type needs to be retrieved from DB at the token validation phase which will allow the new decorator to quickly check for it.
- [ ] Create an endpoint-based authorization, similar to the locations authorization, to allow the student id authorization without any additional code in the controller body. To accomplish this we need:
    -  All institution's endpoints that require student data will be following a pattern like the below ones.
        - api/institutions/institution/student/**:studentId**/profile
        - api/institutions/institution/student/**:studentId**/applications
        - api/institutions/institution/student/**:studentId**/notes
    -  Create a decorator, for instance, @AllowStudentDataAccess("studentId") similar to `@HasLocationAccess("locationId")`
        - This decorator can validate if the `studentId` provided in the URL can be accessed by the institution, which will then ensure that all queries needed inside the controller body will simply rely on the student id and no other authorization query will be required. This will also make it possible to reuse some Ministry queries since they are also based on the student id.
        - The query executed on the `@AllowStudentDataAccess` can potentially be too expensive to be executed too frequently. We can then explore the Typeorm out-of-box query cache since we have Redis available. If the cache works we can extend the cache to other queries used for token authorization (not part of this ticket).
- [ ] Apply the new decorator to the already existing endpoints.
- [ ] Apply E2E tests to the existing endpoints to test the new authorization decorators.

PR #1857, opened 2023-04-01
#1786 - Enable Search Student for Institutions - part 1

andrepestana-aot:
In this PR
- moved SearchStudents.vue to SearchStudentsCommon.vue to reuse the same UI for institutions and ministry users;
- added link for BC Public institutions;
- created API to search students by institutions;

I will create another PR to:
- create e2e tests
- fix bug on UI (responsiveness)
- check an issue with the search when the user application has any pending exception. In that case the location_id is not defined and the user won't be found by the institution.

## BC Public Institution
image

## Other Institutions
image


sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Student } from "@sims/sims-db";
+import { IInstitutionUserToken } from "../../auth";
+
+/**
+ * Student controller for institutions.
+ */
+@AllowAuthorizedParty(AuthorizedParties.institution)
+@Controller("student")
+@ApiTags(`${ClientTypeBaseRoute.Institution}-student`)
+@Injectable()
```
guru-aot, line 23: Do we need a injectable here?

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -705,6 +708,16 @@ export class InstitutionService extends RecordDataModelService<Institution> {
     return INSTITUTION_TYPE_BC_PRIVATE === institutionType.institutionType.id;
   }
 
+  /**
+   * Check if institution type is BC public.
+   * @param institutionId institution id.
+   * @returns whether the institution type is public.
+   */
+  async isBCPublicInstitution(institutionId: number): Promise<boolean> {
```
guru-aot, line 716: Just a suggestion, we already have a method to check if the institution is private, which has one reference. Can we just create a single method that return either the institution is INSTITUTION_TYPE_BC_PRIVATE or INSTITUTION_TYPE_BC_PUBLIC?. Thoughts @ann-aot @andrewsignori-aot @sh16011993 @dheepak-aot

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      )
+
+      .innerJoin("application.location", "pirLocation")
+      .innerJoin("pirLocation.institution", "pirInstitution")
+
+      .leftJoin("application.currentAssessment", "studentAssessment")
+      .leftJoin("studentAssessment.offering", "offering")
+      .leftJoin("offering.institutionLocation", "offeringLocation")
+      .leftJoin("offeringLocation.institution", "offeringInstitution")
+
```
guru-aot, line 534: please remove additional space.

sources/packages/backend/apps/api/src/app.institutions.module.ts
```diff
@@ -42,6 +42,7 @@ import {
   EducationProgramOfferingInstitutionsController,
   EducationProgramOfferingControllerService,
   ConfirmationOfEnrollmentControllerService,
+  StudentInstitutionController,
```
dheepak-aot, line 45: StudentInstitution**_s_**Controller?

While using student or institution as client type we use plural names

sources/packages/backend/apps/api/src/route-controllers/index.ts
```diff
@@ -56,3 +56,4 @@ export * from "./confirmation-of-enrollment/confirmation-of-enrollment.aest.cont
 export * from "./overaward/overaward.controller.service";
 export * from "./overaward/overaward.aest.controller";
 export * from "./overaward/overaward.students.controller";
+export * from "./student/student.institution.controller";
```
dheepak-aot, line 59: same here `student.institutions.controller`

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
     return searchQuery.getMany();
   }
 
+  /**
+   * Search students based on the search criteria.
+   * @param searchCriteria options to search by firstName,
+   * lastName, appNumber or sin.
+   * @returns list of students.
+   */
+  async searchStudentApplicationForInstitution(
```
guru-aot, line 503: Cant we reuse the existing query searchStudentApplication to use the instituion id as an optional parameter and return?

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -93,27 +93,27 @@ export class UpdateStudentAPIInDTO
 /**
  * Student AEST search parameters.
```
dheepak-aot, line 94: Please remove AEST from the comment as well.

ann-aot, line 94: pls update the comment too

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
dheepak-aot: When `@Controller` is used, `@Injectable` is not required.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+   * Search students based on the search criteria.
+   * @param searchCriteria criteria to be used in the search.
+   * @returns searched student details.
+   */
+  @Post("search")
+  async searchStudents(
+    @UserToken() userToken: IInstitutionUserToken,
+    @Body() searchCriteria: StudentSearchAPIInDTO,
+  ): Promise<SearchStudentAPIOutDTO[]> {
+    if (!this.institutionService.isBCPublicInstitution) {
```
dheepak-aot, line 42: Is this required with #1830 going to handle the authorization.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+import { InstitutionService, StudentService } from "../../services";
+import { ClientTypeBaseRoute } from "../../types";
+import { AuthorizedParties } from "../../auth/authorized-parties.enum";
+import { AllowAuthorizedParty, UserToken } from "../../auth/decorators";
+import BaseController from "../BaseController";
+import {
+  StudentSearchAPIInDTO,
+  SearchStudentAPIOutDTO,
+} from "./models/student.dto";
+
```
ann-aot, line 12: pls remove the extra line

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
andrewsignori-aot: I see the point @guru-aot.
Maybe we could create read-only properties a the entity model that will do the check based on the constants ` INSTITUTION_TYPE_BC_PRIVATE ` and `INSTITUTION_TYPE_BC_PUBLIC`.
IMO this is something that can be considered an entity-level concern. Would be like an emulation of a DB computed property.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
andrewsignori-aot: It will not be required after #1830. It will be handled prior to the controller body code being executed as @dheepak-aot mentioned.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
ann-aot: looks like its used in quite a few controllers, here are few 
image
image
image

andrepestana-aot: I just moved this code, it already this way. Thanks!

andrewsignori-aot: `@Controller` does not require the `@Injectable` because they are not injected anywhere. They were probably added by mistake someday and kept being added. Anyone that finds one can have them removed or create a PR to remove them all.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: I don't like the idea of doing that because it would make a simple thing like a query to a more complex method with conditions that will be harder to maintain. Thoughts @ann-aot @dheepak-aot @andrewsignori-aot  @sh16011993 ?

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+  ): Promise<SearchStudentAPIOutDTO[]> {
+    if (!this.institutionService.isBCPublicInstitution) {
+      return [];
+    }
+    const searchStudentApplications =
+      await this.studentService.searchStudentApplicationForInstitution(
+        userToken.authorizations.institutionId,
+        searchCriteria,
+      );
+    return searchStudentApplications.map((eachStudent: Student) => ({
```
andrewsignori-aot, line 50: If AEST and student are sharing the same DTO, the "map" logic can be centralized also on `student.controller.service.ts`.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
andrewsignori-aot: > I just moved this code, it already this way. Thanks!

I did not get the part that states that this code was "moved". This is a brand-new file, right?
If the file is new and the code was copied, the statement "it already this way" does not seem accurate, IMO.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      .leftJoin("application.currentAssessment", "studentAssessment")
+      .leftJoin("studentAssessment.offering", "offering")
+      .leftJoin("offering.institutionLocation", "offeringLocation")
+      .leftJoin("offeringLocation.institution", "offeringInstitution")
+
+      .innerJoin("student.user", "user")
+      .innerJoin("student.sinValidation", "sinValidation")
+      .where("user.isActive = true")
+      .andWhere(
+        "(offeringInstitution.id = :institutionId or pirInstitution.id = :institutionId)",
```
andrewsignori-aot, line 539: Typeorm also has the "or" syntax.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
...
+        "user.lastName",
+        "sinValidation.sin",
+      ])
+      .innerJoin(
+        Application,
+        "application",
+        "application.student.id = student.id",
+      )
+
+      .innerJoin("application.location", "pirLocation")
```
andrewsignori-aot, line 527: Maybe it would be safe to use left join here since we still are not saving the PIR location right on submission.
Even if we start saving, I believe that it would be better to not rely on it using the INNER.

sources/packages/web/src/components/common/SearchStudentsCommon.vue
andrewsignori-aot, line 1: I would prefer emoving the suffix "Common" from the component name.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+      id: eachStudent.id,
+      firstName: eachStudent.user.firstName,
+      lastName: eachStudent.user.lastName,
+      birthDate: getISODateOnlyString(eachStudent.birthDate),
+      sin: eachStudent.sinValidation.sin,
+    }));
```
dheepak-aot, line 56: As the AEST controller has same logic, it can be share using controllerService

sources/packages/web/src/components/common/SearchStudentsCommon.vue
```diff
@@ -0,0 +1,208 @@
+<template>
+  <full-page-container>
+    <body-header
+      title="Search Students"
+      subTitle="Lookup students by entering their information below."
+    >
+    </body-header>
+    <v-form ref="searchStudentsForm">
```
andrewsignori-aot, line 8: I am not sure if the shared component should be at the `full-page-container`. I would prefer to have it at the `v-form` level and the `full-page-container` and `body-header` would be defined at the specific client level.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: Yes, I see that but when I use .orWhere() it doesn't add a parenthesis to the query like in:
```sql
AND     ("offeringInstitution"."id" = $1   OR      "pirInstitution"."id"      = $2)
```
I'm not sure how to add those parenthesis through the DSL.

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
```diff
@@ -38,6 +40,18 @@ export default defineComponent({
         },
       },
     ]);
+
+    if (institutionState.value.institutionType === "BC Public") {
```
andrewsignori-aot, line 44: We already have `isBCPrivate`  on `InstitutionStateForStore` and I believe that we should add an `isBCPublic` for checks like this in the same way.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
dheepak-aot: I agree that it should just be a entity level concern.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: Maybe I am wrong and I did not tested it, but would the change be something about moving the inner/left and one where inside an if like below?
```ts
    if (institutionId) {
      searchQuery
        .innerJoin("application.location", "pirLocation")
        .innerJoin("pirLocation.institution", "pirInstitution")
        .andWhere(
          "(offeringInstitution.id = :institutionId or pirInstitution.id = :institutionId)",
          { institutionId },
        );
    }
```

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: Please see the below example:
```ts
   if (paginationOptions.searchCriteria) {
      activeApplicationQuery
        .andWhere(
          new Brackets((qb) => {
            qb.where(
              "CONCAT(user.firstName,' ', user.lastName) Ilike :searchCriteria",
            ).orWhere("application.applicationNumber Ilike :searchCriteria");
          }),
        )
        .setParameter(
          "searchCriteria",
          `%${paginationOptions.searchCriteria.trim()}%`,
        );
    }
```

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
dheepak-aot: @guru-aot I get what you are saying, but to have a single method returning private or public, we need to have a thrid category to pull all these institution types.

image

sources/packages/web/src/components/common/SearchStudentsCommon.vue
dheepak-aot: I see the point that there is already a a file in /aest/views SearchStudents.vue. 
So we can use a different name for the component as StudentSearch or StudentSearchResult not having the common in file name.

sources/packages/web/src/components/common/SearchStudentsCommon.vue
```diff
@@ -0,0 +1,208 @@
+<template>
+  <full-page-container>
+    <body-header
```
dheepak-aot, line 3: body-header-container?

sources/packages/web/src/components/common/SearchStudentsCommon.vue
andrewsignori-aot: I would be ok with having another suffix or even keeping this one, my comment was more of a suggestion.
That said unless I am missing something there is no technical issue with having both files named the same, right?

sources/packages/web/src/components/common/SearchStudentsCommon.vue
```diff
@@ -0,0 +1,208 @@
+<template>
+  <full-page-container>
```
dheepak-aot, line 2: This component can have just the search and result content and not the full page container.

They way how we follow is we create sepeate views for AEST and Institution at /views.

We use the Common component inside both the views like this.


image

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
guru-aot: I get the point team, I was just talking about combining 2 methods and returning a single value either public or private. I guess its going way out of the scope of this PR. Pausing the discussion for now.

sources/packages/web/src/constants/search-types.ts
```diff
@@ -0,0 +1,4 @@
+export enum SearchTypes {
```
dheepak-aot, line 1: I believe this enum is not required if we follow the typical component and view approach as per https://github.com/bcgov/SIMS/pull/1857/files#r1158939706

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
dheepak-aot: @guru-aot I am exactly trying to comprehend your idea. for e.g. if there is a method called `checkIfPrivateOrPublic(): Private | Public`, what will it return if the institution type is International? let me know if I am missing somehthing.

andrewsignori-aot: @guru-aot feel free to pause the discussion, from your end, please leave the comment open.
@andrepestana-aot I believe that this method will no longer be needed because we will build the decorator in the future.
If it is still needed let's move it to the entity model alongside the `isBCPrivate`.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      .leftJoin("offeringLocation.institution", "offeringInstitution")
+
+      .innerJoin("student.user", "user")
+      .innerJoin("student.sinValidation", "sinValidation")
+      .where("user.isActive = true")
+      .andWhere(
+        "(offeringInstitution.id = :institutionId or pirInstitution.id = :institutionId)",
+        { institutionId },
+      )
+      .andWhere("application.applicationStatus != :overwrittenStatus", {
```
dheepak-aot, line 542: Are we not ignoring the draft applications? as we are using the left join above for all associations created by submitted application.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      .innerJoin(
+        Application,
+        "application",
+        "application.student.id = student.id",
+      )
+
+      .innerJoin("application.location", "pirLocation")
+      .innerJoin("pirLocation.institution", "pirInstitution")
+
+      .leftJoin("application.currentAssessment", "studentAssessment")
```
dheepak-aot, line 530: This one can be an inner join right as per the AC.
image

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -494,6 +494,78 @@ export class StudentService extends RecordDataModelService<Student> {
...
+    const searchQuery = this.repo
+      .createQueryBuilder("student")
+      .select([
+        "student.id",
+        "student.birthDate",
+        "user.firstName",
+        "user.lastName",
+        "sinValidation.sin",
+      ])
+      .innerJoin(
```
dheepak-aot, line 521: why is the join explicitly mentioned here?

sources/packages/backend/apps/api/src/services/student/student.service.ts
dheepak-aot: I understood the context from our conversation.

sources/packages/backend/apps/api/src/services/student/student.service.ts
dheepak-aot: I understood the context from our conversation.

sources/packages/backend/apps/api/src/route-controllers/student/student.institution.controller.ts
andrepestana-aot: It was already that way in the other file, I just moved the code from there to this one. I'm not saying that I should not have changed it and made it right. Sorry!

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: Ok. Changed.

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
andrepestana-aot: I removed the method isBCPublicInstitution

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+  @Post("search")
+  async searchStudents(
+    @UserToken() userToken: IInstitutionUserToken,
+    @Body() searchCriteria: StudentSearchAPIInDTO,
+  ): Promise<SearchStudentAPIOutDTO[]> {
+    const institutionType =
+      await this.institutionService.getInstitutionTypeById(
+        userToken.authorizations.institutionId,
+      );
+    if (institutionType.id !== INSTITUTION_TYPE_BC_PUBLIC) {
```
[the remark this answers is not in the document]
andrepestana-aot: Well, so shouldn't I do any checks then and just wait for the decorator?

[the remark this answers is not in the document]
andrepestana-aot: ok. Removed and added TODO.

sources/packages/web/src/components/common/SearchStudents.vue
```diff
@@ -0,0 +1,174 @@
...
+              </div>
+            </template>
+          </Column>
+          <Column header="Action">
+            <template #body="slotProps">
+              <v-btn
+                color="primary"
+                class="p-button-raised"
+                data-cy="viewStudent"
+                @click="$emit('goToStudentView', slotProps.data.id)"
```
dheepak-aot, line 96: As an alternate approach because the params are same between both routes, the component can receive a a route prop and do the navigation here. 
The benefit it is avoid the emit. `<search-students :view-student-route="InstitutionRoutesConst.STUDENT_PROFILE">`

Just a suggestion.

dheepak-aot:
Thanks for making the changes. Just one outstanding question.

sources/packages/web/src/components/common/SearchStudents.vue
andrepestana-aot: Yes, there are many ways of doing that. I'm following https://github.com/bcgov/SIMS/pull/1857#pullrequestreview-1373523716

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -457,16 +462,48 @@ export class StudentService extends RecordDataModelService<Student> {
...
-      .leftJoin(
+      ]);
+    if (institutionId) {
+      searchQuery
+        .innerJoin(
+          Application,
+          "application",
+          "application.student.id = student.id",
+        )
+        .leftJoin("application.location", "pirLocation")
```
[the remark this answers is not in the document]
andrepestana-aot: Yes, I'm changing the submit logic to save the location in a different PR.

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Student controller for institutions.
+ */
+@AllowAuthorizedParty(AuthorizedParties.institution)
+@Controller("student")
+@ApiTags(`${ClientTypeBaseRoute.Institution}-student`)
+export class StudentInstitutionsController extends BaseController {
+  constructor(
+    private readonly studentService: StudentService,
+    private readonly institutionService: InstitutionService,
```
andrewsignori-aot, line 23: Please remove the `institutionService` if it is no longer used.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: I noticed the ` if (institutionId)` twice in the method. Was it because the proposed code on the comment didn't work or because it looks cleaner?

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+   * TODO add decorator to restrict to BC Public institutions.
+   * @param searchCriteria criteria to be used in the search.
+   * @returns searched student details.
+   */
+  @Post("search")
+  async searchStudents(
+    @UserToken() userToken: IInstitutionUserToken,
+    @Body() searchCriteria: StudentSearchAPIInDTO,
+  ): Promise<SearchStudentAPIOutDTO[]> {
+    return await this.studentService.searchStudentApplication(
```
andrewsignori-aot, line 39: No need to await.

andrewsignori-aot, line 39: Maybe I am missing something but is the service result directly returned through the API?
The results must be converted to a DTO.

sources/packages/backend/apps/api/src/route-controllers/student/student.aest.controller.ts
```diff
@@ -228,17 +226,9 @@ export class StudentAESTController extends BaseController {
...
-    const searchStudentApplications =
-      await this.studentService.searchStudentApplication(searchCriteria);
-    return searchStudentApplications.map((eachStudent: Student) => ({
-      id: eachStudent.id,
-      firstName: eachStudent.user.firstName,
-      lastName: eachStudent.user.lastName,
-      birthDate: getISODateOnlyString(eachStudent.birthDate),
-      sin: eachStudent.sinValidation.sin,
-    }));
+    return await this.studentService.searchStudentApplication(searchCriteria);
```
andrewsignori-aot, line 231: Same comments from https://github.com/bcgov/SIMS/pull/1857/files#r1160095243

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
andrewsignori-aot: Now I see that the service is returning the DTO directly. The services, unless they are controller services, do not return DTO directly. API DTOs are supposed to be a controller concern only, that is why we usually deal with them at controller or service controller only.

dheepak-aot (approved):
Thanks for doing the changes. LGTM 👍

sources/packages/backend/apps/api/src/route-controllers/institution/models/institution.dto.ts
```diff
@@ -132,6 +132,7 @@ export class InstitutionDetailAPIOutDTO extends InstitutionProfileAPIOutDTO {
   legalOperatingName: string;
   institutionTypeName?: string;
   isBCPrivate?: boolean;
+  isBCPublic?: boolean;
```
andrewsignori-aot, line 135: Is there a reason for this one to be nullable?
If it is just following the `isBCPrivate` maybe both can be not nullable.

sources/packages/backend/apps/api/src/services/student/student.service.ts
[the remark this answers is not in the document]
andrewsignori-aot: > Are we changing the submit logic to save the location in a different PR?
> 
> And if that is going to happen, this whole part can be removed right?
> 
> ```
>         .leftJoin("application.currentAssessment", "studentAssessment")
>         .leftJoin("studentAssessment.offering", "offering")
>         .leftJoin("offering.institutionLocation", "offeringLocation")
>         .leftJoin("offeringLocation.institution", "offeringInstitution");
> ```

During previous conversations, we had in mind doing both checks, but since we will save the `application.location` during the application submission and assuming that the location can only be changed by editing the application, it could be a nice simplification of the query.

andrewsignori-aot (approved):

Issue #1761, opened 2023-02-27
Handling full time BCSL lifetime and federal maximums


JasonCTang:
**BCSL Lifetime Maximum**
The student's cumulative full-time BCSL awards received over a lifetime need to be tracked by SIMS and identify when the student has hit the full-time BCSL Lifetime Maximum.

This will require some calculations at the time of e-Cert generation to determine if the current disbursement will cause the student to hit the maximum.  If it will, then it will be necessary to reduce the award amount accordingly so that the final total exactly equals the full-time BCSL lifetime maximum that is in effect for that program year.  

Once the maximum is hit, a restriction will be applied to the student. All subsequent disbursements will effectively zero out the full-time BCSL award amounts (and all BC funding) at the time of e-Cert generation.

Restriction applies only to full-time BCSL as part-time maximums follow completely different rules.

**Federal lifetime maximum changes**
Federal lifetime maximum restriction codes impact Full Time funding only.  The actions for the following federal restriction codes need to be modified to stop full time disbursements only, allowing part time funding to proceed without impact. 

Codes:
AV, 5, 7 and 9

**Acceptance Criteria for Federal lifetime maximum**
- [x] Modify actions for the codes (AV, 5, 7, 9) to stop full time disbursements only. (Remove "stop part time" part of action)
- [x] Perform Database Update together with the addition of a new restriction for BCSL lifetime maximum ticket #1761

**Acceptance Criteria for BCSL lifetime maximum**
- [x] Ensure that the restriction (to be created) has only `Stop full time BC funding` action.
- [x] Consider only full-time BCSL for the maximum calculation but once the restriction is in place, all BC Funding is blocked. This is because the grants can only be awarded to students who are eligible for Loans. 
    - When there are two disbursements, the BCSL is part of the second one. The system should determine the total BCSL for the current and past applications, even if the current disbursement is for the first disbursement of an application with two disbursements.
- [x] The amounts used for maximums can change and need to be stored within a program year. If a reassessment of a past application occurs, the maximum values that were in effect for the program year of the application are still in effect. The values will never change in the middle of a program year.
    - There are three different categories for maximum (PD/non-doctorate/doctorate) but the system should consider all as always the same, hence 
just one maximum configuration is needed and only one provincial restriction is needed.
    - Create a new column `max_lifetime_bc_loan_amount` on `sims.program_years` table to store the BCSL max lifetime amount. `(max_lifetime_bc_loan_amount is 50,000CAD for all program year now)`
[...]

PR #1886, opened 2023-04-13
#1761 - Handling full time BCSL lifetime and federal maximums

ann-aot:
- Added LIfetime maximum logic to e-cert generation during `calculateEffectiveValue` calculation.
- When the maximum is hit, the `BCLM`  restriction is added to the student.
- Total BCSL disbursement amount from legacy is processed from `sfasApplication` table with sin, birthdate, and last name. ( IF the student has a temporary SIN in SFAS and a permanent SIN in SIMS, we expect to miss the details from SFAS.)
- Moved certain restriction-related methods to shared services like `createRestrictionToSave`, `getRestrictionByCode`.


sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -281,27 +306,109 @@ export class ECertGenerationService {
...
+    application: Application,
+    entityManager: EntityManager,
+  ): Promise<void> {
+    if (disbursementValue.valueType === DisbursementValueType.BCLoan) {
+      const student = application.student;
+      const maxLifetimeBCLoanAmount =
+        application.programYear.maxLifetimeBCLoanAmount;
+      const auditUser = await this.systemUsersService.systemUser();
+
+      const totalLifeTimeAmount =
```
andrewsignori-aot, line 364: Please take a look at the `getOverawardBalance`. The idea for the e-Cert is that we can execute the methods in bulk and process the result in memory. Is the current way, unless I am missing something, for every single e-cert record there will be other 2 queries executed to get the sum from SIMS + SFAS.

sources/packages/backend/apps/workers/src/controllers/disbursement/disbursement.controller.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from "zeebe-node";
 // TODO: In the upcoming tasks, either DisbursementScheduleService will be renamed at shared library
 // or MSFAA related methods will move to shared library.
-import { DisbursementScheduleService as DisbursementScheduleSharedService } from "@sims/services";
+import { DisbursementScheduleSharedService } from "@sims/services";
```
andrewsignori-aot, line 11: Thanks for doing the refactor 👍

sources/packages/backend/libs/services/src/restriction/model/restriction.model.ts
```diff
@@ -28,4 +28,10 @@ export enum RestrictionCode {
    * this restriction is applied case the SIN expiry date is before the offering end date.
    */
   SINF = "SINF",
+  /**
+   * BCSL Lifetime Maximum is the student's cumulative full-time BCSL awards received over
+   * a lifetime. Thus restriction is added to a student when they reach BC lifetime maximum
```
andrewsignori-aot, line 33: Was the intention to use "Thus" here or is it supposed to be "This"?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -192,14 +200,30 @@ export class ECertGenerationService {
         END`,
         "stopFullTimeBCFunding",
       )
+      .addSelect(
```
andrewsignori-aot, line 203: Is it not the case of, instead of having 2 subqueries, we start to bring all active restrictions and do the verifications on the in-memory list, which in this case will be "all active restrictions that can impact disbursement"?
I am assuming also the possibles `StopFullTimeDisbursement` and `StopPartTimeDisbursement` will be already excluded by the where condition at line 244, since these records will not even be selected.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -281,27 +306,109 @@ export class ECertGenerationService {
...
+      if (totalLifeTimeAmount >= maxLifetimeBCLoanAmount) {
+        // Amount subtracted when lifetime maximum is reached.
+        const amountSubtracted = totalLifeTimeAmount - maxLifetimeBCLoanAmount;
+        // Ideally disbursementValue.effectiveAmount should be greater or equal to amountSubtracted.
+        // The flow will not reach here if the ministry ignore the restriction for the previous
+        // disbursement/application and money went out to the student, even though they reach the maximum.
+        const newEffectiveAmount =
+          disbursementValue.effectiveAmount - amountSubtracted;
+        /**
+         * Create {@link RestrictionCode.BCLM} restriction when lifetime maximum is reached/exceeded.
```
andrewsignori-aot, line 384: I like the `Create {@link RestrictionCode.BCLM}` and I would support its usage in the methods comments.
At this level, I believe that it leaves the code less readable, also because you need the comment to be placed with `/***/`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -281,27 +306,109 @@ export class ECertGenerationService {
...
+         */
+        const bclmRestriction =
+          await this.studentRestrictionSharedService.createRestrictionToSave(
+            student.id,
+            RestrictionCode.BCLM,
+            auditUser.id,
+            application.id,
+          );
+
+        if (bclmRestriction) {
```
andrewsignori-aot, line 394: Edge case but a possible scenario.
Depending on certain circumstances we may be processing at the same time 2 disbursements for the same students, for instance, assuming that we have two applications as below.

**Application A**:  Jan, 1st - April, 30th
**Application B**: May, 1st - July, 31st.
Wonder that Application A can be reassessed near the offering end date on April, 30th and it may be having the e-Cert generated on May, 1st, which will lead both to be part of the same e-Cert file.

The e-Cert SQL query will get both records where no BCSL restriction will be in place, but then the first one will be adding the BCSL restriction and the second one must be aware of it to stop the BC funding.

When we have 2 records for the same student, we may need to have a way to refresh the restrictions in a way that the second disbursement will not have the BC fundings disbursed.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -332,8 +439,35 @@ export class ECertGenerationService {
...
         }, 0);
+      // Calculate total BC grants subtracted due to restriction.
+      let bcAmountSubtractedRestrictionId = NaN;
+      const bcTotalGrantAmountSubtracted =
+        disbursementSchedule.disbursementValues
+          // Filter all BC grants and filter out the null BC grants subtracted.
+          .filter((disbursementValue) => {
+            if (
+              disbursementValue.valueType === DisbursementValueType.BCGrant &&
+              disbursementValue.restrictionAmountSubtracted !== null
```
andrewsignori-aot, line 450: Why do we need additional checks here? The idea of this method was to be concerned only about doing the SUM of the BC Grants if there is a value to be ignored it probably must be set to zero before this calculation happen.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -332,8 +439,35 @@ export class ECertGenerationService {
         .reduce((previousValue, currentValue) => {
           return previousValue + currentValue.effectiveAmount;
         }, 0);
+      // Calculate total BC grants subtracted due to restriction.
```
andrewsignori-aot, line 442: Is it not supposed to be taken care of during the effective value calculation?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: mm, That is a good edge case, are we managing a similar scenario in overwards.? ya we need to find a way

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: ya, get the whole list and assign the latest restriction to the bc grants and loans when needed?..mm ok...actually, I was trying to avoid the number of queries. But i can give it a try

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: I added it here because the total bc grant itself was here at this stage, after the effective calculation
image

sources/packages/backend/apps/workers/src/controllers/disbursement/disbursement.controller.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from "zeebe-node";
 // TODO: In the upcoming tasks, either DisbursementScheduleService will be renamed at shared library
 // or MSFAA related methods will move to shared library.
-import { DisbursementScheduleService as DisbursementScheduleSharedService } from "@sims/services";
```
dheepak-aot, line 11: 👍

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
```diff
@@ -281,27 +306,109 @@ export class ECertGenerationService {
...
    */
-  private calculateEffectiveValue(disbursements: ECertDisbursementSchedule[]) {
+  private async calculateEffectiveValue(
+    disbursements: ECertDisbursementSchedule[],
+    entityManager: EntityManager,
+  ): Promise<void> {
     for (const disbursement of disbursements) {
       for (const disbursementValue of disbursement.disbursementValues) {
-        if (this.shouldStopFunding(disbursement, disbursementValue)) {
+        if (this.shouldStopFullTimeBCFunding(disbursement, disbursementValue)) {
```
dheepak-aot, line 317: I believe the method here must be more Generic in name(as it was before) and action, not just to stop full time and only BC funding.

sources/packages/backend/libs/services/src/sfas/sfas-application.service.ts
```diff
@@ -58,6 +58,31 @@ export class SFASApplicationService extends DataModelService<SFASApplication> {
...
+  async totalLegacyBCSLAmount(
+    sin: string,
+    birthDate: string,
+    lastName: string,
+  ): Promise<number> {
+    const totalQuery = await this.repo
+      .createQueryBuilder("sfasApplication")
+      .select("SUM(sfasApplication.bslAward)")
+      .innerJoin("sfasApplication.individual", "sfasFTstudent")
+      .where("lower(sfasFTstudent.lastName) = lower(:lastName)", { lastName })
```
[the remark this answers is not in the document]
ann-aot: nice point

sources/packages/backend/libs/services/src/sfas/sfas-application.service.ts
```diff
@@ -58,6 +58,31 @@ export class SFASApplicationService extends DataModelService<SFASApplication> {
...
+   * @param birthDate birthdate of the student.
+   * @param lastName last name of the student.
+   * @returns total BCSL amount that the student received from the legacy(sfas) system.
+   */
+  async totalLegacyBCSLAmount(
+    sin: string,
+    birthDate: string,
+    lastName: string,
+  ): Promise<number> {
+    const totalQuery = await this.repo
```
dheepak-aot, line 75: The totalQuery could be typed.  Currently it shows of type any.

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(dataSource: DataSource) {
+    super(dataSource.getRepository(Restriction));
+  }
+
+  /**
+   * Fetch the restriction with requested restriction code.
+   * @param restrictionCode restriction code.
+   * @returns restriction.
+   */
+  async getRestrictionByCode(
```
dheepak-aot, line 20: typeorm find instead of queryBuilder?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: @dheepak-aot  @guru-aot @andrepestana-aot  @sh16011993  I and @andrewsignori-aot were discussing this and for the PR I am planning to add a directory that will be updated when a student gets a restriction for the first time, and as this directory as the reference the next disbursement of the same student will forwarded to execute restriction logic

sources/packages/backend/libs/services/src/disbursement-schedule/disbursement-schedule-shared.service.ts
```diff
@@ -671,4 +671,31 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
...
     return [...new Set(awards.map((award) => award.valueCode))];
   }
+
+  /**
+   * Total BCSL amount that the student received from the SIMS.
+   * @param studentId student id.
+   * @returns total BCSL amount that the student received from the SIMS.
+   */
+  async totalDisbursedBCSLAmount(studentId: number): Promise<number> {
+    const totalQuery = await this.repo
```
dheepak-aot, line 681: could be typed and variable renamed not to have Query in it?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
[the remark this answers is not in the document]
ann-aot: @dheepak-aot  that is not part of the PR, The first 2 order is already implemented in our system, this PR is implementing the 3rd order and some changes in the 2nd order. That point for just to make sure we follow the order

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: https://github.com/bcgov/SIMS/pull/1886#discussion_r1171972543 but I have added 
image

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: removed the logic got total bc grant

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: https://github.com/bcgov/SIMS/pull/1886#discussion_r1171973155

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
ann-aot: As discussed we will handle this in a separate ticket

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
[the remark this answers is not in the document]
dheepak-aot: As discussed, can slightly add context to first point in the story like it is already happening in ecert disbursements selection criteria.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
guru-aot: Lets handle it in a separate ticket, can we add a TODO here?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-generation.service.ts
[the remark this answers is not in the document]
ann-aot: Updated the story, as discssed

sources/packages/backend/libs/services/src/disbursement-schedule/disbursement-schedule-shared.service.ts
```diff
@@ -671,4 +671,31 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
   private getDistinctValueCodes(awards: DisbursementValue[]): string[] {
     return [...new Set(awards.map((award) => award.valueCode))];
   }
+
+  /**
+   * Total BCSL amount that the student received from the SIMS.
+   * @param studentId student id.
+   * @returns total BCSL amount that the student received from the SIMS.
+   */
+  async totalDisbursedBCSLAmount(studentId: number): Promise<number> {
```
andrewsignori-aot, line 680: As part of the upcoming ticket, we can also discuss the possibility of stopping doing the calculation every time and maybe have the SUM for the students consolidated in some other table on SIMS.

sources/packages/backend/libs/services/src/sfas/sfas-application.service.ts
```diff
@@ -58,6 +58,21 @@ export class SFASApplicationService extends DataModelService<SFASApplication> {
       .getOne();
   }
 
+  /**
+   * Total BCSL amount that the student received from the legacy(SFAS) system.
+   * @param studentId student id.
+   * @returns total BCSL amount that the student received from the legacy(sfas) system.
+   */
+  async totalLegacyBCSLAmount(studentId: number): Promise<number> {
```
andrewsignori-aot, line 66: As part of the upcoming ticket, we can also discuss the possibility of stopping doing the calculation every time and maybe have the SUM for the students consolidated in some other table on SIMS.

sources/packages/backend/libs/sims-db/src/entities/student.model.ts
```diff
@@ -82,4 +84,21 @@ export class Student extends RecordDataModel {
...
+  studentRestrictionsIds: number[];
+  /**
+   * List of all student restrictions.
+   */
+  @OneToMany(
+    () => StudentRestriction,
+    (studentRestriction) => studentRestriction.student,
+    {
+      eager: false,
+      cascade: true,
```
andrewsignori-aot, line 100: Since this is new I believe that we can have the cascade as `false` or `["update"]`. Does it make sense?

andrewsignori-aot (approved):

andrepestana-aot (approved):
LGTM.

Issue #91, opened 2021-01-25
Ministry User is Authenticated with IDIR


mzmichellesmith:
**As a** *Ministry User)* **I want** *(my IDIR login credentials to authenticated)* **So That** *(only authorized individuals can access Student Information

**Additional Context**
- enter text here
- enter text here

**Acceptance Criteria**
- [ ] Given (A ministry User), When (passes the authentication), Then (the user will proceed to SIMS)
- [ ] Given (A ministry User), When (does not pass authentication), Then (the system will display a standard error message from IDIR and will not proceed to SIMS)
- [ ] Given (A Ministry User), When (who has previously failed authentication but then passes with subsequent login), Then (the Ministry User is allowed to navigate to SIMS

PR #1887, opened 2023-04-13
#1786 - Enable Search Student for Institutions part 2

andrepestana-aot:
- Added BC Public institution (COLLF) to test seed data;
- Added COLLF users to test-utils;
- Adjusted query to not get either Cancelled applications that weren't assessed or Draft applications;
- Persisted institution location at the submit time;
- Removed application update for institution location from PIR process;
- Updated comment on sims.applications.location_id;
- Added E2E testes for the search student API.
- Fixed responsiveness for the form:
image
image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
@@ -29,7 +29,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
   let collegeFLocation: InstitutionLocation;
   let studentRepo: Repository<Student>;
   let applicationRepo: Repository<Application>;
-  let collegeFInstitutionUserToken;
+  let collegeFInstitutionUserToken: string;
   const endpoint = "/institutions/student/search";
 
   beforeAll(async () => {
@@ -374,6 +374,11 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .expect([]);
   });
 
+  /**
+   * Saves a student and an application for the student for College F.
+   * @param applicationStatus application status.
+   * @returns an object with student and application persisted in the database.
+   */
   const saveStudentWithApplicationForCollegeF = async (
     applicationStatus: ApplicationStatus,
   ) => {
--- a/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
@@ -11,6 +11,7 @@ import {
 } from "../../../../testHelpers";
 import {
   createFakeInstitutionLocation,
+  saveFakeApplication,
   saveFakeApplicationDisbursements,
   saveFakeStudent,
 } from "@sims/test-utils";
@@ -52,14 +53,9 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
   it("Should find the student by application number when student has at least one application submitted for the institution.", async () => {
     // Arrange
     // Student who has application submitted to institution.
-    let student = await saveFakeStudent(appDataSource);
-    // Refresh student object to get a birthDate with date only
-    student = await studentRepo.findOne({
-      where: { id: student.id },
-      relations: { sinValidation: true },
-    });
+    const student = await saveFakeStudent(appDataSource);
 
-    const application = await saveFakeApplicationDisbursements(
+    const application = await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -86,7 +82,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([
         {
           id: student.id,
@@ -101,14 +97,9 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
   it("Should find the student by last name when student has at least one application submitted for the institution.", async () => {
     // Arrange
     // Student who has application submitted to institution.
-    let student = await saveFakeStudent(appDataSource);
-    // Refresh student object to get a birthDate with date only
-    student = await studentRepo.findOne({
-      where: { id: student.id },
-      relations: { sinValidation: true },
-    });
+    const student = await saveFakeStudent(appDataSource);
 
-    await saveFakeApplicationDisbursements(
+    await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -135,7 +126,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .then((response) => {
         expect(response.body).toEqual(
           expect.arrayContaining([
@@ -154,14 +145,9 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
   it("Should find the student by first name when student has at least one application submitted for the institution.", async () => {
     // Arrange
     // Student who has application submitted to institution.
-    let student = await saveFakeStudent(appDataSource);
-    // Refresh student object to get a birthDate with date only
-    student = await studentRepo.findOne({
-      where: { id: student.id },
-      relations: { sinValidation: true },
-    });
+    const student = await saveFakeStudent(appDataSource);
 
-    await saveFakeApplicationDisbursements(
+    await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -188,7 +174,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .then((response) => {
         expect(response.body).toEqual(
           expect.arrayContaining([
@@ -207,14 +193,9 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
   it("Should find the student by sin when student has at least one application submitted for the institution.", async () => {
     // Arrange
     // Student who has application submitted to institution.
-    let student = await saveFakeStudent(appDataSource);
-    // Refresh student object to get a birthDate with date only
-    student = await studentRepo.findOne({
-      where: { id: student.id },
-      relations: { sinValidation: true },
-    });
+    const student = await saveFakeStudent(appDataSource);
 
-    await saveFakeApplicationDisbursements(
+    await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -241,7 +222,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([
         {
           id: student.id,
@@ -258,7 +239,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
     // Student who has application submitted to institution.
     const student = await saveFakeStudent(appDataSource);
 
-    const application = await saveFakeApplicationDisbursements(
+    const application = await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -289,21 +270,16 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([]);
   });
 
   it("Should find the student when application submitted and cancelled after the assessment.", async () => {
     // Arrange
     // Student who has application submitted to institution.
-    let student = await saveFakeStudent(appDataSource);
-    // Refresh student object to get a birthDate with date only
-    student = await studentRepo.findOne({
-      where: { id: student.id },
-      relations: { sinValidation: true },
-    });
+    const student = await saveFakeStudent(appDataSource);
 
-    const application = await saveFakeApplicationDisbursements(
+    const application = await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -331,7 +307,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([
         {
           id: student.id,
@@ -377,7 +353,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([]);
   });
 
@@ -397,7 +373,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       InstitutionTokenTypes.CollegeCUser,
       collegeCLocation,
     );
-    const application = await saveFakeApplicationDisbursements(
+    const application = await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -424,7 +400,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([]);
   });
 
@@ -433,7 +409,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
     // Student who has application submitted to institution.
     const student = await saveFakeStudent(appDataSource);
 
-    const application = await saveFakeApplicationDisbursements(
+    const application = await saveFakeApplication(
       appDataSource,
       {
         institution: collegeF,
@@ -460,7 +436,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       .post(endpoint)
       .send(searchPayload)
       .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.CREATED)
+      .expect(HttpStatus.OK)
       .expect([]);
   });
 
--- a/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
@@ -12,7 +12,6 @@ import {
 import {
   createFakeInstitutionLocation,
   saveFakeApplication,
-  saveFakeApplicationDisbursements,
   saveFakeStudent,
 } from "@sims/test-utils";
 import {
@@ -30,6 +29,8 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
   let collegeFLocation: InstitutionLocation;
   let studentRepo: Repository<Student>;
   let applicationRepo: Repository<Application>;
+  let collegeFInstitutionUserToken;
+  const endpoint = "/institutions/student/search";
 
   beforeAll(async () => {
     const { nestApplication, dataSource } = await createTestingAppModule();
@@ -48,28 +49,15 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       InstitutionTokenTypes.CollegeFUser,
       collegeFLocation,
     );
+    collegeFInstitutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeFUser,
+    );
   });
 
   it("Should find the student by application number when student has at least one application submitted for the institution.", async () => {
     // Arrange
-    // Student who has application submitted to institution.
-    const student = await saveFakeStudent(appDataSource);
-
-    const application = await saveFakeApplication(
-      appDataSource,
-      {
-        institution: collegeF,
-        institutionLocation: collegeFLocation,
-        student,
-      },
-      {
-        applicationStatus: ApplicationStatus.Submitted,
-      },
-    );
-    const endpoint = "/institutions/student/search";
-    const institutionUserToken = await getInstitutionToken(
-      InstitutionTokenTypes.CollegeFUser,
-    );
+    const { student, application } =
+      await saveStudentWithApplicationForCollegeF(ApplicationStatus.Submitted);
     const searchPayload = {
       appNumber: application.applicationNumber,
       firstName: "",
@@ -81,7 +69,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
     await request(app.getHttpServer())
       .post(endpoint)
       .send(searchPayload)
-      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .auth(collegeFInstitutionUserToken, BEARER_AUTH_TYPE)
       .expect(HttpStatus.OK)
       .expect([
         {
@@ -96,23 +84,8 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
 
   it("Should find the student by last name when student has at least one application submitted for the institution.", async () => {
     // Arrange
-    // Student who has application submitted to institution.
-    const student = await saveFakeStudent(appDataSource);
-
-    await saveFakeApplication(
-      appDataSource,
-      {
-        institution: collegeF,
-        institutionLocation: collegeFLocation,
-        student,
-      },
-      {
-        applicationStatus: ApplicationStatus.Submitted,
-      },
-    );
-    const endpoint = "/institutions/student/search";
-    const institutionUserToken = await getInstitutionToken(
-      InstitutionTokenTypes.CollegeFUser,
+    const { student } = await saveStudentWithApplicationForCollegeF(
+      ApplicationStatus.Submitted,
     );
     const searchPayload = {
       appNumber: "",
@@ -125,7 +98,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
     await request(app.getHttpServer())
       .post(endpoint)
       .send(searchPayload)
-      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .auth(collegeFInstitutionUserToken, BEARER_AUTH_TYPE)
       .expect(HttpStatus.OK)
       .then((response) => {
         expect(response.body).toEqual(
@@ -142,30 +115,19 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
       });
   });
 
-  it("Should find the student by first name when student has at least one application submitted for the institution.", async () => {
+  it("Should find the student by part of last name when student has at least one application submitted for the institution.", async () => {
     // Arrange
-    // Student who has application submitted to institution.
-    const student = await saveFakeStudent(appDataSource);
+    const { student } = await saveStudentWithApplicationForCollegeF(
+      ApplicationStatus.Submitted,
+    );
+    student.user.lastName =
+      "last name ef1c3bd5-4e8d-4ac4-adcb-3507a1cc7d43 ilike test";
+    await studentRepo.save(student);
 
-    await saveFakeApplication(
-      appDataSource,
-      {
-        institution: collegeF,
-        institutionLocation: collegeFLocation,
-        student,
-      },
-      {
-        applicationStatus: ApplicationStatus.Submitted,
-      },
-    );
-    const endpoint = "/institutions/student/search";
-    const institutionUserToken = await getInstitutionToken(
-      InstitutionTokenTypes.CollegeFUser,
-    );
     const searchPayload = {
       appNumber: "",
-      firstName: student.user.firstName,
-      lastName: "",
+      firstName: "",
+      lastName: "EF1C3BD5-4E8D-4AC4-ADCB-3507A1CC7D43",
       sin: "",
     };
 
@@ -173,55 +135,7 @@ describe("StudentInstitutionsController(e2e)-searchStudents", () => {
     await request(app.getHttpServer())
       .post(endpoint)
       .send(searchPayload)
-      .auth(institutionUserToken, BEARER_AUTH_TYPE)
-      .expect(HttpStatus.OK)
-      .then((response) => {
-        expect(response.body).toEqual(
-          expect.arrayContaining([
-            {
-              id: student.id,
-              firstName: student.user.firstName,
-              lastName: student.user.lastName,
-              birthDate: student.birthDate,
-              sin: student.sinValidation.sin,
-            },
-          ]),
-        );
-      });
-  });
-
-  it("Should find the student by sin when student has at least one application submitted for the institution.", async () => {
-    // Arrange
-    // Student who has application submitted to institution.
-    const student = await saveFakeStudent(appDataSource);
-
-    await saveFakeApplication(
-      appDataSource,
-      {
-        institution: collegeF,
-        institutionLocation: collegeFLocation,
-        student,
-      },
-      {
-        applicationStatus: ApplicationStatus.Submitted,
-      },
-    );
-    const endpoint = "/institutions/student/search";
-    const institutionUserToken = await getInstitutionToken(
-      InstitutionTokenTypes.CollegeFUser,
-    );
-    const searchPayload = {
-      appNumber: "",
-      firstName: "",
-      lastName: "",
-      sin: student.sinValidation.sin,
-    };
-
-    // Act/Assert
-    await request(app.getHttpServer())
-      .post(endpoint)
-      .send(searchPayload)
-      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .auth(collegeFInstitutionUserToken, BEARER_AUTH_TYPE)
       .expect(HttpStatus.OK)
       .expect([
         {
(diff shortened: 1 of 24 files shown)


sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+      collegeFLocation,
+    );
+  });
+
+  it("Should find the student by application number when student has at least one application submitted for the institution.", async () => {
+    // Arrange
+    // Student who has application submitted to institution.
+    let student = await saveFakeStudent(appDataSource);
+    // Refresh student object to get a birthDate with date only
+    student = await studentRepo.findOne({
```
andrewsignori-aot, line 57: Can the method `createFakeStudent` be fixed to receive the date as expected avoiding the reload here?
```ts
student.birthDate = getISODateOnlyString(faker.date.past(18));
```

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+      InstitutionTokenTypes.CollegeFUser,
+      collegeFLocation,
+    );
+  });
+
+  it("Should find the student by application number when student has at least one application submitted for the institution.", async () => {
+    // Arrange
+    // Student who has application submitted to institution.
+    let student = await saveFakeStudent(appDataSource);
+    // Refresh student object to get a birthDate with date only
```
ann-aot, line 56: period

ann-aot, line 56: mm, we need to refresh, we are getting saved student object right

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+      },
+    );
+    const endpoint = "/institutions/student/search";
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeFUser,
+    );
+    const searchPayload = {
+      appNumber: "",
+      firstName: "",
+      lastName: student.user.lastName,
```
andrewsignori-aot, line 129: I believe that the test should control the `student.user.lastName` instead of relying on `faker.name.lastName()`. If two users are added to the DB with the same last name this test will fail.
The last name can be set even using a guide to ensure uniqueness.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+        applicationStatus: ApplicationStatus.Submitted,
+      },
+    );
+    const endpoint = "/institutions/student/search";
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeFUser,
+    );
+    const searchPayload = {
+      appNumber: "",
+      firstName: student.user.firstName,
```
andrewsignori-aot, line 181: Same from https://github.com/bcgov/SIMS/pull/1887/files#r1167217144

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+    );
+    const endpoint = "/institutions/student/search";
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeFUser,
+    );
+    const searchPayload = {
+      appNumber: "",
+      firstName: "",
+      lastName: "",
+      sin: student.sinValidation.sin,
```
andrewsignori-aot, line 236: Similar to https://github.com/bcgov/SIMS/pull/1887/files#r1167213944

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
ann-aot: got it. resolving

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+        },
+      ]);
+  });
+
+  it("Should not find the student when application submitted and cancelled before the assessment.", async () => {
+    // Arrange
+    // Student who has application submitted to institution.
+    const student = await saveFakeStudent(appDataSource);
+
+    const application = await saveFakeApplicationDisbursements(
```
andrewsignori-aot, line 261: Other `saveFakeApplicationSomething` helpers can be created. In this case, we are calling a method to create the assessments/disbursements/values and then removing them.
For this particular case, maybe the `createFakeApplication` would be enough.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+      lastName: "",
+      sin: "",
+    };
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(searchPayload)
+      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.CREATED)
```
andrewsignori-aot, line 463: Can we "fix" the HTTP return code for the search endpoint executing the post similar to what was done below?
image

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrewsignori-aot, line 181: Can we have one more scenario to test the search with `ILIKE` for a first name or last name?
The last name can be defined as "last name ef1c3bd5-4e8d-4ac4-adcb-3507a1cc7d43 ilike test" and the search input can be something like "EF1C3BD5-4E8D-4AC4-ADCB-3507A1CC7D43".

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+          id: student.id,
+          firstName: student.user.firstName,
+          lastName: student.user.lastName,
+          birthDate: student.birthDate,
+          sin: student.sinValidation.sin,
+        },
+      ]);
+  });
+
+  it("Should not find the student when user is not active.", async () => {
```
ann-aot, line 346: mm... just a general business thought, not related to the PR,  shouldn't we allow a search for nonactive students too at least for ministry or ministry just didn't care @andrepestana-aot  @andrewsignori-aot  @dheepak-aot @guru-aot @JasonCTang

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -186,6 +187,9 @@ export class ApplicationService extends RecordDataModelService<Application> {
       application.studentAssessments = [originalAssessment];
       application.currentAssessment = originalAssessment;
       application.submittedDate = now;
+      application.location = {
```
andrewsignori-aot, line 190: Not related to this PR.
We need to set the location here in two places because of the "application submit" and "application edit". Are we not supposed to do the same with the offering set at line 152? I may be missing something 😄 
@ann-aot @dheepak-aot @guru-aot thoughts?

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -487,15 +484,29 @@ export class StudentService extends RecordDataModelService<Student> {
...
       .where("user.isActive = true");
     if (institutionId) {
-      searchQuery.andWhere(
-        new Brackets((qb) => {
-          qb.where("offeringInstitution.id = :institutionId", {
-            institutionId,
-          }).orWhere("pirInstitution.id = :institutionId", { institutionId });
-        }),
-      );
+      searchQuery
```
andrewsignori-aot, line 487: To fix the point about "Adjusted query to not get either Cancelled applications that weren't assessed or Draft applications;", is it not the case to have `.leftJoin("application.currentAssessment", "studentAssessment")` converted to a `innerJoin` and have it moved inside the `if (institutionId)`?

sources/packages/backend/apps/db-migrations/src/sql/Applications/Update-comment-location-id.sql
```diff
@@ -0,0 +1 @@
+COMMENT ON COLUMN sims.applications.location_id IS 'References the location related to the application at the submission. From the location it is possible to get the institution for example.';
```
andrewsignori-aot, line 1: Thanks for adjusting the comment ❤️

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: did you mean in the draft and submit?. the offering is to assigned to `originalAssessment` before the condition itself, right (line 163) so, it will be updated in both conditions. Not sure, if if i answered the qn or missed anything

sources/packages/backend/apps/workers/src/controllers/program-info-request/program-info-request.controller.ts
andrewsignori-aot, line 1: Please remove the `STUDENT_DATA_SELECTED_LOCATION` from the `fetchVariable`.
Ensure that `ProgramInfoRequestJobInDTO` is adjusted.
`STUDENT_DATA_SELECTED_LOCATION` can also be removed from the code if no longer needed.

sources/packages/backend/apps/api/src/services/student/student.service.ts
ann-aot: mm,so just to refresh,  so if an application had an exception/cra validation and is in progress and the student cancels the application. shouldn't the student be seen when searched?

sources/packages/web/src/components/common/SearchStudents.vue
```diff
@@ -1,60 +1,66 @@
...
-            data-cy="appNumber"
-            variant="outlined"
-            v-model="appNumber"
-            @keyup.enter="searchStudents"
-            hide-details
-          />
+      <v-row>
+        <v-col cols="11" lg="11" md="10" sm="12">
+          <v-row
+            ><v-col cols="12" lg="3" md="6">
```
ann-aot, line 7: actually a total of `md ` is more than 12. I think since you gave `lg`, it's working for you. I believe either one is enough.
since we are using `md` everywhere else and to keep it consistent. maybe we can go for `md`. also fo line 5

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrepestana-aot: if two users have the same last name and also an application for this same institution, it would show two objects in the array. As we're comparing using `expect.arrayContaining()` that wouldn't fail. What I could do to ensure that we're getting only the object that is important for the test is to search for the application number and last name for example. Do you think I should add another test searching for 2 or more student properties?

ann-aot (commented):
Good work @andrepestana-aot . please take a look at the comment. Not a blocker, just a general qn that came to our mind, maybe for the business or devs, while institutions searching for a student, should they be able to search for an archived student? @JasonCTang  @andrewsignori-aot @dheepak-aot  @guru-aot

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: You are right @andrewsignori-aot. Like the way how offering is done, this assignment must happen before line 165 to make sure both scenarios Draft->Submitted and Submit->ReSubmit to be handled.

image

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -186,6 +187,9 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        id: applicationData.selectedLocation,
+      } as InstitutionLocation;
```
dheepak-aot, line 192: Should we directly trust the form data and assign it to the location?? as a end user can always modify the submitted form data. In case of offerings we validate it with DB before assigning it to the application. 

image

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -469,12 +469,9 @@ export class StudentService extends RecordDataModelService<Student> {
...
         )
-        .leftJoin("application.location", "pirLocation")
-        .leftJoin("pirLocation.institution", "pirInstitution")
-        .leftJoin("application.currentAssessment", "studentAssessment")
-        .leftJoin("studentAssessment.offering", "offering")
-        .leftJoin("offering.institutionLocation", "offeringLocation")
-        .leftJoin("offeringLocation.institution", "offeringInstitution");
+        .leftJoin("application.location", "institutionLocation")
+        .leftJoin("institutionLocation.institution", "institution")
+        .leftJoin("application.currentAssessment", "studentAssessment");
```
dheepak-aot, line 474: @andrepestana-aot @andrewsignori-aot @ann-aot @guru-aot 

Considering the context of the story, is it not sufficient to consider  a student having an application submitted with it's location->institution matching the institution who is performing the search and application not being in overwritten status? if my assumption is right we can ignore the join with currentAssessment right? let me know if I am missing something. 

image

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -487,15 +484,29 @@ export class StudentService extends RecordDataModelService<Student> {
...
-      searchQuery.andWhere(
-        new Brackets((qb) => {
-          qb.where("offeringInstitution.id = :institutionId", {
-            institutionId,
-          }).orWhere("pirInstitution.id = :institutionId", { institutionId });
-        }),
-      );
+      searchQuery
+        .andWhere("institution.id = :institutionId", { institutionId })
+        .andWhere(
```
dheepak-aot, line 489: If my assumption is right in https://github.com/bcgov/SIMS/pull/1887/files#r1167302495 we may not need these conditions.

sources/packages/backend/apps/db-migrations/src/sql/Applications/Update-comment-location-id.sql
dheepak-aot, line 1: Thanks for adjusting the comment.

As a suggestion, it could be simplified as `References the institution location of the application.`

sources/packages/web/src/components/common/SearchStudents.vue
```diff
@@ -1,60 +1,66 @@
...
-            label="Application number"
-            density="compact"
-            data-cy="appNumber"
-            variant="outlined"
-            v-model="appNumber"
-            @keyup.enter="searchStudents"
-            hide-details
-          />
+      <v-row>
+        <v-col cols="11" lg="11" md="10" sm="12">
```
andrewsignori-aot, line 5: IMO I was expecting no need to use the cols at this level to achieve the responsiveness here. Now looking at how the code was I was able to understand what was forcing the text inputs to not break in multiple lines.
As a suggestion, I believe that a similar result can be achieved as with the below template.
```vue
<v-row>
        <v-col>
          <v-row class="search">
            <v-text-field
              label="Application Number"
              density="compact"
              data-cy="appNumber"
              variant="outlined"
              v-model="appNumber"
              @keyup.enter="searchStudents"
              hide-details
            />
            <v-text-field
              label="SIN"
              density="compact"
              data-cy="sin"
              variant="outlined"
              v-model="sin"
              @keyup.enter="searchStudents"
              hide-details
            />
            <v-text-field
              label="Given names"
              density="compact"
              data-cy="firstName"
              variant="outlined"
              v-model="firstName"
              @keyup.enter="searchStudents"
              hide-details
            />
            <v-text-field
              label="Last name"
              density="compact"
              data-cy="lastName"
              variant="outlined"
              v-model="lastName"
              @keyup.enter="searchStudents"
              hide-details
            />
          </v-row>
        </v-col>
        <v-col cols="auto">
          <v-btn
            color="primary"
            class="p-button-raised"
            data-cy="searchStudents"
            @click="searchStudents()"
          >
            Search
          </v-btn>
        </v-col>
      </v-row>
```

The class used is the below one:
```scss
<style lang="scss">
.search {
  padding: 5px;
  .v-input {
    min-width: 120px;
    margin: 5px;
  }
}
</style>
```

andrewsignori-aot (commented):
Great work and nice E2E scenarios coverage. Please take a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrewsignori-aot: I would still prefer to have a unique name and avoid using the `expect.arrayContaining()`. Open for discussion, but it would be simpler (IMO) and it would also guarantee the test output.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrewsignori-aot: As far as I can remember the concept of active users is in place only for institution users.
I cannot recall if or when we would deactivate a student user, even if we do, the system is not prepared to deal with it.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: @dheepak-aot my comment was about what @ann-aot replied. I missed the point that the `originalAssessment` is used in both scenarios.
About your comment @dheepak-aot, if I followed correctly, in case an "application edit" is happening, a new application object (`newApplication`) is created and we need to populate it also, unless I am missing something again 😉

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: Yes @ann-aot, but I do not see how my comment would change it. My suggestion would still return any application that had an assessment created (on Submit) even if the workflow was never executed.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: We can follow what @dheepak-aot suggested to avoid letting the DB throw the exception.

sources/packages/web/src/components/common/SearchStudents.vue
andrepestana-aot: Thanks for the suggestion, but with the code above depending on the width the layout can be weird like this:
image

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: @dheepak-aot if your suggestion is about checking the `application.location` and, because `application.location` is only set for submitted applications, we can rely on that, it may be acceptable.
Checking if the assessment record was present is juts, for instance, to exclude a possible canceled draft application.

andrepestana-aot: The join with currentAssessment if only for the case of the cancelled applications. The student can create a draft application and cancel it and we want to avoid this case to be found by the institution. So, to find only the cancelled ones that have been submitted, we check if the application is cancelled AND had an assessment.

sources/packages/web/src/components/common/SearchStudents.vue
andrewsignori-aot: If the "weird" part is related to the application filed, it can be adjusted with the width, as shown below. It the "weird" is because the last name takes up all the space then it will need more work.
Either way, the idea was to suggest a different approach where there was no need to have every single input wrapped with a `v-col`. If the current solution is working no need to spend more time on this.

image

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: @andrepestana-aot if the application is created and canceled while in the draft we are not setting the `application.location`, right?

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: Oh, yes. I completely missed it. Got it now.

sources/packages/backend/apps/api/src/services/student/student.service.ts
dheepak-aot: @andrewsignori-aot @andrepestana-aot A cancelled draft application will not have locationId right? as it is populated on submission, so it will be automatically excluded right?

Let me know if I am missing something.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+    // Arrange
+    // Student who has application submitted to institution.
+    let student = await saveFakeStudent(appDataSource);
+    // Refresh student object to get a birthDate with date only
+    student = await studentRepo.findOne({
+      where: { id: student.id },
+      relations: { sinValidation: true },
+    });
+
+    const application = await saveFakeApplicationDisbursements(
```
dheepak-aot, line 62: If we use only the locationId, as per https://github.com/bcgov/SIMS/pull/1887/files#r1167302495 then we can just save an application with which is created through 
```ts
const application = createFakeApplication({
      location: collegeFLocation,
      student,
    });
```

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: Yes, you're right. Also, we can do inner joins to location and institution.
They are already inside the `if(institutionId)`.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+            },
+          ]),
+        );
+      });
+  });
+
+  it("Should find the student by sin when student has at least one application submitted for the institution.", async () => {
+    // Arrange
+    // Student who has application submitted to institution.
+    let student = await saveFakeStudent(appDataSource);
```
dheepak-aot, line 210: The arrange part seems to be repetitive with same code for many scenarios. we can create a common method within this .ts file ?

dheepak-aot:
Good work @andrepestana-aot . Please have a look at the comments.

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: @dheepak-aot this was exactly what I meant to say with my comment above, so yes, I believe that we are on the same page 😉

dheepak-aot:
> Good work @andrepestana-aot . please take a look at the comment. Not a blocker, just a general qn that came to my mind, maybe for the business or devs, while institutions search for a student, should they be able to see an archived student application? @JasonCTang @andrewsignori-aot @dheepak-aot @guru-aot

Good question @ann-aot . I am curious to see the reply.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrepestana-aot: The assert validates the student id, the student sin and student birthDate as well. There's no way of all student properties to be equal and we get the wrong student in the assert.  What can happen is to have additional users in the response but it is ok IMO.

sources/packages/backend/apps/api/src/services/student/student.service.ts
ann-aot: no @andrewsignori-aot, it was not about the comment. it was to refresh the logic. Thanks 👍

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
ann-aot: mm... I was asking because I see a logic checking the user is active in the `seachStudent` function, 
image

sources/packages/backend/apps/api/src/services/student/student.service.ts
ann-aot: @dheepak-aot I agree. I was also in the same page 👍

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrepestana-aot: Yes, I added the DB check for the location. Thanks!

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrepestana-aot: I was adding the locationId as a Draft too, but I see what you guys are saying and I'm going to change then and change the query as well. Thanks!

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrewsignori-aot: @ann-aot All that I can say is that I do not remember any requirement for active/inactive user checks for students.
That said, we do have `ActiveUserGuard` that checks in general if a user accessing the API is active or not, but I am not sure about the search.
Below is the ticket that enabled the search for students and added the mentioned condition and it does not state anything about isActive. Maybe it was a decision taken by the Team during the time or maybe we have another ticket that change it later. 
https://github.com/bcgov/sims/issues/91

Which would be the feature that would make the student user (current associated one) be set as inactive?

sources/packages/backend/apps/api/src/services/student/student.service.ts
andrewsignori-aot: @andrepestana-aot were you adding it to the draft? I saw the code doing it on the transition from "Draft->Submit" and this will not change. The application draft is saved in a different API endpoint.

andrepestana-aot: yes, you're right @andrewsignori-aot . I got confused.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrepestana-aot: I create a `saveFakeApplication()` that can be called just for these cases when you need a application. I refactored `saveFakeApplicationDisbursements` to call it.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,470 @@
...
+      {
+        applicationStatus: ApplicationStatus.Submitted,
+      },
+    );
+    const endpoint = "/institutions/student/search";
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeFUser,
+    );
+    const searchPayload = {
+      appNumber: application.applicationNumber,
```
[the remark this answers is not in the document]
andrepestana-aot: A duplicate application number alphanumeric with 10 digits can happen but it is such a remote possibility and there's a unique key for application number in the database that would shows us the cause in case that happened. I'm not sure it is worth to invest in that approach.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
andrepestana-aot: I changed this part to save the application without disbursements.

guru-aot:
> Good work @andrepestana-aot . please take a look at the comment. Not a blocker, just a general qn that came to my mind, maybe for the business or devs, while institutions search for a student, should they be able to see an archived student application? @JasonCTang @andrewsignori-aot @dheepak-aot @guru-aot

I am not sure whether to see an archived student came as a requirement for us. the better one to answer would be @JasonCTang and business

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -469,12 +469,8 @@ export class StudentService extends RecordDataModelService<Student> {
...
           "application.student.id = student.id",
         )
-        .leftJoin("application.location", "pirLocation")
-        .leftJoin("pirLocation.institution", "pirInstitution")
-        .leftJoin("application.currentAssessment", "studentAssessment")
-        .leftJoin("studentAssessment.offering", "offering")
-        .leftJoin("offering.institutionLocation", "offeringLocation")
-        .leftJoin("offeringLocation.institution", "offeringInstitution");
+        .innerJoin("application.location", "institutionLocation")
+        .innerJoin("institutionLocation.institution", "institution");
```
guru-aot, line 473: 👍

guru-aot (approved):

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -36,11 +38,14 @@ export class StudentInstitutionsController extends BaseController {
...
   /**
    * Search students based on the search criteria.
+   * Return a 200 HTTP status instead of 201 to indicate that the operation
+   * was completed with success but no resource was created.
    * TODO add decorator to restrict to BC Public institutions.
    * @param searchCriteria criteria to be used in the search.
    * @returns searched student details.
    */
   @Post("search")
+  @HttpCode(HttpStatus.OK)
```
ann-aot, line 48: 👍

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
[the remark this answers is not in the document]
andrewsignori-aot: I believe that I am missing something, what is the "unique key for application number" on DB?
Do you mean that, if we try to add two applications with the same application number, the DB will throw an error?

Besides that, about if it is worth it or not, in general, I would prefer not to build the E2E tests based on the "bet" of a remote possibility of them failing when the alternative way can be implemented with a possible quick (minutes or few hours) alternatives.

I will not block this specific PR at this moment, but I believe that the comment is worth it to think about in any upcoming one.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,402 @@
...
+} from "@sims/sims-db";
+
+describe("StudentInstitutionsController(e2e)-searchStudents", () => {
+  let app: INestApplication;
+  let appDataSource: DataSource;
+  let collegeF: Institution;
+  let collegeFLocation: InstitutionLocation;
+  let studentRepo: Repository<Student>;
+  let applicationRepo: Repository<Application>;
+  let collegeFInstitutionUserToken;
```
andrewsignori-aot, line 32: Please type this to avoid VSCode waning.
image

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.searchStudents.e2e-spec.ts
```diff
@@ -0,0 +1,402 @@
...
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(searchPayload)
+      .auth(collegeFInstitutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .expect([]);
+  });
+
+  const saveStudentWithApplicationForCollegeF = async (
```
andrewsignori-aot, line 377: Since this is a new method, hence it would need comments and a return type (which can be an anonymous type to keep it simple), unless there is a team agreement to skip those.
@ann-aot @dheepak-aot @guru-aot @sh16011993

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
andrewsignori-aot, line 48: If you are in the mood, the same can be done on the AEST search if no side effects.

sources/packages/backend/libs/test-utils/src/factories/application.ts
```diff
@@ -140,7 +119,66 @@ export async function saveFakeApplicationDisbursements(
...
+  options?: {
+    applicationStatus?: ApplicationStatus;
+  },
+): Promise<Application> {
+  const userRepo = dataSource.getRepository(User);
+  const studentRepo = dataSource.getRepository(Student);
+  const applicationRepo = dataSource.getRepository(Application);
+  const studentAssessmentRepo = dataSource.getRepository(StudentAssessment);
+  const offeringRepo = dataSource.getRepository(EducationProgramOffering);
+  const applicationStatus =
```
andrewsignori-aot, line 158: This default value made sense while it was in the method saveFakeApplicationDisbursements as per the old comment on top of it.
```ts
// Using Assessment as a default status since it the first status when
// the application has the disbursement already generated.
```
I believe that we can move the default value assessment as it was before to the `saveFakeApplicationDisbursements`.

sources/packages/backend/libs/test-utils/src/factories/application.ts
```diff
@@ -154,5 +192,8 @@ export async function saveFakeApplicationDisbursements(
     fakeOriginalAssessment,
   );
   savedApplication.currentAssessment = savedOriginalAssessment;
+  if (options?.applicationStatus === ApplicationStatus.Draft) {
```
andrewsignori-aot, line 195: Do we really need to make it null at this point? If so, we need also to take care of the `savedApplication.currentAssessment` for instance.
If the problem is related to having an intitutionLocation associated with the offering and application location, we can also receive two locations on the relations options.

andrewsignori-aot (commented):
Thanks for doing the changes. Few minor comments only.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -145,6 +147,16 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
     originalAssessment.submittedDate = now;
     originalAssessment.creator = auditUser;
+    const institutionLocation =
+      await this.institutionLocationService.getInstitutionLocation(
+        applicationData.selectedLocation,
+      );
+    if (!institutionLocation) {
+      throw new CustomNamedError(
+        "Not able to find the institution location.",
+        INSTITUTION_LOCATION_NOT_VALID,
```
dheepak-aot, line 157: This error must be included in the controller to be transformed appropriately. 

image

dheepak-aot:
Thanks for doing the changes @andrepestana-aot. Added one more comment.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -145,6 +147,16 @@ export class ApplicationService extends RecordDataModelService<Application> {
     originalAssessment.submittedBy = auditUser;
     originalAssessment.submittedDate = now;
     originalAssessment.creator = auditUser;
+    const institutionLocation =
```
ann-aot, line 150: I am wondering should this logic be only checked if `applicationData.selectedLocation,` is there or can be moved after the draft logic. because in draft state, a student does have the option to omit the `program` tab and fill other tabs right? And if they do so, this logic will not allow the student to submit the draft application right?

ann-aot: And as per the discussion, I believe we decided not to add location id in the draft state right? unless I am missing something https://github.com/bcgov/SIMS/pull/1887#discussion_r1167328352. so , do we need line 201 and this logic we can move after the draft logic right?

andrewsignori-aot: @ann-aot adding the location for a draft application was never on the table. The code on line 201 is inside the block that converts a draft to a submitted application.
image

ann-aot: oh.. ya... Thanks for that. resolving the comment

andrewsignori-aot (approved):
Thanks for doing the changes and really nice coverage on the search scenarios for E2E tests.

ann-aot (approved):
nICE WORK @andrepestana-aot  Thanks for doing the changes 👍

sh16011993 (approved):
Looks good to me 👍

PR #1888, opened 2023-04-14
#1889 - Zeebe GRPC Error Investigation

andrewsignori-aot:
- Investigation to try to resolve the error "ERROR: Grpc Stream Error: 8 RESOURCE_EXHAUSTED: Expected to activate jobs of type 'some-worker-type'".
- Updated Zeebe server to use version 8.2.1 (https://github.com/camunda/zeebe/releases/tag/8.2.1).
- Updated Zeebe node client to use version 8.1.4.
- Enabled error logs for Workflow E2E tests that were disabled for better readability in a previous PR.

So far there are indications the upgrade of the Zeebe server alone was able to fix the "Grpc Stream Error: 8". The Camunda Cloud clusters on DEV/TEST are now updated and we can monitor from now on if we notice the error. The dev local docker-compose files were also updated which should result in no more errors also.


andrepestana-aot (approved):
👍

dheepak-aot (approved):
Thanks for taking care of the Zeebe issue. 👍

Issue #1864, opened 2023-04-06
Review PT Integration Specifications and confirm our files are compliant.


JasonCTang:
**Describe the task**
Review the finalized PT Integration Specs document and confirm the following files we are already creating are compliant.
- [x] Send PT MSFAA Send File to Finastra
- [x] Receive and process the PT MSFAA Received File
- [x] Send new PT e-Cert file to Finastra
- [x] Receive and process new e-Cert Feedback file

**Additional context**
- [BC Part-Time MSFAA Sent File](https://app.zenhub.com/files/319429847/f277e264-b3ca-4a8a-8d2e-0a9ec548ec02/download)
- [BC Part-Time MSFAA  Received File](https://app.zenhub.com/files/319429847/805e87c6-243f-4b07-be93-bbb15b1efdf9/download)
- [BC Part-Time E-Cert File V 3.0](https://app.zenhub.com/files/319429847/fbeefb54-0e86-4126-a08b-e8b9aa181bfb/download)
- [BC Part-Time E-Cert Feedback File V 1.3](https://app.zenhub.com/files/319429847/358b6ec8-69b0-420c-bb85-03c7b06d24cf/download)

**Acceptance Critera**
- [x] Review finalized specs and compare to the files we are currently generating.
- [x] If we discover that there have been changes to the file format, then modify the files to the new specs. (It's possible there will be zero changes)
- [x] Perform analysis of receiving and processing new e-Cert Feedback file for Part Time.
- [x] Perfom a knowledge transfer session with business area to demonstrate where in the process the files are generated. 
- [x] Review receive and process of the e-Cert Feedback file Part Time
- [x] Save files generated and received as part of the test with Finastra into our Teams channel
- [x] Ensure the country is converted to CAN when Canada for the part-time e-Cert that has a 4-character max field for country information. Other countries will be truncated at 4 characters, for instance, "United States" will be sent as "Unit"

andrewsignori-aot:
## MSFAA Sent File

Questions
- As per the documentation "Gender" is expected to be sent as "Borrower's Gender (F or M)". SIMS is actually sending it as below. Is the type "O" still supported?
    - males: M
    - female: F
    - Not male or female: O

andrewsignori-aot:
## e-Cert File

Questions
- "First Name" is stated in the documentation as mandatory but it will not be true for mononymos names.

Changes
- Originator updated to "NEW PT ENTITLEMENT".
[...]

PR #1890, opened 2023-04-14
#1864 - PT Integration Files Review - MSFAA Send/Request Files

andrewsignori-aot:
- MSFAA sent/received schedulers were named incorrectly (the request was a response and vice-versa). There are no actual changes on files `msfaa-integration/msfaa-part-time-process-integration.scheduler.ts` and `msfaa-part-time-process-response-integration.scheduler.ts` other than the file renaming.
- Replaced some deprecated `substr` by `substring`.
- Added a condition `cancelledDate is null` to pending MSFAA generation to avoid generating an MSFAA to canceled records.

### DIFF
--- a/sources/packages/backend/libs/integrations/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-header.ts
+++ b/sources/packages/backend/libs/integrations/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-header.ts
@@ -37,7 +37,7 @@ export class MSFAAFileHeader implements MSFAARequestFileLine {
     const header = new MSFAAFileHeader();
     header.transactionCode = line.substring(0, 3) as RecordTypeCodes;
     header.processDate = getDateOnlyFromFormat(
-      line.substring(47, 54),
+      line.substring(47, 55),
       DATE_FORMAT,
     );
     return header;
--- a/sources/packages/backend/libs/integrations/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-header.ts
+++ b/sources/packages/backend/libs/integrations/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-header.ts
@@ -2,11 +2,12 @@ import {
   DATE_FORMAT,
   MSFAARequestFileLine,
   MSFAA_SENT_TITLE,
+  NUMBER_FILLER,
   RecordTypeCodes,
   SPACE_FILLER,
+  TIME_FORMAT,
 } from "../models/msfaa-integration.model";
 import { getDateOnlyFromFormat, StringBuilder } from "@sims/utilities";
-import { TIME_FORMAT } from "../../models/esdc-integration.model";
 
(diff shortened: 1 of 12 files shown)


sources/packages/backend/libs/integrations/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-header.ts
```diff
@@ -2,11 +2,12 @@ import {
   DATE_FORMAT,
   MSFAARequestFileLine,
   MSFAA_SENT_TITLE,
+  NUMBER_FILLER,
   RecordTypeCodes,
   SPACE_FILLER,
+  TIME_FORMAT,
```
guru-aot, line 8: Nice one ❤️ .

sources/packages/backend/libs/integrations/src/services/msfaa-number/msfaa-number.service.ts
```diff
@@ -54,6 +54,7 @@ export class MSFAANumberService extends RecordDataModelService<MSFAANumber> {
       .innerJoin("currentAssessment.offering", "offering")
       .innerJoin("offering.institutionLocation", "institutionLocation")
       .where("msfaaNumber.dateRequested is null")
+      .andWhere("msfaaNumber.cancelledDate is null")
```
guru-aot, line 57: Curious, the generation of pending msfaa for a cancelled record, should have a dateRequested already right? or are we considering the old data that was imported for MSFAA?

guru-aot (approved):
LGTM, thanks for doing the refactoring and replacements.

andrepestana-aot (approved):
LGTM.

sources/packages/backend/libs/integrations/src/services/msfaa-number/msfaa-number.service.ts
andrewsignori-aot: Mine though was to add this condition in preparation for the upcoming requirement for MSFAA reactivation which would potentially have SIMS setting the canceled date.
From #1777 
_Update ANY OTHER active (with null cancellation) MSFAA numbers for the student for the SAME OFFERING INTENSITY by populating the cancelled date and a boolean column indicating it is a system cancellation._

Either way, sending a canceled MSFAA would be a mistake.

ann-aot (approved):
👍

sources/packages/backend/libs/integrations/src/services/msfaa-number/msfaa-number.service.ts
dheepak-aot: @guru-aot with the upcoming requirement of re-activating MSFAA number, A cancelled MSFAA number can be re-activated where a pending msfaa(date_signed is null) could be cancelled by SIMS. At this situation, for that msfaa record which is cancelled will never have date signed.

dheepak-aot (approved):
Thanks for taking up the substring refactors and also getting the schedulers right. LGTM. 👍

Issue #582, opened 2021-07-26
Improvement in guards


ann-aot:
**Context**
1. Improvements in frontend routers, currently student routes doesn't have any restrictions  and update institutions routes accordingly
2. When adding anew route, everyone will have access to route in front end. If you don't specify roles you wont have access. If you don't explicitly say what the user has access to you don't have access to anything
**Tasks**

**Acceptance Criteria**
- [ ] Change default pattern so if role is not present the user is forbidden

ann-aot:
1. remove 
![image.png](https://images.zenhubusercontent.com/608872f7d0d25cdfe8dab75d/8068e109-c9af-4de4-b96d-265f7230891e) 
from sources/packages/api/src/auth/guards/institution-location.guard.ts 
2. loads admin locations

PR #1891, opened 2023-04-18
#582 - Improvement in guards

dheepak-aot:
# Improvement in Guards

The following improvements have been implemented to the Guards.

- [x] Defining the interface to route meta to have it typed.
image

- [x] Refactoring the logic of validating institution user access to the institution routes. According the updated logic, the institution user types must be specified in the route and missing to do so will reject the access for all user types.

- [x] Route meta property `allowOnlyBCPublic` created for routes to be allowed only for BCPublic institutions. (Implementation of this logic is not part of the ticket.)
- [x] Route meta property `allowOnlyLegalSigningAuthority` is added and is used to authorize only user with role legal signing authority to access the routes with `allowOnlyLegalSigningAuthority` set to true.
- [x] `useInstitutionAuth` refactored to use the store getters and added additional properties.
- [x] `isAuthenticatedInstitutionUser` created in `useInstitutionAuth` to include the verification of isActive status of the user. And due to which the following nav buttons does not appear during create institutions.
image

- [x] As per the Vue documentation, Global before guards are replaced with Global Resolve guards.

image

- [x] Refactored the interface `UserStateForStore` as the interface properties were not in sync with data stored in state.


sources/packages/web/src/router/InstitutionRouteHelper.ts
```diff
@@ -0,0 +1,111 @@
...
+  // If the user is institution admin, then they have access to all routes
+  // except the routes which are accessible only for legal signing authority.
+  if (isAdmin.value) {
+    // Validate the legal signing authority user role.
+    if (to.meta?.allowOnlyLegalSigningAuthority) {
+      return isLegalSigningAuthority.value;
+    }
+    return true;
+  }
+  // If the user is not an admin, check if the route is allowed for non admin user.
```
andrepestana-aot, line 93: I know that we only have 2 types of institution user for now but if we add a new one, we'll need to come here and change the comment. Maybe a more generic comment here would suit it better.

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -93,7 +93,10 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
         },
         meta: {
           clientType: ClientIdType.Institution,
-          userTypes: [InstitutionUserTypes.admin, InstitutionUserTypes.user],
+          institutionUserTypes: [
```
andrepestana-aot, line 96: We could have a constant for the institutionUserTypes as it repeats many times.

andrepestana-aot (commented):
Good job. I left some minor comments.

sources/packages/web/src/router/index.ts
```diff
@@ -3,14 +3,23 @@ import { studentRoutes } from "./StudentRoutes";
...
-  SharedRouteConst,
-} from "../constants/routes/RouteConstants";
-import { UserAuthorizationService } from "@/services/UserAuthorizationService";
-import { AuthService } from "@/services/AuthService";
 import { ClientIdType } from "../types/contracts/ConfigContract";
 import { supportingUsersRoutes } from "./SupportingUserRoutes";
+import { validateInstitutionUserAccess } from "./InstitutionRouteHelper";
+import { InstitutionUserTypes } from "@/types";
+
+declare module "vue-router" {
```
ann-aot, line 11: for my info, why declare a module here

sources/packages/web/src/router/index.ts
```diff
@@ -25,55 +34,21 @@ const router = createRouter({
...
-    })
-    .catch((error: unknown) => {
-      console.error(error);
-      throw error;
-    });
-}
+// Define error handling on router error.
+router.onError((error: unknown) => {
+  console.error(error);
+  throw error;
```
ann-aot, line 40: In our discussion when are saying we don't want the router to stuck in case of an error right, instead return something

sources/packages/web/src/router/InstitutionRouteHelper.ts
```diff
@@ -0,0 +1,111 @@
...
+import { RouteLocationNormalized, NavigationGuardNext } from "vue-router";
+/**
+ * Institution route params.
+ */
+interface InstitutionRouteParams {
+  locationId?: string;
+}
+
+/**
+ * Validate the access of user to the given route.
```
ann-aot, line 18: 👍

sources/packages/web/src/router/index.ts
dheepak-aot: Just blindly followed the way how it was given in vue documentation.

image
https://router.vuejs.org/guide/advanced/meta.html

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -1,34 +1,60 @@
...
+import { Store, useStore } from "vuex";
 import { useAuth } from "..";
 
-export function useInstitutionAuth() {
-  const store = useStore();
+export function useInstitutionAuth(rootStore?: Store<any>) {
+  const store = rootStore ?? useStore();
+
+  // Use store getters to get user details and authorizations.
+  const institutionUserDetails = store.getters[
```
andrewsignori-aot, line 14: Looking at the getters as they are I do not see a benefit since they are not actually transforming or doing any operation on the state.
IMO we can even get rid of the getters and use the state in this composable as it was before, for instance,
using `store.state.institution.userState`, `store.state.institution.authorizationsState`.

sources/packages/web/src/store/modules/institution/actions.ts
```diff
@@ -57,4 +58,16 @@ export const actions: ActionTree<InstitutionLocationState, RootState> = {
...
+
+  /**
+   * Set the institution user details
+   * during the institution setup.
+   * @param context action context.
+   */
+  setInstitutionUserDetailsOnSetup(
+    context: ActionContext<InstitutionLocationState, RootState>,
+  ): void {
+    const userState = { isInstitutionSetupUser: true } as UserStateForStore;
```
ann-aot, line 70: I was thinking shouldn't `isInstitutionSetupUser` is set to not true, when the intended logic is achieved, since it is stored in the store,  they will be able to access the `INSTITUTION_CREATE` URL  again after they have the institution right?

sources/packages/web/src/router/index.ts
```diff
@@ -3,14 +3,23 @@ import { studentRoutes } from "./StudentRoutes";
...
 import { ClientIdType } from "../types/contracts/ConfigContract";
 import { supportingUsersRoutes } from "./SupportingUserRoutes";
+import { validateInstitutionUserAccess } from "./InstitutionRouteHelper";
+import { InstitutionUserTypes } from "@/types";
+
+declare module "vue-router" {
+  /**
+   * Route meta properties.
+   */
+  interface RouteMeta {
```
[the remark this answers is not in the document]
dheepak-aot: I also would like to have separate interfaces for different client types. But unfortunately RouteMeta is the interface that the router's meta is typed to and we are suppose to use only one single interface.

Are you are saying that for the sake of readability if we should create like this? if so do not see a benefit of creating it as it(RouteMetaInstitution) could never be used anywhere.

```ts

interface RouteMetaInstitution {
    institutionUserTypes?: InstitutionUserTypes[];
    allowOnlyLegalSigningAuthority?: boolean;
    allowOnlyBCPublic?: boolean;
  }

interface RouteMeta extends  RouteMetaInstitution {
    clientType: ClientIdType;
    requiresAuth?: boolean;
    institutionUserTypes?: InstitutionUserTypes[];
    allowOnlyLegalSigningAuthority?: boolean;
    allowOnlyBCPublic?: boolean;
  }
```

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -1,34 +1,60 @@
...
 } from "@/types";
 import { computed } from "vue";
-import { useStore } from "vuex";
+import { Store, useStore } from "vuex";
 import { useAuth } from "..";
 
-export function useInstitutionAuth() {
-  const store = useStore();
+export function useInstitutionAuth(rootStore?: Store<any>) {
+  const store = rootStore ?? useStore();
```
ann-aot, line 11: why `rootStore`

sources/packages/web/src/router/index.ts
dheepak-aot: As discussed, I wrongly assumed that the catch block in the previous code had the re-direction, but it was not. 

So just kept the code inside catch as is and added to on error. @andrewsignori-aot @andrepestana-aot @guru-aot @sh16011993  fyi.

image

sources/packages/web/src/router/index.ts
```diff
@@ -25,55 +34,21 @@ const router = createRouter({
...
+  throw error;
+});
 
-router.beforeEach((to, from, next) => {
-  if (to.meta?.clientType === ClientIdType.Institution)
-    forEachInstitutionRoutes(to, from, next, ClientIdType.Institution);
-  else next();
+// TODO: As per vue documentation, usage of next is discouraged.
+// when vue-router version is upgraded, usage of next must be replaced
+// with returning either true or route location.
```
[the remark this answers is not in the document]
dheepak-aot: As discussed it is about the usage of returning true instead of next.

image

sources/packages/web/src/composables/institution/useInstitutionState.ts
```diff
@@ -1,9 +1,9 @@
 import { InstitutionStateForStore, LocationStateForStore } from "@/types";
 import { computed } from "vue";
-import { useStore } from "vuex";
+import { Store, useStore } from "vuex";
 
-export function useInstitutionState() {
-  const store = useStore();
+export function useInstitutionState(rootStore?: Store<any>) {
```
andrewsignori-aot, line 5: Can we replace `any` with `unknown`?

sources/packages/web/src/composables/institution/useInstitutionState.ts
```diff
@@ -14,8 +14,18 @@ export function useInstitutionState() {
     ).name;
   };
 
+  const initialize = async () => {
+    await store.dispatch("institution/initialize");
+  };
+
+  const setInstitutionUserDetailsOnSetup = async () => {
```
andrewsignori-aot, line 21: I would suggest having it named something like `setInstitutionSetupUser`. It should also just impact/change the `isInstitutionSetupUser` variable, in the way it is right now it is replacing the entire store state.

sources/packages/web/src/store/modules/institution/actions.ts
```diff
@@ -57,4 +58,16 @@ export const actions: ActionTree<InstitutionLocationState, RootState> = {
       await InstitutionService.shared.getMyInstitutionLocationsDetails();
     context.commit("setMyInstitutionLocationsDetailsState", resultComment);
   },
+
+  /**
+   * Set the institution user details
+   * during the institution setup.
+   * @param context action context.
+   */
+  setInstitutionUserDetailsOnSetup(
```
ann-aot, line 67: mm, shouldn't set be happen on `mutation.ts`?

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -1,34 +1,60 @@
...
+
+  // Use store getters to get user details and authorizations.
+  const institutionUserDetails = store.getters[
+    "institution/myDetails"
+  ] as UserStateForStore;
+  const authorizations =
```
ann-aot, line 17: just for my info why `store.state.institution.authorizationsState` are replaced

sources/packages/web/src/views/institution/AppInstitution.vue
```diff
@@ -8,7 +8,7 @@
         class="navigation-btn float-left"
       >
         <v-btn
-          v-if="isAuthenticated"
+          v-if="isAuthenticatedInstitutionUser"
```
andrewsignori-aot, line 11: Thanks for the `isAuthenticatedInstitutionUser` 👍

andrewsignori-aot (commented):
Nice work and great refactor, please take a look at the comments.

sources/packages/web/src/router/index.ts
andrewsignori-aot: So, do we need it at all? I am assuming that an error will be logged either way, right?

ann-aot (commented):
Nice work @dheepak-aot  added some comments

sources/packages/web/src/router/index.ts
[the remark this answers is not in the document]
dheepak-aot: Only interface AFAIK.

image

sources/packages/web/src/router/index.ts
dheepak-aot: Then I can remove the this for now and added it when required if that is ok.

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
dheepak-aot: As discussed when the composable NOT used from vue files, useStore() is undefined and we need to pass the store to the composable. I just followed the pattern of useStudentStore() and didn't investigate why it was done in that way.

sources/packages/web/src/composables/institution/useInstitutionState.ts
dheepak-aot: I did that and saw the following error. Felt that creating an interface and typing that with it could be done later.

image

sources/packages/web/src/store/modules/institution/actions.ts
dheepak-aot: Thanks for bringing this point @ann-aot . As discussed we assume that once the institution setup is complete on click of submit on the same page, the store variables are now created for institution which will eventually overwrite the institution user object in the store.

Again as discussed, tested the same and found the assumption is right.

image

image

sources/packages/web/src/router/index.ts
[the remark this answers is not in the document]
ann-aot: I understood after our discussion

sources/packages/web/src/store/modules/institution/actions.ts
```diff
@@ -57,4 +57,14 @@ export const actions: ActionTree<InstitutionLocationState, RootState> = {
...
   },
+
+  /**
+   * Set the institution setup user property to institution user state.
+   * @param context action context.
+   */
+  setInstitutionSetupUser(
+    context: ActionContext<InstitutionLocationState, RootState>,
+  ): void {
+    context.commit("setInstitutionSetupUser", true);
```
ann-aot, line 68: 👍

andrewsignori-aot (approved):
Thanks for doing the changes and the great refactor, looks good 👍

andrepestana-aot (approved):
LGTM

guru-aot (approved):
LGTM, nice work @dheepak-aot

Issue #1851, opened 2023-03-29
Enable Document Upload Tab for Public Institution User


JasonCTang:
**As a** *Public institution user* **I want** *view the file upload tab of the student view* **So That** **

**Additional Context**
- Public Institution users only need to know that students have uploaded files and should not have access to the contents of the documents.  So they cannot download or view them.
- enter text here

**Acceptance Criteria**
- [ ] Enable authorization paths to the File Upload tab for the public institution user.
- [ ] Disable links to download/view the file.

PR #1893, opened 2023-04-19
#1851 - Enable File Upload for Public Institution User

sh16011993:
The following tasks were completed as a part of this task:

- Remove upload file button for Public Institution User.
- Disable links to download/view the file for Public Institution User.
- Wrote E2E test - This included writing test for inserting fake file upload data into the database and retrieving it using the API endpoint and validating the retrieved data with the inserted data.

Screenshot showing the removed upload file button and disabled links to download/view the file:

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
@@ -1,0 +1,84 @@
+import { HttpStatus, INestApplication } from "@nestjs/common";
+import * as request from "supertest";
+import * as assert from "assert";
+import { DataSource, Repository } from "typeorm";
+import {
+  authorizeUserTokenForLocation,
+  BEARER_AUTH_TYPE,
+  createTestingAppModule,
+  getAuthRelatedEntities,
+  getInstitutionToken,
+  InstitutionTokenTypes,
+} from "../../../../testHelpers";
+import {
+  createFakeInstitutionLocation,
+  createFakeApplication,
+  saveFakeStudent,
+  saveFakeStudentFileUpload,
+} from "@sims/test-utils";
+import { Application, Institution, InstitutionLocation } from "@sims/sims-db";
+
+describe("StudentInstitutionsController(e2e)-getStudentFileUploads", () => {
+  let app: INestApplication;
+  let appDataSource: DataSource;
+  let collegeC: Institution;
+  let collegeCLocation: InstitutionLocation;
+  let applicationRepo: Repository<Application>;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    appDataSource = dataSource;
+
+    const { institution } = await getAuthRelatedEntities(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
+    );
+    collegeC = institution;
+    collegeCLocation = createFakeInstitutionLocation(collegeC);
+    await authorizeUserTokenForLocation(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
+      collegeCLocation,
+    );
+    applicationRepo = appDataSource.getRepository(Application);
+  });
+
+  it("Should get the student file uploads when student has at least one application submitted for the institution.", async () => {
+    // Student who has application submitted to institution.
+    const student = await saveFakeStudent(appDataSource);
+    const application = createFakeApplication({
+      location: collegeCLocation,
+      student,
+    });
+    await applicationRepo.save(application);
+
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeCUser,
+    );
+
+    // Save fake file upload for the student.
+    const studentUploadedFile = await saveFakeStudentFileUpload(appDataSource, {
+      student,
+    });
+
+    // Endpoint to test.
+    const endpoint = `/institutions/student/${student.id}/documents`;
+
+    // Assert.
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .then((res) => {
+        assert(res.body[0].fileName, studentUploadedFile.fileName);
+        assert(res.body[0].uniqueFileName, studentUploadedFile.uniqueFileName);
+        assert(res.body[0].groupName, "Ministry communications");
+        assert(res.body[0].fileOrigin, "Ministry");
+      });
+  });
+
+  afterAll(async () => {
+    await app?.close();
+  });
+});
--- a/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
@@ -167,6 +167,12 @@ export class AESTStudentFileAPIOutDTO extends StudentUploadFileAPIOutDTO {
   updatedAt: Date;
 }
 
+export class InstitutionStudentFileAPIOutDTO extends StudentUploadFileAPIOutDTO {
+  metadata: StudentFileMetadataAPIOutDTO;
+  groupName: string;
+  updatedAt: Date;
+}
+
 export class StudentFileMetadataAPIOutDTO {
   applicationNumber?: string;
 }
--- a/sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
@@ -7,7 +7,7 @@ import {
   Post,
 } from "@nestjs/common";
 import { ApiNotFoundResponse, ApiTags } from "@nestjs/swagger";
-import { StudentService } from "../../services";
+import { StudentService, StudentFileService } from "../../services";
 import { ClientTypeBaseRoute } from "../../types";
 import { AuthorizedParties } from "../../auth/authorized-parties.enum";
 import { AllowAuthorizedParty, UserToken } from "../../auth/decorators";
@@ -16,6 +16,7 @@ import {
   StudentSearchAPIInDTO,
   SearchStudentAPIOutDTO,
   StudentProfileAPIOutDTO,
+  InstitutionStudentFileAPIOutDTO,
 } from "./models/student.dto";
 import { IInstitutionUserToken } from "../../auth";
 import { StudentControllerService } from "./student.controller.service";
@@ -29,6 +30,7 @@ import { StudentControllerService } from "./student.controller.service";
 export class StudentInstitutionsController extends BaseController {
   constructor(
     private readonly studentService: StudentService,
+    private readonly fileService: StudentFileService,
     private readonly studentControllerService: StudentControllerService,
   ) {
     super();
@@ -67,5 +69,27 @@ export class StudentInstitutionsController extends BaseController {
     @Param("studentId", ParseIntPipe) studentId: number,
   ): Promise<StudentProfileAPIOutDTO> {
     return this.studentControllerService.getStudentProfile(studentId);
+  }
+
+  /**
+   * This controller returns all student documents uploaded by student uploader.
+   * @param studentId student id.
+   * @returns list of student documents.
+   */
+  @Get(":studentId/documents")
+  async getInstitutionStudentFiles(
+    @Param("studentId", ParseIntPipe) studentId: number,
+  ): Promise<InstitutionStudentFileAPIOutDTO[]> {
+    const studentDocuments = await this.fileService.getStudentUploadedFiles(
+      studentId,
+    );
+    return studentDocuments.map((studentDocument) => ({
+      fileName: studentDocument.fileName,
+      uniqueFileName: studentDocument.uniqueFileName,
+      metadata: studentDocument.metadata,
+      groupName: studentDocument.groupName,
+      updatedAt: studentDocument.updatedAt,
+      fileOrigin: studentDocument.fileOrigin,
+    }));
   }
 }
--- a/sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
+++ b/sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
@@ -1,0 +1,41 @@
+import * as faker from "faker";
+import { FileOriginType, Student, StudentFile, User } from "@sims/sims-db";
+import { DataSource } from "typeorm";
+import { createFakeStudent } from "./student";
+
+/**
+ * Create fake student file upload object.
+ * @param student entity.
+ * @returns created studentFile object.
+ */
+export function createFakeStudentFileUpload(student?: Student): StudentFile {
+  const studentFile = new StudentFile();
+  studentFile.fileName = faker.system.fileName();
+  studentFile.uniqueFileName =
+    studentFile.fileName +
+    faker.random.number(5) +
+    "." +
+    faker.system.fileType();
+  studentFile.groupName = "Ministry communications";
+  studentFile.mimeType = faker.system.mimeType();
+  studentFile.fileContent = Buffer.from(faker.random.words(50), "utf-8");
+  studentFile.student = student ?? createFakeStudent();
+  studentFile.creator = { id: student?.user.id } as User;
+  studentFile.fileOrigin = FileOriginType.Ministry;
+  return studentFile;
+}
+
+/**
+ * Save fake student file upload.
+ * @param dataSource data source to persist studentFileUpload.
+ * @param relations student entity relations.
+ * @returns persisted studentFile.
+ */
+export async function saveFakeStudentFileUpload(
+  dataSource: DataSource,
+  relations?: { student: Student },
+): Promise<StudentFile> {
+  const studentFile = createFakeStudentFileUpload(relations?.student);
+  const studentFileRepo = dataSource.getRepository(StudentFile);
+  return studentFileRepo.save(studentFile);
+}
--- a/sources/packages/web/src/components/common/students/StudentFileUploads.vue
+++ b/sources/packages/web/src/components/common/students/StudentFileUploads.vue
@@ -1,0 +1,206 @@
+<template>
+  <tab-container>
+    <body-header-container>
+      <template #header>
+        <body-header
+          title="File Uploads"
+          :recordsCount="studentFileUploads?.length"
+        >
+          <template #actions v-if="canUploadFiles">
+            <check-permission-role :role="Role.StudentUploadFile">
+              <template #="{ notAllowed }">
+                <v-btn
+                  color="primary"
+                  data-cy="uploadFileButton"
+                  @click="uploadFile"
+                  prepend-icon="fa:fa fa-plus-circle"
+                  class="float-right"
+                  :disabled="notAllowed"
+                  >Upload file</v-btn
+                >
+              </template>
+            </check-permission-role>
+          </template>
+        </body-header>
+      </template>
+      <content-group>
+        <toggle-content :toggled="!studentFileUploads.length">
+          <DataTable
+            :value="studentFileUploads"
+            :paginator="true"
+            :rows="DEFAULT_PAGE_LIMIT"
+            :rowsPerPageOptions="PAGINATION_LIST"
+          >
+            <template #empty>
+              <p class="text-center font-weight-bold">No records found.</p>
+            </template>
+            <Column
+              field="groupName"
+              header="Document Purpose"
+              :sortable="true"
+            ></Column>
+            <Column field="metadata" header="Application #">
+              <template #body="slotProps">{{
+                slotProps.data.metadata?.applicationNumber
+                  ? slotProps.data.metadata.applicationNumber
+                  : "-"
+              }}</template></Column
+            >
+            <Column field="updatedAt" header="Date Submitted"
+              ><template #body="slotProps">{{
+                dateOnlyLongString(slotProps.data.updatedAt)
+              }}</template></Column
+            >
+            <Column field="updatedAt" header="File">
+              <template #body="slotProps">
+                <div
+                  v-if="canDownloadFiles"
+                  class="file-label"
+                  @click="fileUtils.downloadStudentDocument(slotProps.data)"
+                >
+                  <span class="mr-4">
+                    <v-icon icon="fa:far fa-file-alt" size="20"></v-icon
+                  ></span>
+                  <span>{{ slotProps.data.fileName }}</span>
+                </div>
+                <div v-else>
+                  <span class="mr-4">
+                    <v-icon icon="fa:far fa-file-alt" size="20"></v-icon
+                  ></span>
+                  <span>{{ slotProps.data.fileName }}</span>
+                </div>
+              </template></Column
+            >
+          </DataTable>
+        </toggle-content>
+      </content-group>
+      <formio-modal-dialog
+        :max-width="730"
+        ref="fileUploadModal"
+        title="Upload file"
+        :formData="initialData"
+        formName="uploadStudentDocumentsAEST"
+      >
+        <template #actions="{ cancel, submit }">
+          <v-row class="m-0 p-0">
+            <v-btn color="primary" variant="outlined" @click="cancel"
+              >Cancel</v-btn
+            >
+            <check-permission-role :role="Role.StudentUploadFile">
+              <template #="{ notAllowed }">
+                <v-btn
+                  class="float-right"
+                  @click="submit"
+                  color="primary"
+                  variant="elevated"
+                  :disabled="notAllowed"
+                  >Upload now</v-btn
+                >
+              </template>
+            </check-permission-role>
+          </v-row>
+        </template>
+      </formio-modal-dialog>
+    </body-header-container>
+  </tab-container>
+</template>
+
+<script lang="ts">
+import { onMounted, ref, defineComponent } from "vue";
+import {
+  DEFAULT_PAGE_LIMIT,
+  FormIOForm,
+  PAGINATION_LIST,
+  LayoutTemplates,
+  Role,
+} from "@/types";
+import { StudentService } from "@/services/StudentService";
+import {
+  useFormatters,
+  useFileUtils,
+  ModalDialog,
+  useFormioUtils,
+  useSnackBar,
+} from "@/composables";
+import FormioModalDialog from "@/components/generic/FormioModalDialog.vue";
+import { AESTFileUploadToStudentAPIInDTO } from "@/services/http/dto/Student.dto";
+import CheckPermissionRole from "@/components/generic/CheckPermissionRole.vue";
+
+export default defineComponent({
+  components: {
+    FormioModalDialog,
+    CheckPermissionRole,
+  },
+  emits: ["loaded"],
+  props: {
+    studentId: {
+      type: Number,
+      required: true,
+    },
+    canUploadFiles: {
+      type: Boolean,
+      required: true,
+    },
+    canDownloadFiles: {
+      type: Boolean,
+      required: true,
+    },
+    studentFileUploads: {
+      type: Array,
+      required: true,
+    },
+  },
+  setup(props: any, context) {
+    const fileUploadModal = ref({} as ModalDialog<FormIOForm | boolean>);
+    const { dateOnlyLongString } = useFormatters();
+    const snackBar = useSnackBar();
+    const fileUtils = useFileUtils();
+    const formioUtils = useFormioUtils();
+    const initialData = ref({ studentId: props.studentId });
+
+    const loadStudentFileUploads = async () => {
+      context.emit("loaded", props.studentId);
+    };
+
+    onMounted(async () => {
+      await loadStudentFileUploads();
+    });
+
+    const uploadFile = async () => {
+      const modalResult = await fileUploadModal.value.showModal();
+      if (!modalResult) {
+        return;
+      }
+
+      try {
+        const associatedFiles = formioUtils.getAssociatedFiles(modalResult);
+        const payload: AESTFileUploadToStudentAPIInDTO = {
+          associatedFiles,
+        };
+        await StudentService.shared.saveAESTUploadedFilesToStudent(
+          props.studentId,
+          payload,
+        );
+        snackBar.success(
+          "The documents were submitted and a notification was sent to the student.",
+        );
+        await loadStudentFileUploads();
+      } catch {
+        snackBar.error("An unexpected error happened.");
+      }
+    };
+
+    return {
+      fileUtils,
+      DEFAULT_PAGE_LIMIT,
+      PAGINATION_LIST,
+      dateOnlyLongString,
+      uploadFile,
+      fileUploadModal,
+      initialData,
+      LayoutTemplates,
+      Role,
+    };
+  },
+});
+</script>
--- a/sources/packages/web/src/services/StudentService.ts
+++ b/sources/packages/web/src/services/StudentService.ts
@@ -4,6 +4,7 @@ import { useFormatters } from "@/composables";
 import {
   AESTFileUploadToStudentAPIInDTO,
   AESTStudentFileAPIOutDTO,
+  InstitutionStudentFileAPIOutDTO,
   CreateStudentAPIInDTO,
   SearchStudentAPIOutDTO,
   StudentFileUploaderAPIInDTO,
@@ -130,6 +131,16 @@ export class StudentService {
     studentId: number,
   ): Promise<AESTStudentFileAPIOutDTO[]> {
     return ApiClient.Students.getAESTStudentFiles(studentId);
+  }
+
+  /**
+   * Get all student documents for Institution user.
+   * @return list of student documents.
+   */
+  async getInstitutionStudentFiles(
+    studentId: number,
+  ): Promise<InstitutionStudentFileAPIOutDTO[]> {
+    return ApiClient.Students.getInstitutionStudentFiles(studentId);
   }
 
   /**
--- a/sources/packages/web/src/services/http/StudentApi.ts
+++ b/sources/packages/web/src/services/http/StudentApi.ts
@@ -2,6 +2,7 @@ import HttpBaseClient from "@/services/http/common/HttpBaseClient";
 import {
   AESTFileUploadToStudentAPIInDTO,
   AESTStudentFileAPIOutDTO,
+  InstitutionStudentFileAPIOutDTO,
   CreateStudentAPIInDTO,
   SearchStudentAPIOutDTO,
   StudentFileUploaderAPIInDTO,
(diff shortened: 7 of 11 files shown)


sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
```diff
@@ -0,0 +1,41 @@
...
+  studentFile.student = student ?? createFakeStudent();
+  studentFile.creator = { id: student?.user.id } as User;
+  studentFile.fileOrigin = FileOriginType.Ministry;
+  return studentFile;
+}
+
+/**
+ * Save fake student file upload.
+ * @param dataSource data source to persist studentFileUpload.
+ * @param relations student entity relations.
```
ann-aot, line 31: maybe you can expand the options in the relation

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    // Endpoint to test.
+    const endpoint = `/institutions/student/${student.id}/documents`;
+
+    // Assert.
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .then((res) => {
```
ann-aot, line 73: why you went for this approach instead of something like this
image

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -68,4 +70,26 @@ export class StudentInstitutionsController extends BaseController {
...
+  /**
+   * This controller returns all student documents uploaded by student uploader.
+   * @param studentId student id.
+   * @returns list of student documents.
+   */
+  @Get(":studentId/documents")
+  async getInstitutionStudentFiles(
+    @Param("studentId", ParseIntPipe) studentId: number,
+  ): Promise<InstitutionStudentFileAPIOutDTO[]> {
+    const studentDocuments = await this.fileService.getStudentUploadedFiles(
```
ann-aot, line 83: since it is exactly the same as the AEST controller, we can move the common code to the shared controller service and call the shared service 
image. Maybe you can refer to the profile controller https://vscode.dev/github/bcgov/SIMS/blob/a2442286f936c0ee7621fa39a105aaa36607fe36/sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts#L68

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+          </template>
+        </body-header>
+      </template>
+      <content-group>
+        <toggle-content :toggled="!studentFileUploads.length">
+          <DataTable
+            :value="studentFileUploads"
+            :paginator="true"
+            :rows="DEFAULT_PAGE_LIMIT"
+            :rowsPerPageOptions="PAGINATION_LIST"
```
ann-aot, line 32: `totalRecords`  param is also required, right?

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -167,6 +167,12 @@ export class AESTStudentFileAPIOutDTO extends StudentUploadFileAPIOutDTO {
   updatedAt: Date;
 }
 
+export class InstitutionStudentFileAPIOutDTO extends StudentUploadFileAPIOutDTO {
```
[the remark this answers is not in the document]
dheepak-aot: @ann-aot  IMO, even though the properties are same between AEST and Institution, I would prefer to have separate DTOs considering they both have additional properties from `StudentUploadFileAPIOutDTO` and we trying to put a contextual name for the sake that both AEST and Institution have same properties. But not a very hard feeling.

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
dheepak-aot: Yes we can move this code to controller service and could be used for student as well.

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -68,4 +70,26 @@ export class StudentInstitutionsController extends BaseController {
   ): Promise<StudentProfileAPIOutDTO> {
     return this.studentControllerService.getStudentProfile(studentId);
   }
+
+  /**
+   * This controller returns all student documents uploaded by student uploader.
```
dheepak-aot, line 75: I know the comments may be copied over from other controller. I would suggest the comment to be something like `Get all the documents uploaded by student.`

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -68,4 +70,26 @@ export class StudentInstitutionsController extends BaseController {
...
+   * @param studentId student id.
+   * @returns list of student documents.
+   */
+  @Get(":studentId/documents")
+  async getInstitutionStudentFiles(
+    @Param("studentId", ParseIntPipe) studentId: number,
+  ): Promise<InstitutionStudentFileAPIOutDTO[]> {
```
dheepak-aot, line 82: I would like to have the studentId to be validated by the API before proceeding any further like this in the begining of the method. 

```ts
const student = await this.studentService.studentExists(studentId);
    if (!student) {
      throw new NotFoundException("Student not found.");
    }
```

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
+<template>
+  <tab-container>
```
dheepak-aot, line 2: 👍 thanks for moving it to components.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+  },
+  setup(props: any, context) {
+    const fileUploadModal = ref({} as ModalDialog<FormIOForm | boolean>);
+    const { dateOnlyLongString } = useFormatters();
+    const snackBar = useSnackBar();
+    const fileUtils = useFileUtils();
+    const formioUtils = useFormioUtils();
+    const initialData = ref({ studentId: props.studentId });
+
+    const loadStudentFileUploads = async () => {
```
dheepak-aot, line 161: I would say that the calling service(API) part can be inside the component itself. Because getStudentFiles can be one service which can handle the API call for all the client types. And as per our latest dev discussion we can have a common model returned from service for the Vue perspective.

```ts
const loadStudentFileUploads = async () => {
      studentFileUploads.value =
        await StudentService.shared.getStudentFiles(props.studentId);
    };
```

StudentService.ts

```ts
async getStudentFiles(
    studentId: number,
  ): Promise<StudentFile[]> {
   // Transformation if needed
    return ApiClient.Students.getStudentFiles(studentId);
  }
```

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
sh16011993: @ann-aot I have skipped the properties "updatedAt" and "metadata" in the comparison. For me to skip those, I had to follow this approach. Please let me know if I should add those properties.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
ann-aot: @dheepak-aot `getStudentFiles` method in StudentService.ts is the service for student-side file uploader logic. 

 This common component is for Institution and AEST datatable

ann-aot: But as @dheepak-aot  mentioned we can call the service from here 
image
you can refer to other shared components. and we already have the service probably you can rename and adjust it 
image

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+      context.emit("loaded", props.studentId);
+    };
+
+    onMounted(async () => {
```
ann-aot, line 165: suggestion, you can also do `onMounted(loadStudentFileUploads);`

sources/packages/web/src/components/common/students/StudentFileUploads.vue
sh16011993: We are using `studentFileUploads?.length` directly.

sources/packages/web/src/services/StudentService.ts
```diff
@@ -132,6 +133,16 @@ export class StudentService {
     return ApiClient.Students.getAESTStudentFiles(studentId);
   }
 
+  /**
+   * Get all student documents for Institution user.
+   * @return list of student documents.
+   */
+  async getInstitutionStudentFiles(
```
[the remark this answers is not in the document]
ann-aot: https://github.com/bcgov/SIMS/pull/1893/files#r1171726717

sources/packages/web/src/services/http/StudentApi.ts
```diff
@@ -135,6 +136,18 @@ export class StudentApi extends HttpBaseClient {
     );
   }
 
+  /**
+   * Get all student documents uploaded by student uploader.
+   * @return list of student documents.
+   */
+  async getInstitutionStudentFiles(
```
ann-aot, line 143: we already have a api service, just need to adjust the comments and rename
image

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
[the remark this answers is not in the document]
ann-aot: @dheepak-aot `StudentUploadFileAPIOutDTO` is a different DTO, it's for the student side, I belive as part of this story, the component is shared between AEST and the institution only. we already have created an DTO for AEST `AESTStudentFileAPIOutDTO`  which is exactly the same as this newly created `InstitutionStudentFileAPIOutDTO`, 
image

sources/packages/web/src/services/http/dto/Student.dto.ts
```diff
@@ -75,6 +75,16 @@ export interface AESTStudentFileAPIOutDTO extends StudentUploadFileAPIOutDTO {
   updatedAt: Date;
 }
 
+/**
+ *  Institution user to view student uploaded documents.
+ */
+export interface InstitutionStudentFileAPIOutDTO
```
ann-aot, line 81: https://github.com/bcgov/SIMS/pull/1893/files#r1171547861

sources/packages/web/src/components/common/students/StudentFileUploads.vue
sh16011993: I feel its better to explicitly specify that `loadStudentFileUploads` is an async call.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
andrewsignori-aot: Unless there is a problem adding those, please add them and follow @ann-aot recommendation.
Following it also ensures that anything not expected is returned.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    await authorizeUserTokenForLocation(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
+      collegeCLocation,
+    );
+    applicationRepo = appDataSource.getRepository(Application);
+  });
+
+  it("Should get the student file uploads when student has at least one application submitted for the institution.", async () => {
+    // Student who has application submitted to institution.
```
andrepestana-aot, line 48: We usually comment `// Arrange` at this part of the test.

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
andrewsignori-aot: @dheepak-aot is it not supposed to be already ensured when the authorization decorator is in place?

sources/packages/web/src/views/institution/student/InstitutionStudentFileUploads.vue
```diff
@@ -1,3 +1,36 @@
 <template>
-  <tab-container><div>File Upload component to be built.</div></tab-container>
+  <tab-container>
+    <student-file-uploads
+      :studentId="studentId"
+      :canUploadFiles="false"
```
dheepak-aot, line 5: these properties can be defaulted to false at the component level and not required to mention explicitly.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    collegeCLocation = createFakeInstitutionLocation(collegeC);
+    await authorizeUserTokenForLocation(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
+      collegeCLocation,
+    );
+    applicationRepo = appDataSource.getRepository(Application);
+  });
+
+  it("Should get the student file uploads when student has at least one application submitted for the institution.", async () => {
```
andrepestana-aot, line 47: You could create a record with fileOrgin = "Temporary" and check that the API doesn't display that one. Maybe another test? Just a suggestion.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
andrewsignori-aot: I would be inclined to agree with @ann-aot because the code looks cleaner.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  let applicationRepo: Repository<Application>;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    appDataSource = dataSource;
+
+    const { institution } = await getAuthRelatedEntities(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
```
andrewsignori-aot, line 35: The PR with the public institution is already merged, can we update these tests to use the new public institution?

andrepestana-aot (commented):
Good job! Left some comments.

sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Create fake student file upload object.
+ * @param student entity.
+ * @returns created studentFile object.
+ */
+export function createFakeStudentFileUpload(student?: Student): StudentFile {
+  const studentFile = new StudentFile();
+  studentFile.fileName = faker.system.fileName();
+  studentFile.uniqueFileName =
+    studentFile.fileName +
+    faker.random.number(5) +
```
andrewsignori-aot, line 16: Even the DB not having a constraint to the `unique_file_name` the file names are generated including a guid to ensure their uniqueness, for instance, `Upload Test - Copy-2c0028d5-c681-46d4-84b3-471d890a6f13.doc`.
Can we please ensure that this factory does the same?

sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
```diff
@@ -0,0 +1,41 @@
...
+    "." +
+    faker.system.fileType();
+  studentFile.groupName = "Ministry communications";
+  studentFile.mimeType = faker.system.mimeType();
+  studentFile.fileContent = Buffer.from(faker.random.words(50), "utf-8");
+  studentFile.student = student ?? createFakeStudent();
+  studentFile.creator = { id: student?.user.id } as User;
```
andrewsignori-aot, line 23: The creator should be received as a parameter in the relations.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+              <p class="text-center font-weight-bold">No records found.</p>
+            </template>
+            <Column
+              field="groupName"
+              header="Document Purpose"
+              :sortable="true"
+            ></Column>
+            <Column field="metadata" header="Application #">
+              <template #body="slotProps">{{
+                slotProps.data.metadata?.applicationNumber
```
andrewsignori-aot, line 44: Can you please change it to use the `emptyStringFiller` from `useFormatters`.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+                  ></span>
+                  <span>{{ slotProps.data.fileName }}</span>
+                </div>
+              </template></Column
+            >
+          </DataTable>
+        </toggle-content>
+      </content-group>
+      <formio-modal-dialog
+        :max-width="730"
```
andrewsignori-aot, line 78: Please remove the `:max-width="730"` it shouldn't be needed anymore.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+    canDownloadFiles: {
+      type: Boolean,
+      required: true,
+    },
+    studentFileUploads: {
+      type: Array,
+      required: true,
+    },
+  },
+  setup(props: any, context) {
```
andrewsignori-aot, line 153: Please remove the `any`.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,206 @@
...
+      if (!modalResult) {
+        return;
+      }
+
+      try {
+        const associatedFiles = formioUtils.getAssociatedFiles(modalResult);
+        const payload: AESTFileUploadToStudentAPIInDTO = {
+          associatedFiles,
+        };
+        await StudentService.shared.saveAESTUploadedFilesToStudent(
```
andrewsignori-aot, line 180: @dheepak-aot @andrepestana-aot @dheepak-aot @guru-aot should we move this AEST specific logic outside the component?
Maybe an event that will be triggered with the `associatedFiles` array as an event args?

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
dheepak-aot: @sh16011993 ignore my previous comment. As @andrewsignori-aot mentioned, it will be handled by the decorator.

sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
sh16011993: Are you suggesting that I use a `guid` generator instead of `faker.random.number(5)` that is currently in the code?

sources/packages/web/src/components/common/students/StudentFileUploads.vue
ann-aot: mm... that's a good point...How we are managing in other shared components, I think even though this scenario is AEST specific, it is an action of the component which can be enabled for any client type. I am kind of okay with this.  but would like to know what other devs thought.

ann-aot: Thinking about it in a business context, it is only enabled for the ministry for communications. yep, I am good to move it to the AEST side. it makes sense

sources/packages/web/src/components/common/students/StudentFileUploads.vue
sh16011993: I have made the suggested change.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
andrepestana-aot: Yes, that's a good idea as it's only related to AEST.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
andrewsignori-aot: If the other suggestion about making the API call unique will be implemented, the "loaded" event will be removed, right?

sh16011993: Yes

sources/packages/web/src/components/common/students/StudentFileUploads.vue
guru-aot: I agree

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,201 @@
+<template>
```
guru-aot, line 1: ❤️

sources/packages/web/src/components/common/students/StudentFileUploads.vue
sh16011993: I have made the change

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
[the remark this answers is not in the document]
dheepak-aot: @ann-aot I understood the context from our conversation. I am good on this.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,115 @@
...
+  <tab-container>
```
dheepak-aot, line 2: So this one being a component, we must remove the tab container from here and I see already the view/*.vue files using this are having the tab container in it.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,115 @@
...
+                    <v-icon icon="fa:far fa-file-alt" size="20"></v-icon
+                  ></span>
+                  <span>{{ slotProps.data.fileName }}</span>
+                </div>
+              </template></Column
+            >
+          </DataTable>
+        </toggle-content>
+      </content-group>
+      <slot></slot>
```
dheepak-aot, line 61: What is the purpose of this default slot? I guess it may be carried over from the old file.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentFileUploads.e2e-spec.ts
```diff
@@ -0,0 +1,130 @@
+import { HttpStatus, INestApplication } from "@nestjs/common";
+import * as request from "supertest";
+import * as assert from "assert";
```
andrewsignori-aot, line 3: Please remove the `assert` as it is not used.

sources/packages/web/src/services/StudentService.ts
```diff
@@ -123,13 +123,13 @@ export class StudentService {
   }
 
   /**
-   * Get all student documents for AEST user.
+   * Get all student documents for AEST / Institution user.
```
dheepak-aot, line 126: As per the context of this method, it returns all student files in detail and not a client specific thing.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
dheepak-aot: Looking further I got this part, 
image


But a question here. When a file upload happens, how will the list of files refresh?

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -135,6 +136,27 @@ export class StudentControllerService {
     };
   }
 
+  /**
+   * Get the student uploaded files.
+   * @param studentId student id to retrieve the data.
+   * @returns student file details.
+   */
+  async getStudentUploadedFiles(
```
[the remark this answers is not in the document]
andrewsignori-aot: I am not sure if I would be on the same page about this.

sources/packages/web/src/views/aest/student/FileUploads.vue
```diff
@@ -1,74 +1,26 @@
...
-                  >Upload file</v-btn
-                >
-              </template>
-            </check-permission-role>
+    <student-file-uploads
+      ref="studentFileUploads"
+      :studentId="studentId"
+      :canDownloadFiles="true"
+    >
+      <template #uploadBtn>
```
dheepak-aot, line 8: https://github.com/bcgov/SIMS/pull/1893/files#r1173965852

sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
```diff
@@ -0,0 +1,48 @@
...
+): StudentFile {
+  const studentFile = new StudentFile();
+  studentFile.fileName = faker.system.fileName();
+  studentFile.uniqueFileName =
+    studentFile.fileName + faker.random.uuid() + "." + faker.system.fileType();
+  studentFile.groupName = "Ministry communications";
+  studentFile.mimeType = faker.system.mimeType();
+  studentFile.fileContent = Buffer.from(faker.random.words(50), "utf-8");
+  studentFile.student = relations?.student ?? createFakeStudent();
+  studentFile.creator = { id: relations?.creator?.id } as User;
```
andrewsignori-aot, line 28: It can be simplified to `studentFile.creator = creator;`, unless I am missing something.

sources/packages/web/src/views/aest/student/FileUploads.vue
```diff
@@ -168,23 +93,17 @@ export default defineComponent({
         snackBar.success(
           "The documents were submitted and a notification was sent to the student.",
         );
-        await loadStudentFileUploads();
+        studentFileUploads.value.loadStudentFileUploads();
```
andrewsignori-aot, line 96: The same approach is also used for the modals along the code, that is why I am not going to say, don`t do it but,
IMO, it can cause confusion because someone looking at the setup method may not expect it.

Is there a reason to believe that this would be clear that the providing the method as a parameter to be invoked, as shown below?

image

andrewsignori-aot (commented):
Thanks for doing the changes, please look at the comments.

sources/packages/web/src/views/aest/student/FileUploads.vue
sh16011993: Vue Docs: https://vuejs.org/guide/essentials/template-refs.html#ref-on-component

In the last section [Ref on Component], it discusses the approach I followed but then it also says that the approach "makes it easy to create tightly coupled implementation details between the parent and the child, so component refs should be only used when absolutely needed - in most cases, you should try to implement parent / child interactions using the standard props and emit interfaces first."

sources/packages/backend/libs/test-utils/src/factories/student-file-uploads.ts
```diff
@@ -0,0 +1,48 @@
+import * as faker from "faker";
+import { FileOriginType, Student, StudentFile, User } from "@sims/sims-db";
+import { DataSource } from "typeorm";
+import { createFakeStudent } from "./student";
+
+/**
+ * Create fake student file upload object.
+ * @param student entity.
```
ann-aot, line 8: pls review the commnets

sources/packages/web/src/views/institution/student/InstitutionStudentFileUploads.vue
```diff
@@ -1,3 +1,20 @@
 <template>
-  <tab-container><div>File Upload component to be built.</div></tab-container>
+  <tab-container>
+    <student-file-uploads :studentId="studentId"></student-file-uploads>
```
ann-aot, line 3: suggestion, u can also use `<student-file-uploads :studentId="studentId"/>`

sources/packages/web/src/components/common/students/StudentFileUploads.vue
ann-aot: I am sorry, I didn't get why we need this default slot, is it for the modal? if it's, do we really need it??

ann-aot: but I agree with @dheepak-aot, after file upload, the component needs to be refreshed

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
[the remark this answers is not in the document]
dheepak-aot: Sure. we can talk if needed.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
sh16011993: image

sh16011993: I had added the `slot` you are referring to to render any additional AEST or Institution template specific content.

sources/packages/web/src/views/aest/student/FileUploads.vue
dheepak-aot: IMO, executing an operation/method of child component in parent looks like unorthodox, as mentioned in the Vue documentation it is used for tightly coupled implementation but is that what we want? we could follow the pattern what we followed for other components where we block these ones with v-if inside the component.

dheepak-aot:
Thanks for making the changes. Added more comments.

sources/packages/web/src/views/aest/student/FileUploads.vue
sh16011993: Yes, I am reverting back that change.

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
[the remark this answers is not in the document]
andrewsignori-aot: After our conversation, I got that the 3 endpoints are actually the same resource `/documents` accessed by the three different clients (Students, Ministry, Institution) hence I am ok if your suggestion @dheepak-aot.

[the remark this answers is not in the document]
sh16011993: I did the change.

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -135,6 +137,32 @@ export class StudentControllerService {
...
+    studentId: number,
+    options?: { extendedDetails: boolean },
```
andrewsignori-aot, line 147: Please add some comments to the `options` parameters.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,166 @@
...
+      required: false,
+      default: false,
+    },
+    canDownloadFiles: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
+  },
+  emits: ["uploadFile"],
```
andrewsignori-aot, line 131: Minor, but I would say that we having being declaring the `emits` at the top.

sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -0,0 +1,166 @@
...
+    const fileUtils = useFileUtils();
+    const initialData = ref({ studentId: props.studentId });
+
+    const loadStudentFileUploads = async () => {
+      studentFileUploads.value =
+        await StudentService.shared.getStudentFileDetails(props.studentId);
+    };
+
+    const uploadFile = () => {
+      context.emit("uploadFile", fileUploadModal, () =>
```
andrewsignori-aot, line 145: I am not expecting to provide the modal as a parameter here.
Since the modal itself it is in this component, we can call the event if the modal results in a file uploaded. This would be the recommendation from the comment https://github.com/bcgov/SIMS/pull/1893#discussion_r1171793732 where it says "`associatedFiles` array as an event args".

sources/packages/web/src/components/common/students/StudentFileUploads.vue
andrewsignori-aot: I am not expecting to provide the modal as a parameter here.
Since the modal itself is in this component, we can call the event if the modal results in a file uploaded. This would be the recommendation from the first comment on this thread where it says "`associatedFiles` array as an event args".

sources/packages/web/src/views/aest/student/FileUploads.vue
```diff
@@ -1,156 +1,38 @@
...
-        await StudentService.shared.getAESTStudentFiles(props.studentId);
-    };
-
-    onMounted(async () => {
-      await loadStudentFileUploads();
-    });
 
-    const uploadFile = async () => {
+    const uploadFile = async (
+      fileUploadModal,
```
andrewsignori-aot, line 33: Please check if you have the below waning, we should get rid of them.
image

andrewsignori-aot (commented):
Please take a look at a few minor comments. I believe that the below one is the one that requires more attention https://github.com/bcgov/SIMS/pull/1893#discussion_r1175631845

sources/packages/web/src/components/common/students/StudentFileUploads.vue
dheepak-aot, line 145: I didn't follow why are we emitting back to parent here?

I technically what happen between here and parent, but are we going to follow this approach moving forward?

sh16011993: I have updated this part of the code as suggested by @andrewsignori-aot

Instead of passing the entire `fileUploadModal` as argument, only `associatedFiles` array is being passed to the emitted event handler. We are emitting `uploadFile` to the AEST specific file upload component to allow for the API call `saveAESTUploadedFilesToStudent` to be made by the consumer since it is a consumer specific operation.

ann-aot (approved):
Good work @sh16011993 👍 Thanks for doing the changes

ann-aot (commented):
Pls take a look at the failing test cases

sources/packages/web/src/components/common/students/StudentFileUploads.vue
andrewsignori-aot: @dheepak-aot considering also the last explanation given by @sh16011993 is there anything here to be considered as a new approach? 
If the point was about the modal as a parameter, that was exactly what I asked to be changed but besides that, is there anything else?

andrewsignori-aot (approved):

sources/packages/web/src/components/common/students/StudentFileUploads.vue
dheepak-aot: No I am good.

dheepak-aot (approved):
LGTM

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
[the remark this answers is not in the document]
dheepak-aot: So in that case, I assume we would not consider using a controller service method just to share the transformation code alone.

guru-aot (approved):
LGTM, nice work @sh16011993

andrepestana-aot (approved):
Great job!

PR #1898, opened 2023-04-20
#582 - Restrict non BC Public Institutions from accessing Routes for BC Public institutions

dheepak-aot:
Based on the route meta property `allowOnlyBCPublic` block access for non BC Public Institution users from accessing routes which are for BCPublic institutions.

### DIFF
--- a/sources/packages/web/src/router/InstitutionRoutes.ts
+++ b/sources/packages/web/src/router/InstitutionRoutes.ts
@@ -460,6 +460,7 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
             InstitutionUserTypes.admin,
             InstitutionUserTypes.user,
           ],
+          allowOnlyBCPublic: true,
         },
         children: [
           {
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -468,6 +469,7 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
             component: InstitutionStudentProfile,
             meta: {
               clientType: ClientIdType.Institution,
+              allowOnlyBCPublic: true,
```
ann-aot, line 472: adding `allowOnlyBCPublic` to the parent (line 457) will it protect all children? As discussed, you can just try and check

dheepak-aot: I tested by using allowOnlyBCPublic in parent onluy and it works perfectly. thanks.

guru-aot (approved):
LGTM, nice work @dheepak-aot

andrepestana-aot (approved):
LGTM

sources/packages/web/src/router/InstitutionRoutes.ts
ann-aot: Thanks for trying it. As discussed, we can test and remove the other meta from the children in another ticket

ann-aot (approved):
Good work @dheepak-aot 👍

andrewsignori-aot (approved):
Nice work, looks good 👍

Issue #1790, opened 2023-03-09
Enable Applications Tab for Public Institution User


JasonCTang:
**As an** *Public institutions user* **I want** *to view the Student's applications with my institution* **So That** *I can help the student understand their awards, assessments and disbursements.*

**Additional Context**
- Lynn to provide UI/UX

**To be answered**
- [ ] Only return applications that are submitted.

**Acceptance Criteria**
- [ ] Enable authorization paths to the Applications tab in the student view for Public institution users.
- [x] Only return applications that match the institution of the institution user.

PR #1903, opened 2023-04-24
#1790 - Enable Applications Tab for Public Institution User

ann-aot:
- e2e test for the new controller
image
- Enabled applications tab for the public institution user
- Authorization part will be taken care of when https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/1830 is done

Screenshots:
1. Student has submitted an application to the below institution
image
2. Below are the applications that belong to this student
image


sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    appDataSource = dataSource;
+    // College C.
+    const { institution } = await getAuthRelatedEntities(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
+    );
+    collegeC = institution;
+    collegeCLocation = createFakeInstitutionLocation(collegeC);
```
andrewsignori-aot, line 50: Can we use `College F`(public) instead in preparation for the test changes when the authorization decorator is in place?

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      EducationProgramOffering,
+    );
+    institutionLocationRepo = dataSource.getRepository(InstitutionLocation);
+  });
+
+  it("Should get the student application details as summary when student has a submitted application for the institution (application with location id saved).", async () => {
+    // Arrange
+
+    // Student has a submitted application to the institution.
+    const student = await saveFakeStudent(appDataSource);
```
andrewsignori-aot, line 68: Is it not possible to simplify the E2E data creation using the `saveFakeApplication`?

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  it("Should get an empty array as application summary when student has a draft application for the institution (application without the location id saved(i.e draft/cancelled draft status)).", async () => {
+    // Arrange
+
+    // Student has a draft application to the institution.
+    const student = await saveFakeStudent(appDataSource);
+    const application = createFakeApplication({
+      student,
+    });
+    application.data.selectedLocation = collegeCLocation.id;
```
andrewsignori-aot, line 126: `application.data.selectedLocation` has any effect on the test?

sources/packages/web/src/views/student/StudentApplicationSummary.vue
```diff
@@ -7,7 +7,8 @@
         </template>
       </header-navigator>
     </template>
-    <v-row>
+    <v-row
+      >student-applications
```
andrewsignori-aot, line 11: I believe that the string `student-applications` is lost in the code.

sources/packages/web/src/views/aest/student/Applications.vue
```diff
@@ -21,7 +21,7 @@ export default defineComponent({
       required: true,
     },
   },
-  setup(props: any) {
+  setup(props) {
```
andrewsignori-aot, line 24: Thanks 😉

sources/packages/web/src/components/common/students/StudentApplications.vue
```diff
@@ -41,7 +41,11 @@
               >
             </v-btn>
 
-            <span v-if="clientType === ClientIdType.AEST"
+            <span
+              v-if="
+                clientType === ClientIdType.AEST ||
```
andrewsignori-aot, line 46: Since it is an old code (and refactoring would not be part of the PR) I will not request to change it but doing the checks based on the client is what we have been trying to avoid.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
dheepak-aot: From the data seed institution type context (Just to make it easy here)

College F - BC Public 
College C - BC Private
College D - BC Private

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -109,9 +109,10 @@ export class ApplicationApi extends HttpBaseClient {
     studentId?: number,
   ): Promise<PaginatedResultsAPIOutDTO<ApplicationSummaryAPIOutDTO>> {
     let url =
-      AuthService.shared.authClientType === ClientIdType.AEST
+      AuthService.shared.authClientType === ClientIdType.AEST ||
+      AuthService.shared.authClientType === ClientIdType.Institution
```
dheepak-aot, line 113: Should we base this condition on studentId like we do for most of other places?

sources/packages/web/src/views/aest/student/Applications.vue
dheepak-aot, line 24: 👍

sources/packages/web/src/components/common/students/StudentApplications.vue
dheepak-aot: IMO we can do a v-if based on optional prop value which it's consumer can use.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -608,6 +611,17 @@ export class ApplicationService extends RecordDataModelService<Application> {
       .andWhere("application.applicationStatus != :overwrittenStatus", {
         overwrittenStatus: ApplicationStatus.Overwritten,
       });
+    // If institution id is present, get only the applications
+    // linked with the institution.
+
+    if (institutionId) {
```
dheepak-aot, line 617: Suggestion, we can move the where(line 610,611) beyond line 624 to follow the pattern of select, join and where because sometimes where would have a dependency with join(not in this case)

sources/packages/backend/apps/api/src/route-controllers/student/student.institutions.controller.ts
```diff
@@ -76,6 +82,26 @@ export class StudentInstitutionsController extends BaseController {
     return this.studentControllerService.getStudentProfile(studentId);
   }
 
+  /**
+   * Get the list of applications that belongs to a student for the institution.
+   * TODO: Authorization must be enabled to validate if the student has submitted
```
dheepak-aot, line 87: Minor, TODO comment seems to be not complete.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
andrepestana-aot: I think that she's trying to emphasizing that `application.data.selectedLocation` doesn't change the result here.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      auditUser: student.user,
+    });
+    fakeOffering.offeringIntensity = OfferingIntensity.fullTime;
+    const savedOffering = await educationProgramOfferingRepo.save(fakeOffering);
+    fakeOriginalAssessment.offering = savedOffering;
+    const savedOriginalAssessment = await studentAssessmentRepo.save(
+      fakeOriginalAssessment,
+    );
+    savedApplication.currentAssessment = savedOriginalAssessment;
+    await applicationRepo.save(savedApplication);
```
dheepak-aot, line 90: Do we really need this(line 75-90). Is an application for a student with provided location will be sufficient right.?

Given that `createFakeApplication` creates application in submitted status.

andrepestana-aot (commented):
Good job. I just answered some other comment to understand a test and agree with the rest of their comments.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+        count: 1,
+      });
+  });
+
+  it("Should get an empty array as application summary when student has a draft application for the institution (application without the location id saved(i.e draft/cancelled draft status)).", async () => {
+    // Arrange
+
+    // Student has a draft application to the institution.
+    const student = await saveFakeStudent(appDataSource);
+    const application = createFakeApplication({
```
dheepak-aot, line 123: . I understand the intention of going deeper but the authorization will definitely validate for a linked application where on a happy path the result could never be empty. IMO this scenario is not required.

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .expect({
+        results: [],
+        count: 0,
+      });
+  });
+
+  it("Should get an empty array as application summary when student has a submitted application in a different location.", async () => {
+    // Arrange
```
dheepak-aot, line 147: https://github.com/bcgov/SIMS/pull/1903/files#r1178370671

Test scenario can be removed.

dheepak-aot:
Nice work with applications tab @ann-aot  👍. Please have look at the comments.

sources/packages/web/src/views/student/StudentApplicationSummary.vue
ann-aot: oh..thanks for that catch, I will remove it.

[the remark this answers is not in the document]
ann-aot: it was a lost string, the change in the file is 
image

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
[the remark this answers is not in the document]
ann-aot: mm, that is a good point. I will add few TC for that

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
ann-aot: I got your point, we can do either one 
1. add one more application to the institution that belongs to the student and expect only the application that belongs to the institution? 
2. or keep it as it is and adjust the except (to return a forbidden error, I believe),  once you merge public institution authorization logic. 

I prefer the 2nd option thoughts? @andrewsignori-aot @andrepestana-aot @guru-aot

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
ann-aot: https://github.com/bcgov/SIMS/pull/1903#discussion_r1178411676

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
ann-aot: I got it, Actually, I was checking other TCs when I was doing this, and both C and D were used in the public institution tab TCs. So, I randomly chose C 😄 But I got it, I will update it. Thanks

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
andrewsignori-aot: If option 1 is not covered yet by any other E2E I would prefer it @ann-aot because the decorator will not ensure that and it would be good to have it asserted here.

sources/packages/web/src/components/common/students/StudentApplications.vue
ann-aot: I have added props to take care of these actions

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
ann-aot: ya. @andrewsignori-aot  first is already covered in the first TC. So, i will remove these

sources/packages/web/src/services/http/ApplicationApi.ts
ann-aot: mm...since it is the API service and technically it's based on the client type (same as we add client root), I don't see an issue here. Because. if it is a nonstudent client and the student is not present, according to this logic it will throw the expected error (student not found, etc..), but if we add the `student id` check, it will throw an authorization error right (because of the wrong token).?

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: mm, I find this way more readable like till 610, its the default join, and where are beyond the logic and w.r.t conditions, so, keeping as it is

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,221 @@
...
+            },
+          ],
+          count: 2,
+        });
+    },
+  );
+
+  it(
+    `Should get all the two student application details in ${FieldSortOrder.DESC} order of application number ` +
+      "as summary when student has two submitted/Inprogress application for the institution (application with location id saved) " +
```
andrewsignori-aot, line 151: Just a suggestion, I would not mind omitting the explanation about "(application with location id saved)", that said I do support the long text description 😉

sources/packages/web/src/components/common/students/StudentApplications.vue
```diff
@@ -157,6 +155,21 @@ export default defineComponent({
       type: Number,
       required: false,
     },
+    enableViewApplicationBtn: {
```
andrewsignori-aot, line 158: For properties in general I would avoid adding the component type and use an abbreviation like btn or similar.
IMO these three properties could be named as.
- `enableViewApplication`
- `allowManageApplication`
- `displayApplicationName`

sources/packages/web/src/services/http/ApplicationApi.ts
dheepak-aot: When I analyzed, my understanding is different, correct me if wrong.

For this scenario for example, if it is an AEST Client and by mistake studentId is not passed to this method,

it will form the URL like `student/undefined/application-summary`.(I believe parseint pipe will throw bad request error)

If the logic is otherwise(studentId check), and if the same happens URL `student/application-summary` will be formed and returned 404 because any non student endpoint cannot exist without studentId followed by student `student/studentId`.

I was not trying to say to go this way(by studentId way) for a technical reason as I don't see a much tangible difference, my point was as we are doing like this mostly in our application, we can follow the same here unless there is a very strong reason. 

@andrewsignori-aot @andrepestana-aot @guru-aot thoughts?

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: As there is a technical reason behind my suggestion, I would still ask for the other way unless team thinks differently

@andrewsignori-aot @andrepestana-aot @guru-aot ?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -608,6 +611,17 @@ export class ApplicationService extends RecordDataModelService<Application> {
       .andWhere("application.applicationStatus != :overwrittenStatus", {
```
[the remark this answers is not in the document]
guru-aot: As the services is used across client types, the left join actually make sense

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
dheepak-aot: Technically we don't check the offering details of an application to find if it is linked to a student, so my suggestion was to keep the application data more simple and just adequate for the test case.

sources/packages/backend/apps/api/src/utilities/datatable-config.ts
```diff
@@ -1,17 +1,8 @@
...
- */
-export enum FieldSortOrder {
-  DESC = "DESC",
-  ASC = "ASC",
-}
-/**
- * default user summary table sort order
+ *  Pagination option.
  */
-export const DEFAULT_PAGE_LIMIT = 10;
```
andrewsignori-aot, line 11: Nice cleanup 😉

sources/packages/backend/apps/api/src/services/application/application.service.ts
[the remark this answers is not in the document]
andrewsignori-aot: To be clear, the method is supposed to return all applications, even draft ones, unless I am missing something. It is not about bing used across different clients it is about what it needs to return.

andrewsignori-aot (commented):
Just minor comments.

sources/packages/web/src/components/common/students/StudentApplications.vue
ann-aot: Added `btn` only because of the last property, the last property is the `view application` action in the name. 
image.

`displayApplicationName` is confusing. do you guys have a better name, I can update it

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
dheepak-aot: @ann-aot I am not able to see the scenario where student has applications for more than one institution and the list returns only application which belongs to the particular institution.

dheepak-aot:
Thanks for making the changes. Added some clarifications and one more comment.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: I would prefer to keep it as it is right now because IMO it makes it easier to read and identify the difference for the `if (institutionId)`.
Breaking the first part of the query in this scenario would add a bit more visual complexity.

sources/packages/web/src/services/http/ApplicationApi.ts
andrewsignori-aot: I was on a chat we @ann-aot and we quickly talk about it. I believe that the suggestion is valid @dheepak-aot and we can change it as per your recommendation.

ann-aot: make sense after reading the expalination

sources/packages/backend/apps/api/src/route-controllers/student/_tests_/e2e/student.institutions.controller.getStudentApplicationSummary.e2e-spec.ts
```diff
@@ -0,0 +1,221 @@
...
+  let collegeF: Institution;
+  let collegeFLocation: InstitutionLocation;
+  let applicationRepo: Repository<Application>;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    appDataSource = dataSource;
+    // College F.
+    const { institution } = await getAuthRelatedEntities(
```
andrepestana-aot, line 37: Minor: if you do:
```
    const { institution: collegeF } = await getAuthRelatedEntities(
      appDataSource,
      InstitutionTokenTypes.CollegeFUser,
    );
```
then you won't need line 41.

andrepestana-aot (approved):
LGTM!

andrewsignori-aot (approved):

Issue #1759, opened 2023-02-27
# 13 – Ministry Re-Issuing MSFAA Number


JasonCTang:
**As a** *Ministry Staff User* **I want** *The ability to re-issue an MSFAA Number* **because** *there are instances where the status of an MSFAA issued by Province of BC (ie, issued by SIMS/SFAS) may be overriden at the federal level by another province and need the ability to issue a new MSFAA without having the student resubmit their application.

A "Re-issue MSFAA Number" button will need to be available to Ministry staff when the current MSFAA for a student is "cancelled" (ie, the cancel date is populated).

When a new MSFAA is issued, the student must be notified.
When a new MSFAA is issued, it needs to be reassociated with all pending disbursements.
If NOA has disbursements with different MSFAA, we will show both numbers separately in the same NOA.
Any prior assessments/reassessments NOAs will display the previous MSFAA number. 
For each MSFAA, we are going to display messaging indicating if it is in pending signing state and cancelled state. 
 (Signed/received state will have no messaging)

**Whether a new MSFAA is created by the action of the Reissue button or a new Application, reassociate with ALL pending disbursements for the student and for that offering intensity.**
**Reissue button should only be available if the pending disbursements for the same offering intensity has a cancelled MSFAA.**
**Same logic is going to be centralized between the Application Workflow and the Ministry Reissue Button.**

**UI/UX and messaging**

Mockup: https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=11206-138465&t=skhfaZTJUvv6bkH7-0

Ministry journey (in workflows): https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=11467-157807&t=skhfaZTJUvv6bkH7-0

- [ ] MSFAA waiting to be signed messaging
- [ ] MSFAA cancelled messaging
- [ ] MSFAA cancelled email notification

**Additional Context**
- MSFAA is an agreement between student and federal government.  We just produce the number for the student and provide it to the federal government and we receive a notification from the federal government through a file when it is signed.  There's no real need to understand the cause of a cancellation, we just need to understand that the federal government can also tell us through the file when an MSFAA is cancelled.
- There is already logic in place to hold the disbursement when the cancellation date is populated for the MSFAA record associated to the application.   
- MSFAA can be rendered inactive at the federal level at any time in the study period. I.e. it can happen between two disbursements.
[...]

PR #1928, opened 2023-05-05
#1759 - Reissue MSFAA - Part 1 - New SIMS API Endpoint

andrewsignori-aot:
- Methos for MSFAA creation (used during assessment) and reissue were centralized to execute the below process in the same way.
    - Ensure that any pending MSFAA is canceled. A pending MSFAA is a never signed or canceled record.
    - Create the new MSFAA number.
    - Associate the new MSFAA number with any pending disbursement for the same student and same offering intensity.
- Created a new API endpoint under the Ministry application controller to reissue the MSFAA.

### E2E Tests for reissuing the MSFAA from SIMS API
ApplicationAESTController(e2e)-reissueMSFAA
    √ Should reissue an MSFAA and associate with both disbursements when both disbursements are pending and the current MSFAA is signed but canceled. (449 ms)
    √ Should reissue an MSFAA and associate with only the pending disbursement when one disbursement is not pending and the other is pending and the current MSFAA is canceled. (228 ms)  
    √ Should reissue an MSFAA and associate with only pending disbursements across multiple applications for the same offering intensity and same student when some disbursements are pending, some are not and there is at least on application from a different intensity. (427 ms)
    √ Should throw a UnprocessableEntityException when the associated MSFAA is not cancelled. (152 ms)
    √ Should throw a UnprocessableEntityException when there is no pending disbursement associated with the application. (141 ms)
    √ Should throw a UnprocessableEntityException when the application is not in the expected status. (59 ms)
    √ Should throw a NotFoundException when the application does not exist. (11 ms)


sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
```diff
@@ -0,0 +1,266 @@
...
+      // This will ensure that the newly created MSFAA will be the only one valid.
+      await entityManager.getRepository(MSFAANumber).update(
+        {
+          student: { id: studentId },
+          offeringIntensity,
+          dateSigned: IsNull(),
+          cancelledDate: IsNull(),
+        },
+        {
+          cancelledDate: nowISODate,
```
ann-aot, line 131: mm, I was thinking do we need any other indication to know this process happened as part of reissuing the new MSAFAA and was not canceled as part of the file integration? or do we care about that?

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
```diff
@@ -0,0 +1,266 @@
...
+   * pending MSFAA for the particular offering intensity and also associating the
+   * new MSFAA number to any pending disbursement for the same offering intensity.
+   * Reissuing an MSFAA is only valid for an application with pending disbursement
+   * where the MSFAA is cancelled.
+   * @param referenceApplicationId reference application id.
+   * @returns the newly created MSFAA.
+   */
+  async reissueMSFAA(
+    referenceApplicationId: number,
+    auditUserId: number,
```
ann-aot, line 71: pls add it to the comment

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.reissueMSFAA.e2e-spec.ts
```diff
@@ -0,0 +1,355 @@
...
+    db = createE2EDataSources(dataSource);
+  });
+
+  it("Should reissue an MSFAA and associate with both disbursements when both disbursements are pending and the current MSFAA is signed but canceled.", async () => {
+    // Arrange
+    const student = await saveFakeStudent(db.dataSource);
+    const currentMSFAA = createFakeMSFAANumber(
+      { student },
+      {
+        state: MSFAAStates.Signed | MSFAAStates.CancelledOtherProvince,
```
ann-aot, line 40: won't this give the sum of 1 and 4 that is 5 
image
does, it satisfy the condition that we need?

sources/packages/backend/libs/test-utils/src/factories/msfaa-number.ts
```diff
@@ -7,26 +7,76 @@ import {
...
-  msfaaNumber.dateRequested = now;
-  msfaaNumber.dateSigned = dateOnly;
-  msfaaNumber.serviceProviderReceivedDate = dateOnly;
-  msfaaNumber.offeringIntensity = OfferingIntensity.fullTime;
-  msfaaNumber.cancelledDate = null;
-  msfaaNumber.newIssuingProvince = null;
-  msfaaNumber.student = relations.student;
-  msfaaNumber.referenceApplication = relations?.referenceApplication;
+
+  if (!options?.state || options.state & MSFAAStates.Pending) {
```
ann-aot, line 58: mm, I am not following these conditions

sources/packages/backend/libs/test-utils/src/factories/msfaa-number.ts
```diff
@@ -7,26 +7,76 @@ import {
...
+/**
+ * Represents the different MSFAA scenarios.
+ * More than one state can be combined, for instance,
+ * to create a signed MSFAA that was also cancelled.
+ */
+export enum MSFAAStates {
+  Pending = 0,
+  Signed = 1,
+  CancelledSystem = 2,
+  CancelledOtherProvince = 4,
```
ann-aot, line 19: are we not using `3` due to bitwise operations?

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -56,4 +72,43 @@ export class ApplicationAESTController extends BaseController {
...
   }
+
+  /**
+   * Creates a new MSFAA number to be associated with the student, cancelling any
+   * pending MSFAA for the particular offering intensity and also associating the
+   * new MSFAA number to any pending disbursement for the same offering intensity.
+   * @param applicationId reference application id.
+   * @returns the newly created MSFAA.
+   */
+  @Post(":applicationId/reissue-msfaa")
```
dheepak-aot, line 83: While reissuing msfaa, wouldn't it be better to have the student as part of URI as we issue/re-issue msfaa for a student even though it happens on an application.?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.reissueMSFAA.e2e-spec.ts
andrewsignori-aot: Yes, please review how the check is done.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
andrewsignori-aot: Honestly, I do not see any benefit, feel free to clarify if needed.
The MSFAA is issued for a student because of an application. Having the application and a student as a separate parameters opens the possibility of even providing a student from a different application.

ann-aot (commented):
Good work @andrewsignori-aot. Added some comments

ann-aot (approved):
LGTM 👍 Thanks for doing the changes

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    offeringIntensity: OfferingIntensity,
+    auditUserId: number,
+  ): Promise<MSFAANumber> {
+    return this.dataSource.transaction(async (entityManager) => {
+      const auditUser = await this.systemUsersService.systemUser();
+      const now = new Date();
+      const nowISODate = getISODateOnlyString(now);
+      // Cancel any pending MSFAA record that was never reported as signed or cancelled.
+      // This will ensure that the newly created MSFAA will be the only one valid.
+      await entityManager.getRepository(MSFAANumber).update(
```
dheepak-aot, line 124: I see that is is a very rare edge case scenario where 

Student submit an application where MSFAA was valid during submission and before the disbursement was sent it expired and a new application got created with new MSFAA in the mean time. 

Hence when reissuing msfaa we are checking for this MSFAA which was never reported or cancelled.

But I have a question here, while re-issuing should we not use this non-reported msfaa number? Is there anything like a policy which stops us doing it?

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
```diff
@@ -0,0 +1,269 @@
...
+        id: referenceApplicationId,
+      } as Application;
+      newMSFAANumber.offeringIntensity = offeringIntensity;
+      newMSFAANumber.creator = { id: auditUserId } as User;
+      newMSFAANumber.createdAt = now;
+      await entityManager.getRepository(MSFAANumber).save(newMSFAANumber);
+      // Associate pending disbursements with the new MSFAA.
+      const disbursementScheduleRepo =
+        entityManager.getRepository(DisbursementSchedule);
+      const schedulesToUpdate = await disbursementScheduleRepo.find({
```
dheepak-aot, line 153: We can directly use update like this right?

```ts

await disbursementScheduleRepo.update(
        {
          disbursementScheduleStatus: DisbursementScheduleStatus.Pending,
          studentAssessment: {
            application: { student: { id: studentId } },
            offering: { offeringIntensity: offeringIntensity },
          },
        },
        {
          msfaaNumber: newMSFAANumber,
          modifier: auditUser,
          updatedAt: now,
        },
      );
```

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
```diff
@@ -0,0 +1,269 @@
...
+        `Application id ${applicationId} was not found.`,
+        APPLICATION_NOT_FOUND,
+      );
+    }
+    const nowAllowedApplicationStatuses = [
+      ApplicationStatus.Draft,
+      ApplicationStatus.Cancelled,
+      ApplicationStatus.Overwritten,
+    ];
+    if (nowAllowedApplicationStatuses.includes(application.applicationStatus)) {
```
dheepak-aot, line 254: Why didn't we include this condition in the query?

Also the Query can be based on status [Assessment,Enrolment,Completed] which will pick just the eligible one.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.reissueMSFAA.e2e-spec.ts
andrepestana-aot: Sorry but I didn't understand why using the bitwise operations here.

dheepak-aot, line 40: It is as good as using `state:MSFAAStates.Signed` right?  or am I missing something?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.reissueMSFAA.e2e-spec.ts
```diff
@@ -0,0 +1,355 @@
...
+      { student, msfaaNumber: currentMSFAA },
+      {
+        applicationStatus: ApplicationStatus.Completed,
+        createSecondDisbursement: true,
+      },
+    );
+    const [firstSchedule, secondSchedule] =
+      application.currentAssessment.disbursementSchedules;
+    const endpoint = `/aest/application/${application.id}/reissue-msfaa`;
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
```
dheepak-aot, line 55: Just curious to know, couldn't find the AEST group who can re-issue msfaa, is it BusinessAdministrators?

dheepak-aot:
Great work @andrewsignori-aot . Added few comments.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
dheepak-aot: I was thinking back and forth but what you said makes sense.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.reissueMSFAA.e2e-spec.ts
andrewsignori-aot: @andrepestana-aot please let me know if need more information and please review the code for the `createFakeMSFAANumber` method.
https://github.com/bcgov/SIMS/pull/1928#discussion_r1189112182

sources/packages/backend/libs/test-utils/src/factories/msfaa-number.ts
andrepestana-aot: So, wouldn't it be good to comment here that a MSFAAStates 3 should be a "Signed" + "CancelledSystem"? That was my understanding at least.

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
andrewsignori-aot: This was my first attempt. Typeorm does not allow it apparently and, unless I am missing something, this is probably the reason why we do not have any other similar update like this.
Please notice that the main difference is that the "where" clause for this would include relations.

andrepestana-aot (commented):
Good job. I'm just not very sure about the bitwise operations idea and left some comments.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.reissueMSFAA.e2e-spec.ts
andrewsignori-aot: As mentioned today at the daily meeting, there was no requirement for a specific role. It will be added to the ticket and taken care of in the upcoming PR for this ticket 😉

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
andrewsignori-aot: Related to "_Why didn't we include this condition in the query_?": the idea was to provide a more a specific error instead of a generic "Application not found" for this situation. I believe that in the past we talked about both approaches being acceptable.

About checking for "_Assessment,Enrolment,Completed_", the MSFAA as per the assessment workflow right now can be generated while the application is "In Progress", for instance (unless we change it). The idea was only to provide a certain level of validation and ensure that the offering would be present.

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
andrewsignori-aot: I was trying to recall the exact reason to have them canceled actually. The goal was to ensure that when a new MSFAA is created there will be only one "not signed".
For instance, the student created an application, generated an MSFAA, and never completed it. After 3 years the same student comes back and a new MSFAA will be generated, in this situation that old record will be "cancelled by the system".
I believe that the idea was just to keep the DB looking better with only one possible not-signed/not-canceled MSFAA.

sources/packages/backend/libs/test-utils/src/factories/msfaa-number.ts
andrewsignori-aot: The combination between the enum items is free. It can be as below or others.
- Pending 
- Signed 
- Signed | CancelledSystem 
- Signed | CancelledSystem |  CancelledOtherProvince

andrepestana-aot (approved):
Thanks for the explanation.

dheepak-aot (approved):
Thanks for the clarifications. LGTM. 👍

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -56,4 +72,43 @@ export class ApplicationAESTController extends BaseController {
...
+  @ApiNotFoundResponse({ description: "Application id not found." })
+  @ApiUnprocessableEntityResponse({
```
guru-aot, line 85: Cool implementation 👍

sources/packages/backend/libs/services/src/msfaa-number/msfaa-number-shared.service.ts
```diff
@@ -0,0 +1,269 @@
...
+    const application = await this.getApplicationForMSFAACreation(
+      referenceApplicationId,
+    );
+    const pendingDisbursement =
+      application.currentAssessment.disbursementSchedules.find(
+        (schedule) =>
+          schedule.disbursementScheduleStatus ===
+          DisbursementScheduleStatus.Pending,
+      );
+    if (!pendingDisbursement) {
```
guru-aot, line 83: 👍

guru-aot (approved):
Nice work @andrewsignori-aot

PR #1932, opened 2023-05-09
Segregate db migrations deployment

guru-aot:


.github/workflows/build-deploy-dev.yml
```diff
@@ -125,23 +125,22 @@ jobs:
...
         run: |
           oc login --token=${{ secrets.SA_TOKEN }} --server=https://api.silver.devops.gov.bc.ca:6443
-      # Build the images
-      - name: Build SIMS-API Image
+          if [ $? -ne 0 ]; then
+            echo "Failed to log in to OpenShift. Please check the credentials and connection details."
+            exit 1
+          fi
+       # Build the images in parallel.
+      - name: Build db-migrations, api, workers, queues and web Images in parallel
```
ann-aot, line 133: `db-migrations` will run first, then rest in parallel right?

.github/workflows/build-deploy-dev.yml
```diff
@@ -125,23 +125,22 @@ jobs:
...
-        working-directory: "./devops"
+            make oc-build-db-migrations &
+            make oc-build-api &
+            make oc-build-workers &
+            make oc-build-queue-consumers &
+            make oc-build-web
+      - name: Wait for parallel builds to finish
         run: |
-          make oc-build-web
+          wait
```
ann-aot, line 143: for my info, what happens here

.github/workflows/build-deploy-dev.yml
ann-aot: oh.. since it is deployed we are doing everything in parallel?

.github/workflows/build-deploy-dev.yml
```diff
@@ -197,23 +196,27 @@ jobs:
...
-      - name: Deploy Web/Frontend
-        working-directory: "./devops/"
+          make oc-deploy-api &
+          make oc-deploy-workers &
+          make oc-deploy-queue-consumers &
+          make oc-deploy-web &
+      - name: Wait for parallel deployments to finish
         run: |
-          make oc-deploy-web
+          wait
```
ann-aot, line 222: will it wait for all processes to finish or it wait for a certain time? how is it for my info

devops/Makefile
```diff
@@ -34,6 +35,17 @@ export FORMIO_ROOT_EMAIL := $(or ${FORMIO_ROOT_EMAIL}, dev_sabc@gov.bc.ca)
 export MONGODB_URI := $(or ${MONGODB_URI}, $$MONGODB_URI)
 export QUEUE_PREFIX := $(or $(QUEUE_PREFIX), {sims-local})
 
+export MAX_WAIT=300 # Maximum wait time in seconds
```
ann-aot, line 38: is this related to https://github.com/bcgov/SIMS/pull/1932/files#r1191748847

ann-aot (commented):
Added some questions @guru-aot . I would like to get a walk through

.github/workflows/build-deploy-dev.yml
guru-aot: Yes, you are right

.github/workflows/build-deploy-dev.yml
guru-aot: It just waits until all the parallel process is successful

.github/workflows/build-deploy-dev.yml
guru-aot: Not for certain time, once the parallel tasks are executed, it just waits like promise.all

devops/Makefile
guru-aot: No this variable is used to check whether the dbmigration job is completed within the specified time

.github/workflows/build-deploy-dev.yml
```diff
@@ -125,23 +125,22 @@ jobs:
...
-        working-directory: "./devops"
-        run: |
-          make oc-build-workers
-      - name: Build Queue Consumers Image
-        working-directory: "./devops"
-        run: |
-          make oc-build-queue-consumers
-      - name: Build Web/Frontend Image
-        working-directory: "./devops"
+            make oc-build-db-migrations &
```
andrewsignori-aot, line 136: Why this approach was taken instead of using the steps as they were before? 
The benefit from the previous way was also to have better UI process descriptions on Github actions, as shown below.
image

We also have other examples about how to run jobs in parallel, for instance on file `tests.yml`, and also how to run in parallel and create a dependency where one job must be finished to another start, for instance on the file `deploy-all.yml` as shown below.
image

If there are benefits in the approach used in this PR, can you please clarify them?

.github/workflows/build-deploy-dev.yml
```diff
@@ -197,23 +196,27 @@ jobs:
...
         run: |
           oc login --token=${{ secrets.SA_TOKEN }} --server=https://api.silver.devops.gov.bc.ca:6443
-      # Deploy SIMS-Api
-      - name: Deploy SIMS-API
-        working-directory: "./devops/"
-        run: |
-          make oc-deploy-api
-      # Deploy Workers
-      - name: Deploy Workers
+          if [ $? -ne 0 ]; then
```
andrewsignori-aot, line 199: When the steps failed they usually will indicate it in a nice way in the UI, why do we need this additional check here?
For instance,
image

.github/workflows/deploy-all.yml
```diff
@@ -8,6 +8,7 @@ on:
         required: true
         default: ""
       environment:
+        description: 'Openshift Environment'
```
andrewsignori-aot, line 11: I guess that we have been using double quotes along all files.

andrewsignori-aot, line 11: If `Openshift Environment` is related to the `Openshift Namespace` it would not be accurate because we have DEV/TEST/PROD namespaces but we have, for instance, DEV/TEST/QA environments.
I would avoid calling this `Openshift Environment`.

devops/Makefile
```diff
@@ -285,6 +307,33 @@ create-new-db:
...
+		-p NAME=$(DB_MIGRATIONS) \
+		-p DB_SERVICE=$(DB_SERVICE_NAME) \
+		-p BUILD_NAMESPACE=$(BUILD_NAMESPACE) \
+		-p IMAGE_STREAM_TAG="$(DB_MIGRATIONS_BUILD_REF):$(BUILD_REF)" \
+		-p BUILD_TAG=$(BUILD_REF) \
+		-p DATABASE_NAME_KEY=$(DATABASE_NAME_KEY) \
+		-p JOB_NAME=$(DB_MIGRATIONS_JOB_NAME) \
+		| oc -n $(NAMESPACE) apply -f - 
+
+wait-for-db-migrations:
```
andrewsignori-aot, line 329: Are we taking this approach using `wait-for-db-migrations` because there is no way to call something **similar** to `$(call rollout_and_wait,dc/$(API))` as API, Workers, etc.?

devops/openshift/api-deploy.yml
```diff
@@ -245,6 +248,18 @@ objects:
...
+    metadata:
+      name: ${NAME}-hpa
+    spec:
+      scaleTargetRef:
+        apiVersion: apps/v1
+        kind: ReplicationController
+        name: ${NAME}-rc
+      minReplicas:  "${{REPLICAS}}"
+      maxReplicas: 10
+      targetCPUUtilizationPercentage: 80
```
andrewsignori-aot, line 262: ❤️

devops/openshift/api-deploy.yml
```diff
@@ -245,6 +248,18 @@ objects:
...
+    kind: HorizontalPodAutoscaler
+    metadata:
+      name: ${NAME}-hpa
+    spec:
+      scaleTargetRef:
+        apiVersion: apps/v1
+        kind: ReplicationController
+        name: ${NAME}-rc
+      minReplicas:  "${{REPLICAS}}"
+      maxReplicas: 10
```
andrewsignori-aot, line 261: Does it take into consideration the available quota and stop creating more replicas?
I noticed that all PODs are configured to a max of 10, just wondering if we need to plan the maxReplicas as per the POD application type.

devops/openshift/db-migrations-job.yml
```diff
@@ -0,0 +1,87 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: ${NAME}
+  annotations:
+    description: "Template for running a job that performs a database migrations"
+labels:
+  project: ${PROJECT}
+  service: ${SERVICE_NAME}
+parameters:
```
andrewsignori-aot, line 10: Is there a benefit to keeping the parameters at the top? I believe that other files have them all at the bottom.

sources/packages/backend/apps/db-migrations/tsconfig.app.json
```diff
@@ -2,7 +2,6 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "declaration": false,
-    "sourceMap": false,
```
andrewsignori-aot, line 5: Was it removed because by default it is already false?

.github/workflows/build-deploy-dev.yml
andrewsignori-aot: The make `oc-something` here, in general, are appended with `--wait` does it not mean that they will be waiting to be executed one after another?

sources/packages/backend/apps/db-migrations/tsconfig.app.json
guru-aot: no if we have this sourceMap defined here, it will override the tsconfig.json configuration. The default configuration is true in our tsconfig, so to ignore the overriding and generate the map.js files, I have removed it here.

devops/openshift/api-deploy.yml
guru-aot: We cannot create pods more than the available quota, it will not create more replicas if the resource is not available in openshift. During that situation, sysdig alerts come into place, that alerts the users that the overloading is happening and the cluster needs some attention. I mentioned the max replicas as 10, because depending on the load, different containers will scaleup to different replica sets, and it will come down once the utilization is less than 80%. I believe at a point not all the containers will go to replicate 10. Hopefully not

guru-aot: We cannot create pods more than the available quota, it will not create more replicas if the resource is not available in openshift. During that situation, sysdig alerts come into place, that alerts the users that the overloading is happening and the cluster needs some attention. I mentioned the max replicas as 10, because depending on the load, different containers will scaleup to different replica sets, and it will come down once the utilization is less than 80%. I believe at a point not all the containers will go to replicate 10. Hopefully not

devops/Makefile
guru-aot: Yes for jobs we have only wait, and check the status. So enhanced a little bit with some max timeout and checking status for every 5 sec. As this is the main job before the other jobs run.

.github/workflows/build-deploy-dev.yml
guru-aot: No, if the oc-build-db-migrations target in the make command includes --wait in its command, it does not automatically wait for the other make commands to finish. The --wait option typically applies to the specific build process triggered by that particular make target.

.github/workflows/build-deploy-dev.yml
guru-aot: You are correct that failed steps in a GitHub Actions workflow typically indicate the failure in the GitHub UI, making it easy to identify the cause of the failure. The additional check in the provided code snippet is a manual validation step that provides an explicit error message and exits the workflow with a failure status code.
Also by including the check and exit statement, the workflow consistently handles OpenShift login failures. It ensures that the workflow stops and reports a failure in a predictable manner, regardless of any potential variations in the behavior or error reporting of the oc login command itself.

.github/workflows/build-deploy-dev.yml
guru-aot: Both are similar approaches, but the choice I made provides additional checks and error handling to ensure that the database migrations are successfully executed and that all the parallel deployments complete before continuing with the workflow. It enhances the reliability of the deployment process by explicitly handling potential failures and ensuring proper synchronization.
Also the approach on handing the build and deployments as a single line, is just an option, I can revert it back, but we are not going to lose any logs and start and end of the make targets, using this approach and its descriptive in knowing that the jobs run in parallel.

devops/openshift/db-migrations-job.yml
```diff
@@ -0,0 +1,87 @@
+apiVersion: template.openshift.io/v1
+kind: Template
```
dheepak-aot, line 2: Correct me if I am wrong, why is the kind given here as template and not job?

The reference from user story shows a sample job specification like this?

image

devops/openshift/db-migrations-job.yml
```diff
@@ -0,0 +1,87 @@
...
+    kind: Job
+    metadata:
+      name: ${JOB_NAME}
+    spec:
+      template:
+        metadata:
+          name: ${JOB_NAME}
+        spec:
+          containers:
+            - name: db-migrations-container
```
dheepak-aot, line 58: Timeout and retries? 

image

sources/packages/backend/apps/db-migrations/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Base Image
+FROM registry.access.redhat.com/ubi8/nodejs-16:1-5
+
+LABEL maintainer="BC GOV"
+
+# Copying package.json and package-lock.json
+COPY package*.json ./
+
+# Installing dependencies
+RUN npm ci --ignore-scripts
```
dheepak-aot, line 10: I see that we are trying to limit the packages here to what has been present in package.json. Are we sure of no side effects?

sources/packages/backend/apps/db-migrations/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Copying sources
+COPY apps/db-migrations ./apps/db-migrations
+COPY tsconfig.json ./
+COPY tsconfig.build.json ./
```
dheepak-aot, line 15: We can combine line 14 and 15 to be 

COPY ./*.json ./

sources/packages/backend/apps/api/Dockerfile
```diff
@@ -3,7 +3,6 @@ FROM registry.access.redhat.com/ubi8/nodejs-16:1-5
 
 LABEL maintainer="BC GOV"
 
-
```
dheepak-aot, line 6: Was this an intentional modification?

devops/openshift/db-migrations-job.yml
guru-aot: The template contains the job, a template commonly used to encapsulates the necessary configuration parameters and objects, making it easier to be used in the job or the deployment configs consistently.
image

devops/openshift/db-migrations-job.yml
guru-aot: Yes the job has the maximum duration the job can run and number of retries, I've started handling it in the "wait-for-db-migrations" of makefile by checking the status of the job every 5 sec and so did not add them here, it will be a duplicate and during the github actions run, the failure is shown in the logs.

devops/openshift/workers-deploy.yml
```diff
@@ -77,6 +77,18 @@ objects:
                 requests:
                   cpu: ${CPU_REQUEST}
                   memory: ${MEMORY_REQUEST}
+  - apiVersion: autoscaling/v1
+    kind: HorizontalPodAutoscaler
```
dheepak-aot, line 81: Should we consider this for StatefulSet as well?

As the horizontal pod autoscaling is not in the scope of the ticket, we can discuss with team to either add this to ticket or do the autoscaling as a seperate effort considering if we need to include the statefulset or not and any further analysis required including the testing of this capability.

sources/packages/backend/apps/db-migrations/Dockerfile
guru-aot: using npm i --ignore-scripts is entirely different from using npm ci --ignore-scripts
npm i will execute lifecycle scripts associated with those packages. These scripts can include pre-install, post-install, pre-publish, and other custom scripts defined by the package author. Using --ignore-scripts will entirely skip all such scripts.
npm ci is used to download dependencies using the package-lock.json, by default npm ci does not run any  lifecycle scripts. However for the sake of sonarcloud, we are adding --ignore-scripts in our Dockerfile.
I am sure this will not have any side effects.

devops/openshift/workers-deploy.yml
dheepak-aot: Aslo disruptionBudget is missing for this deployment config. Any particular reason?

guru-aot: Nice catch I have added it now

sources/packages/backend/apps/api/Dockerfile
guru-aot: no just syncing with theother dockerfiles

.github/workflows/deploy-all.yml
```diff
@@ -33,89 +34,180 @@ on:
...
         run: |
           make oc-deploy-web
+
+  # Deploy Camunda Definitions
   deployCamundaDefinitions:
     if: ${{ github.event.inputs.deployCamundaDefinitions == 'true' }}
     name: Deploy BPMNs and DMNs to Camunda
     needs:
-      - deploy
+      - deploy-sims-api
```
dheepak-aot, line 196: Should needs consider workers also?

dheepak-aot: May be queue consumers also.

.github/workflows/build-deploy-dev.yml
andrewsignori-aot: I got the point when we talked about the commands concatenation with the "&".
Either way, as per the Team discussion, the approach with the GitHub parallel jobs looks great.

.github/workflows/build-deploy-dev.yml
andrewsignori-aot: Thanks for doing the changes with the parallel GitHub actions as per the Team discussions 😉

devops/Makefile
andrewsignori-aot: Thanks for doing the changes and removing the `wait-for-db-migrations`.

sources/packages/backend/apps/db-migrations/tsconfig.app.json
andrewsignori-aot: Thanks for reverting the file.

.github/workflows/build-deploy-dev.yml
andrewsignori-aot: Thanks for reverting these changes as per the Team discussion.

.github/workflows/deploy-all.yml
andrewsignori-aot: It is a good point. Maybe we can wait for all the deployments to be successful before deploying definitions.

andrewsignori-aot (approved):
Great work and thanks for doing the changes 👍

.github/workflows/deploy-all.yml
guru-aot: If I am not wrong,  they do not have any dependencies with any of the deployments. We can deploy it parallel if the inputs are checked. What you say?? @dheepak-aot @andrewsignori-aot @andrepestana-aot @sh16011993

andrepestana-aot (approved):
Great job! Looks good!

sources/packages/backend/apps/db-migrations/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Base Image
+FROM registry.access.redhat.com/ubi8/nodejs-16:1-5
+
+LABEL maintainer="BC GOV"
+
+# Copying package.json and package-lock.json
+COPY ./package*.json ./
```
sh16011993: for my info, is it ok to copy in the container's root directory. Shouldn't a `WORKDIR` be specified here?

.github/workflows/deploy-all.yml
andrewsignori-aot: I believe that if we enforce that all the changes will be backward compatible (and we will as much as possible), we can have them deployed in parallel also.

sources/packages/backend/apps/db-migrations/Dockerfile
guru-aot: No need at this context we are using the default /opt/app-root/src.

.github/workflows/deploy-all.yml
andrewsignori-aot: For Camunda definitions, assuming that we have a new worker task that will need an additional input parameter, the new worker app will need to be compatible with current and new versions of the bpmn, but the opposite may be not right.
My point is, maybe forms can be dependent on the API, and Camunda definitions may be dependent on workers.

dheepak-aot (approved):
Great work @guru-aot . Thanks for making the changes.

sh16011993 (approved):
LGTM @guru-aot  👍

.github/workflows/deploy-all.yml
dheepak-aot: Also deploying Camunda definitions could be dependent on Queue consumers, where the structure of payload to start workflow instance or something like that may have changed and could be in sync with bpmn updates.

But we can discuss this in upcoming PR.

Issue #1966, opened 2023-05-25
Fix SonarCloud critical and major Issues


sh16011993:
**As a** *developer* **I want** *to fix the critical and major sonarcloud issues* **So As** *to improve the code quality*

**Acceptance Criteria**
- Remove the critical and major issues showing up in the sonarcloud.

PR #1973, opened 2023-05-29
#1966 - Fix sonarcloud critical and major issues (part 2)

sh16011993:
As a part of this PR, fixed the major issues (bugs) reported by SonarCloud.

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/main.ts
+++ b/sources/packages/backend/apps/db-migrations/src/main.ts
@@ -1,7 +1,7 @@
 import { DataSource } from "typeorm";
 import { ormConfig } from "./data-source";
 
-(async function bootstrap() {
+(async () => {
   const migrationDataSource = new DataSource({
     ...ormConfig,
     logging: ["error", "warn", "info"],
--- a/sources/packages/backend/apps/db-migrations/src/main.ts
+++ b/sources/packages/backend/apps/db-migrations/src/main.ts
@@ -23,4 +23,6 @@ async function bootstrap() {
     await dataSource.destroy();
   }
 }
-bootstrap();
+(async () => {
+  await bootstrap();
+})();
--- a/sources/packages/backend/apps/db-migrations/src/main.ts
+++ b/sources/packages/backend/apps/db-migrations/src/main.ts
@@ -1,7 +1,7 @@
 import { DataSource } from "typeorm";
 import { ormConfig } from "./data-source";
 
-async function bootstrap() {
+(async function bootstrap() {
   const migrationDataSource = new DataSource({
     ...ormConfig,
     logging: ["error", "warn", "info"],
@@ -22,7 +22,4 @@ async function bootstrap() {
   } finally {
     await dataSource.destroy();
   }
-}
-(async () => {
-  await bootstrap();
 })();
--- a/sources/packages/backend/apps/queue-consumers/src/main.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/main.ts
@@ -9,7 +9,7 @@ import { Queue } from "bull";
 import { QueueConsumersModule } from "./queue-consumers.module";
 import * as basicAuth from "express-basic-auth";
 
-(async function bootstrap() {
+(async () => {
   const app = await NestFactory.create(QueueConsumersModule);
   const config = app.get<ConfigService>(ConfigService);
   // Queue service.
--- a/sources/packages/backend/apps/queue-consumers/src/main.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/main.ts
@@ -42,4 +42,6 @@ async function bootstrap() {
 
   await app.listen(config.queueConsumersPort);
 }
-bootstrap();
+(async () => {
+  await bootstrap();
+})();
--- a/sources/packages/backend/apps/queue-consumers/src/main.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/main.ts
@@ -9,7 +9,7 @@ import { Queue } from "bull";
 import { QueueConsumersModule } from "./queue-consumers.module";
 import * as basicAuth from "express-basic-auth";
 
-async function bootstrap() {
+(async function bootstrap() {
   const app = await NestFactory.create(QueueConsumersModule);
   const config = app.get<ConfigService>(ConfigService);
   // Queue service.
@@ -41,7 +41,4 @@ async function bootstrap() {
   );
 
   await app.listen(config.queueConsumersPort);
-}
-(async () => {
-  await bootstrap();
 })();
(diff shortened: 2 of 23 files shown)


sources/packages/backend/apps/db-migrations/src/main.ts
```diff
@@ -23,4 +23,6 @@ async function bootstrap() {
     await dataSource.destroy();
   }
 }
-bootstrap();
+(async () => {
```
andrewsignori-aot, line 26: The code follows the Nestjs boilerplate.
If we need to change the `bootstrap` to be "auto-invoked" I believe that we may do it on the main method then, as shown below, or mark these ones as a false positive on Sonarcloud. Thoughts?
```ts
(async ( ) => {
  const migrationDataSource = new DataSource({
    ...ormConfig,
    logging: ["error", "warn", "info"],
  });
  const dataSource = await migrationDataSource.initialize();
  try {
    console.info("**** Running setupDB ****");
    await dataSource.query(`CREATE SCHEMA IF NOT EXISTS ${ormConfig.schema};`);
    await dataSource.query(`SET search_path TO ${ormConfig.schema}, public;`);
    await dataSource.query(`SET SCHEMA '${ormConfig.schema}';`);
    console.info(`**** Running migration ****`);
    await dataSource.runMigrations();
    console.info(`**** Running setupDB: [Complete] ****`);
  } catch (error) {
    console.error(`Exception occurs during setup db process: ${error}`);
    console.dir(ormConfig);
    throw error;
  } finally {
    await dataSource.destroy();
  }
})();
```

sources/packages/backend/libs/services/src/disbursement-schedule/disbursement-schedule-shared.service.ts
```diff
@@ -129,7 +129,7 @@ export class DisbursementScheduleSharedService extends RecordDataModelService<Di
       assessment.modifier = auditUser;
       // Adjust the saved grants disbursements with the values already disbursed.
       // !Intended to process only grants (CanadaGrant/BCGrant)
-      this.applyGrantsAlreadyDisbursedValues(
+      await this.applyGrantsAlreadyDisbursedValues(
```
andrewsignori-aot, line 132: I believe that the method is wrongly marked with `async`. Please check if the `async` can be removed from the method and remove the `await` add here.

sources/packages/web/src/main.ts
```diff
@@ -33,7 +33,7 @@ import FooterButtons from "@/components/generic/FooterButtons.vue";
 import TitleValue from "@/components/generic/TitleValue.vue";
 import StatusInfoLabel from "@/components/generic/StatusInfoLabel.vue";
 
-AppConfigService.shared.init().then(() => {
+await AppConfigService.shared.init().then(() => {
```
andrewsignori-aot, line 36: The await does not make sense in this case because it is not inside an `async` method. Was this a Sonarcloud suggestion?

andrewsignori-aot (commented):
Thanks for checking the issues, please take a look at the comments.

sources/packages/backend/apps/db-migrations/src/main.ts
andrepestana-aot: That's fine having the IIFE, I think.

sh16011993: I would incline to following the above suggested approach rather than marking these as false positives by Sonar

andrepestana-aot (commented):
Good job. Just take a look at Andrew's comments.

sources/packages/backend/libs/services/src/disbursement-schedule/disbursement-schedule-shared.service.ts
sh16011993: There are multiple promises being resolved inside the anonymous function passed as an argument to `this.dataSource.transaction`, therefore I feel `async` is required in this case.

guru-aot (commented):
Done with the review, please have a look on the other Devs comments.

sources/packages/web/src/main.ts
sh16011993: Yes, you are correct. `Await` does not make sense here because it is already being handled in the `.then()`. However, the Sonarcloud was complaining about this issue because the `.then()` was being used without the rejection handler. I have added `.catch()` now and this handles the rejection scenario fixing the sonar complaint.

sources/packages/web/src/main.ts
```diff
@@ -33,29 +33,32 @@ import FooterButtons from "@/components/generic/FooterButtons.vue";
...
+      .component("HeaderNavigator", HeaderNavigator)
+      .component("Banner", Banner)
+      .component("formio", formio)
+      .component("ToggleContent", ToggleContent)
+      .component("FormioContainer", FormioContainer)
+      .component("FooterButtons", FooterButtons)
+      .component("TitleValue", TitleValue)
+      .mount("#app");
+  })
+  .catch((e) => e);
```
andrewsignori-aot, line 64: What is the purpose of the `catch` here?

sh16011993: Callback when the promise is rejected

sources/packages/backend/apps/queue-consumers/src/main.ts
```diff
@@ -9,7 +9,7 @@ import { Queue } from "bull";
 import { QueueConsumersModule } from "./queue-consumers.module";
 import * as basicAuth from "express-basic-auth";
 
-async function bootstrap() {
+(async () => {
```
ann-aot, line 12: just for my understanding. what was the error sonarcloud was complaining about here?

sources/packages/backend/libs/services/src/disbursement-schedule/disbursement-schedule-shared.service.ts
```diff
@@ -487,10 +487,10 @@ export class DisbursementScheduleSharedService extends RecordDataModelService<Di
    * @param totalAlreadyDisbursedValues sum of the awards already
    * disbursed per award code.
    */
-  private async applyGrantsAlreadyDisbursedValues(
+  private applyGrantsAlreadyDisbursedValues(
```
ann-aot, line 490: 👍

sources/packages/web/src/main.ts
andrewsignori-aot: I see that, but it is not doing anything, right?

sh16011993: Yes actually. I can log it or I can remove it but if I remove it, then Sonar complains

sh16011993: Removed it as per the discussion

sources/packages/backend/apps/db-migrations/src/main.ts
ann-aot: ya, we can go with option 1

sources/packages/backend/apps/queue-consumers/src/main.ts
sh16011993: Initially, bootstrap was called separately after its definition. Its return type was Promise<void> and therefore Sonar was complaining that the promise had to be awaited.

andrewsignori-aot (approved):

ann-aot (approved):
👍 Good work @sh16011993

andrepestana-aot (approved):
👍

Issue #1929, opened 2023-05-08
Current days of study/studybreaks count is 1 short of total.


JasonCTang:
**Describe the Bug**
The system currently does not count the start and end date inclusively when counting the # of days of study, as well as the # of days for a study break.  

**Expected Behaviour**
The current calculation is [study end date] - [study start date].

**Actual Behaviour**
It should be [study end date] - [study start date] + 1.

** Steps To Reproduce**
Steps to reproduce the behaviour:

PR #1975, opened 2023-05-29
#1929 - Current days of study/studybreaks count is 1 short of total

andrepestana-aot:
- Added 1 day to the calculation of `dateDifference` as all date periods will be calculated with start and end dates inclusive;
- Changed education program offering form to add 1 day in the calculated value for study break days;
- Added unit tests to check the calculation results for `getCalculatedStudyBreaksAndWeeks` method:
    √ Should calculate funded study period and total days when study start and end study dates are available. (4 ms)                                                                      
    √ Should not calculate funded study period days and total days when study start date is not available.                                                                                
    √ Should not calculate funded study period days and total days when study end date is not available. (1 ms)                                                                           
    √ Should calculate study breaks when start and end study break dates available. (1 ms)                                                                                                
    √ Should not calculate study breaks when start study break date is not available.                                                                                                     
    √ Should not calculate study breaks when end study break date is not available.                                                                                                       
    √ Should calculate eligible break days as a maximum of 21 days when a study break exceeds it.

### DIFF
--- a/sources/packages/backend/apps/api/src/services/education-program-offering/_tests_/unit/education-program-offering.service-getCalculatedStudyBreaksAndWeeks.spec.ts
+++ b/sources/packages/backend/apps/api/src/services/education-program-offering/_tests_/unit/education-program-offering.service-getCalculatedStudyBreaksAndWeeks.spec.ts
@@ -18,7 +18,7 @@ describe("EducationProgramOfferingService-getCalculatedStudyBreaksAndWeeks", () 
       );
 
     // Assert
-    expect(calculatedStudyBreaksAndWeeks).toEqual({
+    expect(calculatedStudyBreaksAndWeeks).toStrictEqual({
       allowableStudyBreaksDaysAmount: 1,
       fundedStudyPeriodDays: 10,
       studyBreaks: [],
@@ -46,7 +46,7 @@ describe("EducationProgramOfferingService-getCalculatedStudyBreaksAndWeeks", () 
       );
 
     // Assert
-    expect(calculatedStudyBreaksAndWeeks).toEqual({
+    expect(calculatedStudyBreaksAndWeeks).toStrictEqual({
       allowableStudyBreaksDaysAmount: NaN,
       fundedStudyPeriodDays: NaN,
       studyBreaks: [],
@@ -74,7 +74,7 @@ describe("EducationProgramOfferingService-getCalculatedStudyBreaksAndWeeks", () 
       );
 
     // Assert
-    expect(calculatedStudyBreaksAndWeeks).toEqual({
+    expect(calculatedStudyBreaksAndWeeks).toStrictEqual({
       allowableStudyBreaksDaysAmount: NaN,
       fundedStudyPeriodDays: NaN,
       studyBreaks: [],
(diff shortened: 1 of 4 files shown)


sources/packages/backend/libs/utilities/src/date-utils.ts
```diff
@@ -53,13 +53,16 @@ export const isAfter = (
  * Difference in days between endDate and startDate (endDate-startDate).
  * @param endDate end date.
  * @param startDate start date.
+ * @param inclusiveDiff whether it should consider start and end date as part of the calculation.
```
sh16011993, line 56: For my understanding, in which case might this parameter be set to false?

andrepestana-aot: As discussed in our meeting, all the dates will be inclusive, then I removed the param. Thanks for the question.

sources/packages/backend/apps/api/src/services/education-program-offering/_tests_/unit/education-program-offering.service-getCalculatedStudyBreaksAndWeeks.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+      };
+
+    // Act
+    const calculatedStudyBreaksAndWeeks =
+      EducationProgramOfferingService.getCalculatedStudyBreaksAndWeeks(
+        offeringStudyBreakCalculationContext,
+      );
+
+    // Assert
+    expect(calculatedStudyBreaksAndWeeks).toEqual({
```
andrewsignori-aot, line 228: Is there a benefit of using `toStrictEqual` instead `toEqual` for these cases?

sources/packages/backend/apps/api/src/services/education-program-offering/_tests_/unit/education-program-offering.service-getCalculatedStudyBreaksAndWeeks.spec.ts
andrewsignori-aot, line 1: Great to see some tests added to this ❤️

andrewsignori-aot (approved):
Great work and thanks for adding the unit tests. Looks good 👍

sources/packages/backend/apps/api/src/services/education-program-offering/_tests_/unit/education-program-offering.service-getCalculatedStudyBreaksAndWeeks.spec.ts
andrepestana-aot: In this case we can use `.toStrictEqual` because the service is creating an object literal. 

Benefits of using  `.toEqual`:

- keys with undefined properties are checked, e.g. {a: undefined, b: 2} will not equal {b: 2};
- undefined items are taken into account, e.g. [2] will not equal [2, undefined];
- array sparseness is checked, e.g. [, 1] will not equal [undefined, 1];
- object types are checked, e.g. a class instance with fields a and b will not equal a literal object with fields a and b.

I changed the comparions to `.toStrictEqual`. Thanks!

guru-aot (approved):
Nice work @andrepestana-aot , thanks for adding the test cases 👍

Issue #1959, opened 2023-05-24
Stage SIMS PROD environment


JasonCTang:
**Describe the task**
Setup activities not relating to the actual deployment of the application.

**Acceptance Criteria**
- [x] Create the secret variables
- [x] Allocate 64GB divided by 3 for Patroni
- [x] Patroni
- [x] Forms/MongoDB
- [x] Redis

- [x] Security-init.yml (Check if this is needed)
- [x] Networkpolicy.yml
- [x] Update wiki/front page of repo https://github.com/bcgov/SIMS/blob/main/docs/howto/devops.md#openshift-setup

**Additional context**
- Jason to populate secret values

PR #1987, opened 2023-06-02
#1959 - Staging sims prod environment -Part 1

guru-aot:
Changes as part of this PR:
- Patroni deployment is changed, going forward we are not building in tools namespace and deploying them. We use the reference as per sample shared by BCGOV https://github.com/bcgov/patroni-postgres-container/blob/master/samples/deploy.yaml for further deployments.
- Network Policies removed which are not used.
- Role Bindings  for github actions will be added as part of init-oc command.
- Removed the secrets that are not used and removed the obselete methods
  - WORKFLOW_SA_USER_NAME
  - WORKFLOW_SA_PASSWORD
  - SIMS_API_CLIENT_ID
  - SIMS_API_CLIENT_SECRET


sources/packages/backend/apps/api/src/services/auth/keycloak/keycloak.service.ts
```diff
@@ -99,24 +99,6 @@ export class KeycloakService {
     return this.getKeyCloakToken(payload);
   }
 
-  /**
-   * Retrieve an token from Keycloak using a provided client id and client secret.
-   * @param clientSecret Client secret.
-   * @param clientId Client ID.
-   * @returns token
-   */
-  public async getTokenFromClientSecret(
```
ann-aot, line 108: why was this removed

ann-aot (approved):
👍 good work @guru-aot

sources/packages/backend/apps/api/src/services/auth/keycloak/keycloak.service.ts
guru-aot: Yes it was obsolete and do not have any reference

andrewsignori-aot: @ann-aot it was not for form.io but it was no longer used.

devops/openshift/database/patroni-deploy.yml
```diff
@@ -186,16 +215,21 @@ parameters:
...
+- name: IMAGE_NAME
+  description: |
+    The Patroni image stream name
+  value: patroni-postgres
+- name: IMAGE_TAG
   description: |
-    Patroni ImageTag
-  value: patroni:12-latest
+    The image tag used to specify which image you would like deployed.
+    Don't use `latest`.
```
andrewsignori-aot, line 231: The comment is confusing 😄 
"Don't use `latest`" and then we use "12.4-latest".

guru-aot: :) Copy paste from the repo sample i got from bcgov. I will remove that line, might be an old comment i guess.

sources/packages/backend/apps/api/src/services/auth/tokens.service.ts
andrewsignori-aot, line 1: Thanks for doing the cleanup 😉

andrewsignori-aot (approved):
Thanks for doing the cleanups and updating Patroni. Looks good 👍

dheepak-aot (approved):
Great work with patroni and thanks for the clean up. 👍

Issue #1953, opened 2023-05-19
Request an Offering Change - DB Migrations and List


andrewsignori-aot:
As an institution, I need to allow the student to request an offering change for a completed application, which will then trigger a new assessment upon student and Ministry approval.

**Additional Context**
https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?type=design&node-id=11898-153649&t=8iHQAtz9roK7z0ti-0

- [x] A "Request a Change" have the below statuses:
    - In progress with student (once the institution creates the request)
    - In progress with SABC (after student approval)
    - Approved (Ministry approval)
    - Declined by student
    - Declined by SABC
- [x] Create a new menu for locations "Request a Change" to list
    - "Available to change" applications: display all complete applications that are not archived yet and do not have a "Request a Change" in progress (`In Progress with student` or `In Progress with SABC`)
    - "In progress" requests: display requests of type `In progress with student` and `In progress with SABC`.
    - "Completed" requests: display all requests with statuses `Approved`, `Declined by student`, and `Declined by SABC`, including the archived applications to allow the search.
- [x] Server pagination.
- [x] Search for the student's full name and application number.
- [x] Sort by the student's full name and application number.
- [x] View button can be present but will not redirect the user.
- [x] Order by application number.
- [x] New tables to store the requests, and assessment changes (new trigger type and new column for requests relationship). See analysis on #1900.
- [x] E2E tests for new API endpoints.

PR #1995, opened 2023-06-07
#1953 - Request an Offering Change - DB Migrations 

ann-aot:
- Created `application_offering_change_requests` table and `application_offering_change_request_status_types` enum type

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/migrations/1686094364995-CreateApplicationOfferingRequestsAndStatusTypes.ts
+++ b/sources/packages/backend/apps/db-migrations/src/migrations/1686094364995-CreateApplicationOfferingRequestsAndStatusTypes.ts
@@ -22,7 +22,7 @@ export class CreateApplicationOfferingRequestsAndStatusTypes1686094364995
   public async down(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
       getSQLFileData(
-        "Drop-application-offering_change-requests.sql",
+        "Drop-application-offering-change-requests.sql",
         "ApplicationOfferingChangeRequests",
       ),
     );
--- a/sources/packages/backend/apps/db-migrations/src/migrations/1686094364995-CreateApplicationOfferingRequestsAndStatusTypes.ts
+++ b/sources/packages/backend/apps/db-migrations/src/migrations/1686094364995-CreateApplicationOfferingRequestsAndStatusTypes.ts
@@ -1,0 +1,36 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class CreateApplicationOfferingRequestsAndStatusTypes1686094364995
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData(
+        "Create-application-offering-change-request-status-types.sql",
+        "Types",
+      ),
+    );
+    await queryRunner.query(
+      getSQLFileData(
+        "Create-application-offering-change-requests.sql",
+        "ApplicationOfferingChangeRequests",
+      ),
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData(
+        "Drop-application-offering_change-requests.sql",
+        "ApplicationOfferingChangeRequests",
+      ),
+    );
+    await queryRunner.query(
+      getSQLFileData(
+        "Drop-application-offering-change-request-status-types.sql",
+        "Types",
+      ),
+    );
+  }
+}
--- a/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
@@ -8,16 +8,12 @@ CREATE TABLE sims.application_offering_change_requests (
   assessed_by INT REFERENCES sims.users(id),
   student_action_date TIMESTAMP WITH TIME ZONE,
   reason VARCHAR(500) NOT NULL,
-  ministry_note_id INT REFERENCES sims.notes(id),
+  assessed_note_id INT REFERENCES sims.notes(id),
   -- Audit columns
   created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
   updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
-  creator INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
-  SET
-    NULL,
-    modifier INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
-  SET
-    NULL
+  creator INT NULL DEFAULT NULL REFERENCES sims.users(id),
+  modifier INT NULL DEFAULT NULL REFERENCES sims.users(id)
 );
 
 -- ## Comments
@@ -27,7 +23,7 @@ COMMENT ON COLUMN sims.application_offering_change_requests.id IS 'Auto-generate
 
 COMMENT ON COLUMN sims.application_offering_change_requests.application_id IS 'Foreign key reference to the applications table. Application on which the application specific offering change was requested.';
 
-COMMENT ON COLUMN sims.application_offering_change_requests.change_offering_id IS 'Foreign key reference to the education programs offerings table. The new offering created by the institution for the application specific offering change.';
+COMMENT ON COLUMN sims.application_offering_change_requests.change_offering_id IS 'Foreign key reference to the education programs offerings table. The new or existing offering assigned by the institution for the application specific offering change.';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.active_offering_id IS 'Foreign key reference to the education programs offerings table. The actual offering of the application that was requested for the application specific offering change.';
 
@@ -41,7 +37,7 @@ COMMENT ON COLUMN sims.application_offering_change_requests.student_action_date 
 
 COMMENT ON COLUMN sims.application_offering_change_requests.reason IS 'The reason for application specific offering request change added by institution.';
 
-COMMENT ON COLUMN sims.application_offering_change_requests.ministry_note_id IS 'Note added by the Ministry while approving or denying the application specific offering request change.';
+COMMENT ON COLUMN sims.application_offering_change_requests.assessed_note_id IS 'Note added by the Ministry while approving or denying the application specific offering request change.';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.created_at IS 'Record creation timestamp.';
 
--- a/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
@@ -7,7 +7,7 @@ CREATE TABLE IF NOT EXISTS sims.application_offering_change_requests (
   assessed_date TIMESTAMP WITH TIME ZONE,
   assessed_by INT REFERENCES sims.users(id),
   student_action_date TIMESTAMP WITH TIME ZONE,
-  reason VARCHAR(500),
+  reason VARCHAR(500) NOT NULL,
   ministry_note_id INT REFERENCES sims.notes(id),
   -- Audit columns
   created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
@@ -25,13 +25,13 @@ COMMENT ON TABLE sims.application_offering_change_requests IS 'Represents the li
 
 COMMENT ON COLUMN sims.application_offering_change_requests.id IS 'Auto-generated sequential primary key column.';
 
-COMMENT ON COLUMN sims.application_offering_change_requests.application_id IS 'Foreign key reference to the applications table. Application on which the eligible offering change was requested.';
+COMMENT ON COLUMN sims.application_offering_change_requests.application_id IS 'Foreign key reference to the applications table. Application on which the offering change was requested.';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.change_offering_id IS 'Foreign key reference to the education programs offerings table. The new offering created by the institution for the offering change.';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.active_offering_id IS 'Foreign key reference to the education programs offerings table. The actual offering of the application that was requested for the change.';
 
-COMMENT ON COLUMN sims.application_offering_change_requests.application_offering_change_requests_status IS 'Current status of application offering request change (e.g. In progress with the student, In progress with SABC, Approved, Declined by student, Declined by SABC.';
+COMMENT ON COLUMN sims.application_offering_change_requests.application_offering_change_requests_status IS 'Current status of application offering request change (e.g. In progress with the student, In progress with SABC, Approved, Declined by student, Declined by SABC).';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.assessed_date IS 'Date that the Ministry approved or denied the application offering request change.';
 
@@ -40,8 +40,6 @@ COMMENT ON COLUMN sims.application_offering_change_requests.assessed_by IS 'Mini
 COMMENT ON COLUMN sims.application_offering_change_requests.student_action_date IS 'Date that the Student approved or denied the application offering request change.';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.reason IS 'The reason for application offering request change added by institution.';
-
-COMMENT ON COLUMN sims.application_offering_change_requests.ministry_note_id IS 'Note added by the Ministry while approving or denying the application offering request change.';
 
 COMMENT ON COLUMN sims.application_offering_change_requests.ministry_note_id IS 'Note added by the Ministry while approving or denying the application offering request change.';
 
--- a/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
@@ -1,0 +1,54 @@
+CREATE TABLE IF NOT EXISTS sims.application_offering_change_requests (
+  id SERIAL PRIMARY KEY,
+  application_id INT NOT NULL REFERENCES sims.applications(id),
+  change_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  active_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  application_offering_change_requests_status sims.application_offering_change_request_status_types NOT NULL,
+  assessed_date TIMESTAMP WITH TIME ZONE,
+  assessed_by INT REFERENCES sims.users(id),
+  student_action_date TIMESTAMP WITH TIME ZONE,
+  reason VARCHAR(500),
+  ministry_note_id INT REFERENCES sims.notes(id),
+  -- Audit columns
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+  creator INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
+  SET
+    NULL,
+    modifier INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
+  SET
+    NULL
+);
+
+-- ## Comments
+COMMENT ON TABLE sims.application_offering_change_requests IS 'Represents the list of application (completed and not archived) specific offering change request, which is requested by institution.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.id IS 'Auto-generated sequential primary key column.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.application_id IS 'Foreign key reference to the applications table. Application on which the eligible offering change was requested.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.change_offering_id IS 'Foreign key reference to the education programs offerings table. The new offering created by the institution for the offering change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.active_offering_id IS 'Foreign key reference to the education programs offerings table. The actual offering of the application that was requested for the change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.application_offering_change_requests_status IS 'Current status of application offering request change (e.g. In progress with the student, In progress with SABC, Approved, Declined by student, Declined by SABC.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.assessed_date IS 'Date that the Ministry approved or denied the application offering request change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.assessed_by IS 'Ministry user that approved or denied the application offering request change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.student_action_date IS 'Date that the Student approved or denied the application offering request change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.reason IS 'The reason for application offering request change added by institution.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.ministry_note_id IS 'Note added by the Ministry while approving or denying the application offering request change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.ministry_note_id IS 'Note added by the Ministry while approving or denying the application offering request change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.created_at IS 'Record creation timestamp.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.updated_at IS 'Record update timestamp.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.creator IS 'Creator of the record.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.modifier IS 'Modifier of the record.';
--- a/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
@@ -1,7 +1,7 @@
 CREATE TABLE sims.application_offering_change_requests (
   id SERIAL PRIMARY KEY,
   application_id INT NOT NULL REFERENCES sims.applications(id),
-  change_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  requested_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
   active_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
   application_offering_change_request_status sims.application_offering_change_request_status_types NOT NULL,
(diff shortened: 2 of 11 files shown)


sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,52 @@
+CREATE TABLE IF NOT EXISTS sims.application_offering_change_requests (
```
andrepestana-aot, line 1: As I understood from a previous PR, we won't add `IF EXISTS`/`IF NOT EXISTS` anymore.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Drop-application-offering_change-requests.sql
```diff
@@ -0,0 +1,2 @@
+-- DROP Table
+DROP TABLE IF EXISTS sims.application_offering_change_requests;
```
andrepestana-aot, line 2: https://github.com/bcgov/SIMS/pull/1995/files#r1222068667

sources/packages/backend/apps/db-migrations/src/migrations/1686094364995-CreateApplicationOfferingRequestsAndStatusTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+        "Create-application-offering-change-requests.sql",
+        "ApplicationOfferingChangeRequests",
+      ),
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData(
+        "Drop-application-offering_change-requests.sql",
```
andrewsignori-aot, line 25: Please remove the underscore from the file name.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
ann-aot: I had it in my mind, but then I was confused about whether our discussion was not to use `IF EXISTS/IF NOT EXISTS` for ALTER tables or in all `.sql` comments, I actually tried to find the old chat in dev, and I couldn't find it. 
so I thought, I will wait for the dev comment. 
 Thanks for the comments. I will address it

sources/packages/backend/apps/db-migrations/src/sql/Types/Create-application-offering-change-request-status-types.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TYPE sims.application_offering_change_request_status_types AS ENUM (
+  'In progress with the student',
+  'In progress with SABC',
+  'Approved',
```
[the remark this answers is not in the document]
andrepestana-aot: ok, got it from the comment.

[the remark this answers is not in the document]
ann-aot: yes

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-requests.model.ts
```diff
@@ -0,0 +1,162 @@
...
+  @PrimaryColumn({
+    name: "id",
+    nullable: false,
+  })
+  id: number;
+
+  /**
+   * Student Application id related to application specific offering change request.
+   */
+  @RelationId(
```
andrepestana-aot, line 38: I see that adding `@RelationId` is to get the object's id without doing the join but what's the advantage of this instead of just adding `@Column`? @sh16011993 and I were trying to do a query using `findOne()` selecting a property decorated with `@RelationId` and it wasn't possible. We did some tests and checked selecting is just for columns. Then we changed it to `@Column` as test and everything seemed to work.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,52 @@
+CREATE TABLE sims.application_offering_change_requests (
+  id SERIAL PRIMARY KEY,
+  application_id INT NOT NULL REFERENCES sims.applications(id),
+  change_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  active_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  application_offering_change_request_status sims.application_offering_change_request_status_types NOT NULL,
+  assessed_date TIMESTAMP WITH TIME ZONE,
+  assessed_by INT REFERENCES sims.users(id),
+  student_action_date TIMESTAMP WITH TIME ZONE,
```
andrewsignori-aot, line 9: I would prefer to be precise about the "action".
I would suggest something like `student_approval_date` or `student_acceptance_date`.

andrepestana-aot (commented):
Great job. Left minor comments and one for my knowledge. Thanks!

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,52 @@
...
+  reason VARCHAR(500) NOT NULL,
+  ministry_note_id INT REFERENCES sims.notes(id),
```
andrewsignori-aot, line 11: I believe that we never used `ministry_note_id` and we try to create something more close to the context. For this one I would suggest `assessed_note_id`.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,52 @@
+CREATE TABLE sims.application_offering_change_requests (
+  id SERIAL PRIMARY KEY,
+  application_id INT NOT NULL REFERENCES sims.applications(id),
+  change_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  active_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
```
andrewsignori-aot, line 5: Should we keep the `active_offering_id` in the table or just use the one from the current assessment?

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
ann-aot: my idea was it can have either the approved date or the declined date

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,52 @@
...
+  application_offering_change_request_status sims.application_offering_change_request_status_types NOT NULL,
+  assessed_date TIMESTAMP WITH TIME ZONE,
+  assessed_by INT REFERENCES sims.users(id),
+  student_action_date TIMESTAMP WITH TIME ZONE,
+  reason VARCHAR(500) NOT NULL,
+  ministry_note_id INT REFERENCES sims.notes(id),
+  -- Audit columns
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+  creator INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
```
andrewsignori-aot, line 15: I believe that we stooped using the `ON DELETE` also for these ones, right?
I found some examples in the source code no longer using them.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
ann-aot: actually, when I checked another table, we were either using `note_id` or `entity_note_id`, eg, `restrction_note_id`. so I thought if I use `application-offering-change-requests_note_id` if it confuses dev with the reason column. Thats why I gave this name, but the name you suggested also makes sense. I will update

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,52 @@
...
+);
+
+-- ## Comments
+COMMENT ON TABLE sims.application_offering_change_requests IS 'Represents the list of application specific offering change requests, which is requested by institution.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.id IS 'Auto-generated sequential primary key column.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.application_id IS 'Foreign key reference to the applications table. Application on which the application specific offering change was requested.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.change_offering_id IS 'Foreign key reference to the education programs offerings table. The new offering created by the institution for the application specific offering change.';
```
andrewsignori-aot, line 30: From a system perspective, it does not need to be a new offering.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
ann-aot: mm, that makes sense. I was following the DB structure from the analysis ticket. @andrepestana-aot @dheepak-aot @guru-aot thoughts?

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
ann-aot: mm, I when was doing this I found it on most of the recent ones, but I will remove it

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
guru-aot: I think we already had a discussion, the thought of having the oactive offering id is when we create a componenet which shows both the offeringside byside, the requirement for it would be pass the 2 offering id's and it should render. Yes we can use the current_assessment, but again do we need to go through the application-> current_assessment->offering_id every time? I am okay with not having the offering id explicitly too, but looking at the row in the table can give the active and changed offering directly. Just my thoughts

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TABLE sims.application_offering_change_requests (
+  id SERIAL PRIMARY KEY,
+  application_id INT NOT NULL REFERENCES sims.applications(id),
+  change_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
```
guru-aot, line 4: I know i created the DB column names :) , but I feel it would be better to have change_offering_id as requested_offering_id? Thoughts? @andrewsignori-aot @dheepak-aot @andrepestana-aot @sh16011993

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
andrewsignori-aot: We do not build the DB and API based on how the UI will work, right?
Yes @guru-aot I remember the conversation and I did not remember the output of it.
My point goes beyond saving the current offering id or not. What happens if between the request and the Ministry approval the application went through another reassessment? The real concern is that we would be copying the offering id from the current assessment into this table only to make it better for the UI.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
andrewsignori-aot: I see, it makes sense 😉

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
guru-aot: I get your point. Does a reassessment happens inbetween? Is that a valid scenario? IF yes, dont we need to stop the reassessment from happening?

sources/packages/backend/libs/sims-db/src/entities/index.ts
```diff
@@ -82,3 +82,5 @@ export * from "./disbursement-overaward.model";
 export * from "./notification-permanent-failure-error.type";
 export * from "./queue_configuration.model";
 export * from "./student-assessment-status.type";
+export * from "./application-offering-change-request-status.type";
+export * from "./application-offering-change-requests.model";
```
andrewsignori-aot, line 86: This follows the entity class name, so it should be singular.

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-requests.model.ts
andrewsignori-aot: Apparently `@RelationId` is causing more questions than benefits. I would be ok to stop using them.

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-requests.model.ts
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Represents the list of application specific offering change request,
+ * which is requested by institution.
+ */
+@Entity({ name: TableNames.ApplicationOfferingChangeRequests })
+export class ApplicationOfferingChangeRequest extends RecordDataModel {
+  /**
+   * Auto-generated sequential primary key column.
+   */
+
```
andrewsignori-aot, line 28: Please remove the blank line.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
andrewsignori-aot: The idea is that using IF EXISTS/IF NOT EXISTS can hide some issues in the migrations. The migrations should enforce whether the item exists or not because the DB is supposed to be changed only through a migration.
If a migration is executed to remove an item and the item is not present, either the migration is wrong or the DB it is in a state where the migration should not be executed.

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-requests.model.ts
```diff
@@ -0,0 +1,162 @@
...
+   * Current status of application specific offering request change (e.g. In progress with the student,
+   * In progress with SABC, Approved, Declined by student, Declined by SABC).
+   */
+  @Column({
+    name: "application_offering_change_request_status",
+    type: "enum",
+    enum: ApplicationOfferingChangeRequestStatus,
+    enumName: "ApplicationOfferingChangeRequestStatus",
+  })
+  applicationOfferingChangeRequestsStatus: ApplicationOfferingChangeRequestStatus;
```
andrewsignori-aot, line 111: Can we rename it to `applicationOfferingChangeRequetStatus` or `changeRequetStatus`?

andrewsignori-aot (commented):
Great start, please take a look at the comments.
The other migrations for student assessments will be done as part of the other ticket?

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-requests.model.ts
ann-aot: Removed

ann-aot:
> Great start, please take a look at the comments. The other migrations for student assessments will be done as part of the other ticket?

yes @andrewsignori-aot

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
dheepak-aot: @andrewsignori-aot  @guru I read through the conversations, and I strongly feel that we need to we need to have `active_offering_id`  here for the sake of accuracy of the application_offering_change_record.

Assume the following scenario without `active_offering_id`

An application with applicationId 100 and offeringId 8

-> applicationId - 100, requested_offering_id-10, status-approved

The application might go through a re-assessment and get associated to an offering 12 later.

If that is the case, then this application offering change record data will say that the offerring change happened from offering 12 to offering 10 which is not accurate. I am talking from the data perspective.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TABLE sims.application_offering_change_requests (
+  id SERIAL PRIMARY KEY,
+  application_id INT NOT NULL REFERENCES sims.applications(id),
+  requested_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  active_offering_id INT NOT NULL REFERENCES sims.education_programs_offerings(id),
+  application_offering_change_request_status sims.application_offering_change_request_status_types NOT NULL,
+  assessed_date TIMESTAMP WITH TIME ZONE,
```
dheepak-aot, line 7: For the sake of clarification? will assessed_by and assessed_date be updated only upon a ministry update? because I feel like that is more a ministry assessment. 

If suppose a student approves or decline the offering change request, it is more or less like a consent right?

@andrewsignori-aot  @andrepestana-aot @guru-aot ?

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,48 @@
...
+  creator INT NULL DEFAULT NULL REFERENCES sims.users(id),
+  modifier INT NULL DEFAULT NULL REFERENCES sims.users(id)
+);
+
+-- ## Comments
+COMMENT ON TABLE sims.application_offering_change_requests IS 'Represents the list of application specific offering change requests, which is requested by institution.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.id IS 'Auto-generated sequential primary key column.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.application_id IS 'Foreign key reference to the applications table. Application on which the application specific offering change was requested.';
```
dheepak-aot, line 24: May be, we can avoid technical terms like Foreign key in the comments as it is implicit from code.

ann-aot: we are using it in other places, i was trying to follow this same,
image

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Drop-application-offering-change-requests.sql
```diff
@@ -0,0 +1,2 @@
+-- DROP Table
+DROP TABLE sims.application_offering_change_requests;
```
dheepak-aot, line 2: Are we starting to avoid `IF EXISTS` and `IF NOT EXISTS` ? I remember that dev chat we had.

ann-aot: yes we are

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
```diff
@@ -0,0 +1,48 @@
...
+
+COMMENT ON COLUMN sims.application_offering_change_requests.requested_offering_id IS 'Foreign key reference to the education programs offerings table. The new or existing offering assigned by the institution for the application specific offering change.';
+
+COMMENT ON COLUMN sims.application_offering_change_requests.active_offering_id IS 'Foreign key reference to the education programs offerings table. The actual offering of the application that was requested for the application specific offering change.';
```
dheepak-aot, line 28: `The active offering of the application when application specific offering change was requested.` ?

ann-aot: this is what we discussed just now @dheepak-aot.

sources/packages/backend/apps/db-migrations/src/sql/Types/Create-application-offering-change-request-status-types.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TYPE sims.application_offering_change_request_status_types AS ENUM (
+  'In progress with the student',
```
dheepak-aot, line 2: As per the second AC in the ticket `the` is not present here. But first AC has.

@hellolynn-tbtb ? which one should we go for?

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request-status.type.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Different status of application specific offering change request.
+ */
+export enum ApplicationOfferingChangeRequestStatus {
+  /**
+   * The status is set when the institution creates a
+   * application specific offering change request.
+   */
+  InProgressWithTheStudent = "In progress with the student",
```
dheepak-aot, line 9: https://github.com/bcgov/SIMS/pull/1995/files#r1222277272

sources/packages/backend/apps/db-migrations/src/sql/Types/Create-application-offering-change-request-status-types.sql
ann-aot: this is what I had in my story, https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/1953. Open for suggestion @hellolynn-tbtb

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request.model.ts
```diff
@@ -0,0 +1,136 @@
...
+} from "typeorm";
+import { ColumnNames, TableNames } from "../constant";
+import { RecordDataModel } from "./record.model";
+import { Application } from "./application.model";
+import { EducationProgramOffering } from "./education-program-offering.model";
+import { User } from "./user.model";
+import { Note } from "./note.model";
+import { ApplicationOfferingChangeRequestStatus } from "./application-offering-change-request-status.type";
+
+const REASON_MAX_LENGTH = 500;
```
dheepak-aot, line 17: This max length could be a const like NOTES_MAX_LENGTH which could be used on POST API for class validators.

ann-aot: I didn't get the point here, i believe it should done in the submit story

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request.model.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Represents the list of application specific offering change request,
+ * which is requested by institution.
+ */
+@Entity({ name: TableNames.ApplicationOfferingChangeRequests })
+export class ApplicationOfferingChangeRequest extends RecordDataModel {
+  /**
+   * Auto-generated sequential primary key column.
+   */
+  @PrimaryColumn({
```
dheepak-aot, line 27: We can go with @PrimaryGeneratedColumn() which is more accurate.

Our application has a mix of both @PrimaryGeneratedColumn() and @PrimaryColumn().

I see that all the usages @PrimaryColumn() are either very old code or the primary key is not auto-generated.

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request.model.ts
dheepak-aot: I meant this constant must be more global as we see that it has more potential usage. like this.

image

dheepak-aot:
Great start @ann-aot . Few minor comments.

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request.model.ts
ann-aot: image
I believe we can add export when we are actually using it

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request.model.ts
ann-aot: ok..I was not aware of that. updated

andrepestana-aot (approved):
Looks good. Thanks for the explanations.

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
ann-aot: keeping as it is and as discussed will confirm with @JasonCTang and will update in next PR

sources/packages/backend/libs/sims-db/src/entities/application-offering-change-request.model.ts
dheepak-aot: Yes just the export is missing 😊

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
dheepak-aot: If we have it already, may be we must stop using it IMO if other devs are onboard with the same idea. 

Btw. it is not a blocker.

dheepak-aot (approved):
LGTM

sources/packages/backend/apps/db-migrations/src/sql/Types/Create-application-offering-change-request-status-types.sql
andrewsignori-aot: @ann-aot please use the Figma one and adjust the ticket.
image

ann-aot: as @hellolynn-tbtb  confirmed going without `the`

sources/packages/backend/apps/db-migrations/src/sql/ApplicationOfferingChangeRequests/Create-application-offering-change-requests.sql
sh16011993: I am ok with not using the foreign key term explicitly.

sh16011993 (approved):
LGTM 👍

andrewsignori-aot (approved):
Thanks for doing the changes and lets ensure that we are going to have a follow up on the active offering column 👍

guru-aot (approved):
LGTM

PR #2008, opened 2023-06-12
#1959-Stage SIMS Prod

guru-aot:
Redis cluster command changed, as part of makefile to run in any terminal(Windows/ mac/ Linux). 
 - Issue was when trying to get the jsonpath `oc get pods -n $(NAMESPACE) -l app=redis -o jsonpath='{range .items[*]}{.status.podIP}:6379 ` in mac it appends an additional ` :6379 `at the end, so to solve it removing the generated one using the sed command.
 - Also running the oc commands inside the redis-cli was not working, so created variables to fetch the commands output and ran the cluster creation.
 
Documentation and wiki documentation are updated


README.md
```diff
@@ -23,7 +23,7 @@ High, new features released weekly
 - [Technologies Overview](./docs/technologies-overview.md)
 - [BCeID-Web Service Integration](./docs/bceid-webservice-integration.md)
 - How To
-  - [DevOps](./docs/howto/devops.md)
+  - [DevOps](https://github.com/bcgov/SIMS/wiki/DevOps-and-Running-the-Application#devops)
   - [Database Migration](./docs/howto/db-migration.md)
   - [Keycloak Configuration](./docs/hotwot/keycloak-config.md)
```
andrewsignori-aot, line 28: Can we change the Keycloak URL also to point to https://github.com/bcgov/SIMS/wiki/Keycloak-Config, please?

devops/Makefile
```diff
@@ -286,7 +286,9 @@ deploy-redis:
 # Command to initialize the redis cluster.
 # Make sure that all the redis pods are up and running before initializing the cluster.
 init-redis-cluster:
-	@oc exec -n $(NAMESPACE) -it redis-0 -- redis-cli -a $(oc get secret -n $(NAMESPACE) redis-creds -o jsonpath='{.data.password}' | base64 -d) --cluster create --cluster-replicas 1 $(oc get pods -n $(NAMESPACE) -l app=redis -o jsonpath='{range.items[*]}{.status.podIP}:6379 ')
+	REDIS_PASSWORD=$$(oc get secret -n $(NAMESPACE) redis-creds -o jsonpath='{.data.password}' | base64 -d); \
+	REDIS_NODES=$$(oc get pods -n $(NAMESPACE) -l app=redis -o jsonpath='{range .items[*]}{.status.podIP}:6379 ' | sed 's/ :6379 $$/ /'); \
```
andrewsignori-aot, line 290: Executing the command 
`oc get pods -n 0c27fb-dev -l app=redis -o jsonpath='{range .items[*]}{.status.podIP}:6379 ' | sed 's/ :6379 $$/ /'`
Is still returning the extra `:6379` at the end, as shown below. Is it the expectation?
`10.97.50.200:6379 10.97.38.75:6379 10.97.70.236:6379 10.97.94.99:6379 10.97.56.98:6379 10.97.101.99:6379 :6379`

guru-aot: The idea is when the command is run via a makefile it trims the last 6379
image

image

Dont ask me how it does, but it does 😄

andrewsignori-aot: I see that `echo` is correctly printing it. Thanks for the explanation 😉

andrewsignori-aot (approved):
Nice fix, looks good 👍

dheepak-aot (approved):
Thanks for the fix. As discussed we can have a ticket created for platform agnostic script which works on all types of terminal.

ann-aot (approved):
👍

PR #1998, opened 2023-06-08
#1911 - ECE Response file notification part 1

dheepak-aot:
## ECE Response notification part 1

 - [x] DB Migration to add integration_contacts column and notification message type `ECE Response file processing`
 - [x] Added validation to capture log the invalid data for 1.enrolement confirmation flag 2. enrolment confirmation date and 3.pay to school amount(tuition remittance).
 - [x] Added E2E tests for the new validations. 
 - [x] Modified the processing summary result to always populate the total processing counts as it will be an important piece of information in every scenario.
 - [x] Added E2E tests for enrolment confirmation date outside valid approval period as discussed during the demo of #1501 

image

### To be done in the upcoming PR
 - [ ] Email template for ECE Response notification.
 - [ ] Notification action service.
 - [ ] Save notification on ECE response file processing.

Issue #1911, opened 2023-04-27
Send email notification to predefined email addresses when ECE file load has errors


JasonCTang:
**Describe the task**
Send email notification to predefined email addresses for the institution when ECE file load has errors

**Additional context**
- These email addresses do not change often, so we can live with any changes being performed directly in the database, via query.
- There can be more than one email address per institution/location.

**Acceptance Criteria**
- [x] Store email addresses in the sims.institution_locations table in a new array of strings type column called integration_contacts. 
- [x] Create GCNotify template for the email address.
- [x] Extend E2E tests from #1501.
- [x] If there is an error that prevents the file from being processed at all, send email with error message.
- [x] Whenever there is an error in a data record, add to list of errors to send in a email at the end of processing, then skip record and continue processing.
- [x] After processing complete, send email with list of all errors encountered in data records. 
- [x] Investigate possiblity to send error report as email attachment. 
- [x] Content fix in the E2E tests and add an E2E test for validating enrolment confirmation period.

***Valdiation and Email criteria***
Error emails that SFAS currently sends out if there is an error encountered processing the ECE Response file (CONR_008.txt)

Items that need to be validated:
- 1.	Header record is present and record type is 1.
  - a.	“Header record error. Processing aborted.”
- 2.	Trailer record is present and record type is 3.
  - a.	“Trailer record error. Processing aborted.”
- 3.	At least 1 data record is present.
  - a.	“No data records found. Processing aborted.”
- 4.	Number of data records = the number of records expected which is specified by institution in the trailer record.
  - a.	“Number of data records found does not match expected number. Processing aborted.”
- 5.	In Data records, validate the following:
  - a.	Data record type is 2.
    - i.	“Data record type invalid: [Record Type Value].  Record with Application Number: [Application No] & Disbursement ID: [Disbursement ID].”
  - b.	Application number is numeric.
    - i.	“Application number invalid. Record with Application Number: [Application No] & Disbursement ID: [Disbursement ID].”
  - c.	Disbursement ID is numeric.
    - i.	“Disbursement ID invalid. Record with Application Number: [Application No] & Disbursement ID: [Disbursement ID].”
  - d.	Confirmation flag is Y or N
    - i.	“Confirmation flag invalid: [Flag value]. Record with Application Number: [Application No] & Disbursement ID: [Disbursement ID].”
  - e.	Confirmation date is formatted as a date.
[...]

Issue #1501, opened 2022-10-27
Process Confirmation of Enrolment file for institutions


abschwenker:
Additional Context
- File layout:
  -[https://app.zenhub.com/files/319429847/123a8930-13e2-47b1-80f0-b720cff7b7cb/download](https://app.zenhub.com/files/319429847/123a8930-13e2-47b1-80f0-b720cff7b7cb/download) 
- Sample file:
 - https://bcgov.sharepoint.com/:t:/r/teams/03177/Shared%20Documents/General/Analysis/Institution%20File%20Specs/CONR_008.TXT?csf=1&web=1&e=alE9WO
- File is based off all applications that have an eligible COE request waiting
- When SIMS creates ECE file and passes it to SFAS/HARRY, the SFAS COEs will be appended before sending to institutions.  So the response file will contain approvals/denials for both systems meaning we need to skip over any apps that don't match.
- If one award is approved, then assume all awards are approved for that disbursement. 
- Sum all tuition remittance from awards for a single disbursement. 
- Notifications ***WILL NOT*** be included in this ticket.

Acceptance Criteria
- [x] System user to read file and update database.
- [x] For each disbursement, sum all tuition remittance amounts for all awards.
- [x] When processing the file, treat it the same as the manual COE confirmation process and perform the same validations and notifications.
- [x] Move API COE approval process to shared services to be used by schedule
- [x] When COE is declined, there needs to be a reason.  If response file does not include a reason for declines, then we need to have a default reason to populate the field.
- [ ] Search institutions by the institution code (it's a field in the file and also in the file name)
- [x] Search disbursement by the Award_disbursement_id we sent them
- [x] Assume a record can be received twice, update the first time only. 
- [x] If parsing the file fails, stop processing the file. 
- [x] When processing the records after the file has been passed, if a record does not pass validation, skip over the WHOLE DISBURSEMENT and continue.
- [x] In the log, identify which records did not pass validation and report how many applications were not found and report how many applications were duplicates.
- [ ] If a COE is denied, there is no "reason" provided in the CONR_008.txt file, so default to "Please speak to your institution."
- [x] Create E2E test.

Questions
- [ ]  What do we do when any records in the response file does not pass validations? How is this handled in SFAS currently?
  - SFAS stops processing the file if it encounters an error and automatically emails the Institution user to let them know there was an issue with the file. Any records that were not updated, get included again in the next file. 
  - ***Notifications WILL NOT be included in this ticket***
[...]

PR #2011, opened 2023-06-12
#1911 - ECE response processing notification part 2

dheepak-aot:
## ECE Response processing notification part 2

- [x] Create a GC Notify template to send emails to integration contacts for ECE
image
image

Email attachment:
image

- [x] Update ECE Response processing service to save notifications on file processing when the file is present.
- [x] Updated the validations in ECE processing service to check for application number to have a valid number as per @JasonCTang 
- [x] Updated the `createFakeApplication` as mentioned above to create application as valid number string with fixed length of 10 characters.
- [x] Added an interface to GCNotify email payload to have some sort of validation in place.
- [x] Added E2E tests for notifications.
- [x] Added E2E test for disbursement with multiple detail records(multiple award lines) as per discussions in previous PR https://github.com/bcgov/SIMS/pull/1998

image

- [x] Updated the E2E tests(few) to check for COE status and tuition remittance after the processing is complete as per the discussion in previous PR https://github.com/bcgov/SIMS/pull/1998


sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
```diff
@@ -31,24 +39,31 @@ export async function createInstitutionLocations(
...
     "FAIL",
     e2eDataSources,
   );
+  // Institution location to test disbursement with multiple detail records.
+  const institutionLocationMULT = await findOrCreateInstitutionLocation(
+    institution,
+    "MULT",
+    e2eDataSources,
+    // Multiple integration contacts.
+    { integrationContacts: [faker.internet.email(), faker.internet.email()] },
```
ann-aot, line 58: how about moving this logic to `findOrCreateInstitutionLocation`  and the contact type will depend on the flag?

dheepak-aot: I feel that letting the consumer replace the default data that the method `findOrCreateInstitutionLocation ` creates will have more flexibility at consumer side to use it. e.g. what if consumer wants the data to be just [] or an invalid data in future.

ann-aot: it can be achieved with the same flag also right

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
```diff
@@ -95,3 +114,22 @@ export async function enableIntegration(
...
+
+/**
+ * Get all unsent ECE Response notifications.
+ * @param e2eDataSources e2e data sources.
+ * @returns Unsent ECE Response notifications.
+ */
+export async function getUnsentECEResponseNotifications(
+  e2eDataSources: E2EDataSources,
+): Promise<Notification[]> {
+  return await e2eDataSources.notification.find({
```
ann-aot, line 126: (https://github.com/bcgov/SIMS/pull/2011#discussion_r1228494095)

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
```diff
@@ -135,9 +160,130 @@ describe(
...
+        process.env.INSTITUTION_RESPONSE_FOLDER,
+        locationMULT.institutionCode,
+        ECE_RESPONSE_FILE_NAME,
+      );
+
+      // Create disbursement to confirm enrolment.
+      const application = await saveFakeApplicationDisbursements(
+        db.dataSource,
+        {
+          // Save disbursement with custom award value to validate tuition remittance.
```
ann-aot, line 193: its not a custom value right

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
```diff
@@ -135,9 +160,130 @@ describe(
...
+      // Expect the notifications to be created.
+      const notifications = await getUnsentECEResponseNotifications(db);
+      expect(notifications).toHaveLength(
+        locationMULT.integrationContacts.length,
+      );
+      const updatedDisbursement = await db.disbursementSchedule.findOne({
+        select: {
+          coeStatus: true,
+          tuitionRemittanceRequestedAmount: true,
+          tuitionRemittanceEffectiveAmount: true,
```
ann-aot, line 274: I see `tuitionRemittanceEffectiveAmount` is not used anywhere

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
dheepak-aot: Sorry, I am not able to follow this. My assumption on hearing a flag was a boolean value passed to the method and and method creating data1 or data2 based on true or false of the boolean. But what If we may need more that 2 types of data ?

Am I missing something here.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
```diff
@@ -260,9 +419,14 @@ describe(
...
       expect(sftpClientMock.delete).toHaveBeenCalled();
+      // Expect the notifications to be created.
+      const notifications = await getUnsentECEResponseNotifications(db);
+      expect(notifications).toHaveLength(
+        locationSKIP.integrationContacts.length,
+      );
     });
 
-    it("Should skip the ECE disbursement when disbursement does not belong to the system.", async () => {
+    it("Should skip the ECE disbursement and create notification when disbursement does not belong to the system.", async () => {
```
ann-aot, line 429: `when disbursement and application does not belong to the system.`

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
dheepak-aot: Can you please be elaborate on why you consider this award value not being custom?

In my view it was replacing the default value of award ($1) provided by the factory to specifically test one more scenarios in my suite and hence I called it custom.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
```diff
@@ -359,6 +532,7 @@ describe(
       expectedResult.summary = [
         `Starting download of file ${confirmEnrolmentResponseFile}.`,
         `The file ${confirmEnrolmentResponseFile} has been deleted after processing.`,
+        "Notification has been created to send email to integration contacts.",
```
ann-aot, line 535: I am not seeing the notification check here, is that the intention here? then why description was changed? is it for this line

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
dheepak-aot: Right, removing it.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
dheepak-aot: Notification check is missed. will add it.

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-files/ece-response-file-detail.ts
```diff
@@ -73,7 +73,7 @@ export class ECEResponseFileDetail extends ECEResponseFileRecord {
     if (isNaN(this.disbursementIdentifier)) {
       errors.push("Invalid unique index number for the disbursement record");
     }
-    if (!this.applicationNumber?.trim()) {
+    if (!this.applicationNumber?.trim() || isNaN(+this.applicationNumber)) {
```
ann-aot, line 76: do we need `+` when using `isNaN` `?

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
```diff
@@ -88,8 +103,10 @@ export class ECEResponseProcessingService {
    * @returns process summary result.
    */
   private async processDisbursementsInECEResponseFile(
-    remoteFilePath: string,
+    institutionFileDetail: InstitutionFileDetail,
```
ann-aot, line 106: pls update the comment

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
```diff
@@ -390,6 +417,81 @@ export class ECEResponseProcessingService {
     }
   }
 
+  /**
+   * Create ECE Response file processing notifications.
+   * @param institutionCode institution of the processing file.
+   * @param auditUserId user who creates notification.
```
ann-aot, line 423: pls review the comment

ann-aot (commented):
Added some comments @dheepak-aot. Thanks for adding the scenarios that we discussed in the test. I think we can also add the new message_payload type to the `SaveNotificationModel` interface

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-files/ece-response-file-detail.ts
ann-aot: if it only excepts then what is the use of not a number check, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN, 
image

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
ann-aot: My thought was a flag (  probably a string or enum like 'MULT', 'SINGLE', or even a boolean) which decides the number of contact on `findOrCreateInstitutionLocation`, and if the option is not present, then contacts can be null

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
ann-aot: IMO, I still don't see it as a custom.  I see it as an actual award.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
```diff
@@ -72,11 +90,12 @@ async function findOrCreateInstitutionLocation(
   if (!institutionLocation) {
     const newInstitutionLocation = createFakeInstitutionLocation(institution);
     newInstitutionLocation.institutionCode = institutionCode;
-    institutionLocation = await e2eDataSources.institutionLocation.save(
-      newInstitutionLocation,
-    );
+    institutionLocation = newInstitutionLocation;
   }
-  return institutionLocation;
+  institutionLocation.integrationContacts = options?.integrationContacts ?? [
```
andrewsignori-aot, line 95: Is it a direct assignment?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
```diff
@@ -71,9 +88,16 @@ describe(
         { hasIntegration: true },
         { hasIntegration: false },
       );
+
+      // Update the date sent to current date to verify newly created notifications
+      // with date sent as null.
+      await db.notification.update(
+        { dateSent: IsNull() },
+        { dateSent: getISODateOnlyString(new Date()) },
```
andrewsignori-aot, line 96: `dateSent` is a regular date and date. We do not need the `getISODateOnlyString`.

hellolynn-tbtb (commented):
Content looks good. Thanks @dheepak-aot

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
```diff
@@ -135,9 +160,129 @@ describe(
...
+      const updatedDisbursement = await db.disbursementSchedule.findOne({
+        select: {
+          coeStatus: true,
+          tuitionRemittanceRequestedAmount: true,
+        },
+        where: { id: disbursement.id },
+      });
+      // Expect the COE status of the updated disbursement to be completed.
+      expect(updatedDisbursement.coeStatus).toBe(COEStatus.completed);
+      // The file has 3 detail records for the disbursement, with pay to school amount of 10 in each record.
```
andrewsignori-aot, line 279: Checking tickets #1911 and #1501 I was not able to find the specification about the SUM only for the Y records. If it is already captured there please disregard this comment, otherwise please update the ticket accordingly.

sources/packages/backend/libs/integrations/src/constants/system-configurations-constants.ts
```diff
@@ -15,3 +15,9 @@ export const ECE_RESPONSE_FILE_NAME = "CONR_008.TXT";
  */
 export const ECE_RESPONSE_COE_DECLINED_REASON =
   "A reason was not provided by your institution. Please speak to your Financial Aid Officer at your institution for details.";
+
+/**
+ * Name of the email attachment file sent in ECE Response processing notification.
+ */
+export const ECE_RESPONSE_ATTACHMENT_FILE_NAME =
+  "processing_summary_report.txt";
```
andrewsignori-aot, line 23: Just a suggestion, I believe that we can use a friendly file name, like `Processing Summary Report.txt`.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
dheepak-aot: I am yet to update the ticket, but thanks for checking and reminding.

sources/packages/backend/libs/utilities/src/base64-utils.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Encode a given string to Base64 string.
+ * @param stringToEncode string content to encode.
+ * @returns Base64 encoded string.
+ */
+export function base64Encode(stringToEncode: string): string {
+  return Buffer.from(stringToEncode, "binary").toString("base64");
```
andrewsignori-aot, line 7: The "binary" does not seem accurate since we are building from a string, Am I missing something?
Can we just use the default utf8 as `Buffer.from(stringToEncode).toString("base64");`?

dheepak-aot: You are right. It works with `Buffer.from(stringToEncode).toString("base64");` will change

sources/packages/backend/libs/test-utils/src/factories/application.ts
```diff
@@ -49,7 +49,12 @@ export function createFakeApplication(relations?: {
   application.applicationStatus = ApplicationStatus.Submitted;
   application.relationshipStatus = RelationshipStatus.Single;
   application.currentAssessment = relations?.currentStudentAssessment;
-  application.applicationNumber = faker.random.alphaNumeric(10);
+  // Application numbers are expected to be a string of number
+  // with fixed length of 10 characters.
+  application.applicationNumber = faker.random
```
andrewsignori-aot, line 54: I am still not 100% sure about this assumption but apparently is what we got from @JasonCTang, right?
Either way, the 10 digits number will not start with zero (I assume), so it would be better to change the `padStart` to a `padEnd` IMO.

dheepak-aot: apparently is what we got from @JasonCTang, right? - Yes. 
Sure, I can also set the min as 1000000000 in faker.random.number so that it always has 10 numbers.

sources/packages/backend/libs/services/src/notifications/notification/gc-notify.model.ts
```diff
@@ -41,3 +41,21 @@ export interface GCNotifyErrorResponse {
   errors: GCNotifyErrorDetail[];
   status_code: HttpStatus;
 }
+
+/**
+ * Email message format.
+ */
+export interface EmailMessage {
```
andrewsignori-aot, line 48: Event this being a `EmailMessage` inside GC Notify just wondering if it would be good to be explicit about the fact that this is notification specific `EmailMessage`, for instance, naming it as `NotificationEmailMessage`.

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -399,7 +408,60 @@ export class NotificationActionsService {
...
+          totalRecords: notification.totalRecords,
+          totalDisbursements: notification.totalDisbursements,
+          disbursementsSuccessfullyProcessed:
+            notification.disbursementsSuccessfullyProcessed,
+          disbursementsSkipped: notification.disbursementsSkipped,
+          duplicateDisbursements: notification.duplicateDisbursements,
+          disbursementsFailedToProcess:
+            notification.disbursementsFailedToProcess,
+          application_file: {
+            file: base64Encode(notification.attachmentFileContent),
```
andrewsignori-aot, line 448: Just double checking, is there a scenario where the file would not be generated?
When everything goes well, are we including the summary also in the file?

dheepak-aot: Yes, even when everything goes well we include the detailed summary in the file.

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
```diff
@@ -159,8 +176,18 @@ export class ECEResponseProcessingService {
       // Delete the ECE response file, if the file exist in remote server.
       if (isECEResponseFileExist) {
         await this.deleteProcessedFile(remoteFilePath, processSummary);
+
+        // Create notification email which gets sent to
+        // the integration contacts of the institution
+        // when a ece file exist for an institution.
```
andrewsignori-aot, line 182: "// when an ECE file exists for an institution."

sources/packages/backend/libs/integrations/src/constants/system-configurations-constants.ts
dheepak-aot: Sure. I assume we can join them with `_` right?

andrewsignori-aot (commented):
Great work and a great log/summary for the emails. Please take a look at the comments.

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-files/ece-response-file-detail.ts
andrepestana-aot: Yes, that's weird but in TS that's the way. I think it still checks for null/undefined.
https://stackoverflow.com/a/42121162/7983709

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
```diff
@@ -390,6 +417,81 @@ export class ECEResponseProcessingService {
     }
   }
 
+  /**
+   * Create ECE Response file processing notifications.
+   * @param institutionCode institution of the processing file.
+   * @param auditUserId user that creates notification.
```
andrepestana-aot, line 423: Please remove comment for `auditUserId` param.

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-files/ece-response-file-detail.ts
ann-aot: ya @andrepestana-aot. it says if we use `as any` then it bypasses compiler check and works as JS, weird
image

andrepestana-aot (commented):
Great job. Just left a minor comment.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
dheepak-aot: Had a discussion with same on @andrewsignori-aot and as per our dev discussion in one of the previous PR, using `options?: { initialValues?: Partial<InstitutionLocation> }` to set the initial values(integrationContacts here).

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -123,7 +127,7 @@ export class NotificationService extends RecordDataModelService<Notification> {
     // Call GC Notify send email method.
     try {
       await this.gcNotifyService.sendEmailNotification(
-        notification.messagePayload,
+        notification.messagePayload as EmailMessage,
```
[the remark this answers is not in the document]
dheepak-aot: As discussed, to use the message in typeorm  select, we have to use the work around of `messagePayload: true as unknown` in the typeorm find. 

We have different opinions from devs on this point. @andrewsignori-aot's opinion was using the work around and moving forward may be misleading for devs who are looking at the code. They may try to use the jsonb like relation objects removing the `as unknown` part. Correct me if wrong @andrewsignori-aot 

Having said that we can discuss this point on dev sync up to agree on how we can use the jsonb columns with typed property.

dheepak-aot:
> Added some comments @dheepak-aot. Thanks for adding the scenarios that we discussed in the test. I think we can also add the new message_payload type to the `SaveNotificationModel` interface

I added the message_payload type to `SaveNotificationModel` interface

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
andrewsignori-aot: To be clear, in the last PR, we talked about having `initialValues?: Partial<InstitutionLocation>` and we have an example of it in the code already but not inside options.
Yesterday during a chat we @dheepak-aot proposed to it be inside options, which I think is a good idea to follow from now on.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
```diff
@@ -58,12 +77,15 @@ export async function createInstitutionLocations(
...
  * @param e2eDataSources e2e data sources.
+ * @param options options.
+ * - `initialValues` initial values of institution location.
  * @returns institution location.
  */
 async function findOrCreateInstitutionLocation(
   institution: Institution,
   institutionCode: string,
   e2eDataSources: E2EDataSources,
+  options?: { initialValues?: Partial<InstitutionLocation> },
```
ann-aot, line 88: The code that already has, has `initialvalue` as a param (which was done as per the suggestion), 
image
Maybe we can finalize either one

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
```diff
@@ -72,11 +94,11 @@ async function findOrCreateInstitutionLocation(
   if (!institutionLocation) {
     const newInstitutionLocation = createFakeInstitutionLocation(institution);
     newInstitutionLocation.institutionCode = institutionCode;
-    institutionLocation = await e2eDataSources.institutionLocation.save(
-      newInstitutionLocation,
-    );
+    institutionLocation = newInstitutionLocation;
   }
-  return institutionLocation;
+  institutionLocation.integrationContacts = options?.initialValues
```
ann-aot, line 99: @dheepak-aot I remember. In our discussion, you were saying about a scenario where `integrationContacts` is null. is that not handled here? Or planning for the future?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
andrewsignori-aot: @ann-aot I comment about it here https://github.com/bcgov/SIMS/pull/2011#discussion_r1229876459
The idea of having it inside the options is also to not have to call methods providing undefined sometimes. The idea is to follow this, assuming that it would also make the development easier unless someone sees some drawback.

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
```diff
@@ -390,6 +417,86 @@ export class ECEResponseProcessingService {
...
+   */
+  private buildEmailAttachmentBody(
+    processSummaryResult: ProcessSummaryResult,
+  ): string {
+    const attachmentBody = new StringBuilder();
+    attachmentBody.appendLine("Summary:");
+    attachmentBody.appendLine(processSummaryResult.summary.join(END_OF_LINE));
+    const warnings = processSummaryResult.warnings.length
+      ? processSummaryResult.warnings.join(END_OF_LINE)
+      : "NONE";
```
ann-aot, line 489: @dheepak-aot sorry for asking this now, can u pls share the email attachment screenshot, I did;t able to find it in the PR description

dheepak-aot: No problem. I am going to update PR description.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
dheepak-aot: We can plan if that is required in future @ann-aot

sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
dheepak-aot: I updated the email attachment. Please have a look.

andrewsignori-aot (approved):

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
ann-aot: okay, so moving forward the plan is to add the `initialValues` in the options param

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-helper.ts
ann-aot: but null doesn't mean an empty array right? Anyway resolving the conversation as it is not in the scope of the PR

ann-aot (approved):
@dheepak-aot, Thank you for making the adjustments and for sharing the screenshot. 👍  As previously noted, if you have any concerns with my/devs' comments, kindly let us know. Please notify us first before proceeding to the dev chat and making a decision 😄 . Thanks for understanding 👍

dheepak-aot:
> @dheepak-aot, Thank you for making the adjustments and for sharing the screenshot. 👍 As previously noted a gentle reminder, if you have any concerns with my/devs' comments, kindly let us know first before proceeding to the dev chat and making a decision 😄 . Thanks for understanding 👍

Thanks for understanding too. To be clear the concern is not personal and if it was personal, direct message or call would have been the first thing that I will do. But the concern was about deviating from the path of making the code better which I strongly felt that it is better to share in the group. Having said that I know we are a great team and we will always get better in any given situation. 👍

PR #2013, opened 2023-06-13
#1953 - Request an Offering Change - Part 2

ann-aot:
- Added "available to change" UI ( used vuetify data table) and API.
- Used `@leftjoin` to join the `application.applicationOfferingChangeRequest` and `applicationOfferingChangeRequest` as there will always be one in-progress item for an application at a time.
- For pagination options in the controller, I have reused the same `ApplicationStatusPaginationOptionsAPIInDTO` used by the `active application` as it was exactly the same. let me know your thoughts.
- Added migration to add `application-offering-change-assessment-trigger-types`  and to add `application-offering-change-to-assessment` for the `StudentAssessments` table.
- There was a issue, in the previous `v-data table-server` during the search. When we search for something on any other page other than the first one, there was an unexpected behavior - resolved

**_Screenshots_**
image
image
image
image

Next PRs
1. In progress and completed tabs
2. E2E tests

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
@@ -1,0 +1,62 @@
+import { Controller, Get, Param, ParseIntPipe, Query } from "@nestjs/common";
+import { ApiTags } from "@nestjs/swagger";
+import { AuthorizedParties } from "../../auth/authorized-parties.enum";
+import { AllowAuthorizedParty, HasLocationAccess } from "../../auth/decorators";
+import { ClientTypeBaseRoute } from "../../types";
+import { getUserFullName } from "../../utilities";
+import {
+  OfferinfChangePaginationOptionsAPIInDTO,
+  PaginatedResultsAPIOutDTO,
+} from "../models/pagination.dto";
+import BaseController from "../BaseController";
+import { ApplicationOfferingChangeSummaryAPIOutDTO } from "./models/application-offering-change-request.institutions.dto";
+import { ApplicationOfferingChangeRequestService } from "../../services";
+
+/**
+ * Application offering change request controller for institutions client.
+ */
+@AllowAuthorizedParty(AuthorizedParties.institution)
+@Controller("location/:locationId/application-offering-change-request")
+@ApiTags(`${ClientTypeBaseRoute.Institution}-location-application-offering-change-request`)
+export class ApplicationOfferingChangeRequestInstitutionsController extends BaseController {
+  constructor(private readonly applicationOfferingChangeRequestService: ApplicationOfferingChangeRequestService) {
+    super();
+  }
+
+  /**
+   * Gets all eligible applications that can be requested for application
+   * offering change.
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible applications that can be requested for
+   * application offering change.
+   */
+  @HasLocationAccess("locationId")
+  @Get("available")
+  async getEligibleApplicationOfferingChangeRecords(
+    @Param("locationId", ParseIntPipe) locationId: number,
+    @Query() pagination: OfferinfChangePaginationOptionsAPIInDTO,
+  ): Promise<
+    PaginatedResultsAPIOutDTO<ApplicationOfferingChangeSummaryAPIOutDTO>
+  > {
+    const applications =
+      await this.applicationOfferingChangeRequestService.getEligibleApplicationOfferingChangeRecords(
+        locationId,
+        pagination,
+      );
+
+    return {
+      results: applications.results.map((eachApplication) => {
+        const offering = eachApplication.currentAssessment?.offering;
+        return {
+          applicationNumber: eachApplication.applicationNumber,
+          applicationId: eachApplication.id,
+          studyStartPeriod: offering?.studyStartDate,
+          studyEndPeriod: offering?.studyEndDate,
+          fullName: getUserFullName(eachApplication.student.user),
+        };
+      }),
+      count: applications.count,
+    };
+  }
+}
--- a/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
@@ -55,8 +55,8 @@ export class ApplicationOfferingChangeRequestInstitutionsController extends Base
         return {
           applicationNumber: eachApplication.applicationNumber,
           applicationId: eachApplication.id,
-          studyStartPeriod: offering?.studyStartDate,
-          studyEndPeriod: offering?.studyEndDate,
+          studyStartPeriod: offering.studyStartDate,
+          studyEndPeriod: offering.studyEndDate,
           fullName: getUserFullName(eachApplication.student.user),
         };
       }),
--- a/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
@@ -37,14 +37,14 @@ export class ApplicationOfferingChangeRequestInstitutionsController extends Base
    * application offering change.
    */
   @Get("available")
-  async getEligibleApplicationOfferingChangeRecords(
+  async getEligibleApplications(
     @Param("locationId", ParseIntPipe) locationId: number,
     @Query() pagination: OfferingChangePaginationOptionsAPIInDTO,
   ): Promise<
     PaginatedResultsAPIOutDTO<ApplicationOfferingChangeSummaryAPIOutDTO>
   > {
     const applications =
-      await this.applicationOfferingChangeRequestService.getEligibleApplicationOfferingChangeRecords(
+      await this.applicationOfferingChangeRequestService.getEligibleApplications(
         locationId,
         pagination,
       );
--- a/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
@@ -5,7 +5,7 @@ import { AllowAuthorizedParty, HasLocationAccess } from "../../auth/decorators";
 import { ClientTypeBaseRoute } from "../../types";
 import { getUserFullName } from "../../utilities";
 import {
-  OfferinfChangePaginationOptionsAPIInDTO,
+  OfferingChangePaginationOptionsAPIInDTO,
   PaginatedResultsAPIOutDTO,
 } from "../models/pagination.dto";
 import BaseController from "../BaseController";
@@ -17,9 +17,14 @@ import { ApplicationOfferingChangeRequestService } from "../../services";
  */
 @AllowAuthorizedParty(AuthorizedParties.institution)
 @Controller("location/:locationId/application-offering-change-request")
-@ApiTags(`${ClientTypeBaseRoute.Institution}-location-application-offering-change-request`)
+@ApiTags(
+  `${ClientTypeBaseRoute.Institution}-location-application-offering-change-request`,
+)
+@HasLocationAccess("locationId")
 export class ApplicationOfferingChangeRequestInstitutionsController extends BaseController {
-  constructor(private readonly applicationOfferingChangeRequestService: ApplicationOfferingChangeRequestService) {
+  constructor(
+    private readonly applicationOfferingChangeRequestService: ApplicationOfferingChangeRequestService,
+  ) {
     super();
   }
 
@@ -31,11 +36,10 @@ export class ApplicationOfferingChangeRequestInstitutionsController extends Base
    * @returns list of eligible applications that can be requested for
    * application offering change.
    */
-  @HasLocationAccess("locationId")
   @Get("available")
   async getEligibleApplicationOfferingChangeRecords(
     @Param("locationId", ParseIntPipe) locationId: number,
-    @Query() pagination: OfferinfChangePaginationOptionsAPIInDTO,
+    @Query() pagination: OfferingChangePaginationOptionsAPIInDTO,
   ): Promise<
     PaginatedResultsAPIOutDTO<ApplicationOfferingChangeSummaryAPIOutDTO>
   > {
--- a/sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
@@ -47,6 +47,7 @@ import {
   InstitutionLocationFormAPIInDTO,
   InstitutionLocationDetailsAPIOutDTO,
   InstitutionLocationsAPIOutDTO,
+  ApplicationOfferingChangeSummaryAPIOutDTO,
 } from "./models/institution-location.dto";
 import { FormNames } from "../../services/form/constants";
 import { transformAddressDetailsForAddressBlockForm } from "../utils/address-utils";
@@ -264,5 +265,43 @@ export class InstitutionLocationInstitutionsController extends BaseController {
         name: location.name,
       };
     });
+  }
+
+  /**
+   * Gets all eligible application that can be requested for application
+   * offering change.
+   * @param {number} locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible application that can be requested for
+   * application offering change.
+   */
+  @HasLocationAccess("locationId")
+  @Get(":locationId/active-applications/request-change")
+  // todo: ann add max validation to location id
+  async getEligibleApplicationOfferingChangeApplications(
+    @Param("locationId", ParseIntPipe) locationId: number,
+    @Query() pagination: ApplicationStatusPaginationOptionsAPIInDTO,
+  ): Promise<
+    PaginatedResultsAPIOutDTO<ApplicationOfferingChangeSummaryAPIOutDTO>
+  > {
+    const applications =
+      await this.applicationService.getEligibleApplicationOfferingChangeApplications(
+        locationId,
+        pagination,
+      );
+
+    return {
+      results: applications.results.map((eachApplication) => {
+        const offering = eachApplication.currentAssessment?.offering;
+        return {
+          applicationNumber: eachApplication.applicationNumber,
+          applicationId: eachApplication.id,
+          studyStartPeriod: getISODateOnlyString(offering?.studyStartDate),
+          studyEndPeriod: getISODateOnlyString(offering?.studyEndDate),
+          fullName: getUserFullName(eachApplication.student.user),
+        };
+      }),
+      count: applications.count,
+    };
   }
 }
--- a/sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
@@ -47,7 +47,6 @@ import {
   InstitutionLocationFormAPIInDTO,
   InstitutionLocationDetailsAPIOutDTO,
   InstitutionLocationsAPIOutDTO,
-  ApplicationOfferingChangeSummaryAPIOutDTO,
 } from "./models/institution-location.dto";
 import { FormNames } from "../../services/form/constants";
 import { transformAddressDetailsForAddressBlockForm } from "../utils/address-utils";
@@ -265,42 +264,5 @@ export class InstitutionLocationInstitutionsController extends BaseController {
         name: location.name,
       };
     });
-  }
-
-  /**
-   * Gets all eligible application that can be requested for application
-   * offering change.
-   * @param locationId location id.
-   * @param paginationOptions options to execute the pagination.
-   * @returns list of eligible application that can be requested for
-   * application offering change.
-   */
-  @HasLocationAccess("locationId")
-  @Get(":locationId/active-applications/request-change")
-  async getEligibleApplicationOfferingChangeRecords(
-    @Param("locationId", ParseIntPipe) locationId: number,
-    @Query() pagination: ApplicationStatusPaginationOptionsAPIInDTO,
-  ): Promise<
-    PaginatedResultsAPIOutDTO<ApplicationOfferingChangeSummaryAPIOutDTO>
-  > {
-    const applications =
-      await this.applicationService.getEligibleApplicationOfferingChangeRecords(
-        locationId,
-        pagination,
-      );
-
-    return {
-      results: applications.results.map((eachApplication) => {
-        const offering = eachApplication.currentAssessment?.offering;
-        return {
-          applicationNumber: eachApplication.applicationNumber,
-          applicationId: eachApplication.id,
-          studyStartPeriod: getISODateOnlyString(offering?.studyStartDate),
-          studyEndPeriod: getISODateOnlyString(offering?.studyEndDate),
-          fullName: getUserFullName(eachApplication.student.user),
-        };
-      }),
-      count: applications.count,
-    };
   }
 }
--- a/sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
@@ -270,22 +270,21 @@ export class InstitutionLocationInstitutionsController extends BaseController {
   /**
    * Gets all eligible application that can be requested for application
    * offering change.
-   * @param {number} locationId location id.
+   * @param locationId location id.
    * @param paginationOptions options to execute the pagination.
    * @returns list of eligible application that can be requested for
    * application offering change.
    */
   @HasLocationAccess("locationId")
   @Get(":locationId/active-applications/request-change")
-  // todo: ann add max validation to location id
-  async getEligibleApplicationOfferingChangeApplications(
+  async getEligibleApplicationOfferingChangeRecords(
     @Param("locationId", ParseIntPipe) locationId: number,
     @Query() pagination: ApplicationStatusPaginationOptionsAPIInDTO,
   ): Promise<
     PaginatedResultsAPIOutDTO<ApplicationOfferingChangeSummaryAPIOutDTO>
   > {
     const applications =
-      await this.applicationService.getEligibleApplicationOfferingChangeApplications(
+      await this.applicationService.getEligibleApplicationOfferingChangeRecords(
         locationId,
         pagination,
       );
--- a/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
@@ -92,3 +92,9 @@ export class PaginatedResultsAPIOutDTO<T> {
   results: T[];
   count: number;
 }
+
+export class OfferinfChangePaginationOptionsAPIInDTO extends PaginationOptionsAPIInDTO {
+  @IsOptional()
+  @IsIn(["applicationNumber", "fullName"])
+  sortField?: string;
+}
--- a/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
@@ -93,6 +93,9 @@ export class PaginatedResultsAPIOutDTO<T> {
   count: number;
 }
 
+/**
+ * Extended pagination option for the application offering change.
+ */
 export class OfferinfChangePaginationOptionsAPIInDTO extends PaginationOptionsAPIInDTO {
   @IsOptional()
   @IsIn(["applicationNumber", "fullName"])
--- a/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
@@ -96,7 +96,7 @@ export class PaginatedResultsAPIOutDTO<T> {
 /**
  * Extended pagination option for the application offering change.
  */
-export class OfferinfChangePaginationOptionsAPIInDTO extends PaginationOptionsAPIInDTO {
+export class OfferingChangePaginationOptionsAPIInDTO extends PaginationOptionsAPIInDTO {
   @IsOptional()
   @IsIn(["applicationNumber", "fullName"])
   sortField?: string;
--- a/sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
+++ b/sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
@@ -1,0 +1,103 @@
+import { Injectable } from "@nestjs/common";
+import { InjectRepository } from "@nestjs/typeorm";
+import {
+  Application,
+  ApplicationOfferingChangeRequestStatus,
+  ApplicationStatus,
+  getUserFullNameLikeSearch,
+  transformToEntitySortField,
+} from "@sims/sims-db";
+import { Brackets, Repository } from "typeorm";
+import { PaginatedResults, PaginationOptions } from "../../utilities";
+
+@Injectable()
+export class ApplicationOfferingChangeRequestService {
+  constructor(
+    @InjectRepository(Application)
+    private readonly applicationRepo: Repository<Application>,
+  ) {}
+
+  /**
+   * Gets all eligible applications that can be requested for application
+   * offering change.
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible applications that can be requested for
+   * application offering change.
+   */
+  async getEligibleApplicationOfferingChangeRecords(
+    locationId: number,
+    paginationOptions: PaginationOptions,
+  ): Promise<PaginatedResults<Application>> {
+    const applicationQuery = this.applicationRepo
+      .createQueryBuilder("application")
+      .select([
+        "application.applicationNumber",
+        "application.id",
+        "currentAssessment.id",
+        "offering.studyStartDate",
+        "offering.studyEndDate",
+        "student.id",
+        "user.firstName",
+        "user.lastName",
+      ])
+      .leftJoin("application.currentAssessment", "currentAssessment")
+      .leftJoin("currentAssessment.offering", "offering")
+      .leftJoin(
+        "application.applicationOfferingChangeRequest",
+        "applicationOfferingChangeRequest",
+      )
+      .leftJoin("application.student", "student")
+      .leftJoin("student.user", "user")
+      .where(
+        new Brackets((qb) =>
+          qb
+            .where("applicationOfferingChangeRequest.id IS NULL")
+            .orWhere(
+              "applicationOfferingChangeRequest.applicationOfferingChangeRequestStatus NOT IN (:...status)",
+              {
+                status: [
+                  ApplicationOfferingChangeRequestStatus.InProgressWithSABC,
+                  ApplicationOfferingChangeRequestStatus.InProgressWithStudent,
+                ],
+              },
+            ),
+        ),
+      )
+      .andWhere("application.location.id = :locationId", { locationId })
+      .andWhere("application.applicationStatus = :applicationStatus", {
+        applicationStatus: ApplicationStatus.Completed,
+      })
+      .andWhere("application.isArchived = false");
+
+    if (paginationOptions.searchCriteria?.trim()) {
+      applicationQuery
+        .andWhere(
+          new Brackets((qb) => {
+            qb.where(getUserFullNameLikeSearch()).orWhere(
+              "application.applicationNumber Ilike :searchCriteria",
+            );
+          }),
+        )
+        .setParameter(
+          "searchCriteria",
+          `%${paginationOptions.searchCriteria}%`,
+        );
+    }
+
+    applicationQuery
+      .orderBy(
+        transformToEntitySortField(
+          paginationOptions.sortField,
+          paginationOptions.sortOrder,
+        ),
+      )
+      .offset(paginationOptions.page * paginationOptions.pageLimit)
+      .limit(paginationOptions.pageLimit);
+    const [result, count] = await applicationQuery.getManyAndCount();
+    return {
+      results: result,
+      count,
+    };
+  }
+}
--- a/sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
+++ b/sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
@@ -5,7 +5,7 @@ import {
   ApplicationOfferingChangeRequestStatus,
   ApplicationStatus,
   getUserFullNameLikeSearch,
-  transformToEntitySortField,
+  transformToApplicationEntitySortField,
 } from "@sims/sims-db";
 import { Brackets, Repository } from "typeorm";
 import { PaginatedResults, PaginationOptions } from "../../utilities";
@@ -25,7 +25,7 @@ export class ApplicationOfferingChangeRequestService {
    * @returns list of eligible applications that can be requested for
    * application offering change.
    */
-  async getEligibleApplicationOfferingChangeRecords(
+  async getEligibleApplications(
     locationId: number,
     paginationOptions: PaginationOptions,
   ): Promise<PaginatedResults<Application>> {
@@ -41,14 +41,14 @@ export class ApplicationOfferingChangeRequestService {
         "user.firstName",
         "user.lastName",
       ])
-      .leftJoin("application.currentAssessment", "currentAssessment")
-      .leftJoin("currentAssessment.offering", "offering")
+      .innerJoin("application.currentAssessment", "currentAssessment")
+      .innerJoin("currentAssessment.offering", "offering")
       .leftJoin(
         "application.applicationOfferingChangeRequest",
         "applicationOfferingChangeRequest",
       )
-      .leftJoin("application.student", "student")
-      .leftJoin("student.user", "user")
+      .innerJoin("application.student", "student")
+      .innerJoin("student.user", "user")
       .where(
         new Brackets((qb) =>
           qb
@@ -87,7 +87,7 @@ export class ApplicationOfferingChangeRequestService {
 
     applicationQuery
       .orderBy(
-        transformToEntitySortField(
+        transformToApplicationEntitySortField(
           paginationOptions.sortField,
           paginationOptions.sortOrder,
         ),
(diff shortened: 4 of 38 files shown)


sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
```diff
@@ -265,4 +266,41 @@ export class InstitutionLocationInstitutionsController extends BaseController {
...
+  /**
+   * Gets all eligible application that can be requested for application
+   * offering change.
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible application that can be requested for
+   * application offering change.
+   */
+  @HasLocationAccess("locationId")
+  @Get(":locationId/active-applications/request-change")
```
andrewsignori-aot, line 279: The endpoint seems to be too generic to represent the offering change.
Should we start a new controller specific to the entire feature?
For instance:

@Get(":locationId/offering-change-request/[available,InProgress,etc]") - Institution/Ministry
@Get(":locationId/offering-change-request/") - Ministry
@Get(":locationId/offering-change-request/:offeringChangeRequestId") - Institution/Ministry
@Post(":locationId/offering-change-request") - Institution only
@Patch(":locationId/offering-change-request") - Ministry/Student

The above endpoints do not represent what we actually need, they were added to help to demonstrate the point that we will need a variety of endpoints for Institution/Ministry/Students, and a new folder under the `route-controllers` seems to be more reasonable. Thoughts

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
```diff
@@ -265,4 +266,41 @@ export class InstitutionLocationInstitutionsController extends BaseController {
...
+        pagination,
+      );
+
+    return {
+      results: applications.results.map((eachApplication) => {
+        const offering = eachApplication.currentAssessment?.offering;
+        return {
+          applicationNumber: eachApplication.applicationNumber,
+          applicationId: eachApplication.id,
+          studyStartPeriod: getISODateOnlyString(offering?.studyStartDate),
```
andrewsignori-aot, line 298: Offering `studyStartDate` and `studyEndDate` are already date-only columns that should be ready to be returned by the API and they are without the need to call `getISODateOnlyString`.

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
```diff
@@ -265,4 +266,41 @@ export class InstitutionLocationInstitutionsController extends BaseController {
       };
     });
   }
+
+  /**
+   * Gets all eligible application that can be requested for application
```
andrewsignori-aot, line 271: Minor but the `application` should be plural to enforce the idea of a list being returned. Same for the service method.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        "application.applicationNumber",
+        "application.id",
+        "currentAssessment.id",
+        "offering.studyStartDate",
+        "offering.studyEndDate",
+        "student.id",
+        "user.firstName",
+        "user.lastName",
+      ])
+      .leftJoin("application.currentAssessment", "currentAssessment")
```
andrewsignori-aot, line 1571: This and the next line can be changed to an `innerJoin`, unless I am missing something. Please let me know.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot, line 1: This service is way too big. We can break these services into more specific services. For instance, we can talk about having the queries related to this feature more centralized in one service instead of spread across multiple services.
The services, the API ones, and even more have been too centralized into only the "main table". When we look at services files like this one with 1633 I believe that it is way passed the time to start creating more services.
@dheepak-aot @guru-aot @andrepestana-aot @sh16011993 thoughts?

sources/packages/backend/apps/db-migrations/src/sql/Types/Add-application-offering-change-assessment-trigger-type.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TYPE sims.assessment_trigger_types
+ADD
+  VALUE if not exists 'Application offering change';
```
andrewsignori-aot, line 3: Can we remove the `if not exists`?

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -164,6 +165,26 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
...
+      {
+        path: AppRoutes.RequestAChange,
+        name: InstitutionRoutesConst.REQUEST_A_CHANGE,
+        components: {
+          default: RequestApplicationChange,
+          sidebar: InstitutionHomeSideBar,
+        },
+        props: {
+          default: (route: RouteLocationNormalized) => ({
+            locationId: parseInt(route.params.locationId as string),
```
andrewsignori-aot, line 177: ❤️

sources/packages/web/src/helpers/helpers/uri.ts
```diff
@@ -61,6 +62,9 @@ export const getPaginationQueryString = (
 ) => {
   if (enableZeroPage) {
     paginationOptions.page = paginationOptions.page - 1;
+    if (paginationOptions.searchCriteria) {
```
andrewsignori-aot, line 65: I see the point here but does it work when there is a filter that results in multiple pages? Can the user still navigate across the pages with a filter applied?

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
andrepestana-aot: I think it's a good idea to have a separare controller for "offering change request".

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
```diff
@@ -265,4 +266,41 @@ export class InstitutionLocationInstitutionsController extends BaseController {
...
+   * offering change.
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible application that can be requested for
```
dheepak-aot, line 275: https://github.com/bcgov/SIMS/pull/2013/files#r1232859566

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
dheepak-aot: I agree that having a seperate controller will make it better to organize the endpoints for the feature.
Having said that, for the endpoints, using offering-change-request will be more precise from API side.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+            ),
+        ),
+      )
+      .andWhere("application.location.id = :locationId", { locationId })
+      .andWhere("application.applicationStatus = :applicationStatus", {
+        applicationStatus: ApplicationStatus.Completed,
+      })
+      .andWhere("application.isArchived = false");
+
+    if (paginationOptions.searchCriteria) {
```
dheepak-aot, line 1600: `if(paginationOptions.searchCriteria?.trim())` ?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .where(
+        new Brackets((qb) =>
+          qb
+            .where("applicationOfferingChangeRequest.id IS NULL")
+            .orWhere(
+              "applicationOfferingChangeRequest.applicationOfferingChangeRequestStatus NOT IN (:...status)",
+              {
+                status: [
+                  ApplicationOfferingChangeRequestStatus.InProgressWithSABC,
+                  ApplicationOfferingChangeRequestStatus.InProgressWithStudent,
```
andrepestana-aot, line 1588: Should not include "not in `DeclinedByStudent` and `DeclinedBySABC`"?

sources/packages/backend/apps/db-migrations/src/sql/Types/Add-application-offering-change-assessment-trigger-type.sql
dheepak-aot, line 3: Should we consider using AFTER or BEFORE an existing enum value?

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
```diff
@@ -327,6 +328,20 @@ export class Application extends RecordDataModel {
...
+    () => ApplicationOfferingChangeRequest,
+    (applicationOfferingChangeRequest) =>
+      applicationOfferingChangeRequest.application,
+    {
+      eager: false,
+      cascade: false,
+      nullable: true,
+    },
+  )
+  applicationOfferingChangeRequest?: ApplicationOfferingChangeRequest[];
```
dheepak-aot, line 344: @ann-aot @andrewsignori-aot @guru-aot  I am confused here. I was expecting the relationship to be established only between student_assessment and application_offering_change similar to the student appeal. 

Just trying to understand what could be benefit of establishing this one to many relationship?

andrepestana-aot (commented):
Great job! Left a comment. Please take a look.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
```diff
@@ -207,6 +208,19 @@ export class StudentAssessment extends RecordDataModel {
     },
   )
   disbursementSchedules?: DisbursementSchedule[];
+  /**
+   * When the reassessment happen due to an application offering change request (by institution),
+   * this will provide to the workflow the data that need be changed.
+   */
+  @ManyToOne(() => ApplicationOfferingChangeRequest, {
```
dheepak-aot, line 215: Is it not one-to-one?

Correct me if wrong.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .leftJoin(
+        "application.applicationOfferingChangeRequest",
+        "applicationOfferingChangeRequest",
+      )
+      .innerJoin("application.student", "student")
+      .innerJoin("student.user", "user")
+      .where(
+        new Brackets((qb) =>
+          qb
+            .where("applicationOfferingChangeRequest.id IS NULL")
```
dheepak-aot, line 1582: we could use `currentAssessment.applicationOfferingChangeRequest` right?

sources/packages/web/src/components/institutions/request-a-change/AvailableToChangeSummary.vue
```diff
@@ -0,0 +1,169 @@
...
+            <v-data-table-server
+              :headers="AvailableToChangeOfferingChangeSummaryHeaders"
+              :items="applications.results"
+              :items-length="applications.count"
+              :loading="loading"
+              :items-per-page="DEFAULT_PAGE_LIMIT"
+              @update:options="paginationAndSortEvent"
+            >
+              <template #[`item.fullName`]="{ item }">
+                <span data-cy="fullName">{{ item.columns.fullName }} </span>
```
dheepak-aot, line 41: Just to clarify. Are we still considering adding data-cy?

sources/packages/web/src/constants/routes/RouteConstants.ts
```diff
@@ -82,6 +82,7 @@ export const InstitutionRoutesConst = {
   STUDENT_APPEAL_REQUEST: Symbol(),
   ASSESSMENT_AWARD_VIEW: Symbol(),
   NOTICE_OF_ASSESSMENT_VIEW: Symbol(),
+  REQUEST_A_CHANGE: Symbol(),
```
dheepak-aot, line 85: REQUEST_CHANGE ?

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: no, you are correct. I actually copied from active application query and missed to change

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
```diff
@@ -120,6 +120,18 @@ export default defineComponent({
                   },
                 },
               },
+              {
+                title: "Request a Change",
```
dheepak-aot, line 124: Question here to @hellolynn-tbtb 

Is it ok to have a button called `Request a Change` for student(request an appeal) and in institutions for different purpose?

image

sources/packages/web/src/router/InstitutionRoutes.ts
dheepak-aot, line 177: 👍

sources/packages/web/src/types/AppRoutes.ts
```diff
@@ -36,6 +36,7 @@ export enum AppRoutes {
   LocationProgramsOfferingsCreate = "location-offerings/location/:locationId/education-program/:programId/create",
   LocationUsers = "location-users/",
   ActiveApplicationsSummary = "active-applications/location/:locationId/summary",
+  RequestAChange = "request-change/location/:locationId/summary",
```
dheepak-aot, line 39: RequestChange?

sources/packages/backend/apps/api/src/services/application/application.service.ts
guru-aot: I do agree @andrewsignori-aot , instead of running around multiple services for this implementation, we can have a specific service for this implementations.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: in that case we need to include Approved as well. Above condition is relatively better IMO.

dheepak-aot:
Great work. Please have a look at the comments.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: Same page.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+            .orWhere(
```
guru-aot, line 1583: Can you please explain why we need this applicationOfferingChangeRequestStatus Or condition? Is that to show the applications that are already approved or declined? If that is the case, IMO we can just change the condition of Id and add an or to have the where with status approved or denied

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: There are methods in this service file related to PIR, those can be moved out into a separate service file to keep this one more targeted to application requests.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        "user.firstName",
+        "user.lastName",
+      ])
+      .leftJoin("application.currentAssessment", "currentAssessment")
+      .leftJoin("currentAssessment.offering", "offering")
+      .leftJoin(
+        "application.applicationOfferingChangeRequest",
+        "applicationOfferingChangeRequest",
+      )
+      .innerJoin("application.student", "student")
```
sh16011993, line 1577: Should this be a left or an inner join? Since we only need the students for the application and not all the students in my understanding.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .innerJoin("student.user", "user")
```
sh16011993, line 1578: Same as https://github.com/bcgov/SIMS/pull/2013/files#r1234425399

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible application that can be requested for
+   * application offering change.
+   */
+  async getEligibleApplicationOfferingChangeRecords(
+    locationId: number,
+    paginationOptions: PaginationOptions,
+  ): Promise<PaginatedResults<Application>> {
+    const applicationQuery = this.repo
+      .createQueryBuilder("application")
```
sh16011993, line 1560: Are we not using the other way of writing the query like below:
`
select: {},
relations: {},
where: {}
`

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
guru-aot: The application can have as many offering change requests, so this is a direct relationship to the change requests from the institution. Its more like a starting point.
But in the student appeal, the student assessment is connected when the ministry approves the appeal request.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
guru-aot: Yes you are right, once the application request is approved by ministry it will be updated. It should be OneToOne relation.

guru-aot (commented):
Added some comments, overall intensive and nice work @ann-aot

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: I think Declined requests become re-eligible for offering change. @ann-aot can confirm.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
     });
   }
 
+  /**
+   * Gets all eligible application that can be requested for application
```
sh16011993, line 1548: Shouldn't this be "Get all the eligible applications"?

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: we have been using `trim()` inside the query search for all the pagination, that's the reason I was also following the same. let me know the way devs prefer

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: no @andrepestana-aot, if a request is in any completed state (declined/approved) then institution will be again able to request another change if it's still not achieved

sources/packages/backend/apps/db-migrations/src/sql/Types/Add-application-offering-change-assessment-trigger-type.sql
ann-aot: for the assessment trigger type, I don't think it matters. correct me if Iam wrong @andrewsignori-aot @JasonCTang @guru-aot @andrepestana-aot

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+                  ApplicationOfferingChangeRequestStatus.InProgressWithSABC,
+                  ApplicationOfferingChangeRequestStatus.InProgressWithStudent,
+                ],
+              },
+            ),
+        ),
+      )
+      .andWhere("application.location.id = :locationId", { locationId })
+      .andWhere("application.applicationStatus = :applicationStatus", {
+        applicationStatus: ApplicationStatus.Completed,
```
sh16011993, line 1596: Do we need to check here that the application disbursement has not been processed. In that case, from my understanding, the application offering change request cannot be made, right?

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
ann-aot: If I got your question correctly @dheepak-aot , we have been following iOneToMany` in a lot of places wherever needed to make the query easier. I believe we have also used it in the student appeal that you mentioned, 
image

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      })
+      .andWhere("application.isArchived = false");
```
sh16011993, line 1598: For my understanding, after the disbursements for the application are processed, does the application go to the archived state?

sources/packages/web/src/components/institutions/request-a-change/AvailableToChangeSummary.vue
ann-aot: so, there wn;t be any automation test happing anything? I don't know the current status, but  I was under the impression that there will be automation testing happing and we decided to add data_cy for future developments, I don't remember any discussion mentioning we need to stop it. @andrewsignori-aot  @JasonCTang @guru-aot @andrepestana-aot

sources/packages/web/src/types/AppRoutes.ts
ann-aot: we already had a RequestChange, 
image. I think I will give a different name to avoid confusion, even though there is an institution comment

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: This query is for the `available for request` applications list,  image

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: As I mentioned in a standup. I initially started with the Typeform query and i was not able to achieve the full name search with the typeom query, that's why I chose querybuilder

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: I agree with @dheepak-aot. There is no need to create the query if the `searchCriteria` is a bunch of white space and line terminator characters.

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: mm, My undeserving is that it can happen anytime when the application is completed (not archived) irrespective of the disbursement status. If a disbursement already happened, then the offering change assessment will create over wards. correct me if I am wrong @andrepestana-aot  @JasonCTang @dheepak-aot @guru-aot @andrepestana-aot

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: there is an archived day configured, so, once a completed application is past that day from the offering's last day, the application will be archived, and another way application gets archived is when there is a scholastic standing happening for an application.
Reference
https://vscode.dev/github/bcgov/SIMS/blob/feature/sims-1953/part-2/sources/packages/backend/apps/queue-consumers/src/services/application/application.service.ts#L26
https://vscode.dev/github/bcgov/SIMS/blob/feature/sims-1953/part-2/sources/packages/backend/apps/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts#L251

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+          paginationOptions.sortField,
+          paginationOptions.sortOrder,
+        ),
+      )
+      .offset(paginationOptions.page * paginationOptions.pageLimit)
+      .limit(paginationOptions.pageLimit);
+    const [result, count] = await applicationQuery.getManyAndCount();
+    return {
+      results: result,
+      count: count,
```
sh16011993, line 1627: Can we reduce it to:
`
return {
results: result,
count
}
`

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    }
+
+    applicationQuery
+      .orderBy(
+        this.transformToEntitySortField(
+          paginationOptions.sortField,
+          paginationOptions.sortOrder,
+        ),
+      )
+      .offset(paginationOptions.page * paginationOptions.pageLimit)
```
sh16011993, line 1622: Can we use `skip` instead of `offset` here as `offset` may not work correctly in case of joins.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1543,6 +1544,90 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .limit(paginationOptions.pageLimit);
```
sh16011993, line 1623: Can we use `take` method instead of `limit` here as `limit` may not work as expected in case of joins.

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
```diff
@@ -265,4 +266,41 @@ export class InstitutionLocationInstitutionsController extends BaseController {
...
+   * application offering change.
+   */
+  @HasLocationAccess("locationId")
+  @Get(":locationId/active-applications/request-change")
+  async getEligibleApplicationOfferingChangeRecords(
+    @Param("locationId", ParseIntPipe) locationId: number,
+    @Query() pagination: ApplicationStatusPaginationOptionsAPIInDTO,
```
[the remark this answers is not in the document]
andrewsignori-aot: I would not mind keeping these DTOs specific for the target endpoint.
@ann-aot just double-checking, the idea would be to reuse the same pagination DTO for multiple endpoints?

sources/packages/backend/apps/db-migrations/src/sql/Types/Add-application-offering-change-assessment-trigger-type.sql
andrewsignori-aot: My understanding is that in the way it is it will be just appended to the end. I do not see a problem with it.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
andrewsignori-aot: I believe that we have not been enforcing the one-to-one for similar ones in the same entity.
For instance, to be a real one-to-one on DB should we create a unique constraint on the foreign key column, right?
We can talk about having this as one-to-one on Typeorm. Is a suggestion about having the DB structure changed or only the decorator for Typeorm?

sources/packages/web/src/helpers/helpers/uri.ts
ann-aot: nice catch

sources/packages/web/src/components/institutions/request-a-change/AvailableToChangeSummary.vue
andrewsignori-aot: I prefer to avoid using it at all for now since we have no directions.
If we have the commitment to continue to add it we also need to be clear about how we will be adding.

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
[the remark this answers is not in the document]
ann-aot: the extended pagination dto 
image
as its the same @andrewsignori-aot

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
```diff
@@ -207,6 +208,19 @@ export class StudentAssessment extends RecordDataModel {
     },
   )
   disbursementSchedules?: DisbursementSchedule[];
+  /**
+   * When the reassessment happen due to an application offering change request (by institution),
+   * this will provide to the workflow the data that need be changed.
```
sh16011993, line 213: Shouldn't it be "Data that needs to be changed."?

sources/packages/backend/apps/api/src/route-controllers/institution-locations/institution-location.institutions.controller.ts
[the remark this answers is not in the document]
ann-aot: as mentioned, created a separate one for the endpoint

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: so, we will need a ticket to update all the pagination query

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: Applications with disbursements sent can be reassessed. There is nothing in this feature stating otherwise.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
ann-aot: yes, we are not using OneToOne in similar scenarios

sources/packages/web/src/components/institutions/request-a-change/AvailableToChangeSummary.vue
ann-aot: okay, then removing the `data-cy`

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: what is the issue you are pointing here @sh16011993? Actually, we here using offset as there was some issue with the skip previously, I don't quite recall it.  @andrewsignori-aot @dheepak-aot @guru-aot @andrepestana-aot do you guys recall it

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: As part of the PR, created separate service for the feature

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: I was just trying to suggest the below:

image

I also found an additional link for the same:
https://www.kindacode.com/snippet/pagination-in-typeorm-find-options-querybuilder/#Using_QueryBuilder

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: https://github.com/bcgov/SIMS/pull/2013#discussion_r1234565461

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: @andrewsignori-aot Do you mean applications with disbursement sent can be requested for offering intensity change?

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: Thanks @ann-aot

sh16011993:
Nice work 👍 . Please have a look at the comments.

sources/packages/web/src/components/layouts/Institution/sidebar/HomeSideBar.vue
hellolynn-tbtb: @dheepak-aot @ann-aot hey yup - business is okay with it. From UX - I don't see it being a confusing experience as it is within their own portal. Thanks

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
andrewsignori-aot, line 1: This file does not look completely formatted. Can you please check it?

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible applications that can be requested for
+   * application offering change.
+   */
+  @HasLocationAccess("locationId")
+  @Get("available")
+  async getEligibleApplicationOfferingChangeRecords(
+    @Param("locationId", ParseIntPipe) locationId: number,
+    @Query() pagination: OfferinfChangePaginationOptionsAPIInDTO,
```
dheepak-aot, line 38: Minor. g instead of f  in offering in the variable name.

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Gets all eligible applications that can be requested for application
+   * offering change.
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible applications that can be requested for
+   * application offering change.
+   */
+  @HasLocationAccess("locationId")
```
andrewsignori-aot, line 34: Since we have the `location/:locationId` at the controller level, I would prefer to have the decorator move to the class level also.

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
ann-aot: you are correct, I think there is some issue with my formatter, its not properly formatting on save. I will take a look

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
andrewsignori-aot, line 38: Typo on `OfferinfChangePaginationOptionsAPIInDTO`.

sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Gets all eligible applications that can be requested for application
+   * offering change.
+   * @param locationId location id.
+   * @param paginationOptions options to execute the pagination.
+   * @returns list of eligible applications that can be requested for
+   * application offering change.
+   */
+  async getEligibleApplicationOfferingChangeRecords(
```
andrewsignori-aot, line 28: The `records` is not adding any value here IMO.
I would prefer to have it named `getEligibleApplicationOfferingChanges`.

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+        pagination,
+      );
+
+    return {
+      results: applications.results.map((eachApplication) => {
+        const offering = eachApplication.currentAssessment?.offering;
+        return {
+          applicationNumber: eachApplication.applicationNumber,
+          applicationId: eachApplication.id,
+          studyStartPeriod: offering?.studyStartDate,
```
andrewsignori-aot, line 58: `studyStartDate` and `studyEndDate` cannot be null anymore (even DB states otherwise) and they definitely cannot be null for completed applications.
I would prefer that these ones and the DTO would express them as required.

sources/packages/backend/apps/api/src/route-controllers/models/pagination.dto.ts
andrewsignori-aot, line 1: This file also has some changes if I save it on my VS Code. Are we not using the same configurations?

sources/packages/backend/apps/db-migrations/src/sql/Types/Add-application-offering-change-assessment-trigger-type.sql
dheepak-aot: Got it.  It was off my memory that it appends to the end. I am good 😊

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
dheepak-aot: I understood this from the context of https://github.com/bcgov/SIMS/pull/2013#discussion_r1234459532

sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
andrewsignori-aot: `getEligibleApplicationOfferingChanges` makes sense to me.
When appending `Records` I was just concerned that it can become a pattern and we would start to see `Records` appended everywhere 😉

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: I am not able to see this change. Is it expected to be changed already?

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
dheepak-aot: @ann-aot @andrewsignori-aot when I entered the comment it was keeping in mind the typeorm decorator. What you mentioned makes sense to have the enforcement in db as well. 

I saw application exception as an instance following one to one decorator. Let me know if I am missing anything.

Either way @ann-aot this comment is not a blocker as we do not have an agreement currently on this.

sources/packages/backend/libs/sims-db/src/pagination-utils.ts
andrewsignori-aot, line 1: The file is not formatted.

sources/packages/backend/libs/sims-db/src/pagination-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+import { OrderByCondition } from "typeorm";
+
+ /**
+   * Transformation to convert the data table column name to database column name.
+   * Any changes to the data object (e.g data table) in presentation layer must be adjusted here.
+   * @param sortField database fields to be sorted.
+   * @param sortOrder sort order of fields (Ascending or Descending order).
+   * @returns OrderByCondition
+   */
+ export function transformToEntitySortField(
```
andrewsignori-aot, line 11: Is this method supposed to be generic?
When we need to use column references like `application.applicationNumber` I would assume that it would be dedicated to certain queries and it should be named accordingly.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
andrewsignori-aot: @dheepak-aot @guru-aot thoughts?

sources/packages/web/src/helpers/helpers/uri.ts
andrewsignori-aot: I small ticket would be better @ann-aot. Do you mind creating one?

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
ann-aot: we do was some old data with no offerings, which is causing below error, 
image
so, adding the `?` back @andrewsignori-aot

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: Yes, they can. The disbursements being sent are not a problem or a blocker. Even if both disbursements were sent the application can still be reassessed to give more money to a student or generate an overaward. 
Disbursement sent, so far, does not block anything.

sources/packages/backend/libs/sims-db/src/pagination-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  sortField = "applicationNumber",
+  sortOrder = FieldSortOrder.ASC,
+): OrderByCondition {
```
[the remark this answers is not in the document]
ann-aot: it was already there, I just moved it to a new file. but, make sense to add an interface

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: The last time that we went through this the conclusion was:
`take` + `skip` will give the precise results but they are not executed on DB, meaning, typeorm will not add anything extra to the query and the pagination will happen in memory, basically killing the idea of why we do pagination.
`limit` and `offset` will actually add extra instructions to the query execution of the pagination on DB. The possible incorrect results are probably because if the query results in relationships where a `one to many` will result in `many`, the count can be affected. So, whether it's a problem or not depends on the query that we are executing.

I am not sure if the behavior state above changed or not. The basic way to check it is by updating the query to use `take` + `skip` and checking if the generated SQL changes.

dheepak-aot (approved):
Thanks for making the changes @ann-aot 👍

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: Got it. Thank You @andrewsignori-aot

andrepestana-aot (approved):
Thanks for the changes and explanations.  Looks good to me!

sources/packages/web/src/services/ApplicationOfferingChangeRequestService.ts
```diff
@@ -0,0 +1,35 @@
+import { PaginationOptions } from "../types";
```
andrewsignori-aot, line 1: Please change to import from the index `"@/types"`.

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: @andrewsignori-aot I understood the part that if the pagination happens in memory for `take` + `skip` , the purpose is defeated since all records are returned from the db and the filtering happens in memory. If the `offset`+ `limit` returns the records correctly, then we can keep it as is.

andrewsignori-aot (approved):
Thanks for doing the changes, its look good 👍

guru-aot (approved):
LGTM, nice work @ann-aot

Issue #2006, opened 2023-06-12
Remediate high risk vulnerabilities from WAVA report


JasonCTang:
**Describe the task**
Review and remediate the high risk vulnerabilities from the attached WAVA report

**Acceptance Criteria**
- [ ] Fix pagination vulnerability.
- [ ] Deploy fixes to Test for re-scan.

**Additional context**
[https://app.zenhub.com/files/319429847/4acbc6e6-7de0-4a90-b056-50e69ea61aed/download](https://app.zenhub.com/files/319429847/4acbc6e6-7de0-4a90-b056-50e69ea61aed/download)

andrewsignori-aot:
The `API Mass Assignment` was not considered critical to fix for the v1.0.0 release based on the below.
- All data received by the API is mapped through specific DTOs;
- The API framework ([Nestjs](https://docs.nestjs.com/)) `whitelist` configuration is enabled which ensures that any additional property received will be removed before the actual process happen. More information on https://docs.nestjs.com/techniques/validation.
- The team agreed to have the configuration `forbidNonWhitelisted` also enabled for the upcoming releases ensuring that the API will return an error instead of a success 200 which will comply better with the Wava requirements. The decision was made based on the fact that globally enabling `forbidNonWhitelisted` right now can potentially introduce side effects and the application may not be fully tested prior to release v1.0.0.

PR #2019, opened 2023-06-15
#2006 - Remediate high risk vulnerabilities from WAVA report

andrewsignori-aot:
- Added a max value to the page number that is globally used for all API endpoints that have pagination enabled. This fix the 45 issues on Wava identified in the category `Integer Overflow`.
- Added an E2E test to one of the endpoints mentioned on the Wava scan: `confirmation-of-enrollment/enrollmentPeriod/current?page=99999999999999999999&pageLimit=10`. This E2E test case file was used because it already has tests created and it was easier to have this one added.
- The issues on the category `API Mass Assignment` are not part of this PR. Further details on the associated ticket.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/confirmation-of-enrollment/_tests_/e2e/confirmation-of-enrollment.institutions.getCOESummary.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/confirmation-of-enrollment/_tests_/e2e/confirmation-of-enrollment.institutions.getCOESummary.e2e-spec.ts
@@ -128,6 +128,30 @@ describe("ConfirmationOfEnrollmentInstitutionsController(e2e)-getCOESummary", ()
       });
   });
 
+  it("Should return a BadRequest error when the page number has an invalid integer.", async () => {
+    // Arrange
+    const collegeCLocation = createFakeInstitutionLocation(collegeC);
+    await authorizeUserTokenForLocation(
+      appDataSource,
+      InstitutionTokenTypes.CollegeCUser,
+      collegeCLocation,
+    );
+    const invalidPage = Number.MAX_SAFE_INTEGER + 1;
+    const endpoint = `/institutions/location/${collegeCLocation.id}/confirmation-of-enrollment/enrollmentPeriod/${EnrollmentPeriod.Current}?page=${invalidPage}&pageLimit=10&sortField=disbursementDate&sortOrder=ASC`;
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/api/src/route-controllers/confirmation-of-enrollment/_tests_/e2e/confirmation-of-enrollment.institutions.getCOESummary.e2e-spec.ts
```diff
@@ -128,6 +128,30 @@ describe("ConfirmationOfEnrollmentInstitutionsController(e2e)-getCOESummary", ()
       });
   });
 
+  it("Should return a BadRequest error when the page number has an invalid integer.", async () => {
```
andrepestana-aot, line 131: Should we create a similar test for every endpoint with pagination also? Are all pagination endpoint using that same dto?

andrepestana-aot (approved):
Looks good. Just left a comment for double checking as I think we had in the past some pagination that didn't follow the same logic but I'm not sure if it has been changed.

sources/packages/backend/apps/api/src/route-controllers/confirmation-of-enrollment/_tests_/e2e/confirmation-of-enrollment.institutions.getCOESummary.e2e-spec.ts
andrewsignori-aot: I see the point. I believe that we do not need to revalidate the same over and over again but we do need to enforce that all pagination uses the same DTO. In the past, we did an effort to enforce this for all paginations. For this PR I did not check every single one.

andrewsignori-aot: I did not see it as a critical point to be tested.
The intention was to have one test case added to validate the scenario. I would like to have it deployed to TEST as soon as possible to request a new Wava scan.

sh16011993 (approved):
Looks good to me 👍 Thank You for fixing the Integer Overflow vulnerability.

ann-aot (approved):
Good work @andrewsignori-aot 👍

dheepak-aot (approved):
Nice work 👍

Issue #2018, opened 2023-06-14
Patroni update on dev, test and prod


guru-aot:
Update the Patroni db changes as per BC gov standards to dev, test and then to prod
Deployment activities and debugging on prod namespace

PR #2021, opened 2023-06-16
#2018 - Patroni update on Dev Test and Prod, DB-migration revert enable in API

guru-aot:
- Removed the fields in the service definition, which does not allow the connections of other pods to the stateful set.
-  Enable db-migration scripts to run on atleast one pod when revert is needed, so enabled it in API pod


devops/openshift/database/patroni-deploy.yml
```diff
@@ -31,10 +31,6 @@ objects:
       role: master
       cluster-name: ${NAME}
       app.kubernetes.io/name: patroni
-      sessionAffinity: None
```
andrewsignori-aot, line 34: I notice that these configs were not present in the past either, nice catch.
Was it tested on DEV?

andrewsignori-aot (approved):
Nice catch, looks good 👍

ann-aot (approved):
👍

dheepak-aot (approved):
👍

devops/openshift/database/patroni-deploy.yml
andrepestana-aot, line 34: Is those properties the cause the problem to not work and need to do some commands for this to have a main instance?

andrepestana-aot (approved):

devops/openshift/database/patroni-deploy.yml
sh16011993: So basically removing these properties allows for connection with services outside this cluster?

sh16011993, line 34: So basically removing these properties allows for connection with services outside this cluster?

sh16011993 (approved):
👍

sources/packages/backend/apps/api/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN npm ci
 # Copying sources.
 COPY ./apps/api ./apps/api
 COPY ./libs ./libs
+COPY ./apps/db-migrations ./apps/db-migrations
```
dheepak-aot, line 18: 👍

Issue #1997, opened 2023-06-07
Upgrade zeebe-node package


andrewsignori-aot:
Update the Zeebe client package to version [8.2.3](https://github.com/camunda-community-hub/zeebe-client-node-js/releases/tag/8.2.3) to resolve the dependabot ticket https://github.com/bcgov/SIMS/pull/1993

There is a breaking change that impacts the Worflows E2E tests and needs to be addressed alongside the upgrade.

PR #2031, opened 2023-06-20
#1997 - Upgrade Zeebe node version

dheepak-aot:
## Update Zeebe node version

 - [x] Updated to zeebe node version 8.2.3
 - [x] Updated the workflow test-util models to extend JSONDoc to be compliant with `createProcessInstanceWithResult()` of Zeebe client.
 
### Note
Initial effort was to upgrade to Zeebe 8.2.4 latest version of zeebe node (https://www.npmjs.com/package/zeebe-node?activeTab=versions) but upgrading to latest version brought back the GRPC resource exhausted error logs. 
Hence upgrading with second latest/penultimate version 8.2.3.

### DIFF
--- a/sources/packages/backend/workflow/test/models/assessment.model.ts
+++ b/sources/packages/backend/workflow/test/models/assessment.model.ts
@@ -4,8 +4,9 @@ import {
   YesNoOptions,
   OfferingDeliveryOptions,
 } from "@sims/test-utils";
+import { JSONDoc } from "zeebe-node";
 
-export interface StudentDependent {
+export interface StudentDependent extends JSONDoc {
   dateOfBirth: string;
   attendingPostSecondarySchool: YesNoOptions;
   declaredOnTaxes: YesNoOptions;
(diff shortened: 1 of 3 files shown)


guru-aot (approved):
LGTM,, nice work @dheepak-aot

sources/packages/backend/workflow/test/models/assessment.model.ts
```diff
@@ -86,7 +87,7 @@ export interface AssessmentConsolidatedData {
   studentDataPartnerStudyWeeks?: number;
   studentDataPartnerEmploymentInsurance?: YesNoOptions;
   studentDataPartnerFedralProvincialPDReceiptCost?: number;
-  studentDataParentDependentTable?: unknown;
+  studentDataParentDependentTable?: JSONDoc;
```
ann-aot, line 90: just to check, should this be a list of JSONDoc

ann-aot (approved):
LGTM 👍 . Added a minor qn.

andrewsignori-aot (approved):
Nice catch, looks good 👍

sources/packages/backend/workflow/test/models/assessment.model.ts
dheepak-aot: May be. But we never populated data to these values in tests, so kept it as JSONDoc. When we create an interface to this value in future we can update to list of that interface that extends JSONDoc.

andrepestana-aot (approved):
Great job! Looks good!

PR #2038, opened 2023-06-21
#1997 - Fix camunda deployments

dheepak-aot:
## Close Zeebe connection on camunda deployments

We have an ongoing issue on camunda definition deployments when deploying to camunda SaaS not in local.

The deployment gets executed but the command does not terminate(happens in local and github actions as well). 

`zeebeClient.close()` has been added to code and tested that it fixes the issue.

### DIFF
--- a/sources/packages/backend/workflow/src/deploy.ts
+++ b/sources/packages/backend/workflow/src/deploy.ts
@@ -125,6 +125,6 @@ import {
     console.error(error);
     throw error;
   } finally {
-    zeebeClient.close();
+    await zeebeClient.close();
   }
 })();
--- a/sources/packages/backend/workflow/src/deploy.ts
+++ b/sources/packages/backend/workflow/src/deploy.ts
@@ -124,5 +124,7 @@ import {
     console.error("Error while executing the deployment.");
     console.error(error);
     throw error;
+  } finally {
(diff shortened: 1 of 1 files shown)


sh16011993 (approved):
LGTM 👍 Thank You Dheepak for fixing the issue.

sources/packages/backend/workflow/src/deploy.ts
```diff
@@ -124,5 +124,7 @@ import {
     console.error("Error while executing the deployment.");
     console.error(error);
     throw error;
+  } finally {
+    zeebeClient.close();
```
andrewsignori-aot, line 128: I believe that there is an `await` missing. Can you check, please?

guru-aot (approved):
LGTM

sources/packages/backend/workflow/src/deploy.ts
dheepak-aot: Yes it is a promise and added await.

andrewsignori-aot (approved):
Thanks for doing the fix, looks good 👍

andrepestana-aot (approved):
👍

ann-aot (approved):
nice catch 👍

Issue #2020, opened 2023-06-15
Remediate high risk vulnerabilities from WAVA report (API Mass Assignment)


andrewsignori-aot:
- Enable `forbidNonWhitelisted` on the SIMS API to force the API to return an error when the DTO received has unexpected properties.
- Change the Vue formio-related API calls to ensure that only the expected DTO properties will be sent to the API.

PR #2054, opened 2023-06-28
#2020 - WAVA  (API Mass Assignment) - Fix location creation

andrewsignori-aot:
Fix an error during location creation. The DTO was not properly typed causing an API error during the submission due to the extra properties sent in the payload.


dheepak-aot (approved):
👍

sh16011993 (approved):
👍

Issue #2055, opened 2023-06-28
Institution Missing Menu Options


andrewsignori-aot:
After a new institution is created using a business BCeID the menus are not rendered as expected due to an issue with the vue store reactivity.

Main page rendered without the menus.
![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/ca43724d-543e-458c-a231-54945c426094)

Expected page with the expected menus.

![image.png](https://images.zenhubusercontent.com/5fcebe3c076e6aaed83693fb/c33bacf8-3d1f-49e3-b384-ccd041fecb86)

hellolynn-tbtb:
DEVS demod with Hugh/Josh on June 29, 2023.

PR #2056, opened 2023-06-28
#2055 - Institution menus not displayed after creation (Business BCeID only)

andrewsignori-aot:
When a Business BCeID creates the institution the main applications menus are not displayed to allow the user to navigate in the application. 
The way the store was previously being accessed was apparently interrupting the reactivity, so even when the store was updated the computed properties were not broadcasting the changes.
The way it is right now is working and was the quick way to have it fixed for the upcoming release.

From Vue documentation (https://vuejs.org/guide/extras/reactivity-in-depth.html#how-reactivity-works-in-vue)
_When you assign or destructure a reactive object's property to a local variable, accessing or assigning to that variable is non-reactive because it no longer triggers the get/set proxy traps on the source object. Note this "disconnect" only affects the variable binding - if the variable points to a non-primitive value such as an object, mutating the object would still be reactive._

### DIFF
--- a/sources/packages/web/src/composables/institution/useInstitutionAuth.ts
+++ b/sources/packages/web/src/composables/institution/useInstitutionAuth.ts
@@ -1,4 +1,7 @@
-import { InstitutionLocationState, InstitutionUserRoles } from "@/types";
+import {
+  InstitutionUserAuthRolesAndLocation,
+  InstitutionUserRoles,
+} from "@/types";
 import { computed } from "vue";
 import { Store, useStore } from "vuex";
 import { useAuth } from "..";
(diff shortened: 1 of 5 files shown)


sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -1,50 +1,55 @@
...
-  // From institution details in store, get institution user details and authorizations.
-  const institutionUserDetails = institutionDetails.userState;
-  const authorizations =
-    institutionDetails.authorizationsState.authorizations ?? [];
-
   const { isAuthenticated } = useAuth();
-  const isAuthenticatedInstitutionUser = computed(
-    () => isAuthenticated.value && institutionUserDetails.isActive,
+  const isAuthenticatedInstitutionUser = computed(() => {
+    return isAuthenticated.value && store.state.institution.userState.isActive;
```
andrepestana-aot, line 14: 👍

andrepestana-aot (approved):
Looks good!

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
sh16011993, line 14: Nice catch 👍

sh16011993 (approved):
Thank You for the fix 👍

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -1,50 +1,55 @@
...
+    store.state.institution.authorizationsState?.authorizations.some(
+      (auth: InstitutionUserAuthRolesAndLocation) =>
+        auth.locationId === locationId,
     );
-  const [userAuthorization] = authorizations;
+
   // User type Admin | User.
-  const userType = computed(() => userAuthorization?.userType);
+  const userType = computed(() => {
+    if (!store.state.institution.authorizationsState?.authorizations) {
```
ann-aot, line 44: just a suggestion, returning `store.state.institution.authorizationsState?.authorizations` instead of lines 44 to 49 will also work, right?

ann-aot (approved):
LGTM 👍 nice catch. added  a small suggestion

guru-aot (approved):

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
dheepak-aot: userType is  single userAuthorization object and store.state.institution.authorizationsState?.authorizations is array

dheepak-aot (approved):
Great catch and thanks for doing the change.

Issue #1882, opened 2023-04-11
Recovery and backup strategy of Patroni/PostGres database in DEV/TEST and implement in PROD


guru-aot:
~Currently we have Patroni and Mongo db backup are available already, check for redis backup~

Modifying this ticket to focus on the backup strategy for the Patroni/PostGres db.

***Acceptance Criteria***
 - [x] Removed MongoDB backups from all environments.
 - [x] Verify backup is functioning
 - [x] Verify that we can restore from backup
 - [x] Establish a basic restore/rollback procedure
 - [x] Backups to retain:
- Daily: 6
- Weekly: 4
- Monthly: 1
 - [ ] Jason to do: Determine if OCIO provides off server storage of backups (Currently being stored on PV in OpenShift) - Will be continued in the next story #2067
 - [ ] Implement backups in PROD - Will be continued in the next story #2067

guru-aot:
As a part of this story verification of the backup container is not done and will be completed in the upcoming story.

PR #2057, opened 2023-06-28
#1882 - Recovery and backup startegy patroni

guru-aot:
- Updated the base image of database backup container to support till Postgres 14 - https://github.com/BCDevOps/backup-container/blob/master/docker/Dockerfile
- The base build template was changed from sample build given in https://github.com/BCDevOps/backup-container/blob/master/openshift/templates/backup/backup-build.yaml
- Removed the mongo db backup from makefile and removed in all the namespaces
- The backup.conf should be created for databases seperately, so renamed it for simsdb and refactored to reflect the changes in the config names, pvc's and containers created
- The deploy.yaml is updated with the latest from bcgov recommendations https://github.com/BCDevOps/backup-container/blob/master/openshift/templates/backup/backup-deploy.yaml
- Backup stratergy is followed by rolling backup with default bcgov recommendations on backup
- 
image


devops/Makefile
```diff
@@ -13,9 +13,9 @@ export HOST_PREFIX := $(or $(HOST_PREFIX), $$HOST_PREFIX)
 export NEW_DB := $(or $(NEW_DB), $$NEW_DB)
 export JOB_NAME := $(or $(JOB_NAME), $$JN)
 export SITE_MINDER_LOGOUT_URL := $(or $(SITE_MINDER_LOGOUT_URL), https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi)
-export BACKUP_POSTGRESQL_APP_NAME := $(or $(BACKUP_POSTGRESQL_APP_NAME), postgresql-backup)
-export BACKUP_MONGODB_APP_NAME := $(or $(BACKUP_MONGODB_APP_NAME), mongodb-backup)
-export BACKUP_CONFIGMAP_NAME := $(or $(BACKUP_CONFIGMAP_NAME), backup-conf)
+export BACKUP_POSTGRESQL_APP_NAME := $(or $(BACKUP_POSTGRESQL_APP_NAME), patroni-simsdb-backup)
+export PATRONI_SIMSDB_BACKUP_CONFIGMAP_NAME := $(or $(PATRONI_SIMSDB_BACKUP_CONFIGMAP_NAME), patroni-simsdb-backup-conf)
```
andrewsignori-aot, line 17: When this was mentioned "The backup.conf should be created for databases separately" does it means that they need to be created per database or per database type? For instance, Mongo/Postgres? In case we have a second DB on Postgres, do we need to create a second backup structure?

guru-aot: Yeah for seperate database not the database type, because individual databases can have different schedule of backups

devops/openshift/database-backup/backup-build.yaml
```diff
@@ -73,4 +71,4 @@ parameters:
     displayName: FROM Image Tag
     description: Base image to build from.  Docker creds or Artificatory setup may be needed to alleviate docker rate-limiting
     required: true
-    value: artifacts.developer.gov.bc.ca/docker-remote/centos/postgresql-12-centos7:20200917-804ef01
+    value: docker.io/centos/postgresql-12-centos7:20200917-804ef01
```
andrewsignori-aot, line 74: `artifacts.developer.gov.bc.ca/docker-remote` is not the same as `docker.io`?
Are we not supposed to be using the `artifacts` URL as much as possible (if not for everything)?

guru-aot: Yes you are right, but in this case, the build.yaml has been copied from the https://github.com/BCDevOps/backup-container/blob/master/openshift/templates/backup/backup-build.yaml. And the base image is actually not uses this too instead it uses quay.io/fedora/postgresql-14:14, as per the suggestion from the bcgov developer who did the changes to the dockerfile, which supports postgres 14.

devops/Makefile
andrewsignori-aot: So, "The backup.conf should be created for databases separately" actually means that the backups can be configured for multiple databases but we are actually structuring it in a way that it will be per database assuming that a new database will have a different backup schedule, right?

Does it mean that in the future, if we have more than one DB, instead of a new line in a config file we will create a separate structure to handle the new backup?

devops/openshift/database-backup/backup-deploy.yaml
```diff
@@ -170,16 +169,16 @@ objects:
                   value: ${MONGODB_AUTHENTICATION_DATABASE}
                 - name: TABLE_SCHEMA
                   value: ${TABLE_SCHEMA}
-                - name: ${DATABASE_SERVER_NAME}_USER
+                - name: DATABASE_USER
```
andrewsignori-aot, line 172: Are we removing the "_USER" because we will no longer support the backup for multiple DBs?
"_USER" reference: https://github.com/BCDevOps/backup-container/tree/master#backupconf

devops/Makefile
```diff
@@ -490,127 +489,72 @@ deploy-forms:
...
-	oc -n $(NAMESPACE) create configmap $(BACKUP_CONFIGMAP_NAME) --from-file=./openshift/database-backup/backup.conf
-
-db-backup-configmap-delete:
-	test -n "$(NAMESPACE)"
-	@echo "+\n++ Removing configmap resoures for database backups on $(NAMESPACE)\n+"
-	oc -n $(NAMESPACE) delete configmap/$(BACKUP_CONFIGMAP_NAME)
-
 # Create the POSTGRES structure
-db-backup-build-postgresql:
+db-backup-build-patroni:
```
andrewsignori-aot, line 493: Should this also include SIMS to be aligned with the other renamed variables?

andrewsignori-aot (commented):
Thanks for the research on this. Few comments to get moe clarity 😉

devops/Makefile
```diff
@@ -143,16 +143,15 @@ oc-build-forms: | print-status build-forms
 oc-deploy-ha-mongo: | print-status deploy-ha-mongo
 oc-deploy-forms: | print-status deploy-forms
 
+# Build backup structure
```
ann-aot, line 146: minor - preriod

devops/Makefile
guru-aot: i specifically did not add SIMS, becuase build is going to be same for all, but while deploying we can have multiple databases with different names.

devops/Makefile
guru-aot: yes you got it right, that was the plan, as individual backup stratergy for different databases.

ann-aot (commented):
Whenever available, I would like to have a walk through for my information

devops/Makefile
dheepak-aot: I went through the conversation @guru-aot @andrewsignori-aot .

@guru-aot what could be the benefit of having individual backup strategy per db?

devops/openshift/database-backup/backup-build.yaml
dheepak-aot: @guru-aot does `artifacts.developer.gov.bc.ca/docker-remote/centos/postgresql-12-centos7:20200917-804ef01` work? as artifacts.developer.gov.bc.ca/docker-remote is just a proxy. 
Correct me if wrong.

IMO only if it does not work we should start thinking about docker.io IMO.

dheepak-aot, line 74: Can we claim that current backup supports postgres 14 with this image?

devops/openshift/database-backup/patroni-simsdb-backup.conf
```diff
@@ -1,7 +1,4 @@
 postgres=patroni-master:5432/SIMSDB
-postgres=patroni-master:5432/FFA_API_DB
-postgres=patroni-master:5432/FFA_BPM_DB
```
dheepak-aot, line 3: 👍

devops/openshift/database-backup/backup-deploy.yaml
dheepak-aot: Should we decide to move away from back-up of multiple db in same backup container?

If there is a requirement in future to have a another db, or we if decide to back up redis(by any chance) current way will be easily adaptable right?

dheepak-aot:
Great work. Added some comments for my understanding and clarity.

devops/openshift/database-backup/backup-build.yaml
guru-aot: Yes it supports Postgres 14, as per the fedora postgres 14 base image

devops/Makefile
```diff
@@ -13,9 +13,10 @@ export HOST_PREFIX := $(or $(HOST_PREFIX), $$HOST_PREFIX)
 export NEW_DB := $(or $(NEW_DB), $$NEW_DB)
 export JOB_NAME := $(or $(JOB_NAME), $$JN)
 export SITE_MINDER_LOGOUT_URL := $(or $(SITE_MINDER_LOGOUT_URL), https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi)
-export BACKUP_POSTGRESQL_APP_NAME := $(or $(BACKUP_POSTGRESQL_APP_NAME), postgresql-backup)
-export BACKUP_MONGODB_APP_NAME := $(or $(BACKUP_MONGODB_APP_NAME), mongodb-backup)
```
andrepestana-aot, line 17: For my knowledge, why are we removing MongoDB from our backups? We are not doing backup for MongoDB because we can restore from de formio definitions, right?

dheepak-aot (approved):
Thanks for the walk through 👍

ann-aot (approved):
Thanks for the walk through @guru-aot 👍

andrepestana-aot (approved):
Thanks for the explanations!

andrewsignori-aot (approved):
Thanks for making this happen. Looks good 👍

Issue #1962, opened 2023-05-24
Favicon replacement


JasonCTang:
**Describe the task**
Replace vuetify favicon with government favicon.
Perhaps grab the assets from bccamping website
https://camping.bcparks.ca

![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/170519af-ffb3-4f55-b06a-e6463d4a8122)

Favicon for apple is for when you save a website on your iphone home screen and the favicon turns into an "app icon"

**Acceptance Criteria**
- [x] Replace favicon
- [x] Rename label to "StudentAid BC"
- [x] Lynn to output ico, favicon sizes

**Additional context**
- Add any other context about the task here.
- Or here

PR #2068, opened 2023-06-30
#1962 - Favicon replacement

ann-aot:
- Added new favicons and logic.
 - Rename the tab title, for now, all the page title is the same, we can customize them in the future if we want as per the router.
- @hellolynn-tbtb  will be testing the below scenarios, after merging,
image

_Screenshots_
Page tile
image
bookmark
image


sources/packages/web/src/router/index.ts
```diff
@@ -40,6 +40,8 @@ const router = createRouter({
 // When tried to return true or route location instead of next with current version of vue-router
 // results in router not working as expected on page refresh.
 router.beforeResolve(async (to, _from, next) => {
+  // For now, all the page title is same, we can customize in future, if we want.
+  document.title = "StudentAid BC";
```
andrepestana-aot, line 44: I was wondering if that should come from any constant and be used in BCLogo.vue too for example. Thoughts?

guru-aot (approved):
LGTM, nice work @ann-aot

sh16011993 (approved):
👍

sources/packages/web/src/router/index.ts
ann-aot: mm, actually I was under the assumption that this is for now and we may be customizing each router title in the future as usual.

@hellolynn-tbtb can you confirm the future plan?

dheepak-aot: I don't have hard feeling to use a constant. I'm ok either way.

dheepak-aot (approved):
👍

sources/packages/web/src/router/index.ts
andrewsignori-aot: If there is no intention now to be specific per portal why we cant just set the title in the index.html?
`<title>StudentAid BC</title>`

andrepestana-aot (approved):
Looks good! I just commented about the constant but that is minor. Either way is ok. Thanks!

andrewsignori-aot (approved):
👍

sources/packages/web/src/router/index.ts
hellolynn-tbtb: Although I like it - likely low priority anyway for a specified page title. Let's go with something simple and what we know for now (e.g. StudentAid BC). My two cents. Thanks!

ann-aot: ok... for now going with `title` tag

Issue #1658, opened 2023-01-12
Address fields display for institutions and ministry users


JasonCTang:
**Critical?**
- [x] Yes - we need it for august launch
- [ ] No - it can wait after august launch

When a new instittuion location is created, the text entered into the Address Line 2 field is being merged into the Address Line 1 column leaving the Address Line 2 column blank when it is displayed in the Manage Locations for institutions.  We are going to leave Address Line 2 merged in the Address Line 1 field and it will be the second line of the address; between the actual line 1 and line 3 (city and postal code).  

The heading for Address line 1 will be changed to just Address and the Address Line 2 column will be completely removed.
Will need to make the same changes for the institution location viewed as a Ministry user as well.  

**Additional Information**
- There is also a bug where both the values of Address Line 1 and Address Line 2 are concatenated to the city and postal code in line 3.  This duplication needs to be removed so there is only city and postal code on line 3.

**Acceptance Criteria**
- [x] Rename Address Line 1 column heading to just say Address
- [x] Remove Address Line 2 column completely
- [x] Fix line 3 so that it only displays City and postal code
- [x] Ensure Address Line 2 value is displayed on line 2

**Additional context**
![image.png](https://images.zenhubusercontent.com/6047d6ac906f4ec904f7286c/aa5a8755-6ab5-4710-8f56-78b54cecadab)

PR #2070, opened 2023-06-30
#1658 - Address fields display for institutions and ministry users

andrewsignori-aot:
- Institution locations address formatted as below. The same component is shared between Mnistry and Institutions so the change affects both places.
- Removed column for address line 2.
- Added `orderBy` to the query returning locations because every time that one item was edited they kept switching orders.

image

### DIFF
--- a/sources/packages/web/src/components/common/LocationSummary.vue
+++ b/sources/packages/web/src/components/common/LocationSummary.vue
@@ -23,11 +23,12 @@
     </template>
     <content-group
       v-for="item in institutionLocationList"
+      class="mb-4"
       :key="item"
       data-cy="institutionLocation"
     >
       <v-row>
-        <v-col md="10">
+        <v-col>
           <div>
             <v-icon icon="mdi-map-marker-outline"></v-icon>
             <span data-cy="locationName" class="category-header-medium mx-1">{{
@@ -58,25 +59,18 @@
         </v-col>
       </v-row>
       <v-row>
-        <!-- Address 1 -->
+        <!-- Address -->
         <v-col>
-          <title-value propertyTitle="Address line 1" />
+          <title-value propertyTitle="Address" />
           <span
             class="label-value muted-content clearfix"
-            v-for="addressLine in addressList1(item)"
+            v-for="addressLine in addressList(item)"
             :key="addressLine"
             data-cy="institutionAddress1"
           >
             {{ addressLine }}
           </span>
         </v-col>
-
-        <!-- Address 2 -->
-        <v-col>
-          <title-value propertyTitle="Address line 2" />
-          <span data-cy="institutionAddress2">---</span>
-        </v-col>
-
         <!-- Primary contact -->
         <v-col>
           <title-value propertyTitle="Primary contact" />
@@ -129,7 +123,7 @@ export default defineComponent({
     },
   },
   setup(props) {
-    const formatter = useFormatters();
+    const { getFormattedAddressList } = useFormatters();
     const router = useRouter();
     const clientType = computed(() => AuthService.shared.authClientType);
     const goToAddNewLocation = () => {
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -115,6 +115,40 @@ export function useFormatters() {
     return formattedAddress.join(", ");
   };
 
+  /**
+   * Format an address in the expected order where each expected
+   * address component represents an entry in the array.
+   * @param address address to be formatted.
+   * @returns string array with ordered address components.
+   * @example
```
andrepestana-aot, line 123: 👌

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -115,6 +115,40 @@ export function useFormatters() {
...
+   */
+  const getFormattedAddressList = (address: Address): string[] => {
+    const formattedAddress: string[] = [];
+    // Address line 1
+    formattedAddress.push(address.addressLine1);
+    if (address.addressLine2) {
+      // Address line 2
+      formattedAddress.push(address.addressLine2);
+    }
+    // Country, province, postal code.
```
andrepestana-aot, line 139: `City` ?

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -115,6 +115,40 @@ export function useFormatters() {
...
+    const cityPostalCountry: string[] = [];
+    cityPostalCountry.push(address.city);
+    if (address.provinceState) {
```
andrepestana-aot, line 142: I see that `provinceState` is optional in `Address`. Shouldn't it be mandatory?

andrepestana-aot (approved):
Looks good. Just minor comments.

sources/packages/web/src/components/common/LocationSummary.vue
```diff
@@ -58,25 +59,18 @@
...
+            v-for="addressLine in addressList(item)"
             :key="addressLine"
             data-cy="institutionAddress1"
           >
             {{ addressLine }}
           </span>
         </v-col>
-
-        <!-- Address 2 -->
-        <v-col>
```
ann-aot, line 75: for my info, so, there won't be address 2 ?

sources/packages/web/src/composables/useFormatters.ts
andrewsignori-aot: It is mandatory because the addressees outside Canada when we do not capture this information.

sources/packages/web/src/components/common/LocationSummary.vue
andrewsignori-aot: Not as a separate column. It was supposed to be the "Address Line 2" (if I am not wrong).

ann-aot (approved):
👍 LGTM. just a small question

sources/packages/web/src/composables/useFormatters.ts
andrewsignori-aot: Adjusted.

sources/packages/web/src/components/common/LocationSummary.vue
ann-aot: mm, actually I think, the old plan was that there can be 2 addresses for the institution (although we don't have anything in place to capture the 2nd address now), `Address 1` and `Address 2`

dheepak-aot (approved):
👍

sources/packages/web/src/components/common/LocationSummary.vue
andrewsignori-aot: Yes, maybe that also. There was a plan for a sort of `home address`/`mailing address` but it was dropped a long time ago.
Either way, seems to be pretty simple now.

Issue #2074, opened 2023-07-04
Forbidden 403 error on clicking Search Student in Public Institution Portal


sh16011993:
**Describe the Bug**
Forbidden User 403 error is shown when clicking on Search Student in Public Institution Portal.

**Expected Behaviour**
Search Student page should show up as expected.

**Actual Behaviour**
Forbidden User 403 error is shown.

**Error**
![image.png](https://images.zenhubusercontent.com/63e1969c3f013ebfc60703f3/9bb8a1b8-bb84-4d3d-9223-ef6e17b23c11)

** Steps To Reproduce**
1. Login to Public Institution Portal.
2. Click on 'Search Student'.
3. Forbidden User 403 error shows up.

The store reference was wrong at this part of the code and due to the same, useInstitutionAuth was always returning undefined for isBCPublic.

![image](https://github.com/bcgov/SIMS/assets/54600590/2ffef871-e6f0-4eb0-ba41-b1bc6321f50e)

the store refernce must be `store.state.institution.institutionState?.isBCPublic`

PR #2075, opened 2023-07-04
#2074 - Fix the store reference

dheepak-aot:
The store reference was wrong at this part of the code and due to the same, useInstitutionAuth was always returning undefined for isBCPublic.

image

It has been fixed and it is working now.

image


sh16011993 (approved):
Thank You for the fix @dheepak-aot 👍

ann-aot (approved):
👍 nice catch

andrewsignori-aot (approved):
Nice catch, thanks for the fix.

guru-aot (approved):
LGTM

Issue #2022, opened 2023-06-16
Form.io definitions cache update for releases


andrewsignori-aot:
Currently, the form.io definitions downloaded by the Vue/Web application are kept in the browser session storage. When a new deployment happens it will not be refreshed automatically, it only happens when the browser is closed.

- [x] When a new deployment happens we should force the web application to download again the form.io definition. One suggestion would be to make available the version in the Vue application and use it as part of the storage key ([source reference](https://github.com/bcgov/SIMS/blob/86dd260e23f3878a149878b28f7b62a403a10473/sources/packages/web/src/components/generic/formio.vue#L97)), in this way, when a new application is deployed all the form.io storage cache will be "invalidated".

PR #2084, opened 2023-07-05
#2022 - Form.io definitions cache update for releases

ann-aot:
- version added to form name
_screenshots_
image

### DIFF
--- a/sources/packages/backend/libs/utilities/src/config/config.service.ts
+++ b/sources/packages/backend/libs/utilities/src/config/config.service.ts
@@ -247,7 +247,8 @@ export class ConfigService {
   }
 
   /**
-   * Api version.
+   * Gets the current API version.
+   * @returns current API version.
    */
   get getAPIVersion(): string {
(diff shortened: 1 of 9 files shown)


sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -245,4 +245,12 @@ export class ConfigService {
       process.env.KEYCLOAK_AUTH_URL,
     ).href;
   }
+
+  /**
+   * Gets the current API version.
+   * @returns current API version.
+   */
+  get getAPIVersion(): string {
```
dheepak-aot, line 253: It is already a getter. So the name could be just apiVersion. `get apiVersion()`

dheepak-aot (approved):
Nice work @ann-aot . One minor comment.

guru-aot (approved):
LGTM

sh16011993 (approved):
LGTM @ann-aot  👍 . Nice work. Thank You for the doubt clarification via code walkthrough.

PR #2098, opened 2023-07-10
#2042 - Review axios timeout configuration

ann-aot:
- Implemented nestjs HTTP service
- Added `5000ms` as timeout
-  HTTP service returns an `Observable` object. https://docs.nestjs.com/techniques/http-module#full-example.
- `firstValueFrom`  converts an observable to a promise  https://rxjs.dev/api/index/function/firstValueFrom.
-  Imported `HttpModule` as global, in order to avoid the import in all feature modules. (https://stackoverflow.com/questions/71742545/nestjs-http-module-needs-to-be-imported-in-every-feature-module)
-  Latest version of `@nestjs/axios`  is `3.0.0`, but I used `1.0.1` because of Axios and soap dependency (as the latest soap is compatible with axios@"^0.27.2" and lower )
image
image
image

_Screenshots_,
image

- FEEDBACKS
1. I forcefully upgraded the axios@3. did a smoke test, looks like it's working fine
2. removed `firstValueFrom` and used `httpService.axiosRef`


sources/packages/backend/apps/api/src/services/auth/keycloak/keycloak.service.ts
```diff
@@ -48,7 +53,9 @@ export class KeycloakService {
    */
   public async getOpenIdConfig(): Promise<OpenIdConfig> {
     try {
-      const response = await axios.get(this.authConfig.openIdConfigurationUrl);
+      const response = await firstValueFrom(
```
sh16011993, line 56: We can directly get a promise using: `httpService.axiosRef.get` instead of getting an Observable and then converting it into a `AxiosResponse` promise. Currently implemented way is fine too.

andrewsignori-aot, line 56: I got the point of using the `firstValueFrom`, just wondering if using the axiosRef directly would not be acceptable and leave the usage of `rxjs` if it is needed.

sh16011993: Something like below:

`this.httpService.axiosRef.get(
        this.authConfig.openIdConfigurationUrl,
);`

sources/packages/backend/apps/queue-consumers/src/queue-consumers.module.ts
```diff
@@ -56,9 +56,18 @@ import { StudentAssessmentService, ApplicationService } from "./services";
 import { SFASIntegrationModule } from "@sims/integrations/sfas-integration";
 import { ATBCIntegrationModule } from "@sims/integrations/atbc-integration";
 import { ECEIntegrationModule } from "@sims/integrations/institution-integration/ece-integration";
+import { HttpModule } from "@nestjs/axios";
 
 @Module({
   imports: [
+    {
+      ...HttpModule.register({
```
andrewsignori-aot, line 64: Same from https://github.com/bcgov/SIMS/pull/2098/files#r1260032489

sources/packages/backend/apps/api/src/app.module.ts
```diff
@@ -21,9 +21,18 @@ import { ConfigModule } from "@sims/utilities/config";
 import { DatabaseModule } from "@sims/sims-db";
 import { NotificationsModule } from "@sims/services/notifications";
 import { QueueModule } from "@sims/services/queue";
+import { HttpModule } from "@nestjs/axios";
 
 @Module({
   imports: [
+    {
```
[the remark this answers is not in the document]
ann-aot: no, Andrew, there were 2methods to make it global, I chose it as it can be directly added to the parent module itself, but I can do it with decorator

sources/packages/backend/apps/api/src/app.module.ts
```diff
@@ -21,9 +21,18 @@ import { ConfigModule } from "@sims/utilities/config";
...
+      ...HttpModule.register({
+        // `timeout` specifies the number of milliseconds before the request times out.
+        // If the request takes longer than `timeout`, the request will be aborted.
+        timeout: 5000,
```
andrewsignori-aot, line 32: I believe that this value should be at least 30 seconds. Is there a best practice around it? Either way, 5 seconds does not seem too much.

sources/packages/backend/apps/queue-consumers/src/queue-consumers.module.ts
```diff
@@ -56,9 +56,18 @@ import { StudentAssessmentService, ApplicationService } from "./services";
...
+        // `timeout` specifies the number of milliseconds before the request times out.
```
andrewsignori-aot, line 65: I would remove the comments here.

sources/packages/backend/apps/api/src/services/auth/keycloak/keycloak.service.ts
ann-aot: thanks for that, I think, I overlooked it

ann-aot: https://github.com/bcgov/SIMS/pull/2098#discussion_r1260039266

andrewsignori-aot (commented):
Thanks for doing the investigation, great work, and there are a few comments to be discussed.
If the soap library is the only one holding the Axios upgrade, our SOAP usage is pretty limited and it is not hard to check if there will be some compatibility issue if we force the Axios upgrade. Can we evaluate the option?

sources/packages/backend/apps/api/src/app.module.ts
ann-aot: mm, Actually,  when I was looking for an ideal timeout, I got mixed answers, and most of the articles that I looked were saying 5s, that's the reason I added 5sec. I can change it.

sh16011993 (commented):
Thank You for the nice work 👍

sources/packages/backend/apps/api/src/app.module.ts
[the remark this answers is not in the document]
andrewsignori-aot: The main difference is that creating the new module will have it centralized and it will be easier to be consumed, IMO.

sources/packages/backend/apps/api/src/app.module.ts
dheepak-aot, line 32: Shouldn't we have this coming from a config(an env property) ? I get that Idea that we are trying to keep it simple but I felt it must be something which could be modified at environment level when needed.

sources/packages/backend/apps/api/src/app.module.ts
```diff
@@ -15,7 +15,7 @@ import { AppInstitutionsModule } from "./app.institutions.module";
 import { ClientTypeBaseRoute } from "./types";
 import { AppStudentsModule } from "./app.students.module";
 import { AppSupportingUsersModule } from "./app.supporting-users.module";
-import { ZeebeModule } from "@sims/services";
+import { GlobalHttpModule, ZeebeModule } from "@sims/services";
```
dheepak-aot, line 18: Just because it is global, the name starts with Global?, I would suggest AppHttpModule.

dheepak-aot:
Great Job! Nice to see us using  NestJS wrapper for http. Just 2 comments.

sources/packages/backend/apps/api/src/app.module.ts
andrewsignori-aot: IMO the problem is that HttpModule is already taken and we are just creating a wrapper around it. Since it can be also used in some library modules, I am not sure if it makes sense to have it prefixed with "App".

sources/packages/backend/apps/api/src/app.module.ts
andrewsignori-aot: To be clear, did you find 5s as a recommendation to be followed or just as an example for those articles?
If there was a recommendation to be 5s, can you please share one article?

andrewsignori-aot: As it was not defined in the ticket I refrain from mentioning it. If you are going ahead with this please keep in mind that the new ENV variable should be referenced on all PODs using the HTTP module.

sh16011993 (approved):
Nice work. Thank You for making the change. LGTM 👍

sources/packages/backend/apps/api/src/app.module.ts
ann-aot: so keeping as it is

dheepak-aot: We can create a small ticket for it(using env to be used in local/sources/devops)?

ann-aot: no @andrewsignori-aot I didn't find any recommendation, it was more like examples, I have updated it to 30s

dheepak-aot (approved):
LGTM 👍

ann-aot:
> Thanks for doing the investigation, great work, and there are a few comments to be discussed. If the soap library is the only one holding the Axios upgrade, our SOAP usage is pretty limited and it is not hard to check if there will be some compatibility issue if we force the Axios upgrade. Can we evaluate the option?

I forcefully upgraded the `axios@3`. did a smoke test, looks its working fine

guru-aot (approved):
LGTM, Nice work @ann-aot

sources/packages/backend/apps/api/src/app.module.ts
andrewsignori-aot: I would not mind having it exposed as an ENV. Can be in a separated ticket.

ann-aot: Added it as a system config constants

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -70,3 +70,8 @@ export const MSFAA_PART_TIME_FILE_CODE = "PBC.EDU.MSFA.SENT.PT.";
  * all the time.
  */
 export const SFAS_IMPORT_RECORDS_PROGRESS_REPORT_PACE = 1000;
+
+/**
+ * Https service timeout
```
andrewsignori-aot, line 75: Period 😄

andrewsignori-aot, line 75: Please remove the s from HTTP, is may lead to a wrong interpretation.

andrewsignori-aot (approved):

Issue #2134, opened 2023-07-27
Increase size of legal operating name for institution in database


JasonCTang:
**Describe the task**
Increase size of legal operating name and operating name for institution in database.  Jason to confirm the size. 

**Acceptance Criteria**
- [ ] Increase size of legal operating name and operating namefor institution in database.

**Additional context**
- The size is currently limited to 64 characters and is throwing an unexpected error if the length is greater than that.

PR #2140, opened 2023-07-27
#2134 - Increase the max length of operating and legal operating name to 250

dheepak-aot:
## Increase the max length in formio and db for operating name and legal operating name

- [x] DB Migations
- [x] Form IO validation update.

### Institution Creation exceeding 250 characters

image

### Institution Creation within 250 characters

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/institutionprofile.json
+++ b/sources/packages/forms/src/form-definitions/institutionprofile.json
@@ -1,5 +1,5 @@
 {
-  "title": "Institution Profile",
+  "title": "InstitutionProfile",
   "name": "institutionprofile",
   "path": "institutionprofile",
   "type": "form",
--- a/sources/packages/forms/src/form-definitions/institutionprofile.json
+++ b/sources/packages/forms/src/form-definitions/institutionprofile.json
@@ -278,7 +278,7 @@
           "validate": {
             "required": true,
             "minLength": "",
-            "maxLength": "",
+            "maxLength": 250,
             "minWords": "",
             "maxWords": "",
             "pattern": "",
@@ -297,7 +297,7 @@
           "tags": [],
           "properties": {},
           "conditional": {
-            "show": null,
+            "show": "",
             "when": null,
             "eq": "",
             "json": ""
@@ -346,8 +346,7 @@
           "addons": [],
           "inputType": "text",
           "id": "eiloy9e",
-          "defaultValue": "",
-          "isNew": false
+          "defaultValue": ""
         },
         {
           "label": "Operating name",
@@ -440,8 +439,7 @@
           "addons": [],
           "inputType": "text",
           "id": "eaql8hg",
-          "defaultValue": "",
-          "isNew": false
+          "defaultValue": ""
         },
         {
           "label": "Institution type",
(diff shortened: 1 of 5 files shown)


guru-aot (approved):
LGTM, nice work @dheepak-aot

sources/packages/forms/src/form-definitions/institutionprofile.json
```diff
@@ -395,7 +396,7 @@
           "validate": {
             "required": true,
             "minLength": "",
-            "maxLength": 100,
+            "maxLength": 250,
```
andrewsignori-aot, line 399: Nice catch updating this form also.
Can we have an AC added to the ticket to make it clear that the update process by the Ministry is also a scenario to be tested by the business?
@JasonCTang FYI

andrewsignori-aot (approved):
Nice work, looks good 👍

sources/packages/forms/src/form-definitions/institutionprofile.json
```diff
@@ -346,7 +346,8 @@
           "addons": [],
           "inputType": "text",
           "id": "eiloy9e",
-          "defaultValue": ""
+          "defaultValue": "",
+          "isNew": false
```
ann-aot, line 350: just wondering, are we not adding the max length to legalOperatingName

andrewsignori-aot: We consider that @andrewsignori-aot but the `legalOperatingName` is displayed here for info only. The value is retrieved from the token directly, not from the UI, so there is no real point in adding a validation here since we have the confirmation about the 250 characters.

sources/packages/forms/src/form-definitions/institutionprofilecreation.json
```diff
@@ -858,7 +860,7 @@
           "validate": {
             "required": true,
             "minLength": "",
-            "maxLength": 100,
+            "maxLength": 250,
```
sh16011993, line 863: For my info, do we also not validate the max length for the legal operating name field?

sources/packages/forms/src/form-definitions/institutionprofile.json
dheepak-aot: Yes, the value is read from the response of bceid web service. 

image

sh16011993 (approved):
LGTM 👍 Thank You for the quick fix.

sources/packages/forms/src/form-definitions/institutionprofile.json
dheepak-aot: @andrewsignori-aot , I just thought twice, we are re-using the same form(institution profile) to create institution by ministry right. should we use the validation then?

image

ann-aot: mm, ya...  but I was wondering, what if they increased the length (and we have 250), and now there is a new institution with more than 250 characters of legal name, then the same error which we saw today will happen again right? but if we add the max length here, at least the user will not be able to submit due to validation and the user will at least know the issue right?

but, I am good with what we have in the PR, just sharing my thoughts

andrewsignori-aot: Yes, that makes sense @dheepak-aot, nice catch.

ann-aot (approved):
NIce work @dheepak-aot and nice catch

PR #2141, opened 2023-07-28
#2134 - Institution Search Fix

dheepak-aot:
## Update institution search for ministry user to allow 250 characters on operating name and legal operating name

- [x] Update the max length constants

## The institution search result seems to be broken when the characters of operating name and legal name are lengthy

image

After enabling scrollable

image

- [x] Enable scrollable as true


andrewsignori-aot (approved):
Looks good 👍

ann-aot (approved):
👍 nice catch

guru-aot (approved):
LGTM

Issue #2116, opened 2023-07-14
Request an Offering Change - Submit/View UI - Validations and E2E test


andrewsignori-aot:
Continuation of ticket #1954.

## Acceptance Criteria

- [x] UI to allow study end date in the past and banner to be displayed as per [Figma](https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?type=design&node-id=11898%3A160641&t=32AKlqR1RDHn7hCN-1).
- [x] After submitting, redirect to the "In progress" tab. The tabs must be changed to be route based to allow the redirect back.
- [x] Ensure that API is executing the same validations from PIR. Try to centralize as much as possible the validations.
    - Selected offering belongs to the location (same as PIR).
    - Selected offering do not cause overlaps with other students' applications (SIMS, SFAS, same as PIR).
    - Offering intensity (apparently not checked for PIR but apparently was checked in the past because there is a related error const).
    - Offering belongs to the application program year (nice to have for this ticket). This validation is not currently in place for PIR (or even student application submission). It would be good to be added (not a must) or have another ticket created to ensure that it would be added.
- [x] Review PIR validations and follow the same rules.
- [x] E2E test

PR #2149, opened 2023-07-31
#2116 - Request an Offering Change - Submit/View UI - Validations

ann-aot:
- Added banner for the past end date.
- Tabs are now router based and the page is redirected to the in-progress tab after successful submission.
- Centralized PIR validation logic and used in application offering submission.
- Error code and error msg in student-assessment.service.ts was incorrect and adjusted.


sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
```diff
@@ -258,21 +277,59 @@ export class ApplicationOfferingChangeRequestInstitutionsController extends Base
...
+          payload.applicationId,
+          payload.offeringId,
+          payload.reason,
+          userToken.userId,
+        );
+      return { id: applicationOfferingChangeRequest.id };
+    } catch (error: unknown) {
+      if (error instanceof CustomNamedError) {
+        switch (error.name) {
+          case APPLICATION_NOT_FOUND:
```
andrewsignori-aot, line 314: So far, the `ApiProcessError` is being used, as far as I know, to help to display expected errors to API used in a friendly way.
We handle this in two different ways.
1 - Use `ApiProcessError` when it is an expected business scenario where an API consumer (UI) needs to inform the user.
2 - Regular API exception when the error can be caused only by a user manipulating the payload unexpectedly.
To put in the current context, `APPLICATION_NOT_FOUND`, `OFFERING_PROGRAM_YEAR_MISMATCH`, `OFFERING_DOES_NOT_BELONG_TO_LOCATION` would be regular API errors instead of `ApiProcessError`.
Open to talking about maybe changing the pattern to have them all always expressed as `ApiProcessError`.

sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts
```diff
@@ -258,21 +277,59 @@ export class ApplicationOfferingChangeRequestInstitutionsController extends Base
...
+            throw new UnprocessableEntityException(
+              new ApiProcessError(error.message, error.name),
+            );
+          case OFFERING_DOES_NOT_BELONG_TO_LOCATION:
+            throw new UnauthorizedException(
+              new ApiProcessError(error.message, error.name),
+            );
+        }
+      }
+      throw new InternalServerErrorException(
```
andrewsignori-aot, line 330: We need to have the `error` logged here otherwise we will never know its content.

sources/packages/backend/apps/api/src/constants/error-code.constants.ts
```diff
@@ -127,3 +127,12 @@ export const INSTITUTION_LOCATION_NOT_VALID = "INSTITUTION_LOCATION_NOT_VALID";
...
  */
 export const DISABILITY_REQUEST_NOT_ALLOWED = "DISABILITY_REQUEST_NOT_ALLOWED";
+/**
+ * Offering program year mismatch.
+ */
+export const OFFERING_PROGRAM_YEAR_MISMATCH = "OFFERING_PROGRAM_YEAR_MISMATCH";
+/**
+ * Offering does not belong to the location.
+ */
+export const OFFERING_DOES_NOT_BELONG_TO_LOCATION =
```
dheepak-aot, line 137: Suggestion: on the same context `OFFERING_LOCATION_MISMATCH` ?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1517,6 +1521,142 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+   * overlapping study dates for the student, is the offering is
+   * matching with the existing offering and the newly selected
+   * offering belong to the same program year.
+   * @param applicationId existing application id.
+   * @param selectedOffering newly selected offering id.
+   * @param locationId location id.
+   * @param options more options of the validation.
+   * - `isPir` set true, if its PIR related.
+   */
+  async applicationOfferingValidation(
```
dheepak-aot, line 1598: As a suggestion, the method name can be a verb like `validateApplicationOffering`

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1517,6 +1521,142 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    applicationId: number,
+    selectedOffering: number,
+    locationId: number,
+    options?: {
+      isPir: boolean;
```
dheepak-aot, line 1603: `isPIR`?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1517,6 +1521,142 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    },
+  ): Promise<void> {
+    // Validate if the application exists and the location has access to it.
+    const application = options?.isPir
+      ? await this.getProgramInfoRequest(locationId, applicationId)
```
dheepak-aot, line 1608: I was under the impression that we are going to retrieve the result required for performing the common validations here. I think we are doing the non-centralized part of validation also here based on option.(e.g. Only PIR specific).

was thinking if we should keep `applicationOfferingValidation` exclusively doing common validations or being versatile like the way it is right now? @andrewsignori-aot @guru-aot @sh16011993 thoughts?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1517,6 +1521,142 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      throw new CustomNamedError(
+        "Application not found.",
+        APPLICATION_NOT_FOUND,
+      );
+    }
+    // Validates if the offering exists and belongs to the location.
+    const offering = await this.offeringService.getOfferingLocationId(
+      selectedOffering,
+    );
+    if (offering?.institutionLocation.id !== locationId) {
```
dheepak-aot, line 1621: May be we should check if offering exist before doing this validation? Because if the offeringId is not a valid one `offering not found ` would be more accurate

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1517,6 +1521,142 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        location: {
+          id: locationId,
+        },
+        applicationStatus: Not(ApplicationStatus.Overwritten),
+      },
+    });
+  }
+
+  /**
+   * Set of validations for existing application and newly
```
dheepak-aot, line 1586: Just a thought. As this class(ApplicationService) is having more than 1500 lines of code already, should we consider creating a service like `ApplicationOfferingService` and move this method there?

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/InProgress.vue
```diff
@@ -0,0 +1,20 @@
+<template>
```
dheepak-aot, line 1: The file path has `request-a-change.vue`

sources/packages/web/src/views/institution/locations/request-a-change/RequestAChange.vue
```diff
@@ -72,13 +37,37 @@ export default defineComponent({
   },
   setup(props) {
     const { getLocationName } = useInstitutionState();
-    const tab = ref(ActiveRequestAChangeTab.AvailableToChangeTab);
+    const tabItems = ref([
```
sh16011993, line 40: Nice to see the tabs implemented as routes 👍

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -170,22 +173,45 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
...
             InstitutionUserTypes.admin,
             InstitutionUserTypes.user,
           ],
         },
+        children: [
+          {
+            path: AppRoutes.RequestApplicationOfferingChangeAvailableToChange,
+            name: InstitutionRoutesConst.REQUEST_CHANGE_AVAILABLE_TO_CHANGE,
+            component: AvailableToChange,
+            props: true,
```
sh16011993, line 200: I am explicitly passing the props to the child route and it works. So, I believe the below would work:
```
props: (route) => {
  return {
    locationId: parseInt(
      route.params.locationId as string,
    ),
  };
},
```

sources/packages/web/src/router/InstitutionRoutes.ts
```diff
@@ -170,22 +173,45 @@ export const institutionRoutes: Array<RouteRecordRaw> = [
...
+          },
+          {
+            path: AppRoutes.RequestApplicationOfferingChangeInProgress,
+            name: InstitutionRoutesConst.REQUEST_CHANGE_IN_PROGRESS,
+            props: true,
```
sh16011993, line 205: https://github.com/bcgov/SIMS/pull/2149/files#r1282352951

Similar can be done for the below route as well.

sources/packages/backend/apps/api/src/route-controllers/program-info-request/program-info-request.institutions.controller.ts
```diff
@@ -122,10 +119,7 @@ export class ProgramInfoRequestInstitutionsController extends BaseController {
     // assessment always available for submitted student applications).
     // PIR process happens only during original assessment.
     // Offering available only when PIR is completed.
```
dheepak-aot, line 121: Are the comments at line 118-120 still relevant now? I thought it was describing the find original assessment code which was deleted.

sources/packages/backend/apps/api/src/utilities/program-year-utils.ts
```diff
@@ -21,3 +26,18 @@ export function getSupportingUserForm(
...
+
+/**
+ * Check if the offering belong to the program year.
+ * @param newOffering offering details.
+ * @param programYear program year details.
+ * @returns returns true if the offering belong to the program
+ * year.
+ */
+export function offeringBelongToProgramYear(
+  newOffering: EducationProgramOffering,
```
dheepak-aot, line 38: the param name could be `offering` without new.

sources/packages/backend/apps/api/src/utilities/program-year-utils.ts
```diff
@@ -21,3 +26,18 @@ export function getSupportingUserForm(
...
+  programYear: ProgramYear,
+): boolean {
+  // Check if the program start start belong to the program year.
```
dheepak-aot, line 41: offering start date?

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -304,6 +309,21 @@ export function useFormatters() {
     return `${negativeValueDisplay}$${Math.abs(currencyValue).toFixed(2)}`;
   };
 
+  /**
+   * Validates if the given date 1 is before the given date 2.
+   * @param date1 given date 1.
+   * @param date2 given date 2.
+   * @returns if the given date 1 is before the given date 2.
+   */
+  const isBeforeDate = (
```
dheepak-aot, line 318: This util method looks good.  But why do we do the format here?
We have same method slightly more generic (without doing the format ) in the API side.  May be we can just use the same way. Let me know if I am missing something.

image

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/Completed.vue
```diff
@@ -0,0 +1,20 @@
+<template>
```
dheepak-aot, line 1: https://github.com/bcgov/SIMS/pull/2149/files#r1281298269

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/AvailableToChange.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <available-to-change-summary :location-id="locationId" />
```
dheepak-aot, line 2: https://github.com/bcgov/SIMS/pull/2149/files#r1281298269

dheepak-aot:
Thanks for the great effort on centralization and adding required validations. Please have a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/program-info-request/program-info-request.institutions.controller.ts
```diff
@@ -122,10 +119,7 @@ export class ProgramInfoRequestInstitutionsController extends BaseController {
...
-    const originalAssessmentOffering = application.studentAssessments.find(
-      (assessment) =>
-        assessment.triggerType === AssessmentTriggerType.OriginalAssessment,
-    ).offering;
+    const originalAssessmentOffering = application.currentAssessment.offering;
```
andrewsignori-aot, line 122: I am not following why this one was changed. The PIR offering belongs to the "Original Assessment" only.
The current assessment may contain an offering from a reassessment that is no longer the PIR-provided offering. If I am missing something, please let me know.

sh16011993 (commented):
Nice work. Left a few comments.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -482,24 +487,23 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      .leftJoin("currentAssessment.offering", "offering")
       .leftJoin("offering.educationProgram", "educationProgram")
       .innerJoin("student.user", "user")
       .leftJoin("application.pirDeniedReasonId", "PIRDeniedReason")
       .where("application.id = :applicationId", {
         applicationId,
       })
       .andWhere("location.id = :locationId", { locationId })
-      .andWhere("studentAssessments.triggerType = :triggerType", {
+      .andWhere("currentAssessment.triggerType = :triggerType", {
```
andrewsignori-aot, line 506: I am not sure if I am following this. The PIR will have an original assessment at the start but does this query is intended to work also when a PIR assessment is reassessed?

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: On the same topic, I believe that we can have the `application` retrieve by only one method. I am not able to easily find the reason why we need two different methods. Can you please elaborate or can we have a sync up on this?

sources/packages/backend/apps/api/src/utilities/program-year-utils.ts
andrewsignori-aot, line 38: I would prefer to have the paramter cleary stating what is required, for instance
`Pick<EducationProgramOffering, "studyStartDate"` or `{studyStartDate: string}`.

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -304,6 +309,21 @@ export function useFormatters() {
...
+    date1: string | Date,
+    date2: string | Date,
+  ): boolean => {
+    return dayjs(dayjs(date1).format("YYYY-MM-DD")).isBefore(
```
andrewsignori-aot, line 322: `dayjs` has a second parameter that allows providing the expected format, something like:
`dayjs(date1, "YYYY-MM-DD")`

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/AvailableToChange.vue
andrewsignori-aot, line 2: Is `available-to-change-summary` created as a component with the intention to be shared?
If it will not be shared or its creation will not bring better code readability, should we have a component for such cases?
Same for the similar ones in the PR.

andrewsignori-aot (commented):
Great work and thanks for taking this forward. Please tale a look at the comments. There is one or another that maybe I am missing something and just need some clarification.

sources/packages/backend/apps/api/src/constants/error-code.constants.ts
ann-aot: IMO @dheepak-aot `MISMATCH` is when there is a direct comparison between two values. Let me know your thoughts devs

sources/packages/backend/apps/api/src/services/application/application.service.ts
ann-aot: I also had similar thought when I was doing this, but this validation is not specific to `Application Offering change request`, so I thought its better to do it in `application.service`

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/InProgress.vue
ann-aot: Thanks @dheepak-aot . Nice catch

sources/packages/backend/apps/api/src/route-controllers/program-info-request/program-info-request.institutions.controller.ts
ann-aot: I kept it because I felt it was giving some additional context

sources/packages/web/src/composables/useFormatters.ts
ann-aot: without formatting date is a timestamp, here the comparison is only for date,
image

sources/packages/web/src/composables/useFormatters.ts
ann-aot: @andrewsignori-aot the 2nd parameter is dayjs(), is actually for the validation check of the 1st parameter. For the format, `.format()` is to be used
https://day.js.org/docs/en/parse/is-valid#docsNav
https://day.js.org/docs/en/parse/string-format

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/AvailableToChange.vue
ann-aot: we don't have plans to share I believe. I just kept it as a component, but I can revert, as it is the only logic inside the view/

sources/packages/web/src/composables/useFormatters.ts
ann-aot: As discussed, updated comment and updated the method name

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: @dheepak-aot as per my understanding, since the method is centralized and no longer has different behavior for PIR and non-PIR, I believe that we achieved the needed centralization, right? 
It looks good to me, at least, unless I am missing something else.

sources/packages/backend/apps/api/src/route-controllers/program-info-request/program-info-request.institutions.controller.ts
andrewsignori-aot: The code was reintroduced, so I believe that the comment is meaningful again.

sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -304,6 +309,23 @@ export function useFormatters() {
     return `${negativeValueDisplay}$${Math.abs(currencyValue).toFixed(2)}`;
   };
 
+  /**
+   * Validates if the given date 1 is before the given date 2.
+   * The comparison is only done with dates, if the date has a
+   * time object, its not considered for the comparison.
```
andrewsignori-aot, line 315: Can we change it to "time object it's not considered for the comparison.", minor but changes too much the meaning IMO.

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/AvailableToChange.vue
andrewsignori-aot: I saw the change. IMO it looks better @ann-aot.

sources/packages/backend/apps/api/src/route-controllers/program-info-request/program-info-request.institutions.controller.ts
```diff
@@ -265,11 +247,18 @@ export class ProgramInfoRequestInstitutionsController extends BaseController {
...
+          case APPLICATION_NOT_FOUND:
+            throw new NotFoundException(error.message);
+          case STUDY_DATE_OVERLAP_ERROR:
           case OFFERING_INTENSITY_MISMATCH:
             throw new UnprocessableEntityException(
               new ApiProcessError(error.message, error.name),
             );
+          case OFFERING_PROGRAM_YEAR_MISMATCH:
+            throw new UnprocessableEntityException(error.message);
+
```
andrewsignori-aot, line 259: The extra line can be removed to be aligned with the other `case` .

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -649,6 +651,7 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
       ])
       .innerJoin("offerings.institutionLocation", "location")
       .where("offerings.id = :offeringId", { offeringId })
+      .where("location.id = :locationId", { locationId })
```
andrewsignori-aot, line 654: Need to be changed to `andWhere`.

andrewsignori-aot (commented):
Thanks for doing the changes. Minor comments left only.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
ann-aot: nice catch 👍

andrewsignori-aot (approved):

guru-aot (approved):
LGTM, nice work @ann-aot

sources/packages/web/src/views/institution/locations/request-a-change/request-a-change.vue/InProgress.vue
dheepak-aot: Still wondering why request-a-change is repeated in the directory names 
image

dheepak-aot:
Thanks for doing the changes, Just one minor comment https://github.com/bcgov/SIMS/pull/2149#discussion_r1287788608

dheepak-aot (approved):
LGTM 👍Thanks for explaining the comment https://github.com/bcgov/SIMS/pull/2149#discussion_r1287788608

sh16011993 (approved):
Thank You for doing the changes. Nice work 👍

Issue #2190, opened 2023-08-15
IER 12 Implementation Part 1: DB Migration Centralization and Worker Creation


dheepak-aot:
As an outcome of the analysis ticket #2122 this effort will be towards implementing the solution to save workflow data to the `sims.student_assessments` table which includes creating new column and also creating a worker to the assessment-gateway workflow to save the workflow data.

**Acceptance Criteria**
- [ ] Given (Context), When (action carried out), Then (expected outcome)
- [ ] Given (Context), When (action carried out), Then (expected outcome)

PR #2218, opened 2023-08-25
#2190 - IER 12 Implementation Part 1: DB Migration Worker and Centralization

andrepestana-aot:


sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -157,6 +159,32 @@ export class AssessmentController {
...
+      ZeebeJob<SaveWorkflowDataJobInDTO, ICustomHeaders, IOutputVariables>
+    >,
+  ): Promise<MustReturnJobActionAcknowledgement> {
+    try {
+      await this.studentAssessmentService.updateAssessmentStatusAndSaveWorkflowData(
+        job.variables.assessmentId,
+        job.variables.workflowData,
+      );
+      return job.complete();
+    } catch (error: unknown) {
```
dheepak-aot, line 180: As per the discussion today, I believe we should not add explicit try catch

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
```diff
@@ -582,32 +577,6 @@
       <bpmn:outgoing>Flow_01iq5cw</bpmn:outgoing>
     </bpmn:exclusiveGateway>
     <bpmn:sequenceFlow id="Flow_1atzf7v" sourceRef="associate-msfaa-task" targetRef="Gateway_1m2js1c" />
-    <bpmn:serviceTask id="update-noa-status-to-required-task" name="Update NOA status to &#39;Required&#39;">
```
dheepak-aot, line 585: Is there a way to make this service task not appear as modification?

Same to other properties which are not part of the Update that we made.

sources/packages/backend/apps/db-migrations/src/sql/StudentAssessments/Add-workflow-data-col.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    sims.student_assessments
+ADD
+    COLUMN workflow_data JSONB;
+
+COMMENT ON COLUMN sims.student_assessments.workflow_data IS 'Workflow details data.';
```
andrewsignori-aot, line 6: We can expand the comment as below:
"Output of workflow calculations and data used as calculations inputs. Represents workflow variables that must be persisted after the workflow is executed for easy application consumption."

sources/packages/backend/libs/utilities/src/integration-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Round a number to the nearest decimal number with
+ * the specific decimal places.
+ * @param decimalValue decimal value to be rounded.
+ * @param decimalPlaces number of decimal places to be considered
+ * for rounding.
+ * @returns if a valid number, returns the rounded number, otherwise null.
+ */
+export function decimalRound(decimalValue: number, decimalPlaces = 2): number {
```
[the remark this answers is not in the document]
andrewsignori-aot: I agree with @dheepak-aot, can we have it moved to the utilities lib?

dheepak-aot:
Nice work @andrepestana-aot . Please have a look at the comments.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.dto.ts
```diff
@@ -139,3 +140,8 @@ export interface UpdateNOAStatusHeaderDTO {
 export interface UpdateNOAStatusHeaderDTO {
   status: AssessmentStatus;
 }
+
+export interface SaveWorkflowDataJobInDTO {
+  [ASSESSMENT_ID]: number;
+  workflowData: WorkflowData;
```
andrewsignori-aot, line 146: The variable should be declared with the const `WORKFLOW_DATA` following the same from `ASSESSMENT_ID`.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
```diff
@@ -221,6 +221,15 @@ export class StudentAssessment extends RecordDataModel {
     referencedColumnName: ColumnNames.ID,
   })
   applicationOfferingChangeRequest?: ApplicationOfferingChangeRequest;
+  /**
+   * Assessment workflow data.
```
andrewsignori-aot, line 225: I made a comment adding a bit more context to this in the SQL migration. If that is changed, please change it here also.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -157,6 +159,32 @@ export class AssessmentController {
     return job.complete();
   }
 
+  /**
+   * Updates status and saves workflow data.
+   */
+  @ZeebeWorker(Workers.UpdateAssessmentStatusAndSaveWorkflowData, {
```
andrewsignori-aot, line 165: I believe that we can make this method name more generic in a way that makes it explicit that this is the worker to be called last to wrap up any workflow task before it is finally finished.
Maybe we can call it something like `workflow-wrap-up` and make a clear comment that this is the worker dedicated to be executed at the very end of the flow before the workflow reaches the `end event`.

andrewsignori-aot (commented):
Great work, great start. Please take a look at the comments.

sources/packages/backend/libs/sims-db/src/entities/student-assessment.model.ts
```diff
@@ -276,3 +285,13 @@ interface PartTimeAssessment extends BaseAssessment {
  * assessment payload created by camunda workflow.
  */
 export type Assessment = FullTimeAssessment | PartTimeAssessment;
+
+export interface WorkflowData {
```
sh16011993, line 289: We can provide a short comment here

sh16011993 (approved):
Nice work @andrepestana-aot LGTM 👍

guru-aot (approved):

sources/packages/backend/workflow/test/test-utils/constants/workflow-variables-constants.ts
```diff
@@ -24,6 +24,7 @@ export enum WorkflowServiceTasks {
   CheckIncomeRequest = "check-income-request-task",
   // Workflow - Supporting User Information Request
   CheckSupportingUserResponseTask = "check-supporting-user-response-task",
+  WorkflowWrapUp = "workflow-wrap-up",
```
dheepak-aot, line 27: workflow-wrap-up-task?

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
andrewsignori-aot: When doing the `job.fail` I believe that would be important to do some logging also, as mentioned in the previous comment, otherwise, the error log will be present only on Camunda.
We can discuss that maybe as a global review of error handling on workers. Also maybe it would be good to ensure that we do a JSON.stringnify(error, null, 2) for better formatting while concatenating the error messages.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
andrewsignori-aot: @dheepak-aot @guru-aot @sh16011993 @ann-aot FYI

andrewsignori-aot (commented):
Thanks for doing the changes, I've resolved almost all comments, just minor ones left.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -243,4 +244,17 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+   * Updates assessment status and save workflow data.
+   * @param assessmentId updated assessment.
+   * @param workflowData workflow data to be saved.
+   */
+  async updateAssessmentStatusAndSaveWorkflowData(
+    assessmentId: number,
+    workflowData: WorkflowData,
+  ): Promise<void> {
+    await this.repo.update({ id: assessmentId }, { workflowData });
+    // TODO: update assessment status.
```
ann-aot, line 258: 👍

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
andrepestana-aot: Ok, added logs as well.

andrewsignori-aot: Please use `jobLogger.error` instead.
The log message and the `job.fail` are the same, can we please move it to a `const`?
The use of the `JSON.stringify("", null, 2)` was a suggestion, if we agree to use it we need to create a small method to avoid repeating it everywhere also.

dheepak-aot (approved):
Thanks for doing the changes 👍

andrewsignori-aot (approved):

ann-aot (approved):
Good work @andrepestana-aot 👍

Issue #2207, opened 2023-08-17
Investigate Root Cause of Camunda' 'Workers Getting Stuck


michesmith:
- [ ] Replicating the issue in local  - using debugging in the worker (https://vscode.dev/github/bcgov/SIMS/blob/feature/sims-%232123/sources/packages/backend/node_modules/zeebe-node/dist/lib/interfaces-1.0.d.ts#L216) 

- [ ] Replication in dev with huge volume of load. (If not successful in replication in local)

- [ ] Check kibana logs in test if it happens again - https://kibana-openshift-logging.apps.silver.devops.gov.bc.ca/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-1h,mode:quick,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'8841c680-a15b-11eb-a4dc-e5bf19f04239',key:kubernetes.container_name,negate:!f,params:(query:test-workers-sims,type:phrase),type:phrase,value:test-workers-sims),query:(match:(kubernetes.container_name:(query:test-workers-sims,type:phrase))))),index:'8841c680-a15b-11eb-a4dc-e5bf19f04239',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc))

- [ ] Check the Camunda in test if there are any stuck workers - start rollout the deployment configs for workers in Test (This will not have any downtime for the testers)

- [ ] If we are not able to resolve the issue with workers or the load testing is successful with the required resources in workers, we may need to talk to Camunda on the resolution.

PR #2219, opened 2023-08-25
#2207 - Investigate Root Cause of Camunda Workers Getting Stuck

ann-aot:
- The issue was able to replicate in one of the workers when we submitted an application with a long income, which was an error in the worker logic during the assessment. However, the worker is not getting any acknowledgment from the code and it is waiting
- So, the next time a new assessment comes, it gets stuck in the same worker for an infinite time.
- checked our worker global error handle in https://vscode.dev/github/bcgov/SIMS/blob/fix/sims-%232207/sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts#L57
image
and found that, the `return await jobHandler(job);` is not throwing any error, when there is an error, as a result, the `catch` is not invoked and the worker is not getting the acknowledgment. 
- fixed the global handler, the jobhandler is returning an observable, so, added a logic to await for it when the return type is observable, 
image

- And added try-catch, and if there is any error in the controller level, the code acknowledges the worker as a failed job, so that the worker is free and can focus on the next jobs.
- This causes an incident in the Camunda, Which developers can take a look at (Before this fix, Camunda never caused an incident for these scenarios).

-**Replicate**
1. To replicate the issue, I decreased the `maxJobsToActivate: ` to `1` and decreased the `timeout` to `10`.
image
2. And created/edited 2 applications with a lengthy income,
image
3.  which made the `create income verification` worker stuck in both the case
image
4.  And then, when you submit an application with a proper income, the new assessment is also stuck in the same worker.
5.  After the fix, step 2 will create incidents in Camunda, and step 4, will work as expected.
Step 2, after the fix
image
image
image
Step 4, after the fix,
image

Ref: https://docs.camunda.io/docs/components/best-practices/development/writing-good-workers/#nodejs-client

### DIFF
--- a/sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
@@ -65,6 +65,7 @@ export class ApplicationController {
         jobLogger.error(message);
         return job.error(APPLICATION_STATUS_NOT_UPDATED, message);
       }
+      jobLogger.log("Updated the application status.");
       return job.complete();
     } catch (error: unknown) {
       const errorMessage = `Unexpected error while updating the application status. ${error}`;
@@ -105,6 +106,7 @@ export class ApplicationController {
       }
       if (application.applicationException) {
         // The exceptions were already processed for this application.
+        jobLogger.log("Exceptions were already processed for the application.");
         return job.complete({
           applicationExceptionStatus:
             application.applicationException.exceptionStatus,
@@ -120,10 +122,12 @@ export class ApplicationController {
             job.variables.applicationId,
             exceptions,
           );
+        jobLogger.log("Exception created.");
         return job.complete({
           applicationExceptionStatus: createdException.exceptionStatus,
         });
       }
+      jobLogger.log("Verified application exception.");
       return job.complete({
         applicationExceptionStatus: ApplicationExceptionStatus.Approved,
       });
--- a/sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
@@ -52,6 +52,7 @@ export class ApplicationController {
       >
     >,
   ): Promise<MustReturnJobActionAcknowledgement> {
+    const jobLogger = new Logger(job.type);
     try {
       const updateResult = await this.applicationService.updateStatus(
         job.variables.applicationId,
@@ -59,14 +60,13 @@ export class ApplicationController {
         job.customHeaders.toStatus,
       );
       if (!updateResult.affected) {
-        return job.error(
-          APPLICATION_STATUS_NOT_UPDATED,
-          "The application status was not updated either because the application id was not found or the application is not in the expected status.",
-        );
+        const message =
+          "The application status was not updated either because the application id was not found or the application is not in the expected status.";
+        jobLogger.error(message);
+        return job.error(APPLICATION_STATUS_NOT_UPDATED, message);
       }
       return job.complete();
     } catch (error: unknown) {
-      const jobLogger = new Logger(job.type);
       const errorMessage = `Unexpected error while updating the application status. ${error}`;
       jobLogger.error(errorMessage);
       return job.fail(errorMessage);
@@ -92,13 +92,16 @@ export class ApplicationController {
       >
     >,
   ): Promise<MustReturnJobActionAcknowledgement> {
+    const jobLogger = new Logger(job.type);
     try {
       const application = await this.applicationService.getApplicationById(
         job.variables.applicationId,
         { loadDynamicData: true },
       );
       if (!application) {
-        return job.error(APPLICATION_NOT_FOUND, "Application id not found.");
+        const message = "Application id not found.";
+        jobLogger.error(message);
+        return job.error(APPLICATION_NOT_FOUND, message);
       }
       if (application.applicationException) {
         // The exceptions were already processed for this application.
@@ -125,7 +128,6 @@ export class ApplicationController {
         applicationExceptionStatus: ApplicationExceptionStatus.Approved,
       });
     } catch (error: unknown) {
-      const jobLogger = new Logger(job.type);
       const errorMessage = `Unexpected error while verifying the application exceptions. ${error}`;
       jobLogger.error(errorMessage);
       return job.fail(errorMessage);
--- a/sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
@@ -52,18 +52,24 @@ export class ApplicationController {
       >
     >,
   ): Promise<MustReturnJobActionAcknowledgement> {
-    const updateResult = await this.applicationService.updateStatus(
-      job.variables.applicationId,
-      job.customHeaders.fromStatus,
-      job.customHeaders.toStatus,
-    );
-    if (!updateResult.affected) {
-      return job.error(
-        APPLICATION_STATUS_NOT_UPDATED,
-        "The application status was not updated either because the application id was not found or the application is not in the expected status.",
+    try {
+      const updateResult = await this.applicationService.updateStatus(
+        job.variables.applicationId,
+        job.customHeaders.fromStatus,
+        job.customHeaders.toStatus,
+      );
+      if (!updateResult.affected) {
+        return job.error(
+          APPLICATION_STATUS_NOT_UPDATED,
+          "The application status was not updated either because the application id was not found or the application is not in the expected status.",
+        );
+      }
+      return job.complete();
+    } catch (error: unknown) {
+      return job.fail(
+        `Unexpected error while updating the application status. ${error}`,
       );
     }
-    return job.complete();
   }
 
   /**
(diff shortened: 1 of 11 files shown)


andrepestana-aot (approved):
Looks good! Great job investigating the bug!

dheepak-aot:
Thanks for doing the testing @ann-aot to find out that global error handler wasn't working and we had a dev discussion this morning and agreed on the point that we must fix the global error handler instead of enforcing developer to explicitly add the try catch everywhere.

andrewsignori-aot:
> Thanks for doing the testing @ann-aot to find out that the global error handler wasn't working we had a dev discussion this morning and agreed on the point that we must fix the global error handler instead of enforcing the developer to explicitly add the try-catch everywhere.
> 
> Will continue the review after the global error handler fix.

The global error handler seems to be failing because an unhandled expectation inside the Zeebe worker method would be handled by the [rpc-exceptions-handler.ts](https://github.com/nestjs/nest/blob/master/packages/microservices/exceptions/rpc-exceptions-handler.ts) and this will return an Observable that is not expected currently.

To test the theory, if we change the global handler as below it will report the `job.fail` properly and we will able to see the `retries` counter going down and the Camunda incident generated after 3 attempts.

image

Still, if the unhandled exception is not of the type `RpcExecption` Nest will generate a generic "Internal Server Error" exception message that will be sent to Camunda. The log with the proper exception will be logged in the worker's POD but Camunda will receive only the generic message.

IMO we need to fix the global error handler and also talk about the possibility to ensure the controllers will always provide a better error handling.
@dheepak-aot @guru-aot @andrepestana-aot @sh16011993 FYI

sources/packages/backend/apps/workers/src/controllers/program-info-request/program-info-request.controller.ts
```diff
@@ -37,29 +37,33 @@ export class ProgramInfoRequestController {
...
       );
-    }
-    if (application.pirStatus) {
-      // PIR status was already set, just return it.
       return job.complete({
-        programInfoStatus: application.pirStatus,
+        programInfoStatus: job.customHeaders.programInfoStatus,
       });
+    } catch (error: unknown) {
+      return job.fail(`Unexpected error while seting the PIR status. ${error}`);
```
andrewsignori-aot, line 66: Typo on "seting".

andrewsignori-aot (commented):
Great research and findings. Please let's talk about having the global error handler fixed and also I would like to check if the default timeouts are actually woking.

sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
```diff
@@ -63,7 +64,15 @@ export class ZeebeTransportStrategy
...
     );
     try {
-      return await jobHandler(job);
+      // The return is an `Observable`for a jobHandler, which needs to be awaits.
+      // The lastValueFrom is almost exactly the same as toPromise()
+      // meaning that it will resolve with the last value that has arrived
+      // when the Observable completes.
+      const jobResult = await jobHandler(job);
+      if (isObservable(jobResult)) {
+        await lastValueFrom(jobResult);
```
andrewsignori-aot, line 73: I believe that the comment should focus more on the two possible results and why `await jobHandler(job);` can receive one or another instead of explaining the `lastValueFrom`.
As a suggestion:
"The jobHandler can potentially return a Promise or an Observable. The Promise will be returned from a controller when it finishes as expected returning some value. The Observable will be returned as a result of an unhandled exception when a RpcExecption will be generated and the RpcExpectionHandler will wrap this exception as an Observable object."

sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
```diff
@@ -85,36 +93,48 @@ export class ApplicationController {
...
+          applicationExceptionStatus: createdException.exceptionStatus,
+        });
+      }
+      jobLogger.log("Verified application exception.");
       return job.complete({
-        applicationExceptionStatus: createdException.exceptionStatus,
+        applicationExceptionStatus: ApplicationExceptionStatus.Approved,
       });
+    } catch (error: unknown) {
+      const errorMessage = `Unexpected error while verifying the application exceptions. ${error}`;
```
andrewsignori-aot, line 135: Because we would be repeating this along all controller and jobs, I was expecting that a basic helper would be created, as shown below.
```ts
} catch (error: unknown) {
  return createUnexpectedJobFail(error, job);
}
```

The helper method can be something like below.
```ts
createUnexpectedJobFail(
    error: unknown,
    job: ZeebeJob,
    logger?: Logger,
  ): MustReturnJobActionAcknowledgement {
    const jobLogger = logger ?? new Logger(job.type);
    const errorMessage = `Unexpected error while processing job. ${parseException(error)}`;
    jobLogger.error(errorMessage);
    return job.fail(errorMessage);
  }
```

As an out-of-PR-scope suggestion, I believe that we can create the below error formatter to avoid logging as `jobLogger.log(`${error.name} ${error.message}`);`.

```ts
parseException(error: unknown): string {
  return JSON.stringify(error, null, 2);
}
```

sources/packages/backend/apps/workers/src/utilities/error-handler.ts
```diff
@@ -0,0 +1,33 @@
...
+  jobLogger.error(job);
+  jobLogger.error(errorMessage);
+  return job.fail(errorMessage);
+}
+
+/**
+ * Parse the error in a prettier format for better
+ * readability.
+ */
+function parseException(error: unknown): string {
```
andrewsignori-aot, line 31: This would a generic method to be used every where, not only by workers. It can be moved to the generic lib.

andrewsignori-aot, line 31: Missing error param comment.

sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
```diff
@@ -70,20 +81,27 @@ export class SupportingUserController {
...
+      }
+      const outputVariables = filterObjectProperties(
+        supportingUser.supportingData,
+        job.customHeaders,
       );
+      jobLogger.log("Supporting user data loaded.");
+      return job.complete(outputVariables);
+    } catch (error: unknown) {
+      const errorMessage = `Unexpected error while loading supporting user data. ${error}`;
+      return createUnexpectedJobFail(errorMessage, job, jobLogger);
```
andrewsignori-aot, line 104: I was expecting that we no longer need line 103.
The helper would be called as `return createUnexpectedJobFail(error, job, jobLogger);`

sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
```diff
@@ -63,7 +64,21 @@ export class ZeebeTransportStrategy
       `Starting job for processInstanceKey ${job.processInstanceKey}. Retries left: ${job.retries}.`,
     );
     try {
-      return await jobHandler(job);
+      // The jobHandler can potentially return a Promise or
```
dheepak-aot, line 67: Thanks for comments. 👍

sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
```diff
@@ -70,20 +80,26 @@ export class SupportingUserController {
...
+        const message =
+          "Supporting user not found while checking for supporting user response.";
+        jobLogger.error(message);
+        job.error(SUPPORTING_USER_NOT_FOUND, message);
+      }
+      const outputVariables = filterObjectProperties(
+        supportingUser.supportingData,
+        job.customHeaders,
       );
+      jobLogger.log("Supporting user data loaded.");
```
dheepak-aot, line 99: Why didn't we use logger service like other places? 

image

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -152,11 +171,17 @@ export class AssessmentController {
...
+    try {
+      await this.studentAssessmentService.updateNOAApprovalStatus(
+        job.variables.assessmentId,
+        job.customHeaders.status,
+      );
+      jobLogger.log("NOA status updated.");
+      return job.complete();
+    } catch (error: unknown) {
+      return createUnexpectedJobFail(error, job, jobLogger);
+    }
```
dheepak-aot, line 184: We can refactor the error handling for the worker committed by @andrepestana-aot yesterday.

image

sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
andrewsignori-aot: This would create a logger at the "class" level, which would be okay also, but I believe that the jobs can produce their own logs.
For instance, in the way that it is right now, it puts the `job.name` in the format below (in yellow).
image

sources/packages/backend/apps/workers/src/utilities/error-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param error error message to be logged.
+ * @param job zeebe job.
+ * @param logger logger object if any,
+ * else a new logger instance will be created.
+ * @returns a job fail acknowledgement to the worker.
+ */
+export function createUnexpectedJobFail(
+  error: unknown,
+  job: ZeebeJob,
+  logger?: Logger,
```
dheepak-aot, line 17: options ?

dheepak-aot:
Great work @ann-aot  and great team effort putting our heads together to resolve this issue. Added few minor comments.

sources/packages/backend/libs/utilities/src/parse-json.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Parse the json error in a prettier format for better
+ * readability.
+ * @param value to be parsed.
+ * @returns stringify JSON object is a prettier format.
```
sh16011993, line 5: How about `json string in a prettier format`?

dheepak-aot (approved):
Thanks for making the changes 👍

andrewsignori-aot (approved):

guru-aot (approved):
LGTM

PR #2237, opened 2023-08-31
initial-commit

guru-aot:


devops/openshift/web-deploy.yml
```diff
@@ -97,6 +94,17 @@ objects:
...
       targetCPUUtilizationPercentage: 80
+  - apiVersion: policy/v1
+    kind: PodDisruptionBudget
+    metadata:
+      name: ${NAME}-pdb
+    spec:
+      selector:
+        matchLabels:
+          app: ${NAME}
+      minAvailable: 2
```
dheepak-aot, line 105: Correct me if wrong, I see the kubernetes documentation suggesting differently for frontends.

image

devops/openshift/api-deploy.yml
```diff
@@ -240,6 +237,17 @@ objects:
...
       targetCPUUtilizationPercentage: 80
+  - apiVersion: policy/v1
+    kind: PodDisruptionBudget
+    metadata:
+      name: ${NAME}-pdb
+    spec:
+      selector:
+        matchLabels:
+          app: ${NAME}
+      minAvailable: 2
```
dheepak-aot, line 248: As per openshift/kubernetes, `maxUnavailable` and `minAvailable` are mutually exclusive. Should we use both?

image

guru-aot: Yes, we can use both minAvailable and maxUnavailable in a PodDisruptionBudget, in our current scenario we set minAvailable to 2 and maxUnavailable to 1, it ensures there are always at least 2 pods available, and only 1 pod can be undergoing disruption at any given time.

devops/openshift/web-deploy.yml
guru-aot: In the documentation you have shared the minAvailable should not be reduced by the serving capacity by more than 10%. In our scenario the HorizontalPodAutoscaler we have a maxReplica of 10 and PDB - minAvailable as 2, this means the overall reducing capacity is still at 20% which actually follows the documentation.

devops/openshift/api-deploy.yml
dheepak-aot: Not sure if my understanding is right, but document says only one of the properties can be used for a single pdb.

ann-aot (approved):
👍

andrepestana-aot (approved):
Looks good. Thank you for the explanation.

devops/openshift/web-deploy.yml
```diff
@@ -105,9 +112,9 @@ parameters:
   - name: SERVICE_NAME
     value: web
   - name: CPU_LIMIT
-    value: "0.5"
+    value: "0.25"
   - name: MEMORY_LIMIT
-    value: "256M"
+    value: "512M"
```
andrewsignori-aot, line 117: SIMS-Api will do way more processing than the Web POD, why we would justify more memory allocation to Web than API?

devops/openshift/api-deploy.yml
```diff
@@ -240,6 +237,16 @@ objects:
...
+  - apiVersion: policy/v1
+    kind: PodDisruptionBudget
+    metadata:
+      name: ${NAME}-pdb
+    spec:
+      selector:
+        matchLabels:
+          app: ${NAME}
+      maxUnavailable: 1
+      disruptionsAllowed: 1
```
andrewsignori-aot, line 249: Do you mind elaborating on how the `disruptionsAllowed` configuration affects the PDB?

devops/openshift/database/redis-ha-deploy.yml
```diff
@@ -148,6 +148,17 @@ objects:
...
           storageClassName: ${STORAGE_CLASS}
+  - apiVersion: policy/v1
+    kind: PodDisruptionBudget
+    metadata:
+      name: ${NAME}-pdb
+    spec:
+      selector:
+        matchLabels:
+          app: ${NAME}
+      minAvailable: 6
```
andrewsignori-aot, line 159: Do we have a source of recommendation for Redis PDB configuration?

devops/openshift/database/patroni-deploy.yml
```diff
@@ -182,6 +182,17 @@ objects:
         resources:
           requests:
             storage: ${PVC_SIZE}
+- apiVersion: policy/v1
```
andrewsignori-aot, line 185: Do we have a source of recommendation for Patroni PDB configuration?

devops/openshift/database/patroni-deploy.yml
```diff
@@ -182,6 +182,17 @@ objects:
...
+  kind: PodDisruptionBudget
+  metadata:
```
andrewsignori-aot, line 187: How we would apply those changes to PROD? Is there a plan for it?

andrewsignori-aot (commented):

devops/openshift/web-deploy.yml
guru-aot: initially we had only one pod API with more memory, but due to the change in replication controller to have max replicas of 10 we changed it to smaller numbers. In the case of web, even though there are more processing done by API, the web had to render the application faster, so for safer side, I had the values in the vertical pod bumped up.

devops/openshift/api-deploy.yml
guru-aot: https://kubernetes.io/docs/tasks/run-application/configure-pdb/

image

disruptionsAllowed is a status configuration, I thought i can configure them. Removing it now.

devops/openshift/web-deploy.yml
andrewsignori-aot: The POD will not render anything, it will just allow the download of static files (HTML, js, CSS, ect.). I did not follow the explanation.

devops/openshift/api-deploy.yml
andrewsignori-aot: I was not sure about what it was after reading some docs that why I asked 😉

devops/openshift/database/redis-ha-deploy.yml
guru-aot: As all the databases we use are statefulsets and I followed the documentation below
https://kubernetes.io/docs/tasks/run-application/configure-pdb/#identify-an-application-to-protect

The numbers I have put in there are values I have like maxUnavailable for each pods, when the disruption happens.

devops/openshift/database/patroni-deploy.yml
guru-aot: https://github.com/bcgov/SIMS/pull/2237#discussion_r1313588539

devops/openshift/database/patroni-deploy.yml
guru-aot: The idea is to deploy the PDB configuration in our namespace manually, than running the yaml.

devops/openshift/database/redis-ha-deploy.yml
andrewsignori-aot: Just wondering if there is a recommendation from Gov or if we can get something from RocketChat.

devops/openshift/database/patroni-deploy.yml
andrewsignori-aot: https://github.com/bcgov/SIMS/pull/2237#discussion_r1313591274

andrewsignori-aot (approved):
Thanks for making the changes. I am approving the PR but I would like to have some extra info about the Redis/Patroni configurations, for instance, a double check on how gov in general have it configured. Can we do it?

dheepak-aot (approved):
Thanks for the explanations. Still have some clarification for my understanding for which I will sync up later.

Issue #88, opened 2021-01-22
Analysis to Migrate the Vue Server


andrewsignori-aot:
Currently, the Vue static files are being served using a basic node server that can be checked under the Dockerfile on the Web package.
The idea is to replace it with an NGINX (https://www.nginx.com/) server or similar.
- [ ] Conduct Analysis 
- [ ] if it works - do it 
- [ ] if not create a ticket to do it

PR #2284, opened 2023-09-08
#88 - Analysis to Migrate the Vue Server

andrepestana-aot:
Using npm `serve` package or vue-cli to host a Vue.js app in a production environment is generally not recommended for several important reasons like limited performance, security concerns, lack of Load Balancing, low customization, inefficient static file serving, bad redundancy and scalability and lack of features for monitoring and logging.

## Choosing a web server

According to w3techs and Netcraft websites Nginx is the most used web server at the moment:
image

Among the benefits of adopting Nginx for the project the items below can be highlighted:

- High Performance: Nginx is renowned for its high-performance capabilities. It is designed to handle a large number of concurrent connections efficiently and can serve static content quickly. This makes it well-suited for websites with high traffic and performance demands.
- Low Resource Usage: Nginx is known for its efficiency in using system resources. It consumes significantly less memory compared to some other web servers like Apache, which can be especially beneficial when dealing with limited server resources or a large number of simultaneous connections.
- Security: Nginx offers several security features, including DDoS protection, rate limiting, SSL/TLS termination, and the ability to restrict access to certain resources or IP addresses. Its modular architecture allows for easy integration with security solutions and Web Application Firewalls.
- Community and Support: Nginx has a large and active user community. You can find a wealth of documentation, tutorials, and community-contributed modules to extend its functionality. Commercial support is also available from Nginx, Inc. for those who need it.
- Caching: Nginx provides robust caching capabilities, which can significantly improve website performance by serving cached content to users, reducing the load on your backend servers.
- Ease of Configuration: Nginx's configuration syntax is straightforward and easy to work with. It allows you to set up complex configurations with relative ease. Additionally, it can reload configurations without downtime, making it suitable for production environments.
- Cost-Efficiency: Nginx is open-source and free to use.

During the analysis it's been noticed an increasing number of [projects](https://github.com/search?q=org%3Abcgov+caddy&type=code&p=2) using [Caddy Server](https://caddyserver.com/) among the BCGov projects.
Although there isn't any direction to use Nginx in BC Gov projects, it's widely used and, as mentioned above, there are several reasons why we should use Nginx in SIMS web app.

## Docker build recommendations:
- 2 step build process (using node for build-stage and a new image with nginx + built files). This way the resulting image will not include node, npm etc.
- Using "daemon off;" to make nginx to work in a foreground process letting Docker to monitor it and kill the container in case the process dies.

## Configuration
The configuration for the webserver is setup on [nginx.conf](https://github.com/bcgov/SIMS/pull/2284/files#diff-1abef20a58ec56ef3062ccadeb861ba575b5d5db7bd8c3666ce6efc4704601ec) and it is set with the suggested conf from vue-cli docs and a few tweaks. A full example with configuration options can be found [here](https://www.nginx.com/resources/wiki/start/topics/examples/full/).
To use variables in nginx configuration it's needed to use [`envsubst` and templating](https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration-new-in-119).

## Notes
- Sonar complaint about the image running as root user should be ignored as it is only for Dockerfile.dev and shouldn't affect production env.
- While running the Vuejs app on Vue CLI [it sets running mode to "development" as default](https://github.com/vuejs/vue-cli/blob/eaa2b7341f174260f6ebc2345bd8e838f85a2ea3/packages/%40vue/cli-service/lib/Service.js#L123). When running the web app on Nginx in a local machine, setting the mode is necessary. Otherwise the `process` is not set and the `basesUrl` for the app is [set as in "production" mode](https://github.com/bcgov/SIMS/blob/main/sources/packages/web/src/services/http/common/HttpClient.ts#L4).

## Next steps:
- Configure the server_name to only accept requests with the appropriate hostname. For that it's needed to set a "catch-all" server configuration and block it: https://serverfault.com/a/946623
E.g.:
```
server {
  listen ${PORT} default_server;
  server_name _;
  # Configuration for the default server block
  return 404;
}
server {
  listen       ${PORT};
  server_name  dev.hostname-example.com hostname-example.com;
  location / {
    root   /app;
    index  index.html;
    try_files $uri $uri/ /index.html;
  }
}
```
- Add Headers required by Web App Vulnerability Scan and recheck it through a new Wava scan.

## References
- https://w3techs.com/technologies/overview/web_server
- https://www.netcraft.com/blog/january-2023-web-server-survey/
- https://stackoverflow.com/questions/25970711/what-is-the-difference-between-nginx-daemon-on-off-option
- https://docs.nginx.com/nginx/
- https://cli.vuejs.org/guide/mode-and-env.html#modes
- https://cli.vuejs.org/guide/deployment.html#docker-nginx
- https://github.com/search?q=org%3Abcgov+nginx.conf&type=code&p=2
- https://www.nginx.com/resources/wiki/start/topics/examples/full/

### DIFF
--- a/sources/Makefile
+++ b/sources/Makefile
@@ -153,6 +153,11 @@ api: ## <Helper> :: Executes into the workspace container.
 	@echo "+\n++Shelling into local workspace ...++\n+"
 	@docker-compose exec api bash
 
+# Local Web (nginx)
+web-local:
+	@echo "+\n++ Make: Run local web (nginx).\n+"
+	@docker-compose -f docker-compose.yml up web
+
 local-db:
 	@echo "Make: Shelling into database workspace ..."
 	@docker-compose -f docker-compose.yml exec postgres psql -U $(POSTGRESQL_USER) -W $(POSTGRES_DB)
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -1,27 +1,29 @@
 FROM registry.access.redhat.com/ubi8/nodejs-16:1-5 AS builder
 
+# Application Port.
+ENV PORT 3030
 
 COPY ./package*.json ./
 RUN npm ci
 COPY . ./
+
+# Replace ${PORT} variable in the template and save as default.conf.
+RUN sed 's/${PORT}/'"${PORT}"'/g' default.conf.template > default.conf
+
 RUN npm run build
 
 FROM registry.access.redhat.com/ubi8/nginx-122 AS deployer
 
-RUN mkdir /app
-
-WORKDIR /app
+WORKDIR /opt/app-root/src
 
 # Copy the app built on build-stage to the resulting container.
 COPY --from=builder /opt/app-root/src/dist .
 
+# Copy the default.conf to /etc/nginx/conf.d
+COPY --from=builder /opt/app-root/src/default.conf /etc/nginx/conf.d
+
 # Copying the main configuration file.
 COPY nginx.conf /etc/nginx/nginx.conf
-
-# Copying the configuration template file to templates dir. 
-# It will be processed to transform the vars in the template files into 
-# evaluated values and save the resulting file as /etc/nginx/conf.d/default.conf.
-COPY default.conf.template /etc/nginx/templates/
 
 # Daemon off makes nginx to run on the foreground with only one process.
 # Docker will kill the container if the process dies.
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -6,10 +6,16 @@ RUN npm ci
 COPY . ./
 RUN npm run build
 
-FROM registry.access.redhat.com/ubi8/nodejs-16:1-5 as deployer
+FROM registry.access.redhat.com/ubi8/nginx-120 as deployer
 
-RUN npm i -g serve
-COPY --from=builder /opt/app-root/src/dist .
-COPY --from=builder /opt/app-root/src/serve.json .
+RUN mkdir /app
 
-CMD ["sh", "-c","serve", "-p", "$PORT", "-s", "."]
+# Copy the app built on build-stage to the resulting container.
+COPY --from=builder /opt/app-root/src/dist /app
+
+# Copying the configuration file.
+COPY nginx.conf /etc/nginx/nginx.conf
+
+# Daemon off makes nginx to run on the foreground with only one process.
+# Docker will kill the container if the process dies.
+CMD ["nginx", "-g", "daemon off;"]
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/nodejs-16:1-5 as builder
+FROM registry.access.redhat.com/ubi8/nodejs-16:1-5 AS builder
 
 
 COPY ./package*.json ./
@@ -6,7 +6,7 @@ RUN npm ci
 COPY . ./
 RUN npm run build
 
-FROM registry.access.redhat.com/ubi8/nginx-120 as deployer
+FROM registry.access.redhat.com/ubi8/nginx-120 AS deployer
 
 RUN mkdir /app
 
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -15,8 +15,13 @@ WORKDIR /app
 # Copy the app built on build-stage to the resulting container.
 COPY --from=builder /opt/app-root/src/dist .
 
-# Copying the configuration file.
+# Copying the main configuration file.
 COPY nginx.conf /etc/nginx/nginx.conf
+
+# Copying the configuration template file to templates dir. 
+# It will be processed to transform the vars in the template files into 
+# evaluated values and save the resulting file as /etc/nginx/conf.d/default.conf.
+COPY default.conf.template /etc/nginx/templates/
 
 # Daemon off makes nginx to run on the foreground with only one process.
 # Docker will kill the container if the process dies.
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -6,12 +6,14 @@ RUN npm ci
 COPY . ./
 RUN npm run build
 
-FROM registry.access.redhat.com/ubi8/nginx-120 AS deployer
+FROM registry.access.redhat.com/ubi8/nginx-122 AS deployer
 
 RUN mkdir /app
 
+WORKDIR /app
+
 # Copy the app built on build-stage to the resulting container.
-COPY --from=builder /opt/app-root/src/dist /app
+COPY --from=builder /opt/app-root/src/dist .
 
 # Copying the configuration file.
 COPY nginx.conf /etc/nginx/nginx.conf
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -14,6 +14,8 @@ RUN npm run build
 
 FROM registry.access.redhat.com/ubi8/nginx-122 AS deployer
 
+USER default
+
 WORKDIR /opt/app-root/src
 
 # Copy the app built on build-stage to the resulting container.
--- a/sources/packages/web/Dockerfile.dev
+++ b/sources/packages/web/Dockerfile.dev
@@ -24,20 +24,19 @@ RUN npm run build-local
 # Get a nginx image to have only the compiled app in the resulting image.
 FROM nginx:1.22 AS deployer
 
-RUN mkdir /app
-
+RUN mkdir -p /opt/app-root/src
 
 RUN mkdir /etc/nginx/templates
 
 # Copy the app built on builder stage to the resulting container.
-COPY --from=builder /app/dist /app
+COPY --from=builder /app/dist /opt/app-root/src
 
 # Copying the main configuration file.
 COPY nginx.conf /etc/nginx/nginx.conf
 
 # Copying the configuration template file to templates dir. 
 # It will be processed to transform the vars in the template files into 
-# evaluated values and save the resulting file as /etc/nginx/conf.d/default.conf.
+# evaluated values and save the resulting files into /etc/nginx/conf.d/ dir.
 COPY default.conf.template /etc/nginx/templates/
 
 # Changing ownwership of some nginx directories as it will run as nginx user.
--- a/sources/packages/web/Dockerfile.dev
+++ b/sources/packages/web/Dockerfile.dev
@@ -1,5 +1,5 @@
-# "build-stage", based on Node.js, to build and compile the frontend.
-FROM node:16.17.0-alpine3.16 AS build-stage
+# "builder" stage, based on Node.js, to build and compile the frontend.
+FROM node:16.17.0-alpine3.16 AS builder
 
 LABEL maintainer="BC GOV"
 
@@ -21,16 +21,16 @@ COPY . .
 # Building app
 RUN npm run build-local
 
-# Get a nginx image to have only the compiled app ready for production.
-FROM nginx:1.22 AS production-stage
+# Get a nginx image to have only the compiled app in the resulting image.
+FROM nginx:1.22 AS deployer
 
 RUN mkdir /app
 
 
 RUN mkdir /etc/nginx/templates
 
-# Copy the app built on build-stage to the resulting container.
-COPY --from=build-stage /app/dist /app
+# Copy the app built on builder stage to the resulting container.
+COPY --from=builder /app/dist /app
 
 # Copying the main configuration file.
 COPY nginx.conf /etc/nginx/nginx.conf
--- a/sources/packages/web/Dockerfile.dev
+++ b/sources/packages/web/Dockerfile.dev
@@ -1,37 +1,31 @@
-# "build-stage", based on Node.js, to build and compile the frontend
+# "build-stage", based on Node.js, to build and compile the frontend.
 FROM node:16.17.0-alpine3.16 AS build-stage
 
 LABEL maintainer="BC GOV"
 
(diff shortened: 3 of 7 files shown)


sources/packages/web/Dockerfile
```diff
@@ -1,15 +1,21 @@
...
+FROM registry.access.redhat.com/ubi8/nodejs-16:1-5 AS builder
 
 
 COPY ./package*.json ./
 RUN npm ci
 COPY . ./
 RUN npm run build
 
-FROM registry.access.redhat.com/ubi8/nodejs-16:1-5 as deployer
+FROM registry.access.redhat.com/ubi8/nginx-120 AS deployer
```
andrewsignori-aot, line 9: Apparently, we also have version 1.22 available. Can we use it instead?
https://catalog.redhat.com/software/containers/ubi8/nginx-122/627245493b950f9f4eb01218

sources/packages/web/Dockerfile
andrewsignori-aot, line 1: Can we set a `WORKDIR /app` to make the `COPY --from=builder /opt/app-root/src/dist /app`
(e.g. `COPY --from=build-stage /app/dist /app`) easier to follow?

sources/packages/web/Dockerfile
```diff
@@ -1,15 +1,21 @@
...
 
-CMD ["sh", "-c","serve", "-p", "$PORT", "-s", "."]
\ No newline at end of file
+# Copy the app built on build-stage to the resulting container.
+COPY --from=builder /opt/app-root/src/dist /app
+
+# Copying the configuration file.
+COPY nginx.conf /etc/nginx/nginx.conf
+
+# Daemon off makes nginx to run on the foreground with only one process.
```
ann-aot, line 19: 👍

sources/packages/web/Dockerfile.dev
```diff
@@ -22,10 +22,25 @@ RUN npm ci
 COPY . .
 
 # Building app
-RUN npm run build
+RUN npm run build-local
+
+# Set a nginx a user to run the process
```
ann-aot, line 27: minor `Set a nginx user to run the process`

sources/packages/web/Dockerfile.dev
```diff
@@ -1,5 +1,5 @@
-# Base Image
-FROM node:16.17.0-alpine3.16
+# "build-stage", based on Node.js, to build and compile the frontend
```
ann-aot, line 1: minor, maybe a period

sources/packages/web/nginx.conf
```diff
@@ -0,0 +1,30 @@
...
+  server {
+    listen       8080;
+    server_name  localhost;
+    location / {
+      root   /app;
+      index  index.html;
+      try_files $uri $uri/ /index.html;
+    }
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
```
ann-aot, line 26: Just for my understanding, what is `/50x.html`. is there an HTML file?

ann-aot:
Good work @andrepestana-aot 👍 Just some minor comments and a question

sources/packages/web/Dockerfile.dev
```diff
@@ -22,10 +22,25 @@ RUN npm ci
...
 
 # Building app
-RUN npm run build
+RUN npm run build-local
+
+# Set a nginx a user to run the process
+USER nginx
 
 # Exposing application port
 EXPOSE ${PORT}
```
andrewsignori-aot, line 31: Does it still make sense to have it exposed at the `build-stage`?

sources/packages/web/Dockerfile.dev
andrewsignori-aot, line 1: I don't quite remember why we need the instruction on line 7 (`ENV PATH $PATH:/app/node_modules/.bin`).
Is it possible to have it removed?
The general idea IMO would be to make this as close as possible to the PROD one besides Openshift images.

sources/packages/web/nginx.conf
```diff
@@ -0,0 +1,30 @@
+user  nginx;
+worker_processes  1;
```
andrewsignori-aot, line 2: We may need to review those configurations. Leaving this as the default value 1 can maybe lead to an underutilization of the available resources.

sources/packages/web/Dockerfile.dev
guru-aot: I believe, setting the path variaable is not needed we can remove it for dockerfile.dev

sources/Makefile
```diff
@@ -153,6 +153,11 @@ api: ## <Helper> :: Executes into the workspace container.
 	@echo "+\n++Shelling into local workspace ...++\n+"
 	@docker-compose exec api bash
 
+# Local Web (nginx)
```
guru-aot, line 156: Nice one 👍

guru-aot (commented):
Nice work @andrepestana-aot , we can have a discussion on the nginx configuration with the team and can close it.

sources/packages/web/nginx.conf
```diff
@@ -0,0 +1,30 @@
...
+  default_type  application/octet-stream;
+  log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                    '$status $body_bytes_sent "$http_referer" '
+                    '"$http_user_agent" "$http_x_forwarded_for"';
+  access_log  /var/log/nginx/access.log  main;
+  sendfile        on;
+  keepalive_timeout  65;
+  server {
+    listen       8080;
+    server_name  localhost;
```
andrewsignori-aot, line 19: Should this not be updated to the real hostname?
I guess that, if present, nginx will try to validate the hostname.

sources/packages/web/nginx.conf
andrepestana-aot: That is the html file for the 50x errors. We don't have one. Should we create a place holder one or remove that configuration for now? Ideas? @dheepak-aot @guru-aot @andrewsignori-aot  @sh16011993

sources/packages/web/Dockerfile
```diff
@@ -1,15 +1,21 @@
...
+# Docker will kill the container if the process dies.
+CMD ["nginx", "-g", "daemon off;"]
```
dheepak-aot, line 21: 👍

andrewsignori-aot (commented):
Great work! Please take a look at the comments.
The below BC Gov app has an nginx config com a section named "Headers required by Web App Vulnerability Scan". Maybe that is something else that we will need to look into in a separate ticket. After this effort, we can request a new Wava Scan to get an updated report.
https://github.com/bcgov/moh-hnweb/blob/main/frontend/.nginx/nginx.conf

sources/packages/web/Dockerfile.dev
dheepak-aot: Same PATH setting is there in api,queue-consumers and workflow docker dev files.

image

sources/packages/web/nginx.conf
andrewsignori-aot: I would be ok to have it removed. We can track improvements in a new ticket to be created.

sources/packages/web/Dockerfile.dev
```diff
@@ -22,10 +22,25 @@ RUN npm ci
 COPY . .
 
 # Building app
-RUN npm run build
+RUN npm run build-local
+
+# Set a nginx a user to run the process
+USER nginx
```
dheepak-aot, line 28: This line has no impact before line 34 right?  where from nginx is not used? 

should this come before line 46?

sources/packages/web/nginx.conf
```diff
@@ -0,0 +1,30 @@
...
+  include       /etc/nginx/mime.types;
+  default_type  application/octet-stream;
+  log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                    '$status $body_bytes_sent "$http_referer" '
+                    '"$http_user_agent" "$http_x_forwarded_for"';
+  access_log  /var/log/nginx/access.log  main;
+  sendfile        on;
+  keepalive_timeout  65;
+  server {
+    listen       8080;
```
dheepak-aot, line 18: when nginx listens on 8080, it serves on port 80 and 443 (http and https). Should we consider changing the openshift route?

image

dheepak-aot:
Great work bringing on NGINX. Please have a look at the comments.

sources/packages/web/nginx.conf
andrewsignori-aot: I belive that we may need to update the Docker file to have something like `EXPOSE ${PORT}:8080`.

sources/packages/web/Dockerfile
sh16011993, line 21: Should we use `daemon off` in production?

sources/packages/web/nginx.conf
andrepestana-aot: if we just change the configuration for server_name it won't block. Nginx uses the the first server block as the default server. However if we set a default server to return 404, then that will prevent the server to accept requests without the configured hosts.
```
  server {
    listen 8080 default_server;
    server_name _;
    # Configuration for the default server block
    return 404;
  }
  server {
    listen       8080;
    server_name  test.com;
    location / {
      root   /app;
      index  index.html;
      try_files $uri $uri/ /index.html;
    }
    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
      root   /usr/share/nginx/html;
    }
  }
```

Should we configure it now in this ticket?

sources/packages/web/nginx.conf
andrepestana-aot: We can use `auto` that will config the server to have the same number as available cores for now.

sources/packages/web/Dockerfile
andrepestana-aot: Anywhere we have a docker image. That way (with `daemon off`) if the process dies, the container is destroyed and a new container starts.

sources/packages/web/Dockerfile.dev
andrepestana-aot: removed.

sources/packages/web/Dockerfile.dev
andrepestana-aot: I removed it as it's not used.

sources/packages/web/nginx.conf
andrepestana-aot: @dheepak-aot I'm going to change to use the env var: `"$PORT"`

sh16011993 (commented):
Very nice analysis work @andrepestana-aot 👍 .

sources/packages/web/nginx.conf
andrepestana-aot: This will help to understand how it works: http://nginx.org/en/docs/http/request_processing.html

andrewsignori-aot: Thanks for the further explanation, I got it now.
For this PR I believe that we can remove the `server_name localhost` and add the improvement suggested above using the `server_name_;` in an upcoming ticket.

sources/packages/web/Dockerfile.dev
```diff
@@ -1,31 +1,52 @@
...
-# Copying sources
+# Copying sources.
 COPY . .
 
 # Building app
-RUN npm run build
+RUN npm run build-local
+
+# Get a nginx image to have only the compiled app ready for production.
+FROM nginx:1.22 AS production-stage
```
andrewsignori-aot, line 25: Minor but can we make the layer name `production-stage` aligned with the other Dockerfile. Mentioning production here and in the line 24 comment can be misleading.

andrewsignori-aot (approved):
Thanks for making the changes and for the research. I left just one minor comment as a suggestion, hence approving it. Looks good 👍

ann-aot (approved):
LGTM 👍 Good work with the analysis

sources/packages/web/nginx.conf
dheepak-aot: EXPOSE ${PORT} is missing in dockerfile. is that expected?

dheepak-aot: Also are we not changing the port in make file. ?
image

andrepestana-aot: Changing the port is not needed as it'll use the env variable `PORT`. I'm trying to change as minimum as possible. The `EXPOSE` is not needed either.

andrepestana-aot: image

guru-aot (approved):
LGTM, nice work

sh16011993 (approved):
Thank You for the analysis and migration to nginx.

sources/packages/web/nginx.conf
andrepestana-aot: @dheepak-aot I had to set `ENV PORT 3030` in Dockerfile and did some changes in Dockerfile to replace the variable ${PORT} in default.conf because the ubi8 image works different from the image used in Dockerfile.dev.

sources/packages/web/Dockerfile
```diff
@@ -1,15 +1,32 @@
...
 
+# Application Port.
+ENV PORT 3030
 
 COPY ./package*.json ./
 RUN npm ci
 COPY . ./
+
+# Replace ${PORT} variable in the template and save as default.conf.
+RUN sed 's/${PORT}/'"${PORT}"'/g' default.conf.template > default.conf
```
andrewsignori-aot, line 11: Is it not the same as having the PORT hardcode directly at the file?
Just wondering what is the benefit between having it hardcoded in the file vs the like 4.
Either way it is just a question for my information, not a blocker.

andrewsignori-aot (approved):
Looks good to me 👍

sources/packages/web/Dockerfile
andrepestana-aot: In Dockerfile.dev it uses 8080 but for Dockerfile it uses 3030. We can change that later. It's just that in some places it has 3030 as the default port. We can change our local to 3030 too.

ann-aot (approved):
👍

Issue #2305, opened 2023-09-15
Nginx - change configuration to output logs to stdout


andrepestana-aot:
Nginx is configured to output logs to /var/log/nginx/access.log and /var/log/nginx/error.log but not to stdout. Logging to stdout is needed for Kibana to keep the logs.

**Describe the task**
Change configuration and start command.
Reference: https://stackoverflow.com/a/59710227

**Acceptance Criteria**
- [ ] Have the logs output to stdout and visible on Logs tab on openshift console.

PR #2313, opened 2023-09-18
#2305 - Nginx - change configuration to output logs to stdout

andrepestana-aot:
- Configured error_log and access_log to be redirected only to stdout. Kibana should be responsible for capturing the logs output on stdout.

Notice that error_log is now configured to "info" log level which means that all the higher [log levels](https://www.keycdn.com/support/nginx-error-log#nginx-error-log-severity-levels) will be also logged.

### DIFF
--- a/sources/packages/web/default.conf.template
+++ b/sources/packages/web/default.conf.template
@@ -7,10 +7,6 @@ log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
 server {
   # Log to stdout
   access_log /dev/stdout main;
-  
-  # Log to a file
-  access_log /var/log/nginx/access.log main;
-  
   listen       ${PORT};
   location / {
     root   /opt/app-root/src;
--- a/sources/packages/web/default.conf.template
+++ b/sources/packages/web/default.conf.template
@@ -1,5 +1,16 @@
 # nginx default configuration.
+
+log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                  '$status $body_bytes_sent "$http_referer" '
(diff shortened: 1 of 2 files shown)


sources/packages/web/default.conf.template
```diff
@@ -1,5 +1,16 @@
 # nginx default configuration.
+
+log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                  '$status $body_bytes_sent "$http_referer" '
+                  '"$http_user_agent" "$http_x_forwarded_for"';
+
 server {
+  # Log to stdout
+  access_log /dev/stdout main;
```
guru-aot, line 9: I believe this will print the logs on the console, also after deployment we can check if this is outputting it to kibana.

guru-aot (approved):

sh16011993 (approved):
LGTM @andrepestana-aot . Nice work 👍

sources/packages/web/default.conf.template
```diff
@@ -1,5 +1,16 @@
...
+  
+  # Log to a file
+  access_log /var/log/nginx/access.log main;
```
andrewsignori-aot, line 12: As we talked, let's avoid enabling the logs saved to a file right now and leave it for later to be enabled with some sort of file rotation.
Same for `sources/packages/web/nginx.conf`.
The idea is to keep the same behavior already in place with Kibana capturing the logs.

andrewsignori-aot (approved):

ann-aot (approved):
👍

dheepak-aot (approved):
Looks Good 👍

Issue #2180, opened 2023-08-10
Queued Assessments - Assessment Workflow Enqueuer Scheduler - Start Assessment


andrewsignori-aot:
Create a new scheduler to be executed every few seconds to check the database for assessments that must be queued to start its execution or to start its cancellation.

**Acceptance criteria**
- [x] Create a new scheduler `assessment-workflow-enqueuer`.
- [x] Add a configuration to the `sims.queue_configurations` table as below (no need for a retry since the execution interval should be less than a minute).
```json
{ "cron": "*/30 * * * * *", "cleanUpPeriod": 3600000, "dashboardReadonly": false }
```
- For every interaction:
  - [x] Select the oldest record from all applications where
    - application `current_processing_assessment_id` is null or different than `current_assessment_id`, 
    - and there is no assessment already being processed.
- For every selected assessment record:
  - [x] Update the `current_processing_assessment_id` with the assessment id.
  - [x] Update the assessment status to `Queued`.
  - [x] Ensure that the above DB operations are part of the same DB transaction.
  - [x] Add the assessment to the `start-application-assessment`. Currently, the queue is receiving `workflowName` and `assessmentId` as parameters. The `start-application-assessment` can be changed to optionally receive the `workflowName`, which would allow the scheduler and the current implementation to continue to work till the feature is completed.
- [x] Create the process taking into consideration that the same scheduler will also be processing the cancelation requests.
- Stop adding items to `start-application-assessment` queue on the below points.
  - [x] Application submission [application.students.controller.ts](https://github.com/bcgov/SIMS/blob/d5050a6dc8fa3dc6abe745cb17cd868e35cc8540/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts#L248)
  - [x] Student Appeal Request [student-appeal.aest.controller.ts](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.aest.controller.ts#L120)
  - [x] Offering bulk change [education-program-offering.service.ts](https://github.com/bcgov/SIMS/blob/81923206ba137731f7fd4e3e6f1776369f8a6898/sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts#L1204)
[...]

PR #2326, opened 2023-09-21
#2180 - Additional log improvement

andrewsignori-aot:
### Log improvements
- The date and time are now prepended in the job logger.
- Added a visual hierarchy to identify better the child logs.
- Changed the hierarchy organization to allow child process to be chronologically added, as seen in the image below (before the green ones, for instance, would be displayed altogether).

image

Created a new method to inspect the summary message to indicate if there is something that needs attention in the logs.

image

### Scheduler Enqueuer Query Change

The query was not filtering by application status and I thought that it would be better to ensure that only `Submitted` and `Completed` applications would be considered to have a workflow queued.

### Refactor

Adjusted the logs for the `assessment-workflow-enqueuer` and `start-application-assessment` to exclusively use the new `ProcessSummary` class instead of the `QueueProcessSummary`. The suggestion is intended to be a POC to have the `QueueProcessSummary` marked as deprecated.

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/processors/models/processors.models.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/models/processors.models.ts
@@ -1,5 +1,4 @@
 import { LoggerService } from "@nestjs/common";
-import { ProcessSummary } from "@sims/utilities/logger";
 
 export interface QueueProcessSummaryResult {
   /**
@@ -19,6 +18,7 @@ export interface QueueProcessSummaryResult {
 /**
  * Allow the logs aggregation for a process summary and optionally
  * allow the same logs to be saved to different outputs.
+ * @deprecated please use ProcessSummary from sims/utilities/logger.
  */
 export class QueueProcessSummary {
   private summary: string[] = [];
@@ -81,27 +81,6 @@ export class QueueProcessSummary {
       warnings: this.warnings.length ? this.warnings : undefined,
       errors: this.errors.length ? this.errors : undefined,
     };
-  }
-
-  /**
-   * Writes all the log entries.
-   * @param processSummary process summary logs.
-   * @param logger optionally provides a job logger
-   * in case one is not present.
-   */
-  async logProcessSummaryToJobLogger(
-    processSummary: ProcessSummary,
-    logger?: JobLogger,
-  ): Promise<void> {
-    const jobLogger = logger ?? this.loggers?.jobLogger;
-    if (!jobLogger) {
-      throw new Error("No job logger was provided.");
-    }
-    for (const logEntry of processSummary.flattenLogs()) {
-      await jobLogger.log(
-        `${logEntry.level.toUpperCase()}: ${logEntry.message}`,
-      );
-    }
   }
 }
 
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/workflow/assessment-workflow-enqueuer.scheduler.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/workflow/assessment-workflow-enqueuer.scheduler.ts
@@ -1,18 +1,18 @@
 import { InjectQueue, Process, Processor } from "@nestjs/bull";
 import { Job, Queue } from "bull";
 import { BaseScheduler } from "../base-scheduler";
-import { QueueNames, parseJSONError } from "@sims/utilities";
+import { QueueNames } from "@sims/utilities";
 import { QueueService } from "@sims/services/queue";
-import {
-  QueueProcessSummary,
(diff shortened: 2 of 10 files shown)


sources/packages/backend/apps/queue-consumers/src/processors/schedulers/workflow/assessment-workflow-enqueuer.scheduler.ts
```diff
@@ -35,39 +35,33 @@ export class AssessmentWorkflowEnqueuerScheduler extends BaseScheduler<void> {
...
     const processSummary = new ProcessSummary();
     try {
+      processSummary.info(
+        "Checking application assessments to be queued for start.",
+      );
+      // Process summary to be populated by the enqueueStartAssessmentWorkflows.
+      // In case an unexpected error happen the finally block will still be able to
+      // output the partial information captured by the processSummary.
+      const serviceProcessSummary = new ProcessSummary();
+      processSummary.children(serviceProcessSummary);
```
dheepak-aot, line 48: Let me know if I am missing something. this part can be done inside process summary class constructor right?

sources/packages/backend/apps/queue-consumers/src/processors/models/processors.models.ts
```diff
@@ -19,6 +18,7 @@ export interface QueueProcessSummaryResult {
 /**
  * Allow the logs aggregation for a process summary and optionally
  * allow the same logs to be saved to different outputs.
+ * @deprecated please use ProcessSummary from sims/utilities/logger.
```
ann-aot, line 21: 👍

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/workflow/assessment-workflow-enqueuer.scheduler.ts
andrewsignori-aot: Sorry, I am not following the comment. Do you mean the association of a child process?
Here we have a parent log and the creating of a child log to capture the service-related logs. I am not sure what could be done inside the `ProcessSummary` constructor.

sources/packages/backend/libs/utilities/src/logger/process-summary.ts
```diff
@@ -36,26 +36,32 @@ export enum LogLevels {
...
 /**
  * Logs grouped by different scopes.
  */
 export type ScopedLogLevel = Record<LogScopes, LogEntry[]>;
 
+/**
+ * Allows a log entry that represents either a single log
+ * or a new collection of child entries.
+ */
+type LogInfoEntry = LogEntry | ProcessSummary[];
```
dheepak-aot, line 57: 👍

sources/packages/backend/apps/queue-consumers/src/services/application/application.service.ts
```diff
@@ -94,6 +94,9 @@ export class ApplicationService {
             );
           }),
         )
+        .andWhere("application.applicationStatus IN (:...status)", {
```
dheepak-aot, line 97: 👍

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/workflow/assessment-workflow-enqueuer.scheduler.ts
dheepak-aot: I get it now. it is for the first level of indentation. 
image

My understanding is better now.

With that in mind, just wondering if service process summary could be retuned by a getter in the class like this.  So that we can avoid doing this assignment at processor level.

```ts
await this.workflowEnqueuerService.enqueueStartAssessmentWorkflows(
        processSummary.getServiceProcessSummary(),
      );
```

dheepak-aot (approved):
Looks good. Thanks for the additional effort 👍

I just have a minor question. not a blocker.

ann-aot (approved):
👍 Nice work @andrewsignori-aot

andrepestana-aot (approved):
LGTM!

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/workflow/assessment-workflow-enqueuer.scheduler.ts
andrewsignori-aot: The idea of having it created at the processor lever and passing it as a reference is to allow access to it in case some unhandled exception happens inside the service. Having it at the processor level allows us to dump any information added to it in the final block. Does it make sense?

guru-aot (commented):
LGTM

Issue #1910, opened 2023-04-27
API - Student profile returns SIN


andrepestana-aot:
Review getStudentProfile() method from StudentStudentsController that returns the student profile data with SIN. The API should never return SIN information for the own student user. 

- [ ] Review the API endpoint for student only (ministry and aest will have access to the student's SIN).
- [ ] Change StudentInstitutionsController(e2e)-getStudentProfile E2E test.

PR #2354, opened 2023-09-26
#1910 - Student profile returns SIN

andrepestana-aot:
- Added a new DTO to contain sensitive data and changed the logic to make the student api to not bring the SIN data.
- Removed the SIN property from the student profile retrieved by the student.

## Student view
image

## Institution view
image

## Ministry view
image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
@@ -181,10 +181,13 @@ export class StudentProfileAPIOutDTO {
   contact: ContactInformationAPIOutDTO;
   validSin: boolean;
   disabilityStatus: DisabilityStatus;
+}
+
+export class SensitiveDataStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
   sin: string;
 }
 
-export class AESTStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
+export class AESTStudentProfileAPIOutDTO extends SensitiveDataStudentProfileAPIOutDTO {
   hasRestriction: boolean;
 }
 
--- a/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
@@ -183,9 +183,11 @@ export class StudentProfileAPIOutDTO {
   disabilityStatus: DisabilityStatus;
 }
 
-export class SensitiveDataStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
+class SensitiveDataStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
   sin: string;
 }
+
+export class InstitutionStudentProfileAPIOutDTO extends SensitiveDataStudentProfileAPIOutDTO {}
 
 export class AESTStudentProfileAPIOutDTO extends SensitiveDataStudentProfileAPIOutDTO {
   hasRestriction: boolean;
--- a/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
@@ -183,13 +183,11 @@ export class StudentProfileAPIOutDTO {
   disabilityStatus: DisabilityStatus;
 }
 
-class SensitiveDataStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
+export class InstitutionStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
   sin: string;
 }
 
-export class InstitutionStudentProfileAPIOutDTO extends SensitiveDataStudentProfileAPIOutDTO {}
-
-export class AESTStudentProfileAPIOutDTO extends SensitiveDataStudentProfileAPIOutDTO {
+export class AESTStudentProfileAPIOutDTO extends InstitutionStudentProfileAPIOutDTO {
   hasRestriction: boolean;
 }
 
--- a/sources/packages/backend/apps/api/src/route-controllers/student/student.aest.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/student.aest.controller.ts
@@ -48,7 +48,6 @@ import {
   SINValidationsAPIOutDTO,
   UniqueFileNameParamAPIInDTO,
   UpdateSINValidationAPIInDTO,
-  SensitiveDataStudentProfileAPIOutDTO,
 } from "./models/student.dto";
 import { Response } from "express";
 import { FileInterceptor } from "@nestjs/platform-express";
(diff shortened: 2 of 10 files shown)


sources/packages/backend/apps/api/src/route-controllers/student/student.aest.controller.ts
```diff
@@ -242,13 +243,18 @@ export class StudentAESTController extends BaseController {
...
+        withSensitiveData: true,
+      }),
       this.studentRestrictionService.getStudentRestrictionsById(studentId, {
         onlyActive: true,
       }),
     ]);
 
-    return { ...student, hasRestriction: !!studentRestrictions.length };
+    return {
+      ...(student as SensitiveDataStudentProfileAPIOutDTO),
```
andrewsignori-aot, line 255: The lastest PRs have been creating overloads to better explicit the return type and avoid the cast.
For instance, the method `getStudentProfile` can receive an overload and the cast can be eliminated.

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
```diff
@@ -181,10 +181,13 @@ export class StudentProfileAPIOutDTO {
   contact: ContactInformationAPIOutDTO;
   validSin: boolean;
   disabilityStatus: DisabilityStatus;
+}
+
+export class SensitiveDataStudentProfileAPIOutDTO extends StudentProfileAPIOutDTO {
```
andrewsignori-aot, line 186: We probably have web dtos counterparts for these DTOs. Do they require any change also?

sources/packages/backend/apps/api/src/route-controllers/student/student.aest.controller.ts
andrepestana-aot: Ok. Did the overload.

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
andrepestana-aot: I don't think we should change the web. In the web we are using the same DTO for all cases:
image
and in runtime JS doesn't complain about it: 

image
Institution

image
Student

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -113,14 +114,36 @@ export class StudentControllerService {
...
+  /**
+   * Get the student information that represents the profile.
+   * @param studentId student id to retrieve the data.
+   * @param options options:
+   * - `withSensitiveData` boolean option to return sensitive data such as SIN.
+   * @returns student profile details.
+   */
+  async getStudentProfile(
+    studentId: number,
+    options: { withSensitiveData: boolean },
```
andrewsignori-aot, line 127: Minor that we missed in some PRs, if line 127 is not declared as `options: { withSensitiveData: true },` the method consumer can provide it as false which would lead to an incorrect return type.

guru-aot (approved):

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
dheepak-aot: But in any case, student profile DTO must not contain SIN right?

ann-aot: then shouldn't be make `sin` as optional in web?

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -113,14 +114,36 @@ export class StudentControllerService {
...
+  /**
+   * Get the student information that represents the profile.
+   * @param studentId student id to retrieve the data.
+   * @param options options:
+   * - `withSensitiveData` boolean option to return sensitive data such as SIN.
+   * @returns student profile details.
+   */
+  async getStudentProfile(
+    studentId: number,
+    options: { withSensitiveData: true },
```
ann-aot, line 127: are we able to set default values inside options? I remember it was not working, and I had to do something like this 
image
correct me if Iam wrong

sources/packages/backend/apps/api/src/route-controllers/student/models/student.dto.ts
andrewsignori-aot: I believe that the below one can reflect both possible DTOs received from the API.
https://github.com/bcgov/SIMS/blob/43ca26b8b71e1329c9c01819bbfbaf28ae48b9ed/sources/packages/web/src/services/http/StudentApi.ts#L46-L52

The service is already returning a `StudentProfile` that can have optional properties like SIN, as @ann-aot mentioned.

https://github.com/bcgov/SIMS/blob/43ca26b8b71e1329c9c01819bbfbaf28ae48b9ed/sources/packages/web/src/services/StudentService.ts#L52-L62

Does it make sense?

sources/packages/backend/apps/api/src/route-controllers/student/student.aest.controller.ts
```diff
@@ -242,13 +242,18 @@ export class StudentAESTController extends BaseController {
     @Param("studentId", ParseIntPipe) studentId: number,
   ): Promise<AESTStudentProfileAPIOutDTO> {
     const [student, studentRestrictions] = await Promise.all([
-      this.studentControllerService.getStudentProfile(studentId),
+      this.studentControllerService.getStudentProfile(studentId, {
+        withSensitiveData: true,
```
sh16011993, line 246: 👍

sh16011993 (approved):
Nice work. 👍

sources/packages/web/src/components/common/students/StudentProfile.vue
```diff
@@ -98,8 +98,15 @@
...
 import { StudentService } from "@/services/StudentService";
 import { useFormatters } from "@/composables";
-import { StudentProfile } from "@/types";
 import { AddressAPIOutDTO } from "@/services/http/dto";
+import { StudentProfile } from "@/types";
+
+/**
+ *  Used to combine institution and ministry DTOs and make SIN explicitly mandatory.
+ */
+interface SharedStudentProfile extends Omit<StudentProfile, "sin"> {
```
ann-aot, line 107: What exactly are we trying to do here?

andrewsignori-aot (approved):
Thanks for making the changes and for the discussions. Looks good 👍

ann-aot (approved):
Thanks @andrepestana-aot for doing the changes and for the info 👍

sources/packages/web/src/components/common/students/StudentProfile.vue
andrepestana-aot: That's an interface to represent institution and ministry properties as we can have both here in this case.

dheepak-aot (approved):
Thanks for doing the changes 👍

Issue #1807, opened 2023-03-15
Fix css style for dividers (<hr />)


andrepestana-aot:
The HR is there but its style is set to be transparent:

![image.png](https://images.zenhubusercontent.com/601c6b4f5322cf5c519d503f/6e91ed10-5551-43a1-8d9f-cb5a4bf934ab)

![image.png](https://images.zenhubusercontent.com/601c6b4f5322cf5c519d503f/cee33e5a-8d39-4c89-ba81-5faab5ac9e76)

It's needed to check why it changed to transparent and make it visible again.

ann-aot:
As discussed, keeping the PR as it is and created a ticket with the fix https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2359 to implement after the release signoff

PR #2356, opened 2023-09-26
#1807 - Fix css style for dividers (<hr />)

ann-aot:
- There were some new properties and classes for v-divider in Vuetify 3, added those properties for all `v-divider` for the fix. 
- The default opacity of the divider is `--v-border-opacity: 0.12;` updated it to `--v-border-opacity: 1 `
- Tried to make it as close to Figma
image
https://vuetifyjs.com/en/components/dividers/
**Screenshots**
image
image
image
image
image
image
image
image
image

### DIFF
--- a/sources/packages/web/src/components/aest/institution/modals/ApproveDenyDesignationModal.vue
+++ b/sources/packages/web/src/components/aest/institution/modals/ApproveDenyDesignationModal.vue
@@ -26,7 +26,7 @@
                 variant="outlined"
                 :rules="[checkStringDateFormatRule]"
                 hide-details="auto" /></v-col></v-row
-          ><v-divider></v-divider>
+          ><v-divider class="border-opacity-100" :thickness="2" />
           <h4 class="label-bold color-blue">Designate locations</h4>
           <v-row
             v-for="(item, index) in formModel.locationsDesignations"
--- a/sources/packages/web/src/components/common/restriction/ViewRestriction.vue
+++ b/sources/packages/web/src/components/common/restriction/ViewRestriction.vue
@@ -38,7 +38,7 @@
           ></v-row>
         </content-group>
         <template v-if="showResolution">
-          <v-divider></v-divider>
+          <v-divider class="border-opacity-100" :thickness="2" />
           <h3 class="category-header-medium mb-5">Resolution</h3>
           <v-textarea
             v-if="allowUserToEdit"
--- a/sources/packages/web/src/components/generic/ModalDialogBase.vue
+++ b/sources/packages/web/src/components/generic/ModalDialogBase.vue
@@ -19,12 +19,13 @@
           </h2>
         </slot>
       </v-card-title>
-      <v-divider class="mx-6 mt-1 mb-4"></v-divider>
+      <v-divider class="border-opacity-100" :thickness="2" inset />
       <v-card-text class="pt-0">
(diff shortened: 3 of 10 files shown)


sources/packages/web/src/components/aest/institution/modals/ApproveDenyDesignationModal.vue
```diff
@@ -26,7 +26,7 @@
                 variant="outlined"
                 :rules="[checkStringDateFormatRule]"
                 hide-details="auto" /></v-col></v-row
-          ><v-divider></v-divider>
+          ><v-divider class="border-opacity-100" :thickness="2" />
```
andrewsignori-aot, line 29: Just to think about it, should we consider that the project will use the opacity 100 as a default which means that we can override the default v-divider opacity globally?

andrewsignori-aot (commented):
Nice catch, just a comment to be considered.

sources/packages/web/src/components/aest/institution/modals/ApproveDenyDesignationModal.vue
dheepak-aot: In that case(if border-opacity-100 is going to be set for horizontal and vertical divider) we can do like this to set the default class. 

image

guru-aot: I agree we can have it globally

andrepestana-aot: We can just change the CSS:

image

dheepak-aot, line 29: I had the same feeling towards thickness as well. Instead of setting it 2 everywhere is there a way that we can bring that from constant. I know that we stay away from using typescript constants on vue components. @ann-aot tried to check if that works by setting defaults for a component at vuetify.ts and it didn't work. 

This is not a blocker comment. but trying to see if anyone has a better suggestion .

@andrewsignori-aot @andrepestana-aot @guru-aot ?

andrepestana-aot: I think that should be set via CSS: 
```css
.v-divider {
  opacity: 1 !important;
  border-top: 2px solid $common-border-color !important;
}
```

andrewsignori-aot: The styles can be used as @andrepestana-aot mentioned, I'm just not sure about the "border-top" for instance.
I wold look at the most specific style and override it.
For instance, probably vertical and horizontal may have different ones.

sh16011993 (approved):
Nice catch 👍 LGTM.

ann-aot:
As discussed, keeping the PR as it is and created a ticket https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2359 to implement after the release signoff

dheepak-aot (approved):
Looks good. Thanks for the investigation.

guru-aot (approved):
LGTM

andrewsignori-aot (approved):
Thanks for the research and the ticket creation. Looks good 👍

andrepestana-aot (approved):
👍

Issue #2287, opened 2023-09-12
Load Testing of Form.io using Student Application submission.


JasonCTang:
**Describe the task**
Load testing for Form.io submission.  Test using the Student Application to test the amount of submission that the form.io server can handle under different configurations and pod numbers. As a suggestion, we can have a Student Application payload sample submitted multiple times directly to the form.io server and check how the form.io server handles the requests till it stops.

**Acceptance Criteria**
- [x] Determine and document load testing strategy in wiki.
- [x] Document testing instructions in wiki so that the process can be repeated.
- [x] Run multiple test to check the performance by increasing the number of PODs and modifying CPU/memory per POD.
- [x] Produce reporting of output vs pod#s/CPU/memory configuration. 
  - Jason/Michelle to identify the location to save this report.
- [x] Identify the best configuration and modify the YAML file to standardize future deployment.

**Additional context**

andrewsignori-aot:
dev-forms
- 2 PODs
- 

checks.........................: 100.00% ✓ 977      ✗ 0
data_received..................: 8.7 MB  65 kB/s
data_sent......................: 3.7 MB  27 kB/s
http_req_blocked...............: avg=81.24ms  min=0s       med=35.86ms max=1.23s   p(90)=157.88ms p(95)=367.93ms
http_req_connecting............: avg=16.38ms  min=0s       med=15.56ms max=89.78ms p(90)=18.51ms  p(95)=19.93ms
http_req_duration..............: avg=11.78s   min=199.45ms med=11.44s  max=22.52s  p(90)=16.07s   p(95)=16.84s
  { expected_response:true }...: avg=11.78s   min=199.45ms med=11.44s  max=22.52s  p(90)=16.07s   p(95)=16.84s
http_req_failed................: 0.00%   ✓ 0        ✗ 978
http_req_receiving.............: avg=893.38µs min=0s       med=0s      max=16.09ms p(90)=1.65ms   p(95)=5.45ms
http_req_sending...............: avg=49.36µs  min=0s       med=0s      max=1ms     p(90)=53.82µs  p(95)=505.54µs
http_req_tls_handshaking.......: avg=64.71ms  min=0s       med=19.82ms max=1.22s   p(90)=138.56ms p(95)=350.97ms
http_req_waiting...............: avg=11.78s   min=185.99ms med=11.44s  max=22.51s  p(90)=16.07s   p(95)=16.84s
http_reqs......................: 978     7.300393/s
[...]

PR #2407, opened 2023-10-11
#2287 - Load Testing of Form.io using Student Application submission

andrewsignori-aot:
- Checked some possible tools for load testing,
- Chose K6 to create a POC with k6 based on the below facts:
  - It has an Open Souce version.
  - Tests can be written in javascript. "JavaScript is not generally well suited for high performance. To achieve maximum performance, the tool itself is written in Go, embedding a JavaScript runtime allowing for easy test scripting." (source https://k6.io/docs/).
  - Out-of-box output visualization can be extended (https://k6.io/blog/ways-to-visualize-k6-results).
  - It seems to be a well-known and largely used framework.
  - Extensive documentation and articles about it.
- Created a new package called `load-test` with two examples of tests targeting form.io. The package was created using a merge between the below two available templates. The main difference was that those templates were using `babel` and I replaced them with `ts-loader` which is the same already in place for the backend.
  - https://github.com/grafana/k6-template-typescript
  - https://github.com/grafana/k6-template-es6
- Adapted the solution to transpile typescript into javascript that K6 can execute.
- Added to the ticket comments many samples with test outputs with PODs variation and the `Horizontal Pod Autoscaler` working.
- Currently, the `Horizontal Pod Autoscaler` in OpenShift is not working and they will need to be adjusted. This PR is targeting the adjustment in the forms POD only.
- Adjusted the `forms-deploy.yml` with a new configuration for the  `Horizontal Pod Autoscaler`, CPU and Memory. The CPU and memory currently configured were not reflecting the values on DEV/TEST/PROD.
- The adjusted values, including the 50% CPU for the  `Horizontal Pod Autoscaler` were based on a 10min load test using 60 virtual users where 10334 submissions were made with 99.98% success and PODs auto-scaling from 2 to 6 to attend the requests. The average response time was 3.49s.

**Below is what looks like the deployment config when the autoscaler is active.**

image

**The below image shows the PODs number increasing in a load test where the CPU went over the limit.**

image

### K6 considerations
- It does not execute on top of Nodejs, it is written using Go and comes with JavaScript runtime allowing for easy test scripting, which means that Nodejs-specific libraries won't work.

### References
- https://nodeteam.medium.com/load-testing-tools-for-node-js-developers-98291ed75a4b
- https://medium.com/@sebastian.southern/performance-testing-in-typescript-with-k6-a65377f547e6
- https://k6.io/docs/get-started/installation/
- https://k6.io/docs/
- https://k6.io/docs/using-k6/test-lifecycle/
- https://k6.io/blog/ways-to-visualize-k6-results/
- https://k6.io/docs/testing-guides/api-load-testing/
- https://levelup.gitconnected.com/load-test-your-nodejs-app-using-k6-f7b2bd8fa5cf
- https://chat.developer.gov.bc.ca/channel/devops-how-to?msg=nMAruEuBv7iL67LEu

### Documentation created (work in progress)
- https://github.com/bcgov/SIMS/wiki/Load-testing


devops/openshift/forms-deploy.yml
```diff
@@ -164,11 +164,11 @@ objects:
     spec:
       scaleTargetRef:
         apiVersion: apps/v1
-        kind: ReplicationController
-        name: ${NAME}-rc
+        kind: DeploymentConfig
+        name: ${NAME}
```
guru-aot, line 168: Cool, thanks for doing this change

devops/openshift/forms-deploy.yml
```diff
@@ -164,11 +164,11 @@ objects:
...
       minReplicas: "${{REPLICAS}}"
-      maxReplicas: 10
-      targetCPUUtilizationPercentage: 80
+      maxReplicas: 6
+      targetCPUUtilizationPercentage: 50
```
guru-aot, line 171: I had this doubt yesterday when you are demoing, if the targetCPUUtilizationPercentage is cannot be raised above 50%. Can we change the CPU Request to a higher value like 1, and see if something changes??

andrepestana-aot: As I understood it was tested with 80% but there were many failed requests because it took long for the replicas to be up and running, so with 50% it had better results.

andrewsignori-aot: @guru-aot the CPU can go beyond 50% and it will because the new replicas will take a couple of minutes to start getting the extra load. As @andrepestana-aot mentioned, when leaving at 80% it was degrading too much the success responses.

andrepestana-aot (approved):
Look good!

devops/openshift/forms-deploy.yml
```diff
@@ -164,11 +164,11 @@ objects:
...
       scaleTargetRef:
         apiVersion: apps/v1
-        kind: ReplicationController
-        name: ${NAME}-rc
+        kind: DeploymentConfig
+        name: ${NAME}
       minReplicas: "${{REPLICAS}}"
-      maxReplicas: 10
-      targetCPUUtilizationPercentage: 80
+      maxReplicas: 6
```
ann-aot, line 170: just for my understanding my `minReplicas` is set as a variable and for `maxReplicas` we directly assigned the value

ann-aot (approved):
Nice research and work @andrewsignori-aot 👍 Thanks for the walkthrough. Just minor question.

sources/packages/load-test/src/_tests_/form-io/form-io-get-application/form-io-get-application.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  return { formIOHeader };
+}
+
+/**
+ * Test scenario to executed as many times as defined by the test options.
+ * Part of the K6 life cycle.
+ * @param setupData setup data returned by setup method.
+ */
+export default function (setupData: SetupData) {
+  const response = getFormByAlias("sfaa2023-24", setupData.formIOHeader);
```
guru-aot, line 42: in the future, if we are setting forms for the next fiscal year and the form has some changes, do you think still we need to run these load testing parameter and find an optimal value for the CPU and replicas?

andrewsignori-aot: If we run the load testing every few months we will end up needing to have it updated but I do not believe that we have its content changed in a way that would impact the test results in such a way that it would be a concern.

devops/openshift/forms-deploy.yml
andrewsignori-aot: Yes, right now I am not changing it but the REPLICAS is also set in the parameters as 2 right now.

guru-aot (approved):
LGTM, nice work and start @andrewsignori-aot

dheepak-aot (approved):
Great job and thanks for the walk through.

Issue #2379, opened 2023-10-04
Contribution from Applicant and Spouse- Fix Contribution Expected from Applicant and Spouse Field in IER12 implementation  


HRAGANBC:
**Description of Task:**

**Description of Error:**
In the file below, one of the student applications has a record length that is longer than all the others. The error is located within application number 2023000696 - It seems to be Position number 516 in the IER record where the error starts. This corresponds to Contribution expected from Applicant and spouse

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHdIQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9840b7829b2bbaa321ebf8e695dbff8996145fd0/image.png)
 

[IER_012_2023-10-03_17.30.11.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeGdIQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1f0f8515e9e421d434aec63ba7f6e9c70094f976/IER_012_2023-10-03_17.30.11.txt)

**Acceptance Criteria**
- [x] Ensure the combined decimal places is used for the mentioned fields 
- [x] Check if the value is without decimal places. 
- [x] Check other numeric fields and confirm no decimal as shown below there are at least two decimals 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUFOQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--61d4acc64a4886ca9a5eafe70a17da69beb962a9/image.png)

PR #2419, opened 2023-10-16
#2379 - Fix computer multiplication in combine decimal places

dheepak-aot:
## Floating point arithmetic issue in programming languages

image

https://stackoverflow.com/questions/45221535/javascript-floating-point-multiply-by-100-still-has-errors

## Issue and fix

Certain decimal numbers when multiplied by 100 may not return precise mathematical result as we expect

image

`Math.round()` seems to be one of the solutions to handle this issue.

But in the `combineDecimalPlaces` method we are dividing and then multiplying by 100 again, which re-creates the same issue which was fixed by `Math.round()`

image

Hence: Divide and multiply of the rounded value by multiplier is removed.

### File output
image

### DIFF
--- a/sources/packages/backend/libs/utilities/src/math-utils.ts
+++ b/sources/packages/backend/libs/utilities/src/math-utils.ts
@@ -31,7 +31,5 @@ export function combineDecimalPlaces(
 ): number {
   const multiplier = Math.pow(10, decimalPlaces);
   // Round the number up to the number of decimal places.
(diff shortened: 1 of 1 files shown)


andrewsignori-aot (approved):
Looks good 👍 
Thanks for the nice chat 😉

ann-aot (approved):
👍

guru-aot (approved):

Issue #2440, opened 2023-10-18
Investigate backup failures in TEST and PROD database


JasonCTang:
- [x] Investigate backup failures in TEST and PROD database
- [x] upgrade to latest version of backup container for ALL environments

PR #2454, opened 2023-10-23
#2440-Update_DB_Backup_container_image

guru-aot:
Upgrade to latest version of backup container


devops/Makefile
```diff
@@ -14,6 +14,7 @@ export NEW_DB := $(or $(NEW_DB), $$NEW_DB)
 export JOB_NAME := $(or $(JOB_NAME), $$JN)
 export SITE_MINDER_LOGOUT_URL := $(or $(SITE_MINDER_LOGOUT_URL), https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi)
 export PATRONI_SIMSDB_BACKUP_APP_NAME := $(or $(PATRONI_SIMSDB_BACKUP_APP_NAME), patroni-simsdb-backup)
+export PATRONI_SIMSDB_BACKUP_TAG_NAME := $(or $(PATRONI_SIMSDB_BACKUP_TAG_NAME), 2.7.0)
```
andrepestana-aot, line 17: What's "2.7.0". Shouldn't it be a constant? I know it is only used here but it would easier to understand what's this about.  Maybe a comment would be ok.

cditcher (commented):
Looks ok to me.

devops/Makefile
cditcher: Looks like 2.7.0 is the tag for the backup container version: https://github.com/BCDevOps/backup-container/tags but good question, a comment would be a good idea as I didn't understand at first either.

andrepestana-aot:
> Upgrade to latest version of backup container

What was the issue?

devops/Makefile
guru-aot: The reason is as @cditcher mentioned, added a comment above to explain it. Also the assignment is a default value and its similar to mocking the package.json versioning, which represents that the tags minimum version used for proper working of the container.

guru-aot:
> > Upgrade to latest version of backup container
> 
> What was the issue?

The users which are used for backup have their privileges removed for CONNECT ON DATABASE, USAGE ON SCHEMA, SELECT ON ALL TABLES and USAGE, SELECT ON ALL SEQUENCES for our schema sims. So I have provided the necessary permission and it was able to do the backup successfully.

Note: As I have mentioned, the user that is used to do db backup should NOT be a superuser and the user who does restore should be a super user.

cditcher (approved):

devops/Makefile
```diff
@@ -14,6 +14,8 @@ export NEW_DB := $(or $(NEW_DB), $$NEW_DB)
 export JOB_NAME := $(or $(JOB_NAME), $$JN)
 export SITE_MINDER_LOGOUT_URL := $(or $(SITE_MINDER_LOGOUT_URL), https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi)
 export PATRONI_SIMSDB_BACKUP_APP_NAME := $(or $(PATRONI_SIMSDB_BACKUP_APP_NAME), patroni-simsdb-backup)
+# PATRONI_SIMSDB_BACKUP_TAG_NAME is both the version of the backup container and the tag name created for the build.
```
andrepestana-aot, line 17: "...is either the version of the backup container or the tag name...", right?

andrewsignori-aot (approved):
Nice work, looks good 👍

ann-aot (approved):
👍

andrepestana-aot (approved):
Thanks for the explanation.

sh16011993 (approved):
👍

dheepak-aot (approved):
Looks good. thanks for the walkthrough 👍

Issue #2147, opened 2023-07-28
Student Application Updates: PD/ PPD Form Update on application 


michesmith:
As a Student I want to apply for PD Status on my application too and not just when I first create an account with SABC as my situation may have changed and I need to understand the funding options if I claim PD 'PPD Status or not

**Additional Context**

Account PD/PPD status = False  AND Application PD/PPD question = Yes THEN no funds will go out until one of two events occurs:  
A. student requests change
B. Student auto reassessed at 2 weeks before end date.   

*If a student is already confirmed as PD at the account profile level then we STILL ask again. They have to be YES at Account Level and YES in the application in order to get PD Funding assessment. 
Technical analysis
There are 3 major areas which needs to be updated as part of this change.

1. Submitting PD request on application submit: Based on current PD status of the student and option 1 being selected in new PD question call to ATBC needs to happen. Should we consider queue for this? if so we need to have a way to handle the requests which failed to be sent to ATBC during their outage

2. Workflow code changes
The workflow derive the value for `studentPDStatus` based on the new question(application dynamic data)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaWtKIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--ae78503238d38898e6a7a127adb0982b4228e814/image.png)

 new logic must be based on `newQuestion.option1`

Note: When we implemented PPD changes #2001 we left the existing(going to be obsolete) columns to not disturb the workflow logic. Now that workflow logic is being modified, should we create a seperate ticket to do the cleanup ?

3. Ecert Logic
Ecert validation must ensure that if option1 is selected in new question, then the PDStatus must be approved.
Possibly we must save the studentPDStatus variable to workflow_data.

**Note for Devs**: 
The functionality listed  below is correct , however the content wording may change pending usability testing. The Red arrow below indicates where a new section for "Disability Information" Is to be inserted
![image.png](https://images.zenhubusercontent.com/64b6f330bf2c3f481d13bc8d/2e82bdae-e98a-4cc8-8167-0bb5275f200f)

**Acceptance Criteria**
- [x] Add A new section header is added within the ‘personal information tab’ which is called “Disability Information”  
- [x] Under that header add the question, "Are you a student with disabilities and wish to be assessed for disability funding?”. This question should have the following three options: 
[...]

PR #2460, opened 2023-10-26
#2147 - Student Application Updates: PD/ PPD Form Update on application - part time issue fix

ann-aot:
- While copying and pasting there was a filter for full-time in the new disability section, because of that the disability section was not showing for part-time applications.
- Full-time
image
image
- Part-time
image
image
- When we copy and paste an existing radio component, and if they have any input with the same value(key), then both inputs with the same value have the same HTML `id`. eg, there is a radio input question with 2 choices with value example 1.`yes` 2. `no`. and when we copy and paste this radio component to create a new question 2, with 2 choices with value 1. `yes` and 2. `ggg`. Then question 1 choice 1 and qn 2 choice 1 have the same HTML id, because when you select question 2 choice 1, then sometimes question 1 choice 1 is also selected. As a fix, I removed the copied radio component and added a new radio component for the question.
image


sh16011993 (approved):
LGTM @ann-aot 👍

andrewsignori-aot (approved):
Looks good 👍

dheepak-aot (approved):
Nice catch. Looks good 👍

guru-aot (approved):
LGTM, nice work @ann-aot

Issue #2102, opened 2023-07-11
Establishing independent status for Youth in care students


JasonCTang:
**Describe the task**
Revision of the "former youth in care" question (refer to this as Parent Question) and introduce new sub-question for anyone who answers ‘yes.'

Current Parent Question: 
"Were you ever a youth in the continuing custody of a director of child welfare in BC (ward of the court this means the provincial government is/ was your legal guardian)? "

Requested Change to Parent Questions:
"Were you ever considered a child or youth under government care" with tooltip that reads "Ministry of Children and Family Development (MCFD) or Indigenous Child and Family Service Agency [ICFSA] care, care through the Ministry of Social Development and Poverty Reduction [SDPR] Child in the Home of Relative Program or government care through a Canadian province or territory outside of B.C.

New question: (this should only be presented to the applicant if they answer YES to parent question.
“Are you, or were you at the time of your 19th birthday, in continuing care or custody of a director of child welfare in B.C.?"

Also add an associated tooltip: 
“This means that the provincial government is currently your legal guardian if you are younger than 19 or was your legal guardian at the time of your 19th birthday”

If student answers no, no effect. 
 - No effect means that we still follow the rules of answering YES to parent question.   
 - There will be no impact on the family size calculations.  If student has dependents, it will be counted as normal.

If student answers ‘yes’ AND the student is dependent status, convert student to independent status and treated as independent for the entire assessment workflow. This independent status would overrule any other application logic that determines the student to be a dependent. *Answering 'yes' to second question has no impact on student's exemption from StudentFixedContribution that was determined by virtue of answering yes to the Parent question

This is the same for both Part-Time and Full-Time

***Acceptance Criteria***
- [x]  Check workflow where dependent vs independent is applied.
- [x]  Update the form to the current question. 
- [x]  If student answers Yes to secondary question, convert student to independent status and treat as independent for entire assessment workflow.
- [x]  Only override dependent to single independent. (i.e. if student is married, preserve them as maried)
- [x]  Question only presents to applicant in form if they answer ‘yes’ to being a former youth in care. 
- [x]  Add tool tip to question.

ann-aot:
As part of this ticket there is no workflow changes. The `dependantstatus` logic is set the form itself, so as part of the ticket the `dependantstatus` logic is updated in the form, which will used in the workflow for the calculation.

PR #2464, opened 2023-10-31
#2102 - Establishing independent status for Youth in care students

ann-aot:
- Updated current youth in care qn with new qn with tooltip
image
-  when "yes" is selected for the parent qn, then a subquestion will appear (it also has a tooltip),
image
- When the student selects "yes" for the subquestion, consider the student as "independent", (that means, they won't have a parent tab)
- case 1
image
image
image
image
- case 2
image
image
image
image
- case 3
image
image
image
image
- Updated the `showParentInformation` logic in forms, which is used to determine the `dependant status` in the form, which is used in workflow
- As part of this workflow, there are no Camunda updates.
- The parent qn (youthInCare) is still used in the fulltime-assessment for below logic,
image


sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -10169,6 +10169,70 @@
               "id": "en74evs",
               "defaultValue": ""
             },
+            {
+              "clearOnHide": false,
```
andrewsignori-aot, line 10173: I do not recall the impact of this being `false`. I guess that if the component is hidden it will not be part of the payload at all and the only effect would be if the user answers "Yes" to the parent question it will be displayed with the last value populated. Does it make sense?

sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -10169,6 +10169,70 @@
...
+              "key": "everInCustodyOfChildWelfareInformationPanel",
```
andrewsignori-aot, line 10174: Does `ever` here can potentially imply "always" which would change the meaning of the variable name?
Maybe just `custodyOfChildWelfareInformationPanel` would be enough.

sources/packages/forms/src/form-definitions/sfaa2021-22.json
ann-aot: mm, I am wondering why this value is `false` in definition, In formio, its checked, 
image

sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -15531,7 +15595,7 @@
           "encrypted": false,
           "redrawOn": "data",
           "customDefaultValue": "",
-          "calculateValue": "if (data.hasDependents === 'no' && (data.relationshipStatus === 'single') && data.outOfHighSchoolFor4Years === 'no' && data.howWillYouBeAttendingTheProgram === \"Full Time\") {\n  instance.setValue (true);\n}\nelse {\n  instance.setValue (false);\n}",
+          "calculateValue": "if (\r\n  data.hasDependents === \"no\" &&\r\n  data.relationshipStatus === \"single\" &&\r\n  data.outOfHighSchoolFor4Years === \"no\" &&\r\n  data.howWillYouBeAttendingTheProgram === \"Full Time\" &&\r\n  !(data.everInCustodyOfChildWelfare === \"yes\")\r\n) {\r\n  instance.setValue(true);\r\n} else {\r\n  instance.setValue(false);\r\n}",
```
andrewsignori-aot, line 15598: Just pasting the code below to make the discussion easier 😄 
Is there any particular reason to not use the "no" in the condition, for instace.
```js
data.everInCustodyOfChildWelfare === "no"
```

```js
if (
  data.hasDependents === "no" &&
  data.relationshipStatus === "single" &&
  data.outOfHighSchoolFor4Years === "no" &&
  data.howWillYouBeAttendingTheProgram === "Full Time" &&
  !(data.everInCustodyOfChildWelfare === "yes")
) {
  instance.setValue(true);
} else {
  instance.setValue(false);
}
```

sources/packages/forms/src/form-definitions/sfaa2021-22.json
ann-aot: Got it, this is the panel, for the panel there is no option to `hide on clear`, but I will update the form definition to `true`

andrewsignori-aot (commented):
Great work and analysis. Just minor comments 😉

sources/packages/forms/src/form-definitions/sfaa2021-22.json
dheepak-aot, line 15598: Minor. I see a warning in form editor. 
image
 
we can use `data.everInCustodyOfChildWelfare !== "yes"` instead

sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -10071,21 +10071,21 @@
               "breadcrumb": "default",
               "id": "ef3gfa",
               "addons": [],
-              "lazyLoad": false
+              "lazyLoad": false,
+              "tags": []
             },
             {
-              "label": "<strong>Were you ever a youth in continuing care or custody of a director of child welfare in BC (ward of the court – this means the provincial government is/was your legal guardian)? </strong>",
+              "label": "<strong>Were you ever considered a child or youth under government care?</strong>",
```
dheepak-aot, line 10078: Thanks for stating the workflow reference in the PR description. 

image

I have a question here to business. 

Should the condition in workflow change from `studentDataYouthInCare === "yes"` to `(studentDataYouthInCare === "yes" && studentDataEverInCustodyOfChildWelfare === "yes")`  ?

dheepak-aot:
Good work @ann-aot . Just 2 comments.

sources/packages/forms/src/form-definitions/sfaa2021-22.json
ann-aot: no, as mentioned in the PR description, for the above logics, the parent question is to be considered,, That's what Hugh told me.

andrewsignori-aot: Checked with @HRAGANBC during development and the `studentDataEverInCustodyOfChildWelfare` should not affect the workflow. The checks still must happen based on `studentDataYouthInCare`.

sh16011993 (approved):
Nice work @ann-aot 👍

sources/packages/forms/src/form-definitions/sfaa2021-22.json
ann-aot: As discussed, this is for the panel, so Iam not updating it. It has already added for new subquestion.

sources/packages/forms/src/form-definitions/sfaa2021-22.json
ann-aot: update the key for new subquestion as well

sources/packages/forms/src/form-definitions/sfaa2021-22.json
dheepak-aot: Thanks for the clarification

dheepak-aot (approved):
Thanks for making the change 👍

andrewsignori-aot (approved):
Thanks for doing the changes @ann-aot even the minor ones, looks good 👍

andrepestana-aot (approved):
Looks good! Good job!

guru-aot (approved):
LGTM, nice work @ann-aot

Joshua-Lakusta:
[Open Requirement in TestQuality](https://web.testquality.com/site/gov2/project/16463?requirementId=16056)

Issue #2112, opened 2023-07-13
Business validations for Bulk Withdrawal Upload


JasonCTang:
**Describe the task**
Business validations for Bulk Withdrawal Upload.
Assuming the file parsing and file format validation is already done the dependency ticket https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2111

**Reference figma Design**
https://www.figma.com/file/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?type=design&node-id=12566-163067&mode=design&t=LyYYv68zdOfihIDs-0

**Additional Context**

- File Layout: [https://app.zenhub.com/files/319429847/d246b205-1c6f-449d-b998-b8c75d8e08b2/download](https://app.zenhub.com/files/319429847/d246b205-1c6f-449d-b998-b8c75d8e08b2/download)
- File is based off all applications that have had an assessment since the last file was sent
- Workaround: Institutions that use a bulk withdrawal file still have the ability to enter withdrawals through Insitution Portal.
- Process each record as if it were an manually entered withdrawal. 
- Although this is an "integration" file, it currently still requires a manual process to be executed by the Institution user in SFAS where they have the opportunity to review the warnings prior to committing the changes to SFAS. Therefore, we are suggesting the same approach as the Bulk Offering Upload.
- Each record are proceeded as if it were a manually entered withdrawal.
- Only BC Public institutions are allowed to use and perform this upload.
- Only Institution Administrator role are allowed to perform this upload.
- Affected Student Application belongs to the institution uploading the file.
- Follow the same approach as bulk offering file upload, only process file if all records pass validation (all or nothing), display issues on the page, etc.

**Acceptance Criteria**
- [ ] Ensure that the affected Student Application belongs to the institution uploading the file.
- [ ] If the application number is not found, assume it is an SFAS application and just skip over it, do not perform any other validations on the record and proceed onto the next record to continue processing the file.  Generate warning to user: "Application number not found.  The withdrawal request for this application will not be processed."
- [ ] Create an interface in the SIMS application to upload the file, similar to the bulk offering file upload, using the same validation models structure.
    - [ ] In case one or multiple validations failed, return the validation errors.
    - [ ] Allow the process to be executed as validation only, that will check for the next for possible business validation errors but should not persist any data to the database. 
[...]

PR #2465, opened 2023-11-01
#2112 - Business validations: Application Bulk Withdrawal

sh16011993:
## As a part of this PR, the following tasks were completed:

- [x]  Checked to ensure that the affected Student Application belongs to the institution uploading the file.
- [x]  In case one or multiple validations failed, the validation errors are returned. Below are the validation criteria:
-  If the application number is not found, assuming it is an SFAS application, it is just skipped and the rest of the file is processed. A warning is generated to the user: "Application is not present in SIMS and is a part of the SFAS system."
-  Validated if the combination of the SIN, Application number and institution matched with a record on SIMS. If not, returned the following line as a warning message "The record was not found and will be skipped."
-  If the file contains a line for an application that already has a withdrawal recorded, returned the line as a warning message: "This application is already withdrawn with the date: [withdrawal date]."
- Verified that the application is in "Completed" status and not archived to execute the withdrawal, if not, returned an error with the message "The application is not in the completed status." or "The application is already archived and cannot be withdrawn."
- [x]  For every single file line that has an error or warning, a summary is displayed:
-  Line number: record number including the header in the count.
-  Application Number (even if invalid): displayed to show which "Application Number" was interrupted.
-  Withdrawal Date (even if invalid): displayed to show the "Withdrawal Date" of the interrupted "Application Number".
-  Validations: displayed to show a user-friendly message about why the record is not considered valid.
- [x] Centralized the error validation code to the `validation-utils` file.

### Screenshots [validation errors and warnings]:

image

------------------------------------------------------------------------------------------------------------------------------------------

image

------------------------------------------------------------------------------------------------------------------------------------------

image

------------------------------------------------------------------------------------------------------------------------------------------

image

------------------------------------------------------------------------------------------------------------------------------------------

image

------------------------------------------------------------------------------------------------------------------------------------------

image


sources/packages/backend/apps/api/src/app.institutions.module.ts
```diff
@@ -166,6 +168,8 @@ import {
     ApplicationOfferingChangeRequestService,
     ApplicationOfferingChangeRequestControllerService,
     ApplicationWithdrawalImportTextService,
+    ApplicationBulkWithdrawalImportBusinessService,
```
andrewsignori-aot, line 171: I believe that we do not use the word "Business" in the services. The context can be inferred by where they are imported. Either way, if we follow that asa pattern a great amount of our services would need to be renamed.
If there is a benefit, please let me know 😉

sources/packages/backend/apps/api/src/constants/error-code.constants.ts
```diff
@@ -150,3 +150,17 @@ export const APPLICATION_WITHDRAWAL_TEXT_CONTENT_FORMAT_ERROR =
  */
 export const APPLICATION_WITHDRAWAL_INVALID_TEXT_FILE_ERROR =
   "APPLICATION_WITHDRAWAL_INVALID_TEXT_FILE_ERROR";
+/**
+ * An error was found during the application bulk withdrawal validation and the process
+ * must be stopped. Every offering validation will generate an error but
```
andrewsignori-aot, line 155: Please review the comment and remove the "offering" from the context.

sources/packages/backend/apps/api/src/constants/error-code.constants.ts
```diff
@@ -150,3 +150,17 @@ export const APPLICATION_WITHDRAWAL_TEXT_CONTENT_FORMAT_ERROR =
...
+ * some errors are classified as warnings and therefore are not considered
+ * critical, which means that critical errors are the only ones that will
+ * cause the process to be interrupted.
+ */
+export const APPLICATION_BULK_WITHDRAWAL_VALIDATION_CRITICAL_ERROR =
+  "APPLICATION_BULK_WITHDRAWAL_VALIDATION_CRITICAL_ERROR";
+/**
+ * One or more application withdrawal business validation errors have occurred.
+ */
+export const APPLICATION_WITHDRAWAL_BUSINESS_VALIDATION_ERROR =
```
andrewsignori-aot, line 165: Can we follow similar naming used for offerings and avoid using "business"?

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -196,14 +201,23 @@ export class ScholasticStandingInstitutionsController extends BaseController {
...
   ): Promise<PrimaryIdentifierAPIOutDTO[]> {
     // Read the entire file content.
     const fileContent = file.buffer.toString();
+    // Get the file header model.
+    let textHeader: ApplicationBulkWithdrawalHeader;
     // Convert the file raw content into text models.
     let textModels: ApplicationWithdrawalImportTextModel[];
     try {
+      textHeader =
+        this.applicationWithdrawalImportTextService.readText(
```
andrewsignori-aot, line 214: Please avoid calling `this.applicationWithdrawalImportTextService.readText` twice.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -196,14 +201,23 @@ export class ScholasticStandingInstitutionsController extends BaseController {
...
     @Query("validation-only", new DefaultValuePipe(false), ParseBoolPipe)
     validationOnly: boolean,
+    @UserToken() userToken: IInstitutionUserToken,
   ): Promise<PrimaryIdentifierAPIOutDTO[]> {
     // Read the entire file content.
     const fileContent = file.buffer.toString();
+    // Get the file header model.
+    let textHeader: ApplicationBulkWithdrawalHeader;
     // Convert the file raw content into text models.
     let textModels: ApplicationWithdrawalImportTextModel[];
```
andrewsignori-aot, line 211: Is there a reason to prefer this approach instead of using the `FileData` object returned from `this.applicationWithdrawalImportTextService.readText`?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    });
+    const businessValidationModels: ApplicationBulkWithdrawalImportBusinessValidationModel[] =
+      [];
+    textValidations.forEach((textValidation) => {
+      const businessValidationModel =
+        {} as ApplicationBulkWithdrawalImportBusinessValidationModel;
+      const applicationData =
+        applicationSINDataMap[textValidation.textModel.applicationNumber];
+      if (applicationData) {
+        businessValidationModel["applicationFound"] = true;
```
andrewsignori-aot, line 64: Is there a reason to access the properties as `string` when we have the type available? If not, please use the type.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  ApplicationData,
+} from "./application-bulk-withdrawal-import-business-validation.models";
+
+type ApplicationSINDataMap = Record<string, ApplicationData>;
+/**
+ * Handles the application withdrawal bulk insert preparation.
+ */
+@Injectable()
+export class ApplicationBulkWithdrawalImportBusinessService extends RecordDataModelService<Application> {
+  constructor(private readonly dataSource: DataSource) {
```
andrewsignori-aot, line 20: Please use the `@InjectRepository` for new classes.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  })
+  recordType: RecordType;
+  /**
+   * SIN.
+   */
+  @ValidateIf(
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  @IsNumberString(
```
andrewsignori-aot, line 111: Can we use `@IsValidSIN()` instead?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    return businessValidationModels;
+  }
+
+  /**
+   * Get the SIN and location information for each of the provided applications.
+   * @param applicationNumbers application numbers for which the SIN and location information needs to be retrieved.
+   * @param institutionId institution id.
+   * @returns applications containing the required information.
+   */
+  async getSINForApplications(
```
andrewsignori-aot, line 100: The name of the method does not seem to be aligned with what it really does.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    applicationNumbers: string[],
+    institutionId: number,
+  ): Promise<Application[]> {
+    return await this.repo.find({
```
andrewsignori-aot, line 104: The `await` is not needed in this case.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
andrewsignori-aot, line 100: If this method is not intended to be shared by other services it can be marked as private. maybe the mapping part can happen here or is another private method to reduce the amount of line in `generateBusinessValidationModels`.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+        studentScholasticStandings: true,
+      },
+      where: {
+        applicationNumber: In(applicationNumbers),
+        location: { institution: { id: institutionId } },
+      },
+    });
+  }
+
+  @InjectLogger()
```
andrewsignori-aot, line 126: Not a blocker but we can have it added as needed.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+        isArchived: record.isArchived,
+        sin: record.student.sinValidation.sin,
+        locationId: record.location.id,
+        locationCode: record.location.institutionCode,
+        hasPreviouslyBeenWithdrawn: !!record.studentScholasticStandings.length,
+      };
+    });
+    const businessValidationModels: ApplicationBulkWithdrawalImportBusinessValidationModel[] =
+      [];
+    textValidations.forEach((textValidation) => {
```
andrewsignori-aot, line 58: Not a blocker but it seems that it can be done as a `map`.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  ApplicationWithdrawalTextValidationResult,
+} from "./application-bulk-withdrawal-import-text.models";
+import { DataSource, In } from "typeorm";
+import { Application, RecordDataModelService } from "@sims/sims-db";
+import {
+  ApplicationBulkWithdrawalImportBusinessValidationModel,
+  ApplicationData,
+} from "./application-bulk-withdrawal-import-business-validation.models";
+
+type ApplicationSINDataMap = Record<string, ApplicationData>;
```
andrewsignori-aot, line 14: It seems that just `ApplicationDataMap` would be more meaningful.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+/**
```
andrewsignori-aot, line 15: Please add a blank line above the comment.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
```diff
@@ -0,0 +1,128 @@
...
+      institutionId,
+    );
+    applicationSINData.forEach((record) => {
+      applicationSINDataMap[record.applicationNumber] = {
+        applicationStatus: record.applicationStatus,
+        isArchived: record.isArchived,
+        sin: record.student.sinValidation.sin,
+        locationId: record.location.id,
+        locationCode: record.location.institutionCode,
+        hasPreviouslyBeenWithdrawn: !!record.studentScholasticStandings.length,
```
andrewsignori-aot, line 53: Withdrawal is one of the types of scholastic standings. We should be probably checking by "Student withdrew from program". Maybe we can restrict the SQL query to return only the specific type.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
sh16011993: Changed it. But I still need the header model (the institution code) for validation purpose.

sources/packages/web/src/views/institution/WithdrawalUpload.vue
```diff
@@ -224,6 +224,10 @@ export default defineComponent({
             },
           );
         if (uploadResults.length) {
+          // Check for any errors. If no errors found (only warnings found), show success message
```
andrewsignori-aot, line 227: Please add a period to the sentence.

sources/packages/web/src/views/institution/WithdrawalUpload.vue
```diff
@@ -287,7 +290,6 @@ export default defineComponent({
       BannerTypes,
       uploadForm,
       ACCEPTED_FILE_TYPE,
-      //textFileUploadKey,
```
andrewsignori-aot, line 290: ❤️

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+   * Warnings, if any.
+   */
+  warnings: ValidationResult[];
+  /**
+   * Users friendly errors list, if any.
+   */
+  errors: string[];
+}
+
+export interface ApplicationData {
```
andrewsignori-aot, line 303: Missing comments.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  sin: string;
+  locationId: number;
+  locationCode: string;
+  applicationStatus: ApplicationStatus;
+  isArchived: boolean;
+  hasPreviouslyBeenWithdrawn: boolean;
+}
```
andrewsignori-aot, line 310: None of those properties would be potentially `null`?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Results of the application bulk withdrawal model validation.
+ */
+export interface ApplicationWithdrawalValidationResult {
+  /**
+   * Record index in the list of records. Headers are not considered.
+   */
+  index: number;
+  /**
+   * validated application bulk withdrawal model.
```
andrewsignori-aot, line 290: Please start the sentence with an upper case character.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  isCompleted: "be completed",
+  isArchived: "Is archived",
+  archived: "archived",
+  recordMatch: "Record match",
+  record: "The record was not found",
+};
+
+/**
+ * Possible warnings unique identifiers.
+ * !These keys are also consumed in the UI to display/hide warning banners.
```
andrewsignori-aot, line 42: This comment does not apply to the withdraw, right?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  SIN_NUMBER_LENGTH,
+} from "@sims/sims-db";
+import {
+  ApplicationBulkWithdrawalHeader,
+  ApplicationWithdrawalImportTextModel,
+  DataTextHeaders,
+  RecordType,
+} from "./application-bulk-withdrawal-import-text.models";
+
+const userFriendlyNames = {
```
andrewsignori-aot, line 24: The idea of having a similar approach for the offerings was to make field names, that are internally named differently, to be more "user friendly". It seems that the idea here is not the same.
Also, please review if all the items are used, for instance, I believe that `isCompleted` is not used.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  hasCorrectInstitutionCode: "Has correct institution code",
+  institutionCode: "The institution code",
+  hasPreviouslyBeenWithdrawn: "Has previously been withdrawn",
+  applicationStatus: "Application status",
+  isCompleted: "be completed",
+  isArchived: "Is archived",
+  archived: "archived",
+  recordMatch: "Record match",
+  record: "The record was not found",
+};
```
andrewsignori-aot, line 38: The idea of having a similar approach for the offerings was to make field names, that are internally named differently, to be more "user friendly". It seems that the idea here is not the same.
Also, please review if all the items are used, for instance, I believe that `isCompleted` is not used.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+   * Data record type.
+   */
+  @ValidateIf(
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  @Equals(RecordType.ApplicationBulkWithdrawalDataRecordType, {
+    message: `${DataTextHeaders.recordType} must be valid.`,
+  })
+  recordType: RecordType;
```
andrewsignori-aot, line 103: Do we need the `recordType` for these validations?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  /**
+   * SIN.
+   */
+  @ValidateIf(
```
andrewsignori-aot, line 107: I do not see a problem allowing the SIN validation message to be displayed if the application was not found. The SIN came from the imported file and can be validated on its own, right?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  @IsNumberString(
+    { no_symbols: true },
+    {
+      message: `${DataTextHeaders.applicationNumber} must be a valid number.`,
+    },
+  )
+  @Length(APPLICATION_NUMBER_LENGTH, APPLICATION_NUMBER_LENGTH, {
```
andrewsignori-aot, line 134: Same https://github.com/bcgov/SIMS/pull/2465/files#r1379325084

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+  })
+  withdrawalDate: Date;
+  /**
+   * Application found.
+   */
+  @IsBoolean({
+    message: `${userFriendlyNames.applicationFound} must be a boolean value.`,
+  })
+  @IsIn([true], {
+    message: `${userFriendlyNames.application} is not present in SIMS and is a part of the SFAS system.`,
```
andrewsignori-aot, line 156: Please let's ensure that these messages are reviewed by @HRAGANBC and/or @JasonCTang and persisted to the ticket to allow later QA validations also.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+      object.applicationFound,
+  )
+  @IsDate({
+    message: `${DataTextHeaders.withdrawalDate} must be a valid withdrawal date.`,
+  })
+  withdrawalDate: Date;
+  /**
+   * Application found.
+   */
+  @IsBoolean({
```
andrewsignori-aot, line 152: For the information that we are producing (not a user input), do we need to validate the type? Same for others.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+   * Valid SIN.
+   */
+  @ValidateIf(
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  @IsBoolean({
+    message: `${userFriendlyNames.validSIN} must be a boolean value.`,
+  })
+  validSIN: boolean;
```
andrewsignori-aot, line 186: I would assume that `valid` here means the provided one matches the student one, right?
If it is right, can we have it renamed to `studentSINMactch` or similar?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,310 @@
...
+   */
+  @ValidateIf(
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  @IsEnum(ApplicationStatus)
+  @IsIn([ApplicationStatus.Completed], {
+    message: `${userFriendlyNames.application} is not in the ${ApplicationStatus.Completed} status.`,
+  })
+  applicationStatus: ApplicationStatus;
```
andrewsignori-aot, line 212: If the application is not found many properties will be potentially not defined, rght?
The properties in the model must express it.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.service.ts
```diff
@@ -0,0 +1,149 @@
...
+              .applicationNumber,
+          withdrawalDate:
+            validation.applicationBulkWithdrawalBusinessValidationModel
+              .withdrawalDate,
+          sin: validation.applicationBulkWithdrawalBusinessValidationModel.sin,
+          errors: validation.errors,
+          warnings: validation.warnings,
+          infos: [],
+        }));
+      throw new BadRequestException(
```
andrewsignori-aot, line 104: We do not throw HTTP errors from the service layer, only from `controller services`.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Inspect the validation error and its children when an error
+   * happen as it can have an additional context (must be considered a warning)
+   * or it is a critical error (has no warning context).
+   * @param error error to be inspected.
+   * @returns errors and warnings.
+   */
+  private extractValidationsByType(error: ValidationError): {
```
andrewsignori-aot: Is there a reason to have this method duplicated instead of centralized and reused for offerings and withdrawals?

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -196,13 +202,14 @@ export class ScholasticStandingInstitutionsController extends BaseController {
     @UploadedFile() file: Express.Multer.File,
     @Query("validation-only", new DefaultValuePipe(false), ParseBoolPipe)
     validationOnly: boolean,
+    @UserToken() userToken: IInstitutionUserToken,
   ): Promise<PrimaryIdentifierAPIOutDTO[]> {
     // Read the entire file content.
     const fileContent = file.buffer.toString();
-    // Convert the file raw content into text models.
-    let textModels: ApplicationWithdrawalImportTextModel[];
+    // Get the file data.
```
andrewsignori-aot, line 209: I do not believe that the comment is adding any value here. If you would like to keep it maybe move above the like 212.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -196,13 +202,14 @@ export class ScholasticStandingInstitutionsController extends BaseController {
...
+    let fileData: FileData;
```
andrewsignori-aot, line 210: I am struggling with the `FileData` as generic as it is.
Maybe it is named something like `BulkWithdrawalFileData` the line 240 would be something like `withdrawalFileData.models,` or even `withdrawalFileData.records` considering that the header is also a model.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -61,10 +63,8 @@ import {
   uploadLimits,
 } from "../../utilities";
 import { PrimaryIdentifierAPIOutDTO } from "../models/primary.identifier.dto";
-import {
-  ApplicationWithdrawalImportTextModel,
-  ApplicationWithdrawalTextValidationResult,
-} from "../../services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.models";
+import { ApplicationWithdrawalTextValidationResult } from "../../services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.models";
+import { FileData } from "../../services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models";
```
ann-aot, line 67: How about adding these files to the index and import from there?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
sh16011993: @andrewsignori-aot Is it better to use `map` compared to `forEach` here? I felt `forEach` to be a better fit here since we are populating `businessValidationModels` array rather than updating `textValidations` array. Please let me know if I missed something.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
sh16011993: But the application could have been withdrawn for any of those 4 reasons and if it was withdrawn earlier for any reason, it would be the same behavior, right? Please let me know if I am missing something.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
sh16011993: Nope, I don't think so.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
sh16011993: I have removed this part.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+type ApplicationDataMap = Record<string, ApplicationData>;
+
+/**
+ * Handles the application withdrawal bulk insert preparation.
+ */
+@Injectable()
+export class ApplicationBulkWithdrawalImportService {
+  constructor(
+    private readonly dataSource: DataSource,
```
ann-aot, line 22: its not used

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
sh16011993: Same for the `withdrawalDate` ?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,127 @@
...
+        locationCode: record.location.institutionCode,
+        hasPreviouslyBeenWithdrawn: !!record.studentScholasticStandings.length,
+      };
+    });
+    const businessValidationModels: ApplicationBulkWithdrawalImportBusinessValidationModel[] =
+      [];
+    textValidations.forEach((textValidation) => {
+      const businessValidationModel =
+        {} as ApplicationBulkWithdrawalImportBusinessValidationModel;
+      const applicationData =
```
ann-aot, line 65: `applicationData` name is already used in line 45, pls use a different variable

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,127 @@
...
+        applicationDataMap[textValidation.textModel.applicationNumber];
+      if (applicationData) {
+        businessValidationModel.applicationFound = true;
+        businessValidationModel.applicationBelongsToInstitution =
```
ann-aot, line 69: I didn't get this, what does institution admin have to do with `applicationBelongsToInstitution` here?

ann-aot: if the plan is to check if the application belongs to the institution, it is already achieved in the query where clause right? line 122 - 123

andrewsignori-aot: Here I would need some clarification also. If the query is also checking if the institution, an application not found can be because ii belongs to SFAS or because it does not belong to the institution. Is it right to assume that we do not need to have this distinction?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,127 @@
...
+        textValidation.textModel.applicationNumber;
+      businessValidationModel.withdrawalDate =
+        textValidation.textModel.withdrawalDate;
+      businessValidationModels.push(businessValidationModel);
+    });
+    return businessValidationModels;
+  }
+
+  /**
+   * Get the SIN and location information for each of the provided applications.
```
andrewsignori-aot, line 97: The comment is potentially misleading as it indicates SIN and location but it gets more than data and all data retrieved seems to bo equally important.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
andrewsignori-aot: Same for every single one that is under our control: applicationFound, applicationBelongsToInstitution, studentSINMatch, hasCorrectInstitutionCode, isArchived, applicationStatus, hasPreviouslyBeenWithdrawn, isRecordMatch, withdrawalDate.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-validation.service.ts
```diff
@@ -0,0 +1,149 @@
...
+   * Validate offering models and provide the result for every model.
+   * @param businessValidationModels application bulk withdrawal business validation models to be validated.
+   * @param silently if true, no exception is generated case the validation fails
+   * and the success or failure can be determined from the result object.
+   * @returns validation result or an exception in the case of a failed validation
+   * combined with the silently parameter defined as false.
+   */
+  validateApplicationBulkWithdrawalModels(
+    businessValidationModels: ApplicationBulkWithdrawalImportBusinessValidationModel[],
+    silently = false,
```
ann-aot, line 35: what is the benefit of this variable, I see this method is only used once

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,127 @@
...
+   * @param applicationNumbers application numbers for which the SIN and location information needs to be retrieved.
+   * @param institutionId institution id.
+   * @returns applications containing the required information.
+   */
+  private async getApplicationBulkWithdrawalValidationDetails(
```
andrewsignori-aot, line 102: I like well-descriptive names but inside a service named `ApplicationBulkWithdrawal` do we need a name named `getApplicationBulkWithdrawalValidationDetails`. Can it be just `getApplicationValidationDate`?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-validation.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Validate offering models and provide the result for every model.
+   * @param businessValidationModels application bulk withdrawal business validation models to be validated.
+   * @param silently if true, no exception is generated case the validation fails
+   * and the success or failure can be determined from the result object.
+   * @returns validation result or an exception in the case of a failed validation
+   * combined with the silently parameter defined as false.
+   */
+  validateApplicationBulkWithdrawalModels(
+    businessValidationModels: ApplicationBulkWithdrawalImportBusinessValidationModel[],
```
andrewsignori-aot, line 34: If `business` is removed everywhere I believe that it would not reduce the meaning.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,248 @@
...
+ */
+export interface ApplicationWithdrawalValidationResult {
+  /**
+   * Record index in the list of records. Headers are not considered.
+   */
+  index: number;
+  /**
+   * Validated application bulk withdrawal model.
+   */
+  applicationBulkWithdrawalBusinessValidationModel: ApplicationBulkWithdrawalImportBusinessValidationModel;
```
andrewsignori-aot, line 227: Can we have `applicationBulkWithdrawalBusinessValidationModel` renamed to `validationModel`?

andrewsignori-aot (commented):
Good work. Please take a look at the comments.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
andrewsignori-aot: `map` is intended to be used exactly to generate an array based on a second array, transforming the last one, which means, creating the `businessValidationModels` based on the `textValidations` where there will an one-to-one output.
I would not state this as a blocker but I have a feeling that along the solution whatever a `map` can be used we have been opting for it instead of a `foreach`.
Reason that would lead me to use the `foreach`:
- Performance (not the case IMO)
- Code readability (not the case IMO)

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
andrewsignori-aot: The application can be archived for any of the cases but, as far as I know, withdrawal is one of the cases of the Scholastic Standing. For instance, one of the types of Scholastic Standing is called "Early Completion" which is not a withdrawal at all.
Please let me know if I am missing something or please confirm with @JasonCTang or @HRAGANBC as needed.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-validation.service.ts
andrewsignori-aot: @ann-aot I guess that it is supposed to work similarly to offerings where we use it to allow the method to perform a "validation-only" or the "real" import, but I believe that it does not completely follow the same idea, hence a change is needed.

Reference code (please note that it is calling a controller service).
https://github.com/bcgov/SIMS/blob/81923206ba137731f7fd4e3e6f1776369f8a6898/sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.institutions.controller.ts#L551-L555

sources/packages/backend/apps/api/src/app.institutions.module.ts
```diff
@@ -29,6 +29,8 @@ import {
   StudentAppealRequestsService,
   ApplicationOfferingChangeRequestService,
   ApplicationWithdrawalImportTextService,
+  ApplicationBulkWithdrawalImportService,
```
dheepak-aot, line 32: What is the benefit of having this import service, when the actual import already done?

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -230,12 +232,26 @@ export class ScholasticStandingInstitutionsController extends BaseController {
...
+        withdrawalFileData.applicationWithdrawalModels,
       );
     // Assert successful validation.
     this.assertTextValidationsAreValid(textValidations);
-    // TODO Add business validation for application bulk withdrawal.
-
+    const validationModels =
+      await this.applicationWithdrawalImportService.generateApplicationBulkWithdrawalValidationModels(
+        textValidations,
+        withdrawalFileData.header,
```
dheepak-aot, line 242: I believe only the originator is used further, we can just send that alone.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -230,12 +232,26 @@ export class ScholasticStandingInstitutionsController extends BaseController {
...
+        userToken.authorizations.institutionId,
+        userToken.authorizations.adminLocationsIds,
```
dheepak-aot, line 244: Why should we worry about location ids when we have only institution administrators going to do the upload?

image

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
dheepak-aot: The message in AC, "Record was not found and skipped" me assume that

image

sources/packages/backend/apps/api/src/constants/error-code.constants.ts
```diff
@@ -150,3 +150,17 @@ export const APPLICATION_WITHDRAWAL_TEXT_CONTENT_FORMAT_ERROR =
...
 export const APPLICATION_WITHDRAWAL_INVALID_TEXT_FILE_ERROR =
   "APPLICATION_WITHDRAWAL_INVALID_TEXT_FILE_ERROR";
+/**
+ * An error was found during the application bulk withdrawal validation and the process
+ * must be stopped. Every application bulk withdrawal validation can generate an error but
+ * some errors are classified as warnings and therefore are not considered
+ * critical, which means that critical errors are the only ones that will
+ * cause the process to be interrupted.
+ */
+export const APPLICATION_BULK_WITHDRAWAL_VALIDATION_CRITICAL_ERROR =
```
dheepak-aot, line 160: We have the text file format errors with error name `APPLICATION_WITHDRAWAL_TEXT_CONTENT_FORMAT_ERROR` isn't that going to fall under this same category of critical error?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,126 @@
...
+    });
+    const businessValidationModels: ApplicationBulkWithdrawalImportBusinessValidationModel[] =
+      [];
+    textValidations.forEach((textValidation) => {
+      const businessValidationModel =
+        {} as ApplicationBulkWithdrawalImportBusinessValidationModel;
+      const applicationData =
+        applicationDataMap[textValidation.textModel.applicationNumber];
+      if (applicationData) {
+        businessValidationModel.applicationFound = true;
```
dheepak-aot, line 67: the following code will avoid the if else
```ts
businessValidationModel.applicationFound = !!applicationData;
```

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,248 @@
...
+  })
+  applicationBelongsToInstitution?: boolean;
+  /**
+   * Valid SIN.
+   */
+  @ValidateIf(
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  studentSINMatch?: boolean;
```
dheepak-aot, line 133: `applicationFound` and this `studentSINMatch` has same effect as the the AC. Why is this not treated as warning?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,248 @@
...
+   * Has correct institution code.
+   */
+  @ValidateIf(
+    (object: ApplicationBulkWithdrawalImportBusinessValidationModel) =>
+      object.applicationFound,
+  )
+  @IsIn([true], {
+    message: "The institution code provided is incorrect.",
+  })
+  hasCorrectInstitutionCode?: boolean;
```
dheepak-aot, line 144: https://github.com/bcgov/SIMS/pull/2465/files#r1380689890

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
```diff
@@ -0,0 +1,248 @@
...
+export class ApplicationBulkWithdrawalImportBusinessValidationModel {
+  /**
+   * SIN.
+   */
+  @IsValidSIN({
+    message: `${DataTextHeaders.sin} must be a valid SIN.`,
+  })
+  @Length(SIN_NUMBER_LENGTH, SIN_NUMBER_LENGTH, {
+    message: `${DataTextHeaders.sin} must be of length ${SIN_NUMBER_LENGTH}.`,
+  })
```
ann-aot, line 82: Just wondering, this validation will come under line 77 right? so, do we need it?

dheepak-aot:
Nice first effort on bulk withdrawals. Please have a look at the comments. I would advise to get clarity on level of details on warning messages to be displayed and refactor code if required.

ann-aot (commented):
nice work @sh16011993 added some comments, most of the other comments are addressed captured by the devs

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
sh16011993: Changed it.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-validation.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Inspect the validation error and its children when an error
+   * happen as it can have an additional context (must be considered a warning)
+   * or it is a critical error (has no warning context).
+   * @param error error to be inspected.
+   * @returns errors and warnings.
+   */
+  private extractValidationsByType(error: ValidationError): {
```
andrepestana-aot, line 121: Should we call it an "error" or just "log" here because it can just be a warning or info, right?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-validation.service.ts
sh16011993: Removed it.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
sh16011993: hmm... but the scope of this one in line 65 is limited to the callback function, so it shouldn't be a problem. However, to keep clarity, I am renaming the one in line 45 to `applicationValidationData`.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
sh16011993: Removed this property.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-validation.service.ts
andrewsignori-aot: From the `class-validator` perspective, from where the object `ValidationError` came, for instance, it is a validation error, so I believe that we can keep it as it is.
The method comment also starts with the statement: "Inspect the validation error" which is aligned with the parameter.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
sh16011993: removed it.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        validationModel.applicationFound = false;
+      }
+      validationModel.sin = textValidation.textModel.sin;
+      validationModel.applicationNumber =
+        textValidation.textModel.applicationNumber;
+      validationModel.withdrawalDate = textValidation.textModel.withdrawalDate;
+      return validationModel;
+    });
+    return validationModels;
```
andrewsignori-aot, line 80: Minor it can be returned directly without using the `validationModels` variable.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
andrewsignori-aot: Does the `applicationStatus` need the `@IsEnum(ApplicationStatus)`?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,114 @@
...
+      institutionId,
+    );
+    applicationValidationData.forEach((record) => {
+      applicationDataMap[record.applicationNumber] = {
+        applicationStatus: record.applicationStatus,
+        isArchived: record.isArchived,
+        sin: record.student.sinValidation.sin,
+        locationId: record.location.id,
+        locationCode: record.location.institutionCode,
+        hasPreviouslyBeenWithdrawn: !!record.studentScholasticStandings.length,
```
andrewsignori-aot, line 53: This is the property that I am not following and that also supports the comment: https://github.com/bcgov/SIMS/pull/2465#discussion_r1379375219
If the property is named `hasPreviouslyBeenWithdrawn` but is actually checking for any "Scholastic Standing" it does not seem right.

Depending on what the business answered, at least the property must be changed to `hasScholasticStanding` and the message should also be adjusted from "The application is already withdrawn with the date..." to "The application already has a Scholastic Standing".

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.controller.service.ts
```diff
@@ -40,4 +50,38 @@ export class ScholasticStandingControllerService {
...
+        validationResult.errors.length || validationResult.warnings.length,
+    );
+    if (validations.length) {
+      // At least one error or warning was detected.
+      const errorValidationResults: ApplicationBulkWithdrawalValidationResultAPIOutDTO[] =
+        validations.map((validation) => ({
+          recordIndex: validation.index,
+          applicationNumber: validation.validationModel.applicationNumber,
+          withdrawalDate: validation.validationModel.withdrawalDate,
+          sin: validation.validationModel.sin,
```
andrewsignori-aot, line 73: Is there a reason to the SIN be returned if we do not display it in the UI?

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.controller.service.ts
```diff
@@ -40,4 +56,70 @@ export class ScholasticStandingControllerService {
       ...transformToActiveApplicationDataAPIOutDTO(application, offering),
     };
   }
+
+  /**
+   * Verify if all text file validations were performed with success and throw
+   * a BadRequestException in case of some failure.
+   * @param textValidations validations to be verified.
+   */
+  assertTextValidationsAreValid(
```
andrewsignori-aot, line 65: Thanks for moving this from the controller to the controller service 😉

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.institutions.controller.ts
```diff
@@ -230,12 +230,27 @@ export class ScholasticStandingInstitutionsController extends BaseController {
     // Validate the text models.
     const textValidations =
       this.applicationWithdrawalImportTextService.validateTextModels(
-        textModels,
+        withdrawalFileData.applicationWithdrawalModels,
```
andrewsignori-aot, line 233: With the renaming of the class, my intention was to have the properties simplified as, for instance, `withdrawalFileData.records`, since we already have `withdrawalFileData.header`. Does it make sense?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
sh16011993: Sorry, missed it. Removed.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
```diff
@@ -0,0 +1,108 @@
...
+  ) {}
+
+  /**
+   * Generates the application bulk withdrawal validation model.
+   * @param textValidations text validation models to be converted to the validation models.
+   * @param originator institution that sent the file.
+   * @param institutionId institution id.
+   * @returns application bulk withdrawal models to be validated and persisted.
+   */
+  async generateApplicationBulkWithdrawalValidationModels(
```
andrewsignori-aot, line 30: Not a blocker but `generateApplicationBulkWithdrawalValidationModels` inside a service already named `ApplicationBulkWithdrawalImport` can be renamed to `generateValidationModels`.

sources/packages/web/src/views/institution/WithdrawalUpload.vue
```diff
@@ -55,7 +55,7 @@
             :clearable="true"
             :accept="ACCEPTED_FILE_TYPE"
             density="compact"
-            v-model="withdrawalFiles"
+            :v-model="withdrawalFiles"
```
andrewsignori-aot, line 58: Did the ":" add any difference? I did not see anywhere else where it was required or in Vue docs.
If it was needed, can you please elaborate on the why?

sources/packages/backend/apps/api/src/utilities/class-validation/validation-utils.ts
```diff
@@ -61,3 +61,111 @@ function generateFlattenedErrors(
...
 }
+
+/**
+ * Inspect the validation error and its children checking when an error
+ * happen as has an additional context (must be considered a warning or a info)
+ * or it is a critical error (has no warning or info context).
+ * @param error error to be inspected.
+ * @returns errors, warnings and infos.
+ */
+export function extractValidationsByType<T, S>(
```
andrewsignori-aot, line 72: I am not sure if the generics here are bringing the expected benefit.
Maybe we can have the `ValidationResult` code as `string`.

```ts
export interface ValidationResult {
  typeCode: string;
  message: string;
}
```

The `extractValidationsByType` as:
```ts
export function extractValidationsByType(error: ValidationError): {
  errors: string[];
  warnings: ValidationResult[];
  infos: ValidationResult[];
} {
  const errors: string[] = [];
  const warnings: ValidationResult[] = [];
  const infos: ValidationResult[] = [];
  const flattenedErrors = flattenErrors(error);
  flattenedErrors.forEach((error) => {
    Object.keys(error.constraints).forEach((errorConstraintKey) => {
      let associatedContext: ValidationContext<string> = undefined;
      if (error.contexts) {
        associatedContext = error.contexts[
          errorConstraintKey
        ] as ValidationContext<string>;
      }
      const errorDescription = error.constraints[errorConstraintKey];
      if (associatedContext?.type === ValidationContextTypes.Warning) {
        warnings.push({
          typeCode: associatedContext.typeCode,
          message: errorDescription,
        });
      } else if (
        associatedContext?.type === ValidationContextTypes.Information
      ) {
        infos.push({
          typeCode: associatedContext.typeCode,
          message: errorDescription,
        });
      } else {
        errors.push(errorDescription);
      }
    });
  });
  return { errors, warnings, infos };
}
```

And the `ValidationContext` as
```ts
export class ValidationContext<T> {
  /**
   * Creates an error context that will make the error downgrade to
   * a condition of a warning information.
   * @param warningTypeCode warning code that uniquely identifies this condition.
   * @returns the warning context.
   */
  static CreateWarning<T>(warningTypeCode: T): ValidationContext<T> {
    const newContext = new ValidationContext<T>();
    newContext.type = ValidationContextTypes.Warning;
    newContext.typeCode = warningTypeCode;
    return newContext;
  }

  /**
   * Creates an error context that will make the error downgraded to
   * a condition of a simple information.
   * @param infoTypeCode information code that uniquely identifies this condition.
   * @returns the information context.
   */
  static CreateInfo<T>(infoTypeCode: T): ValidationContext<T> {
    const newContext = new ValidationContext<T>();
    newContext.type = ValidationContextTypes.Information;
    newContext.typeCode = infoTypeCode;
    return newContext;
  }

  /**
   * Context type.
   */
  type: ValidationContextTypes;
  /**
   * Unique identifier of the validation error.
   */
  typeCode: T;
}
```

andrewsignori-aot (commented):
Thanks for making the changes so far. I will do one more round of review once the new changes come.

sources/packages/web/src/views/institution/WithdrawalUpload.vue
sh16011993: Nope. It was a mistake. I reverted it back.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
sh16011993: Keeping it as is after the discussion with business [@HRAGANBC]

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business.service.ts
sh16011993: Updated as per today's discussion.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-business-validation.models.ts
sh16011993: Updated the ticket with the messages received from business [@HRAGANBC and @JasonCTang]

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.controller.service.ts
sh16011993: Thanks for suggesting the change.

guru-aot (commented):
Please work on the other dev comments, I do not have anything new.

sources/packages/backend/apps/api/src/route-controllers/student-scholastic-standings/student-scholastic-standings.controller.service.ts
guru-aot: Good one to move

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.service.ts
```diff
@@ -115,6 +133,100 @@ export class ApplicationWithdrawalImportTextService {
...
+      },
+      relations: {
+        location: true,
+        student: { sinValidation: true },
+        studentScholasticStandings: true,
+      },
+      where: {
+        applicationNumber: In(applicationNumbers),
+        location: {
+          institution: { id: institutionId },
```
andrewsignori-aot, line 220: I trying to remember the decision about changing the query to use the location code from the file header in addition to the institution id. Is it still pending?
I can see that there is a message "The institution location code provided is incorrect." for the incorrect location code but should we not let that fall into the same "Not found" scenario?

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import.service.ts
andrewsignori-aot: It is ok to keep it now since the SQL query is changed and `studentScholasticStandings` will contain only the withdrawals.

guru-aot (approved):
LGTM, nice work @sh16011993

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.service.ts
```diff
@@ -115,6 +133,100 @@ export class ApplicationWithdrawalImportTextService {
...
+      validationModel.sin = textValidation.textModel.sin;
+      validationModel.applicationNumber =
+        textValidation.textModel.applicationNumber;
+      validationModel.withdrawalDate = textValidation.textModel.withdrawalDate;
+      validationModel.applicationFound = false;
+      if (applicationData) {
+        validationModel.applicationFound = true;
+        validationModel.studentSINMatch =
+          textValidation.textModel.sin === applicationData.sin;
+        validationModel.hasCorrectInstitutionCode =
```
dheepak-aot, line 179: Are we suppose to do further validation, when originator is incorrect?

sh16011993: I just updated this part of the code.

dheepak-aot (approved):
Thanks for making the changes @sh16011993 
LGTM considering the need to wrap up the PR 👍

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.service.ts
```diff
@@ -115,6 +133,102 @@ export class ApplicationWithdrawalImportTextService {
...
+   * Get the application validation details for all the applications.
+   * @param applicationNumbers application numbers for which the application details need to be retrieved.
+   * @param institutionId institution id.
+   * @param originator institution that sent the file.
+   * @returns applications containing the required information.
+   */
+  private async getApplicationValidationData(
+    applicationNumbers: string[],
+    institutionId: number,
+    originator: string,
```
ann-aot, line 199: minor, maybe instead of `originator`, we can name it as `institutionCode`, same for line 145 in the service

andrewsignori-aot (approved):
Thanks for making the changes, adjusting the ticket, and making it happen. Looks good  👍

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.service.ts
```diff
@@ -115,6 +133,102 @@ export class ApplicationWithdrawalImportTextService {
...
+        applicationNumber: In(applicationNumbers),
+        location: {
+          institutionCode: originator,
+          institution: { id: institutionId },
+        },
+        studentScholasticStandings: {
+          changeType:
+            StudentScholasticStandingChangeType.StudentWithdrewFromProgram,
+        },
+      },
```
ann-aot, line 228: I am having a hard time understanding the logic,  this method is to retrieve the info of the application that are requested for scholastic standing right? then why this logic? so, the ideal scenario is the application won't have any previous scholastic standing, so with this `where` condition, those application info won't be retrieving right? Let me know if I am missing anything

sh16011993: Yes. Thanks @ann-aot . Corrected it.

sources/packages/backend/apps/api/src/services/application-bulk-withdrawal/application-bulk-withdrawal-import-text.service.ts
```diff
@@ -115,6 +133,102 @@ export class ApplicationWithdrawalImportTextService {
...
+      originator,
+    );
+    applicationValidationData.forEach((record) => {
+      applicationDataMap[record.applicationNumber] = {
+        applicationStatus: record.applicationStatus,
+        isArchived: record.isArchived,
+        sin: record.student.sinValidation.sin,
+        locationId: record.location.id,
+        locationCode: record.location.institutionCode,
+        hasPreviouslyBeenWithdrawn: !!record.studentScholasticStandings?.filter(
```
ann-aot, line 164: use can use `.some()` and it will directly return a boolean

andrepestana-aot (approved):
Looks good!

ann-aot (approved):
Nice effort @sh16011993 👍  thanks for doing the changes 👍

Issue #2383, opened 2023-10-05
Horizontal autoscaler target needs to be changed from Replication controller to deployment config


guru-aot:
Horizontal autoscaler target needs to be changed from Replication controller to deployment config

**Additional Context**
- Our environments have Horizontal Autoscalers defined and they are attached to the replication controllers currently.
- The problem here is, the replication controllers, die as soon as the pods are spin up. SO the Horizontal autoscalers are not able to scale up or live.
- So change them their scale target to deployment config.

**Acceptance Criteria**
- [x] Horizontal autoscaler target needs to be changed from Replication controller to deployment config.
- [x] Apply the Horizontal autoscaler for the pods, change the deploy.yaml and apply them during the deployments.
- [x] Add manual steps for release notes as need. 
- [x] Check the different between autoscaling/v1 and autoscaling/v2 and use the best of two.

guru-aot:
Difference between  autoscaling/v1 and autoscaling/v2

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelhiQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--63c3e16d08ebeec50ce2d5e8b8c839a6b41a6925/image.png)

https://docs.openshift.com/container-platform/3.11/dev_guide/pod_autoscaling.html

PR #2485, opened 2023-11-08
#2383 - Horizontal Autoscaler from replication controller to deployment config

guru-aot:
- [x] Horizontal autoscaler target needs to be changed from Replication controller to deployment config.
- [x] Apply the Horizontal autoscaler for the pods, change the deploy.yaml and apply them during the deployments.

Without HPA not properly assigned
image

With HPA properly assigned
image
image

Formio Load testing
WIth 2 pods it scaled up to 6 pods for 400 Vu
image
image
image

With 6 pods running the same sets of load for 400 Vu
image
image
image

With 6 pods running 300 Vu
image
image
image

With 6 pods running 200 Vu
image
image
image

With 6 pods running 100 Vu
image
image
image

With 2 pods running 100 Vu it started the 3rd pod
image
image
image


devops/openshift/api-deploy.yml
```diff
@@ -225,18 +225,30 @@ objects:
       to:
         kind: Service
         name: ${NAME}
-  - apiVersion: autoscaling/v1
+  - apiVersion: autoscaling/v2
```
dheepak-aot, line 228: 👍

devops/openshift/api-deploy.yml
```diff
@@ -225,18 +225,30 @@ objects:
...
       minReplicas:  "${{REPLICAS}}"
       maxReplicas: 10
-      targetCPUUtilizationPercentage: 80
+      metrics:
+          - type: Resource
+            resource:
+              name: cpu
+              target:
+                type: AverageValue
+                averageValue: '0.18'
```
dheepak-aot, line 245: Asking for my clarification, how did we arrive at this average value?

devops/openshift/api-deploy.yml
```diff
@@ -225,18 +225,30 @@ objects:
...
+          - type: Resource
+            resource:
+              name: memory
+              target:
+                type: AverageValue
+                averageValue: "230M"
```
dheepak-aot, line 251: Same here.

devops/openshift/forms-deploy.yml
```diff
@@ -170,7 +170,19 @@ objects:
...
+              name: cpu
+              target:
+                type: AverageValue
+                averageValue: '0.8'
+          - type: Resource
+            resource:
+              name: memory
+              target:
+                type: AverageValue
+                averageValue: "1Gi"
```
dheepak-aot, line 185: I see higher numbers for form.io server than API for autoscaling, is this with consideration that form.io server is going to have lighter load than API?

devops/openshift/api-deploy.yml
guru-aot: 90% of the CPU limit

devops/openshift/forms-deploy.yml
guru-aot: No, the averageValue is a percentage of the CPU and memory limits and here for formio, we have used the 50% as we used in the previous load testing iterations, this inturn helps for pods scaling without many requests failing.

devops/openshift/api-deploy.yml
guru-aot: 90% of the Memory limit

devops/openshift/forms-deploy.yml
dheepak-aot: Got it. Thanks for explaining. If 50% then cpu must be 0.75 ?

dheepak-aot (approved):
Great job 👍

devops/openshift/forms-deploy.yml
guru-aot: Yeah you are right, i did increased a bit and tested around so gave those value.

andrepestana-aot (approved):
Looks good! Good job!

devops/openshift/api-deploy.yml
andrewsignori-aot, line 1: Please ensure that this and all other documents are formatted. There are some minor differences. Please also ensure that VS Code is applying the format when saving 😉 
image

andrewsignori-aot (commented):
Great work. Just one really minor comment.
Let's talk about extending or having a second ticket to implement the `scaleDown` behavior if needed.
Besides that, once deployed I would like to see the API autoscaling load test 😉

ann-aot (approved):
👍 Thanks for the details

PR #2526, opened 2023-11-22
2283 - HTTP server for nest

andrepestana-aot:
## Context

The SIMS API needs to have security HTTP headers set for the endpoint responses to satisfy some security issues raised by the WAVA scans. Among the alternatives to achieve that, two of them can be highlighted as most common: 
  1. Configure and set headers through the API itself;
  2. Configure a reverse proxy to receive the client requests, redirect to the API and change the response headers to the client.

## Configure and set headers through the API itself

Helmet.js is the middleware module recommended by [Nest.js](https://docs.nestjs.com/security/helmet) and [Express.js](https://expressjs.com/en/advanced/best-practice-security.html#use-helmet). As its documentation mentions, we can achieve something similar [without Helmet.js](https://helmetjs.github.io/faq/you-might-not-need-helmet/). The middleware will set some headers by default and allow the developer to customize it. Besides, it will organize it in a programmatically [way](https://helmetjs.github.io/faq/conditional-options/). 

Other  middleware options:
- Lusca
- Security-Headers
- CSP-Header
- Express-Secure-Headers

It's also possible to set headers for the API responses without any additional module by adding a middleware function: 
```js
  app.use((_, res, next) => {
    res.setHeader("Cache-Control", "no-cache");
    next();
  });
```

## Configure a reverse proxy to receive the client requests, redirect to the API and change the response headers to the client.

In the existing architecture, the existing Nginx server used for the web module could also be configured as a reverse proxy for the API, adding headers to the API requests.

image

Sample Nginx configuration:
```
location /api/ {
    proxy_pass http://dev-api-sims:3000;
    proxy_set_header Host $host;
    proxy_set_header X-Content-Type-Options nosniff;
    ...
}
```

Although the Nginx configuration is not complex, having this redirection of requests to the Nginx server would imply in changing API OpenShift route to the Nginx Server, new log messages would be logged to Nginx regarding the API requests and a new point of failure would be added.

Also, a reverse proxy is usually used to help us solve issues like CORS configuration, Load Balancing, SSL/TLS Termination, Request Rate limiting, Scalability, among others issues. In our architecture hosted on OpenShift these issues are already or can be addressed either by configuring OpenShift features or by setting in the API through Express.js. For this reason this option was not chosen to be implemented now.

## Implementation

In this PR it was added Helmet.js npm package and addressed some of the issues presented by the Wava scan for the institution module. Some headers were added without using Helmet because the Helmet module itself doesn't support them. There's an additional module called [nocache](https://github.com/helmetjs/nocache) that would be supposed to deal with "Cache-Control" header but is not worth it to add a new npm package just to add a [small middleware function](https://github.com/helmetjs/nocache/blob/main/index.js).

### New Headers
image

## References

- https://docs.nestjs.com/security/helmet
- https://expressjs.com/en/advanced/best-practice-security.html#use-helmet
- https://helmetjs.github.io/
- https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/
- https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/How-to-setup-Nginx-reverse-proxy-servers-by-example

### DIFF
--- a/sources/packages/backend/apps/api/src/main.ts
+++ b/sources/packages/backend/apps/api/src/main.ts
@@ -25,18 +25,10 @@ async function bootstrap() {
   // Setting global prefix
   app.setGlobalPrefix("api");
 
-  // Using helmet
-  app.use(
-    helmet({
-      strictTransportSecurity: {
-        maxAge: 31536000,
-        includeSubDomains: true,
-        preload: true,
-      },
-    }),
-  );
+  // Using helmet.
+  app.use(helmet());
 
-  // Adding headers not covered by helmet
+  // Adding headers not covered by helmet.
   app.use((_, res, next) => {
     res.setHeader("Cache-Control", "no-store");
     res.setHeader("Pragma", "no-cache");
--- a/sources/packages/backend/apps/api/src/main.ts
+++ b/sources/packages/backend/apps/api/src/main.ts
@@ -30,8 +30,7 @@ async function bootstrap() {
 
   // Adding headers not covered by helmet.
   app.use((_, res, next) => {
-    res.setHeader("Cache-Control", "no-store");
-    res.setHeader("Pragma", "no-cache");
+    res.setHeader("Cache-Control", "no-cache");
     next();
   });
 
--- a/sources/packages/backend/apps/api/src/main.ts
+++ b/sources/packages/backend/apps/api/src/main.ts
@@ -10,6 +10,7 @@ import { setGlobalPipes } from "./utilities/auth-utils";
 import { DocumentBuilder, SwaggerModule } from "@nestjs/swagger";
 import { Request, Response } from "express";
 import { KeycloakConfig } from "@sims/auth/config";
+import helmet from "helmet";
 
 async function bootstrap() {
   await KeycloakConfig.load();
@@ -23,6 +24,22 @@ async function bootstrap() {
 
   // Setting global prefix
   app.setGlobalPrefix("api");
+
+  app.use(
+    helmet({
+      noSniff: true, // X-Content-Type-Options
+      contentSecurityPolicy: {
+        directives: {
+          defaultSrc: ["'self'"],
+        },
+      },
+      strictTransportSecurity: {
+        maxAge: 31536000,
+        includeSubDomains: true,
+        preload: true,
+      },
+    }),
+  );
 
   // Exception filter
   const { httpAdapter } = app.get(HttpAdapterHost);
--- a/sources/packages/backend/apps/api/src/main.ts
+++ b/sources/packages/backend/apps/api/src/main.ts
@@ -25,14 +25,9 @@ async function bootstrap() {
   // Setting global prefix
   app.setGlobalPrefix("api");
 
+  // Using helmet
   app.use(
     helmet({
-      noSniff: true, // X-Content-Type-Options
-      contentSecurityPolicy: {
-        directives: {
-          defaultSrc: ["'self'"],
-        },
-      },
       strictTransportSecurity: {
         maxAge: 31536000,
         includeSubDomains: true,
@@ -40,6 +35,13 @@ async function bootstrap() {
       },
     }),
   );
+
+  // Adding headers not covered by helmet
+  app.use((_, res, next) => {
+    res.setHeader("Cache-Control", "no-store");
+    res.setHeader("Pragma", "no-cache");
+    next();
+  });
 
   // Exception filter
   const { httpAdapter } = app.get(HttpAdapterHost);
--- a/sources/packages/backend/package-lock.json
+++ b/sources/packages/backend/package-lock.json
@@ -32,6 +32,7 @@
         "dayjs": "^1.10.4",
         "dotenv": "^8.2.0",
         "express-basic-auth": "^1.2.1",
+        "helmet": "^7.1.0",
         "ioredis": "^5.2.4",
         "jsonpath": "^1.1.1",
         "jwt-decode": "^3.1.2",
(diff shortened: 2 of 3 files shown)


sources/packages/backend/apps/api/src/main.ts
```diff
@@ -24,6 +25,24 @@ async function bootstrap() {
   // Setting global prefix
   app.setGlobalPrefix("api");
 
+  // Using helmet
+  app.use(
+    helmet({
+      strictTransportSecurity: {
```
andrewsignori-aot, line 31: Is the default configuration not enough in this case for `Strict-Transport-Security`.
If this is following the Wava recommendation I would say that either one would work and Wava was probably complaining about the missing configuration.

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -24,6 +25,24 @@ async function bootstrap() {
...
+        includeSubDomains: true,
+        preload: true,
+      },
+    }),
+  );
+
+  // Adding headers not covered by helmet
+  app.use((_, res, next) => {
+    res.setHeader("Cache-Control", "no-store");
+    res.setHeader("Pragma", "no-cache");
```
andrewsignori-aot, line 42: The`Pragma` seems deprecated: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma
Besides that, it seems that it can be replaced by `Cache-Control: no-cache`.
Are these headers added based on some best practice or was it based on some Wava recommendation?

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -24,6 +25,24 @@ async function bootstrap() {
...
+        maxAge: 31536000,
+        includeSubDomains: true,
+        preload: true,
+      },
+    }),
+  );
+
+  // Adding headers not covered by helmet
+  app.use((_, res, next) => {
+    res.setHeader("Cache-Control", "no-store");
```
andrewsignori-aot, line 41: If we would like to enable cache for a particular API endpoint, `config` for instance, with the configuration at this level would it still be possible?

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -24,6 +25,24 @@ async function bootstrap() {
...
+    helmet({
+      strictTransportSecurity: {
+        maxAge: 31536000,
+        includeSubDomains: true,
+        preload: true,
+      },
+    }),
+  );
+
+  // Adding headers not covered by helmet
```
andrewsignori-aot, line 39: Missing period in the comment.

andrewsignori-aot (commented):
Great research and thanks for the great PR description documentation. Please take a look at the comments.

sources/packages/backend/apps/api/src/main.ts
andrepestana-aot: Yes, that would be something like this:
```js
  app.use((req, res, next) => {
    if (req.url.contains("/api/config")) {
      res.setHeader("Cache-Control", "public, max-age=3600");
      res.setHeader("Pragma", "public");
    } else {
      res.setHeader("Cache-Control", "no-store");
      res.setHeader("Pragma", "no-cache");
    }
    next();
  });
```

sources/packages/backend/apps/api/src/main.ts
andrepestana-aot: Yes, "Pragma" header was used for HTTP/1.0 and is deprecated and shouldn't there be browsers old enough to use it. Wava scan still recommends it for compatibility:
image
Mozilla [mentions ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma)that it is ok to be used for compatibility if we have "Cache-Control" header.
image

sources/packages/backend/apps/api/src/main.ts
andrepestana-aot: yes, I changed the configuration and let the helmet default configuration as it should be enough.

sources/packages/backend/apps/api/src/main.ts
andrepestana-aot: Wava scan recommends to prevent caching for the `config` endpoint though.
image
At least it was not clear if just having any values for "Cache-Control" and "Pragma" headers would satisfy that issue.

ann-aot (approved):
Nice work and great research @andrepestana-aot 👍 and thanks for the detailed description and references

guru-aot (approved):
Extensive research @andrepestana-aot LGTM.

sources/packages/backend/apps/api/src/main.ts
andrewsignori-aot: For Wava, we do not need to follow the recommendation if we have a justification.
The recommendation from Mozilla is that we do not need to use it if we use `Cache-control: no-cache`, right?
I would recommend just using the `Cache-control: no-cache` which apparently would supersede the `no-store`. Please let me know we I am missing something.

sources/packages/backend/apps/api/src/main.ts
andrewsignori-aot: Maybe I was not clear when talking about adding a cache to an endpoint I was referring to the below.
https://docs.nestjs.com/controllers#headers
```ts
@Get()
@Header('Cache-Control', 'none')
getConfig() {
  return { ourConfig };
}
```

andrepestana-aot: Yes, that can be achieved that way too if needed and I did a test here to confirm that the `@Header()` decorator will be applied even if we have a middleware function setting the same header globally.

andrewsignori-aot (approved):
Thanks for doing the changes, looks great 👍

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -24,6 +25,15 @@ async function bootstrap() {
   // Setting global prefix
   app.setGlobalPrefix("api");
 
+  // Using helmet.
+  app.use(helmet());
+
+  // Adding headers not covered by helmet.
+  app.use((_, res, next) => {
+    res.setHeader("Cache-Control", "no-cache");
```
dheepak-aot, line 33: 👍

dheepak-aot (approved):
Nice work @andrepestana-aot 👍

Great to see the depth of analysis on alternate options.

Issue #2489, opened 2023-11-08
Camunda load testing - Part 2


dheepak-aot:
**Describe the task**
Implement/do analysis on remaining parts of camunda load testing from #2288 

**Acceptance Criteria**
- [x] Observe the camunda optimize in dev to identify performance benchmarks
- [x] Run multiple test to check the performance (for workers) by increasing the number of PODs and modifying CPU/memory per POD
- [x] Check if there is the best way to configure workers maxJobsToActivate.
- [x] Produce reporting of output vs pod#s/CPU/memory configuration.
- [x] Identify the best configuration and modify the YAML file to standardize future deployment.

dheepak-aot:
500 requests 15 VU unmodified worker configuration

![image](https://user-images.githubusercontent.com/54600590/283540912-1cf958ac-6165-495f-93b0-c7cc9568b525.png)

500 requests 15 VU - worker `associate-workflow-instance` updated with `maxJobsToActivate` as `MaxJobsToActivate.High`

![image](https://user-images.githubusercontent.com/54600590/283544349-2a60ff2e-1879-49fc-9c77-c26d9b90cab3.png)

dheepak-aot:
So far with the load testing executed in local env, 2 issues are observed in the workers node application

1 - At the worker `associate-workflow-instance `  `getSystemUser` is called inside transaction and the execution of getSystemUser does not use transaction entity manager. 
It has been observed that with considerable concurrent load to the worker, typeorm is stuck at this point where findOne is executed inside `getSystemUser`. And it seems to be stuck there forever.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVA5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--739afbc1819b0898c034f4ef30c6abaff89f356f/image.png)

it does not produce any typeorm query logs and it seems not reach the database.
Solutions which worked: If we move the `getSystemUser` outside the transaction, this issue does not occur. Or if we use the transactional entity manager inside `getSystemUser` it solves the issue as well.

dheepak-aot:
Optimize heat map on average duration after load test with 5000 iterations and 20 VUs

[...]

PR #2528, opened 2023-11-23
#2489 - ORM fixes to workers based on load test results

dheepak-aot:
### DB opertions performed inside trancation being stuck in deadlock situation

_Description of the issue_: 
It has been observed during the load testing of workers that, when a code which has db operations happening in a transaction is executed in parallel AND given that the transaction involves one or more operation using global entity manager(not the transaction entity manager) AND the number of parallel executions are sufficient enough to get the datasource connection pool at or beyond it's capacity, then we encounter a dead lock situation and all the requests made to db are waiting forever.

Identified root cause:
With continuous debugging by  @andrewsignori-aot and myself, we found the root cause of this issue as described.

Let us assume the connection pool is at default size 10. And we execute a code with transaction(Transaction connection) and one or more operation inside the transaction using global entity manager(Global connection) and number of parallel executions are 10 to fill up the connection pool.

In such circumstance, here is how the deadlock happen. 

image

Fixes implemented towards the issue and also on the course of fixing the issue
 - [x] All the workers have been investigated and code involving database transaction in any of the worker implementation are refactored to use the `transactionEntityManager` to perform all the db operations irrespective of if the operation takes part in transaction or not.
 - [x] Updated the `ormConfig` with connection pool options by increasing the size of the connection pool and providing timeout for waiting connection requests when the pool is at capacity. Added `@types/pg` to have the options typed.

image

 - [x] Typeorm upgrade (Typeorm upgrade was the first step taken towards the errors happening in workers at db/orm level)  

### System User Service refactor
 WHY - In the system user service, to obtain a system user could be a db operation if the variable cache is not loaded. As obtaining a system user could be needed anywhere including inside a transaction, instead of passing the entity manager downstream it is identified as better approach to load the system user on start up, so that it can be assessed anywhere in the code without being a db operation.

While doing this update, there is minor concern on the concurrency that when there is no system user existing in db and multiple applications start in parallel, one or more applications with try to insert system user at same time and can have a error at insert due unique username constraint. This will result in one or more applications not setting the systemUser variable. 
To avoid and adapt to this situation, the insert code has been set to do nothing on unique constraint error assuming that the system user is already created.

image

Finally on this part, as we are loading the variable of a shared instance of `SystemUserService` and this is available as a provider from global module. Hence we need to import `SystemUserModule` to any module requiring the provider and not the provider alone which could create multiple shared instances of  `SystemUserService`. 

Found this was happening in Queue Consumers Module from E2E tests as they were failing because the instance of `SystemUserService` loaded on startup and the instance consumed by the E2E test implementation were different.

So made a refactor to use the SystemUserService always imported as Module instead of being added as provider.

image

image

 - [x] Refactored `SystemUserService` to have a load method which is meant to be loaded on start up of applications(API/Workers/Queue-consumer/E2E Tests)
 - [x] Added insert on onflict do nothing to adapt for concurrency 
 - [x]  Refactor to use the SystemUserService always imported as Module instead of being added as provider.


sources/packages/backend/libs/integrations/src/esdc-integration/e-cert-integration/e-cert-integration.module.ts
```diff
@@ -22,11 +22,11 @@ import {
   ECertGenerationService,
   SshService,
 } from "../../services";
-import { SystemUsersService } from "@sims/services/system-users";
+import { SystemUserModule } from "@sims/services/system-users";
 import { SFASApplicationService } from "@sims/services/sfas";
 
 @Module({
-  imports: [ConfigModule],
+  imports: [ConfigModule, SystemUserModule],
```
ann-aot, line 29: 👍

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
```diff
@@ -41,21 +35,32 @@ export class SystemUsersService {
...
-    if (this.systemUserDetails) {
-      return this.systemUserDetails;
-    }
+  async loadSystemUser(): Promise<void> {
+    this.systemUser = await this.getOrCreateSystemUser();
+  }
 
-    this.systemUserDetails = await this.getSystemUser();
-    return this.systemUserDetails;
+  private async getSystemUser(): Promise<User> {
```
ann-aot, line 56: pls add a breif comments to the method

ann-aot (approved):
Nice work and appreciate extensive research/work to hunt down the issue and the reason, thanks for the nice description. Just a minor comment added

sources/packages/backend/apps/workers/src/controllers/assessment/_tests_/e2e/assessment.controller-updateNOAStatus.e2e-spec.ts
```diff
@@ -54,7 +54,7 @@ describe("AssessmentController(e2e)-updateNOAStatus", () => {
     expect(expectedAssessment.noaApprovalStatus).toBe(
       AssessmentStatus.completed,
     );
-    const auditUser = await systemUsersService.systemUser();
+    const auditUser = systemUsersService.systemUser;
```
andrewsignori-aot, line 57: I understand the "renaming" here but I would not enforce that every time for upcoming PRs.
IMO using it in the below way would be also acceptable unless the code would be less readable.
For this PR I do belive that it is all looking good 😉 
```ts
expect(expectedAssessment.modifier).toEqual(systemUsersService.systemUser);
```

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
```diff
@@ -10,7 +10,8 @@ import {
 
 @Injectable()
 export class SystemUsersService {
-  private systemUserDetails: User = undefined;
+  /** System user */
```
andrewsignori-aot, line 13: Can you please review this comment?

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
```diff
@@ -41,21 +35,32 @@ export class SystemUsersService {
...
-   * and return the user details.
-   * @return system user details.
+   * Load system user.
    */
-  async systemUser(): Promise<User> {
-    if (this.systemUserDetails) {
-      return this.systemUserDetails;
-    }
+  async loadSystemUser(): Promise<void> {
+    this.systemUser = await this.getOrCreateSystemUser();
```
andrewsignori-aot, line 53: If this is the only place calling `getOrCreateSystemUser` do we need two methods?
Can `loadSystemUser` do exactly what `getOrCreateSystemUser` is doing?

sources/packages/backend/libs/services/src/system-users/system-users.service.ts
```diff
@@ -10,7 +10,8 @@ import {
 
 @Injectable()
 export class SystemUsersService {
-  private systemUserDetails: User = undefined;
+  /** System user */
+  systemUser: User = undefined;
```
andrewsignori-aot, line 14: I believe that it should be converted to a read-only property, right?

sources/packages/backend/libs/utilities/src/system-configurations-constants.ts
```diff
@@ -30,6 +30,17 @@ export const ORM_CACHE_LIFETIME = 10 * 60 * 1000;
  */
 export const ORM_CACHE_REDIS_COMMAND_TIMEOUT = 5 * 1000;
 
+/**
+ * Maximum number of open connections that
+ * are allowed for connection pool.
+ */
+export const CONNECTION_POOL_MAX_CONNECTIONS = 20;
```
andrewsignori-aot, line 37: Should we match this with the number used by the `processInParallel` method?
Wonder that by default the `processInParallel` will use 32 as a baseline depending on the POD cores which means that 32 parallel operations will be started and there will be 20 connections available meaning that 12 parallel operations will be waiting for a connection to be acquired from the pool (that now has 2min timeout).
It does not seem a big deal under regular operation but it can be pretty impactful if the DB is under some pressure or taking too long to respond.

sources/packages/backend/package.json
```diff
@@ -80,7 +80,7 @@
     "soap": "^1.0.0",
     "ssh2-sftp-client": "^9.1.0",
     "swagger-ui-express": "^4.3.0",
-    "typeorm": "^0.3.14",
+    "typeorm": "^0.3.17",
```
andrewsignori-aot, line 83: Thanks for doing the upgrade ❤️

andrewsignori-aot (commented):
Great work, great research. Only minor comments, please take a look.

andrepestana-aot (approved):
Great work and extensive search for the root causes and fixes! Just look for the other devs comments.

sources/packages/backend/apps/queue-consumers/test/helpers/testing-modules/testing-modules-helper.ts
```diff
@@ -52,8 +52,14 @@ export async function createTestingAppModule(): Promise<CreateTestingModuleResul
 
   const nestApplication = module.createNestApplication();
   await nestApplication.init();
+
```
guru-aot, line 55: blank line, please remove

sources/packages/backend/libs/services/src/notifications/notification-message/notification-message.service.ts
```diff
@@ -15,12 +15,17 @@ export class NotificationMessageService extends RecordDataModelService<Notificat
   /**
    * Retrieves the template id of a notification message.
    * @param notificationMessageTypeId id of the user who will receive the message.
+   * @param options options.
+   * - `entityManager` external entity manager to run in a transaction.
    * @returns template Id of the notification message.
    */
   async getTemplateId(
     notificationMessageTypeId: NotificationMessageType,
+    options?: { entityManager?: EntityManager },
```
guru-aot, line 24: Nice one

sources/packages/backend/libs/sims-db/src/data-source.ts
```diff
@@ -55,12 +55,18 @@ import {
...
+  CONNECTION_POOL_MAX_CONNECTIONS,
+  CONNECTION_REQUEST_TIMEOUT,
   ORM_CACHE_LIFETIME,
   ORM_CACHE_REDIS_COMMAND_TIMEOUT,
   ORM_CACHE_REDIS_RETRY_INTERVAL,
 } from "@sims/utilities";
 import { ConfigService } from "@sims/utilities/config";
+import { PoolConfig } from "pg";
 
+interface ConnectionOptions extends PostgresConnectionOptions {
```
guru-aot, line 67: 👍

guru-aot (approved):
LGTM, nice exhaustive analysis @dheepak-aot

guru-aot (approved):
LGTM, nice exhaustive analysis @dheepak-aot

sources/packages/backend/libs/utilities/src/system-configurations-constants.ts
dheepak-aot: Thanks for bringing this point. As discussed we can take this up on upcoming dev sync up.

andrewsignori-aot (approved):

PR #2543, opened 2023-11-29
#2489 - Worker and ORM changes based on load test results

dheepak-aot:
## Worker and ORM Changes based on load test results

**Background:** 
While load testing the workers with a chunk of 5000 assessments with different worker configuration scenarios, It has been observed that, when the pods are auto-scaling and going at and beyond 6 pods for workers, an immense load has been put on the patroni db and the patroni db was going to recovery state for few seconds and then became functional without any issues. 

As a side effect of patroni db going to recovery state, the inference proves that  the worker `associate-workflow-instance` was executed twice for the same assessment, during such an occurrence it is expected that second workflow instance must be cancelled and it must not raise an incident. The workflow instance was getting cancelled but the code to cancel workflow was not terminating/conveying the terminal state of the job and the job keeps waiting forever. Due to this the worker `associate-workflow-instance` does not process any jobs which keeps all the assessment gateway instances waiting forever once after this error occurred. 

image
 
**_Solution for worker issue:_** Updated the worker code which cancel the workflow to await for cancellation and send complete status.

**_Solution for DB going to recovery state:_** Updated the connection pool property and worker configuration to control the load on database and the configuration provided in this PR has been tested to make sure it did not make the database enter recovery mode
**Note: This is not a final solution for the database. There will be an upcoming ticket for the analysis and stabilization of DB.

- [x] Updated the worker code which cancel the workflow to await for cancellation and send complete status.
- [x] As we are not raising incident in the event of `INVALID_OPERATION_IN_THE_CURRENT_STATUS` or `ASSESSMENT_ALREADY_ASSOCIATED_WITH_DIFFERENT_WORKFLOW` in the worker `associate-workflow-instance` changed the log type to warn from error
- [x] Reduced the quota of `maxJobsToActivate` for `associate-workflow-instance` to medium to control the load applied to patroni db
- [x] Reduced the  `maxJobsToActivate` to High for all workers which had Maximumto control the load applied to patroni db
- [x] Set the connection pool to default value 10. Still have the property explicitly defined for better understanding from code.

### DIFF
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -88,6 +88,14 @@ export class AssessmentController {
           case INVALID_OPERATION_IN_THE_CURRENT_STATUS:
           case ASSESSMENT_ALREADY_ASSOCIATED_WITH_DIFFERENT_WORKFLOW:
             logger.warn(error.getSummaryMessage());
+            // The below method also returns JOB_ACTION_ACKNOWLEDGEMENT ideally
+            // considered sufficient to acknowledge that the job is completed.
+            // But during the load test it has been identified that calling `job.cancelWorkflow()`
+            // was not acknowledging that job is complete after the instance was cancelled.
+            // At this moment, it proves to be bug in the framework and hence the approach of calling `job.complete()`
+            // was followed.
+            // TODO: Raise the issue regarding the bug to zeebe-node community and modify
+            // the code if required based on their response or solution.
             await job.cancelWorkflow();
             return job.complete();
         }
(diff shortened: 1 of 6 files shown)


sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -87,8 +87,9 @@ export class AssessmentController {
             return job.error(error.name, error.message);
           case INVALID_OPERATION_IN_THE_CURRENT_STATUS:
           case ASSESSMENT_ALREADY_ASSOCIATED_WITH_DIFFERENT_WORKFLOW:
-            logger.error(error.getSummaryMessage());
-            return job.cancelWorkflow();
+            logger.warn(error.getSummaryMessage());
+            await job.cancelWorkflow();
+            return job.complete();
```
andrepestana-aot, line 92: 👍

andrepestana-aot (approved):
LGTM!

guru-aot (approved):
LGTM, nice find

andrewsignori-aot (approved):
Great research and geat findings, looks good 👍

ann-aot (commented):
Thanks for the walk through. 👍

ann-aot (approved):

Issue #2546, opened 2023-11-30
Create readiness probe and liveness probe for Queue consumers and Workers - Part 2


cditcher:
Continuation of ticket #1580 

Acceptance Criteria

- [x] Have the implemented health check created from #1580 and create readiness and liveliness probe for queue consumers
- [x] Have the implemented health check created from #1580 and create readiness and liveliness probe for workers

Issue #1580, opened 2022-11-29
Create readiness and liveness probe for Queue consumers and Workers


dheepak-aot:
**Describe the task**
Queue consumers and Workers currently do not have a  readiness and liveness probe as it does not have a non-authenticated public endpoint.

Once this is set podDisruptionBudget set as part of #1875 will work automatically.

Node Js documentation on setting health check reference : https://docs.nestjs.com/recipes/terminus

**Acceptance Criteria**
- [x] Create a health check/heartbeat API and create readiness and liveness probe for queue consumers and workers.
- [x] Research to find out-of-box approach to do the health check from the pods.
- [x] If time allows, have it implemented, otherwise, create a new ticket for implementation. For instance: https://docs.nestjs.com/recipes/terminus

PR #2550, opened 2023-12-01
#2546 - Readiness and Liveliness Probe for workers and Queue consumers

guru-aot:
- [x] Have the implemented health check created from #1580 and create readiness and liveliness probe for queue consumers
- [x] Have the implemented health check created from #1580 and create readiness and liveliness probe for workers

Tested the scenarios with the setting mentioned in the PR for Workers and queue consumers in dev by manually setting the liveliness and readiness probe. Tested by making redis and/or postgres down, both the pods restarted after the recovery successfully.

### DIFF
--- a/devops/openshift/queue-consumers-deploy.yml
+++ b/devops/openshift/queue-consumers-deploy.yml
@@ -213,6 +213,7 @@ objects:
                   memory: ${MEMORY_REQUEST}
               readinessProbe:
                 failureThreshold: 3
+                successThreshold: 1
                 httpGet:
                   path: /health/
                   port: "${{PORT}}"
@@ -221,6 +222,7 @@ objects:
                 timeoutSeconds: 1
               livenessProbe:
                 failureThreshold: 3
+                successThreshold: 1
                 httpGet:
                   path: /health/
                   port: "${{PORT}}"
--- a/devops/openshift/queue-consumers-deploy.yml
+++ b/devops/openshift/queue-consumers-deploy.yml
@@ -211,6 +211,23 @@ objects:
                 requests:
                   cpu: ${CPU_REQUEST}
                   memory: ${MEMORY_REQUEST}
+              readinessProbe:
+                failureThreshold: 3
+                httpGet:
+                  path: /health/
+                  port: "${{PORT}}"
+                initialDelaySeconds: 10
+                periodSeconds: 10
+                timeoutSeconds: 1
+              livenessProbe:
+                failureThreshold: 3
+                httpGet:
+                  path: /health/
+                  port: "${{PORT}}"
+                  scheme: HTTP
+                initialDelaySeconds: 30
+                periodSeconds: 30
+                timeoutSeconds: 1
   - apiVersion: v1
     kind: Service
     metadata:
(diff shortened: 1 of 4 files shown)


devops/openshift/queue-consumers-deploy.yml
```diff
@@ -211,6 +211,23 @@ objects:
                 requests:
                   cpu: ${CPU_REQUEST}
                   memory: ${MEMORY_REQUEST}
+              readinessProbe:
```
andrewsignori-aot, line 214: some items like `periodSeconds: 10`, `timeoutSeconds: 1`, and `failureThreshold: 3`  are set explicitly with the default values. Was it to make it clear somehow? If that is true, should we include others like `successThreshold`?

devops/openshift/queue-consumers-deploy.yml
```diff
@@ -211,6 +211,23 @@ objects:
...
+                failureThreshold: 3
+                httpGet:
+                  path: /health/
+                  port: "${{PORT}}"
+                initialDelaySeconds: 10
```
dheepak-aot, line 219: Same for inital delay seconds.

guru-aot: initial delay for applications are determined, to how long to wait before the container health checks needs to be done. For workers and queue consumers, i determined this value by doing tests from 1, 5, 10, 20 and 30. If we have more initial delay that will cause time delay to get the application to started state, if we have less initial delay the container will restart again and again, thinking the health check is failing. So we have to find the median.

devops/openshift/queue-consumers-deploy.yml
```diff
@@ -211,6 +211,23 @@ objects:
                 requests:
                   cpu: ${CPU_REQUEST}
                   memory: ${MEMORY_REQUEST}
+              readinessProbe:
+                failureThreshold: 3
```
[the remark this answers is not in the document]
andrewsignori-aot: IMO option we do not need to change Patroni, lest keep using what was provided.
For API/Web we can change now to be 3 also. It is out of the scope of the ticket so it is up to you @guru-aot.

andrewsignori-aot (approved):
Thanks for making the changes for Workers and Queue-Consumers and for doing the research needed. Looks good 👍 
Let's see how both applications behave and if there will be no unnecessary restarts in the upcoming days.

devops/openshift/queue-consumers-deploy.yml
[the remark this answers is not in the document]
guru-aot: Done the changes Team

devops/openshift/queue-consumers-deploy.yml
dheepak-aot: I got the initial delay for this file. But api-deploy.yml was having 40. I suspect we might have considered the migration run time initially when db migrations were running during deployment. 

It is outside the PR so we can consider that outside the ticket.

dheepak-aot (approved):
Nice work @guru-aot. Thanks for extending the work to api-deploy config.

I just have point on api readiness and liveliness probe for api, where it does not use terminus and other properties in api-deploy for readiness and liveliness probe needs to be revisited. We can create a ticket brining this to the attention of the team and work on it later.

ann-aot (approved):
nice work @guru-aot 👍

andrepestana-aot (commented):
Looks good!

Issue #2401, opened 2023-10-11
(Discussion+Estimate) Revise Calculations Dictating # of Disbursements


HRAGANBC:
**Description of Task:**
Conduct sufficient grooming of below acceptance criteria to provide workload estimate to support SABC decision-making. If approved, ACs will be created into a ticket to implement prior to MVP release. 

**Context:**
- Many Study Period Offerings configured by the school will be calculated as 18 weeks in the SIMS system, and therefore be spread across two disbursements.  

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDdiQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e6cc12ff81bcc1dc5a3d6863a0308f22c18de6c2/DisbursementSchedule.JPG)

**Acceptance Criteria**
Revise FT and PT disbursement scheduling calculations such that:
- [x] If offering length is **19 weeks or greater**,  preserve status quo in dividing most funding types across both disbursements (w/ exception of CSL and BCSL which are all-or-nothing) 

**19 weeks or greater disbursement Example Breakdown**
Disbursement 1 
Canada loan 100%
CSG 50%
BC Loan 0%
BC Grant 50%
 
Disbursement 2 
Canada loan 0%
CSG 50%
BC Loan 100%
BC Grant 50%

- [x] If study weeks are **=18 weeks** , change disbursement configuration such that all award types are consolidated onto the first disbursement, EXCEPT Canada Student Grant awards which are distributed equally across two disbursements.
**= to 18 weeks offering disbursement Example break down**
Disbursement 1
CSL 100%
CSG 50%
BC Loan 100%
BC Grant 100%

Disbursement 2
CSL 0%
CSG 50%
BC Loan 0%
BC Grant 0%

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFVWQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3713bb465b6ae0919f7dd67686242fc37ec6f94d/image.png)

Additional Context 
Disbursement breakdown for 17 weeks or less
remains status quo of 
1 Disbursement
100% of all funds

ann-aot:
Screenshots of the demo,
[...]

PR #2569, opened 2023-12-06
#2401 - (Discussion+Estimate) Revise Calculations Dictating # of Disbursements

ann-aot:
- Adjusted the PT and FT disbursement logic
- Current logic, where disbursements are splitting into 2, it was checking `>=17`. I changed it to `>17` and for one disbursement `< 17` is updated to `<= 17`, in both FT and PT.
- PT
image
image
- FT
image
image

- **Screenshots**
- PT
image
image
image
image
image
image
image
- FT
image
image
image
image
image
image
image
image
- If any disbursement does not have any awards or grants, then it won't be saved in `disbursement_schedules`. some disbursement has certain loan which has `awardEligibility: true`, the logic is not added for those kind of disbursements

### DIFF
--- a/sources/packages/backend/workflow/BPMN/camunda-8/fulltime-configure-disbursement.bpmn
+++ b/sources/packages/backend/workflow/BPMN/camunda-8/fulltime-configure-disbursement.bpmn
@@ -64,7 +64,7 @@
     <bpmn:intermediateThrowEvent id="Event_0aunmqb" name="Configure Payload for 18 weeks">
       <bpmn:extensionElements>
         <zeebe:ioMapping>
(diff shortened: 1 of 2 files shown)


sources/packages/backend/workflow/BPMN/camunda-8/fulltime-configure-disbursement.bpmn
```diff
@@ -61,14 +61,14 @@
       <bpmn:incoming>Flow_1bx16ip</bpmn:incoming>
       <bpmn:outgoing>Flow_1kesibu</bpmn:outgoing>
     </bpmn:intermediateThrowEvent>
-    <bpmn:intermediateThrowEvent id="Event_0aunmqb" name="Configure Payload">
+    <bpmn:intermediateThrowEvent id="Event_0aunmqb" name="Configure Payload for 18 weeks">
```
andrewsignori-aot, line 64: For the 18 weeks, the second disbursement will contain only `Canada Grants`. Is it possible that none of those grants would be eligible making the second disbursement in the payload as an empty array?
Does the worker's code need any adjustment to deal with it?

andrewsignori-aot (commented):
Nice work, please take a look at the comment about a possible second empty disbursement.

sources/packages/backend/workflow/BPMN/camunda-8/fulltime-configure-disbursement.bpmn
ann-aot: @andrewsignori-aot I tested the above, I believe we are already taking care of it, I created and application of 18 weeks which is not eligible for any grants, below are my findings,
assessment flow,
image
`disbursement_schedules` will have 2 records, 
image
but `disbursement_values` will have only eligible disbursement award i.e. disbursement-1 award,
image
And NOA will have only 1 disbursement ,
image

so should we still need to restrict it in camunda level?

ann-aot: updated as suggested

andrewsignori-aot (approved):
Thanks for the analysis and for making the changes, looks good 👍

andrepestana-aot (approved):
Looks good!

dheepak-aot (approved):
Great Job @ann-aot . Looks Good 👍

PR #2577, opened 2023-12-08
2535 - Crash Loop Back offs Error

guru-aot:


devops/openshift/docker-build.yml
```diff
@@ -5,28 +5,6 @@ kind: Template
 metadata:
   name: ${NAME}
 objects:
-  - apiVersion: image.openshift.io/v1
```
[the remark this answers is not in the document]
guru-aot: The imagestream is usually used to mention or represent what is the container's base image and the imagestreamtag is used to associate a specific tag version of the image.
These configurations inturn helped us to override the base image defined in the Dockerfile which was defined at the top of it "FROM". 
In our case, for our deployments we used the same base image for (Api, web, workers, queue-consumers, load-test-gateway & db-migration), so there is no need to override the once in the dockerfile, as we run all our builds using Github action. Even in the future if we wanted to specifically change the base image for a particular container, we can change the dockerfile for the particular base image and run the Github action, it should eventually work.

sources/packages/backend/apps/api/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Base Image
-FROM artifacts.developer.gov.bc.ca/redhat-access-docker-remote/ubi8/nodejs-18:1-71.1697652955
+FROM artifacts.developer.gov.bc.ca/redhat-access-docker-remote/ubi8/nodejs-18:1-81
```
andrewsignori-aot, line 2: While I support the node upgrade either way and the research looks great, we are still missing the root cause.
Can you please evaluate if the below assumptions make sense?

1 - Even the Openshift docker image defining a non-root user 1001, the container will be executed with a random user, as the error in the ticket also points out.
image
The same is also supported by the below documentation.
https://docs.openshift.com/container-platform/4.11/openshift_images/create-images.html#use-uid_create-images

2 - Checking the BC git I found at least one entry applying the recommended solution due to a npm 9 issue.
image

The above would also explain why the error started when we moved from node 16 to 18 (npm 8 to 10).
Should we consider also applying the same?

dheepak-aot: I see and I agree what @andrewsignori-aot about the OpenShift container running with  arbitrarily assigned user.

I am 100% on same page that we need to run a fix permission command to have the highest level of certainty that the issue is taken care at it's root cause.

When I look at the openshift, see that write permissions are not present outside the owner for the directory. 

image

 
There is one more thing which I want to share here. It may also be a passible solution.

Please go through this thread, https://github.com/sclorg/s2i-nodejs-container/issues/396

There is a mention about same error (we are using npm ci)

Following screenshots are the highlights linking to our issue

image

image



image

devops/Makefile
```diff
@@ -242,47 +242,47 @@ build-db-migrations:
 	test -n "$(BUILD_REF)"
 	test -n "$(DB_MIGRATIONS_BUILD_REF)"
 	@echo "+\n++ BUILDING DB migrations with tag: $(BUILD_REF)\n+"
-	@oc -n $(BUILD_NAMESPACE) process -f $(BUILD_TEMPLATE_PATH) -p TAG=$(BUILD_REF) -p SOURCE_REPOSITORY_REF=$(BUILD_REF) -p BASE_IMAGE_NAME="nodejs-18" -p BASE_IMAGE_TAG="1-71.1697652955" -p BASE_IMAGE_REPO="artifacts.developer.gov.bc.ca/redhat-access-docker-remote/ubi8/" -p SOURCE_CONTEXT_DIR=$(SOURCE_CONTEXT_DIR)backend -p DOCKER_FILE_PATH=apps/db-migrations/Dockerfile -p NAME=$(DB_MIGRATIONS_BUILD_REF) | oc -n $(BUILD_NAMESPACE) apply -f -
+	@oc -n $(BUILD_NAMESPACE) process -f $(BUILD_TEMPLATE_PATH) -p TAG=$(BUILD_REF) -p SOURCE_REPOSITORY_REF=$(BUILD_REF) -p SOURCE_CONTEXT_DIR=$(SOURCE_CONTEXT_DIR)backend -p DOCKER_FILE_PATH=apps/db-migrations/Dockerfile -p NAME=$(DB_MIGRATIONS_BUILD_REF) | oc -n $(BUILD_NAMESPACE) apply -f -
```
dheepak-aot, line 245: 👍

sources/packages/backend/apps/api/Dockerfile
```diff
@@ -23,5 +23,10 @@ RUN npm run build api
 # Exposing application port
 EXPOSE ${PORT}
 
+# Grant access to group 0 to allow npm v9 to work
+# See: https://docs.openshift.com/container-platform/4.11/openshift_images/create-images.html#use-uid_create-images
+RUN chgrp -R 0 ./.npm && \
```
andrepestana-aot, line 28: If the problem is in the cache files' permissions, why we don't clean the cache and let it be created with the running (whichever it is) like in https://stackoverflow.com/a/60192811 ?

sources/packages/backend/apps/api/Dockerfile
guru-aot: Thanks for the suggestions @andrewsignori-aot and @dheepak-aot , as suggested, i have updated the group permissions for user in group 0 to have write access to the folder./.npm and tested the build and deployments.

sources/packages/backend/apps/api/Dockerfile
guru-aot: We should not use sudo in our Dockerfile and yes while running cache clean, the arbitrary user that is trying to clean it should have write access to the folder. So as per the suggestions, I have given group write permissions for the group 0, where those users are assigned.

andrepestana-aot: The idea is not to use sudo, the idea is to run  `npm cache clean --force`.

guru-aot: the error actually happens not on the build process, rather when there is an error and the crash loop of a pod happens and it restarts. SO when the pod tries to run again, the cache folder does not have write permission by the arbitrary users and it fails.

dheepak-aot: @andrepestana-aot first of all, the problem is not just about the cache file permissions. The problem is about the access for the arbitary user(which openshift uses to run the container) to the to the directory `/opt/app-root/src/.npm` and it's child directories and files. 

By default, this directory is owned by the default user(1001) but the container is started with arbitrarily assigned user who does not have write access to the `.npm` directory.

image

arbitrarily assigned user
image

Hence it would be required to give permission to `.npm` directory to make sure npm life cycle scrips can work without any issue. 

Please let me know if I am missing something or any further clarification needed.

dheepak-aot (approved):
Thanks for making the changes. 👍
Looks good.

ann-aot (approved):
Great research and thanks for the walk-though @guru-aot 👍

sources/packages/backend/apps/api/Dockerfile
andrepestana-aot: I'm asking because the error we see in the ticket refers to the cache:
image

Did we have that error for any other file outside _cacache dir?

dheepak-aot: Yes, in the same screenshot we have the same issue for _logs directory as well.

andrewsignori-aot (approved):

sources/packages/backend/apps/api/Dockerfile
andrepestana-aot: Hm, I see. I did some test to check if the `npm cache clean --force` would delete the _logs but it doesn't. It does remove the _cacache.

andrepestana-aot (approved):
Looks good. Thanks for the explanations.

sources/packages/backend/apps/api/Dockerfile
andrewsignori-aot: Thanks for the further verifications @dheepak-aot 😉

Issue #2539, opened 2023-11-28
Apply / Receive PD / PPD Profile Verification (Part-2) 


michesmith:
The Below acceptance Criteria was initially listed as supporting tasks in ticket #2518. They have been pulled out in order to reduce the scope of the previous ticket. 

**Acceptance Criteria** 
- [x] Suspend any ATBC integration that is already configured within SIMS (for now)
- [x] Review the current SFAS-SIMS `pd_status` mapping logic, As per the current logic, we have mapped `false` with `declined`. But ATBC is not sending any `declined` status. Else Not Requested.
- [x] Remove Student API endpoint decorators to block ATBC API call

Issue #2524, opened 2023-11-22
Hide FT Programs for PT release


HRAGANBC:
**Description of Task**
Hide all FT-only programs and all FT offerings from the Student application dropdown menu so they are unavailable at the initial SIMS launch date. 

**Context:**
- Institutions will continue inputting FT programs and offerings, and should continue to view/manage them from within their account
- Student applications must be restricted to PT-only offerings to ensure that no FT applications inadvertently get submitted. 

**Acceptance Criteria**
- [x] Hide all FT-only programs from the student dropdown menu
- [x] Hide all FT offerings from the student dropdown menu
- [x] Make solution environmentally aware (based on configuration) so that it can be toggled off/on for test/prod environment.
- [x] Ensure that application submit/draft will accept only PT applications when the toggle is on. Execute the check based on the toggle.
- [x] Ensure that the dropdown for PT/FT in the student application will be hidden and the default value will be PT.
- [x] Disable FT schedulers based on toggle (off/on).

guru-aot:
Leverage the isActive flag in queue configurations trying to make the queues configurable.

andrewsignori-aot:
As pointed out by @dheepak-aot, the system currently allows a part-time offering to be created for a Full-time program. It will generate a Pending offering but the Ministry will have the power to approve it. This specific offering will not be selectable because the program itself would never be displayed to the student as it is right now.
<img width="833" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOGZ2QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--eb21a849aa7fd3a18ce872032941fd1c329f6e98/image.png" alt="image.png" />

Right now we will not be considering this scenario and the ticket will be implemented as stated in the current AC: "Hide all FT-only programs from the student dropdown menu".

@CarlyCotton will follow up from the business side if such scenario is needed.

@HRAGANBC @Joshua-Lakusta - FYI

CarlyCotton:
@andrewsignori-aot Hiding FT Only programs is the expectation. Even if the Ministry mistakenly approves a PT offering for a FT only program.

PR #2604, opened 2023-12-15
#2539 apply receive pd ppd profile verification - Part 1

guru-aot:
**Acceptance Criteria** 
- [x] Suspend any ATBC integration that is already configured within SIMS (for now)
- [x] Review the current SFAS-SIMS `pd_status` mapping logic, As per the current logic, we have mapped `false` with `declined`.  But ATBC is not sending any declined status. Else Not Requested.
- [x] Remove Student API endpoint decorators to block ATBC API call

- As part of suspending any calls to API - commented the atbc student controller to disable it from external calls.
- Created a new column is _active in the queue-configurations which enables the queue to be active or not while fetching them from DB. As we are verifying the queues in the scheduler creation and queue constants too, commenting them will ultimately remove the queues from the bull dashboard and disabling them.
- Skipped the test cases accordingly.
- Added the TODO comments for easy reverting of the changes.
- > "TODO: This is commented as part of #2539 - Suspend any ATBC integration." 
- > "TODO: This is disabled as part of #2539 - Suspend any ATBC integration."
- > "TODO: This is commented as part of #2539 - Suspend any ATBC integration."


sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -687,9 +687,6 @@ export class StudentService extends RecordDataModelService<Student> {
     if (pdVerified) {
       return DisabilityStatus.PD;
     }
-    if (pdVerified === false) {
-      return DisabilityStatus.Declined;
-    }
```
andrepestana-aot, line 692: Shouldn't we return `DisabilityStatus.Requested` ?

andrepestana-aot (commented):
Good job. I asked a question.

sources/packages/backend/apps/api/src/services/student/student.service.ts
guru-aot: That is what we explained in today's meeting. ATBC is not sending any false status. So the default is NotRequested.

sources/packages/backend/apps/api/src/app.students.module.ts
```diff
@@ -76,7 +75,7 @@ import { ATBCIntegrationModule } from "@sims/integrations/atbc-integration";
     StudentAccountApplicationStudentsController,
     EducationProgramOfferingStudentsController,
     RestrictionStudentsController,
-    ATBCStudentController,
+    //ATBCStudentController, // TODO: This is commented as part of #2539 - Suspend any ATBC integration.
```
dheepak-aot, line 78: If the controller is removed from the module, I would recommend not removing the decorator in the controller.

Also, I would say not to keep any commented code. Just having the TODO comment will be mor than sufficient to enable the controller when it is required.

sources/packages/backend/apps/api/src/route-controllers/atbc/atbc.students.controller.ts
```diff
@@ -20,7 +16,7 @@ import { DISABILITY_REQUEST_NOT_ALLOWED } from "../../constants";
 
 @AllowAuthorizedParty(AuthorizedParties.student)
 @RequiresStudentAccount()
-@Controller("atbc")
+//@Controller("atbc") // TODO: This is commented as part of #2539 - Suspend any ATBC integration.
```
dheepak-aot, line 19: As mentioned in https://github.com/bcgov/SIMS/pull/2604/files#r1428580193

I would say decorator can remain as it is.

sources/packages/backend/apps/api/src/services/student/student.service.ts
dheepak-aot: This transformation is applied on sfas data which comes through the feed. 

As per the discussion at standup today, the user story had wrong information about mapping sfas import pd_status = false -> `DisabilityStatus.Requested`

only mapping what we need to do here is pd_status = false -> `DisabilityStatus.PD`

@guru-aot as discussed we can add the context as comment to the method and extend the comment when we have ppd status

sources/packages/backend/apps/db-migrations/src/migrations/1702672829350-AddColQueueConfigurationsIsActive.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class AddColQueueConfigurationsIsActive1702672829350
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Add-col-queue-configurations-is-active.sql", "Queue"),
```
dheepak-aot, line 9: I would say, to keep it simple we can remove the queue from the enum with a TODO comment.

image

dheepak-aot: image

andrepestana-aot (approved):
Looks good!

dheepak-aot:
@guru-aot IMO, in spite of having the `is-active` column in DB, we are not achieving a configurable solution here. i.e. for all the the schedulers/queues that we want to disable, we still need to remove the processor class from the `QueueConsumers` module and from `QueueNames` and if we update the isActive without doing this, the code will fail.

Hence in my suggestion,1.  we can just remove the processor class(ATBCResponseIntegrationScheduler) from the `QueueConsumers` module, which is very much sufficient to disable the scheduler. 2. And optionally if we need to make the the Queue not appear in the Bull dashboard and not part of queue module we can remove from the enum and remove enum reference. 

IMO 1. is more than enough and secure.

We can talk as team about this to proceed further.

sources/packages/backend/libs/utilities/src/queue.constant.ts
```diff
@@ -23,7 +23,7 @@ export enum QueueNames {
   PartTimeMSFAAProcessResponseIntegration = "part-time-msfaa-process-response-integration",
   FullTimeMSFAAProcessResponseIntegration = "full-time-msfaa-process-response-integration",
   SFASIntegration = "sfas-integration",
-  ATBCResponseIntegration = "atbc-response-integration",
+  // ATBCResponseIntegration = "atbc-response-integration", // TODO: This is commented as part of #2539 - Suspend any ATBC integration.
```
dheepak-aot, line 26: Please remove the commented code.

sources/packages/backend/apps/api/src/app.students.module.ts
dheepak-aot, line 78: Still we have commented code.

sources/packages/backend/apps/db-migrations/src/sql/Queue/Add-col-queue-configurations-is-active.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  "queue_configurations"
+ADD
+  COLUMN IF NOT EXISTS "is_active" BOOLEAN NOT NULL DEFAULT TRUE;
```
dheepak-aot, line 4: We stopped using IF NOT EXISTS

sources/packages/backend/apps/db-migrations/src/sql/Queue/Drop-col-queue-configurations-is-active.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE
+  "queue_configurations" DROP COLUMN IF EXISTS "is_active";
```
dheepak-aot, line 2: https://github.com/bcgov/SIMS/pull/2604/files#r1431941446

sources/packages/backend/apps/db-migrations/src/migrations/1702672829350-AddColQueueConfigurationsIsActive.ts
dheepak-aot, line 9: Disable-atbc-response could be here so that we do not need a migration script as revert will remove the column

sources/packages/backend/apps/db-migrations/src/sql/Queue/Drop-col-queue-configurations-is-active.sql
ann-aot, line 2: `sims.queue_configurations`?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/atbc-respone-integration/atbc-response-integration.scheduler.ts
```diff
@@ -12,10 +11,8 @@ import {
 /**
  * Process all the applied PD requests to verify the status with ATBC.
  */
-@Processor(QueueNames.ATBCResponseIntegration)
```
ann-aot, line 15: we are commenting `ATBCResponseIntegrationScheduler` in the queue_consumer module right? so still we need to comment this decorator? and line 18?

ann-aot (commented):
Good work @guru-aot 👍  some same comment as other devs and added some comments and a qn

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/atbc-respone-integration/atbc-response-integration.scheduler.ts
guru-aot: yes we need to, as we are removing the queuenames from the queue constants.

sources/packages/backend/apps/db-migrations/src/migrations/1702672829350-AddColQueueConfigurationsIsActive.ts
guru-aot: The reason for having the script separately is to enable a process in the future tickets if we need to disable any other schedulers. Yes you are right, the revert is going to remove the column completely. But as per the order of the migration and revert, the rollback disable will make the is_active flag true and then remove the column in the next revert.

ann-aot (approved):
LGTM 👍 Thanks for doing the changes

sources/packages/backend/apps/db-migrations/src/migrations/1702672829350-AddColQueueConfigurationsIsActive.ts
dheepak-aot: Im ok with that @guru-aot . It was just a suggestion.

dheepak-aot (approved):
Looks good 👍Thanks for making the changes. Please add a comment or a note in #2524 to leverage the `isActive` flag in queue configurations trying to make it configurable.

Issue #2488, opened 2023-11-08
Ensure offering parent id is always present


andrewsignori-aot:
While creating a new offering or cloning an offering, ensure that the parent offering ID is always present.
This will required to ensure that IER12 always has an offering parent ID present to allow better tracking by the institution.

**Acceptance criteria**
- [x] While an offering is created, ensure that the parent id is also populated with the newly created id.
- [x] Check places cloning offerings and also ensure that they have the parent id populated.
- [x] Display the Parent offering ID on the Offering Form
- [x] Assess the possibility to change parent offering id column to be mandatory and populating it - [_**decision added in the comment below**_]

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeURqQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f999baa54b3a2f95db2d0094d40f5a1f592d678a/image.png)

ann-aot:
**_Assess the possibility to change parent offering id column to be mandatory and populating it_**
when offering is created via bulk upload, we are doing `typeorm .insert()`, with which we cannot assign the primary key to the parent offering id, since the offering does not exist at that point.
`error: error: column "id" does not exist` is the error DB is throwing, when we try to do so.
And we don't want to go that `.save()`, which will solve issue, because of the performance issue that was detected during load test.
So, we decided to keep the insert with the parent offering id (which means the parent offering id can't be a mandatory field), And once the offerings are created, we do a bulk update to update the parent offering id.

PR #2608, opened 2023-12-19
#2488 - Ensure offering parent id is always present

ann-aot:
- Primary ID is added as the parent offering ID when a new offering (both bulk and single) is created.
-  Parent offering ID is also copied and saved during cloning.
- Parent offering ID is displayed on the Offering Form in both the Ministry and Institution portals.
- Assessed the possibility of changing the parent offering ID column to be mandatory and populating it 
> When the offering is created via bulk upload, we are doing typeorm `.insert()`, with which we cannot assign the primary key to the parent offering id, since the offering does not exist at that point.
`error: column "id" does not exist` is the error DB is throwing when we try to do so.
And we don't want to go for the typeorm `.save()`, which will solve the issue, but because of the performance issue that was detected during the load test.
So, after discussing with @andrewsignori-aot  we decided to keep the `.insert()` to create the bulk offering upload without the parent offering ID (which means the parent offering ID can't be a mandatory field during the creation), And once the offerings are created, we do a bulk update to update the parent offering id.
- Migration added to populate the parent offering id as primary id for null parent offering id.
- Currently, I have joined the `parent offering ID` with `left join`, that is the reason the parent_id is optional in `DTOs/Interfaces ` and transformations, let me know your thoughts. Open to suggestions
- sonar cloud is showing duplication in the code, that is not part of the PR.

**Screenshots**
**Institution**
**_Newly created bulk offering_**
image
**_Newly created offering via the offering form_**
image
**Ministry**
image


sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -132,17 +135,49 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
         if (promises.length >= maxPromisesAllowed) {
           // Waits for all be processed.
           const insertResults = await Promise.all(promises);
+          await this.saveBulkOfferingParentId(
+            insertResults.map((result) => result.createdOfferingId),
+            offeringRepo,
+          );
```
lewischen-aot, line 141: I'm trying to understand this part of the code. Can you briefly explain it?  I feel like this part is a little redundant if we can push `insertResults` to `saveBulkOfferingParentId` outside the loop from `allResults` before `allResults` push `finalResults`. I could be wrong about this. Let me know your thoughts.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -132,17 +135,49 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+      );
       allResults.push(...finalResults);
       return allResults;
     });
   }
 
+  /**
+   * Update offering parent ids for given offerings.
+   * * When creating a new offering, parent id and
+   * * the primary id are the same.
```
andrepestana-aot, line 160: Are the double stars intentional?
image

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -311,6 +347,7 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
       .innerJoin("offerings.institutionLocation", "institutionLocation")
       .innerJoin("institutionLocation.institution", "institution")
       .leftJoin("offerings.assessedBy", "assessedBy")
+      .leftJoin("offerings.parentOffering", "parentOffering")
```
andrepestana-aot, line 350: Should we use `innerJoin`? @guru-aot @dheepak-aot @lewischen-aot Thoughts?

andrepestana-aot (commented):
Great job. I just left some comments.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
ann-aot: yes @andrepestana-aot, But I was not expecting it to be a bulletin point. I will remove it.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
ann-aot: @lewischen-aot `allResults` can have all the offerings in the file, which can exceed the IN() limit. That is the reason we are doing the update as a batch.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
ann-aot: yup, I would like to know dev's thoughts on this, that's why I added it in the PR description and dev chat

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
lewischen-aot: Thanks @ann-aot I see that now. Conversation is resolved.

lewischen-aot (approved):
LGTM. Great Work 👍 Please see comments from other DEVs.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
dheepak-aot: I would agree with @andrepestana-aot on inner join as we do not expect any offering without parent offering id(with the migration we added)

@ann-aot I saw the reason in the PR description, IMO I would like to see the API fail when there is no parent_id to an offering as it could be manipulated in DB.

ann-aot: updated to `leftJoin` and made `parentofferingid` mandatory in interfaced and dto

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -132,17 +135,49 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
         if (promises.length >= maxPromisesAllowed) {
           // Waits for all be processed.
           const insertResults = await Promise.all(promises);
+          await this.saveBulkOfferingParentId(
+            insertResults.map((result) => result.createdOfferingId),
```
dheepak-aot, line 139: I would recommend to create a const for the mapped result.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -95,6 +96,8 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
     );
     programOffering.offeringStatus = offeringValidation.offeringStatus;
     programOffering.creator = { id: userId } as User;
+    // When creating a new offering, parent id and the primary id are the same.
+    programOffering.parentOffering = programOffering;
```
dheepak-aot, line 100: Is this still effective with using the insert and update late? I believe we can remove it.

sources/packages/backend/apps/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
```diff
@@ -136,12 +136,18 @@ export class StudentScholasticStandingsService extends RecordDataModelService<St
       // Get existing offering.
       const existingOffering = await this.offeringRepo
         .createQueryBuilder("offering")
-        .select(["offering", "educationProgram.id", "institutionLocation.id"])
+        .select([
```
dheepak-aot, line 139: @ann-aot I can very certainly recollect that we discussed in grooming about setting parent offering id to the cloned offering. Did anything change ? I know it was not very explicit in the AC of the ticket. But I thought it was implied in `while an offering is created`

ann-aot: @dheepak-aot  During scholastic standing cloning, since `parent_offering_id` was not retrieved in the offering to be cloned, the new cloned offering did not have the parent offering id, so, here the change is to retrieve `parent_offering_id`, so it will copied to the new cloned offering.
The AC had a point to check that,
image

dheepak-aot: I got it @ann-aot. Thanks for the explaination.

dheepak-aot (approved):
Looks good @ann-aot Thanks for making the changes.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
andrepestana-aot: Looks like it broke some test.

guru-aot: i believe we should go with the innerjoin @ann-aot, i know the current implementation. BUt the part you made the leftJoin and made parentOfferingId mandatory solves the problem without failing the API.

sources/packages/backend/libs/test-utils/src/factories/application.ts
```diff
@@ -238,6 +238,7 @@ export async function saveFakeApplication(
   });
   fakeOffering.offeringIntensity =
     options?.offeringIntensity ?? OfferingIntensity.fullTime;
+  fakeOffering.parentOffering = fakeOffering;
```
andrepestana-aot, line 241: 👍

andrepestana-aot (approved):
Looks good! Thanks for the discussion and changes!

guru-aot (approved):
LGTM, nice work @ann-aot

Issue #2466, opened 2023-11-01
Request a Change Forms - PD/PPD


HRAGANBC:
**User Story** 
As a Student, I want to change my PD PPD flag on a complete application so that I can.... 

**Description of Task:**
Allow students to change their application PD/PPD flag on a complete application. 

**Context:**
- See #2467 for context about general Request a Change Process

**Acceptance Criteria**

**Student Account**
- [x] Introduce new 'form title' called 'Disability Information' (_See Student Form Selection View_ photo)
- [x] Attach the application question "Do you want to be assessed for disability funding..." to the 'Disability Information' form (_See Student Field Inputting View_ photo)
- [x] Build the Student Review Completed Request Screen (see photo below)

**Ministry Account**
- [x] Build the "Ministry Review Request" screens (See photo below) 
- [x] Change the following logic in the workflows:
- If request is denied, do nothing
- If student _**requests**_ disability assessment and ministry approves request, generate new assessment with PD/PPD funding, but apply standard disbursement rules which require approved disability profile status prior to disbursement. 
- It student _**removes**_ disability assessment request and ministry approved request, generate new assessment without PD/PPD funding, and remove any PD/PPD conditions blocking disbursement. 
- [x] See #2467 for other details about how request interacts with COE requests, other pending reassessment requests, etc... 
- [x] Add E2E Test

**Student Form Selection View**
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMjZTQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9a94f6f5cfbdbece0e97e16e7d10afb4ac1610c2/image.png)

**Student Field Inputting View**
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeGlTQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cb1bf3dd162cd31798b9cb2dd9081b0899e068b6/image.png)

**Ministry Review Request View**
[...]

PR #2636, opened 2024-01-02
#2466 - Fix for UI issue

andrepestana-aot:
After adding an item to the select component, the last item was partially hidden inside the vcard:
image

With the change the select displays outside the vcard:
image

### DIFF
--- a/sources/packages/web/src/assets/css/formio-shared.scss
+++ b/sources/packages/web/src/assets/css/formio-shared.scss
@@ -705,7 +705,3 @@
 .formio-component-file ul li.list-group-item {
   border: 1px solid rgba(0, 0, 0, 0.125) !important;
 }
-
-.overflow-visible .v-card {
-  overflow: visible;
-}
--- a/sources/packages/web/src/assets/css/formio-shared.scss
+++ b/sources/packages/web/src/assets/css/formio-shared.scss
@@ -705,3 +705,7 @@
 .formio-component-file ul li.list-group-item {
   border: 1px solid rgba(0, 0, 0, 0.125) !important;
 }
+
+.overflow-visible .v-card {
+  overflow: visible;
+}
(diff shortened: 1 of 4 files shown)


sh16011993 (approved):
👍

sources/packages/web/src/components/layouts/FullPageContainer.vue
```diff
@@ -11,7 +11,7 @@
     <v-container :fluid="fullWidth">
       <template v-if="layoutTemplate === LayoutTemplates.CenteredCard">
         <v-row justify="center">
-          <v-card class="mt-4 p-4 w-100" :class="widthClass">
+          <v-card class="mt-4 p-4 w-100 overflow-visible" :class="widthClass">
```
dheepak-aot, line 14: Do we need to add this overflow property at container level? because we set the overflow property in different places depending upon the need. I know the default property of overflow is visible but if that is not happening do we need to force it at container level?

image

andrepestana-aot: Yes, I changed that as we talked about it. Thanks for the help checking that.

sources/packages/web/src/assets/css/formio-shared.scss
```diff
@@ -705,3 +705,7 @@
 .formio-component-file ul li.list-group-item {
   border: 1px solid rgba(0, 0, 0, 0.125) !important;
 }
+
+.overflow-visible .v-card {
```
dheepak-aot, line 709: any style connected to components of vuetify we add the class to vuetify.scss. Can you please move it there?

dheepak-aot (approved):

guru-aot (approved):

lewischen-aot (approved):
LGTM 👍

Issue #2567, opened 2023-12-06
Offering index constrain needs to be updated


ann-aot:
- we need to update/refactor the `location_id_program_id_offering_name_study_dates_index` index constrain  for scholastic standing cloning purpose
- when we create or update an offering, we need the `location_id_program_id_offering_name_study_dates_index` constrain. but when we clone an offering for a scholastic standing purpose, where the new study end date can be same for multiple application for the same offering, we don't need/ignore the constrain, i.e system/db should allow to create multiple offering with same location id, program id, offering name and study dates during a scholastic standing cloning

```
query failed: INSERT INTO "sims"."education_programs_offerings"("created_at", "updated_at", "offering_name", "study_start_date", "study_end_date", "actual_tuition_costs", "program_related_costs", "mandatory_fees", "exceptional_expenses", "offering_delivered", "lacks_study_breaks", "offering_type", "offering_intensity", "year_of_study", "show_year_of_study", "has_offering_wil_component", "offering_wil_type", "study_breaks", "offering_declaration", "assessed_date", "submitted_date", "offering_status", "course_load", "creator", "modifier", "program_id", "location_id", "assessed_by", "offering_note", "parent_offering_id", "preceding_offering_id") VALUES ($1, DEFAULT, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, DEFAULT, $24, $25, DEFAULT, DEFAULT, DEFAULT, DEFAULT) RETURNING "created_at", "updated_at", "id" -- PARAMETERS: ["2023-12-06T17:36:37.606Z","Ann test- study break for scholastic standing","2023-10-01","2023-12-30",1,1,1,1,"onsite",0,"Scholastic s...
5052error: error: duplicate key value violates unique constraint "location_id_program_id_offering_name_study_dates_index"
5053[Nest] 17 - 12/06/2023, 5:36:37 PM ERROR [AppAllExceptionsFilter] Unhandled exception
5054[Nest] 17 - 12/06/2023, 5:36:37 PM ERROR [AppAllExceptionsFilter] Request path [/api/institutions/scholastic-standing/location/3/application/106998]
[...]

Issue #2484, opened 2023-11-08
Incorporate 'Year of Study' Field into All Offering Names (mandatory); Change Duplicate Offering Validation   


CarlyCotton:
**Description of Task**
Revise the 'duplicate offering name' validation to allow duplicate offering names **IF** the year of study is different

**Current Behaviour**
- Currently the PSIs cannot see the 'Year of Study' as part of the offering name (when show = yes) when viewing the offerings for a program. 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2V4QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--52a705f1a997d005111281f81e1ff0cde3a605b1/image.png)
- When trying to create offerings, via SIMS or bulk upload, it prevents creation of new offerings when name, start date and end date have a match.  Even though these are unique offerings based on a difference in the year of study.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2l4QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0764b2e3011a4393d0adb1906bdb5a5b926efcbf/image.png)

**Expected Behaviour**
- "Show Year of Study" will no longer be optional and default to "Yes"
- Institutions can view Year of Study in the offering names with viewing offerings.
- When creating new offerings, uniqueness will be determined by: offering name, year of study, start & end dates.
- Reject new offerings from being created if the 4 fields are the same: offering name, year of study, start & end dates.

**Sample data**
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBenV5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d4cc33378cb046f4bde2270de36a7b40e0da77ae/image.png)
Rows being flagged as duplicates that we would like to allow as unique based on year of study being different:

- 2 + 12
- 3 + 11
- 4 + 9
- 5 + 7
- 6 + 10

[...]

PR #2643, opened 2024-01-05
#2484 - Incorporate 'Year of Study' Field into All Offering Names

andrepestana-aot:
- Added `year_of_study` to the constraint and excluded the offering type from the constraint conditions([#2567](https://github.com/bcgov/SIMS/issues/2567));
- Changed blurb according to business confirmation: The old banner should be adapted to display: "The Year of Study will be displayed as part of the offering name. Here's an example of what students will see." and the words just below the Name field should have "Year of Study" removed:
  image
- Displaying the Year of Study in the dropdowns:
  image
- Added tests for bulk upload, add offering and update offering;

### Bulk upload
image

### Add Offering
image

### Update Offering
image

### Observations:
- The whole file for [education-program-offering.model.ts](https://github.com/bcgov/SIMS/pull/2643/files#diff-50f528fd7d90633261e54ea904b6bc255ef9331a6830eae24aa5e5242024d21c) looks different because of the wrong EOL in the file.
- [Update-offering-duplication-unique-index.sql](https://github.com/bcgov/SIMS/pull/2643/files#diff-8ce55f89538e8f5979a8d454684e2c92dc86f587f4b1b424cd4f97ac4a589f53) had a `commit;` at the beginning of the file because of an error with the migration: "driverError: error: unsafe use of new value "Scholastic standing" of enum type offering_types". That means that the `sims.offering_types` was updated but wasn't committed. We had a [discussion](https://github.com/bcgov/SIMS/pull/2643#pullrequestreview-1817023604) and changed it to avoid the issue.


sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.institutions.controller.ts
```diff
@@ -158,11 +159,15 @@ export class EducationProgramOfferingInstitutionsController extends BaseControll
...
   })
+  @ApiNotFoundResponse({
+    description: "Not able to find the program in the provided location.",
+  })
   @ApiUnprocessableEntityResponse({
     description: "Not able to a create an offering due to an invalid request.",
   })
-  @ApiNotFoundResponse({
-    description: "Not able to find the program in the provided location.",
+  @ApiUnprocessableEntityResponse({
```
andrewsignori-aot, line 168: As far as I remember adding multiple `ApiUnprocessableEntityResponse` will not result in the expected description in Swagger. Other codes concatenate the strings when multiple are possible, as shown below.
```ts
@ApiUnprocessableEntityResponse({
    description:
      "Not possible to create an MSFAA due to an invalid application status, or " +
      "not possible to reissue an MSFAA when there is no pending disbursements for the application, or " +
      "not possible to reissue an MSFAA when the current associated MSFAA is not cancelled.",
  })
```

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-import-csv.service.ts
```diff
@@ -55,7 +55,8 @@ export class EducationProgramOfferingImportCSVService {
     return csvModels.map((csvModel) => ({
       offeringName: csvModel.offeringName,
       yearOfStudy: csvModel.yearOfStudy,
-      showYearOfStudy: csvModel.showYearOfStudy === YesNoOptions.Yes,
+      // showYearOfStudy will not be used anymore
+      showYearOfStudy: true,
```
andrewsignori-aot, line 59: Even being part of the CSV should we keep parsing this?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -98,7 +99,23 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
     programOffering.creator = { id: userId } as User;
     // When creating a new offering, parent id and the primary id are the same.
     programOffering.parentOffering = programOffering;
-    return this.repo.save(programOffering);
+
+    try {
+      return await this.repo.save(programOffering);
```
andrewsignori-aot, line 104: `await` is not needed here.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -98,7 +99,23 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+    } catch (error: unknown) {
+      if (error instanceof QueryFailedError) {
```
andrewsignori-aot, line 106: These checks are now happening in many places. As a suggestion, if we change `PostgresDriverError` to a class, can the code be simplified as the below?
```ts
if (error instanceof PostgresDriverError && error.constraint ===
  DatabaseConstraintNames.LocationIDProgramIDOfferingNameStudyDatesYearOfStudyIndex) {
    throw new CustomNamedError(
      "Duplication error. An offering with the same name, year of study, start date and end date was found.",
      OFFERING_SAVE_UNIQUE_ERROR,
    );
  }
}
```

sources/packages/backend/apps/db-migrations/src/migrations/1704494480880-UpdateOfferingUniqueIndex.ts
```diff
@@ -0,0 +1,24 @@
...
+        "Update-offering-duplication-unique-index.sql",
+        "EducationProgramsOfferings",
+      ),
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData(
+        "Rollback-update_offering-duplication-unique-index.sql",
```
andrewsignori-aot, line 19: Please replace the "_" in the file name.

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Add-col-show-year-of-study.sql
```diff
@@ -0,0 +1,7 @@
+-- Add show_year_of_study to education_programs_offerings table.
+ALTER TABLE
+    sims.education_programs_offerings
+ADD
+    COLUMN offering_note;
```
andrewsignori-aot, line 5: Should it be `show_year_of_study` instead `offering_note`?

andrewsignori-aot, line 5: The type is also missing `how_year_of_study BOOLEAN`.
Please ensure that the rollback scripts are tested locally also 😉

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Rollback-update_offering-duplication-unique-index.sql
```diff
@@ -0,0 +1,60 @@
...
+SET
+    offering_name = offering_name || ' - updated to be unique - ' || id :: text
+WHERE
+    id IN (
+        SELECT
+            offerings.id
+        FROM
+            sims.education_programs_offerings offerings
+            INNER JOIN (
+                SELECT
```
andrewsignori-aot, line 16: @JasonCTang can we have this `SELECT` (**the SELECT part only**) query executed on PROD to have a better feeling if it would be updating any PROD record?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-import-csv.service.ts
andrepestana-aot: Ok. Removed that.

sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.institutions.controller.ts
```diff
@@ -156,13 +157,13 @@ export class EducationProgramOfferingInstitutionsController extends BaseControll
   @HasLocationAccess("locationId")
   @ApiNotFoundResponse({
     description:
-      "Program to create the offering not found for the institution.",
+      "Program to create the offering not found for the institution or " +
+      "Not able to find the program in the provided location.",
```
andrewsignori-aot, line 161: Minor, since it is part of the same sentence it should start with lower case to be correctly displayed on swagger.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
andrepestana-aot: If I remove the await the promise will be returned and the thrown error is not caught. Is there any other way to solve this?

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Update-offering-duplication-unique-index.sql
```diff
@@ -0,0 +1,26 @@
+-- Commits the addition of the new offering_type 'Scholastic standing'
+commit;
```
andrewsignori-aot, line 2: Unless I am missing something this solution would not be desirable or we need to at least discuss it as a team.
If my understanding is correct it is actually committing the parent transaction created by typerom to execute all pending SQL migrations.
I am assuming that the error happens when the database is built from scratch to execute, for instance, the E2E tests.
We can change the typeorm to execute the transaction per migration for E2E tests (`npx typeorm migration:run -t each`) and leave the "PROD" migrations as they are right now unless DEVs would be willing to change it all to be execute with the `each` mode. Last time that it was proposed the decision was not to use the `each` for PROD.
@dheepak-aot @guru-aot @sh16011993 @lewischen-aot thoughts?

sources/packages/backend/libs/sims-db/src/entities/education-program-offering.model.ts
```diff
@@ -1,295 +1,287 @@
...
+   * Offering year of study.
+   */
+  @Column({
+    name: "year_of_study",
+  })
+  yearOfStudy: number;
+
+  /**
+   * Determines if the offering has WIL component.
+   * */
```
andrewsignori-aot, line 153: There is white space added between the `* *` that is not part of the main branch right now.

sources/packages/backend/libs/sims-db/src/entities/education-program-offering.model.ts
andrewsignori-aot, line 1: Thanks for fixing the format ❤️

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
andrewsignori-aot: My bad!

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -166,16 +181,29 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
   ): Promise<UpdateResult> {
-    return offeringRepo.update(
-      {
-        id: In(newOfferings),
-      },
-      {
-        parentOffering: {
-          id: () => "id",
+    try {
+      return offeringRepo.update(
```
andrewsignori-aot, line 185: Is the await missing here?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -203,18 +231,17 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
-            validatedOffering,
-            "Duplication error. An offering with the same name, start date, and end date was found. Please remove the duplicated offering and try again.",
-          );
-        }
+      if (
+        error instanceof PostgresDriverError &&
+        error.constraint ===
+          DatabaseConstraintNames.LocationIDProgramIDOfferingNameStudyDatesYearOfStudyIndex
+      ) {
+        throw new CreateFromValidatedOfferingError(
```
andrewsignori-aot, line 239: Please start the methods with lower case 😉

andrewsignori-aot (commented):
Great work handling all the different impacted areas, please take a look in the comments.

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Update-offering-duplication-unique-index.sql
andrepestana-aot: why should the index be targeting only 'Public' offerings?

dheepak-aot:
Please add the required labels to PR.

sources/packages/web/src/views/institution/locations/offerings/OfferingCreate.vue
```diff
@@ -75,10 +80,14 @@ export default defineComponent({
         );
         snackBar.success("Offering created.");
         goBack();
-      } catch {
-        snackBar.error(
-          "Unexpected error happened while creating the offering.",
-        );
+      } catch (error: unknown) {
```
dheepak-aot, line 83: 👍

sources/packages/backend/apps/db-migrations/src/migrations/1704480832432-DropShowYearOfStudyColumn.ts
```diff
@@ -0,0 +1,24 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class DropShowYearOfStudyColumn1704480832432
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
dheepak-aot, line 7: Not a part of the PR. Noticed that we have been using public access specifier explicitly in migrations.

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Update-offering-duplication-unique-index.sql
andrewsignori-aot: Because they are the ones where the duplication would cause some impact and also the ones imported during the bulk upload. Private are student-specific, and, unless businesses argue against it, I believe that the institutions must have the ability to create duplicated ones to satisfy a PIR.
@HRAGANBC @CarlyCotton @Joshua-Lakusta thoughts?

sources/packages/backend/apps/db-migrations/src/migrations/1704480832432-DropShowYearOfStudyColumn.ts
andrewsignori-aot: It is the boilerplate code auto-generated by the Typeorm migration since day one. For this specific situation I did not believe that we should spend time. Removing the `public` does not consume time but having it as part of the review and asking for changes from today till the end of the project will be an unnecessary time consumption IMO.
Unless we can change the way that the code is auto-generated 😉

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Update-offering-duplication-unique-index.sql
guru-aot: i can see the problem here, I am okay to change the execute transaction -t each in e2e tests. If that is the case as per my understanding, we have to taken into account that every migration will be taking a separate transaction so if one migration fails on the way, the rollback will happen for that particular transaction and the db will be in a half migrated state.  Are we okay for this?

dheepak-aot: I agree with @andrewsignori-aot and I strongly believe that

 - First of all, we should do that commit manually in migrations
 - Second, we can modify the query and move forward
 - Third, same issue has happened before. I can see this thread of discussion and we can discuss this broadly in next dev sync up.

image

lewischen-aot: I'm still learning the system, and I can see the problem with the bulk upload indexing or id. I'm just thinking if it would be too much work for having two columns to map each item during bulk upload. One column stores the indexing from bulk upload, the other column stores a unique index from uuid or something. It's a thought here. I could be totally wrong.

sources/packages/backend/libs/sims-db/src/entities/education-program-offering.model.ts
lewischen-aot: Thanks for fixing the format :heart It is probably caused by the wrong file EoL from my previous PR

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Update-offering-duplication-unique-index.sql
andrewsignori-aot: > i can see the problem here, I am okay to change the execute transaction -t each in e2e tests. If that is the case as per my understanding, we have to taken into account that every migration will be taking a separate transaction so if one migration fails on the way, the rollback will happen for that particular transaction and the db will be in a half migrated state. Are we okay for this?

@guru-aot for E2E tests I do not see a problem with that. The concern arises for Staging/PROD deployments only, IMO.

CarlyCotton: Institutions don't need to create duplicate offerings to satisfy a PIR. They aren't expecting to create them as they need some sort of indicator on their end to identify how/why these are different offerings. We've provided tips in their instructional guide on how to create them during the PIR process.
If they're private and can be duplicated there is possibility for confusion on the Institution side.

andrewsignori-aot: Thanks @CarlyCotton for the inputs. As per the team's decision, we will take that into consideration and change the code as suggested.
We will be using `AND offering_type IN ('Public', 'Private');`

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Update-offering-duplication-unique-index.sql
```diff
@@ -0,0 +1,23 @@
...
+  offering_status,
+  year_of_study
+)
+WHERE
+  (
+    offering_status IN ('Approved', 'Creation pending')
+  )
+  AND offering_type IN ('Public', 'Private');
+
+COMMENT ON INDEX sims.loc_id_prog_id_offer_name_study_dts_year_study_index IS 'Ensures offering in "Approved" or "Creation pending" statuses does not have the same name, study start/end dates and year of study.'
```
andrewsignori-aot, line 23: Minor but maybe the comment can be appended with "Applied only to 'Public' and 'Private' offering types."

sources/packages/backend/libs/sims-db/src/error-handler.ts
```diff
@@ -20,3 +21,17 @@ export enum DatabaseConstraintNames {
 export interface PostgresDriverError extends QueryFailedError {
   constraint: DatabaseConstraintNames;
 }
+
+export function isDatabaseConstraintError(
```
andrewsignori-aot, line 25: Please add some comments.

andrewsignori-aot (approved):
Thanks for making the changes. Please take a look at the two minor comments added but they are not blockers, hence approving it. Looks good 👍

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -98,7 +99,23 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+      return await this.repo.save(programOffering);
+    } catch (error: unknown) {
+      if (error instanceof QueryFailedError) {
+        const postgresError = error as PostgresDriverError;
+        if (
+          postgresError.constraint ===
+          DatabaseConstraintNames.LocationIDProgramIDOfferingNameStudyDatesYearOfStudyIndex
+        ) {
+          throw new CustomNamedError(
+            "Duplication error. An offering with the same name, year of study, start date and end date was found.",
```
sh16011993: @HRAGANBC @CarlyCotton @Joshua-Lakusta Should the Duplication check also include the `Offering Intensity` field as well?

dheepak-aot (approved):
Great Job @andrepestana-aot . Looks Good. 👍

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -166,16 +176,30 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
   ): Promise<UpdateResult> {
-    return offeringRepo.update(
-      {
-        id: In(newOfferings),
-      },
-      {
-        parentOffering: {
-          id: () => "id",
+    try {
+      return await offeringRepo.update(
```
guru-aot, line 180: 👍

sources/packages/backend/libs/sims-db/src/error-handler.ts
```diff
@@ -20,3 +21,17 @@ export enum DatabaseConstraintNames {
...
+  error: unknown,
+  constraint: DatabaseConstraintNames,
+): boolean {
+  if (error instanceof QueryFailedError) {
+    const postgresError = error as PostgresDriverError;
+    if (postgresError.constraint === constraint) {
+      return true;
+    }
+    return false;
```
sh16011993, line 34: Is this necessary?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
CarlyCotton: @sh16011993 Does it currently (prior to this change) include Intensity?

sources/packages/backend/libs/sims-db/src/error-handler.ts
andrepestana-aot: yes, otherwise the function will return undefined.

lewischen-aot (approved):
Good work @andrepestana-aot  LGTM 👍

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
andrepestana-aot: > @sh16011993 Does it currently (prior to this change) include Intensity?

No, it didn't include intensity.

guru-aot (approved):

sources/packages/backend/libs/sims-db/src/error-handler.ts
andrewsignori-aot: The method description comment is still missing 😉

sources/packages/backend/apps/db-migrations/src/sql/EducationProgramsOfferings/Rollback-update-offering-duplication-unique-index.sql
```diff
@@ -0,0 +1,60 @@
+DROP INDEX sims.loc_id_prog_id_offer_name_study_dts_year_study_index;
+
+-- Ensures currently duplicated offerings will be renamed to avoid issues
+-- while creating the unique constraint.
+UPDATE
```
dheepak-aot, line 5: Do we really need an update that modifies the offering name?

I had some doubts regarding rollback impacts and had a discussion with @andrewsignori-aot

andrepestana-aot: If we have some changes to the tables and we need to rollback, yes we'll to update it to satisfy the old unique constraint. What would be your suggestion?

sh16011993 (approved):
Nice Work 👍 @andrepestana-aot

PR #2719, opened 2024-01-23
#2524 - Hide FT Programs for PT release [Part 1]

sh16011993:
## As a part of this PR, the following were completed:

-  Hide all FT-only programs from the student dropdown menu
-  Hide all FT offerings from the student dropdown menu
-  Make solution environmentally aware (based on configuration) so that it can be toggled off/on for test/prod environment.
-  Ensure that application submit/draft will accept only PT applications when the toggle is on. Execute the check based on the toggle.
-  Ensure that the dropdown for PT/FT in the student application shows PT as the default value.
-  Added the `IS_FULLTIME_ALLOWED` environment variable to the github environment variables.
-  Updated the release notes for the next release to incorporate the above environment variable update.

### Screenshot:

image


.github/workflows/release-deploy-all.yml
```diff
@@ -72,6 +72,7 @@ env:
   HOST_PREFIX: ${{ secrets.HOST_PREFIX }}
   BUILD_NAMESPACE: ${{ vars.BUILD_NAMESPACE }}
   KEYCLOAK_REALM: ${{ vars.KEYCLOAK_REALM }}
+  IS_FULLTIME_ALLOWED: ${{ vars.IS_FULLTIME_ALLOWED }}
```
dheepak-aot, line 75: what about `repo-checks-tests.yml`? I see that E2Es are not failing but still we can add the default state of application from E2E perspective

dheepak-aot, line 75: During our conversation yesterday, I forgot to mention about `sources` as well. We need to have this environment variable going to docker compose and good to have in sources/make as well

devops/openshift/api-deploy.yml
```diff
@@ -236,21 +238,21 @@ objects:
...
-              target:
-                type: Utilization
-                averageUtilization: 80
-          - type: Resource
-            resource:
-              name: memory
-              target:
-                type: AverageValue
-                averageValue: "230M"
+        - type: Resource
```
dheepak-aot, line 244: Is there a formatting change here?

devops/openshift/api-deploy.yml
```diff
@@ -350,6 +352,8 @@ parameters:
     value: FORMS_URL
   - name: SITE_MINDER_LOGOUT_URL
     require: true
+  - name: IS_FULLTIME_ALLOWED
+    require: true
```
dheepak-aot, line 356: According to openshift documentation this must be `required` not `require`
@andrewsignori-aot @guru-aot , I see we are using 2 different properties (yml key) `require` and `required`

image

image

devops/openshift/api-deploy.yml
sh16011993: On saving the file, the editor auto-formats.

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -256,14 +257,24 @@ export default defineComponent({
           SELECTED_OFFERING_END_DATE_KEY,
         );
       }
-      // If the user after selecting a study period finds that
-      // they need to check my study period not listed, then
-      // the details of previously selected
-      // study period must be cleared.
+      if (
+        event.changed?.component.key === LOCATIONS_DROPDOWN_KEY &&
+        +event.changed.value > 0
```
lewischen-aot, line 262: There is a + sign here. Is it a typo?

sources/docker-compose.yml
```diff
@@ -51,6 +51,7 @@ services:
       - ATBC_ENDPOINT=${ATBC_ENDPOINT}
       - FILE_UPLOAD_MAX_FILE_SIZE=${FILE_UPLOAD_MAX_FILE_SIZE}
       - FILE_UPLOAD_ALLOWED_EXTENSIONS=${FILE_UPLOAD_ALLOWED_EXTENSIONS}
+      - IS_FULLTIME_ALLOWED=${IS_FULLTIME_ALLOWED}
```
dheepak-aot, line 54: `sources/tests/docker-compose.yml` needs to be updated as well

sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.students.controller.ts
```diff
@@ -73,6 +73,13 @@ export class EducationProgramOfferingStudentsController extends BaseController {
     @Query("offeringIntensity", new ParseEnumQueryPipe(OfferingIntensity))
     offeringIntensity?: OfferingIntensity,
   ): Promise<OptionItemAPIOutDTO[]> {
+    const isFulltimeAllowed = process.env.IS_FULLTIME_ALLOWED === "true";
```
dheepak-aot, line 76: We can get the value from config service

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -171,6 +173,7 @@ export class ApplicationStudentsController extends BaseController {
     @Param("applicationId", ParseIntPipe) applicationId: number,
     @UserToken() studentToken: StudentUserToken,
   ): Promise<void> {
+    const isFulltimeAllowed = process.env.IS_FULLTIME_ALLOWED === "true";
```
dheepak-aot, line 176: https://github.com/bcgov/SIMS/pull/2719#discussion_r1465441828

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -38,6 +43,44 @@ export function useFormioDropdownLoader() {
...
+    );
+  };
+
+  const getProgramIntensityDetails = (isFulltimeAllowed: boolean) => {
+    if (isFulltimeAllowed) {
+      return [
+        {
+          id: "Full Time",
+          description: "Full Time",
+        },
```
lewischen-aot, line 68: Please let me know why we need type string for id for program intensity detail array, like why we want to do `id: "Full Time"` but not `id: 1`. I understand that this is connected to interface ProgramIntensityAPIOutDTO, but we could also use `OptionItemAPIOutDTO` for this story. I'm just not sure why we need another interface instead of the existing ProgramIntensityAPIOutDTO.

sources/packages/web/src/components/common/StudentApplication.vue
sh16011993: Just checked. `event.changed.value` is already a number. So, removed `+`. Thanks.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
sh16011993: The label and the values for the program intensity dropdown in formio were:
Part Time: Part Time
Full Time: Full Time

My code change only ensures that the dropdown is rendered dynamically based on the `IS_FULLTIME_ALLOWED` environment variable rather than being statically present in the form. In the method `loadDropdown` in `useFormioDropdownLoader`, there is a conversion happening where `description` is assigned to the `label` and `id` to the value.

sources/packages/web/src/components/common/StudentApplication.vue
lewischen-aot: Thanks Shashank. I now know that + is to convert string to number. We can keep + as it is.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
lewischen-aot: Thanks Shashank. I now understand this approach and how the app works between vue and formIO. This conversation is resolved.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -187,6 +192,15 @@ export class ApplicationStudentsController extends BaseController {
       const offering = await this.offeringService.getOfferingById(
         payload.data.selectedOffering,
       );
+      if (
+        !isFulltimeAllowed &&
```
dheepak-aot, line 196: The validation is happening only when an offering is selected in the application. It must happen otherwise as well.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -325,12 +339,23 @@ export class ApplicationStudentsController extends BaseController {
...
   async updateDraftApplication(
     @Body() payload: SaveApplicationAPIInDTO,
     @Param("applicationId", ParseIntPipe) applicationId: number,
     @UserToken() studentToken: StudentUserToken,
   ): Promise<void> {
+    const isFulltimeAllowed = this.configService.isFulltimeAllowed;
+    if (
+      !isFulltimeAllowed &&
+      payload.data.howWillYouBeAttendingTheProgram &&
+      payload.data.howWillYouBeAttendingTheProgram === "Full Time"
```
dheepak-aot, line 355: Line 354 is not required `payload.data.howWillYouBeAttendingTheProgram === "Full Time"` should be sufficient

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -395,11 +399,17 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
     }
     offeringExistsQuery.select("1");
-    return this.repo
+    const programsQuery = this.repo
       .createQueryBuilder("programs")
       .where("programs.programStatus = :programStatus", {
         programStatus: ProgramStatus.Approved,
-      })
+      });
+    if (!isFulltimeAllowed) {
```
dheepak-aot, line 407: IMO environment based logic must not be used inside the services.

Service must receive an optional parameter which should lead to where condition.

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -70,6 +70,7 @@ export default defineComponent({
     const LOCATIONS_DROPDOWN_KEY = "selectedLocation";
     const PROGRAMS_DROPDOWN_KEY = "selectedProgram";
     const OFFERINGS_DROPDOWN_KEY = "selectedOffering";
+    const PROGRAM_INTENSITY_DROPDOWN_KEY = "howWillYouBeAttendingTheProgram";
```
dheepak-aot, line 73: IMO the variable must be named as `OFFERING_INTENSITY_DROPDOWN_KEY`

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -256,14 +262,14 @@ export default defineComponent({
...
       }
-      // If the user after selecting a study period finds that
-      // they need to check my study period not listed, then
-      // the details of previously selected
-      // study period must be cleared.
       if (
         event.changed?.component.key === OFFERING_NOT_LISTED &&
         event.changed.value?.offeringnotListed
       ) {
+        // If the user after selecting a study period finds that
```
dheepak-aot, line 269: why was the comment moved?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -16,15 +19,17 @@ export function useFormioDropdownLoader() {
   const loadDropdown = async (
     form: any,
     dropdownName: string,
-    loadMethod: Promise<OptionItemAPIOutDTO[]>,
+    loadMethod: Promise<OptionItemAPIOutDTO[]> | ProgramIntensityAPIOutDTO[],
```
dheepak-aot, line 22: Why do we need a new type here?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -38,6 +43,44 @@ export function useFormioDropdownLoader() {
     );
   };
 
+  /**
+   * Populate program intensity dropdown.
+   * @param
```
dheepak-aot, line 48: Please complete the comment.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
dheepak-aot, line 22: I would prefer defining `id: number | string` instead of creating new object

sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -5139,6 +5139,24 @@
           "breadcrumb": "default",
           "id": "epd92wk"
         },
+        {
+          "input": true,
+          "tableView": true,
```
dheepak-aot, line 5144: Is table view required to be selected??

sources/packages/forms/src/form-definitions/sfaa2023-24.json
```diff
@@ -1534,16 +1535,7 @@
               "multiple": false,
               "dataSrc": "values",
               "data": {
-                "values": [
```
dheepak-aot, line 1537: Other forms doesn't have this change. Please make sure all 3 SFAA* forms have same content except title, name and path

dheepak-aot:
Nice work with the toggle @sh16011993 . Please have a look at the comments

sources/packages/web/src/components/common/StudentApplication.vue
sh16011993: RHS says `howWillYouBeAttendingTheProgram` and that is the reason I named it as `PROGRAM_INTENSITY_DROPDOWN_KEY`. It is the same variable impacting both program and offering intensities. Let me know if you still want me to change it.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
sh16011993: Isn't it that when the offering is not selected in the application, the `formService.dryRunSubmission` will fail resulting in the `BadRequestException`. Please let me know if I missed something.

devops/Makefile
```diff
@@ -393,6 +395,7 @@ deploy-api:
 		-p ATBC_LOGIN_ENDPOINT=$(ATBC_LOGIN_ENDPOINT) \
 		-p ATBC_ENDPOINT=$(ATBC_ENDPOINT) \
 		-p BYPASS_APPLICATION_SUBMIT_VALIDATIONS=$(BYPASS_APPLICATION_SUBMIT_VALIDATIONS) \
+		-p IS_FULLTIME_ALLOWED=${IS_FULLTIME_ALLOWED}
```
andrewsignori-aot, line 398: Is there a reason this one is using a different sintax?
Should it not be `-p IS_FULLTIME_ALLOWED=$(IS_FULLTIME_ALLOWED)`?

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -159,7 +162,8 @@ export class ApplicationStudentsController extends BaseController {
       "Selected offering id is invalid or " +
       "invalid study dates or selected study start date is not within the program year" +
       "or APPLICATION_NOT_VALID or INVALID_OPERATION_IN_THE_CURRENT_STATUS or ASSESSMENT_INVALID_OPERATION_IN_THE_CURRENT_STATE " +
-      "or INSTITUTION_LOCATION_NOT_VALID or OFFERING_NOT_VALID.",
+      "or INSTITUTION_LOCATION_NOT_VALID or OFFERING_NOT_VALID." +
```
andrewsignori-aot, line 165: Please remove the period at the end.
Just a point for the future because it did not start in this PR, the idea of the comments is to be more friendly than just listing the error codes. They can potentially be consumed by an audience external to this Team where the API descriptions would be helpful.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: @sh16011993 what @dheepak-aot is saying is what we talked about in the call between the 3 of us.
Either the dryRun must ensure that only Part-time intensity would be selected (which) or we need to add a code here to check it. I believe that the decision was to add the code check.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: is the offering intensity not being validated at this point expecting that the submit endpoint will validate it?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
andrewsignori-aot: @dheepak-aot I see the point but this is supposed to be a temporary solution that would be easier to be removed in this way.
I do not have a strong feeling in one way or another but I also did not get what it would not be recommended to use the config service inside a service.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
andrewsignori-aot: I agree with @dheepak-aot, please avoid adding the `ProgramIntensityAPIOutDTO`. The `OptionItemAPIOutDTO` was created to be generic and the objects must be adapted to it because `OptionItemAPIOutDTO` has what is needed to populate a dropdown.

sources/packages/backend/apps/api/src/route-controllers/config/_tests_/e2e/config.controller.getConfig.e2e-spec.ts
```diff
@@ -51,6 +52,7 @@ describe("ConfigController(e2e)-getConfig", () => {
           },
           externalSiteMinderLogoutUrl: fakeEnvVariables.SITE_MINDER_LOGOUT_URL,
         },
+        isFulltimeAllowed: fakeEnvVariables.IS_FULLTIME_ALLOWED === "true",
```
andrewsignori-aot, line 55: I am not following the reason why the `=== "true"` why required here. Can you please elaborate on that?

sources/packages/forms/src/form-definitions/sfaa2021-22.json
```diff
@@ -1574,7 +1565,7 @@
               "validateOn": "change",
               "validate": {
                 "required": true,
-                "onlyAvailableItems": true,
+                "onlyAvailableItems": false,
```
andrewsignori-aot, line 1568: Having this as false demands that the API validate the values otherwise the user can input any value here.
Please ensure that the API is validating `Full Time`/`Part Time` based on the token configuration.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -38,6 +43,46 @@ export function useFormioDropdownLoader() {
...
+    dropdownName: string,
+  ) => {
+    return loadDropdown(
+      form,
+      dropdownName,
+      getProgramIntensityDetails(isFulltimeAllowed),
+    );
+  };
+
+  const getProgramIntensityDetails = (isFulltimeAllowed: boolean) => {
```
andrewsignori-aot, line 64: Method comments missing. In this case, unless reused, I do not see a need to create a separated method.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -38,6 +43,46 @@ export function useFormioDropdownLoader() {
...
+        {
+          id: "Full Time",
+          description: "Full Time",
+        },
+        {
+          id: "Part Time",
+          description: "Part Time",
+        },
+      ];
+    } else {
```
andrewsignori-aot, line 76: The `else` is not needed here.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
sh16011993: Changing it as per the discussion.

sources/packages/web/src/components/common/StudentApplication.vue
sh16011993: Changing it as per the discussion today.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
sh16011993: I have changed it (for now) considering the explanation from @dheepak-aot that if it is passed as a parameter, then the service becomes more configurable allowing the same parameter to be passed from a different service with a different value.

sources/packages/backend/apps/api/src/route-controllers/config/_tests_/e2e/config.controller.getConfig.e2e-spec.ts
sh16011993: `IS_FULLTIME_ALLOWED: false` is a string in `process.env`. But when retrieved from the `/config` api call, it is a boolean. Therefore the comparison.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
sh16011993: `loadDropdown` requires the 3rd parameter to be a function. That is why created a separate function.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: This is how the concatenated string will look like in Swager:
"or INSTITUTION_LOCATION_NOT_VALID or OFFERING_NOT_VALIDor Invalid program or offering intensity"
Please add a space and a period as below

```ts
 "or INSTITUTION_LOCATION_NOT_VALID or OFFERING_NOT_VALID " +
 "or invalid program or offering intensity.",
```

sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -142,6 +142,17 @@ export class ConfigService {
     );
   }
 
+  /**
+   * Fulltime allowed Configuration.
+   * @returns fulltime allowed true if fulltime is allowed, otherwise false.
+   */
+  get isFulltimeAllowed(): boolean {
```
andrepestana-aot, line 149: 👍

sources/packages/forms/src/form-definitions/sfaa2021-22.json
andrewsignori-aot: Was this verification added?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -16,15 +16,17 @@ export function useFormioDropdownLoader() {
   const loadDropdown = async (
     form: any,
     dropdownName: string,
-    loadMethod: Promise<OptionItemAPIOutDTO[]>,
+    loadMethod: Promise<OptionItemAPIOutDTO[]> | OptionItemAPIOutDTO[],
```
andrepestana-aot, line 19: I think we don't need this change.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -38,6 +40,52 @@ export function useFormioDropdownLoader() {
...
+   * Populate program intensity dropdown.
+   * @param form form.
+   * @param isFulltimeAllowed is fulltime allowed.
+   * @param dropdownName dropdown name.
+   */
+  const loadProgramIntensityDetails = async (
+    form: any,
+    isFulltimeAllowed: boolean,
+    dropdownName: string,
+  ) => {
```
andrepestana-aot, line 53: Please add the return type.

andrepestana-aot (commented):
Good job. I left some minor comments. Please check other comments from other devs.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
sh16011993: typescript error shows up without it.

sources/packages/forms/src/form-definitions/sfaa2021-22.json
sh16011993: Added it now. Apologies, missed it earlier.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -171,22 +176,44 @@ export class ApplicationStudentsController extends BaseController {
...
+    const isFulltimeAllowed = this.configService.isFulltimeAllowed;
     const programYear = await this.programYearService.getActiveProgramYear(
       payload.programYearId,
     );
+    if (
+      ![OfferingIntensity.fullTime, OfferingIntensity.partTime].includes(
+        payload.data.howWillYouBeAttendingTheProgram,
+      )
+    ) {
+      throw new UnprocessableEntityException(
```
andrewsignori-aot, line 188: Can we change this to a `BadRequestException`. IMO it should be the same that the Nestjs validator would throw.
Besides that, can you please add the below comment.
"The check to validate the values for howWillYouBeAttendingTheProgram can be removed once the toggle for IS_FULL_TIME_ALLOWED is no longer needed and the types are hard-coded again in the form.io definition using the onlyAvailableItems as true.".

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -171,22 +176,44 @@ export class ApplicationStudentsController extends BaseController {
...
+    }
     if (!programYear) {
       throw new UnprocessableEntityException(
         "Program Year is not active. Not able to create an application invalid request.",
       );
     }
-
+    if (
+      !isFulltimeAllowed &&
+      payload.data.howWillYouBeAttendingTheProgram ===
```
andrewsignori-aot, line 199: Please keep in mind that while dealing with `payload.data` it can be cast to the `ApplicationData` to avoid the `any` along this method.

andrewsignori-aot (approved):
Great wok. I left two minor comments but they are minor, hence approving it.

dheepak-aot (approved):
Thanks for making the changes @sh16011993 . 👍

.github/workflows/release-deploy-all.yml
guru-aot, line 75: 👍

guru-aot (approved):
LGTM, nice work @sh16011993

lewischen-aot (approved):
Thanks for making the changes. Great work 👍

sources/packages/web/src/composables/useFormioDropdownLoader.ts
andrepestana-aot: OK. I see the issue. Thanks for checking that.

andrepestana-aot (approved):
Looks good.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
sh16011993: Changed it and added the comment.

Issue #2544, opened 2023-11-29
Analysis and troubleshooting on patroni going to recovery mode on load test


dheepak-aot:
**Describe the task**
During the process of camunda load testing #2489, when the load test was performed with 5000 assessments with 20 VUs(virtual users) in DEV environment, it has been consistently observed for every load test (until workers were adjusted to reduce the amount of parallel requests) that the database was temporarily going to recovery mode for few seconds and then coming back up and stable after that. 

Error observed in workers :

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDhvQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d5169e67c9d3d56843909c8389740a6d1dd3b255/image.png)

### How to replicate the issue:

To replicate the issue in DEV environment, run the load test with 5000 iterations and 20 VUs. As the worker configurations are updated to stabilize the load on DB now, before running the load test, update the min replicas for workers to 4 or a higher value.

To install K6 and run load test, follow the wiki: https://github.com/bcgov/SIMS/wiki/Load-testing 

### Points which came from discussion when this error was observed

There was a point which came up in a discussion regarding the type of user that we use to connect database from all the applications(API, Workers, Queue Consumers) that we use superuser instead of a regular user. Need to investigate if that is a possible root cause. 

**Acceptance Criteria**
- [x] Perform analysis to find the root cause and solution of the error where database goes to recovery mode temporarily
- [x] Timebox the effort. Implementation of solution if time permits otherwise have a separate ticket for that. 

**Additional context**
- Add any other context about the task here.
- Or here

guru-aot:
https://cloud.ibm.com/docs/databases-for-postgresql?topic=databases-for-postgresql-managing-connections#postgres-connection-limits

as part of the connection limits to the superuser it is retricted to 15 and the database users 100. The change from the superuser to database user in the load test gives the below results

<img width="1477" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3p5QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e9db9e3eca36e1e0a52f46a5a6f683ba21e6aa02/image.png" alt="image.png" />

[...]

PR #2730, opened 2024-01-24
#2544 - Patroni Recovery mode resolution

guru-aot:
Analysis and troubleshooting on patroni going to recovery mode on load test
- [x] Patroni convert the user from superuser to database user

Tested the simple workflow in our application after changing the database user to non super user with the minimal permissions and it seems to work well.

Also the release notes will be updated to grant the non super user permissions to do application operations.

GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA sims TO "app_database_user";

### DIFF
--- a/devops/openshift/api-deploy.yml
+++ b/devops/openshift/api-deploy.yml
@@ -290,9 +290,9 @@ parameters:
   - name: DB_SECRET_NAME
     value: patroni-creds
   - name: DB_USERNAME_KEY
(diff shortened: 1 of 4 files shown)


dheepak-aot (approved):
Great job. Looks good 👍

sh16011993 (approved):
LGTM @guru-aot 👍 . Nice work.

Issue #2664, opened 2024-01-10
Execute assessment calculations sequentially


michesmith:
## Calculate assessments sequentially

**Purpose**: Assessments of applications submitted by students of same PY must be _**calculated**_ sequentially to determine how much the student already consumed(got assessed) for an award in a given PY.

**Chronology of the sequence**: order by study start date (assessment with oldest study start date gets calculated first)

The process of calculating assessments sequentially means that, everything before the point of calculation will still happen in parallel in the same way as of now (e.g. Exceptions, PIR, Student income verification, Supporting user income verification(s)), but when workflow reaches the point of calculation, we will be having a new `service task`(and worker)  which will check **if this assessment is the first by the order of sequence** to be processed?
 if Yes, the calculations happen for this assessment and until the assessment calculation and workflow is complete, other assessments will be waiting at the same point in `assessment-gateway` workflow.

<img width="833" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN084QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b412dfcd6596882054a783b2a7404b2dcc588aec/image.png" alt="image.png" />

<img width="833" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN204QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9d242b00874da7dc1bfdfef26aded0df9f8d76ec/image.png" alt="image.png" />

**Acceptance Criteria**
- [x] Create a service task and worker in `assessment-gateway` to validate the sequence of processing
- [x] When the processing assessment is not the first in the sequence, then wait until a message is received or use timer to verify again
- [x] Make sure that the sequence of processing assessments must not conflict with the processing of queue order and makes a deadlock. i.e sqeuence in `assessment-gateway` waiting for an assessment with older study start date which is waiting to be queued.
- [x] Enchance the workflow E2E Tests with added mocks for the service task and the worker and fix any that breaks

PR #2745, opened 2024-01-29
#2664 - Execute assessment calculations in sequence

dheepak-aot:
## Execute assessment calculations in sequence

 - [x] Created orchestration in camunda workflow to **_process everything until calculation in parallel_** but starting **_from the calculation step, process assessments sequentially for the same student within same PY_**.

image

### Steps involved in sequential assessment calculation
- [x] STEP1: The first step in verification of an assessment to check if it is `ready for calculation` is to see if there is any ongoing calculation happening for an assessment of same student for same PY. If any ongoing calculation is happening, then without any further verification, the processing assessment must wait until the ongoing calculation is complete.

This state of a workflow is identified by a new column in `sims.student_assessments` table called `calculation_start_date` which indicates that calculation has been started for the assessment in the following date/timestamp.

- [x] STEP2: In STEP 1, if there is no assessment found in calculation step, then check for student assessments in the order of original assessment study start date to find if the **_processing assessment is the first outstanding assessment for calculation_**.

Here is the sql to get all the assessments for a student in a given PY ordered by original assessment study start date
```sql
select sa.id, (select epo.study_start_date from sims.student_assessments sas inner join sims.education_programs_offerings epo on sas.offering_id = epo.id 

where  sas.application_id = sa.application_id and sas.trigger_type  = 'Original assessment')  as original_assessment_date 

from sims.student_assessments sa inner join sims.applications a on sa.application_id = a.id

where sa.student_assessment_status not in ('Completed', 'Cancellation requested', 'Cancellation queued', 'Cancelled')

and sa.offering_id is not null

and a.student_id= :student_id

and a.program_year_id = :program_year_id

order by original_assessment_date, sa.created_at; 
```

Return a flag  `isReadyForCalculation` (true|false) based on if the processing assessment is the first outstanding assessment for calculation for the given student in given PY. 
- [x] STEP 3: If the processing assessment is identified to be ready for calculation `isReadyForCalculation = true` then set the calculation_start_date

- [x] STEP 4: For the assessment(s) which is/are waiting another assessment to complete are handled by `event based gateway` which can be triggered by either `message` or `time duration`. Hence when the assessment processing is complete, a message is sent with correlation id `$assessmentId` to unblock/re-trigger sequential verification for assessments which belong to same student in same PY. 

image

image

###  Technical outputs

Flow when the processing assessment is the first outstanding assessment during its verification

image

Flow when the processing assessment is not the first outstanding assessment during its verification and got processed later.

image

## E2E Tests

Will try to add at least one E2E test in an upcoming PR if possible

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/sql/StudentAssessments/Add-calculation-start-date.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/StudentAssessments/Add-calculation-start-date.sql
@@ -1,0 +1,6 @@
+ALTER TABLE
+    sims.student_assessments
+ADD
+    COLUMN calculation_start_date TIMESTAMP WITH TIME ZONE;
+
+COMMENT ON COLUMN sims.student_assessments.calculation_start_date IS 'Assessment calculation start date.';
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -16,6 +16,7 @@ import {
   SupportingUserJobOutDTO,
   UpdateNOAStatusHeaderDTO,
   UpdateNOAStatusJobInDTO,
+  VerifyAssessmentCalculationOrderJobOutDTO,
 } from "..";
 import { StudentAssessmentService } from "../../services";
 import {
@@ -225,6 +226,43 @@ export class AssessmentController {
       );
       jobLogger.log("Updated assessment status and saved the workflow data.");
       return job.complete();
+    } catch (error: unknown) {
+      return createUnexpectedJobFail(error, job, {
+        logger: jobLogger,
+      });
+    }
+  }
+
+  /**
+   * Verify as per the order of original assessment study start date validate if this assessment
+   * is the first in the sequence to be calculated
+   * for the given student inside the given program year.
+   */
+  @ZeebeWorker(Workers.VerifyAssessmentCalculationOrder, {
+    fetchVariable: [ASSESSMENT_ID],
+    maxJobsToActivate: MaxJobsToActivate.Normal,
+  })
+  async verifyAssessmentCalculationOrder(
+    job: Readonly<
+      ZeebeJob<
+        AssessmentDataJobInDTO,
+        ICustomHeaders,
+        VerifyAssessmentCalculationOrderJobOutDTO
+      >
+    >,
+  ): Promise<MustReturnJobActionAcknowledgement> {
+    const jobLogger = new Logger(job.type);
+    try {
+      const assessment = await this.studentAssessmentService.getById(
+        job.variables.assessmentId,
+      );
+      if (!assessment) {
+        const message = "Assessment not found.";
+        jobLogger.error(message);
+        return job.error(ASSESSMENT_NOT_FOUND, message);
+      }
+      jobLogger.log("Assessment calculation sequence has been verified.");
+      return job.complete({ isReadyForCalculation: true });
     } catch (error: unknown) {
       return createUnexpectedJobFail(error, job, {
         logger: jobLogger,
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -40,8 +40,6 @@ import {
 import {
   ASSESSMENT_DATA,
   ASSESSMENT_ID,
-  PROGRAM_YEAR_ID,
-  STUDENT_ID,
   WORKFLOW_DATA,
 } from "@sims/services/workflow/variables/assessment-gateway";
 import { CustomNamedError } from "@sims/utilities";
@@ -214,7 +212,7 @@ export class AssessmentController {
    * Worker to be executed at very end of the workflow responsible for latest tasks before the `end event`.
    */
   @ZeebeWorker(Workers.WorkflowWrapUp, {
-    fetchVariable: [ASSESSMENT_ID, STUDENT_ID, PROGRAM_YEAR_ID, WORKFLOW_DATA],
+    fetchVariable: [ASSESSMENT_ID, WORKFLOW_DATA],
     maxJobsToActivate: MaxJobsToActivate.Normal,
   })
   async workflowWrapUp(
@@ -224,14 +222,25 @@ export class AssessmentController {
   ): Promise<MustReturnJobActionAcknowledgement> {
     const jobLogger = new Logger(job.type);
     try {
+      const assessment =
+        await this.studentAssessmentService.getAssessmentSummary(
+          job.variables.assessmentId,
+        );
+      if (!assessment) {
+        const message = "Assessment not found.";
+        jobLogger.error(message);
+        return job.error(ASSESSMENT_NOT_FOUND, message);
+      }
       await this.studentAssessmentService.updateAssessmentStatusAndSaveWorkflowData(
         job.variables.assessmentId,
         job.variables.workflowData,
       );
+      const studentId = assessment.application.student.id;
+      const programYearId = assessment.application.programYear.id;
       // Send message to unblock the assessments which are waiting for this assessment if any.
       await this.workflowClientService.sendAssessmentCalculationCompleteMessage(
-        job.variables.studentId,
-        job.variables.programYearId,
+        studentId,
+        programYearId,
       );
       jobLogger.log("Updated assessment status and saved the workflow data.");
       return job.complete();
@@ -265,9 +274,10 @@ export class AssessmentController {
   ): Promise<MustReturnJobActionAcknowledgement> {
     const jobLogger = new Logger(job.type);
     try {
-      const assessment = await this.studentAssessmentService.getById(
-        job.variables.assessmentId,
-      );
+      const assessment =
+        await this.studentAssessmentService.getAssessmentSummary(
+          job.variables.assessmentId,
+        );
       if (!assessment) {
         const message = "Assessment not found.";
         jobLogger.error(message);
@@ -305,14 +315,14 @@ export class AssessmentController {
         `There is no ongoing calculation happening while verifying the order for processing assessment id ${assessmentId}.`,
       );
       // Get the first outstanding assessment waiting for calculation as per the sequence.
-      const [firstOutstandingStudentAssessment] =
+      const firstOutstandingStudentAssessment =
         await this.studentAssessmentService.getOutstandingAssessmentsForStudentInSequence(
           studentId,
           programYearId,
         );
       jobLogger.log(
         `The first outstanding assessment is identified to be assessment id ${firstOutstandingStudentAssessment.id} ` +
-          `with original assessment start date: ${firstOutstandingStudentAssessment.originalAssessmentStudyStartDate} ` +
+          `with original assessment start date  ${firstOutstandingStudentAssessment.originalAssessmentStudyStartDate} ` +
           `while verifying the order for processing assessment id ${assessmentId}.`,
       );
       // If the processing assessment is same as first outstanding assessment
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -273,13 +273,60 @@ export class AssessmentController {
         jobLogger.error(message);
         return job.error(ASSESSMENT_NOT_FOUND, message);
       }
-      const isReadyForCalculation =
-        await this.studentAssessmentService.verifyAssessmentCalculationOrder(
-          job.variables.assessmentId,
-          assessment.application.student.id,
-          assessment.application.programYear.id,
+      const assessmentId = job.variables.assessmentId;
+      const studentId = assessment.application.student.id;
+      const programYearId = assessment.application.programYear.id;
+      jobLogger.log(
+        `Verifying the assessment calculation order for assessment: ${assessmentId} student:${studentId} program year: ${programYearId}.`,
+      );
+      // Check for any assessment with ongoing calculation.
+      const assessmentInCalculationStep =
+        await this.studentAssessmentService.getAssessmentInCalculationStepForStudent(
+          studentId,
+          programYearId,
         );
-      jobLogger.log("Assessment calculation sequence has been verified.");
+
+      // If an ongoing calculation is happening all other assessments for given student in program year
+      // must wait until the calculation is complete and saved to the system.
+      // Also if a worker job is terminated by any chance after updating the calculation start date
+      // for an assessment, then the assessment must proceed for the calculation.
+      if (
+        assessmentInCalculationStep &&
+        assessmentInCalculationStep.id !== assessmentId
+      ) {
+        jobLogger.log(
+          `There is ongoing calculation happening for assessment: ${assessmentInCalculationStep.id} ` +
+            `and hence assessment : ${assessmentId} is waiting for that calculation to complete.`,
+        );
+        return job.complete({ isReadyForCalculation: false });
+      }
+      jobLogger.log(
+        `There is no ongoing calculation happening while verifying the order for assessment: ${assessment}.`,
+      );
+      // Get the first outstanding assessment waiting for calculation as per the sequence.
+      const [firstOutstandingStudentAssessment] =
+        await this.studentAssessmentService.getOutstandingAssessmentsForStudentInSequence(
+          studentId,
+          programYearId,
+        );
+      jobLogger.log(
+        `The first outstanding assessment is identified to be assessment: ${firstOutstandingStudentAssessment.id} ` +
+          `with original assessment start date: ${firstOutstandingStudentAssessment.originalAssessmentStudyStartDate} ` +
+          `while verifying the order for assessment: ${assessmentId}.`,
+      );
+      // If the processing assessment is same as first outstanding assessment
+      // then proceed for calculation.
+      const isReadyForCalculation =
+        firstOutstandingStudentAssessment.id === job.variables.assessmentId;
+      if (isReadyForCalculation) {
+        await this.studentAssessmentService.saveAssessmentCalculationStartDate(
+          assessmentId,
+        );
+      }
+      jobLogger.log(
+        `The assessment calculation order has been verified and ready for calculation status is ${isReadyForCalculation} ` +
+          `for assessment: ${assessmentId}.`,
+      );
       return job.complete({ isReadyForCalculation });
     } catch (error: unknown) {
       return createUnexpectedJobFail(error, job, {
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -249,7 +249,7 @@ export class AssessmentController {
             "waiting to be calculated next.",
         );
         // Send message to unblock the assessment which is waiting for calculation.
-        await this.workflowClientService.sendAssessmentCalculationCompleteMessage(
+        await this.workflowClientService.sendReleaseAssessmentCalculationMessage(
           nextOutstandingAssessmentInSequence.id,
         );
       }
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -246,6 +246,9 @@ export class AssessmentController {
    * Verify as per the order of original assessment study start date if this assessment
    * is the first in the sequence to be calculated
    * for the given student inside the given program year.
+   * If an assessment is identified to be the first in sequence to be calculated
+   * then set calculation start date to hold other assessments until the calculation
+   * is complete.
    */
   @ZeebeWorker(Workers.VerifyAssessmentCalculationOrder, {
     fetchVariable: [ASSESSMENT_ID],
@@ -270,14 +273,14 @@ export class AssessmentController {
         jobLogger.error(message);
         return job.error(ASSESSMENT_NOT_FOUND, message);
       }
-      const isFirstInCalculationSequence =
+      const isReadyForCalculation =
         await this.studentAssessmentService.verifyAssessmentCalculationOrder(
           job.variables.assessmentId,
           assessment.application.student.id,
           assessment.application.programYear.id,
         );
       jobLogger.log("Assessment calculation sequence has been verified.");
-      return job.complete({ isFirstInCalculationSequence });
+      return job.complete({ isReadyForCalculation });
     } catch (error: unknown) {
       return createUnexpectedJobFail(error, job, {
         logger: jobLogger,
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
@@ -277,7 +277,8 @@ export class AssessmentController {
       const studentId = assessment.application.student.id;
       const programYearId = assessment.application.programYear.id;
       jobLogger.log(
-        `Verifying the assessment calculation order for assessment: ${assessmentId} student:${studentId} program year: ${programYearId}.`,
+        `Verifying the assessment calculation order for processing assessment id ${assessmentId} student id ${studentId} ` +
+          `program year id ${programYearId}.`,
       );
       // Check for any assessment with ongoing calculation.
       const assessmentInCalculationStep =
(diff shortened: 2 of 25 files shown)


sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
```diff
@@ -76,358 +77,82 @@
...
-    </bpmn:startEvent>
-    <bpmn:sequenceFlow id="Flow_1btetiq" sourceRef="load-assessment-data-pre-assessment-subprocess" targetRef="Gateway_11qgy63" />
+    <bpmn:sequenceFlow id="Flow_1eip27h" sourceRef="verify-assessment-calculation-order-gateway" targetRef="assessment-calculation-completed-message" />
+    <bpmn:sequenceFlow id="Flow_0bza37i" sourceRef="verify-assessment-calculation-order-gateway" targetRef="Event_05itxic" />
+    <bpmn:sequenceFlow id="Flow_16zz7zb" name="Waiting for another assessment to complete" sourceRef="Gateway_16whgqa" targetRef="verify-assessment-calculation-order-gateway">
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=isReadyForCalculation = false</bpmn:conditionExpression>
+    </bpmn:sequenceFlow>
+    <bpmn:sequenceFlow id="Flow_0wp1s62" sourceRef="verify-assessment-calculation-order-task" targetRef="Gateway_16whgqa" />
+    <bpmn:sequenceFlow id="Flow_0358dxp" sourceRef="assessment-calculation-completed-message" targetRef="verify-assessment-calculation-order-task" />
+    <bpmn:sequenceFlow id="Flow_0umbeot" sourceRef="Event_05itxic" targetRef="verify-assessment-calculation-order-task" />
```
andrewsignori-aot, line 94: Can we also have the timer named with a meaningful name?

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
```diff
@@ -444,111 +169,238 @@
...
-    <bpmn:sequenceFlow id="Flow_0ozt4pg" name="Reassessment" sourceRef="first-assessment-gateway" targetRef="Gateway_11qgy63">
-      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=assessmentTriggerType != "Original assessment"</bpmn:conditionExpression>
-    </bpmn:sequenceFlow>
-    <bpmn:sequenceFlow id="Flow_1wrj7fh" sourceRef="associate-workflow-instance-task" targetRef="load-assessment-data-submit-or-reassessment-subprocess" />
-    <bpmn:serviceTask id="associate-workflow-instance-task" name="Associate workflow instance">
+    <bpmn:intermediateCatchEvent id="Event_05itxic">
+      <bpmn:incoming>Flow_0bza37i</bpmn:incoming>
+      <bpmn:outgoing>Flow_0umbeot</bpmn:outgoing>
+      <bpmn:timerEventDefinition id="TimerEventDefinition_0h74s3m">
+        <bpmn:timeDuration xsi:type="bpmn:tFormalExpression">PT30S</bpmn:timeDuration>
```
andrewsignori-aot, line 234: Since this is a fallback, can we have it set for a few minutes at least?

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.model.ts
```diff
@@ -0,0 +1,5 @@
+import { StudentAssessment } from "@sims/sims-db";
+
+export class StudentAssessmentDetail extends StudentAssessment {
```
andrewsignori-aot, line 3: A business comment would be recommended to explain that even with reassessments the original assessment is the one selected. If the business context was already added somewhere else please disregard this comment.

sources/packages/backend/libs/services/src/workflow/variables/assessment-gateway.ts
```diff
@@ -41,3 +41,12 @@ export const DISBURSEMENT_SCHEDULES = "disbursementSchedules";
  * is executed for easy application consumption.
  */
 export const WORKFLOW_DATA = "workflowData";
+
```
andrewsignori-aot, line 44: Please remove the extra blank line following the same from the other in the same file.

sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts
```diff
@@ -171,4 +171,30 @@ export class WorkflowClientService {
...
+  ): Promise<void> {
+    try {
+      await this.zeebeClient.publishMessage({
+        name: "assessment-calculation-completed",
+        correlationKey: `${studentId}-${programYearId}`,
+        timeToLive: ZEEBE_PUBLISH_MESSAGE_DEFAULT_TIME_TO_LEAVE,
+      });
+    } catch (error: unknown) {
+      this.logger.error(
+        `Error while sending assessment calculation completed message for student: ${studentId} and program year: ${programYearId}`,
```
andrewsignori-aot, line 194: Can we change the sentence as the below? I am not sure if a sentence accepts ":" twice.
"Error while sending assessment calculation completed message for student id ${studentId}, and program year id ${programYearId}."

sources/packages/backend/workflow/test/test-utils/constants/workflow-variables-constants.ts
```diff
@@ -25,6 +25,7 @@ export enum WorkflowServiceTasks {
   // Workflow - Supporting User Information Request
   CheckSupportingUserResponseTask = "check-supporting-user-response-task",
   WorkflowWrapUpTask = "workflow-wrap-up-task",
+  VerifyAssessmentCalculationOrderTask = "verify-assessment-calculation-order-task",
```
andrewsignori-aot, line 28: I believe that it should be moved right below line 21, at least.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -209,7 +214,7 @@ export class AssessmentController {
    * Worker to be executed at very end of the workflow responsible for latest tasks before the `end event`.
    */
   @ZeebeWorker(Workers.WorkflowWrapUp, {
-    fetchVariable: [ASSESSMENT_ID, WORKFLOW_DATA],
+    fetchVariable: [ASSESSMENT_ID, STUDENT_ID, PROGRAM_YEAR_ID, WORKFLOW_DATA],
```
andrewsignori-aot, line 217: Not a blocker but the workers, as much as possible so far, receive the minimum information from the Workflow.
Almost all of them are driven by `assessmentId` or `applicationId` and from that information, they will retrieve data like `studentId` and `programYearId`.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -232,6 +242,99 @@ export class AssessmentController {
...
+      if (!assessment) {
+        const message = "Assessment not found.";
+        jobLogger.error(message);
+        return job.error(ASSESSMENT_NOT_FOUND, message);
+      }
+      const assessmentId = job.variables.assessmentId;
+      const studentId = assessment.application.student.id;
+      const programYearId = assessment.application.programYear.id;
+      jobLogger.log(
+        `Verifying the assessment calculation order for assessment: ${assessmentId} student:${studentId} program year: ${programYearId}.`,
```
andrewsignori-aot, line 280: Can we change the below log as:
"Verifying the assessment calculation order for assessment id ${assessmentId}, student id${studentId}, and program year is ${programYearId}."

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -321,4 +332,108 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
   }
+
+  /**
+   * Get all outstanding student assessments to be calculated
+   * in the order of original assessment study start date.
+   * @param studentId student id.
+   * @param programYearId program year id.
+   * @returns student assessment to be calculated.
+   */
+  async getOutstandingAssessmentsForStudentInSequence(
```
andrewsignori-aot, line 343: Do we need to return more than one record at this time or is it in preparation to some future implementation?

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -232,6 +242,99 @@ export class AssessmentController {
...
+      jobLogger.log(
+        `There is no ongoing calculation happening while verifying the order for assessment: ${assessment}.`,
+      );
+      // Get the first outstanding assessment waiting for calculation as per the sequence.
+      const [firstOutstandingStudentAssessment] =
+        await this.studentAssessmentService.getOutstandingAssessmentsForStudentInSequence(
+          studentId,
+          programYearId,
+        );
+      jobLogger.log(
```
andrewsignori-aot, line 312: Same from https://github.com/bcgov/SIMS/pull/2745/files#r1470173211

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -232,6 +242,99 @@ export class AssessmentController {
     }
   }
 
+  /**
+   * Verify as per the order of original assessment study start date if this assessment
```
andrewsignori-aot, line 246: Maybe it would be worth to mentioned that assessments pending PIR would not be considered. Looking at the query and getting it from `assessment.offering IS NOT NULL` may be tricky.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -321,4 +332,108 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+      .setParameter("triggerType", AssessmentTriggerType.OriginalAssessment)
+      .setParameter("studentAssessmentStatus", [
+        StudentAssessmentStatus.Completed,
+        StudentAssessmentStatus.CancellationRequested,
+        StudentAssessmentStatus.CancellationQueued,
+        StudentAssessmentStatus.Cancelled,
+      ])
+      .setParameter("studentId", studentId)
+      .setParameter("programYearId", programYearId)
+      .orderBy(`"${originalAssessmentStudyStartDateAlias}"`)
```
andrewsignori-aot, line 381: Can it be simplified as `.orderBy(originalAssessmentStudyStartDateAlias)`?

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -321,4 +332,108 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+   * @returns student assessment to be calculated.
+   */
+  async getOutstandingAssessmentsForStudentInSequence(
+    studentId: number,
+    programYearId: number,
+  ): Promise<StudentAssessmentDetail[]> {
+    const originalAssessmentStudyStartDateAlias =
+      "originalAssessmentStudyStartDate";
+    // Sub query to get the original assessment study start date of a given assessment's application.
+    const originalAssessmentDateSubQuery = this.repo
```
andrewsignori-aot, line 350: Even knowing that from a business perspective, we should have a maximum of one record, would it be worth adding a `limit 1` to the query?

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -321,4 +332,108 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+      where: {
+        calculationStartDate: Not(IsNull()),
+        studentAssessmentStatus: Not(
+          In([
+            StudentAssessmentStatus.Completed,
+            StudentAssessmentStatus.CancellationRequested,
+            StudentAssessmentStatus.CancellationQueued,
+            StudentAssessmentStatus.Cancelled,
+          ]),
+        ),
```
andrewsignori-aot, line 414: Was there a benefit using the current way instead the below?
```ts
studentAssessmentStatus: In([
  StudentAssessmentStatus.Submitted,
  StudentAssessmentStatus.Queued,
  StudentAssessmentStatus.InProgress,
]),
```

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -232,6 +242,99 @@ export class AssessmentController {
...
+      // Also if a worker job is terminated by any chance after updating the calculation start date
+      // for an assessment, then the assessment must proceed for the calculation.
+      if (
+        assessmentInCalculationStep &&
+        assessmentInCalculationStep.id !== assessmentId
+      ) {
+        jobLogger.log(
+          `There is ongoing calculation happening for assessment: ${assessmentInCalculationStep.id} ` +
+            `and hence assessment : ${assessmentId} is waiting for that calculation to complete.`,
+        );
```
andrewsignori-aot, line 300: Can we change the log as below?
There is an ongoing calculation happening for the assessment id ${assessmentInCalculationStep.id} ` +
and hence assessment id ${assessmentId} is waiting for that calculation to complete.

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
andrewsignori-aot, line 1: Minor but can the second lane on the workflow have the element contained only inside it as the sample below?
image

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
```diff
@@ -444,111 +169,238 @@
...
+      </bpmn:timerEventDefinition>
+    </bpmn:intermediateCatchEvent>
+    <bpmn:eventBasedGateway id="verify-assessment-calculation-order-gateway">
+      <bpmn:incoming>Flow_16zz7zb</bpmn:incoming>
+      <bpmn:outgoing>Flow_0bza37i</bpmn:outgoing>
+      <bpmn:outgoing>Flow_1eip27h</bpmn:outgoing>
+    </bpmn:eventBasedGateway>
+    <bpmn:exclusiveGateway id="Gateway_16whgqa" name="Is the assessment first in sequence to be calculated">
```
andrewsignori-aot, line 242: Can we have it as a question following the existing one?
"Is the assessment first in sequence to be calculated?"

andrewsignori-aot, line 242: Can we have a meaningful name added to the gateway?

andrewsignori-aot (commented):
Amazing work, please take a look at the comments.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -232,6 +242,99 @@ export class AssessmentController {
...
+      if (isReadyForCalculation) {
+        await this.studentAssessmentService.saveAssessmentCalculationStartDate(
+          assessmentId,
+        );
+      }
+      jobLogger.log(
+        `The assessment calculation order has been verified and ready for calculation status is ${isReadyForCalculation} ` +
+          `for assessment: ${assessmentId}.`,
+      );
+      return job.complete({ isReadyForCalculation });
```
lewischen-aot, line 330: I'm just wondering if it is possible that `isReadyForCalculation` could be false and we aren't handling the case for `isReadyForCalculation = false` here.

sources/packages/backend/apps/db-migrations/src/sql/StudentAssessments/Add-calculation-start-date.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    sims.student_assessments
+ADD
+    COLUMN calculation_start_date TIMESTAMP WITH TIME ZONE;
```
andrewsignori-aot, line 4: Should we add some script to have the current records on DB updated?
For instance, copy the `assessment_date`, when present, to the new `calculation_start_date` or set now() when the status is already `in progress`?

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
dheepak-aot: I agree.

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
lewischen-aot, line 1: Correct me if I'm wrong. From Camunda doc for [event-based gateway](https://docs.camunda.io/docs/components/modeler/bpmn/event-based-gateways/), I thought there should be an exclusive gateway to close the two outgoing sequence flows for the event-based gateway. I'm not sure if the same applies to our case here. Please let me know your thoughts.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
dheepak-aot: We are good here for the condition `isReadyForCalculation = false` here @lewischen-aot . The action happening in line 322 is only meant for `isReadyForCalculation = true` and when it is false then variable must be returned as it is with job complete.

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
dheepak-aot: I appreciate your effort to go further and understand about camunda here.

I read the same documentation, and couldn't infer the same.

As per my understanding from the documentation, an event based gateway must have at least 2 outgoing flows which must connect to a catch event. The catch event leads to what is intended to be done based on the event.

image

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
lewischen-aot: Thanks @dheepak-aot This makes now.

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
lewischen-aot: Thanks @dheepak-aot. I thought we needed an exclusive gateway to catch one of the two events and stop the other one. If the service task `verify-assessment-calculation-order` does the following, then I think we're good.
image

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
guru-aot: For my understanding. This change we are planning to limit to 1 as for a given application there will be only one original assessment right?

sources/packages/backend/apps/db-migrations/src/sql/StudentAssessments/Add-calculation-start-date.sql
guru-aot: i believe this is only a problem for test and dev IMO, as @dheepak-aot said we can leave things as it is or clear the assessment data in dev and test after this implementation.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -219,10 +222,26 @@ export class AssessmentController {
   ): Promise<MustReturnJobActionAcknowledgement> {
     const jobLogger = new Logger(job.type);
     try {
+      const assessment =
+        await this.studentAssessmentService.getAssessmentSummary(
+          job.variables.assessmentId,
+        );
+      if (!assessment) {
```
guru-aot, line 229: i believe at this point the assessment cannot be empty and this check will be obsolete. To fetch the studentId and the programYearId i believe this call can happen after the updateAssessmentStatusAndSaveWorkflowData call.

dheepak-aot: This is to validate the assessment id @guru-aot . We do the same in other places .e.g `loadAssessmentConsolidatedData`

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
andrewsignori-aot: Please check the below log as it seems that there is at least one double white space in the middle of the sentence.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
andrewsignori-aot: It is limited to one because it needs only the first assessment as stated by the comment now: "Get first outstanding student assessment to be calculated".

sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts
```diff
@@ -171,4 +171,28 @@ export class WorkflowClientService {
...
+   * send message to unblock the next assessment which is waiting for this assessment
+   * in the sequence to be calculated.
+   * @param assessmentId assessment waiting for calculation.
+   */
+  async sendAssessmentCalculationCompleteMessage(
+    assessmentId: number,
+  ): Promise<void> {
+    try {
+      await this.zeebeClient.publishMessage({
+        name: "assessment-calculation-completed",
```
andrewsignori-aot, line 186: I am having a hard time figuring out if the name would be easier to understand. It may cause some confusion because the calculation being completed is actually from another assessment workflow.
Should we have it renamed to something that would indicate better that it is from outside the workflow itself? For instance: `related-assessment-calculation-completed`.

sources/packages/backend/workflow/BPMN/camunda-8/load-assessment-consolidated-data.bpmn
andrewsignori-aot, line 1: In the future we can avoid this files from being added to the PR 😉

sources/packages/backend/workflow/BPMN/camunda-8/assessment-gateway.bpmn
andrewsignori-aot: Was this updated?

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -232,6 +262,101 @@ export class AssessmentController {
...
+          programYearId,
+        );
+
+      // If an ongoing calculation is happening all other assessments for given student in program year
+      // must wait until the calculation is complete and saved to the system.
+      // Also if a worker job is terminated by any chance after updating the calculation start date
+      // for an assessment, then the assessment must proceed for the calculation.
+      if (
+        assessmentInCalculationStep &&
+        assessmentInCalculationStep.id !== assessmentId
```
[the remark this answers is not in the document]
andrepestana-aot: nevermind

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -321,4 +323,125 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+          StudentAssessmentStatus.CancellationRequested,
+          StudentAssessmentStatus.CancellationQueued,
+          StudentAssessmentStatus.Cancelled,
+        ])
+        .setParameter("studentId", studentId)
+        .setParameter("programYearId", programYearId)
+        .orderBy(`"${originalAssessmentStudyStartDateAlias}"`)
+        .addOrderBy("assessment.createdAt")
+        .limit(1)
+        .getRawOne<StudentAssessmentDetail>()
```
andrepestana-aot, line 379: Do we need to `limit(1)` if we user `getRawOne()`?

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -321,4 +323,125 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+  }
+
+  /**
+   * Set assessment calculation start date.
+   ** Save only when the calculation start date is not present
+   ** considering the need of workers to be idempotent.
+   * @param assessmentId assessment id.
+   * @returns update result.
+   */
+  async saveAssessmentCalculationStartDate(assessmentId: number) {
```
andrepestana-aot, line 416: return type: `: Promise<UpdateResult>`

andrepestana-aot (commented):
Great job on this PR. Left minor comments.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
dheepak-aot: Yes, mainly with query builder I'm not sure about the order of first result without limit 1. 

I printed the queries with and without limit 1 for getRawOne method and observed that

With limit 1 I can see 

image


And without limit one 


image


Hence without limit1 I am not sure what is the result returned and felt safe to use limit 1.

andrepestana-aot: Thanks for checking that.

andrepestana-aot (approved):
Looks good! Thanks for the explanation.

lewischen-aot (approved):
Thanks for making the changes. Great work @dheepak-aot 👍

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
guru-aot: Yes i get it but this place we will not have a scenario where we will not have an assessment id.

Issue #2451, opened 2023-10-23
Enable typescript reference on backend errors


andrewsignori-aot:
While logging error stack traces we must ensure the trace will have the reference to the typescript instead of the bundled one, as shown below.

**Additional context**
Please use references like the below one looking for the best way to have it enabled.
https://stackoverflow.com/questions/59000552/how-to-print-stack-trace-with-reference-to-typescript-source-in-nest-js

```
TypeError: Cannot read property 'length' of undefined
    at SpecializedStringBuilder.appendWithEndFiller (/opt/app-root/src/dist/apps/queue-consumers/main.js:711:15)
    at SpecializedStringBuilder.appendStringWithFiller (/opt/app-root/src/dist/apps/queue-consumers/main.js:7456:14)
    at IER12FileDetail.getFixedFormat (/opt/app-root/src/dist/apps/queue-consumers/main.js:14115:16)
    at /opt/app-root/src/dist/apps/queue-consumers/main.js:12160:66
    at Array.map (<anonymous>)
    at IER12IntegrationService.uploadContent (/opt/app-root/src/dist/apps/queue-consumers/main.js:12160:47)
    at IER12ProcessingService.uploadIER12Content (/opt/app-root/src/dist/apps/queue-consumers/main.js:14357:48)
    at IER12ProcessingService.processIER12File (/opt/app-root/src/dist/apps/queue-consumers/main.js:14342:52)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
TypeError: Cannot read property 'length' of undefined
    at SpecializedStringBuilder.appendWithEndFiller (/opt/app-root/src/dist/apps/queue-consumers/main.js:711:15)
    at SpecializedStringBuilder.appendStringWithFiller (/opt/app-root/src/dist/apps/queue-consumers/main.js:7456:14)
    at IER12FileDetail.getFixedFormat (/opt/app-root/src/dist/apps/queue-consumers/main.js:14115:16)
    at /opt/app-root/src/dist/apps/queue-consumers/main.js:12160:66
    at Array.map (<anonymous>)
```

**Acceptance criteria**
- [x] Check if this has already be done (Signori may have already updated the config)
- [x] Ensure that error stack traces will contain the typescript file reference.
- [x] Review throw errors with inner exceptions to ensure that the stack trace from nested expectations will be added to logs also.

andrepestana-aot:
It is already referencing to the API source code files:
[...]

PR #2753, opened 2024-01-29
#2451 - Enable typescript reference on backend errors

andrepestana-aot:
- Added flag `--enable-source-maps` to node apps;
- Added `devtool: "source-map"` to webpack.config.js. Followed this comment about monorepo: https://github.com/nestjs/nest/issues/4644#issuecomment-616992927
- Fixed some errors being suppressed found;
- Created a ticket to enforce the error throwing to be an instance of `Error` in TS: https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2773
- Created a ticket to update NestJS in order to allow passing `cause` to HTTP Exceptions: https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2774
- Expected logs in containers now:
image
image

### Notes:
- If an error is thrown as string, the line number is not shown:
image
image
That may require to change TS rules [in a new ticket](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2773) to enforce thrown errors to be an instance of `Error`: https://typescript-eslint.io/rules/no-throw-literal/
- The current @nestjs/common version used doesn't have support for passing `cause` in the HTTP Exception:
    - Current used version: https://github.com/nestjs/nest/blob/v9.0.3/packages/common/exceptions/http.exception.ts
    - Latest version: https://github.com/nestjs/nest/blob/master/packages/common/exceptions/http.exception.ts
    It requires to update NestJS [in another ticket](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2774) to allow passing `cause` to HTTP exceptions.


sources/packages/backend/package.json
```diff
@@ -13,14 +13,14 @@
...
-    "start:prod:load-test-gateway": "node dist/apps/load-test-gateway/main.js",
-    "docker:start:load-test-gateway": "node dist/apps/load-test-gateway/main.js",
+    "start:prod:api": "node --enable-source-maps dist/apps/api/main.js",
+    "docker:start:api": "node --enable-source-maps dist/apps/api/main.js",
+    "start:prod:workers": "node --enable-source-maps dist/apps/workers/main.js",
+    "docker:start:workers": "node --enable-source-maps dist/apps/workers/main.js",
+    "start:prod:queue-consumers": "node --enable-source-maps dist/apps/queue-consumers/main.js",
+    "docker:start:queue-consumers": "node --enable-source-maps dist/apps/queue-consumers/main.js",
+    "start:prod:load-test-gateway": "node --enable-source-maps dist/apps/load-test-gateway/main.js",
+    "docker:start:load-test-gateway": "node --enable-source-maps dist/apps/load-test-gateway/main.js",
```
sh16011993, line 23: FMU, don't we need to add `source-map-support` as a dependency?

andrepestana-aot: NodeJS has already support for that from version 12.12: https://nodejs.org/en/blog/release/v12.12.0/

guru-aot:
This means the cause will not be shown until we upgrade the version and after that do we need to have any change other than the required one?

sources/packages/backend/webpack.config.js
```diff
@@ -1,5 +1,6 @@
 const CopyPlugin = require("copy-webpack-plugin");
 module.exports = {
+  devtool: "source-map",
```
guru-aot, line 3: 👍

andrewsignori-aot (approved):
Nice work, looks good 👍 
Can we add an entry in the wiki about ensuring that an error throw will always contain the cause. This should also be part of the PR reviews.

guru-aot (approved):

lewischen-aot (approved):
LGTM 👍 Nice work!

dheepak-aot (approved):
Looks Good 👍

sh16011993 (approved):
Nice work @andrepestana-aot 👍

andrepestana-aot:
> This means the cause will not be shown until we upgrade the version and after that do we need to have any change other than the required one?

@guru-aot The NestJS version is just to allow us to pass the cause to HTTP Exceptions when it is appropriate. Then, after the update, we can check if it would be good to add a cause to them. All the other cases are already working when we pass cause to the `Error` instantiation like in https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/libs/integrations/src/esdc-integration/e-cert-integration/e-cert-part-time-integration/e-cert-files/e-cert-file-record.ts#L225

andrepestana-aot:
> Nice work, looks good 👍 Can we add an entry in the wiki about ensuring that an error throw will always contain the cause. This should also be part of the PR reviews.

Sure. Will do that.

PR #2755, opened 2024-01-30
#2524 - Hide FT programs for PT release: Disable FT Schedulers

sh16011993:
**As a part of this PR, FT schedulers are disabled/enabled based on toggle off/on for `IS_FULLTIME_ALLOWED` environment variable.**

Note: Currently, `queue.obliterate()` is used to remove the fulltime queues and all jobs associated with them (including any active jobs). Other options: `empty()` and `clean()`.

Good Read: https://blog.taskforce.sh/obliterate/

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
@@ -55,9 +55,17 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
     if (
       isFulltimeAllowed ||
       (!isFulltimeAllowed &&
-        !this.schedulerQueue.name.startsWith("full-time", 0))
+        ![
+          QueueNames.FullTimeMSFAAIntegration,
+          QueueNames.FullTimeECertIntegration,
+          QueueNames.FullTimeFeedbackIntegration,
+          QueueNames.FullTimeDisbursementReceiptsFileIntegration,
+          QueueNames.FullTimeMSFAAProcessResponseIntegration,
+        ].includes(this.schedulerQueue.name as QueueNames))
     ) {
-      await this.schedulerQueue.add(await this.payload());
+      const payload = await this.payload();
+      console.log(this.schedulerQueue.name, " Payload: ", payload);
+      await this.schedulerQueue.add(payload);
     }
   }
 
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
@@ -48,7 +48,8 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
-    // TODO: ConfigService below is not injected since it will be removed when the Fulltime is released.
+    // TODO: Partime logic to be removed when the fulltime is released. ConfigService below is not injected
+    // TODO: since adding it here (base class) causes it to be added in all the multiple subclasses extending from it.
     const isFulltimeAllowed = new ConfigService().isFulltimeAllowed;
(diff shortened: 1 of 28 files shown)


sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -47,9 +49,23 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    const isFulltimeAllowed = this.configService.isFulltimeAllowed;
     await this.deleteOldRepeatableJobs();
     // Add the cron to the queue.
     await this.schedulerQueue.add(await this.payload());
+    // Allow Fulltime Schedulers only if isFulltimeAllowed is true
```
dheepak-aot, line 56: `deleteOldRepeatableJobs` and `await this.schedulerQueue.add(await this.payload());` will not be required if the queue needs to be obliterate

image

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -47,9 +49,23 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
...
+    if (
+      !isFulltimeAllowed &&
+      [
+        QueueNames.FullTimeMSFAAIntegration,
```
dheepak-aot, line 60: IER and ECE schedulers are also pertinent to full time. I would say we can add them to the list instead of them running dummy schedules not picking any data.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -47,9 +49,23 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    const isFulltimeAllowed = this.configService.isFulltimeAllowed;
```
dheepak-aot, line 52: As this is a temporary solution until we go full time, we can construct the config service like this which will avoid the changes in all other scheduler processor files. Revert would be easy.

image

dheepak-aot:
Great research and work. Please have a look at the comments.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
andrewsignori-aot: I would be ok if we add a clear comment that this is not inject with the intention of being removed once full-time is release.

andrewsignori-aot (commented):
Nice work. I just added some comments to the comments already added. Please take a look.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -47,6 +48,25 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    // TODO: ConfigService below is not injected since it will be removed when the Fulltime is released.
```
dheepak-aot, line 51: The TODO must be towards removing the fulltime allowed logic later. 
And must additionally explain the reason of not injecting the config service.

dheepak-aot (approved):
Thanks for making the changes. Looks Good. 👍

lewischen-aot:
Nice work. As someone new to the scheduler system, I can see that the latest solution having `isFulltimeAllowed` in `base-scheduler.ts` file should be the acceptable approach for this PR.

lewischen-aot (approved):
Great work @sh16011993 👍

andrewsignori-aot (approved):

guru-aot (approved):
LGTM, nice work @sh16011993

andrepestana-aot (approved):
Great job!

Issue #1989, opened 2023-06-02
Github action workflow to purge the images generated from main branch


guru-aot:
DevOps tasks to create an automated Github action workflow to purge the images generated from the main branch, this is required as during cluster upgrade, the images are purged and not available. This causes failed deployments in the namespaces.

- [x] Review how the images are purged by BC gov (please see https://stackoverflow.developer.gov.bc.ca/questions/1104). In the past, some of our images apparently were pruned by BC Gov in a way that some PODs entered an unrecoverable "Crash loop" because the image associated with the deployment config (DC) was no longer present.

PR #2797, opened 2024-02-06
#1989 -  Prune images

sslaws:
Created a new script to determine ImageStream tags that can be safely removed for a given environment.  The intended usage would be to run the script against prod and set min_tags to a number like 3 for rollback.  This would list all tags newer than prod + include 3 tags that are older than prod.

After speaking with Guru, I added the ability to prefix a branch which then only targets those tags.

The script itself only dumps a list which then must be used by a command to do something.  For this, I've created a new GitHub Action which calls the fetchOldTags.sh against dev with a min_tags to set at 10.  This is the behaviour that has been run manually.  Ideally in the future we could target Prod but this requires timely releases.  

NOTE:  Currently the script simply echo's the OC commands that should be run, once merged I will test and resubmit.

### DIFF
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -1,5 +1,5 @@
 name: Prune Images
-description: Prune old images from OpenShift
+run-name: Prune old images from OpenShift
 
 on:
   workflow_dispatch:
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -1,0 +1,63 @@
+name: Prune Images
+description: Prune old images from OpenShift
+
+on:
+  workflow_dispatch:
+    inputs:
+      applications:
+        description: 'Applications to prune'
+        required: true
+        default: 'web-sims, api-sims'
+      licensePlate:
+        description: 'License Plate'
+        required: true
+        default: '0c27fb'
+      environment:
+        description: 'Environment (dev, test, prod)'
+        required: true
+        default: 'dev'
+      prefix:
+        description: 'Prefix'
+        required: false
+        default: 'main'
+      minTags:
+        description: 'Minimum number of Tags to keep'
+        required: false
+        default: '10'
+  schedule:
+    - cron: '0 6 * * *' # Runs at 6 AM every day
+
+jobs:
+  prune-images:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Print env
+        run: |
+          echo "License Plate: ${{ inputs.licensePlate }}"
+          echo "Environment: ${{ inputs.environment }}"
+          echo "Prefix: ${{ inputs.prefix }}"
+          echo "Minimum Tags: ${{ inputs.minTags }}"
+      
+      - name: Checkout source code
+        uses: actions/checkout@v3
+      
+      - name: Login to OpenShift
+        run: |
+          oc login --token=${{ secrets.SA_TOKEN }} --server=${{ vars.OPENSHIFT_CLUSTER_URL }}
+      
+      - name: Prune Images
+        run: |
+          IFS=',' read -r -a apps <<< ${{ inputs.applications }}
+          for app in "${apps[@]}"
+          do
+            # Trim spaces
+            trimmed_app=$(echo "$app" | xargs)
+            ./fetchOldTags.sh \
+              --license_plate=${{ inputs.licensePlate }} \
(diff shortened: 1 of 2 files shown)


.github/workflows/prune-images.yml
```diff
@@ -0,0 +1,63 @@
...
+run-name: Prune old images from OpenShift
+
+on:
+  workflow_dispatch:
+    inputs:
+      applications:
+        description: 'Applications to prune'
+        required: true
+        default: 'web-sims, api-sims'
+      licensePlate:
```
andrewsignori-aot, line 11: For every Github environment, we have available a variable that represents the Openshift namespace. This is consumed in other GitHub Actions. Can the `licensePlate` be replaced by this one instead, which means that it does not need to be provided as an input parameter and it would be "connected" to the select env.
```
NAMESPACE: ${{ secrets.OPENSHIFT_ENV_NAMESPACE }}
```

.github/workflows/prune-images.yml
```diff
@@ -0,0 +1,63 @@
+name: Prune Images
+run-name: Prune old images from OpenShift
```
andrewsignori-aot, line 2: It would be recommended add some inputs to the `run-name` to differentiate them easily.
For instance:
```
Release - Deploy all from ${{ inputs.gitRef }} to ${{ inputs.environment }}
```

.github/workflows/prune-images.yml
```diff
@@ -0,0 +1,63 @@
...
+    inputs:
+      applications:
+        description: 'Applications to prune'
+        required: true
+        default: 'web-sims, api-sims'
+      licensePlate:
+        description: 'License Plate'
+        required: true
+        default: '0c27fb'
+      environment:
```
andrewsignori-aot, line 15: For clarification, the images are not environment-specific, what would be the intention to have the environment as an input parameter?
Maybe I am just missing how the script will work and I can hold those questions for the final PR 😉

andrewsignori-aot (commented):
Nice work and thanks for making this move forward.
I am ok to approve these changes if the indentation is to have workflow merged to start running some tests directly.
Please also take a look in the below short section with some PR recommendations 😉 
https://github.com/bcgov/SIMS/wiki/Development---Standards,-Decisions,-and-Troubleshooting#pr-review

.github/workflows/prune-images.yml
dheepak-aot: From my understanding on going through the script, the env is to get the DC deployed image, to get the list of images older than the DC deployed image for a given env.

.github/workflows/prune-images.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/checkout@v3
+      
+      - name: Login to OpenShift
+        run: |
+          oc login --token=${{ secrets.SA_TOKEN }} --server=${{ vars.OPENSHIFT_CLUSTER_URL }}
+      
+      - name: Prune Images
+        run: |
+          IFS=',' read -r -a apps <<< ${{ inputs.applications }}
+          for app in "${apps[@]}"
```
dheepak-aot, line 51: 👍

devops/scripts/fetchOldTags.sh
```diff
@@ -0,0 +1,84 @@
...
+    # All good, continue
+    ;;
+  *)
+    echo "The --env parameter must be one of: dev, test, prod."
+    exit 1
+    ;;
+esac
+
+# Setup internal variables
+DC_NAMESPACE="${LICENSE_PLATE}-${ENV}"
```
dheepak-aot, line 57: Getting the namespace this way would be tricky because, our test and staging environment uses same name space.

dheepak-aot (approved):
Great start @sslaws 👍Can do a more deeper review on the final PR.

andrepestana-aot (approved):
Looks good for my limited knowledge :)

.github/workflows/prune-images.yml
andrewsignori-aot: @dheepak-aot images are not environment-specific so we will not prune images per environment. We may use some env as PROD to figure out which ones can be removed, but it would not fits in the idea of prune per env.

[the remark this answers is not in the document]
guru-aot: I believe for this scenario, we are going to clear the data in 0c27fb-tools environment only as we build the Imagestreams in only one.

Issue #2553, opened 2023-12-04
PROD Datafix required - Ineligible Program Message on Eligible Program


CarlyCotton:
**Describe the Bug**
In Production, when viewing already existing (eligible) programs there is a warning message that the program is ineligible for StudentAid BC funding.
<img width="740" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOU10QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d7c012e7303b8e89cf84bc4794b9ea17996ee7bf/image.png" alt="image.png" />

**Expected Behaviour**
In programs that are eligible for funding, there should not be a warning message. 

**Actual Behaviour**
A data change may have affected the messaging in the program view to show the warning message about being ineligble. They are still **correctly** showing as Approved:
<img width="835" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWt0QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7748f8c81ed4f4fbe6745cc7da2a50f9eb850cdb/image.png" alt="image.png" />

** Steps To Reproduce**
Not reproducible in TEST - these are existing programs in PROD.

Programs in PROD prior to late July/August seem to be affected when they have a specific entrance requirement selected:
-  This program has entrance requirements established by the institution that enable completion of the program study.

Some of the programs affected:
BCIT : Boilermaker Foundation (submitted June 29, 2023)
BCIT : Auto Body Repair and Refinishing Technician Foundation (submitted July 4, 2023)
UVIC : Bachelor of Arts (submitted June 28, 2023)
UVIC : Bachelor of Science (submitted June 28, 2023)
Langara : Bachelor of Science in Nursing (Terms 1, 4 and 7)

**Acceptance Criteria**
- [ ] Not possible to reproduce in dev and test.  please consider the prod data for the investigation 
- [ ] Programs that are eligible in SIMS don't show warning message.
[...]

PR #2808, opened 2024-02-07
#2553 - Ineligible Program Message on Eligible Program

andrepestana-aot:
Form.io's show variable evaluates `null` values as `true`:
image

Could verify the error happening in DEV and in my local environment:
image

Fixed issue doing: 
- Changes to form:
```javascript
show = !!data.entranceRequirements.noneOfTheAboveEntranceRequirements;
```


sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Rollback-col-none-of-entrance-requirements.sql
sh16011993, line 1: Shouldn't the filename be `Rollback-update-col-none-of-entrance-requirements.sql`?

sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Rollback-col-none-of-entrance-requirements.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE
+    sims.education_programs
+ALTER COLUMN
+    none_of_entrance_requirements DROP NOT NULL;
+
+ALTER TABLE
+    sims.education_programs
+ALTER COLUMN
+    none_of_entrance_requirements DROP DEFAULT;
```
sh16011993, line 9: Shouldn't there be a rollback for the `Update` statement?

andrepestana-aot: What's the point of updating on the rollback?

sh16011993 (commented):
LGTM 👍 @andrepestana-aot. Left a few comments.

sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Update-col-none-of-entrance-requirements.sql
andrewsignori-aot, line 1: As a suggestion, the same can be executed using two instructions and be simplified as below.
```sql
ALTER TABLE
    sims.education_programs
ALTER COLUMN
    none_of_entrance_requirements
SET
    DEFAULT false,
ALTER COLUMN
    none_of_entrance_requirements
SET
    NOT NULL;
```

sources/packages/backend/apps/db-migrations/src/migrations/1707346761578-UpdateNoneOfEntranceRequirements.ts
```diff
@@ -0,0 +1,24 @@
...
+        "Update-col-none-of-entrance-requirements.sql",
+        "EducationPrograms",
+      ),
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData(
+        "Rollback-col-none-of-entrance-requirements.sql",
```
andrewsignori-aot, line 19: Minor but the idea was to use the Rollback as a prefix, as shown below.
```
Rollback-update-col-none-of-entrance-requirements.sql
```

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
```diff
@@ -259,9 +259,9 @@ export class EducationProgram extends RecordDataModel {
    */
   @Column({
     name: "none_of_entrance_requirements",
-    nullable: true,
+    nullable: false,
```
andrewsignori-aot, line 262: Are the other columns from the same set of options also setting nullable `false`?

andrewsignori-aot (commented):
Great investigation and throubleshooting, please take a look at the comments.

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
andrepestana-aot: No, they are optional and nullable: 
image

Should we update them too and make them mandatory?

andrewsignori-aot: I believe that we should go one way or another. Either have all of them changed or keep the PR simple and do not change the `none_of_entrance_requirements`, unless it would result in an issue. Would it be an issue not making the column mandatory right now?

I assume that it was an optional field in the past and then it became mandatory.
If we make them all mandatory we just need how complex would be change.
I guess that the EntranceRequirements models are already expecting a not null value.
```ts
export class EntranceRequirements {
  @Expose()
  hasMinimumAge: boolean;
  @Expose()
  minHighSchool: boolean;
  @Expose()
  requirementsByInstitution: boolean;
  @Expose()
  requirementsByBCITA: boolean;
  @Expose()
  noneOfTheAboveEntranceRequirements: boolean;
}
```

sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Rollback-col-none-of-entrance-requirements.sql
sh16011993: It's ok. We can leave the data values rollback.

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
andrepestana-aot: We don't need to make `none_of_entrance_requirements` NOT NULL. The formio change will be enough to fix the bug. I'm going to revert the changes to DB.

andrewsignori-aot (approved):
Great finding, nice work, looks good 👍

lewischen-aot (approved):
Looks good. Thanks for making the changes 👍

guru-aot (approved):

sh16011993 (approved):
Nice work and investigation @andrepestana-aot  👍

dheepak-aot:
Looks Good 👍

PR #2821, opened 2024-02-13
#1989 - Prune images

sslaws:
## Overview:

These ideal method for this script would be that it would be called against the Production project and without a branch specifier and would remove all but N older images from the associated tools imagestreams.  This is not currently the practice as the deployments don't occur frequently enough to prevent a build up of old tags.

As such, the script runs against the dev environment and removing tags that are older than N versions in dev and only tagged against main.

In an even more ideal world we would use environment tags like: dev, test, staging, production which with a minor modification would allow the script to only look at the image stream tags and not have to review the deployment configs.  

## Scripts Overview

1. pruneImages.sh \
    ⚠️ This script is **destructive** and iterates over a comma seperated list of applications calling `fetchOldTags.sh` and then using that output to generate OpenShift CLI commands to remove the images.\
    Logic moved from the GH Action to make it reusable \
    \
    Example call
    ```console
    ./pruneImages.sh --license_plate=0c27fb --env=dev --apps="web-sims, api-sims, queue-consumers-sims, workers-sims" --prefix=main --min_tags=10
    ```
1. fetchOldTags.sh \
    For a given application, the script will look in the defined environment and determine what tag is being used by the OpenShift DeploymentConfig.  It then will generate a list of all tags older than that version and exclude the most recent versions up to `--min_tags`.  If a prefix is supplied those tags will be excluded from the list.

    Example call
    ```console
    ./fetchOldTags.sh --license_plate=0c27fb --app_name=api-sims --env=dev --prefix=main --min_tags=10
    ```

Questions and responses from original PR #2797 

1. Add some inputs to the run-name\
DONE

1. Why don't I use  `${{ secrets.OPENSHIFT_ENV_NAMESPACE }}`\
These appear to be fully qualified and this script needs to look at two projects the tools where the images are stored and then deployment environment.

1. For the environment input we use the specific Github type for it...\
The GitHub environment includes those that are not specific to OpenShift and are also not normalized so would cause issues

1. What is the purpose of the environment input \
It is used to determine what tag is currently deployed to that environment and only remove tags older than that

1. Getting the namespace based on `${LICENSE_PLATE}-${ENV}` would not work for staging
That is true but given the typical use case for the script should only be run against dev or prod.

### DIFF
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -7,7 +7,7 @@ on:
       applications:
         description: 'Comma seperated list of applications to prune'
         required: true
-        default: 'web-sims, api-sims'
+        default: 'web-sims, api-sims, queue-consumers-sims, workers-sims'
       licensePlate:
         description: 'OpenShift License Plate'
         required: true
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -52,7 +52,7 @@ jobs:
           ./pruneImages.sh \
               --license_plate=${{ inputs.licensePlate }} \
               --env=${{ inputs.environment }} \
-              --apps='"${{ inputs.applications }}"' \
+              --apps="${{ inputs.applications }}" \
               --prefix=${{ inputs.prefix }} \
               --min_tags=${{ inputs.minTags}}
           popd
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -4,6 +4,15 @@ run-name: Pruning all but the most recent ${{ inputs.minTags }} tags prior to wh
 on:
   workflow_dispatch:
     inputs:
+      environment:
+        type: choice
+        description: 'Environment'
+        required: true
+        options:
+          - dev
+          - test
+          - prod
+        default: 'test'
       applications:
         description: 'Comma seperated list of applications to prune'
         required: true
@@ -12,10 +21,6 @@ on:
         description: 'OpenShift License Plate'
         required: true
         default: '0c27fb'
-      environment:
-        description: 'Environment (dev, test, prod)'
-        required: true
-        default: 'dev'
       prefix:
         description: 'Branch prefix to restrict pruning to'
         required: false
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -1,15 +1,15 @@
 name: Prune Images
-run-name: Prune old images from OpenShift
+run-name: Pruning all but the most recent ${{ inputs.minTags }} tags prior to what is deployed in ${{ inputs.environment }}
 
 on:
   workflow_dispatch:
     inputs:
       applications:
-        description: 'Applications to prune'
+        description: 'Comma seperated list of applications to prune'
         required: true
         default: 'web-sims, api-sims'
       licensePlate:
-        description: 'License Plate'
+        description: 'OpenShift License Plate'
         required: true
         default: '0c27fb'
       environment:
@@ -17,7 +17,7 @@ on:
         required: true
         default: 'dev'
       prefix:
-        description: 'Prefix'
+        description: 'Branch prefix to restrict pruning to'
         required: false
         default: 'main'
       minTags:
@@ -35,6 +35,7 @@ jobs:
         run: |
           echo "License Plate: ${{ inputs.licensePlate }}"
           echo "Environment: ${{ inputs.environment }}"
+          echo "Applications: ${{ inputs.applications }}"
           echo "Prefix: ${{ inputs.prefix }}"
           echo "Minimum Tags: ${{ inputs.minTags }}"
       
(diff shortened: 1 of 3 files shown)


.github/workflows/prune-images.yml
```diff
@@ -1,23 +1,23 @@
...
 
 on:
   workflow_dispatch:
     inputs:
       applications:
-        description: 'Applications to prune'
+        description: 'Comma seperated list of applications to prune'
         required: true
-        default: 'web-sims, api-sims'
+        default: 'web-sims, api-sims, queue-consumers-sims, workers-sims'
```
guru-aot, line 10: can we have an input called as "all" to prune in a single go? rather selecting each and every time?

devops/scripts/fetchOldTags.sh
```diff
@@ -37,8 +37,8 @@ done
 
 # Validation of required arguments and env values
 if [ -z "$LICENSE_PLATE" ] || [ -z "$ENV" ] || [ -z "$APP_NAME" ]; then
-  echo "License plate, environment, and app name are required."
   echo "Usage: $0 --license_plate=VALUE --env=VALUE --app_name=VALUE [--prefix=VALUE] [--min_tags=VALUE]"
+  echo "License plate, environment, and app name are required."
```
guru-aot, line 41: In the same file below can we please convert the MIN_TAGS to number like
from 
`
  | if $MIN_TAGS > 0 then .[:-$MIN_TAGS] else . end
  | .[]
  `
To 
`
| if $MIN_TAGS > 0 then .[:-($MIN_TAGS | tonumber)] else . end
  | .[]
`

devops/scripts/pruneImages.sh
```diff
@@ -0,0 +1,63 @@
...
+do
+    # Trim spaces
+    trimmed_app=$(echo "$app" | xargs)
+        echo "Processing \"${trimmed_app}\""
+        ./fetchOldTags.sh \
+            --license_plate=${LICENSE_PLATE} \
+            --env=${ENV} \
+            --app_name=${trimmed_app} \
+            --prefix=${PREFIX} | \
+            xargs -I {} oc tag ${LICENSE_PLATE}-tools/${trimmed_app}:{} --delete
```
guru-aot, line 59: Just a suggestion, instead of using the xargs, we can use a  while read loop as an output of the fetchOldTags.sh
`
while IFS= read -r tag; do
        # Delete using the command.
        oc tag "${LICENSE_PLATE}-tools/${trimmed_app}:${tag}" --delete
    done
`

.github/workflows/prune-images.yml
sslaws: This is a single input to prune everything but you have to specify the list.  The reason for the list vs ALL is that each environment is unique.

devops/scripts/fetchOldTags.sh
sslaws: The explicit conversion is not required as jq can handle the value and our run identifiers are always numeric.  Great question!

.github/workflows/prune-images.yml
```diff
@@ -1,23 +1,28 @@
...
 on:
   workflow_dispatch:
     inputs:
+      environment:
+        type: choice
+        description: 'Environment'
+        required: true
+        options:
+          - dev
+          - test
```
dheepak-aot, line 13: As per our discussion, we are using the reference of build deployed to test. Do we still need this environment input?

.github/workflows/prune-images.yml
```diff
@@ -1,23 +1,28 @@
...
+        default: 'test'
       applications:
-        description: 'Applications to prune'
+        description: 'Comma seperated list of applications to prune'
         required: true
-        default: 'web-sims, api-sims'
+        default: 'web-sims, api-sims, queue-consumers-sims, workers-sims'
       licensePlate:
-        description: 'License Plate'
+        description: 'OpenShift License Plate'
```
dheepak-aot, line 21: As mentioned in previous PR, we have this value from existing env value right?

dheepak-aot:
I just glanced through the PR and did review based on the discussion we had in sync up. Great job. Please have a look at the comments.

guru-aot (approved):
LGTM, nice work @sslaws

.github/workflows/prune-images.yml
sslaws: The default is test but we have flexibility by keeping this which the underlying scripts can use.

.github/workflows/prune-images.yml
sslaws: I don't see just the license plate but maybe one of the other guys can confirm for us.

andrewsignori-aot: We do not have only the license plate but it can be added at the same level that we have the `SA_TOKEN`, for instance.
I've just created it `OPENSHIFT_LICENSE_PLATE` with the value `0c27fb` and it can be used as 
```
${{ vars.OPENSHIFT_CLUSTER_URL }}
```
Does it work @sslaws?

.github/workflows/prune-images.yml
andrewsignori-aot: @dheepak-aot I believe that it would be ok keeping it as it is considering that the cleanup would happen only for the "main-*" builds for now.

andrewsignori-aot (approved):
Thanks for the work and for the clarifications. Please let me know if you will be consuming the new `var` related to the below comment, otherwise I will delete it 😉 
https://github.com/bcgov/SIMS/pull/2821/files#r1490137567

.github/workflows/prune-images.yml
sslaws: Yeah we could do that but is it worth the effort?  I don't really look at license plates as secrets or special configs but if you want it replaced then I can do that.

lewischen-aot (approved):
Thanks for making the changes and the excellent PR description.
I learned quite a few things from this PR 👍

.github/workflows/prune-images.yml
andrewsignori-aot: From my perspective @sslaws I have already approved the PR because I did not see this as a blocker. I created the env variable and I just need to know if I should keep or removed it 😉

sslaws: No worries, I've removed the input and referenced the GH var instead.

andrepestana-aot (approved):
Looks good!

PR #2842, opened 2024-02-16
#1989 - prune images

sslaws:
Fixes the time the action is triggered at 6am Pacific is 2pm UTC
Maps the inputs or default values into GH Environment variables
Quotes were changed

### DIFF
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -6,55 +6,60 @@ on:
     inputs:
       environment:
         type: choice
-        description: 'Environment'
+        description: "Environment"
         required: true
         options:
           - dev
           - test
           - prod
-        default: 'test'
+        default: "test"
       applications:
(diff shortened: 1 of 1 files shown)


.github/workflows/prune-images.yml
```diff
@@ -6,55 +6,60 @@ on:
...
+
       - name: Print env
         run: |
-          echo "Environment: ${{ inputs.environment }}"
-          echo "Applications: ${{ inputs.applications }}"
-          echo "Prefix: ${{ inputs.prefix }}"
-          echo "Minimum Tags: ${{ inputs.minTags }}"
-      
+          echo "Environment: ${ENVIRONMENT}"
+          echo "Applications: ${APPLICATIONS}"
```
dheepak-aot, line 45: Asking for my understanding, how does it work without using input.* ?

dheepak-aot (approved):
Looks good @sslaws 

Thanks for the explanations.

PR #2843, opened 2024-02-16
Create VS Code Workspace and Editor Config files

sslaws:
Creates an editor config which sets defaults for the entire project (lf, 2 space tab, etc)
Created a multi-root workspace with extension recommendations, default search excludes, format on save etc.

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+charset = utf-8
--- a/sims.code-workspace
+++ b/sims.code-workspace
@@ -1,0 +1,93 @@
+{
+  "folders": [
+    {
+      "name": "Web UI",
+      "path": "./sources/packages/web"
+    },
+    {
+      "name": "Backend",
+      "path": "./sources/packages/backend"
+    },
+    {
+      "name": "Forms",
+      "path": "./sources/packages/forms"
+    },
+    {
+      "name": "DevOps",
+      "path": "./devops"
+    },
+    {
+      "name": "Load Tests",
+      "path": "./sources/packages/load-test"
+    },
+    {
+      "name": "All",
+      "path": "."
+    }
+  ],
+  "extensions": {
+    "recommendations": [
+      "vue.volar",
+      "vue.vscode-typescript-vue-plugin",
+      "dbaeumer.vscode-eslint",
+      "esbenp.prettier-vscode",
+      "streetsidesoftware.code-spell-checker",
+      "aaron-bond.better-comments",
+      "ms-vscode.js-debug-nightly",
+      "adpyke.vscode-sql-formatter",
+      "redhat.vscode-yaml"
+    ]
+  },
+  "tasks": {
+    "version": "2.0.0",
+    "tasks": [
+      {
+        "label": "View Outdated Web Packages",
+        "type": "shell",
+        "command": "npm outdated",
+        "options": {
+          "cwd": "${workspaceFolder:Web UI}"
+        },
+        "problemMatcher": []
+      },
+      {
+        "label": "View Outdated Backend Packages",
+        "type": "shell",
+        "command": "npm outdated",
+        "options": {
+          "cwd": "${workspaceFolder:Backend}"
+        },
+        "problemMatcher": []
+      },
+      {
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+charset = utf-8
```
guru-aot, line 9: for files upload test cases we use ASCII encoding, else the test cases uploading files will be converted to UTF-8 and will cause test case failures.

sslaws: ASCII is a subset of UTF-8 so I don't think it would matter!  VS Code also defaults to UTF-8 so its likely that this wouldn't impact anyone but it does ensure that if you have customized it that while editing files in the SIMS root it will be UTF-8.

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
```
sh16011993, line 3: For my understanding, can you please briefly tell what this would do. Thanks!

sh16011993: Please correct me if I am wrong, does it mean that for all the file formats apply the below settings?

sims.code-workspace
sh16011993, line 1: I did not get the idea as to why we need a multi-root workspace. Can you please briefly tell that. Thanks!

.editorconfig
sslaws: root means that we won't continue up the path looking for more .editorconfg files and [*] means that these rules apply to all files.

sims.code-workspace
sslaws: The purpose of the multi-root workspace is just to give us flexibility going forward but also organize the structure logically.

.editorconfig
sh16011993: Thanks @sslaws

sims.code-workspace
sh16011993: Sounds good. Thanks @sslaws

.editorconfig
andrepestana-aot, line 1: Do I need to use this plugin? https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

.editorconfig
andrewsignori-aot: @guru-aot the settings are following what we have already using and they will not affect the existing files.
For instance, with the settings in place, opening a ts file and saving it would not cause any difference.
The ASCII files will not be affected.
@sslaws, yes, the ASCII would not matter. The problem that @guru-aot was trying to make was about the extended ASCII (8bits) that has issues while converting to UTF-8, but either way would not be a problem and we were not even sure at this moment that we need to support extended ASCII.

andrewsignori-aot (approved):
Thanks for having these configurations unified, looks good 👍

sims.code-workspace
andrepestana-aot, line 1: What should I do to enable that view on my VSCode?
image

andrewsignori-aot: Double-click can work, if not the below should work 😉 
image

.editorconfig
andrewsignori-aot: I do not have it installed but it was working out-of-box.
The configs are the same that we have been using. To test we can change configs like tab to 4 and try to save a file.

lewischen-aot (approved):
LGTM 👍 Thanks for adding the VS Code workspace @sslaws

dheepak-aot (approved):
LGTM 👍

sims.code-workspace
andrepestana-aot: hm, I like it! Thanks!

andrepestana-aot (approved):
Looks good!

sims.code-workspace
```diff
@@ -0,0 +1,93 @@
+{
+  "folders": [
```
andrepestana-aot, line 2: As a suggestion we could have an entry for `sources` because that is the folder where we should be to use the Makefile.

sims.code-workspace
```diff
@@ -0,0 +1,93 @@
...
+      "redhat.vscode-yaml"
+    ]
+  },
+  "tasks": {
+    "version": "2.0.0",
+    "tasks": [
+      {
+        "label": "View Outdated Web Packages",
+        "type": "shell",
+        "command": "npm outdated",
```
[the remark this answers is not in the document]
sslaws: This is mostly an example but this is what I typically run to update the packages that are directly referenced.

Issue #2685, opened 2024-01-15
Import Provincial Restrictions from SFAS Data Bridge into SIMS Account (Part 1) 


HRAGANBC:
**Description**
Ensure that any restrictions imported through the SFAS data bridge are not only logged in SIMS database but **also** translates into those restrictions being applied as 'active' within the student's account. When a restriction code is imported but not recognized by SIMS, convert it to a generic restriction with a warning. 

**Note about the Business Process**
Restrictions will continue to be imported daily from legacy systems even if they've been resolved within SIMS. I.e. SST has to manually resolve them in BOTH systems to prevent them re-appearing through the SFAS data bridge. 

<img width="989" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeVgzQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b09891292f6a9c1e7cc2972af6c23a309db7889e/image.png" alt="image.png" />

**Acceptance Criteria**
- [x] Import active ("restrictionRemovalDate"= null) provincial restrictions from SFAS_restrictionstable, based on "Restriction Code", updating the student account restrictions. Ignore restrictions with "restrictionRemovalDate". 
- [x] Create new provincial restriction type called "Legacy" with code LGCY. Categorized as 'other' within the Ministry interface. 
- [x] Only one active LGCY restriction at a time 
- [x] Check for necessary mapping from the Excel spreadsheet attached below. 
- [x] If a code is imported from SFAS bridge but **not** explicitly mapped to a SIMS restriction, then convert it to a LGCY and apply that restriction in SIMS.
- [x] LGCY restriction should have the following effects:
- Stop part-time disbursement; stop full-time disbursement
- Add banner message with standard restriction 'error' message on student dashboard
- [x] Apply provincial restrictions imported from SFAS onto the student account, according to the mapping done in the attached excel. (see additional context below)

**Additional Context**
- Restrictions List Mapping found here: [SIMS Restrictions List.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1ZEQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e17df147bc512ce164fc8555fec096637ad98128/SIMS%20Restrictions%20List.xlsx)

[...]

PR #2848, opened 2024-02-17
#2685 - Import Provincial Restrictions from SFAS Data Bridge into SIMS

sh16011993:

- [x] **Created migration scripts to:**
        **1.** Add the `is_included` column to the `sfas_restrictions` table. 
        **2.** Add the `Legacy` restriction to the sims `restrictions` table.

- [x] **Created 3 separate sql queries to achieve the following:**
    
**1.** After all the SFAS restrictions are imported,  this query inserts restrictions from `sfas_restrictions` to  `student_restrictions` if those restrictions are valid restrictions in the sims system and are either not present in the sims `student_restrictions` table or they are inactive.
```
INSERT INTO
  sims.student_restrictions (student_id, restriction_id, creator)
SELECT
  filtered_students.student_id,
  $1,
  $2
FROM
  (
    SELECT
      DISTINCT(sfas_individuals.student_id) AS student_id
    FROM
      (
        SELECT
          sfas_restrictions.code,
          CASE
            WHEN sfas_restrictions.code = 'A12' THEN '12'
            WHEN sfas_restrictions.code = 'AF1' THEN 'AF'
            WHEN sfas_restrictions.code = 'B2D' THEN 'B2'
            ELSE sfas_restrictions.code
          END AS mapped_code,
          sfas_restrictions.individual_id,
          sfas_restrictions.processed
        FROM
          sims.sfas_restrictions sfas_restrictions
        WHERE
          sfas_restrictions.removal_date IS NULL
          AND sfas_restrictions.processed = FALSE
      ) mapped_restrictions
      INNER JOIN sims.sfas_individuals sfas_individuals ON mapped_restrictions.individual_id = sfas_individuals.id
      LEFT JOIN sims.restrictions restrictions ON mapped_restrictions.mapped_code = restrictions.restriction_code
    WHERE
      sfas_individuals.student_id IS NOT NULL
      AND restrictions.restriction_code IS NULL
  ) filtered_students
WHERE
  NOT EXISTS (
    SELECT
      1
    FROM
      sims.student_restrictions student_restrictions
    WHERE
      student_restrictions.student_id = filtered_students.student_id
      AND student_restrictions.restriction_id = $1
      AND student_restrictions.is_active = TRUE
  );
```

**2.** After all the SFAS individuals are imported, this query inserts one Legacy restriction entry per student into  `student_restrictions` if they have one or more restrictions imported from the sfas system that are not present in the sims restrictions.
```
INSERT INTO
  sims.student_restrictions (student_id, restriction_id)
SELECT
  (sfas_individuals.student_id),
  restrictions.id AS restriction_id
FROM
  (
    SELECT
      sfas_restrictions.code,
      CASE
        WHEN sfas_restrictions.code = 'A12' THEN '12'
        WHEN sfas_restrictions.code = 'AF1' THEN 'AF'
        WHEN sfas_restrictions.code = 'B2D' THEN 'B2'
        ELSE sfas_restrictions.code
      END AS mapped_code,
      sfas_restrictions.individual_id,
      sfas_restrictions.processed
    FROM
      sims.sfas_restrictions sfas_restrictions
    WHERE
      sfas_restrictions.removal_date IS NULL
      AND sfas_restrictions.processed = false
  ) mapped_restrictions
  INNER JOIN sims.sfas_individuals sfas_individuals ON mapped_restrictions.individual_id = sfas_individuals.id
  INNER JOIN sims.restrictions restrictions ON mapped_restrictions.mapped_code = restrictions.restriction_code
  LEFT JOIN sims.student_restrictions student_restrictions ON student_restrictions.student_id = sfas_individuals.student_id
  AND student_restrictions.restriction_id = restrictions.id
WHERE
  sfas_individuals.student_id IS NOT NULL
  AND (
    student_restrictions.is_active = false
    OR student_restrictions.restriction_id IS NULL
  );
```

**3.** Once all the restrictions have been added to the `student_restrictions` table, mark the `is_included` column to `true` for all the newly imported restrictions.

```
UPDATE
  sims.sfas_restrictions
SET
  processed = true
  and updated_at = now()
WHERE
  processed = false;
```

**Note:**  The mappings for the restriction codes from SFAS to SIMS have been handled in the above sql queries.

**Also, please note that the issue pointed out by Sonarcloud is a false positive as shown below.**

image


sources/packages/backend/apps/db-migrations/src/sql/Restrictions/Rollback-insert-lgcy-restriction.sql
```diff
@@ -0,0 +1,5 @@
+-- Rollback insert Legacy Restriction.
+DELETE FROM
+  sims.restrictions r
```
andrewsignori-aot, line 3: Usually, we do not provide aliases for these queries.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  sims.sfas_restrictions
+ADD
+  COLUMN is_included bool DEFAULT FALSE;
```
andrewsignori-aot, line 4: I am struggling to decide the meaningfulness of the name `is_included`.
I would suggest something like `processed` but I prefer to hear from the other DEVs. If the Team agrees that it is aligned with the purpose of the data I would be ok.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
```diff
@@ -0,0 +1,6 @@
...
+
+COMMENT ON COLUMN sfas_restrictions.is_included IS 'This field indicates whether the sfas restriction has been added to the student restrictions table or not.';
```
andrewsignori-aot, line 6: Minor, SFAS as an acronym in the description can be in upper case.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
andrepestana-aot: Yes, `is_included` is definitely a bad name because doesn't give a clue about the "inclusion". I would use `imported_from_sfas` or `imported_from_integration` to avoid the SFAS name.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
```diff
@@ -17,8 +19,20 @@ export class SFASRestrictionService
...
+    this.bulkInsertStudentRestrictionsSQL = getSQLFileData(
+      "Bulk-insert-restrictions.sql",
+      SFAS_RESTRICTIONS_RAW_SQL_FOLDER,
+    );
+  }
+
+  /**
+   * Bulk operation to insert student restrictions from SFAS restrictions data.
+   */
+  async updateStudentRestrictions(): Promise<void> {
```
andrepestana-aot, line 34: Minor but the method name is "update..." but the operation is to insert records.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -141,10 +141,15 @@ export class SFASIntegrationProcessingService {
 
       if (result.success) {
         try {
+          /**
+           * Bulk operation to update student id in SFAS individuals table after importing data from SFAS.
+           */
+          await this.sfasIndividualService.updateStudentId();
           await this.sfasIndividualService.updateDisbursementOverawards();
+          await this.sfasRestrictionService.updateStudentRestrictions();
```
andrewsignori-aot, line 149: The operation related `updateStudentId`, `updateDisbursementOverawards`, and `updateStudentRestrictions` can have their errors individually logged.
For instance, the below message is trying to comunicating 2 out of 3 possible issues (e.g. `updateStudentId` is not included).
"Error while updating overawards balances or inserting student restrictions imported from SFAS." 

Besides that, I believe that we should not be deleting the file if one of those operations failed. All imported data from SFAS should be able to safely retry. If I am not wrong we can execute the import of the same file over and over again with no harm.

For instance, the comment `Delete the file only if it was processed with success.` states that the file should be deleted if everything went well and I believe that we broke this rule when we did the overawards (unless I am missing something).

In summary, I believe that we should delete the file only when `updateStudentId`, `updateDisbursementOverawards`, and `updateStudentRestrictions` are successfully executed. In case one of them fail the file should be fixed and processed again.

@dheepak-aot @guru-aot @andrepestana-aot @lewischen-aot thoughts?

andrepestana-aot: @andrewsignori-aot  You're right about checking success before deleting the file and that would be good to have separate log messages for the operations.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,93 @@
...
+        WHEN res.code = 'B2D' THEN 'B2'
+        ELSE res.code
+      END AS mapped_code,
+      res.individual_id,
+      res.is_included
+    FROM
+      sims.sfas_restrictions res
+    WHERE
+      res.removal_date IS NULL
+      AND res.is_included IS false
```
andrewsignori-aot, line 36: Is there a need to use `is` operator here? 
`AND res.is_included = false` would not work for some reason?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,93 @@
...
+ * Once all the restrictions have been added,
+ * mark the is_included column to true for all
+ * the newly imported restrictions.
+ */
+UPDATE
+  sims.sfas_restrictions
+SET
+  is_included = TRUE
+WHERE
+  is_included IS FALSE;
```
andrewsignori-aot, line 93: Same from https://github.com/bcgov/SIMS/pull/2848/files#r1496605607

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
andrewsignori-aot, line 1: Looking into the other SQL queries they use a more extended and easy-to-follow alias sometimes even just reperating the table name without the schema. Can you please do the same for this file?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,93 @@
...
+    SELECT
+      res.code,
+      CASE
+        WHEN res.code = 'A12' THEN '12'
+        WHEN res.code = 'AF1' THEN 'AF'
+        WHEN res.code = 'B2D' THEN 'B2'
+        ELSE res.code
+      END AS mapped_code,
+      res.individual_id,
+      res.is_included
```
andrewsignori-aot, line 31: Is it needed in the `select`?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,93 @@
+/* 
+ * After all the SFAS restrictions are imported, 
+ * the below query inserts one Legacy restriction entry 
+ * per student if they have one or more restrictions 
+ * imported from the sfas system that are not present 
+ * in the sims restrictions.
+ */
+INSERT INTO
```
andrewsignori-aot, line 8: Is this query considering the scenario where the student already has an active legacy restriction?
Is the `is_active` check missing?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
andrewsignori-aot, line 31: Are these needed in the `select`?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,93 @@
...
+      res.removal_date IS NULL
+      AND res.is_included IS false
+  ) r
+  INNER JOIN sims.sfas_individuals i ON r.individual_id = i.id
+  INNER JOIN sims.restrictions restrictions ON r.mapped_code = restrictions.restriction_code
+  LEFT JOIN sims.student_restrictions str ON str.student_id = i.student_id
+  AND str.restriction_id = restrictions.id
+WHERE
+  i.student_id IS NOT NULL
+  AND str.is_active IS FALSE
```
andrewsignori-aot, line 80: Is this taking into consideration the scenario where the student never had the restriction and a restriction with the code will not be associated at all for that student?

andrewsignori-aot (commented):
Nice work, please take a look at the comments. Please take a close look at the comments around the SQL commands. I will need a closer look once the comments are addressed or replied to.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
sh16011993: Both work. Changed it to equality.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,93 @@
...
+    FROM
+      sims.sfas_restrictions res
+    WHERE
+      res.removal_date IS NULL
+      AND res.is_included IS false
+  ) r
+  INNER JOIN sims.sfas_individuals i ON r.individual_id = i.id
+  LEFT JOIN sims.restrictions restrictions ON r.mapped_code = restrictions.restriction_code
+WHERE
+  i.student_id IS NOT NULL
```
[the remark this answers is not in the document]
sh16011993: I don't think so. This column is not available with the table in the inner query. After doing the fix as per: https://github.com/bcgov/SIMS/pull/2848#discussion_r1496620252 I believe the query will become more readable.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
sh16011993: Yes, it does (if I am understanding your comment correctly).

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
sh16011993: Nice Catch. Thanks!

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
guru-aot: Yes even a singe operation failure should result in a retrying it by fixing again. I do agree. Also as @andrepestana-aot mentioned it would be better to have some log messages for these.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  sims.sfas_restrictions
+ADD
+  COLUMN is_included bool NOT NULL DEFAULT FALSE;
```
guru-aot, line 4: Why do we have this column? and why I believe we follow keywords BOOLEAN than bool. I am ok with the change but asking to implement the existing standards followed.

sh16011993: I have changed `bool` to `boolean` @guru-aot. I added this column to differentiate the pre-existing SFAS restrictions that have already been inserted into SIMS student restrictions table from the SFAS restrictions that are yet to be inserted.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
sh16011993: I have changed this to `processed` as I too believe that it is a better column name for what we are trying to achieve. Thanks @andrewsignori-aot.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  sims.sfas_restrictions
+ADD
+  COLUMN processed BOOLEAN NOT NULL DEFAULT FALSE;
```
dheepak-aot, line 4: the column name and the file name have a mismatch

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Rollback-add-col-is-included.sql
```diff
@@ -0,0 +1,3 @@
+-- Rollback the add processed column.
+ALTER TABLE
```
dheepak-aot, line 2: https://github.com/bcgov/SIMS/pull/2848/files#r1497959569

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
dheepak-aot, line 4: processed means processed in SIMS right?

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
```diff
@@ -131,7 +131,6 @@ export class SFASIndividualService
    * Updates and inserts new disbursement overaward balances from sfas to disbursement overawards table.
    */
   async updateDisbursementOverawards(): Promise<void> {
-    await this.updateStudentId();
```
dheepak-aot, line 134: Thanks for taking care of this.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
```diff
@@ -0,0 +1,6 @@
...
+
+COMMENT ON COLUMN sfas_restrictions.processed IS 'This field indicates whether the SFAS restriction has been added to the student restrictions table or not.';
```
andrewsignori-aot, line 6: The comment must be adjusted. Not every single record will be added to the student restrictions.
Can you adjust to something like below?
```
Indicates if an SFAS restriction record was verified to determine if a restriction must be created in the student account.
```

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
sh16011993: Yes, `processed` means inserted in sims `student_restrictions` table.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -138,36 +138,42 @@ export class SFASIntegrationProcessingService {
       }
       await Promise.all(promises);
       this.logger.log("Records imported.");
-
+      try {
+        /**
+         * Bulk operation to update student id in SFAS individuals table after importing data from SFAS.
+         */
+        await this.sfasIndividualService.updateStudentId();
```
andrewsignori-aot, line 145: As an alternative to keep this method cleaner, the individual methods could provide an internal try/catch allowing the `catch (error)`  at line 174 to handle the errors and possibly eliminating the need of the new private method `errorHandler`.

sources/packages/backend/apps/db-migrations/src/sql/SFASRestrictions/Add-col-is-included.sql
andrewsignori-aot: Not really, `processed` means that it was checked. There is a possibility to a restrictions be created in the student account but not every single SFAS record, even belonging to a SIMS student, will result in a new restriction inserted.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,118 @@
...
+      WHERE
+        student_restrictions.restriction_id = (
+          SELECT
+            id
+          FROM
+            sims.restrictions
+          WHERE
+            restriction_code = 'LGCY'
+        )
+        AND student_restrictions.is_active = true
```
andrewsignori-aot, line 36: I would assume that this part of the query is trying to avoid inserting a new `LGCY`  if one is already present, right? Would it not simplify the query to add an exists condition to the final where maybe?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,118 @@
...
+      sims.sfas_restrictions sfas_restrictions
+    WHERE
+      sfas_restrictions.removal_date IS NULL
+      AND sfas_restrictions.processed = false
+  ) mapped_restrictions
+  INNER JOIN sims.sfas_individuals sfas_individuals ON mapped_restrictions.individual_id = sfas_individuals.id
+  INNER JOIN sims.restrictions restrictions ON mapped_restrictions.mapped_code = restrictions.restriction_code
+  LEFT JOIN sims.student_restrictions student_restrictions ON student_restrictions.student_id = sfas_individuals.student_id
+  AND student_restrictions.restriction_id = restrictions.id
+WHERE
```
andrewsignori-aot, line 103: For clarity I would prefer to enforce the precedence as below
```sql
WHERE
  sfas_individuals.student_id IS NOT NULL
  AND (student_restrictions.is_active = false
  OR student_restrictions.restriction_id IS NULL);
```

sh16011993: Here, the part after the `OR` i.e. `student_restrictions.restriction_id IS NULL` is redundant. I have changed the equality check `student_restrictions.is_active = false` to `student_restrictions.is_active IS false` to ensure the `NULL` check.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,120 @@
...
+    student_restrictions.is_active = false
+    OR student_restrictions.restriction_id IS NULL
+  );
+
+/*
+ * Once all the restrictions have been added,
+ * mark the processed column to true for all
+ * the newly imported restrictions.
+ */
+UPDATE
```
dheepak-aot, line 115: Is it right to set processed for all the restrictions came. ?

what if a student comes to our system after the restrictions are imported.?

andrewsignori-aot: @dheepak-aot the flag is only for the scope of the bulk insert from the SFAS import.
Students created after the data is imported will have the possible restrictions imported during account creation. Does it make sense?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,120 @@
+/* 
+ * After all the SFAS restrictions are imported, 
+ * the below query inserts one Legacy restriction entry 
+ * per student if they have one or more restrictions 
+ * imported from the sfas system that are not present 
+ * in the sims restrictions.
+ */
+INSERT INTO
+  sims.student_restrictions (student_id, restriction_id)
```
dheepak-aot, line 9: As discussed, the audit user can be received as parameter and inserted to `sims.student_restrictions`

sources/packages/backend/libs/src-sql/sfas-restrictions/Get-legacy-restriction.sql
andrewsignori-aot, line 1: The expectation was to do such a simple query as below. Am I missing something?
```ts
restrictionRepo.findOne({ select: { id: true }, where: { restrictionCode: "THE CODE" } });
```

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,106 @@
...
+
+/* 
+ * After all the SFAS restrictions are imported, 
+ * the below query inserts restrictions from sfas 
+ * if those restrictions are valid restrictions 
+ * in the sims system and are either not present in the 
+ * sims student restrictions table or they are inactive. 
+ */
+INSERT INTO
+  sims.student_restrictions (student_id, restriction_id)
```
dheepak-aot, line 62: audit is missing here.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,106 @@
...
+  $1,
+  $2
+FROM
+  (
+    SELECT
+      DISTINCT(sfas_individuals.student_id) AS student_id
+    FROM
+      (
+        SELECT
+          sfas_restrictions.code,
```
dheepak-aot, line 21: is this required for select?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,106 @@
...
+          CASE
+            WHEN sfas_restrictions.code = 'A12' THEN '12'
+            WHEN sfas_restrictions.code = 'AF1' THEN 'AF'
+            WHEN sfas_restrictions.code = 'B2D' THEN 'B2'
+            ELSE sfas_restrictions.code
+          END AS mapped_code,
+          sfas_restrictions.individual_id,
+          sfas_restrictions.processed
```
dheepak-aot, line 29: Same here. is it required for select?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,106 @@
...
+ */
+INSERT INTO
+  sims.student_restrictions (student_id, restriction_id)
+SELECT
+  (sfas_individuals.student_id),
+  restrictions.id AS restriction_id
+FROM
+  (
+    SELECT
+      sfas_restrictions.code,
```
dheepak-aot, line 69: https://github.com/bcgov/SIMS/pull/2848/files#r1500907393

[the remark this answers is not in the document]
dheepak-aot: Also try to add the context of provinical restrictions to the comment.

dheepak-aot:
Thanks for making the changes. Added more comments.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,107 @@
...
+ */
+INSERT INTO
+  sims.student_restrictions (student_id, restriction_id, creator)
+SELECT
+  (sfas_individuals.student_id),
+  restrictions.id AS restriction_id,
+  $2
+FROM
+  (
+    -- select sfas_restrictions records with mapped restriction (from sfas to sims) codes
```
dheepak-aot, line 68: Please add a similar comment to the first query as well.

dheepak-aot (approved):
Thanks for making the changes. 👍

One minor comment not a blocker.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual.service.ts
```diff
@@ -94,7 +94,13 @@ export class SFASIndividualService
    * Bulk operation to update student id in SFAS individuals table after importing data from SFAS.
    */
   async updateStudentId(): Promise<void> {
-    await this.repo.manager.query(this.bulkUpdateStudentIdSQL);
+    try {
+      await this.repo.manager.query(this.bulkUpdateStudentIdSQL);
+    } catch (error) {
+      throw new Error("Error while updating ids imported from SFAS.", {
```
andrewsignori-aot, line 100: May be adjusted to be explicit about the ids, for instance, "Error while updating student ids imported from SFAS.".

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -138,28 +138,45 @@ export class SFASIntegrationProcessingService {
...
       await Promise.all(promises);
       this.logger.log("Records imported.");
-
+      /**
+       * Bulk operation to update student id in SFAS individuals table after importing data from SFAS.
+       */
+      this.logger.log("Updating student ids for SFAS individuals.");
+      await this.sfasIndividualService.updateStudentId();
+      this.logger.log("Student ids updated.");
+      /**
```
andrewsignori-aot, line 147: The inline comments should follow the below:
```
// Bulk operation to update student id in SFAS individuals table after importing data from SFAS.
```
If the comments are the same already present for the method, (e.g. `updateStudentId`) the comments are not needed IMO.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,113 @@
...
+      INNER JOIN sims.sfas_individuals sfas_individuals ON mapped_restrictions.individual_id = sfas_individuals.id
+      LEFT JOIN sims.restrictions restrictions ON mapped_restrictions.mapped_code = restrictions.restriction_code
+    WHERE
+      sfas_individuals.student_id IS NOT NULL
+      AND restrictions.restriction_code IS NULL
+  ) filtered_students
+WHERE
+  -- The below part of the query checks that for every student id which is a  
+  -- potential candidate to be inserted in the sims student_restrictions table
+  -- (using the above part of the query), is there already a pre-existing active
```
andrewsignori-aot, line 44: I guess that it should be `there is` instead of `is there`, right?

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-restrictions.sql
```diff
@@ -0,0 +1,113 @@
...
+ */
+INSERT INTO
+  sims.student_restrictions (student_id, restriction_id, creator)
+SELECT
+  (sfas_individuals.student_id),
+  restrictions.id AS restriction_id,
+  $2
+FROM
+  (
+    -- select sfas_restrictions records with mapped restriction (from sfas to sims) codes
```
andrewsignori-aot, line 74: Please keep the text consistent while mentioning SFAS. Sometimes it is all upper case and sometimes all lower case.

andrewsignori-aot (approved):
Please take a look at the minor comments, other than that, looks good 👍

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -138,28 +138,36 @@ export class SFASIntegrationProcessingService {
...
         } catch (error) {
-          const logMessage = `Error while deleting SFAS integration file: ${remoteFilePath}`;
-          result.summary.push(logMessage);
-          result.success = false;
-          this.logger.error(logMessage);
-          this.logger.error(error);
+          throw new Error(
+            `Error while deleting SFAS integration file: ${remoteFilePath}`,
+            {
+              cause: error,
```
andrepestana-aot, line 168: 👍

andrepestana-aot (approved):
LGTM!

lewischen-aot (approved):
LGTM 👍 Nice work @sh16011993

Issue #2783, opened 2024-02-01
Patroni going into recovery mode after changing the credentials


guru-aot:
Patroni going into recovery mode after changing the credentials for workers, api, queue-consumers and load-test-gateway from superuser to database user.

<img width="1531" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDBmQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6a196f8102c2c835a01a308bd37a8e7ece77726f/MicrosoftTeams-image%20%286%29.png" alt="MicrosoftTeams-image (6).png" />

<img width="1604" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDRmQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--91051576a4f371120aeb39c77645e0acec796d43/MicrosoftTeams-image%20%287%29.png" alt="MicrosoftTeams-image (7).png" />

<img width="1816" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDhmQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--01b936e8ab8be4c66078bbfab1edcbc9f921b2ed/MicrosoftTeams-image%20%288%29.png" alt="MicrosoftTeams-image (8).png" />

dheepak-aot:
pulling into sprint and added points as per @ninosamson

guru-aot:
As part of the analysis done, the database in dev is going into recovery mode when assessment-workflow-queue-retry queue is running every hour. 

<img width="1367" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3BVQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--01a3235854faff1332ace3969f718a981a945d35/image.png" alt="image.png" />


Points tested and analysis done:
1. This is not happening in test and is not repicable. 
2. Initially changed the permission of the database user accordingly of the superuser and tried to solve the issue, but this failed.
[...]

PR #2851, opened 2024-02-20
#2783 - Patroni Recovery mode Bug

guru-aot:
As a part of the analysis the relationId in the db models created sub query creation, that causes DB to go into recovery mode. So removing it will enhance the queries performance and eventually make 50K records easily.

So as a recommendation, removing it in the all the place we use till date is required. As a part of this effort, the relationId found that do not have references and does not have any test case failures are removed. The next part of this story would be removing the remaining RelationId part of this ticket #2853

### DIFF
--- a/sources/packages/backend/libs/sims-db/src/entities/application-student-file.model.ts
+++ b/sources/packages/backend/libs/sims-db/src/entities/application-student-file.model.ts
@@ -1,10 +1,4 @@
-import {
-  Entity,
-  JoinColumn,
-  ManyToOne,
-  PrimaryGeneratedColumn,
-  RelationId,
-} from "typeorm";
+import { Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from "typeorm";
 import { StudentFile } from ".";
 import { ColumnNames, TableNames } from "../constant";
 import { Application } from "./application.model";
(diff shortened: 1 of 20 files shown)


sources/packages/backend/libs/sims-db/src/entities/application-student-file.model.ts
```diff
@@ -31,9 +25,6 @@ export class ApplicationStudentFile extends RecordDataModel {
   })
   application: Application;
 
-  @RelationId((file: ApplicationStudentFile) => file.studentFile)
```
sh16011993, line 34: @guru-aot Just to confirm, the `@RelationId`(s) that you have removed as a part of this PR aren't used in our queries and the ones for which you have created the story are the ones that are being used (tied up in some way) and therefore need to be removed later?

Also, can you please briefly tell how this was putting the db in recovery mode.

andrepestana-aot (approved):
Looks good to me.

andrewsignori-aot (approved):
Thanks for putting effort and finding the query causing the issue. Looks good 👍

sources/packages/backend/libs/sims-db/src/entities/application-student-file.model.ts
guru-aot: As discussed in the meeting, i believe you might have got the reason.

dheepak-aot (approved):
Great effort to get the issue fixed. Looks good. 👍

Issue #2850, opened 2024-02-20
Accent characters getting translated during Institution file upload(Bulk Offerings | Bulk Withdrawal)


dheepak-aot:
**Describe the task**
During the effort #2504 we noticed a bug after updating the file reading with ASCII, the bug is that the accent characters like `áéíó` are being translated and saved to SIMS db.

NOTE: It is assumed that file integrations are not affected. But confirmation required on the same. 

**Acceptance Criteria**
- [ ] Find the root cause and fix this translation happening. Test with characters `áéíó`

michesmith:
@JasonCTang 7 beats or 8 beats?

andrewsignori-aot:
As stated in the [nodejs documentation](https://nodejs.org/docs/latest/api/buffer.html#buffer_buffer) when the encoding is set to ASCII the most significant bit will be unset causing the characters to be changed, as per the example below.
<img width="556" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNDZEQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4ca8855643180358cf0bea3a9e813ff570c54b92/image.png" alt="image.png" />

The same would not happen it `latin1` is used as per stated by the documentation.

andrewsignori-aot:
As per the meeting, @JasonCTang will schedule some time with IMB to try to settle the questions.
- Considering integrations through file upload and STFP(institutions, ESDC, etc.) is it ok to use regular ASCII (7bits) or can we use ASCII Extended/ISO-8859-1 (8 bits)?

andrewsignori-aot:
### Files generated to the SFTP (`Buffer.from`)

While creating the files to be uploaded to the SFTP both `ascii` and `latin1` resulted in the same bytes.

|Char(áéí)|Encoding|
|---|---|
|UTF-8|195, 161, 195, 169, 195, 173|
|ASCII|225, 233, 237|
|Latin1(ISO8859-1)|225, 233, 237|

#### Sample file generated with UTF8

![image](https://github.com/bcgov/SIMS/assets/61259237/9e5d0806-1da5-4bf0-b11d-8d5083b7d062)

#### Sample file generated with ASCII and Latin1(ISO8859-1) where the same result was produced

![image](https://github.com/bcgov/SIMS/assets/61259237/69b9e648-9749-4e99-b6ce-88d929658052)

[...]

PR #2879, opened 2024-02-23
#2850 - Accent characters getting translated during Institution file upload

andrewsignori-aot:
### Files uploaded to SIMS (`buffer.toString`)

As stated in the [nodejs documentation](https://nodejs.org/docs/latest/api/buffer.html#buffer_buffer) when the encoding is set to ASCII the most significant bit will be unset (to ensure the 7bit characteristic from ASCII) causing the characters to be changed, as per the example below.

image

The same would not happen if `latin1(ISO-8859-1)` is used as stated by the documentation and tested during the file uploads.

The requirement for files integrations, via upload or SFTP, is to use a single-byte encoding, and since UTF-8 is multibyte, `latin1(ISO-8859-1)` seems to be the only way to proceed. 

One E2E test was changed to ensure that the translation issue would be captured.

### Files generated to the SFTP (`Buffer.from`)

While creating the files to be uploaded to the SFTP both `ascii` and `latin1` resulted in the same bytes.

|Char(áéí)|Encoding|
|---|---|
|UTF-8|195, 161, 195, 169, 195, 173|
|ASCII|225, 233, 237|
|Latin1(ISO8859-1)|225, 233, 237|

#### Sample file generated with UTF8

image

#### Sample file generated with ASCII and Latin1(ISO8859-1) where the same result was produced

image

#### Sample file in latin1 using UTF8 characters outside ASCII extended

image

From the student profile below.

image


andrewsignori-aot:
Closed due to a change in the approach that will be done in a new ticket. Details in the associated ticket comments (#2850).

PR #2893, opened 2024-02-26
#2685 - Import Provincial Restrictions from SFAS [Bug Fix]

sh16011993:

**Bug:** Postgres has a limitation that it can execute only one statement with parameters as a part of the sql prepared statement. As a result, the processing of the file `bulk-insert-restrictions.sql` from the previous PR was failing as it had more than one sql queries all of which were a part of the same prepared statement.

**Fix:** Both the queries of the  `bulk-insert-restrictions.sql` file have been moved into their own respective sql files and therefore, each of them is a part of their own prepared statement. These queries form a part of the same transaction.


sources/packages/backend/libs/sims-db/src/entities/sfas-restriction.model.ts
```diff
@@ -58,4 +58,14 @@ export class SFASRestriction extends BaseModel {
     nullable: false,
   })
   extractedAt: Date;
+
```
andrewsignori-aot, line 61: Please remove this while line t keep the same style from along the class.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
```diff
@@ -45,10 +50,21 @@ export class SFASRestrictionService
         select: { id: true },
         where: { restrictionCode: "LGCY" },
       });
-      await this.repo.manager.query(this.bulkInsertStudentRestrictionsSQL, [
-        legacyRestriction.id,
-        creator.id,
-      ]);
+      await this.repo.manager.transaction(async (manager) => {
```
andrewsignori-aot, line 53: I believe that most of the time we start transactions like this from the `dataSource`.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
```diff
@@ -45,10 +50,21 @@ export class SFASRestrictionService
...
+        await manager.query(this.bulkInsertLegacyRestrictionsSQL, [
+          legacyRestriction.id,
+          creator.id,
+        ]);
+        await manager.query(this.bulkInsertSFASMappedRestrictionsSQL, [
+          creator.id,
+        ]);
+        manager
```
andrewsignori-aot, line 61: Missing `await`.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
```diff
@@ -45,10 +50,21 @@ export class SFASRestrictionService
         select: { id: true },
         where: { restrictionCode: "LGCY" },
       });
-      await this.repo.manager.query(this.bulkInsertStudentRestrictionsSQL, [
-        legacyRestriction.id,
-        creator.id,
-      ]);
+      await this.repo.manager.transaction(async (manager) => {
+        await manager.query(this.bulkInsertLegacyRestrictionsSQL, [
```
andrewsignori-aot, line 54: Is there anything that would prevent us from executing `bulkInsertLegacyRestrictionsSQL` and `bulkInsertSFASMappedRestrictionsSQL` in parallel?

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
andrewsignori-aot, line 53: Minor, `manager` is usually named as `transactionalEntityManager` or just `entityManager`.

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
```diff
@@ -0,0 +1,43 @@
...
+ * the below query inserts restrictions from sfas 
+ * if those restrictions are valid restrictions 
+ * in the sims system if they are either not present in the 
+ * sims student restrictions table or they are inactive. All
+ * the restrictions dealt here are provincial restrictions only. 
+ */
+INSERT INTO
+  sims.student_restrictions (student_id, restriction_id, creator)
+SELECT
+  (sfas_individuals.student_id),
```
andrewsignori-aot, line 12: Was there a need to wrap this in parenthesis?

andrewsignori-aot (commented):

andrewsignori-aot (approved):
Thanks for making the changes and for finding the alternative solution quickly. Looks good 👍

andrepestana-aot (approved):
LGTM.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
lewischen-aot: @andrewsignori-aot Correct me if I'm wrong. I thought we need to execute `bulkInsertLegacyRestrictionsSQL` first then execute `bulkInsertSFASMappedRestrictionsSQL`. If we put them into parallel with `Promise.all([])`, there is a chance that `bulkInsertSFASMappedRestrictionsSQL` could be completed before `bulkInsertLegacyRestrictionsSQL`.

sources/packages/backend/libs/integrations/src/services/sfas/sfas-restriction.service.ts
```diff
@@ -45,10 +51,28 @@ export class SFASRestrictionService
...
+          );
+        const bulkInsertSFASMappedRestrictionsPromise =
+          transactionalEntityManager.query(
+            this.bulkInsertSFASMappedRestrictionsSQL,
+            [creator.id],
+          );
+        await Promise.all([
+          bulkInsertLegacyRestrictionsPromise,
+          bulkInsertSFASMappedRestrictionsPromise,
+        ]);
```
lewischen-aot, line 68: @andrewsignori-aot Correct me if I'm wrong. I thought we need to execute `bulkInsertLegacyRestrictionsSQL` first and execute `bulkInsertSFASMappedRestrictionsSQL` after. If we put them into parallel with `Promise.all()`, there is a chance that `bulkInsertSFASMappedRestrictionsSQL` could be completed before `bulkInsertLegacyRestrictionsSQL`.

sh16011993: @lewischen-aot they are independent sql queries because they don't affect each other and have their own specific inserts. `bulkInsertLegacyRestrictionsSQL` is responsible for inserting Legacy restriction entry if there is none active for that particular student and the student gets one or more restrictions from SFAS which are not present is SIMS. The other query is responsible for inserting all the other restrictions which have an available mapping into SIMS and are either not present in SIMS `student_restrictions` table or are inactive.

lewischen-aot: If they are independent sql, then we can use `Promise.all()`. Thanks for the clarification.

lewischen-aot (approved):
LGTM 👍

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
```diff
@@ -0,0 +1,43 @@
...
+SELECT
+  sfas_individuals.student_id,
+  restrictions.id,
+  $1
+FROM
+  (
+    -- select sfas_restrictions records with mapped restriction (from SFAS to SIMS) codes
+    -- and join with restrictions and student_restrictions table to
+    -- get the student ids to create restrictions.
+    SELECT
```
guru-aot, line 20: 👍

guru-aot (approved):
LGTM, nice work @sh16011993

Issue #2775, opened 2024-01-31
PT Request a Change - Add Dependants


HRAGANBC:
**User Story:**
As a student I want to be able to report additional dependants to a complete application so that a new family size can be incorporated after I've received initial funding and result in additional funding to cover my costs. 

**Description:**
Add a new form into the student Request a Change process allowing for reporting of new dependants

**Acceptance Criteria:**
- [x] Add new Request a Change form which is called "Add/Remove Dependants (Part-Time Applications)" in the student dropdown menu
- [x] The form should prompt two questions (and all associated sub-questions) exactly as they appear on the initial application form:
- Do you have any dependants? (+ attending post-secondary + on taxes for disability?)
- Are any of the dependants listed above supported by you financial but not under your sole custody?
- [x] Embedd the file uploader within the question form consistent with how they appear in the Application Form- Files should Only be viewable in the Request a Change Form
- [x] E2E tests

<img width="632" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeW8wQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--015de81c7be4e81dce9ba25049cca4b1e7ccca72/image.png" alt="image.png" />

PR #2901, opened 2024-02-27
#2471 -  Financial information request a change

dheepak-aot:
## Request for change - Financial Information

- [x] New form.io created for financial information - request for change.

image

- [x] Request a change form enhanced with new value in the dropdown to request for change.

image

- [x] During the load of all the change request forms(appeal forms), program year is loaded to form initial data.
- [x] Based on the program year, the tax year in the label for the total income is derived in a dynamic way.
image

- [x] All the money values entered in the form are validated against maximum money value and negative values.
image

-  [x] `MONEY_VALUE_FOR_UNKNOWN_MAX_VALUE` set to 100000000 in the code.
image

Note: E2E tests have been updated due to this change.
- [x] Get API updated to send the program year along with existing basic application details.

**Note:** The existing DTO(was application identifiers DTO) has been used as base DTO in many other application DTOs.
With the program year in it, it cannot be a base DTO anymore and hence no more DTOs are extending this DTO.

- [x] During the dry run submission of all change request forms(appeal forms), program year is added from the server side to perform the dry run submission. 
This action is based on the condition that, if the that particular form requires the program year to perform dry run submission. Created a class `StudentAppealNonSubmissionData` to type/capture all the properties of change request forms which does not belong to the submitted data e.g(programYear)

 - [x] Added E2E tests mocking the dry run submission result but ensuring that the dry run submission is called.

image

PR #2902, opened 2024-02-27
#2775 - PT Request a Change - Add Dependants

lewischen-aot:
- Added new Request a Change dropdown which is called "Add/Remove Dependants (Part-Time Applications)" in the Request a Change page
- Added a new FormIO file for "Appeals - Student Dependants Part Time"
  - All questions exactly as they appear on the initial application form are added
    - Do you have any dependants? (+ attending post-secondary + on taxes for disability?)
    - Are any of the dependants listed above supported by you financial but not under your sole custody?
  - Embedd the file uploader within the question form consistent with how they appear in the Application Form- Files should Only be viewable in the Request a Change Form
- Added E2E tests for `submitStudentAppeal` method
- Added migrations to add "Appeal" to file origin column in student file table 

Screenshot of the dropdown button in Request a Change page
image

Screenshot of the student dependants appeal page
Screenshot 2024-02-26 185226


sources/packages/forms/src/form-definitions/studentrequestchange.json
```diff
@@ -455,6 +455,10 @@
               {
                 "value": "studentdisabilityappeal",
                 "label": "Disability Information"
+              },
+              {
+                "value": "studentdependantsappealparttime",
+                "label": "Add/Remove Dependants (Part-Time Applications)"
```
dheepak-aot, line 461: As of now the change request forms are going to be generic (at least until MVP as per business) not based upon the offering intensity.

i.e the name or title of the change request forms need to be generic e.g. `Add/Remove Dependents`

dheepak-aot, line 461: Minor. Also grammatically `Dependant` is valid but `Dependants`(plural) is not valid. 
So it could be called Dependents. Please confirm with business.

lewischen-aot: Sure. Will check with Huge.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+                    "input": false,
+                    "title": "Dependents that you support financially but do not have sole custody",
+                    "theme": "default",
+                    "tableView": false,
+                    "components": [
+                        {
+                            "input": true,
+                            "tableView": false,
+                            "key": "dependantsSharedCustodyApplicationException",
+                            "label": "Application Exception - Dependant’s shared custody",
```
dheepak-aot, line 727: How are the exceptions considered on the request for change perspective?? There is no workflow logic or consideration on this currently.

Can we check with business on this?

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
+{
+    "title": "Appeals - Student Dependants Part Time",
```
dheepak-aot, line 2: https://github.com/bcgov/SIMS/pull/2902/files#r1505184681

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+    "path": "studentdependantsappealparttime",
+    "type": "form",
+    "display": "form",
+    "tags": [
+        "common"
+    ],
+    "components": [
+        {
+            "key": "dependantsInformationTitle",
+            "label": "Dependants Information (Part-Time Applications)",
```
dheepak-aot, line 13: https://github.com/bcgov/SIMS/pull/2902/files#r1505184681

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+            "theme": "default",
+            "tableView": false,
+            "components": [
+                {
+                    "autofocus": false,
+                    "input": true,
+                    "tableView": true,
+                    "inputType": "radio",
+                    "label": "<strong>Do you have any dependants?</strong>",
+                    "key": "hasDependents",
```
dheepak-aot, line 50: For appeal forms whatever existed so far, we prefixed the variables with `studentNew`.

e.g. here `studentNewHasDependents` 

same for `dependants` and `supportnocustodyDependants`

@andrewsignori-aot @guru-aot @andrepestana-aot @sh16011993 should we have the `studentNew` prefix for api keys in appeal forms as standard practice ??

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+                                                    "labelPosition": "top",
+                                                    "tags": [],
+                                                    "conditional": {
+                                                        "show": "true",
+                                                        "when": "declaredOnTaxes",
+                                                        "eq": "yes"
+                                                    },
+                                                    "properties": {},
+                                                    "storage": "url",
+                                                    "url": "student/files",
```
dheepak-aot, line 534: Are we suppose to re-use the student files table for appeals??

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
+{
```
dheepak-aot, line 1: Please add the hidden variable `programYear` to this form.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+  beforeAll(async () => {
+    const { nestApplication, module, dataSource } =
+      await createTestingAppModule();
+    app = nestApplication;
+    appDataSource = dataSource;
+    appModule = module;
+    applicationRepo = dataSource.getRepository(Application);
+    studentAppealRequestRepo = dataSource.getRepository(StudentAppealRequest);
+    studentAppealRepo = dataSource.getRepository(StudentAppeal);
+    student = await saveFakeStudent(appDataSource);
```
dheepak-aot, line 54: I would not recommend using same student for all tests as we extend the scenarios, there could be data interference.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+    const studentAppealRequest = createFakeStudentAppealRequest();
+    const studentAppeal = createFakeStudentAppeal({
+      application,
+      appealRequests: [studentAppealRequest],
+    });
+    await studentAppealRepo.save(studentAppeal);
+    // Set the student appeal status to Pending
+    studentAppealRequest.studentAppeal = studentAppeal;
+    studentAppealRequest.appealStatus = StudentAppealStatus.Pending;
+    await studentAppealRequestRepo.save(studentAppealRequest);
```
dheepak-aot, line 140: This portion of code must come under `//Arrange` comment

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+
+    // Assert
```
dheepak-aot, line 142: `// Act/Assert ` ?

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+  });
+
+  it("Should throw unprocessable entity exception error when there is already a pending appeal.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(appDataSource, {
+      student: student,
+    });
+    const endpoint = `/students/appeal/application/${application.id}`;
+    const payload = {
+      studentAppealRequests: [{ formData: {}, formName: "test" }],
```
dheepak-aot, line 123: I would suggest to put a very simple real time payload for formData (e.g. student answering no for has dependents) so that when we bring the dry run submission into E2E tests it will be easy.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+      .send(payload)
+      .expect(HttpStatus.BAD_REQUEST)
+      .expect({
+        statusCode: HttpStatus.BAD_REQUEST,
+        message: "Not able to submit student appeal due to invalid request.",
+        error: "Bad Request",
+      });
+  });
+
+  it("Should determine student's new income when the student submits an appeal for student new income.", async () => {
```
dheepak-aot, line 193: Student income appeal is a POC and I would say we should create E2E for that.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+    const studentNewIncome =
+      newStudentAppealRequest.submittedData.studentNewIncome;
+    expect(studentNewIncome).toBe(6000);
+
+    // Set the student appeal status to Approved to allow the next test case to be processed
+    newStudentAppealRequest.appealStatus = StudentAppealStatus.Approved;
+    await studentAppealRequestRepo.save(newStudentAppealRequest);
+  });
+
+  it("Should determine partner's new income when the student submits an appeal for partner's new income.", async () => {
```
dheepak-aot, line 243: This appeal is also a POC and stay away from E2E tests.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+    });
+
+    // Assert
+    const response = await request(app.getHttpServer())
+      .post(endpoint)
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .send(payload)
+      .expect(HttpStatus.CREATED);
+
+    const newStudentAppealRequest = await studentAppealRequestRepo.findOne({
```
dheepak-aot, line 329: just a suggestion, expect newStudentAppealRequest to be defined before doing further checks.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+      where: { studentAppeal: { id: response.body.id } },
+    });
+    const studentNewPDPPDStatus =
+      newStudentAppealRequest.submittedData.studentNewPDPPDStatus;
+    expect(studentNewPDPPDStatus).toBe("yes");
+
+    // Set the student appeal status to Approved to allow the next test case to be processed
+    newStudentAppealRequest.appealStatus = StudentAppealStatus.Approved;
```
dheepak-aot, line 337: This doesn't need to happen when student is created as test case level.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -0,0 +1,560 @@
...
+      .post(endpoint)
+      .auth(studentToken, BEARER_AUTH_TYPE)
+      .send(payload)
+      .expect(HttpStatus.CREATED);
+
+    const newStudentAppealRequest = await studentAppealRequestRepo.findOne({
+      where: { studentAppeal: { id: response.body.id } },
+    });
+    const dependants = newStudentAppealRequest.submittedData.dependants;
+    expect(dependants.length).toBe(2);
```
dheepak-aot, line 429: We can assert the whole input payload against the persisted one using strictEqual

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+                                                    "conditional": {
+                                                        "show": "true",
+                                                        "when": "declaredOnTaxes",
+                                                        "eq": "yes"
+                                                    },
+                                                    "properties": {},
+                                                    "storage": "url",
+                                                    "url": "student/files",
+                                                    "dir": "PD Dependent",
+                                                    "validate": {
```
dheepak-aot, line 536: Also viewing the uploaded files is not rendering properly.

image

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -145,6 +145,43 @@ export class StudentAppealStudentsController extends BaseController {
         } as StudentAppealRequestModel),
     );
 
+    // Obtain the data for validDependent and isavalidDependant APIs for each dependant
+    // for student dependents appeal if it has dependents
+    for (let appealRequest of appealRequests) {
```
dheepak-aot, line 150: I have 2 questions here.

1 - Do we really need the hidden variables `validDependent` `isavalidDependant`, they look like obsolete to me.

2 - If they are required, we can use form io to do the logic as per student application.
To populate the hidden variables on submission, please do it from form.io by enabling this option.

image

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
```diff
@@ -383,6 +383,10 @@ export interface ApplicationData {
    * Offering id selected by the student.
    */
   selectedOffering?: number;
+  /**
+   * Offering start date of the selected offering.
+   */
+  selectedOfferingDate?: string;
```
dheepak-aot, line 389: It can be removed based on the response for the question https://github.com/bcgov/SIMS/pull/2902/files#r1506504337

dheepak-aot:
Nice work @lewischen-aot . Please have a look at the comments.

As discussed, please add new file origin type `Appeal` and update the origin type to Appeal on submission.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
andrewsignori-aot: It may be obsolete but it would require a further looking.
The form.io logic should never be replicated in the server and the `Calculated on server` must be used for similar cases.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
andrewsignori-aot: In my current PR, I also used a student globally defined but I made that explicit naming it as `sharedStudent`. Any test that would demand customization would need to create a new one.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1311,6 +1311,10 @@ export class ApplicationService extends RecordDataModelService<Application> {
         id: true,
         applicationNumber: true,
         isArchived: true,
+        data: {
```
andrewsignori-aot, line 1314: Unfortunately, Typeorm does not support this type of query and this will not produce the expected result. The entire payload will be retrieved either way. To avoid this we need special JSONB instructions that we can use if there is a performance concern. For this case, the below suggestion would be preferred.
Can you please change it as below:
```ts
data: true as unknown
```

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1311,6 +1311,10 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+          studystartDate: true,
```
andrewsignori-aot, line 1315: We use the dynamic data from the application for some specific situations. In general, we will not be consuming, for instance, `studystartDate` and `selectedOfferingDate` from the dynamic data.
Relevant data, like offering data, should be stored somewhere else in DB and should be consumed from there 😉

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
andrewsignori-aot: I noticed that it was added during the PD/PPD appeal implementation but I do not see a reason why we should do it.
IMO we can keep the same value from the original form, when one is available, keeping in mind that the appeal form can potentially receive inputs that are not even in the application form and that would be one more reason that using "new" would not be advised.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+                                            },
+                                            "properties": {},
+                                            "lockKey": true
+                                        },
+                                        {
+                                            "autofocus": false,
+                                            "input": true,
+                                            "tableView": false,
+                                            "inputType": "radio",
+                                            "label": "<strong>Attending post secondary institution</strong>",
```
andrewsignori-aot, line 296: Probably not a blocker for this PR since it seems to be in too many places but we should stop using `<strong>` along the labels and start to use the CSS styles available, for instance `"customClass": "label-bold"`.
It can potentially impact the accessibility and either way does not seem the right way to do it.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
lewischen-aot: @dheepak-aot What is a POC? Where should we create E2E test for this scenario?

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
lewischen-aot: Thanks @dheepak-aot and @andrewsignori-aot. I will keep the same keys as they are in the original family information form.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
lewischen-aot: As discussed in the meeting, we will keep the same url but add "Appeal" as an enum type to the file_origin column.

sources/packages/forms/src/form-definitions/studentrequestchange.json
lewischen-aot: I have a response from the business. @Joshua-Lakusta points out that the British way is the way we are supposed to follow. Auto correct wants "Dependent", but we should be using "Dependant" and "Dependants" as its plural in Canada. We use "Dependant" as the noun and "Dependent" as the adjective.  For example, we are dependent on our developers to do the development work, and my child is my dependant. Thanks @Joshua-Lakusta. I think we will keep using "dependant" and its plural "dependants" throughout the project.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
lewischen-aot: The reason I moved them to the server is because I would need to change multiple DTO files and send data from server to the form.io then back again, but the current approach is do the calculation in a centralized spot for calculation whose data come be obtained directly from the server. I see now that we should do the calculation on the form.io not on the server. I will a dd the calculation for `validDependent` and `isavalidDependant` on the form.io.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
lewischen-aot: It has been settled on the Teams discussion that these "exception" keys will be removed in this PR.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,970 @@
...
+                    "optionsLabelPosition": "right",
+                    "tags": [],
+                    "conditional": {
+                        "show": "",
+                        "when": null,
+                        "eq": ""
+                    },
+                    "properties": {},
+                    "tooltip": "Dependants represent anyone who you support financially.",
+                    "errorLabel": "Dependent",
```
sh16011993, line 85: Is the error label as expected?

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
sh16011993: I concur with the decision.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
andrewsignori-aot: To be clear, as much as possible the custom validations in form.io definition should be as minimal as possible, but once they are there we do not need to have them replicated on the server-side and we must ensure the `Calculate on server` config as true.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
guru-aot: I agree, let not have the studentNew prefixed

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
guru-aot: As @dheepak-aot mentioned, these values calculations can be moved to formio server side calculation and retrieved accordingly.

sources/packages/backend/apps/api/src/services/application/application.service.ts
guru-aot: if it would have be returned like that, that would be of great use :)

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
lewischen-aot: Nice spotting. It is from the original form, but I see we should use dependant instead of dependent. If we use empty string for `errorLabel`, and the result will be the image below.

image

If we keep "dependant", then it will say "dependant is required". Please let me know if we should change it.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
dheepak-aot: I also see the token being generated at before each here. We had suggestions in the past to keep it in test level right?

guru-aot (commented):
I had a look and I echo most of the comments from the other devs, please have a look on the other devs comments

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
sh16011993: @dheepak-aot From my understanding, we will still need the hidden variables `validDependent` and `isavalidDependant` and since we do the calculations on the formio and then send it to the api everywhere else in the application, I believe we can follow the same approach here as well.

sh16011993 (commented):
@lewischen-aot Nice work 👍  Nice e2e coverage. I left a minor comment. Please have a look at the comments from other devs.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
lewischen-aot: After having the discussion with @andrewsignori-aot and @dheepak-aot, we found that `validDependent` and `isavalidDependant` are not required anywhere else, and similar validations are being done in Camunda. Searching validDependent or isavalidDependant in VS Code shows that they appear only in "sfaa" files. For now it's safe to remove them from this PR, and I will create a story to remove them in the application form. It could be a good first issue. FYI @guru-aot @shaangill025 @andrepestana-aot 🙂

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
dheepak-aot: @lewischen-aot I removed the `<strong>` as per my discussion with @andrewsignori-aot and the labels still look bold and good dud to the css that we created. Hence please remove the strong tag here and the label must appear bold without that.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
lewischen-aot: I removed "Dependent" from the "errorLabel", so it will look like the image above now.

sources/packages/forms/src/form-definitions/studentrequestchange.json
```diff
@@ -459,6 +459,10 @@
               {
                 "label": "Partners Income",
                 "value": "partnerincomeappeal"
+              },
```
dheepak-aot, line 462: In my story PR we talked about alphabetical order for the list of forms for  request a change.
As per the order this should be first. 

https://github.com/bcgov/SIMS/pull/2901#discussion_r1506660164

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,933 @@
...
+                "eq": ""
+            },
+            "properties": {},
+            "lockKey": true
+        },
+        {
+            "clearOnHide": false,
+            "key": "dependantsAppealPanel",
+            "input": false,
+            "title": "Dependants Information Part Time Applications Panel",
```
dheepak-aot, line 39: Minor. can you remove the part time from the title? - I know is hidden but could be confusing for a dev looking in future.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,933 @@
...
+                                                    "input": false,
+                                                    "tag": "h3",
+                                                    "attrs": [
+                                                        {
+                                                            "value": "",
+                                                            "attr": ""
+                                                        }
+                                                    ],
+                                                    "className": "category-header-medium primary-color",
+                                                    "content": "Upload proof of your dependants income tax return for 20[XX]",
```
dheepak-aot, line 395: I consider this to be NOT part of the ticket. But can ask business
Does the value of the tax year needs to be precise here and student application replacing XX??

image

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -195,6 +213,385 @@ describe("StudentAppealStudentsController(e2e)-submitStudentAppeal", () => {
...
+      DEPENDANT_INFORMATION_FORM_NAME,
+      {
+        ...dependantInformationData,
+        programYear: application.programYear.programYear,
+      },
+    );
+  });
+
+  it(
+    "Should determine the total number of student appeal requests for a student appeal when the student " +
```
dheepak-aot, line 460: I would say that we have to keep student income and partner income appeal out of the tests totally as they are not built on real requirements.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitStudentAppeal.e2e-spec.ts
```diff
@@ -195,6 +213,385 @@ describe("StudentAppealStudentsController(e2e)-submitStudentAppeal", () => {
...
+      .send(payload)
+      .expect(HttpStatus.BAD_REQUEST)
+      .expect({
+        statusCode: HttpStatus.BAD_REQUEST,
+        message: "Not able to submit student appeal due to invalid request.",
+        error: "Bad Request",
+      });
+  });
+
+  it("Should determine student dependants when the student submits an appeal for student dependants.", async () => {
```
dheepak-aot, line 352: should save student dependent details?

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -510,4 +515,53 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
...
     return orderByCondition;
   }
+
+  /**
+   * Get student dependant appeal from student appeal requests and
+   * update student file origin type to Appeal for any uploaded files
+   * @param userId Student user who submits the appeal.
+   * @param studentAppealRequests Payload data.
+   */
+  async updateStudentAppealFiles(
```
dheepak-aot, line 525: This whole method could be avoided, if we use the existing utility and send the associated files as paylod from Vue.

image

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
```diff
@@ -0,0 +1,933 @@
...
+                                                        "required": true
+                                                    },
+                                                    "lockKey": true
+                                                },
+                                                {
+                                                    "clearOnHide": false,
+                                                    "label": "Columns",
+                                                    "input": false,
+                                                    "tableView": false,
+                                                    "key": "addDependantsUploadColumns",
```
dheepak-aot, line 521: Is this an existing issue in file download ??

https://github.com/bcgov/SIMS/pull/2902#discussion_r1506486077

dheepak-aot:
Thanks for making the changes @lewischen-aot . Added more comments.

sources/packages/backend/apps/db-migrations/src/sql/Types/Add-file-origin-type-appeal.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TYPE sims.file_origin_type
```
dheepak-aot, line 1: We decided recently to replace the enums instead of adding.

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
CarlyCotton: @dheepak-aot Not technically part of this ticket. Will need to be specific based on the program year of the application the change request is for.

sh16011993 (approved):
Nice Work @lewischen-aot 👍 LGTM.

sources/packages/forms/src/form-definitions/studentrequestchange.json
```diff
@@ -444,6 +444,10 @@
           "dataSrc": "values",
           "data": {
             "values": [
+              {
+                "value": "studentdependantsappealparttime",
```
dheepak-aot, line 448: Can you please remove parttime from the value and also the other form??

image

sources/packages/forms/src/form-definitions/studentdependantsappealparttime.json
dheepak-aot: Also on the file name and the path.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -161,6 +166,18 @@ export class StudentAppealStudentsController extends BaseController {
       userToken.userId,
       appealRequests,
     );
+
+    const student = await this.studentService.getStudentByUserId(
+      userToken.userId,
+    );
+
+    await this.studentFileService.updateStudentFiles(
```
dheepak-aot, line 174: This must happen in transaction with the student appeal save.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -161,6 +166,18 @@ export class StudentAppealStudentsController extends BaseController {
...
+      student.id,
+      userToken.userId,
+      payload.studentAppealRequests[0].files,
```
dheepak-aot, line 177: An appeal can be submitted with multiple appeals (financial info, pd status and etc etc)

We cannot assume that first one will have files. 

You can create a flatmap of all the files and call update Student files and this can be done inside the service.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/models/student-appeal.dto.ts
```diff
@@ -23,6 +23,8 @@ export class StudentAppealRequestAPIInDTO {
   @IsDefined()
   @JsonMaxSize(JSON_10KB)
   formData: unknown;
+  @IsDefined()
+  files: string[];
```
dheepak-aot, line 27: Is it expected to come as `[ ]` for appeals without files? or undefined?

lewischen-aot: Yes, it is expected to come as `[ ]` for appeals without files because `getAssociatedFiles(appealForm)` returns an array of strings or an empty array (`const getAssociatedFiles: (form: any) => string[]`)

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -76,7 +81,23 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
           createdAt: currentDateTime,
         } as StudentAppealRequest),
     );
-    return this.repo.save(studentAppeal);
+
+    // Update the student appeal and files if exist using the same entity manager
+    // so that they are processed within the same transaction
+    return await this.dataSource.transaction(async (entityManager) => {
```
dheepak-aot, line 87: this.dataSource.transaction can be returned which will be awaited at consumer end.

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -76,7 +81,23 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
...
+      const student = await this.studentService.getStudentByUserId(userId);
+      const uniqueFileNames: string[] = studentAppealRequests.flatMap(
+        (studentAppeal) => studentAppeal.files,
+      );
+      this.studentFileService.updateStudentFiles(
```
dheepak-aot, line 92: missing await here.

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
```diff
@@ -119,35 +119,78 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
...
+        options.entityManager,
+        studentId,
+        auditUserId,
+        uniqueFileNames,
+        fileOrigin,
+        options,
+      );
 
       // Create student notification on file upload.
       if (options?.saveFileUploadNotification) {
```
dheepak-aot, line 136: Saving file upload notification can go inside the private method.

image

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -76,7 +81,23 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
...
+        student.id,
```
dheepak-aot, line 93: Student id can be sent from the controller through the token. Please use the `StudentUserToken` in the controller.

image

dheepak-aot:
Thanks for doing the changes. Added last set of comments.

dheepak-aot (approved):
Thanks for doing the changes @lewischen-aot 👍 Looks good 👍

guru-aot (approved):
LGTM

PR #2916, opened 2024-03-01
#2685 - Import Provincial Restrictions from SFAS into SIMS [Bug Fix 2]

sh16011993:
### As a part of this PR, the following two bugs were fixed:
-----------------------------------------------------------------------
- **Bug:** The Resolved Column was not showing the date for Resolved restrictions and was instead showing a Date for the Active Restrictions.
  **Fix:** It has been fixed to show the resolved date for the Resolved restrictions and not show a date under the Resolved column for the currently active restrictions.
image

---------------------------------------------------------------------
- **Bug:** The `Bulk-insert-sfas-mapped-restrictions.sql` query was not checking for the scenario when a specific restriction for a student mapped from SFAS is inactive is SIMS and at the same time, the same restriction has entries as an active restriction. If this situation was there, the query was still allowing for this restriction to be inserted.

  **Fix:** This query was adjusted to make sure that if a restriction is found in the `student_restrictions` table that was mapped from the `sfas_restrictions` table but is inactive, then the query now checks that there are no other entries of this same restriction for this particular student that are present in the `student_restrictions` table that are currently active. If active restriction entries are found, then this restriction won't be inserted.
image

### DIFF
--- a/sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
+++ b/sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
@@ -11,7 +11,7 @@ INSERT INTO
 SELECT
   sfas_individuals.student_id,
   restrictions.id,
-  $ 1
+  $1
 FROM
   (
     -- select sfas_restrictions records with mapped restriction (from SFAS to SIMS) codes
--- a/sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
+++ b/sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
@@ -38,6 +38,22 @@ FROM
 WHERE
   sfas_individuals.student_id IS NOT NULL
   AND (
-    student_restrictions.is_active = false
-    OR student_restrictions.restriction_id IS NULL
+    student_restrictions.restriction_id IS NULL
+    OR (
+      -- if the restriction is found in the student_restrictions table
+      -- but is inactive, the below condition checks that there are no 
+      -- other entries of this same restriction for this particular student 
+      -- present in the student_restrictions table that are currently active.
+      student_restrictions.is_active = false
+      AND NOT EXISTS (
+        SELECT
+          1
+        FROM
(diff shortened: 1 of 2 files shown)


sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
```diff
@@ -38,6 +38,22 @@ FROM
 WHERE
   sfas_individuals.student_id IS NOT NULL
   AND (
-    student_restrictions.is_active = false
-    OR student_restrictions.restriction_id IS NULL
+    student_restrictions.restriction_id IS NULL
```
andrewsignori-aot, line 41: Is it not possible to add the `is_active` condition to the `LEFT JOIN` as below or something similar?
```sql
LEFT JOIN sims.student_restrictions student_restrictions
   ON student_restrictions.student_id = sfas_individuals.student_id
      AND student_restrictions.is_active = true
```

sources/packages/web/src/components/common/students/StudentRestrictions.vue
```diff
@@ -48,7 +48,7 @@
           <Column field="updatedAt" header="Resolved">
             <template #body="slotProps">{{
               conditionalEmptyStringFiller(
-                slotProps.data.isActive,
+                !slotProps.data.isActive,
```
andrewsignori-aot, line 51: Thanks for taking care of it.

guru-aot (commented):
Please have a look on the Query, if it can be handled in the Left join rather creating a separate not exists in the where, as mentioned by @andrewsignori-aot .

sources/packages/backend/libs/src-sql/sfas-restrictions/Bulk-insert-sfas-mapped-restrictions.sql
sh16011993: Thanks for suggesting it @andrewsignori-aot . Simplifies the logic a lot.

andrewsignori-aot (approved):
Thanks for making the changes, nice finding 😉 
Look good 👍

guru-aot (approved):
LGTM, nice work @sh16011993

dheepak-aot (approved):
Looks good @sh16011993 👍

Thanks for taking care for the Vue fix as well.

Issue #2700, opened 2024-01-18
MSFAA File Creation Revisions


CarlyCotton:
**User Story**
As a _ministry_, we send two files to show which students have signed up for funding (FT & PT) and it has their MSFAA numbers in them.
TPBC.EDU.MSFA.SENT.PT.XXXXXXXX.XXX
TPBC.EDU.MSFA.SENT.XXXXXXXX.XXX

**Current Behaviour**
There are some formatting issues with the data going into the file that are creating unnecessary soft fails on processing.
- Phone Number (chars 252-272) is "left-aligned" with trailing spaces. "12501112222         "

**Expected Behaviour**
- Phone Number (chars 252-272) should be "right-aligned" with leadings 0s. "00000000001250111222"

**Further Context FYI **
The Federal Government cannot accept Special Characters like a ' ~ '  a tilde,  but they can accepts accents  e'. These files will fail and will trigger an manual process for validation 

**Acceptance Criteria**
- [ ] Convert the phone number text to only numbers and add to MSFAA
- [ ] Update phone number field generated

CarlyCotton:
@HRAGANBC we got soft rejections on phone number, and missing first name. Hard rejections for invalid Institution Code and age. I'm moving this to lather so we can knock some of these off prior to Cycle 2.

I think the age should be addressed prior to getting to the MSFAA file feedback stage. Looking into policies re minimum age so we can look at alternate options. 
Missing first name is something that we get a specific error code on feedback for and has been handled through SABC/NSLSC discussion.

PR #2967, opened 2024-03-08
#2700 - MSFAA File Creation Revisions

sh16011993:

- Converted the phone numbers with hyphens to only numbers padded with zeros to the left.
- Adjusted the e2e tests as per the above fix.

### Screenshot:

image


sources/packages/backend/libs/utilities/src/esdc-utils.ts
```diff
@@ -29,6 +29,15 @@ export function getGenderCode(gender: string): string {
   return gender === "female" ? "F" : "O";
 }
 
+/**
+ * Returns phone number removing the hyphens.
+ * @param phone phone number to be formatted
+ * @returns phone number removing the hyphens in between. (Eg: 123-456-7890 => 1234567890)
+ */
+export function getFormattedPhone(phone: string): string {
```
guru-aot, line 37: 👍 Just another way of doing it would be 

export function getFormattedPhone(phone: string): string {
  return phone.split("-").join("");
}

guru-aot (approved):
LGTM, nice work @sh16011993

lewischen-aot (approved):
LGTM 👍

sources/packages/backend/libs/utilities/src/esdc-utils.ts
```diff
@@ -29,6 +29,15 @@ export function getGenderCode(gender: string): string {
   return gender === "female" ? "F" : "O";
 }
 
+/**
+ * Returns phone number removing the hyphens.
+ * @param phone phone number to be formatted
+ * @returns phone number removing the hyphens in between. (Eg: 123-456-7890 => 1234567890)
```
andrewsignori-aot, line 35: Maybe I am missing something but the task was not about removing possible dashes, it was about scanning a string and keeping only a number.
For instance `+1 (788) 999-7777` would be converted to `17889997777`.
Similarly, `####4*****^%$#@$^9` would be converted to `49`.

sources/packages/backend/libs/integrations/src/esdc-integration/msfaa-integration/msfaa-files/msfaa-file-detail.ts
```diff
@@ -54,7 +54,7 @@ export class MSFAAFileDetail implements MSFAARequestFileLine {
     record.appendWithEndFiller(this.provinceState ?? "", 4, SPACE_FILLER);
     record.appendWithEndFiller(this.postalCode ?? "", 16, SPACE_FILLER);
     record.appendWithEndFiller(this.country, 20, SPACE_FILLER);
-    record.appendWithEndFiller(this.phone ?? "", 20, SPACE_FILLER);
+    record.appendWithStartFiller(this.phone ?? "", 20, NUMBER_FILLER);
```
andrewsignori-aot, line 57: `phone` can be declared as a `number`.

sh16011993: `NUMBER_FILLER` is a string. So, wouldn't it be better to keep it as a string. Plus, the problem of number overflowing (number > `Number.MAX_VALUE`) won't be there if we keep it as string. Am I missing something here?

andrewsignori-aot: About `NUMBER_FILLER` being a string it does not justify at all why the `phone` cannot be a number.
About the second part, the max safe integer would be something around `9007199254740991` which is pretty big for a phone number. Since it is a JSONB and it would be allowed in DB, for this reason and only for this reason I am ok to keep it as string.

sources/packages/backend/libs/utilities/src/esdc-utils.ts
```diff
@@ -29,6 +29,15 @@ export function getGenderCode(gender: string): string {
   return gender === "female" ? "F" : "O";
 }
 
+/**
+ * Returns phone number keeping only digits 0 to 9.
+ * @param phone phone number to be formatted.
+ * @returns the phone number. (Eg: +1 (788) 999-7777 => 17889997777)
```
andrewsignori-aot, line 35: As a sentence I believe that it should be written as 
```
the phone number (e.g. +1 (788) 999-7777 => 17889997777).
```

andrewsignori-aot (approved):
Thanks for making the changes, look good 👍

sources/packages/backend/libs/utilities/src/esdc-utils.ts
sh16011993: Changed it.

Issue #2580, opened 2023-12-11
Data Inventory report analysis - next step (SPRINT PRIORITY)


andrepestana-aot:
### Considerations:
- The query as it is now is filtering applications that are not in `Overwritten` status. 
- All the column names should be revised by the business to be clear for the user. E.g.: 
  - parent1 x parent2 should have a prefix or something in the title to distinct parent1 from parent2. 
  - Titles must have max of 63 characters. Some titles were truncated.
- Change sub selects in the `select` statement of the query to be inner join when possible.
- Titles cannot contain commas if the ticket #2589 is not closed.
- There was a really poor performance on DEV env that should be investigated and improved.

### Acceptance Criteria
- [x] Always return the raw data without conversion 
- [x] Change report to bring the whole history for the applications (all applications records);
- [x] Revise the columns titles with business;
- [x] Change sub selects in the `select` statement of the query to be inner join when possible;
- [x] Change the query to return overwritten applications. Analyze the impact for the columns and change it to bring the whole history for the applications (all applications records). See [comment](https://github.com/bcgov/SIMS/pull/2579#discussion_r1424607276).
- [x] If the ticket #2589 is not closed, revise with business the the column titles with commas;
- [x] Work on the "Missing Columns in previous analysis" described below;
- [ ] (NICE TO HAVE) Investigate and fix the performance on DEV (maybe after the changes in this ticket it won't be an issue). If not achievable create another ticket with outstanding columns 

### Missing Columns in previous analysis

|  Column name |  Description  |  Maybe this |
|---|---|---|
| Youth in Care beyond age 19 Flag | There's already a "Youth in Care Flag" column. What's the difference between them?  | maybe one of these? <img width="1402" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBODVVQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--769345b583314f18a5c5ca3f02f0888d768ad9b7/image.png" alt="image.png" /> | 
[...]

PR #2974, opened 2024-03-08
#2580 - Data Inventory report analysis - next step

andrepestana-aot:
The spreadsheet for the report is in DEVS --> Files --> Reports.

## In this PR:

- Analyzed most of the columns sources/joins; Some of the outstanding columns still require some analysis and some require the data to be saved in the DB as it is calculated by the workflow. Some of the columns were moved to "Nice to have" and some were removed as per discussions with the business analysts.
- Changed some names to better describe or fit in 63 characters and updated the spreadsheet;
- Changed query to select all applications records (Overwritten included); 
- Performance has improved as some subqueries were moved to joins;
- Tried using crosstab to create pivot tables. That had a good performance on DEV DB but could not use on my local setup as it used 100% of disk for a long time and didn't bring any result for after more than 5 minutes. That requires more investigation to understand why it didn't work on our local env.

## Query using crosstab

Here's the version using crosstab: [crosstab_version.sql.txt](https://github.com/bcgov/SIMS/files/14544045/crosstab_version.sql.txt)

### crosstab:
Good example: https://stackoverflow.com/questions/49051959/postgresql-crosstab-simple-example

That is an example of pivot table generated with crosstab:

```sql
SELECT *
FROM crosstab(
	$$SELECT 
		ds.student_assessment_id,
		vc.value_code,
		coalesce(sum(dv.value_amount), 0) AS value_amount
	FROM 
		(VALUES ('BCAG'),('BCSL'),('BGPD'),('CSGD'),('CSGF'),('CSGP'),('CSGT'),('CSLF'),('CSLP'),('SBSD')) AS vc(value_code)
	CROSS JOIN 
		(SELECT id, student_assessment_id FROM sims.disbursement_schedules) AS ds
	LEFT JOIN 
		sims.disbursement_values dv ON vc.value_code = dv.value_code 
							   AND ds.id = dv.disbursement_schedule_id
	group by ds.student_assessment_id, vc.value_code;$$
) AS ct(student_assessment_id INT, BCAG numeric, BCSL numeric, BGPD numeric, CSGD numeric, CSGF numeric, CSGP numeric, CSGT numeric, CSLF numeric, CSLP numeric, SBSD numeric)
``` 
image

### with

Keeps a query as a temporary table in memory. 

[See more.](https://www.postgresql.org/docs/current/queries-with.html)


sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+            parent2.contact_info -> ''address'' ->> ''postalCode'' as "Parent 2 postal code",
+            parent2.supporting_data ->> ''totalIncome'' as "Parent 2 income (line 15000)",
+            current_assessment.workflow_data -> ''calculatedData'' ->> ''totalParentNetIncome'' as "Parent 2 net annual income (calculated value)",
+            parent2.supporting_data ->> ''cppLine30800'' as "Parent 2 Canada pension plan contribution (line 30800)",
+            parent2.supporting_data ->> ''cppLine31000'' as "Parent 2 Canada pension plan contribution (line 31000)",
+            parent2.supporting_data ->> ''eiLine31200'' as "Parent 2 Employment Insurance Premium (Line 31200)",
+            parent2.supporting_data ->> ''totalIncomeTaxLine43500'' as "Parent 2 Income tax paid (Line 43500)",
+            parent2.supporting_data ->> ''parentalContributions'' as "Parent 2 contribution to student",
+            parent2.supporting_data ->> ''parentsOtherDependants'' as "Parent 2 dependants other than applicant",
+            jsonb_array_length(parent1.supporting_data -> ''parentDependentTable'') as "Parent 2 number of other dependants",
```
andrewsignori-aot, line 96: Is `parent1` here supposed to be `parent2`?

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+            parent2_dependant_number_post_sec_school.count as "Parent 2 number of dependants on post secondary school",
+            parent2_dependant_number_disability.count as "Parent 2 number of dependants with disability",
+            -- Spouse Info
```
andrewsignori-aot, line 99: @HRAGANBC @CarlyCotton @Joshua-Lakusta I believe that we started to use "Partner" instead "Spouse" and it must be followed everywhere, right?
If the spreadsheet has it as a "Spouse" I would say that it should be updated also. Does it make sense?
@andrepestana-aot this is not a blocker but something that may need to be adjusted for upcoming efforts if business confirms the convention.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+            partner.supporting_data ->> ''willYourPartnerBeInReceiptOfEmploymentInsuranceBenefitsDuringTheYourStudyPeriodPanel'' as "Partner Employment Insurance amount",
+            partner.supporting_data ->> ''partnerTotalPDBenefits'' as "Partner disability benefits amount",
+            partner.supporting_data ->> ''willYourPartnerBePayingACanadaStudentLoanAndOrAProvincialStudentLoanRegularScheduledPaymentsDuringTheYourStudyPeriodPanel'' as "Partner Canada Student Loan amount",
+            partner.supporting_data ->> ''partnerTotalSupport'' as "Partner child support amount",
+            -- Partner Info Self-Reported by Student
+            applications.data ->> ''estimatedSpouseIncome'' as "Partner estimated income self-reported",
+            -- Family Info
+            current_assessment.workflow_data -> ''calculatedData'' ->> ''familySize'' as "Family size",
+            jsonb_array_length(applications.data -> ''dependants'') as "Number of dependant children in PSE",
+            dependant_11_yo_or_disabled.count as "Number of Dep (0-11 y.o. and disabled 12 y.o. or older)",
```
andrewsignori-aot, line 118: `dependant_11_yo_or_disabled` and `dependant_11_yo_or_disabled` are calculated by Camunda and should be consumed from there. If not saved yet should start to be saved and consumed here.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+            ) as dependant_12_yo_or_other on dependant_12_yo_or_other.application_id = applications.id
+            left join sims.disbursement_schedules disbursement_schedules on disbursement_schedules.student_assessment_id = current_assessment.id
+           	left join (
+            	select su.id, count(1)
+				from
+					sims.supporting_users su,
+					jsonb_array_elements(su.supporting_data->''parentDependentTable'') b
+				where (b->>''attendingPostSecondarySchool'')::text = ''yes''
+				group by su.id
+            ) as parent1_dependant_number_post_sec_school on parent1_dependant_number_post_sec_school.id = parent1.id
```
andrewsignori-aot, line 335: Some variables like `parent1_dependant_number_post_sec_school` will be likely calculated by Camunda. If not, it would be good to have Camunda inspect the data and execute any dynamic data-related calculations.
It can have it captured in the ticket and planned for an upcoming effort.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+            and cra_income_verifications.supporting_user_id is null
+            left join (
+                select
+                    *
+                from
+                    sims.supporting_users supporting_users
+                where
+                    supporting_users.supporting_user_type = ''Parent''
+                order by
+                    supporting_users.created_at asc
```
andrewsignori-aot, line 235: `created_at` may work for the order by if `id` is used it will be guaranteed.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+            left join sims.education_programs_offerings education_programs_offerings on current_assessment.offering_id = education_programs_offerings.id
+            left join sims.student_scholastic_standings student_scholastic_standings on applications.id = student_scholastic_standings.application_id
+            left join (
+            	select sr.student_id, ''Yes'' as current_restriction
+            	from sims.student_restrictions sr
+            	where sr.is_active = true
+            	limit 1
+            ) as current_student_restrictions on current_student_restrictions.student_id = students.id
+            left join (
+            	select sr.student_id, ''Yes'' as previous_restriction
```
andrewsignori-aot, line 270: For my understanding, there was some performance benefit using the `left join` instead of a `select exists` in the main `select` of the query, is it right?

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
```diff
@@ -0,0 +1,527 @@
...
+				group by a.id
+			) as award_disbursements_csgf on award_disbursements_csgf.application_id = applications.id
+			left join (
+				select a.id as application_id, sum(dv.effective_amount) as effective_amount
+				from sims.disbursement_schedules ds, sims.disbursement_values dv, sims.applications a, sims.student_assessments sa
+				where dv.disbursement_schedule_id = ds.id
+				and ds.student_assessment_id = sa.id
+				and sa.application_id = a.id
+				and dv.value_code = ''CSGT''
+				group by a.id
```
andrewsignori-aot, line 491: Was there a reason to use the join clauses as where conditions for these queries?
I would prefer to keep the same style from the other queries, unless I am missing something here.

andrewsignori-aot (commented):
Nice work with the analysis, please take a look at the comments.
I left some comments in the spreadsheet in a new column `Dev Comments for Review`.
The comments that do not start with `[DEV]` are targeting business or the upcoming discussions for upcoming tickets.

guru-aot (commented):
I am good with the PR, please check other Devs comments

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
andrepestana-aot: I removed from the code and updated the spreadsheet.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
andrewsignori-aot: To be clear, I would consider the decision and implementation outside the scope of this PR, IMO.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
andrewsignori-aot: To be clear, I would consider the decision and implementation outside the scope of this PR, IMO.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
andrepestana-aot: No, there's no especial reason. Agree with you about the style. I just couldn't work on that.

andrewsignori-aot (approved):
Thanks for taking this forward and updating the spreadsheet.
Approving based on the understanding that the goal for the ticket in the current sprint is achieved.

sources/packages/backend/apps/db-migrations/src/sql/Reports/Update-data-inventory-report.sql
andrepestana-aot: Not a performance benefit. It could be a "select exists" if it were in the select part but as we are moving from subqueries in the select part to left joins, that's the way I found to do the same in a join.

guru-aot (approved):

lewischen-aot (approved):
LGTM 👍 Nice work @andrepestana-aot

PR #2989, opened 2024-03-12
#2685 - Import Provincial Restrictions from SFAS to SIMS (Bug Fix)

sh16011993:

**Bug:** After student account creation, when the `sfas-integration` scheduler runs, it calls the `insertStudentRestrictions` method which ensures that the restrictions imported for the newly created student, if any, from SFAS are inserted into the SIMS student restrictions. However, as per the current code, the `insertStudentRestrictions` method is only called when there is atleast one SFAS file to process.

**Fix:** Code has been refactored to ensure that this block of code runs even when there are no SFAS file imports to be processed.

**Note:** There is some additional code cleanup in the file `sfas-individual.service.ts`.

### **Screenshot:**

image

### DIFF
--- a/sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.models.ts
+++ b/sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.models.ts
@@ -17,4 +17,5 @@ export class DownloadResult {
 export class ProcessSftpResponseResult {
   summary: string[] = [];
   success: boolean;
+  error: string[] = [];
 }
--- a/sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.models.ts
+++ b/sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.models.ts
@@ -17,5 +17,4 @@ export class DownloadResult {
 export class ProcessSftpResponseResult {
   summary: string[] = [];
   success: boolean;
-  error: string[] = [];
 }
--- a/sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
+++ b/sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
@@ -54,7 +54,28 @@ export class SFASIntegrationProcessingService {
         break;
       }
     }
-
+    // The following sequence of actions take place after the files have been imported and processed.
+    // The below steps need to run even if there were 0 files imported. For instance, in case of a new
+    // student account creation, when this scheduler is run again, even though there are no files to process,
+    // the insertStudentRestrictions method below will be responsible for inserting the restrictions for this
+    // student previously imported from SFAS.
+    this.logger.log("Updating student ids for SFAS individuals.");
+    await this.sfasIndividualService.updateStudentId();
+    this.logger.log("Student ids updated.");
+    this.logger.log(
+      "Updating and inserting new disbursement overaward balances from sfas to disbursement overawards table.",
+    );
+    await this.sfasIndividualService.updateDisbursementOverawards();
+    this.logger.log(
+      "New disbursement overaward balances inserted to disbursement overawards table.",
+    );
+    this.logger.log(
+      "Inserting student restrictions from SFAS restrictions data.",
+    );
+    await this.sfasRestrictionService.insertStudentRestrictions();
+    this.logger.log(
+      "Inserted student restrictions from SFAS restrictions data.",
+    );
     return results;
   }
 
@@ -138,23 +159,6 @@ export class SFASIntegrationProcessingService {
       }
       await Promise.all(promises);
       this.logger.log("Records imported.");
-      this.logger.log("Updating student ids for SFAS individuals.");
-      await this.sfasIndividualService.updateStudentId();
-      this.logger.log("Student ids updated.");
-      this.logger.log(
-        "Updating and inserting new disbursement overaward balances from sfas to disbursement overawards table.",
-      );
-      await this.sfasIndividualService.updateDisbursementOverawards();
-      this.logger.log(
-        "New disbursement overaward balances inserted to disbursement overawards table.",
-      );
-      this.logger.log(
-        "Inserting student restrictions from SFAS restrictions data.",
-      );
-      await this.sfasRestrictionService.insertStudentRestrictions();
-      this.logger.log(
-        "Inserted student restrictions from SFAS restrictions data.",
-      );
       if (result.success) {
         /**
          * Delete the file only if it was processed with success.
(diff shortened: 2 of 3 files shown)


sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -54,7 +54,28 @@ export class SFASIntegrationProcessingService {
...
+    await this.sfasIndividualService.updateDisbursementOverawards();
+    this.logger.log(
+      "New disbursement overaward balances inserted to disbursement overawards table.",
+    );
+    this.logger.log(
+      "Inserting student restrictions from SFAS restrictions data.",
+    );
+    await this.sfasRestrictionService.insertStudentRestrictions();
+    this.logger.log(
+      "Inserted student restrictions from SFAS restrictions data.",
```
andrepestana-aot, line 77: I think we should have a "result" for this part of the process, right?

andrepestana-aot (commented):
Good catch! Added a comment.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
andrewsignori-aot: Right after line 57 the below can be created and replace the `this.logger`
```ts
const postFileImportResult = new ProcessSftpResponseResult();
results.push(postFileImportResult);
result.summary.push("Updating student ids for SFAS individuals.");
...
```
And we can also have a catch as below.
```ts
 } catch (error) {
  const logMessage = `Error while processing SFAS integration file: ${remoteFilePath}`;
  result.summary.push(logMessage);
  result.success = false;
  this.logger.error(logMessage);
  this.logger.error(error);
}
```

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -54,7 +54,28 @@ export class SFASIntegrationProcessingService {
         break;
       }
     }
-
+    // The following sequence of actions take place after the files have been imported and processed.
+    // The below steps need to run even if there were 0 files imported. For instance, in case of a new
+    // student account creation, when this scheduler is run again, even though there are no files to process,
+    // the insertStudentRestrictions method below will be responsible for inserting the restrictions for this
+    // student previously imported from SFAS.
+    this.logger.log("Updating student ids for SFAS individuals.");
```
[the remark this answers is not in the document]
andrewsignori-aot: It was raised as an option during the conversation @dheepak-aot.
There will be one addition to the "post-file-import" that will prevent the overawards execution if no files are imported what would be coupling the methods a bit more.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
sh16011993: @andrewsignori-aot I have adjusted the `catch` as per our discussion.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.models.ts
```diff
@@ -17,4 +17,5 @@ export class DownloadResult {
 export class ProcessSftpResponseResult {
   summary: string[] = [];
   success: boolean;
+  error: string[] = [];
```
andrewsignori-aot, line 20: I would refrain from improving this interface since we have another process summary that would replace this one.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -54,7 +54,47 @@ export class SFASIntegrationProcessingService {
...
+      );
+      postFileImportResult.success = true;
+    } catch (error) {
+      const logMessage =
+        "Error while wrapping up post file processing operations.";
+      postFileImportResult.success = false;
+      postFileImportResult.summary.push(logMessage);
+      this.logger.log(logMessage);
+      this.logger.error(error);
+      postFileImportResult.error.push(error);
```
andrewsignori-aot, line 94: The error as `false` and line 91 should be enough, IMO we can remove this `error.push` and undo the interface change.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -54,7 +54,47 @@ export class SFASIntegrationProcessingService {
         break;
       }
     }
-
+    // The following sequence of actions take place after the files have been imported and processed.
+    // The below steps need to run even if there were 0 files imported. For instance, in case of a new
+    // student account creation, when this scheduler is run again, even though there are no files to process,
```
andrewsignori-aot, line 59: Please adjust the "when this scheduler is run again".

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
```diff
@@ -54,7 +54,47 @@ export class SFASIntegrationProcessingService {
...
+    } catch (error) {
+      const logMessage =
+        "Error while wrapping up post file processing operations.";
+      postFileImportResult.success = false;
+      postFileImportResult.summary.push(logMessage);
+      this.logger.log(logMessage);
+      this.logger.error(error);
+      postFileImportResult.error.push(error);
+    } finally {
+      results.push(postFileImportResult);
```
andrewsignori-aot, line 96: No need for a `finally` block if this code is moved right after line 62.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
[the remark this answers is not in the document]
andrewsignori-aot: To be honest, after the last refactor @sh16011993 I would be leaning also towards creating a separate method.
At this stage, I would not consider it a blocker.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.processing.service.ts
sh16011993: I think you meant success as `false`?

andrewsignori-aot (approved):

dheepak-aot (approved):
Looks good @sh16011993 👍

andrepestana-aot (approved):
Looks good! Thanks for the changes!

Issue #2645, opened 2024-01-08
Institution/Ministry - Deactivate Program - UI Changes


HRAGANBC:
### **User Story**
As an _Institution or Ministry User_, I may need to deactivate a program. This could be due to no longer offering it, or creating it in error resulting in a new revised program being created.

### **Acceptance Criteria**
- [x] Add new state for programs: Inactive (preserve current status at time of deactivation)
- [x] If the state is inactive, update status flag for programs to show the "Inactive" flag in dashboards. 
- [x] Ministry User: Update edit button on Program details screen to drop down with two options: Edit Program, Deactivate. (No change to functionality of "Edit Program")
- [x] Make this feature/role available to **all** Groups within Ministry KeyCloak configs
- [x] Institution User: Update edit button on Program details screen to drop down with two options: Edit Program, Deactivate. (No change to functionality of "Edit Program")
- [x] New pop up when user selects "Deactivate" from edit program details button
- [x] When the Ministry is deactivating, include a basic `Notes` section (not part of the mockup) to save a note to the institution's notes under the `Program` category.
- [x] Update new program creation validation to ignore any deactivated programs. (i.e. so that the same SABC Program Code can be re-used on the new Programs)
- [x] Update bulk upload to ignore deactivated programs when assigning offerings.    

**Technical**

- [x] Create a new column `sims.education_programs.is_active` to represent the inactive state of the program.
- [x] Create new specific audit columns to save the user deactivating and the date, for instance, `deactivated_by`, and `deactivation_date` (we may create columns also as `is_active_updated_on` and `is_active_updated_by`).
- [x] Adjust the badge status component on the web portal to display `Inactive` when the `is_active` is false, otherwise, display the status as it is.
- [x] Create a new role for the Ministry approval action. The new role should be associated with all groups as per stated in the business AC.
- [x] Create API E2E tests for the new endpoints.

### **Mock ups**
**Update to "Edit" button on Program Details to a Drop-down selection with new pop-up modal:**
<img width="682" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBME10QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--70ade47a3a4d48385754b84b3d45b085fb45f971/image.png" alt="image.png" />      
### **Current Behaviour**
Programs exist _forever_ and institution users just update the program name/description to indicate it should not be used.
[...]

PR #3004, opened 2024-03-15
#2645 - Program Deactivation - DB migrations

andrewsignori-aot:
Created the new columns related to the education program deactivation. Used one of the suggested name combinations in the ticket technical discussion.
- isActive
- isActiveUpdatedBy
- isActiveUpdatedOn

_Minor comment, we did not plan to add a specific note id associated with the is_active column. The note requested will be created without it and it was my understanding that it would not be required. Please let me know if it need to be discussed._

_Minor update in the shared workspace `cSpell.words`._

### DIFF
--- a/sims.code-workspace
+++ b/sims.code-workspace
@@ -89,6 +89,14 @@
       }
     },
     "typescript.preferences.importModuleSpecifier": "non-relative",
-    "cSpell.words": ["ESDC", "MSFAA", "NSLSC", "Overaward", "SFAS", "typeorm"]
+    "cSpell.words": [
+      "ESDC",
+      "MSFAA",
+      "NSLSC",
+      "Overaward",
+      "SFAS",
+      "timestamptz",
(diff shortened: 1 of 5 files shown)


sims.code-workspace
```diff
@@ -89,6 +89,14 @@
       }
     },
     "typescript.preferences.importModuleSpecifier": "non-relative",
-    "cSpell.words": ["ESDC", "MSFAA", "NSLSC", "Overaward", "SFAS", "typeorm"]
+    "cSpell.words": [
```
guru-aot, line 92: ❤️

lewischen-aot (approved):
LGTM 👍 Nice work @andrewsignori-aot. Thanks for setting up the DB migrations in advance.

dheepak-aot (approved):
Looks good. 👍Thanks for rolling out the migrations early

Issue #2702, opened 2024-01-18
Fix dependabot security issues


andrewsignori-aot:
Verify and fix dependabot security issues (https://github.com/bcgov/SIMS/security/dependabot).

<img width="761" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3JiQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--028c42c94de00881d6b7f06563a724d928141bc8/image.png" alt="image.png" />

**Acceptance criteria**

- [x] Check the most updated list of security issues on GitHub: https://github.com/bcgov/SIMS/security/dependabot
- [x] High issues must be fixed or justified.
- [ ] Medium issues must be cleaned as must a possible. If any medium issue requires further analysis or extended work please create new tickets accordingly.

PR #3015, opened 2024-03-18
#2702 - Nest upgrade for dependabot fix

dheepak-aot:
## Upgrades for the High severity dependabot issues

Upgrade of several packages in backend has been done to address the high severity dependabot issue #2629

- [x] Upgraded Bull(Including bull board) and Nest JS to the latest version
- [x] Upgraded `@golevelup` packages
- [x] Upgraded the library `swagger-ui-express` to latest version
- [x] Package lock file was re-created,  and hence many of the libraries are expected to be updated to their latest minor version.

### Code changes to support the upgrade

- [x] In date-utils, the type of unit to be added to a given date is modified to `dayjs.ManipulateType` from `dayjs.OpUnitType` as per the recent library version.
- [x] For the Zeebe client, the  `createProcessInstance` method was updated to have a single object parameter and code has been updated accordingly. 
NOTE: The e2e tests use the method `createProcessInstanceWithResult` which were already following the similar format of input.

### Blocker issue in `test-db-seeding` and `load-test-gateway`
 During the implementation of the ticket, we had a blocker issue due to the usage of `@golevelup/ts-jest` library. The issue is that this library imports `@jest/globals` in it which  is not suppose to be imported outside the Jest test context. 
i.e. not to be imported in a code which is not run by jest command. 
Note: All our E2E tests are run by jest and hence it is not a problem for E2E tests.
e.g. API e2e test is run by `cross-env ENVIRONMENT=test jest --collect-coverage --verbose --config ./apps/api/test/jest-e2e.json --forceExit`

The issue now is with `test-db-seeding` and `load-test-gateway` which used `test-utils` library but they are not run by the jest command, instead they are run as regular node application.

DB Seeding run command: `node dist/apps/test-db-seeding/main.js`

 - [x] **Solution:** Since `test-db-seeding` and `load-test-gateway` use the factory methods from `@sims/test-utils` lib, the root index file of this library is loaded during the compile time, which loads all the *.mock.ts and hence `@golevelup/ts-jest` as it's dependency. As a solution all the mock classes and functions are moved out of test-utils/src/index.ts to test-utils/src/mock/index.ts so that the mock libraries are not compiled for `test-db-seeding` and `load-test-gateway` .

- [x] Removing deprecated faker methods: In faker, number and uuid are deprecated under random object and now available in datatype object. Code has been updated everywhere to follow this.


lewischen-aot (approved):
Nice work @dheepak-aot 👍Thanks for making the changes to accommodate the upgrades for packages and the thorough description for the PR ❤️

sources/packages/backend/libs/utilities/src/date-utils.ts
```diff
@@ -205,7 +205,7 @@ export const addHours = (hoursToAdd: number, date?: Date | string): Date => {
 export function addToDateOnlyString(
   date: string | Date,
   toAdd: number,
-  unit: dayjs.OpUnitType,
+  unit: dayjs.ManipulateType,
```
guru-aot, line 208: 👍

guru-aot (approved):
LGTM, Nice and exhaustive work @dheepak-aot

Issue #1801, opened 2023-03-14
Reorganize Workflow folders


andrewsignori-aot:
The code repository contains Camunda workflow definitions from the legacy Camunda 7 (no longer needed) and Camunda 8 (current) which often confuses. The Camunda 7 definitions and all related code can be removed.

*Acceptance criteria**
- [x] Ensure only current Camunda workflow definitions are present in the code repository.

**Technical**
- [x] Move the contents of the folder `backend/workflow/bpmn/camunda-8` to `backend/workflow/src/workflow-definitions` (new folder to be created following the same pattern from forms).
- [x] Delete folders other than `src` and `test` (e.g. BPMN, DMN, Scheduler, scripts).
- [x] Adjust the folder path in [deploy.ts](https://github.com/bcgov/SIMS/blob/b2ea8ce72fc87ae91ebdc2327e69722b4934c307/sources/packages/backend/workflow/src/deploy.ts#L26).
- [x] Test the local deployment using `make deploy-camunda-definitions` from the `sources` folder. Expected to see a "Deployment successful!" message.

PR #3024, opened 2024-03-20
#1801 - Reorganize Workflow Folders

katrina-aot:
Update Camunda workflows to version 8, removing version 7 files


sources/packages/backend/workflow/src/deploy.ts
```diff
@@ -23,7 +23,7 @@ import {
   console.info(`**** Deploying to Camunda ****\n`);
   console.info(`Deploying to Zeebe address ${process.env.ZEEBE_ADDRESS}`);
 
-  const directory = path.resolve(__dirname, `../BPMN/camunda-8`);
+  const directory = path.resolve(__dirname, `../src/workflow-definitions`);
```
andrewsignori-aot, line 26: Can you please use the path below?
```ts
path.resolve(__dirname, "./workflow-definitions")
```

andrewsignori-aot (commented):
Thanks so much for the cleanup. Just one minor comment, please take a look.

sources/packages/backend/workflow/src/deploy.ts
katrina-aot: updated

andrewsignori-aot (approved):
Thanks for making the change, looks good 👍

lewischen-aot (approved):
Thanks for making the change. LGTM 👍

sh16011993 (approved):
Nice work @katrina-aot 👍 LGTM

Issue #2536, opened 2023-11-28
Notify Student About Blocked Disbursements (Email only) PT/FT


HRAGANBC:
**User Story**
As a student, I want to be notified anytime all standard required steps have been fulfilled, yet I have still not received my funding. I want to understand if my disbursements are being withheld. and for what reason, so I can understand how to resolve. 

**Description of Task**
- Send Email notification to student when their disbursements are being withheld

### **Acceptance Criteria**
- [x] **Configure new E-mail to Student** This should get triggered only when eCert is blocked (5 days prior to study start dates).
- [x] GC Notify Email should read: Subject Line: _Your StudentAid BC Funding is Being Withheld_. Email body: _Warning - your StudentAid BC disbursement is being withheld. Please login to your account for more information about how to receive your funding_.") *New Template needed within GC Notify
- [x] **Configure new E-mail to Ministry (per grooming on #2369)** This should get triggered by the same event as the email to the student
- [x] send email to Studentaidbc@gov.bc.ca
- [x] GC Notify Email template: a0bf7cc1-b3f7-463a-a594-d9f10b704e80
- [x] Notifications should stop when application moves to archived or after 3 email notifying about blocked disbursement on specific application. 
- [x] Analysis to allow notifications to track the applications that already received an email notification at least once before 
- [x] Send new notification every 7 days after the first instance, up to maximum of 3 notifications total. 
- [x] E2E Tests

CarlyCotton:
<img width="807" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGo1QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--52d581693a54ac75ac0c86152e69ddd24c29ccc0/image.png" alt="image.png" />

@sh16011993 We'd like the date/time stamp on the notification email. It's on most of the templates, please include.

PR #3032, opened 2024-03-22
#2536 - Notify student about blocked disbursements: email only (PR 2)

sh16011993:

**Business Tasks:**
-  Analysis to allow notifications to track the applications that already received an email notification at least once before.
-  Notifications should stop being sent after 3 emails notifying about blocked disbursement on specific application.
-  Notifications are sent every 7 days after the first instance, up to maximum of 3 notifications total.

**Technical Tasks:**
- Analysis to allow count of the notifications to be tracked involved db migrations to create the `metadata` `jsonb` type column to store the `disbursement_id` related to the notification being sent.
- Code logic updated to store the metadata informationin db and to ensure the notifications are not sent after 3 emails and are sent every 7 days.


sources/packages/backend/libs/sims-db/src/entities/notification.model.ts
```diff
@@ -76,6 +77,13 @@ export class Notification extends RecordDataModel {
     nullable: true,
   })
   permanentFailureError: PermanentFailureError[];
+  /** Metadata information related to the saved notification. */
```
lewischen-aot, line 80: I thought we should use the same comment style as in line 69 - 73. In addition to this, correct me if I'm wrong @andrewsignori-aot. I thought we need to use the same comment (with a small difference) on both the model here and the SQL file above.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -55,17 +56,26 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+        creator: { id: auditUserId } as User,
+        messagePayload: notification.messagePayload,
+        notificationMessage: {
+          id: notification.messageType,
+        } as NotificationMessage,
+      };
+      if (options?.disbursementId) {
+        notificationToSave["metadata"] = {
+          disbursement_id: options?.disbursementId,
+        };
```
lewischen-aot, line 75: Minor. Maybe we can do `notificationToSave.metadata = { ... }` instead of `notificationToSave["metadata"] = { ... }`. If we do the former style, we should probably add `metadata` as optional field to the interface [`SaveNotificationModel`](https://github.com/bcgov/SIMS/blob/52dd32d09387a7b82dfc2d86df7e24c46b8550c6/sources/packages/backend/libs/services/src/notifications/notification/notification.model.ts#L4). Thoughts @dheepak-aot @andrewsignori-aot ?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+      .andWhere(
+        "notifications.metadata->>'disbursement_id' = :disbursementId",
+        {
+          disbursementId,
+        },
+      )
+      .getRawOne<NotificationData>();
+    const sevenDaysLaterDate = getUTC(addDays(7, notification.max_created_at));
+    // Condition check to create notifications: Less than 3 notifications are created previously and there are no failures in sending those created notifications.
+    // Plus, it has been atleast 7 days from the last created notification for this disbursement.
```
lewischen-aot, line 204: Minor. There is small typo "atleast".

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -128,12 +139,19 @@ export abstract class ECertCalculationProcess {
...
-                eCertDisbursement.applicationNumber,
-                entityManager,
-              );
+              const shouldCreateNotification =
+                await this.shouldCreateDisbursementBlockedNotification(
+                  eCertDisbursement.disbursement.id,
+                  entityManager,
+                );
+              if (shouldCreateNotification) {
+                await this.createDisbursementBlockedNotifications(
```
lewischen-aot, line 148: I think we should probably add back the comment "// Create notifications to the student and ministry when the disbursement is blocked." for `await this.createDisbursementBlockedNotifications`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+      })
+      .andWhere("notifications.permanent_failure_error IS NULL")
+      .andWhere(
+        "notifications.metadata->>'disbursement_id' = :disbursementId",
+        {
+          disbursementId,
+        },
+      )
+      .getRawOne<NotificationData>();
+    const sevenDaysLaterDate = getUTC(addDays(7, notification.max_created_at));
```
lewischen-aot, line 202: Should we check for whether `notification is null` and return true before declaring `sevenDaysLaterDate`?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+    // Condition check to create notifications: Less than 3 notifications are created previously and there are no failures in sending those created notifications.
+    // Plus, it has been atleast 7 days from the last created notification for this disbursement.
+    if (
+      !notification.notificationsCount ||
+      (notification.notificationsCount < 3 && sevenDaysLaterDate < getUTCNow())
```
lewischen-aot, line 207: Correct me if I'm wrong about this. I thought it should be `sevenDaysLaterDate >= getUTCNow()`.

lewischen-aot:
Nice work @sh16011993. Like the notification query ❤️ Left a few comments. Please have a look.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+   * @param disbursementId disbursement id.
+   * @param entityManager entity manager to execute in transaction.
+   * @returns boolean indicating if the disbursement blocked notification should be created or not.
+   */
+  private async shouldCreateDisbursementBlockedNotification(
+    disbursementId: number,
+    entityManager: EntityManager,
+  ): Promise<boolean> {
+    // Create notifications to the student and ministry when the disbursement is blocked.
+    const notificationMessageId = 16;
```
andrepestana-aot, line 184: We should have a reference to NotificationMessageType here, right?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
andrepestana-aot, line 202: Can we have a constant for this "7"?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
andrepestana-aot, line 207: I think we could have a constant for this "3" with a meaningful name here.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
andrepestana-aot: Also, instead of `sevenDaysLaterDate` find another more meaningful name. 
"Code logic updated to store the metadata informationin db and to ensure the notifications are not sent after 3 emails and are sent every 7 days."
I think we could call it something like `minimumDateToSendANewMessage`.

andrepestana-aot (commented):
Good job. Left some comments.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+    const notification = await entityManager
+      .getRepository(Notification)
+      .createQueryBuilder("notifications")
+      .select("count(*)::int", "notificationsCount")
```
andrewsignori-aot, line 188: The result of the count should already be returned as a number. Is there a need to have the type cast to `int`?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+      .addSelect("max(notifications.created_at)", "max_created_at")
+      .innerJoin("notifications.notificationMessage", "notificationMessage")
+      .where("notificationMessage.id = :notificationMessageId", {
+        notificationMessageId,
+      })
+      .andWhere("notifications.permanent_failure_error IS NULL")
```
andrewsignori-aot, line 194: Please use the entity column name instead of the column name itself as much as possible.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+    disbursementId: number,
+    entityManager: EntityManager,
+  ): Promise<boolean> {
+    // Create notifications to the student and ministry when the disbursement is blocked.
+    const notificationMessageId = 16;
+    const notification = await entityManager
+      .getRepository(Notification)
+      .createQueryBuilder("notifications")
+      .select("count(*)::int", "notificationsCount")
+      .addSelect("max(notifications.created_at)", "max_created_at")
```
andrewsignori-aot, line 189: Please use the proper name convention `maxCreatedAt`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
andrewsignori-aot, line 194: Can you please elaborate on why the `permanent_failure_error` was included in the conditions?

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+   * Checks whether the disbursement blocked notification should be created or not.
+   * @param disbursementId disbursement id.
+   * @param entityManager entity manager to execute in transaction.
+   * @returns boolean indicating if the disbursement blocked notification should be created or not.
+   */
+  private async shouldCreateDisbursementBlockedNotification(
+    disbursementId: number,
+    entityManager: EntityManager,
+  ): Promise<boolean> {
+    // Create notifications to the student and ministry when the disbursement is blocked.
```
andrewsignori-aot, line 183: The comment is misleading because the method is not creating the notification.
Besides that, it may be pointless to have a general comment here if the method summary is already explaining it.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
     }
   }
 
+  /**
+   * Checks whether the disbursement blocked notification should be created or not.
+   * @param disbursementId disbursement id.
+   * @param entityManager entity manager to execute in transaction.
+   * @returns boolean indicating if the disbursement blocked notification should be created or not.
+   */
+  private async shouldCreateDisbursementBlockedNotification(
```
andrewsignori-aot, line 179: The `ECertCalculationProcess` was created with the idea of keeping a centralized place for the calculation only and it grew too much because of the notifications.
I would suggest creating an `e-cert-notification-services` to keep all notification-related method now in this class.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+      .createQueryBuilder("notifications")
+      .select("count(*)::int", "notificationsCount")
+      .addSelect("max(notifications.created_at)", "max_created_at")
+      .innerJoin("notifications.notificationMessage", "notificationMessage")
+      .where("notificationMessage.id = :notificationMessageId", {
+        notificationMessageId,
+      })
+      .andWhere("notifications.permanent_failure_error IS NULL")
+      .andWhere(
+        "notifications.metadata->>'disbursement_id' = :disbursementId",
```
andrewsignori-aot, line 196: As a column name, I would agree to keep it as `disbursement_id` but here it is supposed to represent metadata. Other metadata properties can be added later and I believe that they must follow `disbursementId`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -18,6 +24,11 @@ interface GroupedDisbursementPerStudent {
   [studentId: number]: EligibleECertDisbursement[];
 }
 
+interface NotificationData {
+  max_created_at: Date;
```
andrewsignori-aot, line 28: The other suggested refactor should make the name aligned with the convention `maxCreatedAt`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -128,12 +139,19 @@ export abstract class ECertCalculationProcess {
               disbursementLog,
             );
             if (!shouldProceed) {
-              // Create notifications to the student and ministry when the disbursement is blocked.
-              await this.createDisbursementBlockedNotifications(
-                eCertDisbursement.studentId,
-                eCertDisbursement.applicationNumber,
-                entityManager,
-              );
+              const shouldCreateNotification =
```
andrewsignori-aot, line 142: When doing the below suggestion, there will just be one method call here to the new service.
https://github.com/bcgov/SIMS/pull/3032/files#r1536263389

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+   */
+  private async shouldCreateDisbursementBlockedNotification(
+    disbursementId: number,
+    entityManager: EntityManager,
+  ): Promise<boolean> {
+    // Create notifications to the student and ministry when the disbursement is blocked.
+    const notificationMessageId = 16;
+    const notification = await entityManager
+      .getRepository(Notification)
+      .createQueryBuilder("notifications")
```
andrewsignori-aot, line 187: Please keep `notifications` singular. Just a few rare places along the source code use it as plural, unless I am missing something.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
andrewsignori-aot: @sh16011993 the constant can be created in the below file:
libs\services\src\constants\system-configurations-constants.ts

andrewsignori-aot, line 202: No need to use `getUTC`. Servers and localhost should be configured to run in UTC already.
This is an old method that no longer should be used.

andrewsignori-aot, line 202: The logic can be changed as shown below.
```ts
const canSendNewNotification = dateDifference(new Date(), notification.maxCreatedAt) > BLOCKED_DISBURSEMENT_NOTIFICATION_MIN_DAYS_INTERVAL;
```

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -55,17 +56,26 @@ export class NotificationService extends RecordDataModelService<Notification> {
     notifications: SaveNotificationModel[],
     auditUserId: number,
     options?: {
-      entityManager: EntityManager;
+      disbursementId?: number;
```
andrewsignori-aot, line 59: This method is generic to save any notification. It should receive the `metadata` object to be saved.

andrewsignori-aot, line 59: Is the metadata not supposed to be in `SaveNotificationModel`?

sources/packages/backend/libs/sims-db/src/entities/notification-metadata.type.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Metadata for notification.
+ */
+export interface Metadata {
+  disbursement_id: number;
```
andrewsignori-aot, line 5: As metadata, I would suggest having it defined as `disbursementId`. 
In the context where it is it should be ok to call it only `Metadata` but I am afraid that it will be too generic.
Can we rename it as `NotificationMetadata`?

sources/packages/backend/libs/sims-db/src/entities/notification.model.ts
```diff
@@ -76,6 +77,13 @@ export class Notification extends RecordDataModel {
...
+  @Column({
+    name: "metadata",
+    type: "jsonb",
+    nullable: true,
+  })
+  metadata: Metadata;
```
andrewsignori-aot, line 86: The type declaration should be nullable also.(`metadata?: Metadata;`).

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -152,13 +170,56 @@ export abstract class ECertCalculationProcess {
...
+    // Condition check to create notifications: Less than 3 notifications are created previously and there are no failures in sending those created notifications.
+    // Plus, it has been atleast 7 days from the last created notification for this disbursement.
+    if (
+      !notification.notificationsCount ||
+      (notification.notificationsCount < 3 && sevenDaysLaterDate < getUTCNow())
+    ) {
+      return true;
+    }
+    return false;
+  }
```
andrewsignori-aot, line 212: Can this be simplified as below?
```ts
return !notification.notificationsCount ||  (notification.notificationsCount < 3 && sevenDaysLaterDate < getUTCNow());
```

sources/packages/backend/apps/db-migrations/src/sql/Notifications/Add-metadata-column.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  sims.notifications
+ADD
+  COLUMN metadata jsonb;
+
+COMMENT ON COLUMN sims.notifications.metadata IS 'Metadata information on notifications sent such as tracking the various relations associated with the sent notification.';
```
andrewsignori-aot, line 6: I would prefer to not indicate the relationships in the comment as it can mislead someone to believe that the intention is so save DB relationships only here.
Suggestion:
```
Metadata information for a notification to allow a more granular identification or identify further the reason why it was generated.
```

andrewsignori-aot (commented):

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
lewischen-aot: `sevenDaysLaterDate < getUTCNow()` is the correct behavior.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
sh16011993: Yes, I did a cast because it was returning the count as a `string`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
sh16011993: I added it because if a notification was not sent because of a `permanent_failure_error`, it should not be counted towards the 3 total max. notifications we would send. Otherwise, it might happen that a notification generated `permanent_failure_error` thrice and hence, a notification never got sent out.

andrewsignori-aot: `permanent_failure_error` means something that cannot be recovered from. If it falls into that category we should not keep generating other messages.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
```diff
@@ -25,7 +20,7 @@ interface GroupedDisbursementPerStudent {
 export abstract class ECertCalculationProcess {
   constructor(
     private readonly dataSource: DataSource,
-    private readonly notificationActionsService: NotificationActionsService,
+    private readonly ecertNotificationService: ECertNotificationService,
```
andrewsignori-aot, line 23: Please rename it to `eCertNotificationService`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/full-time-calculation-process.ts
```diff
@@ -24,7 +24,7 @@ import { NotificationActionsService } from "@sims/services";
 export class FullTimeCalculationProcess extends ECertCalculationProcess {
   constructor(
     dataSource: DataSource,
-    notificationActionsService: NotificationActionsService,
+    ecertNotificationService: ECertNotificationService,
```
andrewsignori-aot, line 27: Please rename it to `eCertNotificationService`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/part-time-calculation-process.ts
```diff
@@ -22,7 +22,7 @@ import { NotificationActionsService } from "@sims/services";
 export class PartTimeCalculationProcess extends ECertCalculationProcess {
   constructor(
     dataSource: DataSource,
-    notificationActionsService: NotificationActionsService,
+    ecertNotificationService: ECertNotificationService,
```
andrewsignori-aot, line 25: Please rename it to `eCertNotificationService`.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
```diff
@@ -0,0 +1,145 @@
...
+      .where("notificationMessage.id = :notificationMessageId", {
+        notificationMessageId,
+      })
+      .andWhere("notification.permanentFailureError IS NULL")
+      .andWhere("notification.metadata->>'disbursementId' = :disbursementId", {
+        disbursementId,
+      })
+      .getRawOne<NotificationData>();
+    const canSendNewNotification =
+      dateDifference(new Date(), notification.maxCreatedAt) >
```
andrewsignori-aot, line 81: `dateDifference` considers a "+1" already. Maybe the comparison here should use ">="
```ts
dateDifference(new Date(), notification.maxCreatedAt) >= BLOCKED_DISBURSEMENT_NOTIFICATION_MIN_DAYS_INTERVAL;
```

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
```diff
@@ -0,0 +1,145 @@
...
+   * @param studentId student id associated with the blocked disbursement.
+   * @param applicationNumber application number associated with the blocked disbursement.
+   * @param entityManager entity manager to execute in transaction.
+   */
+  private async createDisbursementBlockedNotifications(
+    disbursementId: number,
+    studentId: number,
+    applicationNumber: string,
+    entityManager: EntityManager,
+  ) {
```
andrewsignori-aot, line 105: Missing return type.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * Creates disbursement blocked notifications for student and ministry.
+   * @param disbursementId disbursement id associated with the blocked disbursement.
+   * @param studentId student id associated with the blocked disbursement.
+   * @param applicationNumber application number associated with the blocked disbursement.
+   * @param entityManager entity manager to execute in transaction.
+   */
+  private async createDisbursementBlockedNotifications(
+    disbursementId: number,
```
andrewsignori-aot, line 101: Maybe missed in the previous review but `disbursementId` should be enough to get student data and application data, which means that there is no need to provide `studentId` and `applicationNumber` as a parameter.

sources/packages/backend/libs/services/src/constants/notification.constants.ts
```diff
@@ -0,0 +1,4 @@
+// Represents the minimum number of days before another notification for a blocked disbursement will be sent out.
```
andrewsignori-aot, line 1: Please adjust both comments.

sources/packages/backend/libs/services/src/notifications/notification/notification.model.ts
```diff
@@ -1,10 +1,15 @@
 import { NotificationMessageType } from "@sims/sims-db";
 import { NotificationEmailMessage } from "./gc-notify.model";
 
+interface NotificationMetadata {
+  disbursementId: number;
```
andrewsignori-aot, line 5: Since `NotificationMetadata` is supposed to map any future metadata, `disbursementId` would probably needed to nullable.
Not a blocker right now I would recommend changing it.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -25,13 +25,22 @@ import {
...
 import { GC_NOTIFY_PERMANENT_FAILURE_ERROR } from "@sims/services/constants";
+import { NotificationMetadata } from "@sims/sims-db/entities/notification-metadata.type";
 
 /**
  * While performing a possible huge amount of inserts,
  * breaks the execution in chunks.
  */
 const NOTIFICATIONS_INSERT_CHUNK_SIZE = 1000;
 
+interface NotificationToSave {
```
andrewsignori-aot, line 36: Just taking a second look at this interface, can you please elaborate on why we need it?

sources/packages/backend/libs/services/src/notifications/notification/notification.model.ts
```diff
@@ -1,10 +1,15 @@
 import { NotificationMessageType } from "@sims/sims-db";
 import { NotificationEmailMessage } from "./gc-notify.model";
 
+interface NotificationMetadata {
```
andrewsignori-aot, line 4: Why is this not using the DB interface?

sources/packages/backend/libs/sims-db/src/entities/notification-metadata.type.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Metadata for notification.
+ */
+export interface NotificationMetadata {
+  disbursementId: number;
```
andrewsignori-aot, line 5: Since `NotificationMetadata` is supposed to map any future metadata, `disbursementId` would probably needed to nullable.
Not a blocker right now I would recommend changing it.

andrewsignori-aot (commented):
Thanks for doing the changes and noce refactor in the e-Cert notification service, special thanks for that 😉 
Please take a look at the remaining comments.

sources/packages/backend/libs/sims-db/src/entities/notification.model.ts
sh16011993: Sorry, missed it 😅

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
sh16011993: Moved it there.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
```diff
@@ -0,0 +1,145 @@
...
+      .andWhere("notification.permanentFailureError IS NULL")
+      .andWhere("notification.metadata->>'disbursementId' = :disbursementId", {
+        disbursementId,
+      })
+      .getRawOne<NotificationData>();
+    const canSendNewNotification =
+      dateDifference(new Date(), notification.maxCreatedAt) >=
+      BLOCKED_DISBURSEMENT_NOTIFICATION_MIN_DAYS_INTERVAL;
+    // Condition check to create notifications: Less than 3 notifications are created previously and there are no failures in sending those created notifications.
+    // Plus, it has been at least 7 days from the last created notification for this disbursement.
```
andrepestana-aot, line 84: Now  that we have the constants we should remove those hardcoded numbers from the comments.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-calculation/e-cert-calculation-process.ts
sh16011993: Removed.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
```diff
@@ -0,0 +1,157 @@
...
+                  email: true,
+                },
+              },
+            },
+          },
+        },
+        relations: {
+          studentAssessment: { application: { student: { user: true } } },
+        },
+        where: { id: disbursementId },
```
andrewsignori-aot, line 123: Does typeorm not require the ID from the main entity to create the `disbursement` object?

andrewsignori-aot (approved):

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
sh16011993: Just checked. It returns a whole bunch of extra `properties` without adding the `id` from the base table. Added it. Thanks.

lewischen-aot (approved):
LGTM 👍 Nice work @sh16011993

andrepestana-aot (approved):
Looks good!

Issue #2748, opened 2024-01-29
Changed Queues IsActive Flag Behavior 


andrewsignori-aot:
Currently, when changing a Queue configuration to not be active the configuration generates an issue while starting the POD.

**Acceptance criteria**
- [x] Setting is_active to `false` on `sims.queue_configurations` should allow the queue-consumers application to start nicely.
- [x] Change the code creating to queues to ensure that the non-active queues will not be loaded ([code reference](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/libs/services/src/queue/queue.module.ts#L75)).
- [x] If the queue is already executing we need to ensure that it will be stopped and/or removed. The Bull Dashboard will probably no longer display it.
- [x] Review the method `createBullBoard` in [queue-consumers/main.ts](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/queue-consumers/src/main.ts) used to create the Dashboard Queues are created from this `QueueNames` enum.
- [x] Check the solution applied during #2755 to possibly have the queues set to is_active as false "obliterated".
- [x] Non-schedulers queues will not be affected by the is_active flag for now.
- [ ] E2E as nice to have.

andrewsignori-aot:
Moved this ticket to "Get Dressed" as per the discussion with @ninosamson 
It would allow the deployment to queue-consumers disabling some queues directly on DB.

andrewsignori-aot:
Moved this ticket into spring 61 as per team agreement.

PR #3070, opened 2024-04-04
#2748 - Changed Queues IsActive Flag Behavior

lewischen-aot:
- Setting `is_active` to `false` on `sims.queue_configurations` allows the queue-consumers application to start nicely.
- Changed the code for creating queues and ensured that 
  - the inactive, scheduler queues will not be loaded, and
  - the Bull Dashboard will no longer display it.
- Non-schedulers queues will not be affected by the `is_active` flag for now.

Example:
- Disable queues with "part-time" prefix and "start-application-assessment" queues
  - `part-time-e-cert-integration`
  - `part-time-feedback-integration`
  - `part-time-msfaa-process-integration`
  - `part-time-msfaa-process-response-integration`
- "start-application-assessment" queue should still be running because it is a non-scheduler queue.
- All the queues with the "part-time" prefix will not be running as they are scheduler queues.

Screenshot of having queues with "part-time" prefix and "start-application-assessment" queues
 disabled:
image

Screenshot of the "start-application-assessment" queue is running on Bull Dashboard:
image

Screenshot of the queues with "part-time" prefix not displaying on Bull Dashboard:
image

Screenshot of the queues with "part-time" prefix not running (is inactive) on Docker Logs:
Screenshot 2024-04-04 101614

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/main.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/main.ts
@@ -31,7 +31,7 @@ import { SystemUsersService } from "@sims/services";
   serverAdapter.setBasePath("/admin/queues");
   const bullBoardQueues: BullAdapter[] = [];
   queues.forEach((queue) => {
-    if (!queue.isActive && queue.cron) {
+    if (!queue.isActive && queue.isScheduler) {
       logger.log(`Queue service "${queue.name}" is inactive.`);
     } else {
       bullBoardQueues.push(
--- a/sources/packages/backend/apps/queue-consumers/src/main.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/main.ts
@@ -10,6 +10,7 @@ import { QueueConsumersModule } from "./queue-consumers.module";
 import * as basicAuth from "express-basic-auth";
 import { LoggerService } from "@sims/utilities/logger";
 import { SystemUsersService } from "@sims/services";
+import { QueueNames, schedulerQueueNames } from "@sims/utilities";
 
 (async () => {
   const app = await NestFactory.create(QueueConsumersModule);
@@ -29,10 +30,20 @@ import { SystemUsersService } from "@sims/services";
   // Create bull board UI dashboard for queue management.
   const serverAdapter = new ExpressAdapter();
   serverAdapter.setBasePath("/admin/queues");
-  const bullBoardQueues = queues.map((queue) => {
-    return new BullAdapter(app.get<Queue>(`BullQueue_${queue.name}`), {
-      readOnlyMode: queue.dashboardReadonly,
-    });
+  const bullBoardQueues: BullAdapter[] = [];
+  queues.forEach((queue) => {
+    if (
+      !queue.isActive &&
+      schedulerQueueNames.includes(queue.name as QueueNames)
+    ) {
+      logger.log(`Queue service "${queue.name}" is inactive.`);
+    } else {
+      bullBoardQueues.push(
+        new BullAdapter(app.get<Queue>(`BullQueue_${queue.name}`), {
+          readOnlyMode: queue.dashboardReadonly,
+        }),
+      );
+    }
   });
   createBullBoard({
     queues: bullBoardQueues,
--- a/sources/packages/backend/apps/queue-consumers/src/main.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/main.ts
@@ -10,7 +10,6 @@ import { QueueConsumersModule } from "./queue-consumers.module";
 import * as basicAuth from "express-basic-auth";
 import { LoggerService } from "@sims/utilities/logger";
 import { SystemUsersService } from "@sims/services";
-import { QueueNames, schedulerQueueNames } from "@sims/utilities";
 
 (async () => {
   const app = await NestFactory.create(QueueConsumersModule);
@@ -32,10 +31,7 @@ import { QueueNames, schedulerQueueNames } from "@sims/utilities";
   serverAdapter.setBasePath("/admin/queues");
   const bullBoardQueues: BullAdapter[] = [];
   queues.forEach((queue) => {
-    if (
-      !queue.isActive &&
-      schedulerQueueNames.includes(queue.name as QueueNames)
-    ) {
+    if (!queue.isActive && queue.cron) {
       logger.log(`Queue service "${queue.name}" is inactive.`);
     } else {
       bullBoardQueues.push(
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
@@ -1,6 +1,6 @@
 import { LoggerService, OnApplicationBootstrap } from "@nestjs/common";
 import { QueueService } from "@sims/services/queue";
-import { QueueNames, schedulerQueueNames } from "@sims/utilities";
+import { QueueNames } from "@sims/utilities";
 import { ConfigService } from "@sims/utilities/config";
 import { InjectLogger } from "@sims/utilities/logger";
 import { CronRepeatOptions, Queue } from "bull";
@@ -33,9 +33,6 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
     const queueConfig = await this.queueService.getQueueConfiguration(
       this.schedulerQueue.name as QueueNames,
     );
-    if (!queueConfig) {
-      return { cron: "inactive" } as CronRepeatOptions;
-    }
     return queueConfig.repeat as CronRepeatOptions;
   }
 
@@ -51,14 +48,17 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
-    // Stops schedulers from running if their isActive status is false
-    // and they fall into the scheduler-queues category.
-    const inactiveQueueNames =
-      await this.queueService.getInactiveQueueConfigurations();
-    if (
-      inactiveQueueNames.includes(this.schedulerQueue.name as QueueNames) &&
-      schedulerQueueNames.includes(this.schedulerQueue.name as QueueNames)
-    ) {
+    // Stops this scheduler from running if the isActive status is false
+    // and it falls into the scheduler-queues category (cron returns a value).
+    const queueConfigurations =
+      await this.queueService.getAllQueueConfigurations();
+    const isInactiveQueue = queueConfigurations.find(
+      (queue) =>
+        queue.queueName === this.schedulerQueue.name &&
+        queue.isActive === false &&
+        queue.queueConfiguration.cron,
+    );
+    if (isInactiveQueue) {
       await this.schedulerQueue.obliterate({ force: true });
       return;
     }
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
@@ -51,14 +51,14 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
     // Stops this scheduler from running if the isActive status is false
     // and it falls into the scheduler-queues category (cron returns a value).
     const queueConfigurations =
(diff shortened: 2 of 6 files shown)


sources/packages/backend/libs/services/src/queue/queue.module.ts
```diff
@@ -81,6 +81,9 @@ function getQueueModules(): BullModuleAsyncOptions[] {
       queueService: QueueService,
     ): Promise<BullModuleOptions> => {
       const queueConfig = await queueService.getQueueConfiguration(queue);
+      if (!queueConfig) {
+        return { prefix: undefined, defaultJobOptions: undefined };
```
andrepestana-aot, line 85: Why the queueConfig cannot be undefined/null?

sources/packages/backend/libs/services/src/queue/queue.service.ts
```diff
@@ -65,6 +75,14 @@ export class QueueService {
    */
   async getQueueConfiguration(queueName: QueueNames): Promise<Bull.JobOptions> {
     const queueConfig = await this.queueConfigurationDetails(queueName);
+    // If a queue configuration record is inactive due to its false in_active flag
```
andrepestana-aot, line 78: "is_active flag"

andrepestana-aot (approved):
Looks good to me. Just added some minor comments.

sources/packages/backend/libs/services/src/queue/queue.service.ts
```diff
@@ -26,14 +26,23 @@ export class QueueService {
...
       },
     });
     return this.queueConfiguration;
   }
 
+  /**
+   * Get all inactive queue names.
+   * @returns queue names.
+   */
+  async getInactiveQueueConfigurations(): Promise<QueueNames[]> {
```
sh16011993, line 39: Should the function name be `getInactiveQueueNames()` ?

sources/packages/backend/libs/services/src/queue/queue.module.ts
lewischen-aot: I just wanted the return data to be consistent with the format `{ prefix: ... , defaultJobOptions: ...}`, which is passed to [BullModule.registerQueueAsync(...getQueueModules())](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/libs/services/src/queue/queue.module.ts#L27). I tried returning undefined just now, and it also worked. Let me know if we want to return undefined instead @andrepestana-aot @andrewsignori-aot 🙂

sources/packages/backend/libs/services/src/queue/queue.service.ts
lewischen-aot: Yes, it should.

sources/packages/backend/apps/queue-consumers/src/main.ts
```diff
@@ -29,10 +30,20 @@ import { SystemUsersService } from "@sims/services";
...
   serverAdapter.setBasePath("/admin/queues");
-  const bullBoardQueues = queues.map((queue) => {
-    return new BullAdapter(app.get<Queue>(`BullQueue_${queue.name}`), {
-      readOnlyMode: queue.dashboardReadonly,
-    });
+  const bullBoardQueues: BullAdapter[] = [];
+  queues.forEach((queue) => {
+    if (
+      !queue.isActive &&
+      schedulerQueueNames.includes(queue.name as QueueNames)
```
andrewsignori-aot, line 37: The `schedulerQueueNames` array  does not seem required since we can get the same result from 
```ts
Object.values(QueueNames)
```
Either way, is there a reason why we would not be relying only on the DB list?

sources/packages/backend/libs/utilities/src/queue.constant.ts
```diff
@@ -23,8 +23,39 @@ export enum QueueNames {
...
   ProcessNotifications = "process-notifications",
   ProcessArchiveApplications = "archive-applications",
   AssessmentWorkflowEnqueuer = "assessment-workflow-enqueuer",
   AssessmentWorkflowQueueRetry = "assessment-workflow-queue-retry",
 }
+
+/**
+ * List of scheduler queue names.
+ */
+export const schedulerQueueNames: QueueNames[] = [
```
andrewsignori-aot, line 36: https://github.com/bcgov/SIMS/pull/3070/files#r1552559926

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -48,6 +51,17 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    // Stops schedulers from running if their isActive status is false
+    // and they fall into the scheduler-queues category.
+    const inactiveQueueNames =
+      await this.queueService.getInactiveQueueConfigurations();
+    if (
+      inactiveQueueNames.includes(this.schedulerQueue.name as QueueNames) &&
+      schedulerQueueNames.includes(this.schedulerQueue.name as QueueNames)
```
andrewsignori-aot, line 60: This is a base class for schedulers only. Is there a need to check `schedulerQueueNames`?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -48,6 +51,17 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    // Stops schedulers from running if their isActive status is false
+    // and they fall into the scheduler-queues category.
+    const inactiveQueueNames =
+      await this.queueService.getInactiveQueueConfigurations();
```
andrewsignori-aot, line 57: Can we have `isActive` as an optional parameter in the existing method `getAllQueueConfigurations`?

sources/packages/backend/apps/queue-consumers/src/main.ts
lewischen-aot: I wasn't going for this list until I saw one of the AC: "Non-schedulers queues will not be affected by the is_active flag for now." I then thought that we should make sure the non-scheduler queue is still running even though the `is_active` flag false. If we are going to rely on the DB list, I can remove this condition.

sources/packages/backend/libs/utilities/src/queue.constant.ts
lewischen-aot: Thanks for the hint. The non-schedulers records don't have `cron` expression in `queueConfiguration`, and we can use that to tell whether the queue is a scheduler or not.

sources/packages/backend/libs/services/src/queue/queue.service.ts
andrewsignori-aot, line 78: What is the motivation for this constraint?
The method is intended to return a queue configuration and it can still return it even when inactive.

sources/packages/backend/libs/services/src/queue/queue.module.ts
andrewsignori-aot: I do not see a reason to change this. Please note the below comment also.
Maybe I am missing something. Can you please elaborate on the motivation?
https://github.com/bcgov/SIMS/pull/3070/files#r1552604114

sh16011993: I have a question here. Why do we need to change the queue configuration for the inactive queues (set them to undefined etc) if the queue itself is going to be obliterated.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -48,6 +51,17 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    // Stops schedulers from running if their isActive status is false
+    // and they fall into the scheduler-queues category.
+    const inactiveQueueNames =
```
andrewsignori-aot, line 56: We can change the method `getQueueConfiguration` to return as extended interface form `Bull.JobOptions` that would contain also `isActive` and/or `isScheduler`.

andrewsignori-aot (commented):
Nice work, please take a look at the comments and some of them will required some discussion but overall I believe there will be only minor changes.

sources/packages/backend/libs/services/src/queue/model/queue.model.ts
```diff
@@ -3,4 +3,6 @@ import { BullModuleOptions } from "@nestjs/bull";
 export interface QueueModel extends Omit<BullModuleOptions, "redis"> {
   name: string;
   dashboardReadonly?: boolean;
+  isActive: boolean;
```
sh16011993, line 6: Are we not keeping this as optional?

sh16011993 (approved):
Nice work @lewischen-aot 👍  Just one info comment.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -48,6 +48,20 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
    * any old cron job delete it and add the new job to the queue.
    */
   async onApplicationBootstrap(): Promise<void> {
+    // Stops this scheduler from running if the isActive status is false
+    // and it falls into the scheduler-queues category (cron returns a value).
+    const queueConfigurations =
```
andrewsignori-aot, line 53: Can we use the same method used in the mait.ts also here (`queueService.queueConfigurationModel`)?

sources/packages/backend/apps/queue-consumers/src/main.ts
```diff
@@ -29,10 +29,17 @@ import { SystemUsersService } from "@sims/services";
   // Create bull board UI dashboard for queue management.
   const serverAdapter = new ExpressAdapter();
   serverAdapter.setBasePath("/admin/queues");
-  const bullBoardQueues = queues.map((queue) => {
-    return new BullAdapter(app.get<Queue>(`BullQueue_${queue.name}`), {
-      readOnlyMode: queue.dashboardReadonly,
-    });
+  const bullBoardQueues: BullAdapter[] = [];
+  queues.forEach((queue) => {
+    if (!queue.isActive && queue.cron) {
```
andrewsignori-aot, line 34: I believe the model can return a property that would indicate if it is a scheduler.
```ts
isScheduler: !!queue.queueConfiguration.cron,
```

sources/packages/backend/libs/services/src/queue/queue.service.ts
```diff
@@ -55,6 +53,8 @@ export class QueueService {
     return queues.map((queue) => ({
       name: queue.queueName,
       dashboardReadonly: queue.queueConfiguration.dashboardReadonly,
+      isActive: queue.isActive,
+      cron: queue.queueConfiguration.cron,
```
andrewsignori-aot, line 57: Suggestion https://github.com/bcgov/SIMS/pull/3070/files#r1554398463
```ts
isScheduler: !!queue.queueConfiguration.cron,
```

andrewsignori-aot (commented):
Thanks for making the changes. Please take a look at latest comments.

sources/packages/backend/libs/services/src/queue/model/queue.model.ts
```diff
@@ -3,4 +3,6 @@ import { BullModuleOptions } from "@nestjs/bull";
 export interface QueueModel extends Omit<BullModuleOptions, "redis"> {
   name: string;
   dashboardReadonly?: boolean;
+  isActive?: boolean;
```
andrewsignori-aot, line 6: Is there a reason for these properties to be optional?
If not please make them not nullable.

sources/packages/backend/libs/services/src/queue/model/queue.model.ts
andrewsignori-aot: I do not see a reason why this would be optional.

andrewsignori-aot (approved):
Thanks for making the changes. Approving hence there is just a minor comment left.
https://github.com/bcgov/SIMS/pull/3070#discussion_r1554437211

Issue #2852, opened 2024-02-20
SFAS/SAIL Student Data Match/Import Note


CarlyCotton:
**User Story**
As a ministry user, I want to know when there was a successful match between a student's SIMS profile and SFAS/SAIL. so that they know that it is successfully matched so that they can determine if they want to go to the Legacy System to see prior program year awards, restrictions, maximums and notes from Legacy Systems

**Expected Behaviour**
On first import of data or data match with SFAS/SAIL, a note will be placed on student account that is viewable by ministry users to confirm a match was made/data was transferred.

**Acceptance Criteria**
- [ ] Add 'general' note to student profile during **account creation** when match occurs to indicate other data exists from outside of SIMS for student.
_"Successful match with legacy systems - data has been transferred"_

**Technical**
- [ ] Create a new note for the Student in the centralized student account creation method ([source reference](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/services/student/student.service.ts#L173)).
- [ ] Use the system user to create the new note to make it clear in the Notes UI that it was a system-generated note.
- [ ] Use the already-in-place centralized method for creating the student note.

Joshua-Lakusta:
Modified the AC from First match to only on account creation. Can further discuss future tickets or manual processes for other match scenarios.

PR #3074, opened 2024-04-05
#2852 - SFAS SAIL Student Data Match Import Note

katrina-aot-git:
Add a system note upon student creation when there is a legacy system match. 

The note is further down in the code than suggested in the ticket so that the student has been created and has an id to use for the note.

### DIFF
--- a/sources/packages/backend/apps/api/src/services/student/student.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student/student.service.ts
@@ -232,11 +232,12 @@ export class StudentService extends RecordDataModelService<Student> {
           .update({ individualId: sfasIndividual.id }, { processed: false });
 
         // If there is a match in SFAS, record that information is available from legacy system with a note attached to the student.
-        this.addStudentNote(
+        await this.noteSharedService.createStudentNote(
           student.id,
           NoteType.General,
           "Successful match with legacy systems - data has been transferred.",
           this.systemUsersService.systemUser.id,
+          entityManager,
         );
       }
 
--- a/sources/packages/backend/apps/api/src/services/student/student.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student/student.service.ts
@@ -234,7 +234,7 @@ export class StudentService extends RecordDataModelService<Student> {
         // If there is a match in SFAS, record that information is available from legacy system with a note attached to the student.
         this.addStudentNote(
           student.id,
-          NoteType.System,
+          NoteType.General,
           "Successful match with legacy systems - data has been transferred.",
           this.systemUsersService.systemUser.id,
         );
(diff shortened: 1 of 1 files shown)


sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -228,6 +230,14 @@ export class StudentService extends RecordDataModelService<Student> {
         await entityManager
           .getRepository(SFASRestriction)
           .update({ individualId: sfasIndividual.id }, { processed: false });
+
+        // If there is a match in SFAS, record that information is available from legacy system with a note attached to the student.
+        this.addStudentNote(
+          student.id,
+          NoteType.System,
```
andrepestana-aot, line 237: Please, confirm if this is a "System"  or "General" note type. The ticket says: "Add 'general' note to student profile". It's not clear to me.

andrepestana-aot (commented):
Good job working on this ticket. Please take a look at my comment.

sources/packages/backend/apps/api/src/services/student/student.service.ts
katrina-aot-git: Probably is meant to be 'general'! I'd focused on the 'system-generated note' comment. Will update.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -228,6 +230,14 @@ export class StudentService extends RecordDataModelService<Student> {
         await entityManager
           .getRepository(SFASRestriction)
           .update({ individualId: sfasIndividual.id }, { processed: false });
+
+        // If there is a match in SFAS, record that information is available from legacy system with a note attached to the student.
+        this.addStudentNote(
```
andrewsignori-aot, line 235: `await` needs to be added.

andrewsignori-aot, line 235: Please use the method `noteSharedService.createStudentNote`  that would allow the execution in the same transaction `entityManager`.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/services/student/student.service.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/services/student/student.service.ts
sh16011993: FMU: What is the difference b/w `System` and `General` notes?

andrewsignori-aot: @sh16011993 `System` and `General` are just categories under the notes.
image

The request from the business was to add it under the `General` category, this was the AC and we should follow it unless another decision is taken. 
We are using the system user because it seemed to us not consistent to use the student user while saving this particular note. This was the decision during the technical ACs creation.

andrepestana-aot (approved):
Thanks for checking and doing the changes!

sources/packages/backend/apps/api/src/services/student/student.service.ts
sh16011993: Thanks @andrewsignori-aot

Issue #3076, opened 2024-04-08
(Bug) Locked at in progress after approving an exception request.


Joshua-Lakusta:
**Describe the Bug**
In the workflow when an exception is completed the application is getting stuck at the in progress until the timer (12 hours elapses and it resolves itself (note waiting till tomorrow morning to confirm that timer path is working))

**Expected Behaviour**
Once an exception is approved then the application should be able to move past the in progress phase to assessment. 

**Actual Behaviour**
The current process flow is not receiving the message to unblock until the timer resolves it after a period of time (12 hours). 

** Steps To Reproduce**
Steps to reproduce the behaviour:
Create a student application with an exception request include and submit it
Approve the exception request on the ministry portal
Check application status and it should be stuck at in progress

**Additional Technical context**

The message sent to Camunda is using the `exceptionId` while the workflow is expecting the `applicationId`, as shown below. The assumption would be that it was working previously based on the timer only (previously configured to PT30S). There are no errors generated in the API because the message is successfully published to Camunda.

<img width="789" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN1ZDQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d88a245d2eb2b4d3c5445b3bde5d435a5c954415/image.png" alt="image.png" />

<img width="672" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2RDQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--64a01f2028c614b99c8a6c59a6643f944daec920/image.png" alt="image.png" />

Please see the [source reference](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/libs/services/src/workflow/workflow-client.service.ts#L156-L176).

PR #3087, opened 2024-04-09
#3076 - Application exception bug fix

dheepak-aot:
# Fix the correlation id sent on application exception verified message

## Bug
When application exception is verified by ministry (Approved or Declined), message is sent to workflow with correlation id, so that the workflow can proceed for further processing.

The workflow expects a message with name `application-exception-verified` and correlation id `${applicationId}` but it was receiving a message with correlation id `${applicationExceptionId}` and hence the message was not accepted and processed by workflow.

## Fix

Updated the code to send `applicationId` as correlation id for the message.

image


guru-aot (approved):
LGTM, Nice work @dheepak-aot

andrewsignori-aot (approved):
The change looks good, I just have a minor question.
The fix would also be possible to change the expected correlation ID on Camunda which would demand almost no changes other than a string on the Camunda definition. Was there any thought process between choosing one way or another?

sh16011993 (approved):
Nice Work @dheepak-aot 👍

Issue #2369, opened 2023-10-03
Create Automatic E-mail Notifications to SABC Staff 


HRAGANBC:
**User Story**
As a **_Ministry user_**, I want to receive automatic notifications anytime specific trigger events occur within SIMS so that a JIRA ticket gets created for each such event. This will help me manage completion status and assign work tasks to staff members using JIRA functionality. 

**Description of Task:**
Automatically configure simple e-mails to specific SABC e-mail addresses any time particular trigger events occur within SIMS. 

**Context:**
In the absence of custom workload management tools within the various queues in SIMS, Service Delivery staff will have difficulty tracking which staff are working on which pending request, the status of each request, and other ticket management tools. 

**Business Requirements Details**
<img width="1582" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNDNKQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--07fd8cc6c09f4e55507048fb02da1ebc0db33a1c/image.png" alt="image.png" />

Email Body should include the following details, if the value doesn't exist can be blank:
Student Profile:
- Student: [StudentFirst] [StudentLast]
- Date of Birth: [StudentDOB]
- Email: [StudentEmail]
- Application: [ApplicationNumber]
- Date/Time: [Date/Time of Event]

Institution Profile:
- Institution: [InstitutionOperatingName]
- Location: [InstitutionLocationName]
- Program: [ProgramName]
- Offering: [OfferingName]
- Email: [InstitutionPrimaryEmail]
- Date/Time: [Date/Time of Event]

**Acceptance Criteria**
- [ ] Configure e-mails to be automatically configured anytime the below trigger events occur within SIMS:
- - [x] Student File Uploads - TemplateID: `3b37994f-464f-4eb0-ad30-84739fa82377`
- - [x] Student Submits Change Request after COE - TemplateID: `241a360a-07d6-486f-9aa4-fae6903e1cff`
- - [x] Student Submits Application with Exception Request - TemplateID: `997c6625-5571-4c92-94d1-7d7c1b86fd8c`
- - [x] Student Requests BasicBCeID account - TemplateID: `e9cf5143-c553-4937-9a5c-fab4230782aa`
- - [x] Student application is edited for 5th time - TemplateID: `952acba8-8806-4b9e-aee3-a14ea3823bbf`
[...]

PR #3122, opened 2024-04-12
#2369 - Create Automatic Email Notifications to SABC Staff

sh16011993:

**1. Created DB migrations for:**

- Added `email_contacts` column to the `notification_messages` table.
- Added the ministry notification message templates to the `notification_messages` table.

**2. Updated all the pre-existing methods to use the newly created `email_contacts` column for all emails sent to the Ministry.**

**3. Configured e-mails to be automatically saved anytime the below trigger events occur within SIMS:**

- Student Uploads File
- Student Submits Change Request after COE 
- Student Submits Application with Exception Request
- Student Requests Basic BCeID account
- Student application is edited for 5th time
- Change to a Student's Program or Offering (is same as Application Offering Change Request post student approval)
- Institution adds pending program
- Institution adds pending offering
- Institution adds pending offering (via bulk offering interface)
- Institution requests designation

**4. For templates that require the email column to be populated, registered an error log that reads: "Email template id `template_id` requires a configured email to be sent."**

**5. Added `locationName` to the `OfferingValidationModel` to allow for `locationName` to allow for the pending offering notification to have the `locationName` be added to the personalization section.**

**6. Updated the e2e tests to incorporate the changed notification logic for api, queue-consumers and workers.**

### **Student uploads file**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Student Submits Change Request after COE**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Student Submits Application with Exception Request**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Student Requests Basic BCeID account**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Student application is edited for 5th time**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Change to a Student's Program or Offering (is same as Application Offering Change Request post student approval)**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Institution adds pending program**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Institution adds pending offering**
------------------------------------------------------------------------------------------------------------------------------------------
image

### **Institution adds bulk offering**
------------------------------------------------------------------------------------------------------------------------------------------
**CSV File:**
image

**First 2 offering from CSV are in Creation Pending state and the 3rd offering is in Approved state:**
image

**Notification for the first 2 offerings:**
image

### **Institution requests designation**
------------------------------------------------------------------------------------------------------------------------------------------
image

### Part of the next PR:

1. Remove the existing `env` variable that contains the Ministry email.
2. Create an SQL script to be added to the release instructions to have the emails updated on PROD.


sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -80,23 +92,25 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
-   * If a student submit/re-submit and an existing application that is not in draft state
+   * If a student submit/re-submit an existing application that is not in draft state
    * (i.e existing application should be in any one of these status, submitted, In Progress,
    * Assessment, Enrollment), then the existing application status is set to `Overwritten` and
    * applicationStatusUpdatedOn is updated and delete the corresponding workflow and creates a
    * new Application with same Application Number and Program Year as that of the Overwritten
    * Application and with newly submitted payload. And starts a new workflow for the newly created
-   * Application.
-   * a new application with status 'Submitted'
+   * Application. If the application is edited for the 5th time, a notification is saved for the
```
andrepestana-aot, line 112: I'm not sure if it is a good idea to hardcode numbers like '5th time' in the comments because it might change.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -276,8 +290,39 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        });
+        const ministryNotification: ApplicationExceptionRequestNotificationForMinistry =
+          {
+            givenNames: student.user.firstName,
+            lastName: student.user.lastName,
+            email: student.user.email,
+            dob: student.birthDate,
+            applicationNumber: application.applicationNumber,
+          };
+        await this.notificationActionService.saveApplicationEditedFifthTimeNotificationForMinistry(
```
andrepestana-aot, line 320: Please don't create a method with "Fifth" in the name. That number might change and we'll need to refactor the name.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-part-time-process-integration.scheduler.e2e-spec.ts
```diff
@@ -58,6 +58,13 @@ describe(
       processor = app.get(PartTimeECertProcessIntegrationScheduler);
       systemUsersService = app.get(SystemUsersService);
       gcNotifyService = app.get(GCNotifyService);
+      // Insert fake email contact to send ministry email.
+      await db.notificationMessage.update(
+        {
+          id: NotificationMessageType.MinistryNotificationDisbursementBlocked,
+        },
+        { emailContacts: ["shashank.x.shekhar@gov.bc.ca"] },
```
andrepestana-aot, line 66: Are you sure you're going to use your real email?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -69,6 +80,7 @@ export const INSUFFICIENT_APPLICATION_SEARCH_PARAMS =
 
 @Injectable()
 export class ApplicationService extends RecordDataModelService<Application> {
+  private readonly studentRepo: Repository<Student>;
```
andrewsignori-aot, line 83: Please use the already-in-place method `getStudentById` from the `SudentService`. If needed this method can be shared from a more generic library to be reused.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: It can be renamed to something like `saveApplicationEditedTooManyTimesNotification`.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -276,8 +290,39 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        });
+        const ministryNotification: ApplicationExceptionRequestNotificationForMinistry =
+          {
+            givenNames: student.user.firstName,
+            lastName: student.user.lastName,
+            email: student.user.email,
+            dob: student.birthDate,
+            applicationNumber: application.applicationNumber,
+          };
+        await this.notificationActionService.saveApplicationEditedFifthTimeNotificationForMinistry(
```
andrewsignori-aot, line 320: All notification-related logic can be moved to a private method for better readability.
The `submitApplication` is already too long.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -276,8 +290,39 @@ export class ApplicationService extends RecordDataModelService<Application> {
       application.currentAssessment.updatedAt = now;
 
       await applicationRepository.save([application, newApplication]);
+      // Create a notification if the application is edited for the 5th time.
+      const applicationsCount = await applicationRepository.count({
+        where: { applicationNumber: newApplication.applicationNumber },
+      });
+      if (
+        applicationsCount ===
+        APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION + 1
```
andrewsignori-aot, line 299: The const `APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION` can represent the value (including the +1) to avoid having the "+1" here.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -276,8 +290,39 @@ export class ApplicationService extends RecordDataModelService<Application> {
       application.currentAssessment.updatedAt = now;
 
       await applicationRepository.save([application, newApplication]);
+      // Create a notification if the application is edited for the 5th time.
+      const applicationsCount = await applicationRepository.count({
+        where: { applicationNumber: newApplication.applicationNumber },
+      });
+      if (
+        applicationsCount ===
```
andrewsignori-aot, line 298: I was expecting something like saving the application number as metadata and identifying if the application ever sent this notification after the count was exceeded.

sources/packages/backend/apps/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -23,16 +27,22 @@ import {
  */
 @Injectable()
 export class DesignationAgreementService extends RecordDataModelService<DesignationAgreement> {
-  constructor(private readonly dataSource: DataSource) {
+  private readonly institutionRepo: Repository<Institution>;
```
andrewsignori-aot, line 30: We hardly declare the repositories following this approach.
Please check if the `InstitutionService` has a method already similar to the one needed. 
If not, create a method in the service and consume the service.

sources/packages/backend/apps/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -67,7 +77,29 @@ export class DesignationAgreementService extends RecordDataModelService<Designat
...
+      select: {
+        operatingName: true,
+        legalOperatingName: true,
+        primaryEmail: true,
+      },
+      where: { id: institutionId },
+    });
+    const ministryNotification: InstitutionRequestsDesignationNotificationForMinistry =
+      {
+        institutionName: institution.legalOperatingName,
```
andrewsignori-aot, line 90: Is `institutionName` part of the template?

sources/packages/backend/apps/api/src/services/designation-agreement/designation-agreement.service.ts
```diff
@@ -67,7 +77,29 @@ export class DesignationAgreementService extends RecordDataModelService<Designat
...
+        institutionOperatingName: institution.operatingName,
+        institutionPrimaryEmail: institution.primaryEmail,
+      };
+    return this.dataSource.transaction(async (entityManager) => {
+      await this.notificationActionsService.saveInstitutionRequestsDesignationNotificationForMinistry(
```
andrewsignori-aot, line 95: Name suggestion `saveInstitutionRequestsDesignationNotification`.

andrepestana-aot (commented):
Good job! Left some comments. Please take a look.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -64,17 +64,26 @@ import {
 import { EducationProgramOfferingValidationService } from "./education-program-offering-validation.service";
 import * as os from "os";
 import { LoggerService, InjectLogger } from "@sims/utilities/logger";
+import {
+  InstitutionAddsPendingOfferingNotificationForMinistry,
+  NotificationActionsService,
+} from "@sims/services";
 @Injectable()
 export class EducationProgramOfferingService extends RecordDataModelService<EducationProgramOffering> {
+  private readonly institutionLocationRepo: Repository<InstitutionLocation>;
```
andrewsignori-aot, line 73: Similar to https://github.com/bcgov/SIMS/pull/3122/files#r1566229176

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -96,7 +105,41 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+          offeringName: educationProgramOffering.offeringName,
+          institutionPrimaryEmail: institution.primaryEmail,
+        };
+      return await this.dataSource.transaction(async (entityManager) => {
+        if (programOffering.offeringStatus === OfferingStatus.CreationPending) {
+          await this.notificationActionsService.saveInstitutionAddsPendingOfferingNotificationForMinistry(
+            ministryNotification,
+            entityManager,
+          );
+        }
```
andrewsignori-aot, line 138: This can be moved to a private method to avoid extending too much the the method that has the main intention of saving the program.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -35,18 +35,25 @@ import {
...
+import {
+  InstitutionAddsPendingProgramNotificationForMinistry,
+  NotificationActionsService,
+} from "@sims/services";
 
 const OTHER_REGULATORY_BODY = "other";
 @Injectable()
 export class EducationProgramService extends RecordDataModelService<EducationProgram> {
   private readonly offeringsRepo: Repository<EducationProgramOffering>;
+  private readonly institutionRepo: Repository<Institution>;
```
andrewsignori-aot, line 47: https://github.com/bcgov/SIMS/pull/3122/files#r1566285652

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-part-time-process-integration.scheduler.e2e-spec.ts
sh16011993: I had to use one of the emails present in the GC Notify emails to send list. It won't let me send an email to an email address not present in the list.

sources/packages/backend/apps/api/src/services/student-account-applications/student-account-applications.models.ts
```diff
@@ -12,6 +12,20 @@ export interface StudentAccountApplicationCreateModel {
   email: string;
   firstName: string;
   lastName: string;
+  dateOfBirth: string;
```
andrewsignori-aot, line 15: Why was this interface expanded in such a way? Can you please elaborate?

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -41,6 +44,7 @@ import { StudentFileService } from "../student-file/student-file.service";
  */
 @Injectable()
 export class StudentAppealService extends RecordDataModelService<StudentAppeal> {
+  private readonly applicationRepo: Repository<Application>;
```
andrewsignori-aot, line 47: https://github.com/bcgov/SIMS/pull/3122/files#r1566285652

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: image

I think we already have it shared.

sources/packages/backend/apps/db-migrations/src/sql/NotificationMessages/Add-email-contacts-column.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  sims.notification_messages
+ADD
+  COLUMN email_contacts VARCHAR(300) [];
+
+COMMENT ON COLUMN sims.notification_messages.email_contacts IS 'Addresses who receive notification email.';
```
andrewsignori-aot, line 6: Comment suggestion adding more business info.
```
Email addresses to receive a notification when these emails are targeted to someone, not the user, for instance, the Ministry or an external party.
```

sources/packages/backend/apps/db-migrations/src/sql/NotificationMessages/Rollback-insert-ministry-notification-messages.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM
+  sims.notification_messages
+WHERE
```
andrewsignori-aot, line 3: Between can be used also.

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: I feel that the currently implemented way of having the `APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION` as per the business requirement is better since this property can be easily adjusted later on if required without diving into the code part of it where the `+1` logic is handled. IF this constant is moved out to an external properties files, then it becomes easy to update this to a new value without causing any confusion.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-part-time-process-integration.scheduler.e2e-spec.ts
andrewsignori-aot: E2E tests are not supposed to send emails to GC Notify. Can you please elaborate on why are we doing this?

sources/packages/backend/libs/sims-db/src/entities/notification.model.ts
```diff
@@ -165,4 +165,36 @@ export enum NotificationMessageType {
    * Ministry disbursement blocked.
    */
   MinistryNotificationDisbursementBlocked = 17,
+  /**
+   * Student submitted change request after COE.
+   */
+  StudentSubmittedChangeRequestNotificationForMinistry = 18,
```
andrewsignori-aot, line 171: I do not see a reason to have every notification appended with the `ForMinistry`. Maybe it can be expressed in the comment.

sources/packages/backend/libs/sims-db/src/entities/notification-message.model.ts
```diff
@@ -30,4 +30,14 @@ export class NotificationMessage extends RecordDataModel {
     nullable: false,
   })
   templateId: string;
+  /**
+   * Address to send the email notification.
```
andrewsignori-aot, line 34: If the comment suggested for the SQL is accepted please ensure that this one will also be updated.

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -90,3 +90,8 @@ export const BLOCKED_DISBURSEMENT_NOTIFICATION_MIN_DAYS_INTERVAL = 7;
  * Represents the maximum number of disbursement blocked notifications to send.
  * */
 export const BLOCKED_DISBURSEMENT_MAXIMUM_NOTIFICATIONS_TO_SEND = 3;
+
+/**
+ * Represents the count of the number of times an application is edited to trigger a notification to be sent to the ministry.
+ */
+export const APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION = 5;
```
andrewsignori-aot, line 97: There is an equivalent file inside the API. If it is an API-only const it can be moved to the most specific scope.

sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
```diff
@@ -118,12 +129,31 @@ export class ApplicationController {
         application.data,
       );
       if (exceptions.length) {
-        const createdException =
-          await this.applicationExceptionService.createException(
-            job.variables.applicationId,
-            exceptions,
+        let createdException: ApplicationException;
```
andrewsignori-aot, line 132: There is no need to create this externally to the transaction. The return can be from the `job.complete`, right?

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -49,29 +55,41 @@ export class NotificationActionsService {
...
+      await this.notificationMessageService.getNotificationMessageDetails(
+        NotificationMessageType.StudentFileUpload,
+      );
+    if (!notificationDetails.emailContacts?.length) {
+      this.logger.error(
+        `Email template id ${NotificationMessageType.StudentFileUpload} requires a configured email to be sent.`,
+      );
+      return;
+    }
+    const ministryNotificationsToSend = [];
```
andrewsignori-aot, line 68: The use of a `map` would make the method easier to read, IMO. Is there a reason to chose `for` for this scanrio?

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -49,29 +55,41 @@ export class NotificationActionsService {
...
-        template_id: templateId,
-        personalisation: {
-          givenNames: notification.firstName ?? "",
-          lastName: notification.lastName,
-          dob: getDateOnlyFormat(notification.birthDate),
-          applicationNumber: notification.applicationNumber,
-          documentPurpose: notification.documentPurpose,
-          date: this.getDateTimeOnPSTTimeZone(),
+    const notificationDetails =
+      await this.notificationMessageService.getNotificationMessageDetails(
```
andrewsignori-aot, line 59: For `getNotificationMessageDetails` we can maybe enforce the return type using the below.
```ts
Promise<Pick<NotificationMessage, "templateId" | "emailContacts">>
```
And consume the result as below.
```ts
const { templateId, emailContacts } =
  await this.notificationMessageService.getNotificationMessageDetails(
    NotificationMessageType.InstitutionAddsPendingOfferingNotificationForMinistry,
  );
```

andrewsignori-aot (commented):
Great work, please take a look at the comments. I will do a closer look into the `notification-actions.service.ts` once the suggestions are implemented (or not).

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
sh16011993: I remember you explaining me this in call earlier. At the time, I didn't get it. Now I got it 😅 . Will update it. Thanks

sources/packages/backend/apps/api/src/services/designation-agreement/designation-agreement.service.ts
sh16011993: Yes

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-part-time-process-integration.scheduler.e2e-spec.ts
sh16011993: Oops, my bad. Missed the part that here we just check if an entry has been made to the `notifications` table. Updated it to a dummy email. Thanks.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -274,13 +281,50 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+   * @param applicationNumber application number of the related application.
+   * @param studentId related student id.
+   * @param transactionalEntityManager entity manager to be a part of the transaction.
+   */
+  private async saveApplicationEditedTooManyTimesNotification(
+    applicationNumber: string,
+    studentId: number,
+    transactionalEntityManager: EntityManager,
+  ): Promise<void> {
+    // Create a notification if the application is edited for APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION time.
```
andrewsignori-aot, line 306: Since it became a method the method comment is already explaining pretty well the same.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: If the below comment is accepted the "+1" will no longer be needed, right?
https://github.com/bcgov/SIMS/pull/3122#discussion_r1566223186

sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.controller.service.ts
```diff
@@ -163,6 +165,9 @@ export class EducationProgramOfferingControllerService {
       programId,
       institutionId,
     );
+    // Get institution location details.
```
andrewsignori-aot, line 168: Either this check can be moved down line 171 or both queries can be performed in parallel..

sources/packages/backend/apps/api/src/services/application-offering-change-request/application-offering-change-request.service.ts
```diff
@@ -365,6 +366,37 @@ export class ApplicationOfferingChangeRequestService {
     return applicationOfferingChangeRequest;
   }
 
+  /**
+   * Gets the application offering change request details.
+   * @param applicationOfferingChangeRequestId application offering change request id to get the details.
+   * @returns fetched application offering change request details.
+   */
+  async getApplicationOfferingChangeRequestById(
+    applicationOfferingChangeRequestId: number,
```
andrewsignori-aot, line 375: Minor, here just `id` would be enough IMO.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -274,13 +286,69 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    studentId: number,
+    transactionalEntityManager: EntityManager,
+  ): Promise<void> {
+    const applicationRepository =
+      transactionalEntityManager.getRepository(Application);
+    const applicationsCount = await applicationRepository.count({
+      where: { applicationNumber },
+    });
+    if (applicationsCount > APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION) {
+      const notification = await transactionalEntityManager
```
andrewsignori-aot, line 317: There is a difference between requesting a count from the DB vs requesting if a record exists and we should query the DB for what we need, in this situation would be just an exists.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -274,13 +286,69 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        .getRepository(Notification)
```
andrewsignori-aot, line 318: This query belongs to a notification service, right?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-validation.models.ts
```diff
@@ -600,6 +602,11 @@ export class OfferingValidationModel {
     message: "Related institution location was not found or was not provided.",
   })
   locationId: number;
+  /**
+   * Institution location name of the institution that will be associated with this offering.
+   */
+  @IsNotEmpty({ message: "Institution location name is required." })
```
andrewsignori-aot, line 608: Maybe we can extend the comment to mention:
```
Used as supporting data, for instance, as part of the notification to indicate offering in pending status.
```

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -62,19 +68,28 @@ import {
...
+  NotificationActionsService,
+} from "@sims/services";
 @Injectable()
 export class EducationProgramOfferingService extends RecordDataModelService<EducationProgramOffering> {
   constructor(
     private readonly dataSource: DataSource,
     private readonly offeringValidationService: EducationProgramOfferingValidationService,
+    private readonly notificationActionsService: NotificationActionsService,
+    private readonly institutionLocationService: InstitutionLocationService,
+    private readonly entityManager: EntityManager,
```
andrewsignori-aot, line 85: To this day we never follow a similar approach. Is there a benefit to doing it?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -166,6 +217,45 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+   */
+  private async saveEducationProgramOfferingNotification(
+    educationProgramOffering: Pick<OfferingValidationModel, "offeringName"> &
+      Pick<OfferingValidationModel["programContext"], "name">,
+    programOffering: Pick<EducationProgramOffering, "offeringStatus"> & {
+      institutionLocation: Pick<
+        EducationProgramOffering["institutionLocation"],
+        "id" | "name"
+      >;
+    },
```
andrewsignori-aot, line 235: The parameters are not improving the readability or maintainability.
Can you please declare them individually?
Because it is a private method, in this I would recommend even directly accepting the `validatedOffering` which would simplify the consumer code also.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -96,7 +111,24 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
     programOffering.parentOffering = programOffering;
 
     try {
-      return await this.repo.save(programOffering);
+      const institutionLocation =
```
andrewsignori-aot, line 114: Is this call to get the location data needed since this method receives a `offeringValidationModel` that comes from `buildOfferingValidationModel`?

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
sh16011993: Hmm. I don't know of any specific advantage. Just that the framework would inject an object into the service and I can use it. Will change it.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -274,13 +283,58 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        transactionalEntityManager,
+      );
     });
-
     return { application, createdAssessment: originalAssessment };
   }
 
+  /**
+   * Saves a notification when the application is edited too many times
+   * governed by APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION.
```
andrewsignori-aot, line 298: If you would like to use the const in the comment please use it as below.
```ts
governed by {@link APPLICATION_EDIT_COUNT_TO_SEND_NOTIFICATION}.
```

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -166,6 +207,45 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+    const institution = institutionLocation.institution;
+    const ministryNotification: InstitutionAddsPendingOfferingNotification = {
+      institutionName: institution.legalOperatingName,
+      institutionOperatingName: institution.operatingName,
+      institutionLocationName,
+      programName,
+      offeringName,
+      institutionPrimaryEmail: institution.primaryEmail,
+    };
+    if (programOfferingStatus === OfferingStatus.CreationPending) {
```
andrewsignori-aot, line 241: Please move the `if` to the top of the method and do an early return.
If my understanding is correct, if offering is not in `CreationPending` nothing else should be checked and we can avoid, for instance, the call to `getInstitutionLocation`.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -166,6 +207,45 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+  private async saveEducationProgramOfferingNotification(
+    offeringName: string,
+    programName: string,
+    programOfferingStatus: OfferingStatus,
+    institutionLocationId: number,
+    institutionLocationName: string,
+    entityManager: EntityManager,
+  ): Promise<void> {
+    const institutionLocation =
+      await this.institutionLocationService.getInstitutionLocation(
```
andrewsignori-aot, line 229: This method is called inside a `for` for the `bulk-insert`, right?
Is this not the call that we would like to avoid?
Can we not have the data needed loading ahead of time or is this not the one that we talked about before?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -227,7 +240,18 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
     } else {
       program.modifier = auditUser;
     }
-    return this.repo.save(program);
+    return await this.dataSource.transaction(
```
andrewsignori-aot, line 243: Is there a need for the `await` here?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -731,4 +756,32 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+    const institution = await this.institutionService.getInstitutionDetailById(
+      institutionId,
+    );
+    const ministryNotification: InstitutionAddsPendingProgramNotification = {
+      institutionName: institution.legalOperatingName,
+      institutionOperatingName: institution.operatingName,
+      programName: program.name,
+      institutionPrimaryEmail: institution.primaryEmail,
+    };
+    if (program.programStatus === ProgramStatus.Pending) {
```
andrewsignori-aot, line 780: Please move the if condition to the top.
```ts
if (program.programStatus !== ProgramStatus.Pending) {
  return;
}
```

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -731,4 +756,32 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+   * Saves institution adds pending program notification for the ministry.
+   * @param program program
+   * @param institutionId related institution id.
+   * @param entityManager
+   */
+  private async saveInstitutionAddsPendingProgramNotification(
+    program: Pick<EducationProgram, "name" | "programStatus">,
+    institutionId: number,
+    entityManager: EntityManager,
+  ) {
```
andrewsignori-aot, line 770: Please add the return type.

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -170,16 +177,18 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
       .where("institutionLocation.id = :locationId", {
         locationId: locationId,
       });
     if (institutionId) {
-      query
-        .innerJoin("institutionLocation.institution", "institution")
-        .andWhere("institution.id = :id", {
-          id: institutionId,
-        });
+      query.andWhere("institution.id = :id", {
```
andrewsignori-aot, line 189: The `locationId` is mandatory and more specific than the `institutionId`, why do we need the `institutionId`?

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -95,7 +101,22 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
...
+      );
+      const student = application.student;
+      const ministryNotification: StudentSubmittedChangeRequestNotification = {
+        givenNames: student.user.firstName,
+        lastName: student.user.lastName,
+        email: student.user.email,
+        birthDate: student.birthDate,
+        applicationNumber: application.applicationNumber,
+      };
+      await this.notificationActionsService.saveStudentSubmittedChangeRequestNotification(
```
andrewsignori-aot, line 115: These operations can happen in parallel.

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -120,6 +121,34 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+  /**
+   * Checks if a notification exists for the given optional parameters provided.
+   * @param transactionalEntityManager entity manager to be part of the transaction.
+   * @param options related options.
+   * - `applicationNumber` related application number.
+   * @returns boolean true if the notification exists, else false.
+   */
+  async checkNotificationExists(
+    transactionalEntityManager: EntityManager,
+    options?: { applicationNumber?: string },
```
andrewsignori-aot, line 133: This does not seem optional for the method.
Either this method is generic or dedicated to `ApplicationEditedTooManyTimesNotification` and must be renamed.
At this moment do whatever would be faster.

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
andrewsignori-aot: Are we not changing the line 58 to use `const { templateId, emailContacts }`?

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -90,16 +106,17 @@ export class NotificationActionsService {
     auditUserId: number,
     entityManager?: EntityManager,
   ): Promise<void> {
-    const templateId = await this.notificationMessageService.getTemplateId(
-      NotificationMessageType.MinistryFileUpload,
-    );
+    const notificationDetails =
```
andrewsignori-aot, line 109: Same from https://github.com/bcgov/SIMS/pull/3122/files#r1571234632

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -201,12 +220,13 @@ export class NotificationActionsService {
     auditUserId: number,
     entityManager: EntityManager,
   ): Promise<void> {
-    const templateId = await this.notificationMessageService.getTemplateId(
-      NotificationMessageType.ApplicationOfferingChangeRequestCompletedByMinistry,
-    );
+    const notificationDetails =
```
andrewsignori-aot, line 223: Same from https://github.com/bcgov/SIMS/pull/3122/files#r1571234632

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -237,16 +257,17 @@ export class NotificationActionsService {
     auditUserId: number,
     entityManager?: EntityManager,
   ): Promise<void> {
-    const templateId = await this.notificationMessageService.getTemplateId(
-      NotificationMessageType.StudentRestrictionAdded,
-    );
+    const notificationDetails =
```
andrewsignori-aot, line 260: https://github.com/bcgov/SIMS/pull/3122/files#r1571234632

sources/packages/backend/libs/services/src/notifications/notification/notification-actions.service.ts
```diff
@@ -662,35 +703,420 @@ export class NotificationActionsService {
...
+      await this.notificationMessageService.getNotificationMessageDetails(
+        NotificationMessageType.InstitutionAddsPendingOfferingNotification,
+      );
+    if (!notificationDetails.emailContacts?.length) {
+      this.logger.error(
+        `Email template id ${NotificationMessageType.InstitutionAddsPendingOfferingNotification} requires a configured email to be sent.`,
+      );
+      return;
+    }
```
andrewsignori-aot, line 1096: This code is repeated across many methods.
Suggested method.
```ts
private async assertNotificationMessageDetails(
    notificationMessageType: NotificationMessageType,
  ): Promise<
    Pick<NotificationMessage, "templateId" | "emailContacts"> | undefined
> {
  const { templateId, emailContacts } =
    await this.notificationMessageService.getNotificationMessageDetails(
      notificationMessageType,
    );
  if (!emailContacts?.length) {
    this.logger.error(
      `Email template id ${notificationMessageType} requires a configured email to be sent.`,
    );
    return;
  }
  return { templateId, emailContacts };
}
```
Suggested use.
```ts
const { emailContacts, templateId } =
    await this.getNotificationMessageDetails(
      NotificationMessageType.StudentFileUpload,
    );
  if (!emailContacts?.length) {
    return;
  }
```

andrewsignori-aot (commented):

sources/packages/backend/libs/services/src/notifications/notification-message/notification-message.service.ts
```diff
@@ -13,26 +13,29 @@ export class NotificationMessageService extends RecordDataModelService<Notificat
...
+          id: notificationMessageTypeId,
+        },
       },
-      where: {
-        id: notificationMessageTypeId,
-      },
-    });
-    return notificationMessage.templateId;
+    );
+    return { templateId, emailContacts };
```
andrepestana-aot, line 39: Instead of destructuring and creating another object literal you can just return it from the query:

```ts
return await notificationMessageRepo.findOne(
      {
        select: {
          templateId: true,
          emailContacts: true,
        },
        where: {
          id: notificationMessageTypeId,
        },
      },
    );

```

andrepestana-aot: I think await is not need also as you're returning a Promise.

sources/packages/backend/libs/services/src/notifications/notification/notification.model.ts
```diff
@@ -68,11 +69,50 @@ export interface StudentNotification {
...
+  givenNames: string;
+  lastName: string;
+  email: string;
+  birthDate: string;
+  applicationNumber: string;
+}
+
+export interface ApplicationEditedTooManyTimesNotification {
+  givenNames: string;
+  lastName: string;
```
dheepak-aot, line 90: I see almost all the interfaces are the same except one or 2 which does not have application number.

Please have a Base Interface and use extends.

andrewsignori-aot: @dheepak-aot even though them being the same data they are targeting different template IDs that can potentially evolve differently and right now they are also processed individually. An interface would not bring a great improvement here IMO. Please let me know if this is a blocker.

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
sh16011993: I see that this method is being called from multiple places and one of those is `getInstitutionLocation` called from `institution-location.institutions.controller.ts` which passes both location and institution id (seems to be for authorization purpose).

sources/packages/backend/apps/api/src/services/student-appeal/student-appeal.service.ts
```diff
@@ -95,7 +101,30 @@ export class StudentAppealService extends RecordDataModelService<StudentAppeal>
           { entityManager: entityManager },
         );
       }
-      return this.repo.save(studentAppeal);
+      const application = await this.applicationService.getApplicationInfo(
+        applicationId,
+      );
```
dheepak-aot, line 106: As per the lessons learned from the load test, we should not use external repo inside transaction. 

Either move this method outside the transaction, or update `getApplicationInfo` to use optional external entity manager.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-import-csv.service.ts
```diff
@@ -20,8 +20,18 @@ import { CustomNamedError } from "@sims/utilities";
...
+interface InstitutionDetails {
+  operatingName: string;
+  legalOperatingName: string;
+  primaryEmail: string;
+}
+interface LocationDetails {
+  id: number;
+  name: string;
+  institutionDetails: InstitutionDetails;
+}
```
andrewsignori-aot, line 32: Not a blocker, for an object to be consumed internally we could easily keep it simple and use just a flat object.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-validation.models.ts
```diff
@@ -600,6 +603,27 @@ export class OfferingValidationModel {
...
+  /**
+   * Institution location name of the institution that will be associated with this offering.
+   * Used as supporting data, for instance, as part of the notification to indicate offering in pending status.
+   */
+  @IsNotEmpty({ message: "Institution location name is required." })
+  locationName: string;
+  /**
+   * Institution operating name used as supporting data, for instance, as part of the notification to indicate offering in pending status.
+   */
+  @Allow()
```
andrewsignori-aot, line 615: Why this is the only one wth a different decorator?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -731,4 +754,33 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
+  private async saveInstitutionAddsPendingProgramNotification(
+    program: Pick<EducationProgram, "name" | "programStatus">,
+    institutionId: number,
+    entityManager: EntityManager,
+  ): Promise<void> {
+    if (program.programStatus !== ProgramStatus.Pending) {
+      return;
+    }
+    const institution = await this.institutionService.getInstitutionDetailById(
+      institutionId,
```
dheepak-aot, line 773: This is used inside transaction. Please make sure transaction repo is used.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -96,7 +109,23 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+        async (transactionalEntityManager) => {
+          await this.saveEducationProgramOfferingNotification(
+            educationProgramOffering.offeringName,
+            educationProgramOffering.programContext.name,
+            educationProgramOffering.operatingName,
+            educationProgramOffering.legalOperatingName,
+            educationProgramOffering.primaryEmail,
+            programOffering.offeringStatus,
+            programOffering.institutionLocation.name,
+            transactionalEntityManager,
```
dheepak-aot, line 122: Not a blocker, if the arguments for a method is more than 5 it is good to create model to be passed.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -96,7 +109,23 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
     programOffering.parentOffering = programOffering;
 
     try {
-      return await this.repo.save(programOffering);
+      return await this.dataSource.transaction(
```
dheepak-aot, line 112: Are we not missing the offering update part ?? pending offerings can be created during offering update as well

image

sources/packages/backend/libs/services/src/notifications/notification/notification.model.ts
dheepak-aot: More than few interfaces (for personalization) having same properties carrying same functional details seems to be different for me. When the evolution happen we can extend the interface or have individual ones created at the point was what I felt.

Either ways, I wouldn't consider it as a blocker for this PR.

sources/packages/backend/libs/services/src/notifications/notification-message/notification-message.service.ts
```diff
@@ -13,26 +13,26 @@ export class NotificationMessageService extends RecordDataModelService<Notificat
...
-   * Retrieves the template id of a notification message.
+   * Retrieves the notification message details of a notification message.
    * @param notificationMessageTypeId id of the user who will receive the message.
    * @param options options.
    * - `entityManager` external entity manager to run in a transaction.
-   * @returns template Id of the notification message.
+   * @returns notification details of the notification message.
    */
-  async getTemplateId(
+  async getNotificationMessageDetails(
```
andrewsignori-aot, line 22: Please change the method to `private`.

dheepak-aot (commented):
Thanks for doing such a massive amount of work. 👏👏 And also a suggestion to break the work (PR) for the purpose easily of digesting the requirement, betterment of review and reviewers.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -94,9 +108,27 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
...
+    };
     try {
-      return await this.repo.save(programOffering);
+      return await this.dataSource.transaction(
+        async (transactionalEntityManager) => {
+          await this.saveEducationProgramOfferingNotification(
+            educationProgramOfferingNotificationData,
+            transactionalEntityManager,
+          );
+          return transactionalEntityManager
```
andrewsignori-aot, line 127: The return requires the `await` because it is wrapped in the try/catch, right?

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -227,7 +240,16 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
...
       program.modifier = auditUser;
     }
-    return this.repo.save(program);
+    return this.dataSource.transaction(async (transactionalEntityManager) => {
+      await this.saveInstitutionAddsPendingProgramNotification(
+        { name: program.name, programStatus: program.programStatus },
+        institutionId,
+        transactionalEntityManager,
+      );
+      return transactionalEntityManager
```
andrewsignori-aot, line 249: https://github.com/bcgov/SIMS/pull/3122#discussion_r1575462371

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -683,14 +683,25 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+   * `entityManager` entity manager to be used in the transaction.
    * @returns Location retrieved, if found, otherwise returns null.
    */
-  async getInstitutionDetailById(institutionId: number): Promise<Institution> {
-    return this.repo
+  async getInstitutionDetailById(
+    institutionId: number,
+    options?: { entityManager?: EntityManager },
+  ): Promise<Institution> {
+    const repo =
```
andrewsignori-aot, line 694: This can be simplified as below since we are using the base class in the old way for this class.
```ts
const repo =  options?.entityManager?.getRepository(Institution) ?? this.repo;
```

sources/packages/backend/apps/load-test-gateway/src/services/workflow/workflow-data-load.service.ts
```diff
@@ -49,7 +50,7 @@ export class WorkflowDataLoadService {
           student: preliminaryData.student,
           location: preliminaryData.offering.institutionLocation,
         },
-        { initialValue: { data: submittedApplicationData } },
+        { initialValue: { data: submittedApplicationData as ApplicationData } },
```
andrewsignori-aot, line 53: Was this file changed as part of this PR? It seems out of the context.

andrewsignori-aot (commented):
Thanks for making the changes, just a few minor comments left. Please take a look.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
sh16011993: I didn't find a `try/catch` block here. So, didn't add the `await` with the `return`.

sources/packages/backend/apps/api/src/services/student-account-applications/student-account-applications.service.ts
```diff
@@ -90,7 +97,21 @@ export class StudentAccountApplicationsService extends RecordDataModelService<St
...
+      birthDate: studentProfile.dateOfBirth,
+    };
+    return this.dataSource.transaction(async (entityManager) => {
+      await this.notificationActionsService.saveStudentRequestsBasicBCeIDAccountNotification(
+        ministryNotification,
+        entityManager,
+      );
+      return entityManager
+        .getRepository(StudentAccountApplication)
+        .save(newAccountApplication);
```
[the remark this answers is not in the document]
andrewsignori-aot: As much as possible, yes, we would.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
andrewsignori-aot: It was added now, not executing in parallel, but saving.

sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-validation.models.ts
```diff
@@ -664,3 +687,13 @@ export interface CreateValidatedOfferingResult {
   validatedOffering: OfferingValidationResult;
   createdOfferingId?: number;
 }
+
+export interface EducationProgramOfferingNotification {
```
dheepak-aot, line 691: This interface must be in service.models.ts not validation.models.ts

andrepestana-aot (approved):
looks good to me. Great job!

sources/packages/backend/libs/services/src/notifications/notification/notification.service.ts
```diff
@@ -120,6 +121,33 @@ export class NotificationService extends RecordDataModelService<Notification> {
...
+      .getRepository(Notification)
+      .createQueryBuilder("notification")
+      .innerJoin("notification.notificationMessage", "notificationMessage")
+      .where("notificationMessage.id = :notificationMessageId", {
+        notificationMessageId:
+          NotificationMessageType.ApplicationEditedTooManyTimesNotification,
+      })
+      .andWhere(
+        "notification.metadata->>'applicationNumber' = :applicationNumber",
+        {
```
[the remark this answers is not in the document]
dheepak-aot: Assuming that it works based on E2E tests, I would suggest to change to object query in upcoming PR

dheepak-aot (approved):
Great work. Thanks for making the changes @sh16011993 . One minor comment https://github.com/bcgov/SIMS/pull/3122#discussion_r1576660919

PR #3130, opened 2024-04-15
3027 - Investigate Queue-Consumers error in Prod

andrepestana-aot:
- Change to remove `RoleBinding` from `make delete-redis` command as we don't need that in our environments and it causes an error:
image

After the change:
image

- Changed spacing of `storageClassName` to make it be under `spec`. That issue cause an error when testing on TEST env:
image

### DIFF
--- a/devops/Makefile
+++ b/devops/Makefile
@@ -179,7 +179,7 @@ print-status:
 	@echo " +---------------------------------------------------------+ "
 
 init-project:
-	@echo "Run initial set-up for the project, including network security\n" 
+	@echo "Run initial set-up for the project, including network security\n"
 	@oc -n $(NAMESPACE) process -f openshift/networkpolicy.yml -p NAMESPACE=$(NAMESPACE) | oc -n $(NAMESPACE) apply -f -
 	@oc -n $(BUILD_NAMESPACE) policy add-role-to-group system:image-puller system:serviceaccounts:$(NAMESPACE)
 	@oc -n $(NAMESPACE) create rolebinding github-action --clusterrole=edit --serviceaccount=$(BUILD_NAMESPACE):github-action --namespace=$(NAMESPACE)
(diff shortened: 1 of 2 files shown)


devops/Makefile
```diff
@@ -179,7 +179,7 @@ print-status:
 	@echo " +---------------------------------------------------------+ "
 
 init-project:
-	@echo "Run initial set-up for the project, including network security\n"
```
dheepak-aot, line 182: What is the reason behind formatting changes here?

devops/Makefile
```diff
@@ -615,4 +615,4 @@ db-backup-deploy-patroni-delete:
 # Remove redis and resources including secrets from openshift namespace.
 delete-redis:
 	@echo "+\n++ Removing redis and resources on $(NAMESPACE)\n+"
-	@oc delete -n $(NAMESPACE) all,rc,svc,dc,route,pvc,secret,configmap,sa,RoleBinding -l app=redis
```
dheepak-aot, line 618: 👍

devops/Makefile
andrepestana-aot: Comparing the files it seems that there are spaces before the carriage return at some end of lines: 
image
image
image
image

andrewsignori-aot: Checking the file in my local and saving the PR-changed file did not cause additional changes.
It seems that the PR file has the right formatting.

devops/openshift/database/redis-ha-deploy.yml
```diff
@@ -147,7 +147,7 @@ objects:
             resources:
               requests:
                 storage: ${PVC_SIZE}
-          storageClassName: ${STORAGE_CLASS}
+            storageClassName: ${STORAGE_CLASS}
```
andrewsignori-aot, line 150: Just to support the change, `storageClassName` is defined under the specs as per the below doc.
https://docs.openshift.com/container-platform/4.13/rest_api/storage_apis/persistentvolumeclaim-v1.html#spec
Thanks for making the change.

andrewsignori-aot (approved):
Nice work, thanks for the investigation, looks good 👍

devops/openshift/database/redis-ha-deploy.yml
andrepestana-aot: I redid a test on DEV and it worked both ways. 😄 That's good that we tested on TEST.

dheepak-aot, line 150: Thanks for taking care of this. 👍

dheepak-aot (approved):
Great analysis and work 👍

devops/openshift/database/redis-ha-deploy.yml
andrepestana-aot: I went back to the history now and checked that it did happen on DEV but I didn't pay much attention it was just a warning.
image

Issue #2958, opened 2024-03-06
PT Final Award - NOA Summary based on eCert Totals


CarlyCotton:
**User Story**
As a _student_ I want to be able to understand what my final award values are and when they will be available.

As a _ministry user_, I want to be able to tell when the ecert has been sent and what was requested. I also need to be able to direct students to the correct place when they have concerns about the award amount.

**Acceptance Criteria**
- [x] Once the ecert has been submitted for a PART TIME disbursement, change from the "pending" (FIG 1) final award state to the "complete" award state (FIG 2).
- [x] Update the Final award area for **PT** after the ecert has been sent (FIG 2) with new content:
> If you want to confirm the amounts you will receive or when they will be disbursed, please [access](https://protege-secure.csnpe-nslsc.canada.ca/en/public/sign-in-method) your National Student Loan Service Centre account.
- [x] Use the values sent in the ecert disbursement to populate the award amounts in the table (FIG 2). 
- [x] Add a section visible to only ministry users **FOR FT & PT** to display date that the ecert was submitted and the certificate number.

**Technical**
- [x] Do not change Full-time as it is already working as expected retrieving the values from `sims.disbursement_receipt_values` table.
- [x] Change API endpoint to return the same data as it is right now but get the final award values from `sims.disbursement_values.effective_value`.
- [x] "Date eCert Sent"Certificate should be the `sims.disbursement_schedules.date_sent` and should be added under the already existing section that contains the "Certificate Number". Ensure that only Ministry will be able to see it.
- [x] Method [populateDisbursementReceiptAwardValues](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts#L280-L362) should be adapted. Consider creating a new business service to process the final awards for full-time and part-time.
- [x] E2E Tests

**Business Context**
FIG 1:
<img width="334" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFBJQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e5074745fdbb5e2796cd6c289f966ebaeb895f1a/image.png" alt="image.png" />
FIG 2:
[...]

PR #3159, opened 2024-04-19
#2958 - Part-time Final Award From e-Certs - UI

andrewsignori-aot:
## General notes
- The UI page has some components and some duplicated logic already when the application has two disbursements. This behavior has not changed. Some refactors in the future can improve this page.
- Two minor bugs were found from the previous PR, a missing assessment ID in a SQL query and `null` values returned for a second disbursement (the expectation was to not return null awards.)

### Part-time application with two disbursements, the second one waiting for disbursement to be sent.

image

### Part-time application with two disbursements, one sent and one pending.

image

### Part-time application with two disbursements, both sent.

image

### Part-time
- Disbursement pending: "The final award can't be calculated at this time." (previously COE Required)
- Disbursement sent: no message expected since we will display the values.
- COE Declined: "The final award is no longer applicable due to a change. Any scheduled disbursements will be canceled." (Same as before)
 
### Full-time
- Disbursement pending: "The final award can't be calculated at this time." (previously COE Required and same from Part-time)
- Disbursement sent: "The final award will be shown once confirmed by NSLSC."
- COE Declined: "The final award is no longer applicable due to a change. Any scheduled disbursements will be canceled." (Same as before)

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/assessment/_tests_/e2e/assessment.students.controller.getAssessmentAwardDetails.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/assessment/_tests_/e2e/assessment.students.controller.getAssessmentAwardDetails.e2e-spec.ts
@@ -109,6 +109,11 @@ describe("AssessmentStudentsController(e2e)-getAssessmentAwardDetails", () => {
           // Adding date sent to ensure that will not be returned by the API (students should not receive it).
           dateSent: new Date(),
         },
+        secondDisbursementInitialValues: {
+          disbursementScheduleStatus: DisbursementScheduleStatus.Sent,
+          // Adding date sent to ensure that will not be returned by the API (students should not receive it).
+          dateSent: new Date(),
+        },
       },
     );
 
@@ -257,6 +262,7 @@ describe("AssessmentStudentsController(e2e)-getAssessmentAwardDetails", () => {
           disbursementScheduleStatus: DisbursementScheduleStatus.Sent,
         },
         secondDisbursementInitialValues: {
+          disbursementScheduleStatus: DisbursementScheduleStatus.Sent,
           coeStatus: COEStatus.completed,
           tuitionRemittanceRequestedAmount: 9876,
         },
--- a/sources/packages/backend/apps/api/src/route-controllers/assessment/_tests_/e2e/assessment.students.controller.getAssessmentAwardDetails.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/assessment/_tests_/e2e/assessment.students.controller.getAssessmentAwardDetails.e2e-spec.ts
@@ -502,6 +502,7 @@ describe("AssessmentStudentsController(e2e)-getAssessmentAwardDetails", () => {
           dateSent: new Date(),
         },
         secondDisbursementInitialValues: {
+          disbursementScheduleStatus: DisbursementScheduleStatus.Sent,
           coeStatus: COEStatus.completed,
         },
       },
--- a/sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
@@ -293,6 +293,12 @@ export class AssessmentControllerService {
       }
       let index = 1;
       for (const schedule of assessment.disbursementSchedules) {
+        if (
+          schedule.disbursementScheduleStatus !==
+          DisbursementScheduleStatus.Sent
+        ) {
+          break;
+        }
         const awards = this.populateDisbursementReceiptAwardValues(
           disbursementReceipts,
           schedule,
--- a/sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
@@ -306,6 +306,11 @@ export class AssessmentControllerService {
     // values are used instead to provide the best information as possible.
     let index = 1;
     for (const schedule of assessment.disbursementSchedules) {
+      if (
+        schedule.disbursementScheduleStatus !== DisbursementScheduleStatus.Sent
+      ) {
+        break;
+      }
       const awards = this.populateDisbursementECertAwardValues(
         schedule.disbursementValues,
         `${DISBURSEMENT_RECEIPT_PREFIX}${index++}`,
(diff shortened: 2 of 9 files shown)


sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
```diff
@@ -306,6 +306,11 @@ export class AssessmentControllerService {
     // values are used instead to provide the best information as possible.
     let index = 1;
     for (const schedule of assessment.disbursementSchedules) {
+      if (
```
dheepak-aot, line 309: We should have the same validation for full-time while reading from the receipt right?? or am I missing something?

sources/packages/backend/apps/api/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -57,6 +57,7 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
     const query = this.repo
       .createQueryBuilder("assessment")
       .select([
+        "assessment.id",
```
dheepak-aot, line 60: 👍

sources/packages/backend/apps/api/src/route-controllers/assessment/_tests_/e2e/assessment.students.controller.getAssessmentAwardDetails.e2e-spec.ts
```diff
@@ -109,6 +109,11 @@ describe("AssessmentStudentsController(e2e)-getAssessmentAwardDetails", () => {
           // Adding date sent to ensure that will not be returned by the API (students should not receive it).
           dateSent: new Date(),
         },
+        secondDisbursementInitialValues: {
+          disbursementScheduleStatus: DisbursementScheduleStatus.Sent,
+          // Adding date sent to ensure that will not be returned by the API (students should not receive it).
+          dateSent: new Date(),
```
dheepak-aot, line 115: 👍

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
```diff
@@ -57,45 +76,65 @@
...
+                <span class="label-bold">Date eCert sent: </span>
+                <span>{{
+                  dateOnlyLongString(
+                    assessmentAwardData.estimatedAward
+                      .disbursement1DateSent as Date,
+                  )
+                }}</span>
+              </div>
+              <div>
+                <span class="label-bold">Certificate number: </span>
```
dheepak-aot, line 102: Why is the certificate number repeated in final table as well?

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
```diff
@@ -57,45 +76,65 @@
...
           <h3 class="category-header-medium primary-color my-3">Final award</h3>
           <!-- Final award table. -->
-          <award-table
-            :awardDetails="assessmentAwardData.finalAward"
-            :offeringIntensity="assessmentAwardData.offeringIntensity"
-            identifier="disbursementReceipt1"
-            v-if="showFirstFinalAward"
-          />
-          <div v-else>
+          <template v-if="showFirstFinalAward">
```
dheepak-aot, line 85: Is there a reason for using template here instead of div?

sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
andrewsignori-aot: It does not need to be in the same way because the full-time is based in the receipts. We can add a check before the receipt call (not in the for loop) to avoid one DB query when the disbursement is pending.

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
andrewsignori-aot: Is there a reason to use a div here instead of a template 😄 
Unless a `div` is required the template would allow the conditional display in the same way.
The bottom line would be, that we should not add a `div` only for the sake of hiding/showing something. Does it make sense?

dheepak-aot (approved):
Good job! PR looks good to go. 👍

As we discussed, please check on disbursement recent not existing for second disbursement for FT.

sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
andrewsignori-aot: I added the same test inside the `for` for simplicity.

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
```diff
@@ -180,33 +219,78 @@
...
+              <div>
+                <span class="label-bold">Certificate number: </span>
+                <span>{{
+                  assessmentAwardData.estimatedAward.disbursement2DocumentNumber
+                }}</span>
+              </div>
+            </content-group-info>
+            <!-- Part-time final awards are retrieved from e-Cert generated values and requires further explanation.  -->
+            <p
+              v-if="
```
guru-aot, line 247: why cant we have the  check  assessmentAwardData.offeringIntensity === OfferingIntensity.partTime as a consumable and pass it here as isPartTime? and do the check

sources/packages/web/src/services/http/dto/Assessment.dto.ts
```diff
@@ -63,6 +63,26 @@ export interface AssessmentNOAAPIOutDTO {
   applicationStatus: ApplicationStatus;
 }
 
+/**
+ * Dynamic object with awards values.
+ * @example
+ * {
+ *    disbursementReceipt1cslf: 1000,
```
guru-aot, line 70: 👍

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
andrewsignori-aot: By `consumable` do you mean `computed`?
If yes, I did not see a benefit for this case. Please let me know if it is a blocker.

guru-aot (approved):

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
andrewsignori-aot: No difference in the reactivity in this case.
The computed would be better to not target the part-time/full-time for this case, IMO.
Either way at this stage I did not see it as a great improvement. Maybe if more logic is added on top of it.

sources/packages/web/src/components/common/AssessmentAwardDetails.vue
andrewsignori-aot: Please take a look in the official documentation from Vue from where I got the point.
https://vuejs.org/guide/essentials/conditional.html#v-if-on-template.

_Because v-if is a directive, it has to be attached to a single element. But what if we want to toggle more than one element? In this case we can use v-if on a <template> element, which serves as an invisible wrapper. The final rendered result will not include the <template> element._

lewischen-aot (approved):
Nice work @andrewsignori-aot LGTM 👍

Issue #2589, opened 2023-12-12
Use a CSV parser library for the report service


andrepestana-aot:
Analyse the possibility of adding a CSV parser library and implement it to our report service:

https://github.com/bcgov/SIMS/blob/5854e3d9951f5fdf99d7cbc90f72c03dc285c07e/sources/packages/backend/libs/services/src/report/report.service.ts#L96-L115

One of the problems to be resolved is to allow titles and values to contain commas. That can be solved double quoting the csv values.

**Acceptance Criteria**
- [x] Check the options for CSV parser libraries available and pick one with the team; 
- [x] Check the existing library for the bulk upload. 
- [x] Add the parser to the report service;
- [x] Test the report creating titles and values with commas to test it;

michesmith:
@andrepestana-aot- can you please create a first draft Acceptance Criteria for grooming please?

andrewsignori-aot:
### Demo Evidences

Extracted a report from DEV using the new lib and a user with quotes and a comma in the body.
<img width="776" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNm1hQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6677e4994aa17177134e3e33cebc8acfe595f43b/image.png" alt="image.png" />

<img width="1667" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnFhQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c3a36f9fdb8f95c4bfcc02cfe9b75c63e33fa599/image.png" alt="image.png" />

PR #3194, opened 2024-04-25
#2589 - Use a CSV parser library for the report service

andrewsignori-aot:
The backend is currently using the [papaparse ](https://www.npmjs.com/package/papaparse) lib to read CSVs from the bulk file uploads and the same library has a method to convert objects to CSV ([unparse](https://www.papaparse.com/docs)).

Used the `Data Inventory Report` as a POC, changing a user first name to a string with characters that would break the CSV generation.

- Input string: `Testing "quotes" and comma, and more; characters, and values $45.123,21!`
- Escaped string: `"Testing ""quotes"" and comma, and more; characters, and values $45.123,21!"`

#### Generated Report
image

#### CSV escaped string
image

#### Same report without using the lib
image

#### Header is also able to support quotes and other characters

image

Typeorm will produce an object as below. The property names are enclosed in single or double quotes depending on their content. Please see the difference between `'Application Number'`, `'First Name with quotes " in the name and ,coma'`, and  `Gender` (this explanation is not part of the PR and it was added here just for curiosity).
```ts
 {
    'Application ID': 73,
    'Application Number': '5007443876',
    'Application Status': 'Completed',
    'Assessment Date': null,
    'Re-assessment Indicator': 'Original assessment',
    "Marital Status (or 'category')": 'SI',
    'Student Dependency Status': 'dependant',
    'First Name with quotes " in the name and ,coma': 'Maud',
    'Last Name': 'Robel',
    'Date of Birth': 2018-08-01T00:00:00.000Z,
    Gender: 'X',
    SIN: '428314025',
}
```

##### Notes
1. lib was updated to the latest version to take advantage of the opportunity to use the latest package version.
2. The downloaded file is encoded in UTF-8 as the result of writing it to the stream, no changes in this behavior since the lib is currently converting the object to a CSV string that is later written to the desired output.

### DIFF
--- a/sources/packages/backend/libs/services/src/report/report.service.ts
+++ b/sources/packages/backend/libs/services/src/report/report.service.ts
@@ -4,6 +4,8 @@ import { RecordDataModelService, ReportConfig } from "@sims/sims-db";
 import { CustomNamedError, StringBuilder } from "@sims/utilities";
 import { ReportsFilterModel } from "./report.models";
 import { REPORT_CONFIG_NOT_FOUND, FILTER_PARAMS_MISMATCH } from "./constants";
+import { unparse } from "papaparse";
+
 /**
  * Service layer for reports.
  */
@@ -49,7 +51,7 @@ export class ReportService extends RecordDataModelService<ReportConfig> {
       }
     });
     const reportData = await this.connection.query(reportQuery, parameters);
-    return this.buildCSVString(reportData);
+    return unparse(reportData);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


sources/packages/backend/libs/services/src/report/report.service.ts
```diff
@@ -4,6 +4,8 @@ import { RecordDataModelService, ReportConfig } from "@sims/sims-db";
 import { CustomNamedError, StringBuilder } from "@sims/utilities";
```
dheepak-aot, line 4: Please remove the unused import. 

image

sources/packages/backend/libs/services/src/report/report.service.ts
```diff
@@ -4,6 +4,8 @@ import { RecordDataModelService, ReportConfig } from "@sims/sims-db";
...
 import { ReportsFilterModel } from "./report.models";
 import { REPORT_CONFIG_NOT_FOUND, FILTER_PARAMS_MISMATCH } from "./constants";
+import { unparse } from "papaparse";
```
dheepak-aot, line 7: ❤️

dheepak-aot (approved):
Great Job. Thanks for detailed explanation in the PR and upgrading the papaparse version. 👍
Just one minor comment on import.

sources/packages/backend/libs/services/src/report/report.service.ts
andrewsignori-aot: Removed.

PR #3188, opened 2024-04-25
Fix Webpack Config and Enable Multiple Debug

andrewsignori-aot:
Over the past weeks, the debug was not working properly due to the `devtool` configuration (that is the assumption). The configuration we have right now is the [right one for production](https://webpack.js.org/configuration/devtool/#devtool) and to produce the correct stack trace for errors but it was preventing the debugging from properly working. 
A further investigation was not done and the current change is not supposed to affect production, hence it is intended only to quickly bring back the debug functionality.

Besides that, new configurations were added to allow the debugging of the backend applications. Below are the benefits.
- Start the application with the debugger attached.
- Allow application modification keeping the debugger attached.
- Allow the debugging of multiple applications at the same time.

The below image display `api`, `workers`, and `queue-consumers` executed at the same time with the debugger attached to all of them.

image

PR #3202, opened 2024-04-27
Queue-consumers Fail to Start 

andrewsignori-aot:
The PR #3188 introduced an issue where the SQL files expected to be copied by the `copy-webpack-plugin` during the `webpack` build are no longer copied.
The current check was failing because the ENV variable was not present at "build time" and new syntax was also affecting the `plugins` configuration.

A new env variable (`DEBUG`) was introduced only to allow the `webpack` to be aware of the `devtools` expected value.

### DIFF
--- a/sources/packages/backend/webpack.config.js
+++ b/sources/packages/backend/webpack.config.js
@@ -1,5 +1,9 @@
 const CopyPlugin = require("copy-webpack-plugin");
-const exports = {
+module.exports = {
+  // Enable the source maps needed to support proper stack trace errors in production
+  // pointing to typescript files (.ts) instead of javascript (.js) references.
+  // DEBUG is set as 'true' only when using start:debug' script.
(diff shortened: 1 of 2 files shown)


sources/packages/backend/webpack.config.js
```diff
@@ -1,5 +1,9 @@
 const CopyPlugin = require("copy-webpack-plugin");
-const exports = {
+module.exports = {
+  // Enable the source maps needed to support proper stack trace errors in production
+  // pointing to typescript files (.ts) instead of javascript (.js) references.
+  // DEBUG is set as 'true' only when using start:debug' script.
+  devtool: process.env.DEBUG === "true" ? undefined : "source-map",
```
dheepak-aot, line 6: 👍

dheepak-aot (approved):
Thanks for taking care of the issue 👍Looks good.

Issue #3062, opened 2024-04-02
FT + PT Reports - Enable Reports to BC Publics, Build Offering Details Report


HRAGANBC:
**User Story**
As a **BC Public** institution **Admin** user, I need to be able to pull reports about my institution, offerings and students through SIMS. so that I can determine which students have unmet need as Institutions have additional programs that they can provide to students ontop of SABC awards to help students with unmet need. 

**Acceptance Criteria**
- [x] Create a Reports link that is only accessible by Admin users of BC Public Institutions (see link bottom left on second AC).
- [x] Create the page that allows the approved user to select the report type and indicate the filterable field(s):
<img width="1014" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1UxQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7779ab03a72fa93888e6897dc38e74bf88f708b0/image.png" alt="image.png" />
- [x] Program Year is a required drop-down filtering field.
- [ ] Intensity is a required drop-down filtering field with options: Full-time Only, Part-time Only, Full- and Part-time. Treat Full-time as FT Only (Program and Offering), Part-time as FT/PT Program and PT Offering, Full- and Part-time will include anything FT Only & FT/PT (Program) and all offerings. 
- [x] Reports available will have mandatory filters to only show records that are related to their institution code and the students that have applications at their institution. These filters are not shown to the institution user, but should align with what the institution users are able to see when searching for student information manually.
 - [ ] Add report option: Offering Details (see Sample Reports tab in the file linked below, under report titled "Offering Details")
    - Note: Include offerings that could potentially have applications with the following statuses (Creation pending, Approved, Change under review). 
    - Note re Comment from 2024-05-09: Offerings should include Public and Private, not scholastic standing types.
    - Note: Count applications that have been **completed**.

**Business Context**
BC Public institution users have access to their own institution data as well as to student information when they have applied to their offerings/programs. This reporting is to support any manual work on the institution side.
See "Sample Reports" tab of [SIMS Reports & Notification Triggers](https://bcgov.sharepoint.com/:x:/r/teams/03177/_layouts/15/Doc.aspx?sourcedoc=%7B08758022-BBDE-485E-86F9-ED065D92BCFE%7D&file=SIMS%20Reports%20&%20Notification%20Triggers.xlsx=&action=default&mobileredirect=true))

[...]

PR #3205, opened 2024-04-30
#3062 - Enable Reports for BC Public Institutions

sh16011993:


sources/packages/web/src/components/common/Reports.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <full-page-container>
```
andrewsignori-aot, line 2: `<full-page-container>` hardly will be added at the component level. Are we sure about having it shared at this level?

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -0,0 +1,70 @@
...
+    <body-header title="Export reports" />
+    <formio
+      formName="exportfinancialreports"
+      @loaded="formLoaded"
+      @submitted="exportReport"
+    ></formio>
+    <v-row class="justify-center m-4">
+      <check-permission-role :role="Role.AESTReports">
+        <template #="{ notAllowed }">
+          <v-btn color="primary" @click="submitForm" :disabled="notAllowed"
```
andrewsignori-aot, line 15: Can we take the opportunity to add a simple loading to the button?
Righ now there is no feedback at all to the user and the report loading is a long operation.

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -0,0 +1,70 @@
...
+import { FormIOForm, Role } from "@/types";
+import CheckPermissionRole from "@/components/generic/CheckPermissionRole.vue";
+
+export default defineComponent({
+  components: {
+    CheckPermissionRole,
+  },
+  props: {
+    title: { type: String, required: true },
+    reportType: { type: Number, required: true },
```
andrewsignori-aot, line 41: The property should be exposed as the enum that it represents, not the number.

andrewsignori-aot, line 41: This component can receive the list of `OptionItemAPIOutDTO` that is expected to be loaded by each consumer (Ministry or institution).

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -0,0 +1,70 @@
...
+  },
+  setup(props) {
+    const snackBar = useSnackBar();
+    const fileUtils = useFileUtils();
```
andrewsignori-aot, line 45: Not a blocker but I would say that majority of types we consume the functions needed from the composables as below.
```ts
const { downloadReports } = useFileUtils();
```

sources/packages/web/src/components/layouts/Institution/sidebar/ManageInstitutionSideBar.vue
```diff
@@ -94,6 +94,15 @@ export default defineComponent({
           },
         },
       });
+      sidebarBottomItems.value.push({
```
andrewsignori-aot, line 97: I though that the menus were using `mdi` icons but as we can see in this file all icons are from `fa`. Please use the `fa` icon.

sources/packages/web/src/components/common/Reports.vue
sh16011993: @dheepak-aot mentioned it as well when we were debugging the code but I missed to change it. 😅 Will move it outside.

sources/packages/web/src/components/common/Reports.vue
sh16011993: I removed it since after passing the list, I don't need it anymore, atleast for now.

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -0,0 +1,80 @@
...
+} from "@/composables";
+import { FormIOForm, Role } from "@/types";
+import CheckPermissionRole from "@/components/generic/CheckPermissionRole.vue";
+
+export default defineComponent({
+  components: {
+    CheckPermissionRole,
+  },
+  props: {
+    title: { type: String, required: true },
```
andrewsignori-aot, line 46: `title` is no longer used and can be removed.

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -0,0 +1,80 @@
...
+    const formLoaded = async (form: FormIOForm) => {
+      formData = form;
+      await formioDataLoader.loadReportTypes(
+        formData,
+        REPORT_TYPE_DROPDOWN_KEY,
+        props.reportList,
+      );
+    };
+    const submitForm = () => {
+      loading.value = true;
```
andrewsignori-aot, line 65: The loading should go to the API call right?
```ts
const exportReport = async (data: ReportsFilterAPIInDTO) => {
  try {
    loading.value = true;
    await downloadReports(data);
  } catch {
    snackBar.error("Unexpected error while downloading the report.");
  } finally {
    loading.value = false;
  }
};
```

sources/packages/web/src/components/common/Reports.vue
andrewsignori-aot: Not a blocker, for this component and this feature it may not be a problem but even the `body-header-container` should be avoid sometimes.

sources/packages/web/src/components/layouts/aest/AESTHomeSideBar.vue
```diff
@@ -21,7 +21,7 @@
           <template #="{ notAllowed }">
             <v-list-item
               :to="{ name: AESTRoutesConst.REPORTS }"
-              prepend-icon="mdi-home-outline"
+              prepend-icon="fa:fa-regular fa-copy"
```
andrewsignori-aot, line 24: Unfortunately, institutions and the Ministry are using different different libs. Please keep using the same lib for the file.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -189,6 +189,14 @@ export function useFormioDropdownLoader() {
     );
   };
 
+  const loadReportTypes = async (
```
andrewsignori-aot, line 192: What is the benefit of having the method `loadReportTypes`? It looks like a rename from `loadDropdown`?

sources/packages/web/src/views/institution/Reports.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <full-page-container>
+    <template #header
+      ><header-navigator title="Institution" subTitle="Reports" />
+    </template>
+    <Reports title="Institution" :reportList="reportList" />
```
andrewsignori-aot, line 6: Please use it as `reports` and the property as `report-list`, as below.
We have mixed patterns but the below-mentioned one is preferred.
```vue
<reports title="Institution" :report-list="reportList" />
```

sources/packages/web/src/views/aest/Reports.vue
```diff
@@ -1,58 +1,30 @@
...
-    <v-row class="justify-center m-4">
-      <check-permission-role :role="Role.AESTReports">
-        <template #="{ notAllowed }">
-          <v-btn color="primary" @click="submitForm" :disabled="notAllowed"
-            >Export CSV file</v-btn
-          >
-        </template>
-      </check-permission-role></v-row
-    >
+    <Reports title="Ministry" :reportList="reportList" />
```
andrewsignori-aot, line 6: Same from https://github.com/bcgov/SIMS/pull/3205/files#r1588521954

andrewsignori-aot (commented):

sources/packages/web/src/components/common/Reports.vue
sh16011993: Removed it.

sources/packages/web/src/components/common/Reports.vue
sh16011993: Yes, that's correct.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -28,7 +28,7 @@
         </template>
       </header-navigator>
     </template>
-    <StudentApplication
+    <student-application
```
andrewsignori-aot, line 31: Please remove this file from the PR.

andrewsignori-aot (approved):
Nice work and thanks for adding the button loading.
Please remove the non-related file change in the PR.
https://github.com/bcgov/SIMS/pull/3205/files#r1589658904
Look good 👍

Issue #3237, opened 2024-05-07
Cleanup sims-db-migrations ImageStream versions


sslaws:
As a SIMS team, we need to adhere to Platform policy and cleanup old versions from our ImageStreams.

### **Acceptance Criteria**
- [ ] Enhance GH Action to cleanup db-migrations-sims
- [ ] Manually run scripts to cleanup n-4 versions of formal release tags
- [ ] Manually run scripts to remove any scenario based tags
- [ ] Manually run scripts to remove all versions from main older than the Production release

PR #3238, opened 2024-05-07
#3237 - Enhance prune images to support either DCs or Jobs

sslaws:
Enhanced PruneImages by adding a new type and dynamically calling fetch script.
Created new fetchOldJobTags script which looks at the job isntead of the DC.
Updated the GH Action to run both types of prunes.

### DIFF
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -14,9 +14,13 @@ on:
           - prod
         default: "test"
       applications:
-        description: "Comma seperated list of applications to prune"
+        description: "Comma seperated list of application Image Streams to prune"
         required: true
         default: "web-sims, api-sims, queue-consumers-sims, workers-sims"
+      jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/prune-images.yml
```diff
@@ -62,4 +67,11 @@ jobs:
               --apps="${APPLICATIONS}" \
               --prefix=${PREFIX} \
               --min_tags=${MIN_TAGS}
+          ./pruneImages.sh \
```
andrewsignori-aot, line 70: As a suggestion, if DCs and Jobs pruning commands are executed in different Github Actions steps, it will provide better visual feedback to the action execution.

andrewsignori-aot (approved):
Thanks for adjusting the current solution do take care of the jobs also. Looks god 👍

dheepak-aot (approved):
Looks good @sslaws 👍

PR #3264, opened 2024-05-10
Rename variable and initialize properly

sslaws:
The new jobs pruning portion of the script was failing as the variable was not being initialized properly.
I renamed the variable jobs as it is valid yaml syntax and I wanted to be sure we had no conflicts.

### DIFF
--- a/.github/workflows/prune-images.yml
+++ b/.github/workflows/prune-images.yml
@@ -17,7 +17,7 @@ on:
         description: "Comma seperated list of application Image Streams to prune"
         required: true
         default: "web-sims, api-sims, queue-consumers-sims, workers-sims"
-      jobs:
+      ocjobs:
           description: "Comma seperated list of job Image Streams to prune"
           required: true
           default: "db.migrations"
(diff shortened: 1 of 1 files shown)


.github/workflows/prune-images.yml
```diff
@@ -17,7 +17,7 @@ on:
         description: "Comma seperated list of application Image Streams to prune"
         required: true
         default: "web-sims, api-sims, queue-consumers-sims, workers-sims"
-      jobs:
+      ocjobs:
```
andrewsignori-aot, line 20: Can it follow the same camel case pattern from `minTags` which would be `ocJobs`?

.github/workflows/prune-images.yml
```diff
@@ -40,14 +40,15 @@ jobs:
         run: |
           echo "ENVIRONMENT=${{ inputs.environment || 'test' }}" >> $GITHUB_ENV
           echo "APPLICATIONS=${{ inputs.applications || 'web-sims, api-sims, queue-consumers-sims, workers-sims' }}" >> $GITHUB_ENV
+          echo "OCJOBS=${{ inputs.ocjobs || 'db.migrations' }}" >> $GITHUB_ENV
```
andrewsignori-aot, line 43: Can it follow the same naming pattern from `MIN_TAGS`?

andrewsignori-aot (approved):
Just added minor comments. Looks good 👍

dheepak-aot (approved):
Looks good @sslaws 👍

Issue #3055, opened 2024-03-28
Remove API Key prefix in GC Notify Service


dheepak-aot:
**Describe the task**
Remove the hard-coded prefix `ApiKey-v1` from the code while calling the GC Notify API.

**Acceptance Criteria**
- [x] Remove the hard coded prefix from the code https://github.com/bcgov/SIMS/blob/b58f62b0a4aa39fcd30ee83c6067e846e0225e1d/sources/packages/backend/libs/services/src/notifications/notification/gc-notify.service.ts#L42
- [x] Add the prefix `ApiKey-v1` to Github secret `GC_NOTIFY_API_KEY` , deploy secrets and deploy the applications

**Technical**
- [x] Remove the hard-coded `ApiKey-v1` from the `Authorization` header.
- [x] Add the `ApiKey-v1` to DEV and TEST env variables on GitHub envs, deploy secrets, and restart rollout for the `queue-consumers`.
  - While adding the secret to GitHub the value should be added including single quotes.

PR #3266, opened 2024-05-10
#3055 - Remove api key prefix

katrina-aot-git:
Remove the api key prefix from the gc-notify section.


dheepak-aot (approved):
Looks good 👍

Issue #3263, opened 2024-05-10
Refactor Code for Mass Offering Change Request


lewischen-aot:
Story Description
As a developer, I want to refactor existing code for the method [`getApplicationsToSubmitReassessment`](https://github.com/bcgov/SIMS/blob/4fdc2bfef590d20c0802487b09a33b22d5578943/sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts#L1174) to optimize the query.

**Technical Criteria**
- [x] Remove properties "hasAssessmentData", "workflowName", "assessmentWorkflowId" from select.
- [x] Remove property "assessmentAppealId" but keep "studentAppeal.id" from select.
- [x] Remove class "ApplicationAssessmentSummary" and method returns Application array.
- [x] Add `studentAppeal`property to [`application.currentAssessment`](https://github.com/bcgov/SIMS/blob/4fdc2bfef590d20c0802487b09a33b22d5578943/sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts#L1270)

PR #3270, opened 2024-05-13
#3263 - Refactor Code for Mass Offering Change Request

lewischen-aot:
- Removed properties "hasAssessmentData", "workflowName", "assessmentWorkflowId" from select.
- Removed property "assessmentAppealId" but keep "studentAppeal.id" from select.
- Removed class "ApplicationAssessmentSummary" and method returns Application array.
- Added `studentAppeal` property to `application.currentAssessment` if applicationStatus is Completed.

### DIFF
--- a/sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
+++ b/sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
@@ -18,11 +18,9 @@ import {
   StudyBreak,
   ProgramYear,
   DatabaseConstraintNames,
-  mapFromRawAndEntities,
   StudentAssessmentStatus,
   StudyBreaksAndWeeks,
   isDatabaseConstraintError,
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering.service.ts
```diff
@@ -1173,17 +1170,14 @@ export class EducationProgramOfferingService extends RecordDataModelService<Educ
    */
   async getApplicationsToSubmitReassessment(
     offeringId: number,
-  ): Promise<ApplicationAssessmentSummary[]> {
-    const applications = await this.dataSource
+  ): Promise<Application[]> {
+    return await this.dataSource
```
andrewsignori-aot, line 1174: `await` is not needed here.

andrewsignori-aot (approved):
Thanks for doing the refactor. Looks good 👍

Issue #3216, opened 2024-05-02
Additional Transportation weeks field to be made mandatory (PT and FT)


Joshua-Lakusta:
As a ministry user we want the weeks field to be mandatory in the additional transportation question so that we receive the correct information and the calculation works properly.

**Acceptance Criteria**
- [x] Make the "How many weeks during your study period will you require additional transportation costs?" field should be a mandatory field. PartTime and FullTime

**Additional Context**
<img width="1348" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMSt4QkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--83bfa26681a6ff3250a39f6fac32026bc7051ed2/image.png" alt="image.png" />

PR #3273, opened 2024-05-13
#3216 - Make transport weeks a required field

katrina-aot-git:
Make the number of weeks of additional transport costs a required field.


dheepak-aot (approved):
Looks good 👍

Issue #3111, opened 2024-04-10
 4.1.2 Profile Button Student- Accessibility Fixes - Level A -Name, Role, Value


michesmith:
**User Story**
**As a student, I need the profile avatar button to contain a accessible label so that...

**Acceptance Criteria**
- [ ] Add aria-label to profile button. "Account" 

<img width="394" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2xuQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a45bdda718181d5836eeef7815f114c311fdd36f/image.png" alt="image.png" />

**Context:**
- These issues were uncovered through accessibility audit via CITZ. 
- Detailed Spreadsheet for Student s inserted inline and also attached here: [2024_Student Application_Audit.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0xsQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--578821d4addd12c1babcafba5c861e1f5456bc9c/2024_Student%20Application_Audit.xlsx)
- For all user interface components (including but not limited to: form elements, links and components generated by scripts), the name and role can be programmatically determined; states, properties, and values that can be set by the user can be programmatically set; and notification of changes to these items is available to user agents, including assistive technologies.

**The Problem**
Profile avatar does not contain accessible label.

michesmith:
I will get more clarification from Rowan on what the Accessible Text should be

PR #3278, opened 2024-05-14
#3111 - Add label to account button

katrina-aot-git:
Add accessible text labels to account buttons.


andrewsignori-aot (approved):
Looks good 👍

Issue #3040, opened 2024-03-25
Revise Scholastic Standing Functionality and Content


HRAGANBC:
**User Story**
As an institution user, I want the Scholastic Standing reporting options to fully align with policy, and to provide better wording around each option so I am aware which option is more appropriate to use. 

**Acceptance Criteria**
- [ ] **UI Label Changes**
     - change "Student Withdrew From Program" to "Student Withdrew from funded period" 
     - change "Student completed program early" to "Student completed funded study period offering early" 
- [ ] **Display Different Label Conditional on PT or FT application**
    - If FT app: "Student did not complete program" should become "Student was unsuccessful in completing their funded study period offering"
    - If PT app: "Student did not complete program" should become "Student did not complete courses they were funded for"
- [ ] **Remove Label and Back-End Process for "Student went from Full-Time to Part-Time..."**
- [ ] **Add new option** 
    - label should read "Student is transferring to a different institution or campus"
    - sub-question should ask "Last day the student is attending your institution" (ignore this mockup which says something different)
    - response to sub-question should become new end date during re-assessment. Impact should function identically to "completed early", i.e. re-assess with new end date, but apply no restrictions or other penalties. 
    - store this event in database as a distinct option from "completed early" (so they can be distinguished in future reporting)
- [ ] Add accordion with help text beneath the "Report a Change" header. See Fig 2 for text
- [ ] Adjust the sub-question of the "completed early" to be consistent with the mock-up below. 
- [ ] Change IER file mapping:
    - [ ] Remove "CI" as a possible value
    - [ ] Add "ST" to denote 'School Transfer"
    - [ ] Change the IER file specs document accordingly

**Mock-Ups**
**Fig 1**
<img width="903" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMXhoQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b41dc3bc1ace515c45f87eec31f3a278408d7903/image.png" alt="image.png" />

**Fig 2**
<img width="695" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOUlDQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--32e21b73461018df740d56f228645b4a6ded948f/image.png" alt="image.png" />

[...]

PR #3330, opened 2024-05-23
#3040 - Allow the first disbursement to be not present

andrewsignori-aot:
Reassessments were not displaying a card while the assessment was `In progress` because there was an expectation that at least the first disbursement would always be present, which is true only when the assessment workflow is completed.

The UI was already completely ready expecting the `first disbursement` not to be present, but the API was throwing an error every time.

### Current UI without the fix

image

### UI after the fix

image


dheepak-aot (approved):
Looks good. Thanks for taking care of the fix.

Issue #3283, opened 2024-05-15
Nodejs Apps Memory Adjustments


andrewsignori-aot:
The SIMS Api demonstrated some degradation on TEST where a preliminary analysis indicated that it may be caused by memory being consumed and not well handled leading the POD to respond with a major delay causing timeouts in the Web application request (e.g. related ticket #3224).

**Technical**
- [ ] Increase the memory allocated to `SIMS-Api`, `queue-consumers`, and `workers`.
  - Suggested
    - MEMORY_LIMIT: from 256M to 512M
    - MEMORY_REQUEST: from 128M to 384M
- [ ] Edit the `HorizontalPodAutoscaler` memory for `SIMS-Api`, `queue-consumers`, and `workers`.
  - Update `AverageValue` from 230M to 350M
- [ ] Add the `--max-old-space-size=250` to all prod scripts.
  -  "start:prod:api": "node --enable-source-maps dist/apps/api/main.js --max-old-space-size=250"
  -  "start:prod:workers": "node --enable-source-maps dist/apps/workers/main.js  --max-old-space-size=250",
  -  "start:prod:queue-consumers": "node --enable-source-maps dist/apps/queue-consumers/main.js  --max-old-space-size=250",

**Additional context**

Current memory consumption on PROD on May 15th, 2024.
- limits: 26.8%
- requests: 18.9%

Current SIMS-Api memory on PROD on May 15th, 2024.
<img width="903" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelRrQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--73698f2977200a6e7552d10cbb982c2773e49853/image.png" alt="image.png" />

<img width="780" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemJrQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7f51e3269795ace4fad723ed7425a11a2bb62fa7/image.png" alt="image.png" />

andrewsignori-aot:
Below is an article with some additional information on the default Nodejs max memory.

[...]

PR #3338, opened 2024-05-27
#3283 - Nodejs Apps Memory Adjustments

andrewsignori-aot:
- Adjusted memory configurations for `api`, `queue-consumers`, and `workers`.
- Added the same memory limit to docker-compose.


devops/openshift/api-deploy.yml
```diff
@@ -252,7 +252,7 @@ objects:
             name: memory
             target:
               type: AverageValue
-              averageValue: "230M"
+              averageValue: "350M"
```
guru-aot, line 255: Doesnt the average value of the HPA should be 9% of the Memory_Limit ? else it will create replicas when the memory request starts a pod from 384M. Or can you please explain, if my understanding is wrong. This applies to the setting below for queues and workers too.

guru-aot (commented):
A minor clarification, other than that i am good thanks.

devops/openshift/api-deploy.yml
andrewsignori-aot: I am not sure if I am following but the PODs will not "request" the memory up to the "requested" amount. The request is what Openshift will secure to the POD, not necessarily what it will take right from the start. Please let me know if further clarification is needed.

guru-aot: Thanks for the explanation in the call.

guru-aot (approved):

Issue #2807, opened 2024-02-07
Revise Logic for Student Accepting Assessments/Re-assessments


HRAGANBC:
Related to #2665 

**User Story**
As a Student, I must accept an assessment for each application so I understand what I'm agreeing to by sending it to the institution for confirmation.

**Description**
Revise the logic such that students will need to accept **at least one** assessments OR re-assessments, but any assessments or re-assessments generated after that, the NoA does not require accepting. 

**Acceptance Criteria** 
- [x] Require student to accept at least one NoA for every application, but anything after that (prior to complete status) does not require approval. 
- [x] For assessments where the NoA was already approved, and the application is _not_ completed, the workflow should set application status to 'enrolment'
- [x] If a student has not accepted ANY assessment yet and has multiple assessments: Only allow acceptance on the most recent assessment. Remove the ability to accept on the prior assessments.

**Current Behavior**
An assessment requires accepting; re-assessments do not. 
This leaves possibility of an automatic re-assessment happening prior to COE and immediately changing that NoA to 'accepted'. 
<img width="1041" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1E4QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--70751680e454ba9a24e88a06689a78173c3cec19/image.png" alt="image.png" />

User Impact: Student has to potentially accept their assessment multiple times if automatic re-assessments or applications edits occur prior to enrolmet stage. Flip side is that student _does_ need to accept their assesment if there was an automatic re-assessment due to Program Year maximum spillovers

**Technical**

- [x] Load a new variable into Camunda (`load-assessment-consolidated-data`) to define if an NOA was ever accepted by the student for that application.
- [x] Adjust the NOA update in Camunda to reflect the suggested flow below.
<img width="1617" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3hVQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3ef63bb043410ece59445e69f3de08698b1e5764/image.png" alt="image.png" />
- [x] Remove the obsolete data being saved, as shown below.
[...]

PR #3360, opened 2024-05-30
#2807 - Cast to int in router for assessment page

katrina-aot-git:
Follow up to #2807 to add parsing the number for the id.

### DIFF
--- a/sources/packages/web/src/views/student/NoticeOfAssessment.vue
+++ b/sources/packages/web/src/views/student/NoticeOfAssessment.vue
@@ -77,13 +77,13 @@ export default defineComponent({
     const assessmentDataLoaded = (
       applicationStatus: ApplicationStatus,
       noaApprovalStatus: AssessmentStatus,
-      assessmentId: number,
+      currentAssessmentID: number,
     ) => {
       viewOnly.value = !(
         applicationStatus === ApplicationStatus.Assessment &&
         noaApprovalStatus === AssessmentStatus.required
       );
-      currentAssessmentId.value = assessmentId;
+      currentAssessmentId.value = currentAssessmentID;
     };
 
     const confirmAssessment = async () => {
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/views/student/NoticeOfAssessment.vue
```diff
@@ -77,13 +77,13 @@ export default defineComponent({
...
       noaApprovalStatus: AssessmentStatus,
-      assessmentId: number,
+      currentAssessmentID: number,
     ) => {
       viewOnly.value = !(
         applicationStatus === ApplicationStatus.Assessment &&
         noaApprovalStatus === AssessmentStatus.required
       );
-      currentAssessmentId.value = assessmentId;
+      currentAssessmentId.value = currentAssessmentID;
```
andrewsignori-aot, line 86: Please adjust the case in the "ID", it should be `currentAssessmentId`.

andrewsignori-aot, line 86: Missed in the previous PR but the below also need to be adjusted to remove a warning.
```ts
 assessmentDataLoaded: (
      applicationStatus: ApplicationStatus,
      noaApprovalStatus: AssessmentStatus,
      currentAssessmentId: number,
    ) => {
      return (
        !!applicationStatus && !!noaApprovalStatus && !!currentAssessmentId
      );
    },
```
https://github.com/bcgov/SIMS/blob/9c9cbf39c0cb08c2ca3587a62f2c51b87356c0a3/sources/packages/web/src/components/common/NoticeOfAssessmentFormView.vue#L81-L87

katrina-aot-git: I've got this now.

katrina-aot-git: The reason I kept making different names (like the previous currAssessmentId) was to keep the three variables named distinctly. I'd changed this one to be different from the other assessmentId on the page, and different than currentAssessmentId so there wasn't currentAssessmentId.value = currentAssessmentId. I can change it back to assessmentId as that overlap at least is slightly less confusing.

andrewsignori-aot (approved):
Thanks for making the changes and the fix on the routes, look good 👍

Issue #3096, opened 2024-04-10
2.1.1 Menu Navigation- Student- Accessibility Fixes - Level A - Keyboard


michesmith:
**User Story**
**As a student using a screen reader, I need to be able to tab through each of the menu items and navigate to the appropriate page. 

**Acceptance Criteria**

- [x] Update student portal menu so that all items can be tabbed through. Currently, after the menu is opened, tabbing closes the menu. 

Potential solution: Change "tabindex" to "0" from "-2" 
<img width="1194" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3ZHQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b0795f007d02012973eac027030448716ec537bf/image.png" alt="image.png" />

**Context:**
- Theses  issues were uncovered through accessibility audit via CITZ. 
- Detailed Spreadsheet for Student s inserted inline and also attached here: [2024_Student Application_Audit.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0xsQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--578821d4addd12c1babcafba5c861e1f5456bc9c/2024_Student%20Application_Audit.xlsx)
- All functionality of the content is operable through a keyboard interface without requiring specific timings for individual keystrokes, except where the underlying function requires input that depends on the path of the user's movement and not just the endpoints.

**The Problem**
- Menu navigation requires arrow keys which is not intuitive for all users. Changing the tabindex attribute from -2 to 0 or -1 should maintain tab key functionality.

**Student** 
Adjust the following Accessibility Criteria for the following  pages 
- [x]  Login 1 and 2
- [x]  Profile 
- [x] Home page 
- [x] Start new application 
- [x] My applications 
- [x] File uploader 
- [x] Request a change 
- [x] Overawards balance 

**Technical**
- [x] The target is only the Student top nav menu (the same as the business AC).
- [x] Reuse the same solution from the ticket #3141.

HRAGANBC:
@Joshua-Lakusta to complete AC's no earlier than May 1

PR #3363, opened 2024-05-31
#3096 - 2.1.1 Menu Navigation- Student- Accessibility Fixes - Level A - Keyboard

lewischen-aot:
- Update student portal menu so that all items can be tabbed through. Currently, after the menu is opened, tabbing closes the menu.
- Reuse the same solution from the ticket


andrewsignori-aot (approved):
looks good 👍

PR #3372, opened 2024-06-03
#3283 - Adjusted Federal Restrictions Import Parallelism

andrewsignori-aot:
Set the max parallel to 4 instead of the previous OS cores (usually 32 in Openshift).
The current configuration was making the POD unresponsive, leading to Health check fails, and causing the queue job to be stalled.

image


andrepestana-aot (approved):
👍

Issue #3140, opened 2024-04-18
Build Ministry-Specific Reports - Programs and Offering Status


CarlyCotton:
Follow-Up to #2580

### **User Story**
As a Ministry user, I want to be able to extract dedicated reports to support day to day operations and workload. These reports will help me understand the designation status of institutions (and upcoming expiration), the approval status of programs and offerings, and identify students from a specific institution and/or program.

### **Acceptance Criteria:**
- [x] **Build a report titled "Program and Offering Status ".** 
- [x] Access: Make this report available only through Ministry account (to all Ministry user types)
- [x] Offering Intensity Filter: If PT, all PT offerings and FT&PT programs. If FT ONLY, all FT offerings and FT only programs. If FT&PT, all offerings and FT / FT&PT programs.
- [x] Date Range Filter: Include a union of the two data sets based on date
    - Data Set 1:  _Offerings_ that have a start date within the date range selected regardless of program/offering state; AND
    - Data Set 2: _Programs_ that are "Active" or "Pending" at the start of the range (exclude programs that are deactivated, approval declined, or approved but past expiry date).
        - Programs must be included even if they don't have any offerings attached that would be included as part of Data Set 1. 
- [x] Report-specific Optional Filter: Add an optional filter field for SABC Program Code (If entered, only retrieve records that match. If left blank, show all. This is free text field, with 4-character max validation)
- [x] Report-specific Mandatory Filter: Add a filter drop-down for Institution Name (all institution **operating name** listed)
- [x] File Layouts should be consistent with the table in the "Sample Reports" worksheet of the attached file
- [x] Devs have free-choice to add new columns to reflect data if it aligns with the structure of the database. 
- [x] Translation from database variable name to business-friendly name doesn't need to happen, but nice to have 

### **Business Context**
[Sample Reports Worksheet](https://bcgov.sharepoint.com/:x:/r/teams/03177/Shared%20Documents/DEVS/Reports/SIMS%20Reports%20%26%20Notification%20Triggers.xlsx?d=w08758022bbde485e86f9ed065d92bcfe&csf=1&web=1&e=BgvbWd)
This report has the two data sets to account for offerings that have a start date within the date range selected that may have a program which has become inactive prior to the start date of the offerings.

**Technical**
[...]

PR #3398, opened 2024-06-10
#3140 - Build Ministry-Specific Reports - Programs and Offering Status (Part 1)

lewischen-aot:
As part of this story, the frontend UI/UX and partial backend have been completed in this PR
- Added Form.io changes for the report. Update form for the filter logic specific to the report. Also related Vue changes.
- Added new options API to get the institution name dropdown values.
- Added report filters:
  - Institution operating name dropdown
  - SABC program code - Free text input restricted to 4 characters max. (Optional)

Screenshot of the Ministry Reports with new elements
image

Screenshot of the Institution Name dropdown content
image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
@@ -86,4 +86,17 @@ export class InstitutionControllerService {
       description: institutionType.name,
     }));
   }
+
+  /**
+   * Get the list of all institutions names to be returned in an option
+   * list (key/value pair) schema.
+   * @returns institutions names in an option list (key/value pair) schema.
+   */
+  async getInstitutionNameOptions(): Promise<OptionItemAPIOutDTO[]> {
+    const institutions = await this.institutionService.getAllInstitutions();
+    return institutions.map((institution) => ({
+      id: institution.id,
+      description: institution.legalOperatingName,
+    }));
+  }
 }
--- a/sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
@@ -93,7 +93,7 @@ export class InstitutionControllerService {
    * @returns institutions names in an option list (key/value pair) schema.
    */
   async getInstitutionNameOptions(): Promise<OptionItemAPIOutDTO[]> {
-    const institutions = await this.institutionService.getAllInstitutions();
+    const institutions = await this.institutionService.getAllInstitutionNames();
     return institutions.map((institution) => ({
       id: institution.id,
       description: institution.legalOperatingName,
--- a/sources/packages/backend/apps/api/src/services/institution/institution.service.ts
+++ b/sources/packages/backend/apps/api/src/services/institution/institution.service.ts
@@ -998,10 +998,10 @@ export class InstitutionService extends RecordDataModelService<Institution> {
   }
 
   /**
-   * Get all institutions.
-   * @returns all institutions.
+   * Get all institution legal operating names.
+   * @returns all institution legal operating names.
    */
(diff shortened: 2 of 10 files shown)


sources/packages/backend/apps/api/src/route-controllers/institution/institution.controller.service.ts
```diff
@@ -86,4 +86,17 @@ export class InstitutionControllerService {
       description: institutionType.name,
     }));
   }
+
+  /**
+   * Get the list of all institutions names to be returned in an option
+   * list (key/value pair) schema.
+   * @returns institutions names in an option list (key/value pair) schema.
+   */
+  async getInstitutionNameOptions(): Promise<OptionItemAPIOutDTO[]> {
```
andrewsignori-aot, line 95: In this scenario where the method would be unlikely to be shared between two clients, we can have small logic like this inside the controller itself.
The controller's services are intended to share logic between two clients or reduce the amount of code in the controller file.
This is not a blocker.

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -51,6 +51,8 @@ export default defineComponent({
     const REPORT_TYPE_DROPDOWN_KEY = "reportName";
     const loading = ref(false);
     const PROGRAM_YEAR_DROPDOWN_KEY = "programYear";
+    const INSTITUTION_NAME = "institutionName";
```
andrewsignori-aot, line 54: Since it is a dropdown, it should follow the same naming conversion from `REPORT_TYPE_DROPDOWN_KEY` and `PROGRAM_YEAR_DROPDOWN_KEY` which would be `INSTITUTION_NAME_DROPDOWN_KEY`.

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -78,6 +80,17 @@ export default defineComponent({
             PROGRAM_YEAR_DROPDOWN_KEY,
           );
         }
+        if (event.changed.value === PROGRAM_AND_OFFERING_STATUS) {
+          const institutionName = getFirstComponent(form, INSTITUTION_NAME);
```
andrewsignori-aot, line 84: It should follow the same naming convention from the `programYearSelect`, which would be `institutionNameSelect`.

sources/packages/web/src/components/common/Reports.vue
```diff
@@ -78,6 +80,17 @@ export default defineComponent({
             PROGRAM_YEAR_DROPDOWN_KEY,
           );
         }
+        if (event.changed.value === PROGRAM_AND_OFFERING_STATUS) {
```
andrewsignori-aot, line 83: This condition does not need to be added since we have the check `institutionName._visible`.
Removing the condition would allow the form logic to be changed and this method would not need to be updated, following the same logic for the program year.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -204,6 +204,16 @@ export function useFormioDropdownLoader() {
     );
   };
 
+  const loadInstitutionName = async (form: any, dropdownName: string) => {
```
andrewsignori-aot, line 207: Please use `form: FormIOForm` instead of `any`.
The program year using `form: any` was a gap in the previous PR.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -996,4 +996,12 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
     return !!institutionStudentDataAccess;
   }
+
+  /**
+   * Get all institutions.
+   * @returns all institutions.
+   */
+  async getAllInstitutions(): Promise<Partial<Institution>[]> {
+    return this.repo.find();
+  }
```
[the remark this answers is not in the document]
andrewsignori-aot: Please add the order as requested.

andrewsignori-aot (approved):

Issue #3105, opened 2024-04-10
Camunda 8 SDK Adoption


andrewsignori-aot:
Camunda 8.5 is now released and officially supports a Nodejs SDK (https://www.npmjs.com/package/@camunda8/sdk).
As stated in the npm package page: _This is the official supported-by-Camunda Nodejs SDK for Camunda Platform 8._

**Acceptance criteria**
- [ ] Evaluate the impact of replacing the current current Zeebe libraries by the new SDK.
- [ ] Check the impact of the E2E tests.
- [ ] Ensure that Openshift is updated at least to 8.5.
- [ ] If local tests are blocked due to the Camunda local development test, have it upgraded.

andrewsignori-aot:
Instead of analysis, this ticket is delivering the replacement of the Zeebe libraries by the new SDK.

## Main achievements
- Local Camunda docker updated to allow testing the changes locally.
- Workflow E2E tests execution locally executed and on GitHub Actions.
- Openshift deployment change to allow the Camunda connectivity.

## Zeebe package upgrade
- Removed the now deprecated [zeebe-node](https://www.npmjs.com/package/zeebe-node) package.
- Added the [camunda8/sdk](https://www.npmjs.com/package/@camunda8/sdk)

### Changes required due to differences between the packages
- `ZBClient` was renamed `ZeebeGrpcClient` (it caused changes across multiple files), **_this caused an impact across all deployment files_**.
- Types previously imported from "zeebe-node" are now imported from "@camunda8/sdk/dist/zeebe/types", **_this caused an impact across all deployment files_**.
-  `IOutputVariables` variables were defined where they should be `ICustomHeaders` and they were fixed.
- Env variable `ZEEBE_AUTHORIZATION_SERVER_URL` is now called `CAMUNDA_OAUTH_URL`. **_This caused an impact across all deployment files_**.
- There are some differences in the way numbers are handled and it is explained [here](https://github.com/camunda/camunda-8-js-sdk?tab=readme-ov-file#a-note-on-how-int64-is-handled-in-the-javascript-sdk). No issues were detected for the workflow executions other than a small change for E2E tests.

## E2E Tests Execution

### Improvement to prevent resource issues

E2E tests were facing memory leaks and failing multiple times (noticeable in recent past PRs) due to some resource issues.
[...]

PR #3407, opened 2024-06-12
#3105 - Camunda 8 SDK Adoption

andrewsignori-aot:
The massive number of files that were changed ensured that the concepts below worked simultaneously. Multiple files had minor types renaming or receiving the same minor change over and over again, which would reduce the PR complexity.
1. Local Camunda docker updated to allow testing the changes locally.
2. Workflow E2E tests execution locally executed and on GitHub Actions.
3. Openshift deployment change to allow the Camunda connectivity.

## Zeebe package upgrade
- Removed the now deprecated [zeebe-node](https://www.npmjs.com/package/zeebe-node) package.
- Added the [camunda8/sdk](https://www.npmjs.com/package/@camunda8/sdk)

### Changes required due to differences between the packages
- `ZBClient` was renamed `ZeebeGrpcClient` (it caused changes across multiple files), **_this caused an impact across all deployment files_**.
- Types previously imported from "zeebe-node" are now imported from "@camunda8/sdk/dist/zeebe/types", **_this caused an impact across all deployment files_**.
-  `IOutputVariables` variables were defined where they should be `ICustomHeaders` and they were fixed.
- Env variable `ZEEBE_AUTHORIZATION_SERVER_URL` is now called `CAMUNDA_OAUTH_URL`. **_This caused an impact across all deployment files_**.
- There are some differences in the way numbers are handled and it is explained [here](https://github.com/camunda/camunda-8-js-sdk?tab=readme-ov-file#a-note-on-how-int64-is-handled-in-the-javascript-sdk). No issues were detected for the workflow executions other than a small change for E2E tests.

### Improvement to prevent resource issues

E2E tests were facing memory leaks and failing multiple times (noticeable in recent past PRs) due to some resource issues.
It was identified that the majority of E2E tests were not closing the Zeebe client started. To fix this, an `afterAll` method was introduced along almost every file, as shown below. **_This caused an impact across all deployment files_**.
```ts
afterAll(async () => {
  // Closes the singleton instance created during test executions.
  await ZeebeMockedClient.getMockedZeebeInstance().close();
});
```

### Deployment Change

Once workflow E2Es were tried to be executed in GitHub there were a few times when the deployment of the BPMNs was failing because the server was ready to receive the connections but not ready to accept commands from the Zeebe client. To resolve it a routine was introduced in the [deploy.ts](https://github.com/bcgov/SIMS/blob/cc8fb3518423f44a0be0d93b318f782417cb6cd4/sources/packages/backend/workflow/src/deploy.ts#L48-L78) to check if the Zeebe is ready to accept commands.

### Local Development Docker Changes

- The below variables were added to the make file to allow future version upgrades.
  - `CAMUNDA_CONNECTORS_VERSION`
  - `CAMUNDA_PLATFORM_VERSION`
  - `ELASTIC_VERSION`
- The below variables were added to the make file to allow local development.
  - `CAMUNDA_SECURE_CONNECTION` must be set to allow connection without SSL.
  - `CAMUNDA_OAUTH_DISABLED` must be set to allow the local unauthenticated development connection. 
- `camunda-docker-compose-core.yml` was updated from the most recent version from https://github.com/camunda/camunda-platform/tree/main

**_Important node_**: there was an incompatibility between Zeebe version that would prevent the local docker to be upgraded due to a volume issue. The easier way to resolve this would be to remove the associate volumes, allowing the `make camunda` to recreate them.


sources/camunda-docker-compose-core.yml
```diff
@@ -4,28 +4,39 @@
...
+    image: camunda/zeebe:${CAMUNDA_PLATFORM_VERSION}
     container_name: zeebe
     ports:
       - "26500:26500"
       - "9600:9600"
+      - "8088:8080"
     environment: # https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/environment-variables/
       - ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_CLASSNAME=io.camunda.zeebe.exporter.ElasticsearchExporter
       - ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_URL=http://elasticsearch:9200
+      # default is 1000, see here: https://github.com/camunda/zeebe/blob/main/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporterConfiguration.java#L259
```
andrepestana-aot, line 23: This link is showing 404 - page not found.

sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
```diff
@@ -72,7 +71,9 @@ export class ZeebeTransportStrategy
    */
   private async workerHandlerWrapper(
     jobHandler: MessageHandler,
-    job: Readonly<ZeebeJob<IInputVariables, ICustomHeaders, IOutputVariables>>,
+    job: Readonly<
+      ZeebeJob<{ [x: string]: unknown }, ICustomHeaders, IOutputVariables>
```
andrepestana-aot, line 75: Is there any implication if we just use `ZeebeJob<unknown, ICustomHeaders, IOutputVariables>` ?

andrepestana-aot (approved):
Great research and extensive work. Thanks for taking care of the zeebe being ready for connections and the test resource management. I tested on my local env and everyting looks good. I just added minor comments, so approving. Thanks!

sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
andrewsignori-aot: `unknown` can be used but will be less specific in this case.
For this one, the type suggested was used. Since it is more specific than `unknow` I would prefer to leave as it is.
image

sources/camunda-docker-compose-core.yml
andrewsignori-aot: The file `[camunda-docker-compose-core.yml](https://github.com/bcgov/SIMS/pull/3407/files/5c91c115bd1ad10b936a25424e67805772fb8eb2#diff-95b5b59a1b69248d37d3c3e37fccee1eb5eda083cdc7da7e9e3639b0c90d672d)` is a copy from the below source and was kept as it is.
https://github.com/camunda/camunda-platform/blob/main/docker-compose.yaml

Issue #2831, opened 2024-02-15
Update Gender to Include Non-Binary 


michesmith:
**As a** *(Student)* **I want** *(to show my gender identification in alignment with other public services. 

As a Ministry User want to satisfy external partner student verification requirements by capturing Student gender identification. 

**Acceptance Criteria**
- [x] Update the gender selection options and display in the student profile (PT & FT Applications; Student, Institution and Ministry view): 
    - [ ] Male => Man
    - [ ] Female => Woman
    - [ ] Diverse => Non-Binary
    - [ ] Unknown => Prefer not to answer
- [x] Mapping out to integrations for (eCert/MSFAA/ESDC SIN check) will be updated to the following:
    - Man (Male) => M 
    - Woman (Female) => F 
    - Non-binary (Diverse) => X 
    - Prefer not to answer (Unknown) => Blank or white space 
- [x] Convert existing values to new mapped values
**Note:** All above gender changes on student profile apply equally to BasicBCeID students and BCSC students

**Business Context**
- https://www2.gov.bc.ca/assets/gov/government/services-for-government-and-broader-public-sector/information-technology-services/standards-files/312_gender_and_sex_data_standard_guidelines.pdf
- This ticket is handled seperately from SIN check Gender sharing

**Technical**
- [x] Create a DB migration to update the existing saved values to the new ones.
- [x] Business provided the labels, we decide the values to be stored and we should convert them accordingly.
- [x] Check for references to the gender column from the student table to find other possible places to have conversions adjusted as the business AC is mentioning few and also "etc".
- [x] Fix existing E2E.

CarlyCotton:
@michesmith I've added some acceptance criteria under the user story. Can you confirm if this is in line with what was confirmed re gender updates?
It may be ready to lather as I'm sure there will be more technical considerations.

HRAGANBC:
DEVS IGNORE - BUSINESS CONTEXT 

- two things - just need to be consistent about how we ask, per guideline. 
need to carefully consider downstream impacts as we do share this info in our integrations 
it says what to label them: 
o Female o Intersex o Male o Prefer not to answer

Per Sheena March 20th
We will have
Man=M
Woman=F
Non-Binary=X
Prefer not to answer (maybe, if we can incorporate that into the system)= Blank for ecert

andrewsignori-aot:
@HRAGANBC `IER12` does not have gender so no changes are expected unless we are missing something. Can you adjust the AC, please?
[...]

PR #3417, opened 2024-06-13
#2831 - Add a display formatter for the ministry view of student gender

katrina-aot-git:
Add a display formatter for the ministry view of student gender.

### DIFF
--- a/sources/packages/web/src/composables/useFormatters.ts
+++ b/sources/packages/web/src/composables/useFormatters.ts
@@ -261,6 +261,24 @@ export function useFormatters() {
   };
 
   /**
+   * Converts the saved gender to the capitalized display version.
+   * @param gender value to be converted to be displayed.
+   * @returns formatted gender string.
+   */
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/composables/useFormatters.ts
```diff
@@ -260,6 +260,24 @@ export function useFormatters() {
     return DEFAULT_EMPTY_VALUE;
   };
 
+  /**
+   * Converts the saved gender to the capitalized display version.
+   * @param gender value to be converted to be displayed.
+   * @returns formatted gender string.
+   */
+  const genderDisplayFormat = (gender: string): string | undefined => {
```
andrewsignori-aot, line 268: Since the `switch` has a default, `undefined` will never be returned, right?

katrina-aot-git: ah, correct

andrewsignori-aot (approved):
Thanks for taking care of these extra points. Looks good 👍

Issue #3180, opened 2024-04-23
Virus Scanning - Enable Server Deployment OpenShift and locally


sslaws:
**User Story**
As a Security Analyst for SIMS, I would like to ensure that malware is not propagated VIA SIMS due to the reputational risk and as such require that all files received via SIMS are scanned. The scan happens at regular intervals rather than in real-time which means that it can be handled asynchronously instead of right away when the file is received.

**Acceptance Criteria**

- [x] Add a ClamAV docker into our existing docker-compose. If any challenges are found check an alternative way to make it available for local development.
  - [x] Add ClamAV as a API dependency in docker-compose.
- [x] ClamAV configuration must align with SIMS file upload maximum. 
- [x] Create ClamAV automatic deployment to the Openshift.
  - [x] Follow the instructions in the context for extracting the proper YAML file.

**Context**
- Deployed into a6ef19-dev via Helm, requires limits of 3GB RAM and 1CPU per pod
A [Common Service](https://bcgov.github.io/common-service-showcase/services/cavms.html) exists for the signature files.
- Common Service deployment [source](https://github.com/bcgov/common-hosted-clamav-service)  
- Clone and create deployable YAML`helm template dev . --namespace licenseplate-dev`
- Run locally via docker `docker run -it --rm --platform linux/amd64 -p 3310:3310 --name clamav ghcr.io/bcgov/clamav-unprivileged:ca3d42f3dde3c5aa9bcab636f752119bbe6a67e8`
- [Sample code](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1dVQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--736af41447165f1f912cdfc83721e8d68e18ded4/clamtest.zip)
uses Docker above, unzip and perform npm i then npm start

- For a Virus HIT update eicar.txt and replace text with
`X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*`

ninosamson:
Requested confirmation of approach to defer scanning to scheduled scans.

guru-aot:
Wiki updated for clamav installation in openshift and local development.
https://github.com/bcgov/SIMS/wiki/DevOps-and-Running-the-Application#clamav-setup

PR #3427, opened 2024-06-17
#3180 - Helm clam av GitHub action

guru-aot:
- Configuration for clamAV server is complete
- Local starting of the clamav is done using the `make clamav` command.
-  ClamAV configuration is aligned with SIMS file upload maximum.
- ClamAV is added as a dependency to the api startup in local development.

Note: 
- Added some comments inline in the PR files differences for better understanding.
- The security hotspot is just mentioning about running the dockerfile as a root user and for now we are not handling it.


devops/helm/clamAV/_clamav/configurations/1.0/clamd.conf
```diff
@@ -547,7 +547,7 @@ MaxScanSize 150M
 # Technical design limitations prevent ClamAV from scanning files greater than
 # 2 GB at this time.
 # Default: 100M
-MaxFileSize 30M
+MaxFileSize 4M
```
guru-aot, line 550: In sync with with SIMS file upload maximum

sources/packages/clamav/clamd.conf
```diff
@@ -0,0 +1,813 @@
...
+TCPSocket 3310
+
+# TCP address.
+# By default we bind to INADDR_ANY, probably not wise.
+# Enable the following to provide some degree of protection
+# from the outside world. This option can be specified multiple
+# times if you want to listen on multiple IPs. IPv6 is now supported.
+# Default: no
+# Important! 'localhost' won't work
+#TCPAddr 0.0.0.0
```
guru-aot, line 125: This is different configuration from the devops clamd.conf as for localhost development, this tcpaddr should be enabled for local file scanning.

sources/packages/clamav/clamd.conf
```diff
@@ -0,0 +1,813 @@
...
+# daemon (main thread).
+# This file will be owned by root, as long as clamd was started by root.
+# It is recommended that the directory where this file is stored is
+# also owned by root to keep other users from tampering with it.
+# Default: disabled
+# PidFile /var/run/clamd.pid
+
+# Optional path to the global temporary directory.
+# Default: system specific (usually /tmp or /var/tmp).
+#TemporaryDirectory /var/lib/clamav/tmp
```
guru-aot, line 77: This is different configuration from the devops clamd.conf as for local development, this temporary directory is not found and causes error while scanning.

sources/packages/clamAV/clamd.conf
```diff
@@ -0,0 +1,813 @@
...
+# 2 GB at this time.
+# Default: 100M
+MaxFileSize 4M
+
+# Nested archives are scanned recursively, e.g. if a Zip archive contains a RAR
+# file, all files within it will also be scanned. This options specifies how
+# deeply the process should be continued.
+# Note: setting this limit too high may result in severe damage to the system.
+# Default: 17
+MaxRecursion 10
```
andrepestana-aot, line 557: What happens if these values are exceeded? A specific error is thrown? Are we going to capture those type of errors?

devops/helm/clamAV/_clamav/configurations/1.0/clamd.conf
andrepestana-aot, line 550: I know this file was already here but should we keep the parent directory name starting with underscore (_clamav)?

guru-aot: the _calmav is part of the copy we did from the [source](https://github.com/bcgov/common-hosted-clamav-service)  . So do not want to change it.

andrewsignori-aot, line 550: It is well known the file will be increased to 15 MB. I do not see a reason to not leave this ready since there is no harm if we even never do it.

devops/helm/clamAV/_clamav/values.yaml
```diff
@@ -72,9 +72,8 @@ hpa:
   # requests: 1k
 
 podDisruptionBudget:
-  enabled: false
-  # minAvailable: 1
-  # maxUnavailable: 1
+  enabled: true
+  minAvailable: 2
```
andrewsignori-aot, line 76: Since the replica count is also set to 2, should it not be 1 to allow one POD to be moved?

sources/Makefile
```diff
@@ -205,6 +206,12 @@ queue-consumers:
 	@docker-compose -f docker-compose.yml build queue-consumers
 	@docker-compose -f docker-compose.yml up -d queue-consumers
 
+# Runs clamav
+clamav:
+	@echo "+\n++ Make: Clamav only"
```
andrewsignori-aot, line 211: `ClamAV`

sources/packages/clamAV/Dockerfile
andrewsignori-aot, line 1: Following the same from `load-test`, can it be named `clam-av` or is it considered one word only?

sources/packages/clamAV/Dockerfile.dev
andrewsignori-aot, line 1: Is it possible to follow the below recommendations or there is a justification that we can use to not use it?
For instance, it seems that there is a `clamav` user to be used with `init-unprivileged`.

https://docs.clamav.net/manual/Installing/Docker.html#running-clamd-using-non-root-user-using---user-and---entrypoint

sources/packages/clamAV/Dockerfile.dev
```diff
@@ -0,0 +1,6 @@
+FROM clamav/clamav:1.2
+RUN freshclam
```
andrewsignori-aot, line 2: Is it going to run `freshclam` in the background doing regular database upgrades (daemon mode)?

sources/docker-compose.yml
```diff
@@ -68,6 +68,7 @@ services:
     depends_on:
       - postgres
       - redis
+      - clamav
```
andrewsignori-aot, line 71: I believe we should review the AC.
ClamAV will not be a dependency for the API to be executed.
Maybe it should not be a dependency at all since even only one `queue-consumers` queue will depend on it.
We may add it as a dependency for `queue-consumers`, but not mandatory IMO.

sources/packages/clamAV/clamd.conf
```diff
@@ -0,0 +1,813 @@
...
+# Execute a command when virus is found. In the command string %v will
+# be replaced with the virus name and %f will be replaced with the file name.
+# Additionally, two environment variables will be defined: $CLAM_VIRUSEVENT_FILENAME
+# and $CLAM_VIRUSEVENT_VIRUSNAME.
+# Default: no
+#VirusEvent /usr/local/bin/send_sms 123456789 "VIRUS ALERT: %v in %f"
+
+# Run as another user (clamd must be started by root for this option to work)
+# Default: don't drop privileges
+#User clamav
```
andrewsignori-aot, line 226: Is this an option to run it as a non-privileged user?

sources/packages/clamAV/clamd.conf
```diff
@@ -0,0 +1,813 @@
...
+# If you turn off this option, the original files will still be scanned, but
+# without decoding and additional processing.
+# Default: yes
+ScanSWF yes
+
+# This option enables scanning xml-based document files supported by libclamav.
+# If you turn off this option, the original files will still be scanned, but
+# without additional processing.
+# Default: yes
+#ScanXMLDOCS yes
```
andrewsignori-aot, line 421: This one seems to be a good one to be enabled.

andrewsignori-aot (commented):
Great work and research, please take a look at the comments.

sources/packages/clamAV/clamd.conf
guru-aot: The alerts are enabled via configurations, even if these setting are hit until the alerts are enables, clamav does not generate them.

sources/packages/clamAV/Dockerfile.dev
guru-aot: Yes this is just to start the freshclam update when the container starts

sources/packages/clamAV/clamd.conf
guru-aot: no even setting this configuration still the user is root.

sources/packages/clamAV/clamd.conf
guru-aot: Yeah the default value is enabled, so i did not care to configure.

andrewsignori-aot (approved):
Thanks for making the changes and for the research, looks good 👍

Issue #3031, opened 2024-03-21
Content - Overaward Balance


CarlyCotton:
**User Story**
As a part-time student, I don't want to see information that is not relevant to me in part-time as it could create confusion. While the environment is not showing any FT Only content, remove these links on the student profile drop-down. 

**Acceptance Criteria**
- [x] When the environment is set for PT only / no FT, remove the Overawards Balance link from the **student** view:
<img width="460" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOUpCQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--26d31db600151b5ec8fee00a87f183efabb4c6be/image.png" alt="image.png" />

**Technical**
- [x] Use the `IS_FULL_TIME_ALLOWED` in the web app to hide the `Overawards Balance link` only for the student.

PR #3461, opened 2024-06-25
#3031 - Overawards menu item

katrina-aot-git:
Make the overawards section visibility dependent on fulltime being allowed.

### DIFF
--- a/sources/packages/web/src/views/student/AppStudent.vue
+++ b/sources/packages/web/src/views/student/AppStudent.vue
@@ -100,12 +100,13 @@
 
 <script lang="ts">
 import { useRouter } from "vue-router";
-import { computed, ref, defineComponent } from "vue";
+import { computed, ref, defineComponent, onMounted } from "vue";
 import { StudentRoutesConst } from "@/constants/routes/RouteConstants";
 import { ClientIdType, MenuItemModel } from "@/types";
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/views/student/AppStudent.vue
```diff
@@ -151,14 +152,6 @@ export default defineComponent({
...
             },
           },
-          {
-            title: "Overawards Balance",
-            props: {
-              to: {
-                name: StudentRoutesConst.STUDENT_OVERAWARDS,
-              },
-            },
-          },
```
andrepestana-aot, line 161: Instead of removing  the item and adding in case Full-Time is allowed I prefer leaving the list as it is and removing it in the `onMounted()` because that will be easier to maintain and to remove the condition one day.

```ts
    onMounted(async () => {
      const { isFulltimeAllowed } = await AppConfigService.shared.config();
      if (!isFulltimeAllowed) {
        menuItems.value.forEach((item, index) => {
          if (item.title === "Overawards Balance") {
            menuItems.value.splice(index, 1);
          }
        });
      }
    });
```

katrina-aot-git: updated

andrepestana-aot (approved):
Thanks for doing the change. Great job!

Issue #3268, opened 2024-05-10
Database Upgrade - Analysis and implementation


sslaws:
As SIMS technical team, we need to ensure that our tools and platform are maintained at current versions.  Postgres 12 is EOL as of November 2024 and must be upgraded prior to that.  As part of this migration, ideally the team would use technologies that treat the DB more as a service.

**Acceptance Criteria**
- [x] Deploy new DB and PgBouncer to all environments
    - [x] Dev
    - [ ] Test - following regular release activity/schedule. 
    - [ ] Staging - following regular release activity/schedule. 
    - [ ] Production - following regular release activity/schedule. 
- [x] Update application connection to use PgBouncer
- [x] Document, verify, and execute process to migrate Production data
- [ ] Document, test and execute the removal of the old database
- [x] Use Helm for Deployment

**Technical Context**

Decisions:
Use Crunchy Operator and Helm to create the DB

Expectation on the analysis:
1. Helm chart analysis on the crunchy db setup available in BC gov -  https://github.com/bcgov/crunchy-postgres - use dev environment - a6ef19-dev
2. Understand the configurations and values.yaml in the chart and create folders in our local repository for environment.
3. Github action needs to be created, for install, upgrade and backup - if possible try to merge it into a single action.
4. Update the wiki in our repository.

Review Items:
- Ideally use PG 16 for longevity but currently installed operator doesn't support this.
- Current DB name is SIMSDB but Crunchy doesn't like that.
- Do we need to expose the postgres super user account?

Sample yaml using Crunchy Operator to deploy Patroni HA Postgres 15 with pgbouncer and pgbackrest.

```
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
metadata:
  name: simsdb
spec:
  postgresVersion: 15
  instances:
    - name: pgha1
      replicas: 3
      resources:
        limits:
          cpu: 1
          memory: 512Mi
        requests:
          cpu: 250m
          memory: 512Mi
  backups:
    pgbackrest:
      global:
        repo1-retention-archive: "10"
        repo1-retention-full: "7"
        repo1-retention-incremental: "24"
      repos:
      - name: repo1
        schedules:
          full: 0 2 * * *
          incremental: 0 0-1,3-23 * * *
        volume:
          volumeClaimSpec:
            accessModes:
            - ReadWriteOnce
            resources:
              requests:
                storage: 20Gi
  proxy:
    pgBouncer:
      replicas: 3
  users:
[...]

PR #3476, opened 2024-06-28
#3268 - Crunchy Postgres Database setup - Part 1

guru-aot:
Changes in this PR:
- Installation of Cruchy DB using helm.
- Created related make files for install, upgrade and uninstall.
- Created github actions for install/upgrade with options to enable restore helm update.
- Most of the files are copied from https://github.com/bcgov/crunchy-postgres, but restricted to changes and configurations for our project SIMS.
- Connecting our application to the new crunchy db will be created in the next PR.
- Wiki updates will happen once things are finalized and demoed.

**Detailed Architecture**

image

**Services available to connect the database:**
- PgBouncer helps your database handle a larger number of client connections efficiently. It reduces the overhead of creating and destroying connections, leading to faster database access for your applications. Only the database user can be used not the postgres user.
- Replicas-Serive - Crunchy Postgres uses replicas for HA, not a separate "crunchy-replica" service. PGO manages replica deployment and failover within your Kubernetes cluster. Replicas can improve performance by handling read traffic.
- Primary-DB-Service -  The service provides a stable endpoint (hostname and port) for your applications to connect to the primary database, regardless of the underlying pod IP addresses (which can change dynamically in Kubernetes) and  If we scale your primary database by adding more pods (for high availability), the service automatically distributes traffic across all healthy pods.

**Database backups available in Crunchy for our setup we have given the backup retention to be 30days.**
 - pgBackRest in Crunchy Postgres for Kubernetes often included as part of its solution. The Postgres Operator (PGO) can be configured to automate pgBackRest backups, ensuring a robust disaster recovery strategy for your PostgreSQL deployments.
 - image
 - https://access.crunchydata.com/documentation/postgres-operator/latest/tutorials/backups-disaster-recovery/backup-management

**Links on why to use Crunchy for high-availability**
https://access.crunchydata.com/documentation/postgres-operator/latest/tutorials/day-two/high-availability

**Setup to install Crunchy Database**
- The current draft PR contains the setup of the latest 15 version of postgres database installation with crunchy database.
- We used helm installation and followed the documentation of both BCGOV(https://github.com/bcgov/crunchy-postgres) and the official crunchy db documentations(https://access.crunchydata.com/documentation/postgres-operator/latest/installation/helm) to configure the backup, the database name (SIMSDB - simsdb)similar to the one we have already. The database user (app_database_user - app-database-user)
- Once the PR is merged go to --> devops/helm/crunchy-postgres.
- RUN command ```make install NAMESPACE=${NAMESPACE}```, this should install crunchy database.
- To customize the postgresVersion, PVC size, backup retention period, backup repo, CPU, memory limits, appuser, databasename(fullnameOverride), etc can be customized using the ```values.yaml``` in ```devops/helm/crunchy-postgres```.
- Github action can be created for the makefile as part of the subsequent story implementation.

**Backup and Restore**
- Backup for the crunchy database is already included as part of the crunchy implementation. Once this PR is merged and Crunchy is installed, 3 cronjobs(full, differential, incremental) does the backups in the xxxxx-repo-host stateful sets.
- Restore of the crunchy database is analyzed and successful.

**Migration of the existing data into crunchy database (Patroni data to crunchy)**
 - As part of testing dev data is taken as a backup by port forwarding the dev db patroni-master service using the pg_dump tools in dbeaver and was tried to restore the data  into the crunchy database by port-forwarding the connection using the pgbouncer service.
 - Problems in Database name and app user, as postgres helm chart followed by crunchy is using the apiVersion postgres-operator.crunchydata.com/v1beta1, this does not support usernames with ```_``` and database name in all caps so the user name currently we were using in production is ```app_database_user``` which cannot be created has to changed to ```app-database-user``` and  ```SIMSDB``` had to be changed to ```simsdb``.
 - So if we are migrating from patroni to crunchy the data sql/tar file which is backed up had to be manually opened and the above changes has to be done and then the restore process has to continue.


devops/Makefile
```diff
@@ -628,4 +628,4 @@ db-backup-deploy-patroni-delete:
 # Remove redis and resources including secrets from openshift namespace.
 delete-redis:
 	@echo "+\n++ Removing redis and resources on $(NAMESPACE)\n+"
-	@oc delete -n $(NAMESPACE) all,rc,svc,dc,route,pvc,secret,configmap,sa -l app=redis
+	@oc delete -n $(NAMESPACE) all,rc,svc,dc,route,pvc,secret,configmap,sa -l app=redis
```
andrewsignori-aot, line 631: Minor, no need to add this to the PR if no changes are present.

guru-aot: yeah i am trying to remove it and another file devops/openshift/db-migrations-job.yml. But its keep on not saving

devops/openshift/db-migrations-job.yml
andrewsignori-aot, line 1: If no real changes are present there is no need to add it to the PR.

devops/helm/crunchy-postgres/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      cpu: 250m
+      memory: 256Mi
+
+# Postgres Cluster resource values:
+pgmonitor:
+  enabled: false
```
andrewsignori-aot, line 101: To double-check, it is disabled because it would require a license to use it, right?

devops/helm/crunchy-postgres/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  repoName: repo1
+  options:
+    target: "${RESTORE_TARGET}"
+
+instances:
+  name: ha # high availability
+  replicas: 3
+  dataVolumeClaimSpec:
+    storage: 10Gi
+    storageClassName: netapp-block-standard
```
andrewsignori-aot, line 27: Any reason to change from `netapp-file-standard` to `netapp-block-standard`?
Would the first one not be the recommended one?

devops/helm/crunchy-postgres/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+      memory: 256Mi
+  s3:
+    enabled: false
+
+patroni:
+  postgresql:
+    pg_hba: "host all all all md5"
+    parameters:
+      shared_buffers: 128MB # default is 128MB; a good tuned default for shared_buffers is 25% of the memory allocated to the pod
+      wal_buffers: "64kB" # this can be set to -1 to automatically set as 1/32 of shared_buffers or 64kB, whichever is larger
```
andrewsignori-aot, line 83: As per the comment, assuming the 128MB as `shared_buffers` it would be better to leave this one as "-1" since 128MB/32=4MB?

devops/helm/crunchy-postgres/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  enabled: "${RESTORE_ENABLED}"
+  repoName: repo1
+  options:
+    target: "${RESTORE_TARGET}"
+
+instances:
+  name: ha # high availability
+  replicas: 3
+  dataVolumeClaimSpec:
+    storage: 10Gi
```
andrewsignori-aot, line 26: The current patroni storage on PROD is defined as 21GiB. Is this configuration targeting non-prod envs and this will be adjusted on PROD?

devops/helm/crunchy-postgres/values.yaml
guru-aot: yes

devops/helm/crunchy-postgres/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  # If retention-full-type set to 'count' then the oldest backups will expire when the number of backups reach the number defined in retention
+  # If retention-full-type set to 'time' then the number defined in retention will take that many days worth of full backups before expiration
+  retentionFullType: count
+  repos:
+    schedules:
+      full: 0 8 * * *
+      incremental: 0 0,4,12,16,20 * * *
+    volume:
+      accessModes: "ReadWriteOnce"
+      storage: 10Gi
```
andrewsignori-aot, line 59: May need to be adjusted depending on the answer to the below comment unless the backups are compressed and the 30 days rotation would take care of it.
https://github.com/bcgov/SIMS/pull/3476/files#r1706158222

devops/helm/crunchy-postgres/values.yaml
guru-aot: yes for non prod environments

devops/helm/crunchy-postgres/values.yaml
andrewsignori-aot, line 1: Do we have an idea if there are enough resources to have crunchy deployed alongside the current patroni during the transition phase?

.github/workflows/crunchy-db.yml
andrewsignori-aot, line 1: Is this Github action targeting two different operations (install and back restore)?
If yes, is there anything preventing us from having two different Github actions?

devops/helm/crunchy-postgres/values.yaml
guru-aot: image

This is the current crunchy only quota configuration deployed in the new dev namespace, i believe we can tweak a few parameters and can deploy it side of dev but will be very tough when there is a restart or any new replicas created for other applications.

devops/helm/crunchy-postgres/values.yaml
```diff
@@ -0,0 +1,101 @@
+fullnameOverride: simsdb
+appuser: app-database-user
```
andrewsignori-aot, line 2: About having this user created, is it a final and required decision needed for the upcoming DB restore plan?
The secondary possible plan of having the user created manually in the new DB is no longer being considered?

andrewsignori-aot (commented):
Great work and research, looking forward for the demo. Please take a look at the comments.

devops/helm/crunchy-postgres/values.yaml
guru-aot: I did not do any changes from the default values.yml from the bcgov but after your comment, went on researching why this is used. https://docs.redhat.com/en/documentation/openshift_container_platform/4.5/html/scalability_and_performance/optimizing-storage#available-persistent-storage-options_persistent-storage. As per the link given i believe for applications that need full control of storage and operate at a low level on files bypassing the file system. It sounds like our application. If there is any recommendations i would be happy to hear and change it.

devops/helm/crunchy-postgres/values.yaml
guru-aot: these are configurations from bcgov, i believe if needed lets talk and have it changed as a team.

devops/helm/crunchy-postgres/values.yaml
guru-aot: no this creates the user with required permissions and mimic the one we have in out existing database.

devops/helm/crunchy-postgres/values.yaml
andrewsignori-aot: I have a feeling that the current Patroni may not have the best  config 😉

.github/workflows/crunchy-db.yml
andrewsignori-aot: Got it after our chat.

andrewsignori-aot (approved):
Great work. Let's move on with the PR and have the Github actions tested. Looks good 👍

Issue #3383, opened 2024-06-05
Queue-consumers Settings Changes


andrewsignori-aot:
During the execution of memory adjustments for POD (#3283) few issues were noticed that would affect production operations while execution and `federal-restrictions-integration` and other schedulers.

**Problem 1**

It was noticed that the `federal-restrictions-integration` scheduler was not successfully executed because the job was detected as "stalled" due to its long-running process (6 to 7 min) where the CPU-intensive use was not allowing the job to properly report its execution status.

**Problem 2**

During the investigation, it was also noticed that when the job was detected as stalled (after 30 seconds) a second job started an attempt to execute the task concurrently, which caused a problem because the current jobs were not prepared for idempotency.

**Problem 3**

Besides that, the `readinessProbe` and `livenessProbe` for the `queue-consumers` POD were based on DB/Redis checks with 1000ms timeout, which under stress were making the POD to be restarted.

**Acceptance criteria**
- [x] Change the queues settings values for `maxStalledCount`, `lockDuration`, and `lockRenewTime`.
  - Suggested values:
    - `maxStalledCount`: 0. Current default value defined by the framework is 1. Setting it as 0(zero) would make the job not be retried and it would stay failed, avoiding the idempotency requirement.
    - `lockDuration`: 60000. Current default value defined by the framework is 30000. Setting it as 60 would give more time before the job is marked as stalled, which is not expected for the short-living processes. In combination with the `lockRenewTime` defined as 5 seconds, it would give more opportunities for the job report that is still running.
    - `lockRenewTime`: 5000. Current default value defined by the framework is 15000 (half of `lockDuration`).
  - [x] Settings to be loaded from the database in a similar way to how they are loaded now for `defaultJobOptions` (see image in the additional context).
  - [x] Initially, all the queues should be configured with the proposed settings.
- [x] Update `readinessProbe` and `livenessProbe` settings to allow the POD restart if not needed.
  - `readinessProbe` suggested to be kept at 1500ms for timeout.
  - `livenessProbe` suggested to be kept at 3000ms for timeout.
  - See the suggested configuration in the additional context.
- [x] Use the federal restriction file to test if the job can be successfully executed. Do the final test in OpenShift.

**Additional Context**

### Settings Configuration

Please see below where the settings can be configured (number 2 below).
The image also shows the moment when some configurations are already loaded from the database (number 1 below).

[...]

PR #3478, opened 2024-07-02
#3383 - Queue-consumers Settings Changes - Health Check Updated

andrewsignori-aot:
## Main changes

- Changed the health check for the `readinessProbe` and `livenessProbe` to allow configurable timeouts.
- Adjusted the timeout for `readinessProbe` and `livenessProbe`.

_Note_: test details added to the ticket comments. 

## Permission change needed

To allow federal restrictions to work, the DB user should be granted the `truncate` permission using the below.
```sql
GRANT TRUNCATE ON TABLE sims.federal_restrictions TO "<env_db_user>";
```


devops/openshift/queue-consumers-deploy.yml
```diff
@@ -212,21 +212,21 @@ objects:
                 failureThreshold: 3
                 successThreshold: 1
                 httpGet:
-                  path: /health/
+                  path: /health/timeout/1500
                   port: "${{PORT}}"
                 initialDelaySeconds: 10
                 periodSeconds: 10
-                timeoutSeconds: 1
+                timeoutSeconds: 3
```
andrepestana-aot, line 219: Just for my knowledge: why this `timeoutSeconds` has value 3 and the other one is 5?

andrewsignori-aot: It was increased because the HTTP timeout was also increased.

_`timeoutSeconds`: The number of seconds of inactivity after which the probe times out and the container is assumed to have failed. The default is 1. This value must be lower than periodSeconds._
Source: https://docs.openshift.com/container-platform/4.9/applications/application-health.html

andrepestana-aot (approved):
looks good. Thanks for the explanation.

Issue #3255, opened 2024-05-09
CAS Integration 1 - Add CAS Supplier Info to Student Profile (Get Supplier Number)


HRAGANBC:
**User Story:** 
SIMS must record and maintain a student's 'supplier' information as part of their profile, so that a student's SIMS profile can match with their corresponding account in the Corporate Accounting System (CAS), and SIMS can relay critical information about that student to CAS. 

**Acceptance Criteria:**

**Initiate 'Get Supplier' Process** 
- [x] After SIN is successfully validated, begin 'Get Supplier' Process (See **Doc 1**,  Section: 11.1.6, page 13 for context)
- [x] Call CAS using the `Last Name` and `SIN`
- [x] Preserve a history of the following variables received during the response:
    - ```SupplierNumber```
    - ```SupplierName```
    - `status`
    - ```Supplier protected```
    - `lastupdated` 
 And from within ```Supplier Address``` 
    - ```suppliersitecode```
    - `Addressline1`
    - `City`
    - `Province`
    - `country`
    - `postalcode`
    - `status`
    - ```siteprotected```
    - ```lastupdated```
- [ ] If the response returns none of the above values, then see #3260 and #3258 for error handling
- [ ] API should be called only during business days (per #3327)

**Technical**
- [x] Follow a similar process from SIN validation.
  - [x] Create a new table to store the history of CAS supplier requests and allow Ministry manual entry, suggested table name `sims.cas_suppliers`. In the future, we will have a new table for invoices that we may call sims.cas_invoices.
    - [x] New table columns following in general the datatypes received from CAS.
      - `id` SERIAL PRIMARY KEY: Auto-generated sequential primary key column.
      - `supplier_number` VARCHAR(100) NULL: Supplier number received from CAS. `null` when no data was ever retrieved from CAS.
      - `supplier_name` VARCHAR(100) NULL: Supplier name received from CAS. `null` when no data was ever retrieved from CAS.
      - `status` VARCHAR(50) NULL: Supplier status received from CAS. `null` when no data was ever retrieved from CAS.
      - `supplier_protected` BOOLEAN NULL: Protected flag received from CAS which means the student profile was created by SFAS and therefore no system other than SFAS can change it. `null` when no data was ever retrieved from CAS.
      - `last_updated` TIMESTAMP WITH TIME ZONE NULL. Date and time of the last update. `null` when no data was ever retrieved from CAS.
[...]

PR #3490, opened 2024-07-05
#3255 - Add CAS Supplier Info to Student Profile (Get Supplier Number) - Part 2

andrepestana-aot:
In this PR:

- Changed `internalCreateStudent` method to add a CAS supplier record with `PendingSupplierVerification` status;
- Added CASSupplierIntegrationScheduler;
- Created a CASService to manage CAS API requests;
- Added CAS variables to the `.env`
- Created mock functions to mock CASService methods and added `casServiceMock` to `createTestingAppModule`;
- Added 2 tests for the scheduler.


devops/Makefile
```diff
@@ -456,6 +459,9 @@ deploy-queue-consumers:
 		-p ATBC_ENDPOINT=$(ATBC_ENDPOINT) \
 		-p APPLICATION_ARCHIVE_DAYS=$(APPLICATION_ARCHIVE_DAYS) \
 		-p IS_FULLTIME_ALLOWED=$(IS_FULLTIME_ALLOWED) \
+    -p CAS_BASE_URL=$(CAS_BASE_URL) \
+    -p CAS_CLIENT_ID=$(CAS_CLIENT_ID) \
+    -p CAS_CLIENT_SECRET=$(CAS_CLIENT_SECRET) \
```
andrewsignori-aot, line 464: I believe the secrets are not properly mapped. Please take a look at a similar secret mapped to queue-consumers and follow the same approach.
https://github.com/bcgov/SIMS/blob/f3d5fa0250717a3693aec48f7883439e23b68bd1/devops/openshift/queue-consumers-deploy.yml#L137-L141

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -261,8 +263,24 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      casSupplier.creator = auditUser;
+      casSupplier.student = savedStudent;
+      const savedCASSupplier = await entityManager
+        .getRepository(CASSupplier)
+        .save(casSupplier);
+
+      savedStudent.casSupplier = savedCASSupplier;
+      const updatedStudent = await entityManager
+        .getRepository(Student)
+        .save(savedStudent);
```
andrewsignori-aot, line 280: This code should be executed only for new students. Following the same approach from SIN, it can be moved inside the `if` at line 220 (`if (!existingStudent)`), and the update can happen at the same time for SIN and Supplier. Does it make sense?

sources/packages/backend/libs/sims-db/src/entities/cas-supplier.model.ts
```diff
@@ -14,6 +21,20 @@ export class CASSupplier extends RecordDataModel {
   @PrimaryGeneratedColumn()
   id: number;
 
+  /**
+   * Reference to student id in students table.
+   */
+  @OneToOne(() => Student, {
+    eager: false,
+    cascade: ["insert", "update"],
```
andrewsignori-aot, line 29: The insert should not be allowed in this relationship.
As information, the SIN having the cascade as `true` is a mistake.

sources/packages/backend/libs/sims-db/src/entities/cas-supplier.model.ts
```diff
@@ -89,7 +110,7 @@ export class CASSupplier extends RecordDataModel {
   @Column({
     name: "supplier_status_updated_on",
     type: "timestamptz",
-    nullable: true,
+    nullable: false,
```
andrewsignori-aot, line 113: This as `nullable: false` should have the property adjusted to not allow nullable.

sources/packages/backend/libs/sims-db/src/entities/cas-supplier.model.ts
```diff
@@ -1,7 +1,14 @@
-import { TableNames } from "@sims/sims-db/constant";
+import { ColumnNames, TableNames } from "@sims/sims-db/constant";
 import { RecordDataModel } from "@sims/sims-db/entities/record.model";
+import { Student } from "@sims/sims-db/entities/student.model";
```
andrewsignori-aot, line 3: The index can be used instead as `import { Student } from "@sims/sims-db/entities";`.

sources/packages/backend/apps/queue-consumers/test/helpers/testing-modules/testing-modules-helper.ts
```diff
@@ -44,11 +47,14 @@ export async function createTestingAppModule(): Promise<CreateTestingModuleResul
     },
   );
   const sshClientMock = createMock<Client>();
+  const casServiceMock = createCASServiceMock();
```
andrewsignori-aot, line 50: Was there a benefit to creating a method instead of using the `createMock<CASService>`?

sources/packages/backend/libs/utilities/src/config/config.models.ts
```diff
@@ -110,3 +110,9 @@ export interface RedisConfiguration {
   redisPassword: string;
   redisStandaloneMode: boolean;
 }
+
+export interface CASIntegrationConfig {
+  baseUrl: string;
+  clientId: string;
+  clientSecret: string;
```
andrewsignori-aot, line 117: Can you please use the existing interface `ClientCredential`?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  private readonly casIntegrationConfig: CASIntegrationConfig;
+  constructor(
+    config: ConfigService,
+    private readonly httpService: HttpService,
+    @InjectRepository(CASSupplier)
+    private readonly casSupplierRepo: Repository<CASSupplier>,
+  ) {
+    this.casIntegrationConfig = config.casIntegration;
+  }
+  /**
```
andrewsignori-aot, line 24: Please add a blank line before the method.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
```diff
@@ -0,0 +1,72 @@
...
+   * Request to login on CAS API and return CAS auth details with the token used for authentication in all other requests.
+   * @returns CAS auth details.
+   */
+  public async casLogon(): Promise<CASAuthDetails> {
```
andrewsignori-aot, line 28: Please remove the `public` keyword.

andrewsignori-aot, line 28: Not a blocker, suggestion, `cas` as prefix would not be needed is a service already named as `CAS`.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
```diff
@@ -0,0 +1,72 @@
...
+import { Repository } from "typeorm";
+
+@Injectable()
+export class CASService {
+  private readonly casIntegrationConfig: CASIntegrationConfig;
+  constructor(
+    config: ConfigService,
+    private readonly httpService: HttpService,
+    @InjectRepository(CASSupplier)
+    private readonly casSupplierRepo: Repository<CASSupplier>,
```
andrewsignori-aot, line 20: `casSupplierRepo` is not used, please remove it.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
andrewsignori-aot, line 1: The technical AC mentioned having the service created under `integrations`. Am I missing something or something changed?

_Create a new CAS API Integration Service under backend/libs/integrations/cas._

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
```diff
@@ -0,0 +1,72 @@
...
+      Authorization: `Basic ${auth}`,
+    };
+    const body = new URLSearchParams();
+    body.append("grant_type", "client_credentials");
+    const config: AxiosRequestConfig = {
+      headers,
+    };
+    const response = await this.httpService.axiosRef.post(
+      url,
+      body.toString(),
```
andrewsignori-aot, line 44: For Keycloak integration `stringify` was used. Can we keep the same?
https://github.com/bcgov/SIMS/blob/4dde03d07b37a188ce39381f986f5a1452ae90e8/sources/packages/backend/libs/auth/src/services/keycloak/keycloak.service.ts#L118

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+   * Gets a list of CAS suppliers to be updated from CAS supplier table.
+   * @returns a list of CAS suppliers to be updated.
+   */
+  async getStudentsToUpdateSupplierInformation(): Promise<CASSupplier[]> {
+    return this.casSupplierRepo.find({
+      select: {
+        id: true,
+        student: {
+          id: true,
+          sinValidation: { sin: true, isValidSIN: true },
```
andrewsignori-aot, line 172: Is `isValidSIN` used anywhere besides the `where`. If not it does not need to be part of the select.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+          user: { lastName: true },
+        },
+      },
+      relations: {
+        student: { sinValidation: true, user: true },
+      },
+      where: {
+        isValid: Not(true),
```
andrewsignori-aot, line 180: Since `isValid` is not nullable this can be simplified as `isValid: false`, right?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+    return this.casSupplierRepo.update(
+      {
+        id: casSupplier.id,
+      },
+      {
+        supplierNumber: casSupplierResponseItem.suppliernumber,
+        supplierName: casSupplierResponseItem.suppliername,
+        status: casSupplierResponseItem.status,
+        supplierProtected: casSupplierResponseItem.supplierprotected === "Y",
+        lastUpdated: new Date(casSupplierResponseItem.lastupdated),
```
andrewsignori-aot, line 151: I would assume this date will be PST/PDT and it will need to be converted before being saved to DB otherwise it will be saved as UTC. Does it make sense?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/models/cas-supplier-response.dto.ts
andrewsignori-aot, line 1: Even if not wrong, I believe that we have been calling DTO the data transfer objects from our API and anything service/integration related would be a model, for instance, the `bceid.models.ts`.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/models/cas-supplier-response.dto.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+export class CASSupplierResponseItem {
+  suppliernumber: string;
+  suppliername: string;
+  subcategory: string;
+  sin: string;
+  providerid?: string;
+  businessnumber?: null;
+  status: string;
+  supplierprotected?: string | null;
```
andrewsignori-aot, line 17: Since it is part of the documentation it can be typed as "N" | null.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/models/cas-supplier-response.dto.ts
```diff
@@ -0,0 +1,52 @@
+export class CASSupplierResponse {
+  items: CASSupplierResponseItem[];
+  hasMore: boolean;
+  limit: number;
+  offset: number;
+  count: number;
+  links: CASSupplierResponseItemLink[];
```
andrewsignori-aot, line 7: In general, I would suggest to map only what the application will consume. For instance `links` will not be needed as far as I can see.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/models/cas-supplier-response.dto.ts
```diff
@@ -0,0 +1,52 @@
...
+  standardindustryclassification?: string;
+  lastupdated: string;
+  supplieraddress: CASSupplierResponseItemAddress[];
+}
+
+export class CASSupplierResponseItemAddress {
+  suppliersitecode: string;
+  addressline1: string;
+  addressline2?: string;
+  addressline3?: string;
```
andrewsignori-aot, line 27: Maybe the business missed the `addressline2` and `addressline3` in the requirement. Have we ever asked then?
If not we should ask, if not in this PR, we can be added to an upcoming ticket.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/models/cas-supplier-response.dto.ts
```diff
@@ -0,0 +1,52 @@
...
+  providerid?: string;
+  status: string;
+  siteprotected?: string;
+  lastupdated: string;
+}
+
+export class CASAuthDetails {
+  access_token: string;
+  token_type: string;
+  expires_in: number;
```
andrewsignori-aot, line 46: If my understanding is correct, the default token expiration will be 60 minutes of idle time which means we really will never need to cache it. Does it sound right?

_An idle token will expire after a specific number of seconds, typically 3600.  The expiry time may be different for each environment.  The expiry times are shared with all Web Services on the server and will not be changed without consideration of the impact to all web services._

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/cas-supplier-integration.scheduler.ts
```diff
@@ -0,0 +1,80 @@
...
+    const processSummary = new ProcessSummary();
+
+    try {
+      this.logger.log(
+        `Processing CAS supplier integration job ${job.id} of type ${job.name}.`,
+      );
+      processSummary.info("Executing CAS supplier integration...");
+      const pendingCASSuppliers =
+        await this.casSupplierIntegrationService.getStudentsToUpdateSupplierInformation();
+      const serviceProcessSummary = new ProcessSummary();
```
andrewsignori-aot, line 44: Please move `serviceProcessSummary` inside the `if (pendingCASSuppliers.length)` where it will be used.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      const auth = await this.casService.casLogon();
+      if (auth.access_token) {
+        summary.info("Logon successful.");
+        suppliersUpdated = await this.requestCASAndUpdateSuppliers(
+          casSuppliers,
+          summary,
+          auth,
+        );
+      } else {
+        summary.info("Could not authenticate on CAS.");
```
andrewsignori-aot, line 54: A failure to authenticate on CAS should be treated as an error inside the method `casLogon`. If the HTTP response is different than 200 an exception can be thrown. Same for `requestCASAndUpdateSuppliers`.
For  `requestCASAndUpdateSuppliers` a 404 is expected when the student is not found and it should be considered an expected response where the method should return `undefined` or something that would indicate it.

andrewsignori-aot: Maybe we should throw a custom-named error when CAS auth fails to allow its detection in the processor and rethrown it to force the queue to retry the operation. Does it make sense?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      summary.info(`Requesting info for CAS supplier id ${casSupplier.id}.`);
+      let supplierResponse = null;
+      try {
+        supplierResponse = await this.casService.getSupplierInfoFromCAS(
+          auth.access_token,
+          casSupplier.student.sinValidation.sin,
+          casSupplier.student.user.lastName.toUpperCase(),
+        );
+      } catch (error: unknown) {
+        summary.error("Unexpected error while requesting supplier.", error);
```
andrewsignori-aot, line 85: If there was an unexpected error the method should be aborted and not proceed to line 87.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      }
+      if (supplierResponse.items.length > 0) {
```
andrewsignori-aot, line 87: Should we also ensure only ACTIVE ones will be considered and an "INACTIVE" will be handled as not found?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      {
+        supplierNumber: casSupplierResponseItem.suppliernumber,
+        supplierName: casSupplierResponseItem.suppliername,
+        status: casSupplierResponseItem.status,
+        supplierProtected: casSupplierResponseItem.supplierprotected === "Y",
+        lastUpdated: new Date(casSupplierResponseItem.lastupdated),
+        supplierAddress: supplierAddressToUpdate,
+        supplierStatus,
+        supplierStatusUpdatedOn: now,
+        isValid: true,
```
andrewsignori-aot, line 155: Should we ensure the record is `ACTIVE` to save it as `isValid: true`?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * Updates CAS supplier table.
+   * @param casSupplier CAS supplier to be updated.
+   * @param casSupplierResponseItem CAS supplier response item from CAS request.
+   * @param supplierStatus CAS supplier status to be updated.
+   * @returns update result.
+   */
+  async updateCASSupplier(
+    casSupplier: CASSupplier,
```
andrewsignori-aot, line 117: Please do not pass the entire object if only the ID would be required for processing.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+   * @param casSuppliers pending CAS suppliers.
+   * @param summary log summary.
+   * @param auth CAS auth details.
+   * @returns a number of update records.
+   */
+  private async requestCASAndUpdateSuppliers(
+    casSuppliers: CASSupplier[],
+    summary: ProcessSummary,
+    auth: CASAuthDetails,
+  ) {
```
andrewsignori-aot, line 73: Missing return type.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+    let suppliersUpdated = 0;
+    for (const casSupplier of casSuppliers) {
```
andrewsignori-aot, line 75: Just double-checking the team's understanding on this, we will not be performing parallel requests for CAS because the amount of expected daily requests would not be enough to justify it. Does that sound correct?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot, line 87: As a suggestion, if can be simplified as `if (supplierResponse.items.length)`.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      summary.info(`Requesting info for CAS supplier id ${casSupplier.id}.`);
```
andrewsignori-aot, line 76: A new child `ProcessSummary` can be created for each supplier processing. It would create a new orientation in the logs output.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      supplierAddressToUpdate = {
+        supplierSiteCode: activeSupplierAddress.suppliersitecode,
+        addressLine1: activeSupplierAddress.addressline1,
+        city: activeSupplierAddress.city,
+        provinceState: activeSupplierAddress.province,
+        country: activeSupplierAddress.country,
+        postalCode: activeSupplierAddress.postalcode,
+        status: activeSupplierAddress.status,
+        siteProtected: activeSupplierAddress.siteprotected,
+        lastUpdated: new Date(activeSupplierAddress.lastupdated),
```
andrewsignori-aot, line 137: Is the `activeSupplierAddress.lastupdated` will be properly persisted on DB. I believe that it is returning the date in PST/PDT without timezone which will be saved as UTC on DB.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+} from "../../processors/schedulers/cas-integration/models/cas-supplier-response.dto";
+import { Not, Repository, UpdateResult } from "typeorm";
+import { CASIntegrationConfig, ConfigService } from "@sims/utilities/config";
+import { CASService } from "./cas.service";
+
+const CAS_SUPPLIER_ADDRESS_ACTIVE_STATUS = "ACTIVE";
+
+@Injectable()
+export class CASSupplierIntegrationService {
+  private readonly casIntegrationConfig: CASIntegrationConfig;
```
andrewsignori-aot, line 18: `casIntegrationConfig` is not used.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,186 @@
...
+import { ProcessSummary } from "@sims/utilities/logger";
+import {
+  CASAuthDetails,
+  CASSupplierResponseItem,
+} from "../../processors/schedulers/cas-integration/models/cas-supplier-response.dto";
+import { Not, Repository, UpdateResult } from "typeorm";
+import { CASIntegrationConfig, ConfigService } from "@sims/utilities/config";
+import { CASService } from "./cas.service";
+
+const CAS_SUPPLIER_ADDRESS_ACTIVE_STATUS = "ACTIVE";
```
andrewsignori-aot, line 14: Since the possible values are known I would recommend defining the status as "ACTIVE | INACTIVE" or undefined.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/cas-supplier-integration.scheduler.ts
```diff
@@ -0,0 +1,80 @@
...
+      if (pendingCASSuppliers.length) {
+        processSummary.children(serviceProcessSummary);
+        suppliersUpdated =
+          await this.casSupplierIntegrationService.executeCASIntegrationProcess(
+            serviceProcessSummary,
+            pendingCASSuppliers,
+          );
+      }
+      processSummary.info("CAS supplier integration executed.");
+      await this.cleanSchedulerQueueHistory();
```
andrewsignori-aot, line 58: `cleanSchedulerQueueHistory` is taken care in the `finally`.

configs/env-example
```diff
@@ -147,3 +147,8 @@ WORKERS_PORT=
 
 #Load test gateway
 LOAD_TEST_API_PORT=
+
+# CAS integration configuration
+CAS_BASE_URL=
+CAS_CLIENT_ID=
+CAS_CLIENT_SECRET=
```
andrewsignori-aot, line 154: These variables need to be added to the `env-setup-deploy-secrets.yml` to allow their updates in Openshift.
Please use `GC_NOTIFY_API_KEY` as a reference to implement the changes.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
andrewsignori-aot: `cas.service.ts` should not be inside a `cas-supplier` folder.
The `cas.service.ts` will have the login and get supplier methods for now and it will receive more API methods in the near future.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Requests supplier information from CAS API.
+   * @param token auth token.
+   * @param sin student's sin.
+   * @param lastName student's last name.
+   * @returns CAS supplier response.
+   */
+  public async getSupplierInfoFromCAS(
```
andrewsignori-aot, line 57: Same from https://github.com/bcgov/SIMS/pull/3490/files#r1678457053

sources/packages/backend/apps/queue-consumers/test/helpers/mock-utils/cas-service.mock.ts
```diff
@@ -0,0 +1,72 @@
...
+      href: "https://localhost:8080/ords/cas/cfs/supplier/SMITH/lastname/123456789/sin",
+    },
+    {
+      rel: "describedby",
+      href: "https://localhost:8080/ords/cas/metadata-catalog/cfs/supplier/SMITH/lastname/123456789/item",
+    },
+  ],
+};
+
+export function createCASServiceMock(): CASService {
```
andrewsignori-aot, line 61: Missing comments.

sources/packages/backend/apps/queue-consumers/test/helpers/mock-utils/cas-service.mock.ts
```diff
@@ -0,0 +1,72 @@
+import { CASService } from "../../../src/services/cas-supplier/cas.service";
+
+export const casLogonMockedResult = {
+  access_token: "token123",
+  token_type: "123456789",
+  expires_in: 3600,
+};
+
+export const getSupplierInfoFromCASMockedResult = {
```
andrewsignori-aot, line 9: It should follow a const naming convention.

sources/packages/backend/apps/queue-consumers/test/helpers/mock-utils/cas-service.mock.ts
```diff
@@ -0,0 +1,72 @@
+import { CASService } from "../../../src/services/cas-supplier/cas.service";
+
+export const casLogonMockedResult = {
```
andrewsignori-aot, line 3: It should follow a const naming convention.

sources/packages/backend/apps/queue-consumers/test/helpers/mock-utils/cas-service.mock.ts
andrewsignori-aot, line 9: Would this not be the case for creating a factory that would allow the payload generation in different ways?
It can be created as simply as this PR requires and later be expanded, for instance, to add multiple addresses or have its values generated from using faker lib.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/_tests_/cas-supplier-integration.scheduler.e2e-spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    ).toBe(true);
+
+    expect(casServiceMock.casLogon).not.toHaveBeenCalled();
+    expect(casServiceMock.getSupplierInfoFromCAS).not.toHaveBeenCalled();
+  });
+
+  it("Should update CAS supplier table when found pending supplier information to be updated.", async () => {
+    const student = await saveFakeStudent(db.dataSource);
+
+    const casSupplier = new CASSupplier();
```
andrewsignori-aot, line 75: Following the existing patterns we should create a factory (e.g. `createFakeCASSupplier`) to help with the entity creation and evolve from there.

andrewsignori-aot (commented):
Great first effort for CAS integration. Please take a look at the comments.

.github/workflows/env-setup-deploy-secrets.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       ZEEBE_CLIENT_ID: ${{ secrets.ZEEBE_CLIENT_ID }}
       ZEEBE_CLIENT_SECRET: ${{ secrets.ZEEBE_CLIENT_SECRET }}
       CAMUNDA_OAUTH_URL: ${{ secrets.CAMUNDA_OAUTH_URL }}
+      CAS_BASE_URL: ${{ secrets.CAS_BASE_URL }}
+      CAS_CLIENT_ID: ${{ secrets.CAS_BASE_URL }}
+      CAS_CLIENT_SECRET: ${{ secrets.CAS_BASE_URL }}
```
andrewsignori-aot, line 50: Please review, all values are assigned from the same secret `CAS_BASE_URL`.

devops/openshift/queue-consumers-deploy.yml
```diff
@@ -200,6 +200,12 @@ objects:
                       name: ${QUEUE_CONSUMERS_SECRET_NAME}
                 - name: APPLICATION_ARCHIVE_DAYS
                   value: ${APPLICATION_ARCHIVE_DAYS}
+                - name: CAS_BASE_URL
+                  value: "${CAS_BASE_URL}"
+                - name: CAS_CLIENT_ID
+                  value: "${CAS_CLIENT_ID}"
+                - name: CAS_CLIENT_SECRET
+                  value: "${CAS_CLIENT_SECRET}"
```
andrewsignori-aot, line 208: These are secret values and should follow the same as below for every new key.
```yml
- name: GC_NOTIFY_URL
  valueFrom:
       secretKeyRef:
           key: ${GC_NOTIFY_URL_NAME_KEY}
           name: ${QUEUE_CONSUMERS_SECRET_NAME}
```

devops/openshift/queue-consumers-deploy.yml
```diff
@@ -414,3 +429,9 @@ parameters:
     value: esdc-environment-code
   - name: APPLICATION_ARCHIVE_DAYS
     require: true
+  - name: CAS_BASE_URL
+    required: true
+  - name: CAS_CLIENT_ID
+    required: true
+  - name: CAS_CLIENT_SECRET
+    required: true
```
andrewsignori-aot, line 437: Please follow the same pattern from other parameters and use the `value` with the expected value.

andrewsignori-aot, line 437: Please follow the same pattern from other parameters and use the `value` with the expected value.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/models/cas-supplier-response.dto.ts
andrewsignori-aot: Please also remove the `CASSupplierResponseItemLink` that no longer has references.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: Maybe I was not clear in my comments but I need a confirmation, when the authentication fails the HTTP response will just be returned as empty?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,248 @@
...
+        suppliersUpdated = await this.updateActiveAddress(
+          supplierResponse,
+          summary,
+          casSupplier,
+          suppliersUpdated,
+        );
+      } catch (error: unknown) {
+        if (error instanceof Error) {
+          summary.error(error.message);
+        } else {
```
andrewsignori-aot, line 102: Can you please clarify why we need to treat an `Error` differently?
Why the below would not be enough for both scenarios.
```ts
summary.error(
    "Error while requesting or updating CAS suppliers.",
    error,
);
```

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/_tests_/cas-response.factory.ts
```diff
@@ -0,0 +1,45 @@
+import { CASSupplierResponse } from "@sims/integrations/cas/models/cas-supplier-response.model";
+
+export function createFakeCASSupplierResponse(): CASSupplierResponse {
```
andrewsignori-aot, line 3: Comments missing.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/_tests_/cas-supplier.factory.ts
```diff
@@ -0,0 +1,16 @@
...
+  student: Student,
+) {
+  const casSupplier = new CASSupplier();
+  casSupplier.supplierStatus = SupplierStatus.PendingSupplierVerification;
+  casSupplier.supplierStatusUpdatedOn = new Date();
+  casSupplier.isValid = false;
+  casSupplier.creator = auditUser;
+  casSupplier.student = student;
+  return db.casSupplier.save(casSupplier);
+}
```
andrewsignori-aot, line 16: Please check some other factories to follow similar approaches.
- Add comments.
- Relations go under a relations object.
- Student can be created if not provided.
- Missing return type.

The PR comment that probably motivated the creation of this method was requesting the creation of the `createFake` only. The `saveFake` would be a good addition but was not a blocker.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrepestana-aot: I changed that to check the exception instead.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrepestana-aot: @ninosamson confirmed that verbally in a meeting.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+        summary.error("Unexpected error.", error);
+      }
+    }
+    return suppliersUpdated;
+  }
+
+  /**
+   * For each pending CAS supplier, request supplier information to CAS API and update local table.
+   * @param casSuppliers pending CAS suppliers.
+   * @param summary log summary.
```
andrewsignori-aot, line 64: Please update the comment to `parentProcessSummary`.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+      parentProcessSummary.children(summary);
+      summary.info(`Requesting info for CAS supplier id ${casSupplier.id}.`);
+      let supplierResponse = null;
+      try {
+        supplierResponse = await this.casService.getSupplierInfoFromCAS(
+          auth.access_token,
+          casSupplier.student.sinValidation.sin,
+          casSupplier.student.user.lastName.toUpperCase(),
+        );
+        suppliersUpdated = await this.updateActiveAddress(
```
andrewsignori-aot, line 85: The name `updateActiveAddress` can be misleading since the core goal would be to update the supplier itself. I would recommend something like `updateSupplier`.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+   * @param suppliersUpdated
+   * @returns a number of update records.
+   */
+  private async updateActiveAddress(
+    supplierResponse: CASSupplierResponse,
+    summary: ProcessSummary,
+    casSupplier: CASSupplier,
+    suppliersUpdated: number,
+  ) {
+    if (supplierResponse.items.length) {
```
andrewsignori-aot, line 115: As a suggestion, an early return can be done as below to simplify the method.
```ts
if (!supplierResponse.items.length) {
    summary.info("No supplier found on CAS.");
    return;
}
```

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+   * @param casSuppliers pending CAS suppliers.
+   * @param suppliersUpdated
+   * @returns a number of update records.
+   */
+  private async updateActiveAddress(
+    supplierResponse: CASSupplierResponse,
+    summary: ProcessSummary,
+    casSupplier: CASSupplier,
+    suppliersUpdated: number,
+  ) {
```
andrewsignori-aot, line 114: Missing return type.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (supplierResponse.items.length) {
+      const [supplierInfo] = supplierResponse.items;
+      const isThereAnActiveSupplierAddress =
```
andrewsignori-aot, line 117: `isThereAnActiveSupplierAddress` can be misleading because pass the idea that it is a boolean variable.
I would recommend `casSupplier`.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+    }
+    return suppliersUpdated;
+  }
+
+  /**
+   * Gets the first active supplier address from a list supplier response item.
+   * @param casSupplierResponseItem CAS supplier response item.
+   * @returns CAS supplier response item address.
+   */
+  private getActiveSupplier(
```
andrewsignori-aot, line 150: The active status exists at two levels, the supplier itself and his addresses.
Should this method check for both?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,236 @@
...
+    casSupplierResponseItem: CASSupplierResponseItem,
+  ): CASSupplierResponseItemAddress {
+    return casSupplierResponseItem.supplieraddress.find(
+      (address) => address.status === CAS_SUPPLIER_ADDRESS_ACTIVE_STATUS,
+    );
+  }
+
+  /**
+   * Updates CAS supplier table.
+   * @param casSupplier CAS supplier to be updated.
```
andrewsignori-aot, line 160: Please review the comment param name.

sources/packages/backend/apps/queue-consumers/test/helpers/mock-utils/cas-service.mock.ts
```diff
@@ -0,0 +1,28 @@
+import { CASService } from "@sims/integrations/cas/cas.service";
+import { createFakeCASSupplierResponse } from "../../../src/processors/schedulers/cas-integration/_tests_/cas-response.factory";
```
andrewsignori-aot, line 2: I do not see a problem moving `createFakeCASSupplierResponse` file to this same folder.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: I was suggesting removing the const and defining the `status` property as below.
```ts
export class CASSupplierResponseItemAddress {
  suppliersitecode: string;
  addressline1: string;
  addressline2?: string;
  addressline3?: string;
  city: string;
  province: string;
  country: string;
  postalcode: string;
  emailaddress?: string;
  accountnumber?: string;
  branchnumber?: string;
  banknumber?: string;
  eftadvicepref?: string;
  providerid?: string;
  status: "ACTIVE" | "INACTIVE";
  siteprotected?: string;
  lastupdated: string;
}
```

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,241 @@
...
+      parentProcessSummary.children(summary);
+      summary.info(`Requesting info for CAS supplier id ${casSupplier.id}.`);
+      let supplierResponse = null;
+      try {
+        supplierResponse = await this.casService.getSupplierInfoFromCAS(
+          auth.access_token,
+          casSupplier.student.sinValidation.sin,
+          casSupplier.student.user.lastName.toUpperCase(),
+        );
+        suppliersUpdated = await this.updateSupplier(
```
andrewsignori-aot, line 86: The method `this.updateSupplier` can return the number of affected rows or a boolean to indicate "updated or not".
I would refrain from passing the accumulator variable as an input parameter.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/cas-supplier-integration.scheduler.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (error.name === CAS_AUTH_ERROR) {
+          processSummary.error(
+            `Unable to request data to CAS due to an authentication error.`,
+          );
+        }
+      } else {
+        const errorMessage = "Unexpected error while executing the job.";
+        processSummary.error(errorMessage, error);
+        return [errorMessage];
+      }
```
andrewsignori-aot, line 78: The idea oof capturing the CAS auth error would be to try to force the queue to retry the operation, as below. Does it make sense?
```ts
if (error instanceof CustomNamedError) {
  if (error.name === CAS_AUTH_ERROR) {
   // Throw an error to force the queue to retry.
    throw new Error(`Unable to request data to CAS due to an authentication error.`),
    );
  }
  const errorMessage = "Unexpected error while executing the job.";
  processSummary.error(errorMessage, error);
  return [errorMessage];
}
```

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,241 @@
...
+    casSuppliers: CASSupplier[],
+    parentProcessSummary: ProcessSummary,
+    auth: CASAuthDetails,
+  ): Promise<number> {
+    let suppliersUpdated = 0;
+    for (const casSupplier of casSuppliers) {
+      const summary = new ProcessSummary();
+      parentProcessSummary.children(summary);
+      summary.info(`Requesting info for CAS supplier id ${casSupplier.id}.`);
+      let supplierResponse = null;
```
andrewsignori-aot, line 79: No need to declare the `supplierResponse` outside the `try` block.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,241 @@
...
+   */
+  private async updateSupplier(
+    supplierResponse: CASSupplierResponse,
+    summary: ProcessSummary,
+    casSupplier: CASSupplier,
+    suppliersUpdated: number,
+  ): Promise<number> {
+    if (!supplierResponse.items.length) {
+      summary.info("No supplier found on CAS.");
+      return;
```
andrewsignori-aot, line 118: It must return 0 since the method returns a number.

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -0,0 +1,78 @@
...
+    try {
+      const headers = {
+        Authorization: `Bearer ${token}`,
+      };
+      const config: AxiosRequestConfig = {
+        headers,
+      };
+      response = await this.httpService.axiosRef.get(url, config);
+    } catch (error: unknown) {
+      throw new Error("Unexpected error while requesting supplier.", error);
```
andrewsignori-aot, line 74: Should it not be as below for the proper stack trace?
```ts
throw new Error("Unexpected error while requesting supplier.", {
  cause: error,
});
```

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -0,0 +1,78 @@
...
+    };
+    const body = { grant_type: "client_credentials" };
+    const config: AxiosRequestConfig = {
+      headers,
+    };
+    const data = stringify(body);
+    try {
+      return await this.httpService.axiosRef.post(url, data, config);
+    } catch (error: unknown) {
+      throw new CustomNamedError(
```
andrewsignori-aot, line 44: I believe we should add a `logger.error` otherwise we may have no idea what caused the issue.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,235 @@
...
+import {
+  CASAuthDetails,
+  CASSupplierResponse,
+  CASSupplierResponseItem,
+  CASSupplierResponseItemAddress,
+} from "@sims/integrations/cas/models/cas-supplier-response.model";
+import { CAS_AUTH_ERROR } from "@sims/integrations/constants";
+
+const CAS_SUPPLIER_ADDRESS_ACTIVE_STATUS = "ACTIVE";
+const CAS_SUPPLIER_RESPONSE_ITEM_STATUS = "ACTIVE";
```
andrewsignori-aot, line 18: Do we still need those constants?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -0,0 +1,235 @@
...
+   * @returns true if updated a record.
+   */
+  private async updateSupplier(
+    supplierResponse: CASSupplierResponse,
+    summary: ProcessSummary,
+    casSupplier: CASSupplier,
+  ): Promise<boolean> {
+    if (!supplierResponse.items.length) {
+      summary.info("No supplier found on CAS.");
+    } else {
```
andrewsignori-aot, line 112: Early return can be done.
```ts
if (!supplierResponse.items.length) {
  summary.info("No supplier found on CAS.");
  return false;
}
```

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -0,0 +1,84 @@
...
+        headers,
+      };
+      response = await this.httpService.axiosRef.get(url, config);
+    } catch (error: unknown) {
+      throw new Error("Unexpected error while requesting supplier.", {
+        cause: error,
+      });
+    }
+    return response?.data;
+  }
```
andrewsignori-aot, line 81: Please add a blank line.

andrewsignori-aot (approved):
Thanks for making the changes, great effort, looks good 👍

PR #3497, opened 2024-07-08
#2990 - Supporting User Footer

katrina-aot-git:
Change when supporting user footer is hidden (keep the footer visible on repeat searches).

### DIFF
--- a/sources/packages/web/src/views/supporting-user/SupportingInformation.vue
+++ b/sources/packages/web/src/views/supporting-user/SupportingInformation.vue
@@ -237,7 +237,6 @@ export default defineComponent({
               break;
           }
         } else {
-          showNav.value = false;
           snackBar.error(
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/views/supporting-user/SupportingInformation.vue
```diff
@@ -237,6 +237,7 @@ export default defineComponent({
               break;
           }
         } else {
+          showNav.value = false;
```
andrepestana-aot, line 240: I don't think we need this line as it was already defined as false on line 224.

katrina-aot-git: updated, I'd read the else as incorrectly a level out earlier

andrepestana-aot (approved):
Thanks for checking my comment. Great job!

Issue #3377, opened 2024-06-04
Virus Scanning - Perform the Virus Scan


andrewsignori-aot:
User Story
As a Security Analyst for SIMS, I would like to ensure that malware is not propagated VIA SIMS due to the reputational risk and as such require that all files received via SIMS are scanned. The scan happens at regular intervals rather than in real-time which means that it can be handled asynchronously instead of right away when the file is received. This ticket implements the virus execution once the ClamAV server is deployed to OpensShift and is also available for local development.

**Acceptance Criteria**

- [ ] Enable a queue-consumers method to execute the scan in all uploaded files, similar to `start-application-assessment`.
  - [ ] It should be enabled as a queue, not a scheduler, to allow the files to be scanned as soon as possible.
  - [ ] Create new columns on DB to track the scanning processes (the below ones are suggestions).
    -  virus_scan_status: `Pending`, `In-progress`, `Virus detected`, `File is clean` (the enum types are suggestions).
    -  virus_scan_status_updated_on
- [ ] Define the lib to execute the scan, for instance, [client libraries](https://www.npmjs.com/search?q=clamav). If there is a good nest nest-js lib, please try to use it (e.g. https://www.npmjs.com/package/nestjs-clamscan).
- [ ] Validate scan failure using the EICAR Test file
- [ ] Execute a few manual uploads with 15Mb files (or something close to it).

**Context**

- Deployed into a6ef19-dev via Helm, requires limits of 3GB RAM and 1CPU per pod
A [Common Service](https://bcgov.github.io/common-service-showcase/services/cavms.html) exists for the signature files.
- Common Service deployment [source](https://github.com/bcgov/common-hosted-clamav-service)  
- Clone and create deployable YAML`helm template dev . --namespace licenseplate-dev`
- Run locally via docker `docker run -it --rm --platform linux/amd64 -p 3310:3310 --name clamav ghcr.io/bcgov/clamav-unprivileged:ca3d42f3dde3c5aa9bcab636f752119bbe6a67e8`
- [Sample code](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1dVQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--736af41447165f1f912cdfc83721e8d68e18ded4/clamtest.zip)
uses Docker above, unzip and perform npm i then npm start

[...]

PR #3501, opened 2024-07-10
#3377- Virus Scanning bug Fix

sh16011993:
### The following bug was fixed as a part of this PR:

- Since the ClamAV port is passed from the Makefile, it should be set to `required: true` in the queue-consumers-deploy.yml


guru-aot (approved):
LGTM, nice work @sh16011993

Issue #2706, opened 2024-01-19
Student is blocked from accepting Assessment ( Disability Funding not verified, Funding Restrictions, MSFAA not Signed) 


HRAGANBC:
**User Story**
As a Student, I should be blocked from accepting my assessment until:
    - My disability profile status is verified (only applicable if they have requested disability funding on their application)
    - My MSFAA is Signed 
    - I have resolved restrictions which have the 'block funding' effect.
So that I am prompted to deal with outstanding items on my application before my study start date. 

**Acceptance Criteria:**
- [x] Do not allow student to **accept** their assessment if: (i.e. disable the button in the UI/UX) for any of the following situations:
    - [ ]  Application requests disability funding & disability profile status is still NOT verified
    - [ ]  The MSFAA is not valid
    - [ ]  Restrictions are preventing their funding from being issued.  
- [x]  If any above condition is =true, the accept assessment button should not be clickable
- [x]  If any above condition is =true, display a warning banner at the top of the Notice of Assessment Page (the location where the 'accept assessment' button is)
- [x]  If multiple blocking conditions display all information under the same banner as a bulleted list (similar to the blocked disbursement tracker card).

**Content for banners:**  
Disability Check Banner:
>Your account has not been approved for disability funding. You will not be able to accept this application until the account disability status has been approved. If you would like to receive all non disability funding please edit your application.

MSFAA Check Banner:
>Your MSFAA is not valid. Please complete your MSFAA with the National Student Loans Centre to move forward with your application. Please note, there is a one day delay between signing your MSFAA and being able to accept your assessment. 

Restrictions Check Banner:
>You have restrictions that block funding on your account. Please resolve them in order to move forward with your application.

**Additional Context:**
    - To confirm criticality for PT MVP Scope. Can we rely on communications instead to ensure students are aware of the below things blocking disbursement? I.e. a banner message which immediately applies to their account right after accepting their assessment if there's anything blocking their eventual eCert, AND/OR an email notification to the student when any such conditions exists. 
    - This ticket does not impact the student's application status tracker in any way
   
**Technical**
- [x] Reuse the same logic from [application.students.controller.ts](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts#L782) to get the list of all failed validations.
- [x] Adapt the query used by the validation method to allow applications other than `Completed`.
[...]

PR #3503, opened 2024-07-10
#2706 - add a list form for banners

katrina-aot-git:
Extend banner components to display a list of content.
image

### DIFF
--- a/sources/packages/web/src/components/generic/Banner.vue
+++ b/sources/packages/web/src/components/generic/Banner.vue
@@ -12,6 +12,13 @@
       <v-col>
         <div class="label-value-normal">
           <slot name="content">{{ summary }}</slot>
+          <slot name="content-list" v-if="summaryList">
+            <ul>
+              <li v-for="item in summaryList" :key="item">
+                {{ item }}
+              </li>
+            </ul>
+          </slot>
         </div>
       </v-col>
       <v-col cols="auto"> <slot name="actions"></slot></v-col>
@@ -33,6 +40,9 @@ export default defineComponent({
     },
     summary: {
       type: String,
+    },
+    summaryList: {
+      type: Array<string>,
     },
   },
   setup(props) {
--- a/sources/packages/web/src/components/generic/Banner.vue
+++ b/sources/packages/web/src/components/generic/Banner.vue
@@ -14,7 +14,7 @@
           <slot name="content">{{ summary }}</slot>
           <slot name="content-list" v-if="summaryList">
             <ul>
-              <li v-for="item in summaryList" :key="item">
+              <li v-for="(item, index) in summaryList" :key="index">
                 {{ item }}
               </li>
             </ul>
--- a/sources/packages/web/src/views/student/NoticeOfAssessment.vue
+++ b/sources/packages/web/src/views/student/NoticeOfAssessment.vue
@@ -6,8 +6,8 @@
         class="mb-2"
         :type="BannerTypes.Warning"
         header="Warning"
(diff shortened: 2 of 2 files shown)


sources/packages/web/src/views/student/NoticeOfAssessment.vue
```diff
@@ -140,11 +143,11 @@ export default defineComponent({
             break;
           case ECertFailedValidation.MSFAACanceled:
           case ECertFailedValidation.MSFAANotSigned:
-            warningText.value.push(
-              `Your MSFAA is not valid. 
-              Please complete your MSFAA with the National Student Loans Centre to move forward with your application. 
-              Please note, there is a one day delay between signing your MSFAA and being able to accept your assessment.`,
-            );
+            if (!warningText.value.includes(MSFAAWarning)) {
```
andrepestana-aot, line 146: Can it have a "Canceled" and "Not signed" eCert validation for the same assessment?

sources/packages/web/src/views/student/NoticeOfAssessment.vue
```diff
@@ -6,8 +6,8 @@
         class="mb-2"
         :type="BannerTypes.Warning"
         header="Warning"
-        :summary="warningText.join('\n')"
-      />
+        :summaryList="warningText"
```
andrepestana-aot, line 9: I think it makes more sense to call it a `warningList`

sources/packages/web/src/components/generic/Banner.vue
```diff
@@ -12,6 +12,13 @@
       <v-col>
         <div class="label-value-normal">
           <slot name="content">{{ summary }}</slot>
+          <slot name="content-list" v-if="summaryList">
+            <ul>
+              <li v-for="item in summaryList" :key="item">
```
andrepestana-aot, line 17: Good job working with the `v-for`!

I think that would be more appropriate using the index for the key though:

```ts
<div v-for = "(title, index) in news" :key="index">
     {{ title }}
</div>
```

sources/packages/web/src/views/student/NoticeOfAssessment.vue
katrina-aot-git: That is why I added this actually - the test case I was working with this time had that.

andrepestana-aot (commented):
Good job working with the v-for.  I left some comments. Please take a look.

sources/packages/web/src/views/student/NoticeOfAssessment.vue
andrepestana-aot: I'm just wondering if that is a real possible scenario for production.

katrina-aot-git: I'm not sure if it would be realistic, but I don't think it hurts anything having a case there for it. Do you need me to remove it?

sources/packages/web/src/components/generic/Banner.vue
katrina-aot-git: Updated

sources/packages/web/src/views/student/NoticeOfAssessment.vue
katrina-aot-git: Updated

sources/packages/web/src/views/student/NoticeOfAssessment.vue
andrepestana-aot: ok. That's fine to keep it. Thanks.

Issue #3469, opened 2024-06-28
Content: Student: Request a change instructions 


JerPearson:
As a student I want clear instructions when completing a change request so I know what needs to be done.

**Acceptance Criteria**
- [x] Add very clear language to student whenever they initiate a Request a Change
- [x] Revise header: "Fill in the field(s) below" to "Complete all question(s) below"
- [x] Revise text: "StudentAid BC will review your application......" to "All requested changes will be reviewed by StudentAid BC after you submit for review."
- [x] Revise and bold text: "Please note:" to "**Instructions**:" and tab the text and bulleted list down to separate from the above text

<img width="1466" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWk2QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--19dd70b6e6bc46c959865d11f5ad46f09ed0201e/image.png" alt="image.png" />

**Additional Context**
- Miscommunication on ticket [#2846](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/2846) which is why further changes are required.

JerPearson:
@Joshua-Lakusta edits made to content including  revising text from your previous ticket.  Can you take a look and confirm the changes to your previous text.

PR #3517, opened 2024-07-12
#3469 - Request a change update

katrina-aot-git:
image
Update to the second part of the request a change form page.


ninosamson (approved):
Looks good!

Issue #3491, opened 2024-07-08
Content: Ministry: Approve Account Request


JerPearson:
As a Ministry user I want clear instructions when approving a student account request.

**Acceptance Criteria**

- [x] Revise button text: "Create account for student" to "Create student account"
- [x] Revise model title: "Create account for student" to "Create student account"
- [x] Revise: "This will allow the student to access the system using a Basic BCeID account...." to "Attention: Approved account requests are final and cannot be deactivated.  Please ensure all supporting documentation has been carefully reviewed prior to approval." 

<img width="1025" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3EvQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--57f96437ce1d237b86aa629dec1c9f47f8f1885a/image.png" alt="image.png" />

PR #3524, opened 2024-07-15
#3491 - Ministry student account creation

katrina-aot-git:
Update ministry student account creation wording.
image
image

### DIFF
--- a/sources/packages/web/src/views/aest/student/StudentAccountApplicationsApproval.vue
+++ b/sources/packages/web/src/views/aest/student/StudentAccountApplicationsApproval.vue
@@ -24,7 +24,7 @@
                   color="primary"
                   @click="createStudentAccount"
                   :disabled="notAllowed"
-                  >Create account for student</v-btn
+                  >Create student account</v-btn
                 >
               </template>
             </check-permission-role>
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/views/aest/student/StudentAccountApplicationsApproval.vue
```diff
@@ -24,7 +24,7 @@
                   color="primary"
                   @click="createStudentAccount"
                   :disabled="notAllowed"
-                  >Create account for student</v-btn
```
dheepak-aot, line 27: This is the text on the button, I couldn't see an AC to revise button text and would request to have the business add the AC. Let me know if I am missing something.

katrina-aot-git: It was the same "Create account for student" text so I figured the first AC applied to both places that had that phrase on the page.

dheepak-aot: Absolutely makes sense. But the AC doesn't seem to be specific on mentioning both places and requires to be updated.

dheepak-aot: Got confirmation from the business on the same. Approving the PR.

dheepak-aot (approved):
Looks good 👍

PR #3532, opened 2024-07-16
Camunda SDK and Camunda Localhost Upgrade

andrewsignori-aot:
- Upgraded Camunda SDK to the latest version.
- Upgraded local Camunda versions to the latest available from https://github.com/camunda/camunda-platform/blob/main/.env


dheepak-aot (approved):
Looks good 👍

PR #3537, opened 2024-07-18
#3255 - CAS - deploy error fix

andrepestana-aot:
- Removed the params in the Makefile for queue-consumers deploy;
- Renamed params in queue-consumers-deploy.yml to end with "_NAME_KEY" instead of "_KEY".


andrewsignori-aot (approved):

Issue #1366, opened 2022-08-11
Change Deletes to soft deletes Part 2 of 2


abschwenker:

andrewsignori-aot:
As per the analysis, student files are never deleted only the relationship between an application and its files is deleted while the application is in draft. Once the application is summited the files are no longer deleted.
According to the confirmation received from @CarlyCotton and @Joshua-Lakusta, there are no concerns about keeping the current behavior since the application only starts to matter when it is submitted.

PR #3542, opened 2024-07-18
#1366 - Set FileOriginType for Student Application Files

andrewsignori-aot:
As per the analysis, student files are never deleted, only the relationship between an application and its files is deleted while the application is in draft. Once the application is summited the file relations are no longer deleted.
According to the confirmation from @CarlyCotton and @Joshua-Lakusta, there are no concerns about keeping the current behavior since the application only starts to matter when submitted.

### Minor improvement

Student application files were never updated from their `Temporary` status because the status was added later.
This PR is now updating the file origin when an application is saved.


dheepak-aot (approved):
Looks good @andrewsignori-aot 👍

PR #3543, opened 2024-07-18
#3255 - CAS logon bugfix

andrepestana-aot:
- Changed to return `data` prop from CAS logon response.

### DIFF
--- a/sources/packages/backend/libs/integrations/src/cas/cas.service.ts
+++ b/sources/packages/backend/libs/integrations/src/cas/cas.service.ts
@@ -40,7 +40,7 @@ export class CASService {
     };
     const data = stringify(body);
     try {
-      return await this.httpService.axiosRef.post(url, data, config);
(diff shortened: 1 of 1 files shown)


sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -40,7 +40,7 @@ export class CASService {
     };
     const data = stringify(body);
     try {
-      return await this.httpService.axiosRef.post(url, data, config);
+      return (await this.httpService.axiosRef.post(url, data, config)).data;
```
andrewsignori-aot, line 43: For these cases, we create a const most of the time. I would recommend doing the same here.
```ts
const response = await this.httpService.axiosRef.post(url, data, config);
return response .data;
```

andrewsignori-aot (approved):
Approving since the only comment would not be a blocker. Looks good 👍

Issue #3502, opened 2024-07-10
Bug Fix for error in workers pods


guru-aot:
It seems there is some error when the camunda tries to connect with the zeebe libraries configurations, but this is not related to the health check or the process running in workers. Currently we are not facing any downtime or queue processing delays due to the above error. I also confirm there is no crash loop error happening in any of the environments/namespaces.
Also the restart was happening like 3 time from July 4, which is normal and not due to this issue or the health check.

The issue may be a cause of the library used for health check by our health indicator https://github.com/bcgov/SIMS/blob/7b072e5ca8ccc40bb693743385d877ef2c02431c/sources/packages/backend/apps/workers/src/zeebe/zeebe-health-indicator.ts? 
It can be also the camunda_outh_url which has been updated as part of last release Release v1.9.0.

PFB the processing logs.

<img width="1401" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK0hNQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5f4aaa743b4887cdd7bf2333283b52ab08977ef6/image.png" alt="image.png" />

<img width="1425" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMnZTQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--433391976304d7f9da27dcc529b5578598a47aab/image.png" alt="image.png" />

andrewsignori-aot:
## Analysis
 The cancellation error seems to be generated due to conflicted configurations with their default values set to 60s.
- `ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS`: how long in seconds the long poll Job Activation request is held open by a worker.
- `DEFAULT_JOB_ACTIVATION_TIMEOUT`:  timeout for the GRPC call to activate a job.

With both variables set to 60s, the `DEFAULT_JOB_ACTIVATION_TIMEOUT` throws the cancellation error when the timeout is reached.

### Proposed solution

Set the variables as below to allow the long pooling to restart before the job activation reaches its timeouts.
- ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS: 60
- DEFAULT_JOB_ACTIVATION_TIMEOUT: 65

[...]

PR #3546, opened 2024-07-19
#3502 - Workers PODs Logging Cancellation Errors

andrewsignori-aot:
## Problem analysis

Having the long pooling configuration set to a value equal to or greater than 60 was causing `Grpc Stream Error: 1 CANCELLED` errors. Every time such an error happened the worker was marked as disconnected and took 1 to 2 seconds to be marked back as connected which sometimes led to health checks failure.

The previous [Zeebe lib had the long pooling configured as 30s](https://www.npmjs.com/package/zeebe-node#long-polling), after the upgrade to Camunda SDK 8 [the new default value became 60s](https://github.com/camunda/camunda-8-js-sdk/blob/d6acdfddcef8413226e3366932df5b6bda234e47/src/lib/Configuration.ts#L335-L339).

This value directly impacts the timeouts used while [initializing a new GRPC stream to activate jobs](https://github.com/camunda/camunda-8-js-sdk/blob/05b3ecb84d26e6b507a09ac2df8d7794532e5621/src/zeebe/lib/ZBWorkerBase.ts#L556-L588).

### Proposed solution

- The framework allows the pooling configuration using the `ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS` env variable configured to 45 seconds. After setting this value manually in the Openshift Workder DC there was an immediate improvement.
- Removed `readinessProbe` since the POD does not receive network traffic and the check is the same as the `livenessProbe` (as agreed in Teams chat).
- Small improvement in the Zeebe health check to provide a clear message about which workers are failing.


devops/openshift/workers-deploy.yml
```diff
@@ -191,6 +187,9 @@ parameters:
   - name: CAMUNDA_OAUTH_URL_KEY
     value: camunda-oauth-url
     require: true
+  - name: ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS_KEY
+    value: zeebe-grpc-worker-longpoll-seconds
+    require: true
```
dheepak-aot, line 192: required ?

dheepak-aot:
Looks good. 👍Just one comment.

dheepak-aot (approved):
Thanks for the great investigation and fix. Also for making the change. Looks good. 👍

Issue #2973, opened 2024-03-08
File Uploader Bug in Web UI


lewischen-aot:
**User Story**
As a student, I want to be able to upload a file using the File Uploader and also view the file with its original content by downloading it. As a ministry user, I want to view the original file by downloading it from the File Uploader. 

**Current Behavior**
When a student uploads a file and immediately download it by clicking on the file, the downloaded file doesn't have the original content. The same error happens when the ministry user downloads the file from the File Uploader. It is known to happen for `text files. The downloaded text file will display [object object] as shown below.
<img width="508" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFRJQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5afa71372c54016d615d770d98c989b507f7a01c/image.png" alt="image.png" />

**Expected Behavior**
The uploaded file on File Uploaded should have the original file content after downloading.

**Acceptance Criteria**
- [ ] The downloaded file should have the original file content after downloading using the File Uploader
- [ ] Ensure all accepted files (JPG, PNG, DOC, DOCX, PDF, TXT) have the correct behavior.

guru-aot:
There is a change in the feature when the student uploads a file. Currently this bug is created, without virus scanning is enabled. We will have a look on the bug again, once the ticket #3297 for UI enhancement for both Ministry and Student user is complete.

PR #3556, opened 2024-07-23
#2973 - File Uploader Bug

sh16011993:
### As a part of this PR, the following file uploader bug was fixed:

**Bug:** The text of the file was not displaying properly. Instead, it was displaying [Object object]

**Fix:** Updated the `FormUploadService` `downloadFile` method to use the `fileUtils.downloadStudentDocument` method which ensures the proper download of the `blob` type response.

**Screenshot:**

image

### DIFF
--- a/sources/packages/web/src/services/FormUploadService.ts
+++ b/sources/packages/web/src/services/FormUploadService.ts
@@ -1,4 +1,4 @@
-import { FileOriginType, FormUploadFileInfo } from "@/types";
+import { FormUploadFileInfo } from "@/types";
 import { AxiosRequestConfig } from "axios";
 import ApiClient from "../services/http/ApiClient";
 import { useFileUtils } from "@/composables";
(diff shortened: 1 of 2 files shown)


sources/packages/web/src/services/FormUploadService.ts
```diff
@@ -46,18 +47,13 @@ export default class FormUploadService {
...
-        storage: "base64",
-        url: fileContent,
-        originalName: fileInfo.originalName,
-        name: fileInfo.name,
-        type: fileInfo.type,
+      const fileUtils = useFileUtils();
+      const studentDocumentDetails = {
+        fileName: fileInfo.originalName,
+        uniqueFileName: fileInfo.name,
+        fileOrigin: FileOriginType.Temporary,
```
dheepak-aot, line 54: Because of the reason that input model of  `downloadStudentDocument` is expecting  fileOrigin we are manualling setting the value to `FileOriginType.Temporary`. I would say update the type of `downloadStudentDocument` input to expect the properties required.

dheepak-aot: Something like this. 
image

dheepak-aot:
Great work. Just one comment.

dheepak-aot (approved):

PR #3559, opened 2024-07-23
#3560 - Migration revert bug fix

sh16011993:
### As a part of this PR, the following bug has been fixed:

**Bug:** Migration revert for `Rollback-drop-fin-process-provincial-daily-disbursements-integration.sql` was failing on staging environment because of the missing `sims` schema name and the `queue_settings` not nullable column.

**Fix:** Fixed the query.

**Note:** Tested the migration reverts till the v1.9.0 release to ensure that all the migration reverts are happening properly.


dheepak-aot (approved):
Thanks for taking care. Looks good 👍

Issue #3561, opened 2024-07-23
Prettier formatting 


dheepak-aot:
**Describe the Bug**
 

**Expected Behaviour**

**Actual Behaviour**

** Steps To Reproduce**
Steps to reproduce the behaviour:

PR #3563, opened 2024-07-23
#3561 - Fix code formatting

dheepak-aot:
## Code formatting warning

Code formatting in few of the files were not in sync with the common settings that we have been following. Fixed those files listed here.

image


andrewsignori-aot (approved):
Thanks for taking care of it, looks good 👍

Issue #2919, opened 2024-03-04
Warning Message on Offering Edit/Creation


CarlyCotton:
**Describe the Bug**
A warning message is shown when an institution account is editing (and creating?) an offering. This also may require investigation on the bulk upload.

**Expected Behaviour**
No additional warning banner beyond the "_____ is required" message.

**Actual Behaviour**
"_____ is required" message and additional banner message regarding costs that may exceed eligible amounts.
<img width="613" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWF0QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d931391a1f25b3afbb2453ac0de378e47c3c7bc5/image.png" alt="image.png" />
<img width="1213" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWl0QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0abbc5dcc49dd2d8081e5a84522081819c75d722/image.png" alt="image.png" />

** Steps To Reproduce**
1. Edit offering
2. Enter dollar amount for Tuition, Books and Supplies, Mandatory fees, or Exceptional Costs
3. Remove the dollar amount entered for any of the above fields so the field is blank
<img width="1240" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOVd0QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3693788fa3570efbe5f85e678c5c68504c4747e4/image.png" alt="image.png" />

**Technical**
- [x] Change the offering validator to generate the warning only if the four required costs are present.
- [ ] Adjust the warning validator key `offeringCostExceedMaximum` in the [OfferingValidationModel](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/services/education-program-offering/education-program-offering-validation.models.ts).
- [x] Do a manual bulk-offering upload to ensure that the validation is working as expected. There are E2E tests present with file examples.

andrewsignori-aot:
Moved the ticket back to "In progress" to allow a technical change.

andrewsignori-aot:
[...]

PR #3567, opened 2024-07-24
#2919 - Institution offering warning

katrina-aot-git:
Add blocking for institution offering costs warning when a field is empty.


andrewsignori-aot (approved):

Issue #3351, opened 2024-05-30
Student: FORM Additional transportation question form updatesFT/PT 


CarlyCotton:
**User story**
As a _student user_, I want the additional transportation sections updated so that I can make better decisions on my application and the questions should be correct to properly assess.

### **Acceptance Criteria**
**Additional transportation cost application question**
- [ ] All changes should be reflected in the forms for the application and Change Request

- [ ] Remove the "please estimate your weekly transportation costs." field entirely 

- [ ] Add a new radio option selection That will be dynamic and when answered will expand to show all questions after it (see image 2 for visual explanation of dynamic behaviour)

> “Are you listed as a driver on the insurance for the vehicle you will be using?” Radio options of yes or no
- [ ] When the answer to the new question is no red banner should appear 

> **Warning**
> You must be a listed driver on the insurance of the vehicle you will be using to qualify for additional transportation funding.

- [ ] If the answer to this question is "no" also do not show the following questions :
- new question around vehicle ownership
- Please estimate your weekly travel distance from home to school in kilometers:
- How many weeks during your study period will you require additional transportation costs?
- Are you receiving a weekly transportation allowance from your institution for a clinical or practicum placement

- [ ] Add a new radio option selection that will sit above the others questions that are also hidden under the listed driver question.

> Are you the owner of the vehicle? options yes or no

<p></p>

- [ ] Variable naming to match following table

| Question | Response | Old Variable | New Variable | 
|----------|----------|----------|----------|
| 1. Do you want addtl transport?  | Yes/No   | studentDataEligibleForAnAdditionalTransportationAllowance   | studentDataAdditionalTransportRequested  |
| 2. Are you the listed driver?    | Yes/No  | _n/a_  | studentDataAdditionalTransportListedDriver   |
| 3. Are you the owner?    | Yes/No  | _n/a_  | studentDataAdditionalTransportOwner   |
| 4. Weekly KM     | 1-999   | studentDataAdditionalTransportKm   | _no change_   |
| 5. Is their a clinical practicuum?     | Yes/No  | studentDataAdditionalTransportPlacement   | _no change_   |
| 6. Weekly travel costs    | 1-999  | studentDataAdditionalTransportCost | **removed**  |

**Additional information**
IMAGE 1
[...]

PR #3570, opened 2024-07-25
#3351 - Form Additional Transportation Question (FT/PT)

sh16011993:

- Updated the application forms with the additional transportation question
- Updated the additional transportation appeals form with the additional transportation question
- Updated the formio for the updated variable as per the table and removed the prefix `studentData` as the variable given is in camunda format in both the forms - Application and Change Requests

**Note:** I have confirmed with @Joshua-Lakusta and the red banner as shown in the image is as expected.

### Screenshots:

**Application Screenshots:**

image

----------------------------------------------------------------

image

----------------------------------------------------------------

**Appeal Screenshots:**

image

-----------------------------------------------------------------

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2022-23.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2022-23.json
@@ -10684,8 +10684,7 @@
               "allowMultipleMasks": false,
               "addons": [],
               "id": "evyzdwp",
-              "hideLabel": true,
-              "isNew": false
+              "hideLabel": true
             },
             {
               "key": "disabilityInformationPanelContent",
@@ -30386,8 +30385,7 @@
               "fieldSet": false,
               "id": "e26wllo",
               "defaultValue": "",
-              "lockKey": true,
-              "isNew": false
+              "lockKey": true
             },
             {
               "key": "additionalTransportationCostContent",
@@ -30576,9 +30574,9 @@
                   "tags": [],
                   "properties": {},
                   "conditional": {
-                    "show": "true",
-                    "when": "additionalTransportListedDriver",
-                    "eq": "yes",
+                    "show": "",
+                    "when": "",
+                    "eq": "",
                     "json": ""
                   },
                   "customConditional": "",
@@ -30639,9 +30637,9 @@
                   ],
                   "key": "transportationcostsituation",
                   "conditional": {
-                    "show": "true",
-                    "when": "additionalTransportListedDriver",
-                    "eq": "yes"
+                    "show": "",
+                    "when": "",
+                    "eq": ""
                   },
                   "type": "radio",
                   "input": true,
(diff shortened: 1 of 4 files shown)


sources/packages/forms/src/form-definitions/sfaa2023-24.json
```diff
@@ -30565,7 +30746,7 @@
                   "tags": [],
                   "conditional": {
                     "show": "true",
-                    "when": "eligibleForAnAdditionalTransportationAllowance",
+                    "when": "additionalTransportListedDriver",
```
dheepak-aot, line 30749: I know the existing code was applying the hide/show condition to every question, but the condition is at the panel level and we do not have to repeat the same condition to every item inside the panel unless I am missing something. 

image

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -30290,7 +30288,128 @@
...
+              "encrypted": false,
+              "redrawOn": "",
+              "clearOnHide": true,
+              "customDefaultValue": "",
+              "calculateValue": "",
+              "calculateServer": false,
+              "allowCalculateOverride": false,
+              "validate": {
+                "required": true,
+                "onlyAvailableItems": true,
```
dheepak-aot, line 30336: 👍

sources/packages/forms/src/form-definitions/studentadditionaltransportationappeal.json
```diff
@@ -143,6 +143,72 @@
               "properties": {},
               "lockKey": true
             },
+            {
+              "autofocus": false,
+              "input": true,
+              "tableView": true,
+              "inputType": "radio",
+              "label": "Are you listed as a driver on the insurance for the vehicle you will be using?",
```
dheepak-aot, line 151: I believe `onlyAvailableItems:true` should be there for appeal forms as well ? Is there a reason for that nbot being here?

dheepak-aot:
Nice work @sh16011993 . Please have a look at the comments.

sources/packages/forms/src/form-definitions/studentadditionaltransportationappeal.json
```diff
@@ -151,6 +219,49 @@
...
+                  ],
+                  "defaultValue": "",
+                  "protected": false,
+                  "fieldSet": false,
+                  "persistent": true,
+                  "hidden": false,
+                  "clearOnHide": true,
+                  "validate": {
+                    "required": true,
+                    "onlyAvailableItems": true,
```
dheepak-aot, line 249: Thanks for doing the change.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -30290,7 +30288,128 @@
...
+              "conditional": {
+                "show": "true",
+                "when": "additionalTransportRequested",
+                "eq": "yes",
+                "json": ""
+              },
+              "customConditional": "",
+              "logic": [],
+              "attributes": {
+                "data-cy": "eligibleForAnAdditionalTransportationAllowance"
```
dheepak-aot, line 30359: Not a blocker and data-cy is not being used anymore. But if used please do not use the old variable name.

dheepak-aot (approved):
Thanks for doing the changes @sh16011993. 

Approved with one minor comment on data-cy which is not a blocker. But please have look into it.

Issue #3539, opened 2024-07-18
Award Configuration Round - PT


CarlyCotton:
**User Story**
As a ministry user, I need to ensure proper rounding or truncating takes place to ensure our requests via ecert do not inadvertently cause rejected records which would delay or prevent student funding. 

### Acceptance Criteria
- [x] **Single Disbursements:** Award values in the configuration need to always be rounded down
    - $17.80 to $17
    - $17.02 to $17
- [x] **Two disbursements:** Award values in the first disbursement need to be the final award divided in half and rounded up if .5 or more (101/2 = 50.5 results in 51. 102.8/2 = 51.4 results in 51.)
- [x] **Two disbursements:** Award values in the second disbursement need to be the final award rounded down (floor) less the first disbursement amount

### Business Context
| Award | Half | 1st Disb | 2nd Disb |
|----------|----------|----------|----------|
| 350    | 175   | 175   | 175 (350-175)  |
| 375    | 187.5  | 188   |  187 (375-188)  |
| 100.5    | 50.25   | 50   | 50 (100 -50)   |
| 100.9    | 50.45   | 50   | 50 (100 -50)   |
| 101.9    | 50.95   | 51   | 50 (101-51)   |

Ecert records do not accept any partial dollars (cents).
Records that exceed program year or weekly limits could result in unnecessary delays to student funding, or prevent it all together. This is possible for any record that is >$1 the limit.
SFAS Award Configuration p.22 (AW2022P1.docx):

> Loan and grant awards must not have fractional values.   **Where a fractional value could result from a calculation, rounding to the nearest dollar should take place, EXCEPT for the federal grants, for which cents are to be dropped (special SABC instructions this year, as the federal grant per week amounts will have cents within them)**.
Attempts to avoid round-off problems should be used where possible.   

*Technical*
- [x] Update bpmn `parttime-configure-disbursement` with new rounding logic in the configure payload. No further code change outside this workflow is expected.
Ref: [Feel expression numeric functions](https://docs.camunda.io/docs/components/modeler/feel/builtin-functions/feel-built-in-functions-numeric/)
- [x]  While rounding off multiple disbursements, the first disbursement value is subtracted from round down value of actual award. 

- [x] E2E tests with at least one single disbursement scenario and given multiple disbursement scenarios.

| Award | Half | 1st Disb | 2nd Disb |
|----------|----------|----------|----------|
| 375    | 187.5  | 188   |  187 (375-188)  |
[...]

PR #3588, opened 2024-07-30
#3505 - Additional transportation workflow updates

dheepak-aot:
# Additional transportation allowance workflow updates

The story required update on the calculation of the additional transportation weekly allowance based on the updated inputs received from the `student application/student request a change` form inputs.

Update to the student inputs include addition of new inputs, removal of an existing input and also renaming the variable for an existing input. 

## Technical summary of new calculation.

The calculation of weekly additional transportation cost can be simplified in 5 steps as follows.

`STEP 1: DEFINE WEEKLY MINIMUM ` Define the minimum weekly transportation allowance that the student is entitled to receive based on the offering delivery type.

`STEP 2: DEFINE WEEKLY MAXIMUM ` Define the maximum weekly transportation allowance limit that the student is entitled to receive based on `if they are the owner of the vehicle`.

`STEP 3: CALCULATE WEEKLY COST BASED ON STUDENT KM INPUT `
 weekly additional transport requested by student = `additional transport km * km reduction`

`STEP 4:  LIMIT TO MAXIMUM WEEKLY ALLOWANCE` Make sure that the weekly additional transport requested by student (STEP 3) does not go beyond maximum weekly cost(STEP 2).

`STEP 5: APPLY PLACEMENT REDUCTION ` Apply placement reduction to the cost calculated in STEP 4 if required.

`STEP 6: ENSURE MINIMUM WEEKLY ALLOWANCE` Ensure that the weekly cost calculated in STEP 5 is not below the minimum weekly allowance. Once minimum weekly allowance is ensured, then it results in weekly net additional transportation allowance.

## Additional transportation allowance calculation update(BPMN and DMN)

- [x] Added a new DMN output in `dmnPartTimeProgramYearMaximums` for non-owner maximum weekly allowance.

image

- [x] Updated the calculations based on the new requirement and new inputs to calculate additional transportation allowance.

image

## Update the existing student/appeal variables

- [x] Updated the variables as per the requirement in student application and appeal flow.
image

image

## E2E Tests

- [x] Modified the existing E2E tests due to the update in the variables.

- [x] Created E2E Tests for the scenarios listed by business in the story.

image

image

PR #3597, opened 2024-07-31
#3539 - Award Configuration Round - PT

lewischen-aot:
- Updated bpmn `parttime-configure-disbursement` with new rounding logic in the configure payload. No further code change outside this workflow is expected.
  - Turned the calculation for disbursement schedules into two steps: `firstHalfDisbursementValues` and `secondHalfDisbursementValues`
  - While rounding off multiple disbursements, the first disbursement value is subtracted from round down value of actual award.
- Added E2E tests with at least one single disbursement scenario and given multiple disbursement scenarios.
```markdown
| Award | Half  | 1st Disb | 2nd Disb      |
|-------|-------|----------|---------------|
| 350   | 175   | 175      | 175 (350-175) |
| 375   | 187.5 | 188      | 187 (375-188) |
| 100.5 | 50.25 | 50       | 50 (100 -50)  |
| 100.9 | 50.45 | 50       | 50 (100 -50)  |
| 101.9 | 50.95 | 51       | 50 (101-51)   |
```

Screenshot of disbursement schedules and values for 18 weeks or more in Camunda 
image


sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
```diff
@@ -108,7 +108,9 @@
...
+          <zeebe:output source="={&#10;  CSLPAmount: round half up(finalFederalAwardNetCSLPAmount / 2, 0),&#10;  CSGPAmount: round half up(finalFederalAwardNetCSGPAmount / 2, 0),&#10;  CSGDAmount: round half up(finalFederalAwardNetCSGDAmount / 2, 0),&#10;  CSPTAmount: round half up(finalFederalAwardNetCSPTAmount / 2, 0),&#10;  BCAGAmount: round half up(finalProvincialAwardNetBCAGAmount / 2, 0),&#10;  SBSDAmount: round half up(finalProvincialAwardNetSBSDAmount / 2, 0)&#10;}" target="firstHalfDisbursementValues" />
```
andrewsignori-aot, line 111: Can the `firstHalfDisbursementValues` and `secondHalfdisbursementValues` use the rounding methods from the same algorithm, for instance, can we use `round half up`/`round half down` or `ceiling`/`floor`?

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
```diff
@@ -108,7 +108,9 @@
...
+          <zeebe:output source="={&#10;  CSLPAmount: ceiling(floor(finalFederalAwardNetCSLPAmount) - firstHalfDisbursementValues.CSLPAmount),&#10;  CSGPAmount: ceiling(floor(finalFederalAwardNetCSGPAmount) - firstHalfDisbursementValues.CSGPAmount),&#10;  CSGDAmount: ceiling(floor(finalFederalAwardNetCSGDAmount) - firstHalfDisbursementValues.CSGDAmount),&#10;  CSPTAmount: ceiling(floor(finalFederalAwardNetCSPTAmount) - firstHalfDisbursementValues.CSPTAmount),&#10;  BCAGAmount: ceiling(floor(finalProvincialAwardNetBCAGAmount) - firstHalfDisbursementValues.BCAGAmount),&#10;  SBSDAmount: ceiling(floor(finalProvincialAwardNetSBSDAmount) - firstHalfDisbursementValues.SBSDAmount)&#10;}" target="secondHalfdisbursementValues" />
```
andrewsignori-aot, line 112: Was there a reason to provide the second parameter as zero? It seems to be the default behavior already unless I am missing something.
```
round half up(finalFederalAwardNetCSLPAmount / 2, 0)
```

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot, line 111: Was there a reason to provide the second parameter as zero? It seems to be the default behavior already unless I am missing something.
```
round half up(finalFederalAwardNetCSLPAmount / 2, 0)
```

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot, line 112: I was not able to understand why the `ceiling` was used here. Can you please elaborate?

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot, line 111: Following the naming convention from the rest of the source code, the acronym in the beginning would be lowercase, as below.
```
{
  cslpAmount: round half up(finalFederalAwardNetCSLPAmount / 2, 0),
  csgpAmount: round half up(finalFederalAwardNetCSGPAmount / 2, 0),
  csgdAmount: round half up(finalFederalAwardNetCSGDAmount / 2, 0),
  csptAmount: round half up(finalFederalAwardNetCSPTAmount / 2, 0),
  bcagAmount: round half up(finalProvincialAwardNetBCAGAmount / 2, 0),
  sbsdAmount: round half up(finalProvincialAwardNetSBSDAmount / 2, 0)
}
```

sources/packages/backend/workflow/test/2022-2023/parttime-assessment/disbursements/parttime-assessment-disbursements.e2e-spec.ts
```diff
@@ -7,6 +7,35 @@ import { addDays, getISODateOnlyString, getUTCNow } from "@sims/utilities";
 import { createFakeConfigureDisbursementPartTimeData } from "../../../test-utils/factories";
 
 describe(`E2E Test Workflow parttime-assessment-${PROGRAM_YEAR}-disbursements`, () => {
+  // An array of award amounts to verify first and second disbursement value amounts.
+  const testFinalAwardNetAmounts = [
```
andrewsignori-aot, line 11: [minor] In similar work done for the eligibility tests I would say that there are the following differences:
- const was declared following the equivalent to `TEST_FINAL_AWARDNET_AMOUNTS`;
- There was a `describe` common description and the `it` was inside the for.

https://github.com/bcgov/SIMS/blob/363df0d21ea086e3329e0605c9b1aa2147305376/sources/packages/backend/workflow/test/2024-2025/parttime-assessment/eligibility/parttime-assessment-eligibility-BCAG.e2e-spec.ts#L11-L30

This is similar to a previous PR in the sprint. Feel free to take a look at the final approach.

https://github.com/bcgov/SIMS/pull/3588#discussion_r1697757745

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot, line 1: As part of this PR, an assessment with a single disbursement should also be rounded down. An E2E scenario should also be created for this scenario.

image

andrewsignori-aot (commented):

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
lewischen-aot: `ceiling` is not recommended for rounding the first disbursement value, and `round half up` is a better option. We can certainly use `round half down` instead of `floor` to "pair with" `round half up`, but `round half down` requires two parameters -- `round down(n: number, scale: number): number` and `floor` only needs one parameter. If we use `round half down`, then `floor(floor(finalFederalAwardNetCSLPAmount) - firstHalfDisbursementValues.cslpAmount)` will be `round half down(round half down(finalFederalAwardNetCSLPAmount, 0) - firstHalfDisbursementValues.cslpAmount, 0)`. The code is a little longer, but the result is the same. I am okay either way. Please let me know if we want to change `floor` to `round half down`.

PS: `ceiling` has been replaced with `floor` in the second commit as of the comment below https://github.com/bcgov/SIMS/pull/3597#discussion_r1699128373.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
lewischen-aot: `ceiling` has been replaced with `floor`.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
lewischen-aot: Updated.

sources/packages/backend/workflow/test/2022-2023/parttime-assessment/disbursements/parttime-assessment-disbursements.e2e-spec.ts
lewischen-aot: Updated.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
lewischen-aot: Added.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
lewischen-aot: According to the documentation [link](https://docs.camunda.io/docs/components/modeler/feel/builtin-functions/feel-built-in-functions-numeric/#round-half-upn-scale), 0 is used to round to the ones. The calculation will give `null` value for using `round half up` without the scale parameter: `round half up(finalFederalAwardNetCSLPAmount / 2)` --> `null`.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot: Was there some issue that required calling `floor` twice?
For instance, consider the below:
```
floor(floor(finalFederalAwardNetCSLPAmount) - firstHalfDisbursementValues.cslpAmount)
```
The `floor(finalFederalAwardNetCSLPAmount)` will result in an integer and the `firstHalfDisbursementValues.cslpAmount` was already a rounded number, was there a need for the external `floor` to be added?

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot: Sorry, I did not get why it would not be recommended for the "first disbursement value" and then why the `round` would be ok.
Just for clarification.

sources/packages/backend/workflow/test/2022-2023/parttime-assessment/disbursements/parttime-assessment-disbursements.e2e-spec.ts
```diff
@@ -246,6 +280,295 @@ describe(`E2E Test Workflow parttime-assessment-${PROGRAM_YEAR}-disbursements`,
     },
   );
 
+  describe("Should generate 1 disbursement when offering weeks is greater than 17 weeks and ", () => {
```
andrewsignori-aot, line 283: The 17-week logic is not the target of the test, right? We can just do the minimum to ensure one disbursement is generated and the numbers are properly rounded.
IMO the test description should be clear about the target of the test.
```
"Should round down the values for the disbursement when only one disbursement is expected and " +
    `the final award of each type is ${testFinalAwardNetAmount.finalAwardNetAmount}.` 
```

sources/packages/backend/workflow/test/2022-2023/parttime-assessment/disbursements/parttime-assessment-disbursements.e2e-spec.ts
```diff
@@ -246,6 +280,295 @@ describe(`E2E Test Workflow parttime-assessment-${PROGRAM_YEAR}-disbursements`,
...
+                valueType: "BC Grant",
+              },
+            ],
+          },
+        ]);
+      });
+    }
+  });
+
+  describe(
```
andrewsignori-aot, line 362: Is this test not similar to the previous one? Can you please clarify the different goals between both?
Just double-checking if the first one would not be enough to cover the modification.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot: As explained by @lewischen-aot the `ceiling` would be rounding up values between 0 and 0.5 which is not desired and that is why the `round half up` was the selected one.
The floor is also applicable for the other operation since it should act as a truncate.

sources/packages/backend/workflow/src/workflow-definitions/parttime-configure-disbursement.bpmn
andrewsignori-aot: As explained by @lewischen-aot, the second `floor` is to ensure that the value resulting from the operation would not be anyway having some decimals and will be guaranteed to be saved as an integer in the DB.

andrewsignori-aot (approved):

Issue #2970, opened 2024-03-08
Analysis - SIMS Object Storage for Files 


michesmith:
As an SIMS product team, we need to store files in an object store rather than the persistent volume claims. 

**Acceptance Criteria:** 

- [x] Find a Nodejs library to be used.
- [x] Create the env variable for the S3 connection and operations.
- [x] Keep saving the file bytes to SIMS DB as it is right now but start uploading the file to the S3 storage in parallel. If the S3 uploads fail it should not prevent the API from saving the DB. Log the S3 operation time.
- [x] Try to download the file bytes from the S3 storage, if not present or if an error happens retrieve the bytes from SIMS DB. Log the S3 operation time.

**Technical Context**
Buckets and policies were created for sims-local, sims-dev, sims-test, sims-staging, and sims-prod. Out-of-box server-side managed bucket encryption enabled.

Connection information available from Stephen or Jason

michesmith:
@sslaws this was a note from our previous tech analyst. are you able to discuss at Dev Lather?

sslaws:
@michesmith yes I can discuss.

sslaws:
Dev to dress and include the following: secure retrieval of files from object storage, migration of existing files and updates to application/db to store links to object store.

andrewsignori-aot:
As discussed with @ninosamson moving it to Biz Lather for Ministry agreement and decision on how to proceed.

dheepak-aot:
As per retro discussion moving the ticket to IMB Architecture pipeline.

ninosamson:
re: questions from me: 
- I assumed object / file are same level? 
- if we're trying to clarify whether or not files with PI should be saved, then yes, they should.

andrewsignori-aot:
> re: questions from me:
> 
> * I assumed object / file are same level?
> * if we're trying to clarify whether or not files with PI should be saved, then yes, they should.

@ninosamson can we assume after the last chat with @sslaws and yourself that we are ok to have the encryption at rest as a solution and questions about files saved with potential PI data are ok?

ninosamson:
Confirmed @andrewsignori-aot. The solution will meet the requirements.

andrewsignori-aot:
## Tests on Openshift (~4MB file)

[...]

PR #3604, opened 2024-08-02
#2970 - SIMS Object Storage for Files (POC)

andrewsignori-aot:
### S3 Library
- Used the largely used [aws-sdk/client-s3](https://www.npmjs.com/package/@aws-sdk/client-s3) library, the same used by the Common [Object Management Service (COMS)](https://github.com/bcgov/common-object-management-service/blob/master/app/src/services/storage.js).
- Only the methods required for upload/download have been implemented.
- The implemented methods have well-typed exceptions.
- The SDK connection is stateless and no need to establish or keep an open connection. The same client object is created globally and reused for all the requests.
- Without defining a `region` the library throws an error. Even though not important for the implementation an ENV variable was added and it should be set to a value like `canada-bc`.
- Currently saving the files associated with its `content-types` which means the columns `sims.student_files.file_content` and  `sims.student_files.mime_type` can be removed once SIMS API stops saving the file content to DB.

### Deployment
- Created the new ENV variables and associated them with the API POD deployment.

### Temporary code for monitoring
- Changed the API upload and download area to parallel upload and download files from the S3 and logged the time for the operation to happen.


.github/workflows/env-setup-deploy-secrets.yml
```diff
@@ -49,6 +49,11 @@ jobs:
       CAS_BASE_URL: ${{ secrets.CAS_BASE_URL }}
       CAS_CLIENT_ID: ${{ secrets.CAS_CLIENT_ID }}
       CAS_CLIENT_SECRET: ${{ secrets.CAS_CLIENT_SECRET }}
+      S3_ACCESS_KEY_ID: ${{ secrets.CAS_CLIENT_SECRET }}
+      S3_SECRET_ACCESS_KEY: ${{ secrets.S3_SECRET_ACCESS_KEY }}
```
andrepestana-aot, line 53: they are using CAS values.

.github/workflows/env-setup-deploy-secrets.yml
```diff
@@ -49,6 +49,11 @@ jobs:
       CAS_BASE_URL: ${{ secrets.CAS_BASE_URL }}
       CAS_CLIENT_ID: ${{ secrets.CAS_CLIENT_ID }}
       CAS_CLIENT_SECRET: ${{ secrets.CAS_CLIENT_SECRET }}
+      S3_ACCESS_KEY_ID: ${{ secrets.CAS_CLIENT_SECRET }}
```
andrepestana-aot, line 52: This one is using CAS value

andrewsignori-aot: Fixed!

sources/packages/backend/libs/integrations/src/object-storage/models/object-storage.models.ts
```diff
@@ -0,0 +1,11 @@
...
+  key: string;
+  body: Buffer;
+  contentType: string;
+}
+
+export interface GetObjectResult {
+  contentLength: number;
+  contentType: string;
+  body: NodeJS.ReadableStream;
+}
```
dheepak-aot, line 11: What about following DTO convention here?

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -129,14 +132,39 @@ export class StudentControllerService {
...
+      // Populate fil information from DB.
+      console.info("Retrieving from DB.");
+      response.setHeader("Content-Type", studentFile.mimeType);
+      response.setHeader("Content-Length", studentFile.fileContent.length);
+      const stream = new Readable();
+      stream.push(studentFile.fileContent);
+      stream.push(null);
+      stream.pipe(response);
+    } finally {
+      console.timeEnd(stopwatchLabel);
```
dheepak-aot, line 166: For S3 operation time why didn't we just cover the line 141 alone??

sources/packages/backend/libs/integrations/src/object-storage/models/object-storage.models.ts
andrewsignori-aot: I do not see a reason to be honest. Even if it is not wrong to call it a DTO we usually do not use it from service to service models, at least that is my perception.
Even for services like BCeID I believe that we use models instead DTO conventions.

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
andrewsignori-aot: The idea was just to try to get the difference between the overall operation. The same file can be uploaded twice and one can be deleted from the storage which would allow both times to be tested.
In the end, I believe that either way has the potential to give us the data we need.

dheepak-aot (approved):
Great work. Good to see the object storage finally happening. Also thanks for updating the makefile. 👍

andrepestana-aot:
Looks good! 👍

Issue #3456, opened 2024-06-24
Content: Student: Account creation declarations


JerPearson:
As a student I want to complete a declaration when creating my account authorizing third party verification.

**Acceptance Criteria**
- [ ] Update basic BCeID path "Create your profile" declaration with the following: 

"I understand that clicking ‘agree’ means consenting to the following:

**Collection, use and disclosure of personal information**

(1)	The information included in this form is collected under ss. 26(c) and 26(e) of the Freedom of Information and Protection of Privacy Act, R.S.B.C. 1996, c. 165. The information you provide will be used in confirming your identity. If you have any questions about the collection and use of this information, contact the Executive Director, StudentAid BC, Ministry of Advanced Education and Skills Training, PO Box 9173, Stn Prov Govt, Victoria BC, V8W 9H7, telephone 1-800-561-1818 (toll-free in Canada/U.S.) or +1-778-309-4621 from outside North America. This consent takes effect when I click ‘I agree’
 
(2)	I consent to the disclosure of my Social Insurance Number, name, date of birth and gender, for verification against information contained in Employment and Social Development Canada Social Insurance Registry. This information will be disclosed to Employment and Social Development Canada for the purpose of confirming the accuracy of my identification as an applicant to student financial assistance. This consent takes effect when I click ‘I agree’.

**Updates to personal information**  

(3)	I acknowledge that I am responsible for ensuring that my personal profile information is kept up to date.

(4)	I will update my personal profile information where there is a change to my personal information that I can update myself, such as an address change.

(5)	I will request a change from Employment and Social Development Canada, ServiceBC or ICBC   , as applicable, when there is a change to my personal information that I cannot update myself within the StudentAid BC portal , such as a name change.

(6)	I acknowledge that I may be asked to provide documentation to the Ministry of Post-Secondary Education and Future Skills  its contractors, or authorized third party administrators, to support changes to information contained within my personal profile, such as a name change."

<img width="1003" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0dWQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2365f4c9ffbd4a7891c613624ca5c660a98a1687/image.png" alt="image.png" />

- [ ] Update BCSC account creation declaration with the following: 

"I understand that clicking ‘agree’ means consenting to the following:

**Collection, use and disclosure of personal information**

[...]

PR #3605, opened 2024-08-02
#3456 - Content: Student: Account creation declarations

andrepestana-aot:
image
image
image


andrewsignori-aot (approved):
Took me a while to understand why item (5) was not easily checked against the ticket content due to the extra white space before the last comma. Nice work, looks good 👍

Issue #3399, opened 2024-06-10
 Content: Student: Request a Change: Options


JerPearson:
As a student I want to see what options are available through the change request drop-down menu.

## Acceptance Criteria

- Revise the wording on the current drop-down menu

 - [ ] Revise: "Add/Remove Dependants" to "Add/remove dependants"
 - [ ] Revise: "Disability Information" to "Disability information"
 - [ ] Revise: "Financial Information" to "Financial information"

Ticket to remove 'Income' and "Partners Income" from the drop-down #3271[](https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/gh/bcgov/sims/3127) 

<img width="948" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMkZvQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--126659d68331e990fa4f8f96980e2dd617727d20/image.png" alt="image.png" />

PR #3608, opened 2024-08-06
#3399 - Content: Student: Request a Change: Options

andrepestana-aot:
- Case  change for request a change options

image


andrewsignori-aot (approved):
👍

Issue #3486, opened 2024-07-04
Content fixes for previous issues


astridSABC:
Various fixes from other ZH issues

**Acceptance Criteria**

On _Apply for disability status popup information_ (Go to Profile>Apply for disability status)

- [x] Remove "a" and revise "d" and "s" to lower case "disability status" (sentence case) 

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDIyQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--877bc50b10e1e6a6b41b90fb25748a1d5462cfdc/image.png" alt="image.png" />

Ref #3303

On the Additional transportation 

- [x] Add **bold** style to the three occurrences of "and" 
copy of your vehicle **and**
....letter of explanation **and**
....to and from school or daycare **and**

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3kxQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b611c5251a97309dc7794a144af79fd55a76bbe8/image.png" alt="image.png" />

- [x] Add punctuation "." at the end of the "I am in a clinical or practicum placement that requires travel" sentence

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUMyQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--83f32a27ef33308e100fa12b1ae72c612a5c8038/image.png" alt="image.png" />

- [x] Add question mark to the end of the "Do you want to be assessed for additional transportation costs" question

<img width="897" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMU1aQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6a89be88e55c59895edde9c908a6927c91a534e5/image.png" alt="image.png" />

Ref #3311

On the FAQ third option on the Welcome to StudentAid BC page on the "I am not eligible for a BC Services Card account. What can I do?" in the _What you need to do_ section 

[...]

PR #3612, opened 2024-08-06
#3486 - Content fixes for previous issues

andrepestana-aot:
- Removed "a" and revised "d" and "s" to lower case "disability status" (sentence case)
image

- Added bold style to the three occurrences of "and"
image

- Added punctuation "." at the end of the "I am in a clinical or practicum placement that requires travel" sentence
image

- Added question mark to the end of the "Do you want to be assessed for additional transportation costs" question
image

- Made email address a link "[StudentAidBC@gov.bc.ca](mailto:StudentAidBC@gov.bc.ca)"
image

- Fixed style for the tooltip that was cut off by the container
image

- Revised "Metis" to "Métis"
image

- Removed the "." from "Apply for student financial assistance."
image

### DIFF
--- a/sources/packages/web/src/components/layouts/student/StudentPageContainer.vue
+++ b/sources/packages/web/src/components/layouts/student/StudentPageContainer.vue
@@ -2,7 +2,7 @@
   <full-page-container
     :layout-template="layoutTemplate"
     :full-width="fullWidth"
-    class="overflow-visible"
+    v-bind="$attrs"
   >
     <template #header>
       <slot name="header"></slot>
--- a/sources/packages/web/src/components/layouts/student/StudentPageContainer.vue
+++ b/sources/packages/web/src/components/layouts/student/StudentPageContainer.vue
@@ -2,6 +2,7 @@
   <full-page-container
     :layout-template="layoutTemplate"
     :full-width="fullWidth"
+    class="overflow-visible"
   >
     <template #header>
       <slot name="header"></slot>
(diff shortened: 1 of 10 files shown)


sources/packages/web/src/components/layouts/student/StudentPageContainer.vue
```diff
@@ -2,6 +2,7 @@
   <full-page-container
     :layout-template="layoutTemplate"
     :full-width="fullWidth"
+    class="overflow-visible"
```
andrewsignori-aot, line 5: Assuming this will be available for all student pages using the `student-page-container`, should we remove it from the `StudentAppealRequest.vue`?
I guess only pages that needd it would receive the `class` since it overrides the default vue component behavior.
Would be be better to have ot applied only to the page with the issue?

andrewsignori-aot (commented):
Nice work, overall looks good, just a minor comment.

sources/packages/web/src/components/layouts/student/StudentPageContainer.vue
andrepestana-aot: yeah, I removed it from `StudentPageContainer` and I'm setting the class in `FullTimeApplication`. i just needed to add `v-bind="$attrs"` because `FullTimeApplication` is dealing with multiple root nodes: https://vuejs.org/guide/components/attrs#attribute-inheritance-on-multiple-root-nodes

andrepestana-aot: Reverted the v-bind as not needed. You were right. Thanks!

andrewsignori-aot (approved):

PR #3613, opened 2024-08-07
#2857 - Remove formio long variables

dheepak-aot:
# RENAME OR REMOVE FORM IO LONG AND OBSOLETE VARIABLES

- [x] Removed/updated the long variable names and obsolete variables in student application forms (`SFAA*`) and supporting users forms(`supportingusersparent*` and `supportinguserspartner*`)
- [x] Created a spreadsheet which contains all the removed/updated variables and their impact on workflow and data inventory reports. Attached here.

image

[Long_Variable_names.xlsx](https://github.com/user-attachments/files/16531153/Long_Variable_names.xlsx)
- [x] Adjusted the workflow variables which were affected by the changes.
- [x] Renamed variable `federalAwardBCAGAmount` to `provincialAwardBCAGAmount` in all `parttime-assessment-*.bpmn`.
- [x] Adjusted workflow E2E tests on variables which affected workflow.
- [x] Smoke tested with FT and PT application submissions and also validated long variables.


sources/packages/forms/src/form-definitions/supportinguserspartner2024-2025.json
```diff
@@ -4140,7 +4154,7 @@
                     "multiple": false,
                     "unique": false
                   },
-                  "key": "iAgreeToTheTermsAndUseAsNoted",
+                  "key": "iAgreeToTheAboveCraConsent",
```
andrewsignori-aot, line 4157: Would it be `iAgreeToTheAboveCRAConsent`?

andrewsignori-aot (commented):
Great cleanup and renaming. I left just one minor comment but I have a question about the "Panel" variable in the ticket description: "totalUnsubsidizedDayCareCostsDuringClassHoursForTheProposedStudyPeriodForDisabledInfirmedDependents12YearsOfAgeAndOlderPanel".

I believe the above is the variable that triggered the ticket creation but I do not believe its association in the report (only place being used) is correct, meaning that even if it is too long it should never be saved to DB.

Similar to this variable we have some others. I see two possible actions on this.
1 - We rename the panel variables, or
2 - We leave the panel variable and we mention in the ticket that the refactor is only for key peristed to DB.
Thoughts?

sources/packages/forms/src/form-definitions/sfaa2024-25.json
andrewsignori-aot, line 1: The [latest changes from `sfaa*` forms](https://github.com/bcgov/SIMS/pull/3612/files) are not in the PR. Should we expect the merge to take care of it or would it be better to do it in the PR?

dheepak-aot:
> Great cleanup and renaming. I left just one minor comment but I have a question about the "Panel" variable in the ticket description: "totalUnsubsidizedDayCareCostsDuringClassHoursForTheProposedStudyPeriodForDisabledInfirmedDependents12YearsOfAgeAndOlderPanel".
> 
> I believe the above is the variable that triggered the ticket creation but I do not believe its association in the report (only place being used) is correct, meaning that even if it is too long it should never be saved to DB.
> 
> Similar to this variable we have some others. I see two possible actions on this. 1 - We rename the panel variables, or 2 - We leave the panel variable and we mention in the ticket that the refactor is only for key peristed to DB. Thoughts?



> Great cleanup and renaming. I left just one minor comment but I have a question about the "Panel" variable in the ticket description: "totalUnsubsidizedDayCareCostsDuringClassHoursForTheProposedStudyPeriodForDisabledInfirmedDependents12YearsOfAgeAndOlderPanel".
> 
> I believe the above is the variable that triggered the ticket creation but I do not believe its association in the report (only place being used) is correct, meaning that even if it is too long it should never be saved to DB.
> 
> Similar to this variable we have some others. I see two possible actions on this. 1 - We rename the panel variables, or 2 - We leave the panel variable and we mention in the ticket that the refactor is only for key peristed to DB. Thoughts?

As discussed, we can add this item for panel key to be replaced with the right dynamic data key in data inventory report. As a part of this ticket, I am collecting the impacts of variable changes to data inventory report which will be recorded in an appropriate ticket.

andrewsignori-aot (approved):
Thanks again for the cleanups, looks good 👍

Issue #3223, opened 2024-05-06
Create history/journal tables  for auditing purposes


JasonCTang:
***User Story***
As a Privacy and Security Officer (David Malcolm), I need the ability to determine when student profile records were updated, what was updated and who updated them.

***Additional Information***
Tables to create journal table for: 
- users
- students
- supporting_users
- institutions
- institution_locations
- education_programs
- education_programs_offerings

When to create journal entries
- Insert a new record in the original table and then insert a clone of the record in the corresponding journal table
- Update the update_at, modifier & modified fields in the original table, and then insert a clone of the record in the corresponding journal table.

***Acceptance Criteria***
- [x] Create journal tables for the tables listed above
- [x] Implement triggers and stored procedures to insert records into journal tables.

**Technical**
- [x] Create the new triggers on DB for `Insert` and `Update` operations.
- [x] Create new tables appending a suffix like `_history`.
- [x] Duplicate the table structure by adding the below columns.
  - [x] Operation type: `Insert`, `Update`.

**Some References**
https://www.postgresql.org/docs/current/sql-createtrigger.html

sslaws:
Jason to determine what tables we need to journal.

sslaws:
Removing high priority as its pushed to 2.1

andrewsignori-aot:
@JasonCTang @ninosamson we may need to review the apps and also the understanding of the creator/modifier, how they work right now, and the expectation.
Many times the creator will be the "system user", for instance, while processing Camunda workflows. Usually, the DEVs are just taking a best guess.
It was also introduced a long time and many places may not be saving it.
Just to be clear, journal table records are created, creator/modifier should be there most of the time but it is not guaranteed right now.

andrewsignori-aot:
## Demo

Tested on DEV the below scanrions to produce inserts and updates in the required tables.

- Created a new basic BCeID users and created the account  after Ministry approval.
- Ministry creates new institution.
- Institution creates and edit locations.
- Institution creates and edit programs.
- Institution creates and edit offerings.
- Student create an application with supporting users.

Please see below some screenshots with records inserted to the history tables.

![image](https://github.com/user-attachments/assets/b1e999a1-4cb7-480f-9fe5-8941b65681f4)

![image](https://github.com/user-attachments/assets/68fcf254-a292-432e-8841-9e5248a74193)

![image](https://github.com/user-attachments/assets/3bf42e78-7edf-4b26-990a-51f61aa92f95)

[...]

PR #3618, opened 2024-08-08
#3223 - Create history/journal tables for auditing purposes (users and students)

andrewsignori-aot:
## Proposed Solution
- Created a generic function `create_history_entry` to be associated with every table that would need a history.
- Used the tables `sims.users` and `sims.students` as a POC with the expected structure for any history table.
  - Tables were appended with `_history`.
  - New column `history_timestamp` to save the time when the entry was created.
  - New column `history_operation` to save the operation as `INSERT` or `UPDATE`.
  - Index created on `history_timestamp`.
  - `NOT NULL` constraints and relationships were dropped to keep the table as plain as possible.
  - _Migration scripts for the new history table were placed inside the same folder as the target table. When columns need to be added or dropped, the idea would be to use the same SQL script file to manage both tables to keep it simple._

## Sample `sims.user` table where the POC was applied
image

### Pros

- Generic table eliminates the need for the creation of individual functions to deal with different table structures.
- Creation of new columns would not demand changes in the trigger function.
- The tables could be individually configured for the triggers, for instance, a table like `education_programs_offerings` can be configured to generate history only for the `UPDATE` for performance considerations during bulk inserts.

### Cons

- When adding new columns we should be mindful about adding the columns to both tables at the same time to ensure the proper insert order (it may be in the "pros" list 😄 ).
- If new columns need to be added to the `_history` table we must ensure they will be introduced at the beginning of the table which would demand recreation of the `_history` table since there is no mechanism to add a table in the specific order.

### Migrations rollback evidence

image

image

image

### Smoke test validations
- Created a new basic BCeID and login as a student with Ministry approval and checked the history created for `sims.users` and `sims.students` tables.
- Create a new column in `sims.students` and `sims.students_history` and the history continues to be generated as expected.
image

### Todo

- Implement audit history for remaining tables.


ninosamson:
Looks good to me @andrewsignori-aot. Just confirming - we capture the associated user GUID, correct? 
If so, then this is perfect - we can track user updates to the associated data in the new history tables.

andrewsignori-aot:
> Looks good to me @andrewsignori-aot. Just confirming - we capture the associated user GUID, correct? If so, then this is perfect - we can track user updates to the associated data in the new history tables.

@ninosamson I added the Student's table in this PR. The tables part of this requirement have the `creator` and `modifier` columns that are the id from the `sims.users` where the user GUID exists.

sources/packages/backend/apps/db-migrations/src/sql/Student/Create-student-history-table.sql
```diff
@@ -0,0 +1,76 @@
+-- Table to keep historical data.
+-- Columns history_timestamp and history_operation should be default columns for all the history tables.
+-- Columns other than the above-mentioned should reflect the columns from the original table without constraints and relationships.
+-- The order of the columns below follow the exact same order from the existing table.
+CREATE TABLE sims.students_history (
```
dheepak-aot, line 5: Not a blocker. IMO having a primary key for the history table e.g `history_id` will be better unless it is going to hurt the common insert query.

sources/packages/backend/apps/db-migrations/src/sql/Student/Create-student-history-table.sql
```diff
@@ -0,0 +1,76 @@
...
+  history_timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
+  history_operation VARCHAR(20) NOT NULL,
+  id INT,
+  contact_info JSONB,
+  user_id INT,
+  created_at TIMESTAMP WITHOUT TIME ZONE,
+  updated_at TIMESTAMP WITHOUT TIME ZONE,
```
dheepak-aot, line 12: Created_at and updated_at are with timezone right?

sources/packages/backend/apps/db-migrations/src/sql/Student/Create-student-history-table.sql
andrewsignori-aot: I thought about it and decided not to add a primary key or a serial ID because I did not see a benefit.
Some considerations:
- A `history_id` would not be used for relationships.
- A `history_id` as a primary key would also generate an index, which I do not see the benefit.
- This table will never be updated, only insert operations would happen.
- The above-mentioned reasons would add an unnecessary overhead to the history.

Please let me know if the arguments are not accurate or if I am missing something 😉

sources/packages/backend/apps/db-migrations/src/sql/Student/Create-student-history-table.sql
andrewsignori-aot: The schemas for the history tables reflect the current schema from the target table. If we change the target table schema we can then change the history table schema. Does it make sense?

sources/packages/backend/apps/db-migrations/src/sql/User/Create-user-history-table.sql
```diff
@@ -0,0 +1,61 @@
...
+  user_name VARCHAR(300),
+  email VARCHAR(300),
+  first_name VARCHAR(100),
+  last_name VARCHAR(100),
+  created_at TIMESTAMP WITH TIME ZONE,
+  updated_at TIMESTAMP WITH TIME ZONE,
+  is_active BOOLEAN,
+  creator INT,
+  modifier INT,
+  identity_provider_type sims."identity_provider_types"
```
dheepak-aot, line 18: Minor. For enums in other sqls we do not use the double quotes. Is there a particular reason here?

sources/packages/backend/apps/db-migrations/src/sql/User/Create-user-history-table.sql
```diff
@@ -0,0 +1,61 @@
...
+
+COMMENT ON COLUMN sims.users_history.history_timestamp IS 'Date and time the history record was inserted.';
+
+COMMENT ON COLUMN sims.users_history.history_operation IS 'Database operation that generates the history.';
+
+COMMENT ON COLUMN sims.users_history.id IS 'Historical data from the original table. See original table comments for details.';
+
+COMMENT ON COLUMN sims.users_history.user_name IS 'Historical data from the original table. See original table comments for details.';
+
+COMMENT ON COLUMN sims.users_history.email IS 'Historical data from the original table. See original table comments for details.';
```
dheepak-aot, line 35: Should we ignore adding comments for original table copy columns instead of using dummy comment ?

dheepak-aot:
Great work!!. Just a few minor comments.

sources/packages/backend/apps/db-migrations/src/sql/User/Create-user-history-table.sql
andrewsignori-aot: No reason, it was generated from DBeaver and I forgot to change it. I will adjust it.

sources/packages/backend/apps/db-migrations/src/sql/User/Create-user-history-table.sql
andrewsignori-aot: I see the point, I had the same doubt and ended up adding the comments in a way that they would point to the real business comments to keep the idea of always adding a comment to the DB objects.
I would be ok either way in this situation but I preferred the approach in place right now.

sources/packages/backend/apps/db-migrations/src/sql/User/Create-user-history-table.sql
andrewsignori-aot: Changed.

sources/packages/backend/apps/db-migrations/src/sql/Student/Create-student-history-table.sql
andrewsignori-aot: You were right, I got it now. I will update it.

andrewsignori-aot: Fixed!

lewischen-aot (approved):
LGTM

PR #3631, opened 2024-08-14
#3223 - Create history/journal tables for auditing purposes (revert)

andrewsignori-aot:
Revert history tables creation to allow its recreation.
After the deployment to DEV the offerings history failed due to differences in the column's order that could potentially happen in the future. To resolve it a new history table creation SQL will be used to ensure the history tables will follow the same order from the env it is deployed.
The decision to revert these commits was to allow a clean history of the history table creation.


dheepak-aot (approved):
Looks Good 👍

andrewsignori-aot:
All DB migrations manually reverted on DEV.
image

PR #3627, opened 2024-08-12
#3223 - Create history/journal tables for auditing purposes (remaining tables)

andrewsignori-aot:
## Applied the same solution from #3618 on below tables.

- supporting_users
- institutions
- institution_locations
- education_programs
- education_programs_offerings

#### Minor Audit-Related Fix

- Supporting users were not saving the audit user (system user) during the record creation.
- While providing supporting user data, the user was not saved when the authenticated user was new, which means the user token ID was never populated leaving the `modifier` blank.

### Migrations rollback evidence

image

image

image

image

image

PR #3632, opened 2024-08-14
#3223 - Create history/journal tables for auditing purposes

andrewsignori-aot:
This PR is the result a cherry-pick from reverted PRs #3618 and #3627. These PRs were reverted by #3631 to allow the history tables to be created following exactly the same column orders from the DB it is being deployed.
The only real change not reviewed previously is the below. All other files were kept as they were. The description below is also a combination from the previous reverted PRs (the rollback evidences are new).
```sql
CREATE TABLE sims.supporting_users_history AS
SELECT
  CURRENT_TIMESTAMP AS history_timestamp,
  '' :: varchar(50) AS history_operation,
  *
FROM
  sims.supporting_users
WHERE
  false;
```

## Proposed Solution
- Created a generic function `create_history_entry` to be associated with every table that would need a history.
- Created the history tables for the required tables.
  - Tables were appended with `_history`.
  - New column `history_timestamp` to save the time when the entry was created.
  - New column `history_operation` to save the operation as `INSERT` or `UPDATE`.
  - Index created on `history_timestamp`.
  - `NOT NULL` constraints and relationships were dropped to keep the table as plain as possible.
  - _Migration scripts for the new history table were placed inside the same folder as the target table. When columns need to be added or dropped, the idea would be to use the same SQL script file to manage both tables to keep it simple._

## Sample `sims.user` table
image

### Pros

- Generic table eliminates the need for the creation of individual functions to deal with different table structures.
- Creation of new columns would not demand changes in the trigger function.
- The tables could be individually configured for the triggers, for instance, a table like `education_programs_offerings` can be configured to generate history only for the `UPDATE` for performance considerations during bulk inserts.

### Cons

- When adding new columns we should be mindful about adding the columns to both tables at the same time to ensure the proper insert order (it may be in the "pros" list 😄 ).
- If new columns need to be added to the `_history` table we must ensure they will be introduced at the beginning of the table which would demand recreation of the `_history` table since there is no mechanism to add a table in the specific order.

#### Minor Audit-Related Fix

- Supporting users were not saving the audit user (system user) during the record creation.
- While providing supporting user data, the user was not saved when the authenticated user was new, which means the user token ID was never populated leaving the `modifier` blank.

### Migrations rollback evidence

image

image

image

image

image

image

image

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
@@ -237,7 +237,7 @@ export class SupportingUserSupportingUsersController extends BaseController {
       const updatedUser = await this.supportingUserService.updateSupportingUser(
         application.id,
         supportingUserType,
-        userToken.userId,
+        user.id,
         {
           contactInfo,
           sin: submissionResult.data.data.sin,
--- a/sources/packages/backend/apps/db-migrations/src/migrations/1723231015871-CreateHistoryEntryFunction.ts
+++ b/sources/packages/backend/apps/db-migrations/src/migrations/1723231015871-CreateHistoryEntryFunction.ts
@@ -1,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../utilities/sqlLoader";
+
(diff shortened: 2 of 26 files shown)


dheepak-aot (approved):
Thanks for taking the effort to investigate and provide a better solution. 👍Looks good.

guru-aot (approved):
LGTM, i like this implementation. Nice work @andrewsignori-aot

PR #3635, opened 2024-08-15
#3268 - DB values for different Namespaces

guru-aot:
- Created different values.yaml for individual namespaces in openshift, similar to clamav configurations.
- Changed the makefile to accomodate the individual yaml changes with respect to namespaces.
- Added one more non-super-user in the crunchy postgres DB, to use it for DB migrations.


andrewsignori-aot (approved):
Great work, looks good 👍

devops/helm/crunchy-postgres/value-0c27fb-prod.yaml
```diff
@@ -0,0 +1,102 @@
+fullnameOverride: simsdb
```
dheepak-aot, line 1: Is it necessary to have the namespace e.g `a6ef19` in it?

guru-aot: we did the similar exercise for clamav, to have the namespaces values defined for a6ef19

PR #3637, opened 2024-08-15
3268 - DB Connection Parameter change to crunchy

guru-aot:
- Applications DB connection parameters are changed from patroni to crunchy. 
- DB_SERVICE_NAME is moved as part of crunchy secret, so removed it from makefile parameters.


devops/openshift/api-deploy.yml
```diff
@@ -323,13 +326,13 @@ parameters:
   - name: PORT
     required: true
   - name: DB_SERVICE
-    value: patroni-master
+    value: pgbouncer-host
```
dheepak-aot, line 329: I see. This is now an auto generated host name coming from secrets. 
image

devops/openshift/api-deploy.yml
```diff
@@ -58,6 +56,11 @@ objects:
                   value: "${MAXIMUM_IDLE_TIME_FOR_WARNING_INSTITUTION}"
                 - name: MAXIMUM_IDLE_TIME_FOR_WARNING_AEST
                   value: "${MAXIMUM_IDLE_TIME_FOR_WARNING_AEST}"
+                - name: POSTGRES_HOST
+                  valueFrom:
+                    secretKeyRef:
+                      key: ${DB_SERVICE}
```
dheepak-aot, line 62: As per our naming convention of secret keys, this should be `DB_SERVICE_KEY`

devops/openshift/api-deploy.yml
guru-aot: yes

devops/openshift/api-deploy.yml
```diff
@@ -322,14 +325,14 @@ parameters:
...
-  - name: DB_SERVICE
-    value: patroni-master
+  - name: DB_SERVICE_KEY
+    value: pgbouncer-host
   - name: DB_SECRET_NAME
-    value: patroni-creds
+    value: simsdb-pguser-app-database-user
   - name: DB_USERNAME_KEY
-    value: database-user
+    value: user
```
dheepak-aot, line 333: The API and DB Migrations are are using the same user here. I believe the idea for apps(api, workers etc..) is to use non privileged user right?? 

Or is this part going to be handled later?

dheepak-aot:
Good work @guru-aot . Just an outstanding question.

devops/openshift/api-deploy.yml
guru-aot: We have different user, if you check the SECRET_NAME_KEY both are pointing to 2 different secrets, but the DB_USERNAME_KEY looks the same 
image
image

dheepak-aot (approved):
Thanks for the clarification. Looks good.

PR #3643, opened 2024-08-16
#3268- Crunchy DB PVC size change

guru-aot:
- Crunchy PVC DB size changed to 5Gi as initial size.
- -f values are removed in the makefile, as it was going always to the default values.yml and it was not needed. I found while testing so changed it now.

### DIFF
--- a/devops/helm/crunchy-postgres/Makefile
+++ b/devops/helm/crunchy-postgres/Makefile
@@ -18,7 +18,7 @@ helm-dep:
 .PHONY: install
 install: helm-dep
 install:
-	@helm install $(call arguments) --set restore.enabled=false -f values.yaml
+	@helm install $(call arguments) --set restore.enabled=false;
 
 .PHONY: upgrade
 upgrade: helm-dep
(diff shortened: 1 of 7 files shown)


devops/helm/crunchy-postgres/Makefile
```diff
@@ -18,7 +18,7 @@ helm-dep:
 .PHONY: install
 install: helm-dep
 install:
-	@helm install $(call arguments) --set restore.enabled=false -f values.yaml
```
andrewsignori-aot, line 21: Can you clarify why `-f values.yaml` is changed?
The PR description mentioned the size change and did not explain this additional change.

guru-aot: Updated the PR description.

devops/helm/crunchy-postgres/value-0c27fb-prod.yaml
```diff
@@ -24,7 +24,7 @@ instances:
   name: ha # high availability
   replicas: 3
   dataVolumeClaimSpec:
-    storage: 21Gi
+    storage: 5Gi
```
dheepak-aot, line 27: PVC for prod is kept as same as non prod environments. I believe we have a plan to update it before we move production.

andrewsignori-aot (approved):
Thanks for providing the clarification, looks good 👍

devops/helm/crunchy-postgres/value-0c27fb-prod.yaml
guru-aot: Yes

dheepak-aot (approved):
Looks good. Just an observation to confirm.

Issue #3193, opened 2024-04-25
New OCP Routes


sslaws:
As a Product Owner, as we get closer to the launch of sims we need to have URLs that better reflect out branding.

**Acceptance Criteria**
- Existing routes should be maintained and these new routes will be applied.
- Update the routes for dev, test, staging and prod web-sims
    - [x] Dev = https://dev.sims.studentaidbc.ca
    - [x] Test = https://test.sims.studentaidbc.ca
    - [x] Staging = https://staging.sims.studentaidbc.ca
    - [x] Prod = https://sims.studentaidbc.ca
    - [x] Forms = https://*-forms.sims.studentaidbc.ca (not publicly exposed). One per environment. 
- [x] Keycloak clients will need to be updated with the web origins and redirect URLs above per corresponding KC env 
- [x] All routes need the SIMS cert installed
-  For each environment using the web-sims base-url add the following routes
     - [x] Path /api for api-sims
     - [x] Path /swagger for api-sims (one is already in place for Swagger and should be removed before the new deployment happens).
     - [x] Path /admin/queues for queue-consumers 
- [x] All routes should be configured as insecure edge termination redirect
- All routes should have the following annotations applied:
    - [x] haproxy.router.openshift.io/balance: leastconn
    - [x] haproxy.router.openshift.io/disable_cookies: 'true'
    - [x] haproxy.router.openshift.io/hsts_header: max-age=31536000;includeSubDomains;preload
    - [x] haproxy.router.openshift.io/timeout: ??s
        - [x] Replace ?? with the longest query timeout expected - default is 30 seconds
- [x] Create Github secrets **at the repository** level for the new route properties: `certificate`, `key`, and `caCertificate`.
  - [x] Adjust the deployment configs to have the new properties configured. Use this [route](https://console.apps.silver.devops.gov.bc.ca/k8s/ns/0c27fb-dev/routes/tls-dev-edwapi/yaml) to retrieve the values for the properties to be applied to all environments.
  - [x] Update the wiki for the KC changes
  - [x] Ensure release instructions are updated to capture all of the changes
  - _Please note, that these properties are sensitive and should not be exposed in any way._
          
**Other Context**
Sample working routes exist using the cert in dev are called tls-dev-edwapi and tls-dev-swagger
EDW Extract will be /edwapi and is excluded from this ticket
Other environments in dev (performance, load etc) can be migrated to new URLs or left as is.

ninosamson:
[...]

PR #3657, opened 2024-08-26
#3193 - New OCP Routes - Form.io Server

andrewsignori-aot:
- Added missing map for TLS variable for form.io deployment and DOMAIN_PREFIX.
- Forced the forms URL to be e.g. `dev-forms` instead `dev.forms`.
- Fix make file missing "\\".


dheepak-aot (approved):
Looks good 👍

Issue #2801, opened 2024-02-06
Archive SFTP files


JasonCTang:
***Describe the task***
Archive files on the SFTP server.

- When reading files uploaded by external systems, archive the file using the following logic. (Instead of deleting which is what happens now)
     - If file is proccessed succesffully, move them to an ./Archive/ subfolder which will be in the same folder as the file being processsed.  
     - If a non-sequential file fails processing, move them to an ./Archive/ subfolder which will be in the same folder as the file being processed.  
     - If a sequential file fails processing, do not archive.

- These change applies to all files being **read**:
     - Institution files
     - MSFAA files
     - Federal files
     - Bridge files
     - And others, if any.

***Acceptance Criteria***
- [x] Investigate and identify which files MUST be processed sequentially and which ones do not.
- [x] Whenever a file (sequential or non-sequeuntial) is processed successfully, move the file into the ./Archive/ subfolder located in the same folder as the file being processed.
- [ ] Whenever a non-sequential file fails processing, move file into the ./Archive/ subfolder located in the same folder as the file being processed.
- [x] Whenever a sequential file fails processing, do not move file. 
- [x] Provide a list of sequentially executed ones, non-sequentially executed ones, and delete behavior (e.g. deleted on error or not) defined during development.
- [x] For all archived files, append a timestamp to the name of the archived file (example: CONR_008.TXT will be archived as CONR_008_20240823_164947.TXT. 
- [x] E2E tests
  - [x] Adjust the E2E tests checking for the delete operation to start checking for the "move" operation.

JasonCTang:
***Describe the task***
Instead of deleting response files on the SFTP server after they're read into SIMS, copy them to an ./Archive/ subfolder.  The location of this archive subfolder will be in the same folder as the file being read.  This change applies to all files being read:
- Institution response files
- Federal response files
- Bridge files

***Acceptance Crteria***
- [ ] Whenever a file is successfully read into SIMS, move the file into the ./Archive/ subfolder located in the same folder as the file being read.

ninosamson:
Keeping ticket open to clarify/discuss wiki and sequential processing of files.

ninosamson:
Let's close this ticket and open another ticket to track monitoring, errors and operational processes with regards to files.

andrepestana-aot:
**Clarifying the sequential processing of files**
[...]

PR #3664, opened 2024-08-28
#2801 - Archive SFTP files - Archiving non sequential in case of error

andrepestana-aot:
Removed condition to not archive in case of error for:
 - full-time-feedback-integration
 - part-time-feedback-integration


andrepestana-aot:
No need for this one

Issue #3435, opened 2024-06-18
Dashboard Announcement - Institution and Student 


ninosamson:
**User Story**
As an Institution or Student user, I would like to be informed of any system changes or upcoming outages. 

**Acceptance Criteria**
- [x] Add two  alerts to 
     - [ ] dashboard page for Institution and 
     - [ ] dashboard page for Student 
- [x] Info style alert with icon 
    - [ ] System Announcements 
    - [ ] Title - 100 char max, plain text 
    - [ ] Text - 500 char max, plain text 
    - [ ] Announcements can be scheduled - start and end date 
    - [ ] "target" DB column will specify which application page. e.g. "Student Dashboard; Institution Dashboard" 
- [x] updates to banners are direct through database 

Display similar to this:
<img width="934" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd21QQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1c26961aca7d160c0d148a439c38656aec7d9a2c/image.png" alt="image.png" />

**Technical**
- [x] Create a new table to store the messages. The suggested table structure is below.
  - MessageTitle: text
  - Message: text
  - StartDate: date and time
  - EndDate: date and time
  - Target: array of strings (e.g. Student dashboard, Institution dashboard).
- [x] Create a Vue component.
  - Property to receive the array of strings (targets)  to be retrieved.
- [x] Create a shared controller between clients, similar to `DynamicFormController`.
- [x] E2E tests for the new API endpoints.

PR #3678, opened 2024-09-05
#3435 - System announcements

katrina-aot-git:
Add a database table for system announcements, and a way to display them on student and institution dashboards.
Screenshot 2024-09-05 at 4 15 53 PM
image


sources/packages/backend/apps/db-migrations/src/sql/Announcements/Rollback-create-announcements-table.sql
sh16011993, line 1: Can we please add a screenshot of the rollback to the PR description.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS sims.announcements (
+    id SERIAL PRIMARY KEY,
+    message_title VARCHAR(100) NOT NULL,
```
sh16011993, line 3: Have we confirmed on the length of this to be 100?

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
...
+    message VARCHAR(200) NOT NULL,
+    start_date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+    end_date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+    target VARCHAR(200) NOT NULL,
```
sh16011993, line 7: Isn't this supposed to be an array type?

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
...
+);
+
+-- ## Comments
+COMMENT ON TABLE sims.announcements IS 'The table for storing system announcements.';
+
+COMMENT ON COLUMN sims.announcements.message_title IS 'Title for the announcement.';
+
+COMMENT ON COLUMN sims.announcements.message IS 'Message text for the announcement.';
+
+COMMENT ON COLUMN sims.announcements.start_date IS 'When the message starts.';
```
sh16011993, line 26: Suggestion: `Timestamp when the system announcement starts.`

Likewise for the `end_date`.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Rollback-create-announcements-table.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS sims.announcements;
```
sh16011993, line 1: I think we are not using `IF EXISTS` anymore.

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
```diff
@@ -0,0 +1,54 @@
...
+  target: string;
+  /**
+   * Date when the announcement starts.
+   */
+  @Column({
+    name: "start_date",
+    type: "date",
+    nullable: false,
+  })
+  startDate: string;
```
sh16011993, line 44: Shouldn't the type be `timestampz`?

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
```diff
@@ -0,0 +1,54 @@
...
+    name: "target",
+    nullable: false,
+  })
+  target: string;
+  /**
+   * Date when the announcement starts.
+   */
+  @Column({
+    name: "start_date",
+    type: "date",
```
sh16011993, line 41: Shouldn't the type be `timestampz`?

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+  message: string;
+  /**
+   * Target areas for the announcement.
+   */
+  @Column({
+    name: "target",
+    nullable: false,
+  })
+  target: string;
```
sh16011993, line 35: I think this should be array type.

sources/packages/backend/apps/api/src/auth/roles.enum.ts
```diff
@@ -8,6 +8,7 @@ export enum Role {
   AESTManualTriggerReassessment = "aest-manual-trigger-reassessment",
   AESTReports = "aest-reports",
   AESTCreateInstitution = "aest-create-institution",
+  AESTEditCASSupplierInfo = "aest-edit-cas-supplier-info",
```
sh16011993, line 11: Are we using this?

sources/packages/backend/apps/api/src/app.module.ts
```diff
@@ -60,12 +65,18 @@ import { QueueModule } from "@sims/services/queue";
...
+    ConfigController,
+    DynamicFormController,
+    AnnouncementsController,
+  ],
   providers: [
     AppService,
     UserService,
     BCeIDServiceProvider,
     FormService,
+    AnnouncementsService,
```
sh16011993, line 79: I have a doubt here. Doesn't adding the service in this module resolve it for the other modules (aest / institution / student)?

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
```diff
@@ -0,0 +1,20 @@
...
+@Controller("announcements")
+@ApiTags("announcements")
+export class AnnouncementsController extends BaseController {
+  constructor(private readonly announcementsService: AnnouncementsService) {
+    super();
+  }
+
+  @Public()
+  @Get()
+  async getAnnouncements(): Promise<Announcements[]> {
```
sh16011993, line 17: Should we not return `AnnouncementsAPIOutDTO[]`?

sources/packages/backend/apps/api/src/auth/roles.enum.ts
katrina-aot-git: I just added this as when I merged in from main the tests I hadn't touched started failing due to this being missing.

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    return this.repo.find({
+      select: {
+        id: true,
+        messageTitle: true,
+        message: true,
+        target: true,
+        startDate: true,
+        endDate: true,
+      },
+      order: { startDate: "ASC" },
```
sh16011993, line 25: Are we not filtering the announcements based on the start and end dates here?

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  let app: INestApplication;
+  let announcementsRepo: Repository<Announcements>;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    announcementsRepo = dataSource.getRepository(Announcements);
+  });
+
+  it("Should return a saved announcement.", async () => {
```
sh16011993, line 17: I believe the tests can be adjusted to check for the specific start dates (announcements returned only after a specific start date).

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Rollback-create-announcements-table.sql
dheepak-aot: Yes we have stopped using `IF EXISTS` and `IF NOT EXISTS`

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
dheepak-aot: Yes that was mentioned in technical. 
image

sources/packages/web/src/services/http/AnnouncementsApi.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Http API client for system announcements.
+ */
+export class AnnouncementsApi extends HttpBaseClient {
+  /**
+   * Get the list of announcements.
+   * @returns list of announcements.
+   */
+  async getAnnouncements(): Promise<AnnouncementsAPIOutDTO[]> {
+    return this.getCall<AnnouncementsAPIOutDTO[]>(`announcements`);
```
sh16011993, line 13: We can use regular quotes `""` here.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS sims.announcements (
+    id SERIAL PRIMARY KEY,
+    message_title VARCHAR(100) NOT NULL,
+    message VARCHAR(200) NOT NULL,
+    start_date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
+    end_date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW (),
```
dheepak-aot, line 6: There is no requirement to have a default value here.

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,61 @@
...
+  props: {
+    location: {
+      type: String,
+      required: true,
+    },
+  },
+  setup(props) {
+    const { dateOnlyLongString } = useFormatters();
+    // Data-bind
+    const relevantAnnouncements = ref();
```
sh16011993, line 44: Can be initialized as `ref([] as AnnouncementsAPIOutDTO[])`

sources/packages/backend/apps/api/src/auth/roles.enum.ts
dheepak-aot: Please remove anything which is not in the scope of the PR. If the tests fail due to AEST role, it should be fine.

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
dheepak-aot: Yes and more importantly please ensure that entity models are not directly exposed to APIs. 

Here as per what we follow, create a *APIOutDTO  as a class (not an interface) to define the API.

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
```diff
@@ -0,0 +1,20 @@
+import { Controller, Get } from "@nestjs/common";
+import BaseController from "../BaseController";
+import { ApiTags } from "@nestjs/swagger";
+import { Announcements } from "@sims/sims-db";
+import { AnnouncementsService } from "../../services";
+import { Public } from "../../auth/decorators/public.decorator";
+
+@Controller("announcements")
```
dheepak-aot, line 8: we have always followed the singular can convention to name the controller root. In this case, `announcement`

sources/packages/web/src/components/common/AnnouncementBanner.vue
sh16011993, line 44: We can initialize as `ref([] as AnnouncementsAPIOutDTO[])`.

sources/packages/web/src/services/http/dto/Announcements.dto.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Announcements details.
+ */
+export interface AnnouncementsAPIOutDTO {
+  message_title: string;
```
sh16011993, line 5: `messageTitle`? Likewise for the start and end dates.

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <v-alert
+    :type="BannerTypes.Warning"
+    variant="outlined"
+    icon="fa:fa fa-triangle-exclamation"
+    class="sims-banner"
+  >
+    <template #title>
+      <div class="mb-5">System Announcements</div>
```
sh16011993, line 9: We need to add the class for the color.

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,61 @@
...
+    </template>
+    <v-row>
+      <v-col>
+        <div
+          v-for="(announcement, index) in relevantAnnouncements"
+          :key="index"
+        >
+          <div class="label-value-normal">
+            <b>{{ announcement.messageTitle }}</b> -
```
sh16011993, line 18: We can use the class `font-bold` here instead of `<b>`.

sh16011993 (commented):
Left some comments. Please have a look.

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
dheepak-aot: Also, same applies for the file name. `assnouncement.controller.ts`

sources/packages/backend/apps/api/src/route-controllers/announcements/models/announcements.dto.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Announcements detail DTO. This is used for view only purpose.
+ */
+export class AnnouncementsAPIOutDTO {
```
dheepak-aot, line 4: Singular for class name and file name.

sources/packages/backend/apps/api/src/route-controllers/announcements/models/announcements.dto.ts
```diff
@@ -0,0 +1,10 @@
...
+  message_title: string;
+  message: string;
+  start_date: Date;
+  end_date: Date;
+  target: string;
```
dheepak-aot, line 9: As mentioned by @sh16011993 please use camel case for properties e.g `startDate`

sources/packages/backend/apps/db-migrations/src/migrations/1724973005363-AddAnnouncements.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class AddAnnouncements1724973005363 implements MigrationInterface {
```
dheepak-aot, line 4: CreateAnnouncements ?

sources/packages/backend/apps/api/src/app.module.ts
dheepak-aot: Adding service here doesn't resolve the dependency in other modules, but the `AnnouncementsService` is not required to be added in other modules as it is required by the controller which is added only to app module.

sources/packages/backend/apps/api/src/app.aest.module.ts
```diff
@@ -31,6 +31,7 @@ import {
   ApplicationOfferingChangeRequestService,
   StudentLoanBalanceService,
   ProgramYearService,
+  AnnouncementsService,
```
dheepak-aot, line 34: https://github.com/bcgov/SIMS/pull/3678/files#r1747611024

Please remove adding the provider here.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS sims.announcements (
```
dheepak-aot, line 1: please remove IF NOT EXISTS

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
dheepak-aot: Please check with business to check on acceptable message title length.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
...
+    -- Audit columns
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    creator INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
+    SET
+        NULL,
+        modifier INT NULL DEFAULT NULL REFERENCES sims.users(id) ON DELETE
+    SET
+        NULL
+);
```
dheepak-aot, line 17: Audit columns are not required to this table as it is not going to be updated through the system.

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
andrewsignori-aot: There was a decision to avoid returning arrays in the DTO. The DTO must have an array property.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
...
+
+COMMENT ON COLUMN sims.announcements.message_title IS 'Title for the announcement.';
+
+COMMENT ON COLUMN sims.announcements.message IS 'Message text for the announcement.';
+
+COMMENT ON COLUMN sims.announcements.start_date IS 'When the message starts.';
+
+COMMENT ON COLUMN sims.announcements.end_date IS 'When the message finishes.';
+
+COMMENT ON COLUMN sims.announcements.target IS 'Represents the screens the message should appear on, in a CSV list.';
```
dheepak-aot, line 30: CSV list was not the dev agreement unless I am missing something.

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
```diff
@@ -0,0 +1,20 @@
...
+@ApiTags("announcements")
+export class AnnouncementsController extends BaseController {
+  constructor(private readonly announcementsService: AnnouncementsService) {
+    super();
+  }
+
+  @Public()
+  @Get()
+  async getAnnouncements(): Promise<Announcements[]> {
+    return this.announcementsService.getAnnouncements();
```
dheepak-aot, line 18: Should the API return all the system announcements for all the clients ? 

I meant, when this is called from student web, it will still return the announcements that are intended for institutions as well in the API. 

Are we considering API call using a student token receiving institution announcements to be safe?

Thoughts @andrewsignori-aot @andrepestana-aot @guru-aot @sh16011993 @bidyashish ?

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Public } from "../../auth/decorators/public.decorator";
+
+@Controller("announcements")
+@ApiTags("announcements")
+export class AnnouncementsController extends BaseController {
+  constructor(private readonly announcementsService: AnnouncementsService) {
+    super();
+  }
+
+  @Public()
```
dheepak-aot, line 15: I believe security-wise it must follow the same level as dynamic form controller. 

i.e. `@AllowAuthorizedParty` must be used, not the public.

image

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
dheepak-aot, line 17: please add comments to the method.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Rollback-create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: The message is defined to be a max length of 200. The title was not defined but should reasonably be shorter than the message itself. I've been told in past that business should not be bothered with exact db type/implementation details.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Rollback-create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
katrina-aot-git: updated

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
andrewsignori-aot: First, the AC states below where the API must receive the target to allow filtering the messages. So returning all messages was not the expectation.
```
Create a Vue component.
- Property to receive the array of strings (targets) to be retrieved.
```

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
andrewsignori-aot: Second, we may have wrongly planned the controller to be generic. It must be client-specific to allow only the specific targets to be provided for the specific clients.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
```diff
@@ -0,0 +1,33 @@
...
+      where: {
+        startDate: LessThanOrEqual(now.toDateString()),
```
dheepak-aot, line 28: The toDateString() converts to user friendly date format which I doubt if typeorm can process accurately. 

image

dheepak-aot: Please use toISOString()

image

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  async getAnnouncements(): Promise<Announcements[]> {
+    const now = new Date();
+    return this.repo.find({
+      select: {
+        id: true,
+        messageTitle: true,
+        message: true,
+        target: true,
+        startDate: true,
+        endDate: true,
```
dheepak-aot, line 24: As we are selecting all the column, we can ignore the select like this. Please double-check.

image

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <v-alert
```
dheepak-aot, line 2: We already have a banner component, `Banner.vue`. 

Is there a specific reason that we are not re-using it?

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,62 @@
...
+import { defineComponent, onMounted, ref } from "vue";
+import { BannerTypes } from "@/types/contracts/Banner";
+import { AnnouncementsService } from "@/services/AnnouncementsService";
+import { useFormatters } from "@/composables";
+
+export default defineComponent({
+  props: {
+    location: {
+      type: String,
+      required: true,
```
dheepak-aot, line 39: the prop name location is conflicting with the institution location that we use for institutions, can we name that as target ?

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,62 @@
...
+    const { dateOnlyLongString } = useFormatters();
+    // Data-bind
+    const relevantAnnouncements = ref();
+    // Hooks
+    onMounted(async () => {
+      let announcements = await AnnouncementsService.shared.getAnnouncements();
+      relevantAnnouncements.value = announcements.filter((announcement) => {
+        return announcement.target.some(
+          (tar: string) => tar.toLowerCase() === props.location.toLowerCase(),
+        );
```
dheepak-aot, line 52: IMO instead of going to API every time for the component, it would be better to use this from a store. 

Thoughts @andrewsignori-aot @andrepestana-aot @guru-aot @sh16011993 @lewischen-aot @bidyashish ?

sources/packages/web/src/views/student/StudentDashboard.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <student-page-container layout-template="centered">
+    <announcement-banner location="StudentDashboard" />
```
dheepak-aot, line 3: same here. under the slot alerts.

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    // Arrange
+    const now = new Date();
+    const before = new Date(now.setFullYear(now.getFullYear() - 1));
+    const future = new Date(now.setFullYear(now.getFullYear() + 2));
```
dheepak-aot, line 21: We can use the addDays() utility which we already have in place to add or remove days. 
image

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const announcement = new Announcements();
```
dheepak-aot, line 23: It would be good to have at-least 2 scenarios, announcements returned as current date is within the start date end date range and the opposite. I.e something to be started in future or something which ended already.

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    const response = await request(app.getHttpServer())
+      .get(endpoint)
+      .expect(HttpStatus.OK);
+
+    expect(
+      response.body.some(
+        (announcement) =>
+          announcement.message === "test announcement" &&
+          announcement.messageTitle === "test title",
+      ),
```
dheepak-aot, line 42: We must assert almost all the known properties.

dheepak-aot:
Nice work. Please take a look at the comments. 

The most important part of API is that, It must not return the messages that are not intended for it's own client type.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: If audit columns are removed, it can not be saved with the .save function in e2e tests. I originally did not have them.

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: Yes, the generic banner component did not follow the requested display.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
katrina-aot-git: updated

sources/packages/web/src/services/http/AnnouncementsApi.ts
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/migrations/1724973005363-AddAnnouncements.ts
katrina-aot-git: 'Add' seemed to be the more common name in previous migrations.

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
katrina-aot-git: removed

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
dheepak-aot: Start and end date are still having default values. 
image

sources/packages/backend/apps/db-migrations/src/migrations/1724973005363-AddAnnouncements.ts
dheepak-aot: It is not a blocker. Resolving.

andrepestana-aot: "Create" is good for me.

andrewsignori-aot: I see "Create" being used for table creation and "Add" being used also many times for other additions.
Please do not extend this conversation. It seems that "Create" would be the best fit but it would not be a blocker IMO.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
dheepak-aot: When audit columns, apparently we should not be extending the `RecordDataModel` as the `RecordDataModel` and it's child class `BaseModel` are class representing audit columns. 
image

Removing the extends will let us to remove the audit columns from table.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
dheepak-aot: I could not see the CSV list description removed from the comment. 
image

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
andrewsignori-aot: For the message title, I do not see a huge risk, we can increase it to 250 if we want.
The message being 200 seems to be problematic. We can go for a higher number like 2500.

katrina-aot-git: The ticket requested 200 for the message. "Text - 200 char max, plain text".

sources/packages/web/src/components/common/AnnouncementBanner.vue
andrewsignori-aot: I would not see it as a requirement for this ticket right now. I would suggest doing it when and if we expand the announcements to many pages.

sources/packages/web/src/components/common/AnnouncementBanner.vue
andrewsignori-aot: The component in place (`Banner.vue`) should be used.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE sims.announcements (
+    id SERIAL PRIMARY KEY,
+    message_title VARCHAR(100) NOT NULL,
+    message VARCHAR(200) NOT NULL,
+    start_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    end_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    target text [],
```
sh16011993, line 7: I missed this in the first round of review. I believe that this column is a `NOT NULL` column. Either way, this and the model file do not match.

image

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  @Column({
+    name: "message",
+    nullable: false,
+  })
+  message: string;
+  /**
+   * Target areas for the announcement.
+   */
+
```
sh16011993, line 31: Can we please remove the extra line here.

sources/packages/backend/apps/api/src/app.module.ts
sh16011993: Thanks @dheepak-aot . Got it. 👍

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
```diff
@@ -13,16 +13,13 @@ export class AnnouncementsService extends RecordDataModelService<Announcements>
...
-      select: {
-        id: true,
-        messageTitle: true,
-        message: true,
-        target: true,
-        startDate: true,
-        endDate: true,
-      },
       order: { startDate: "ASC" },
+      where: {
```
sh16011993, line 19: Minor: We follow the following order: `select` , `where` , `order`. Please put `where` before `order`.

sh16011993 (commented):
Nice work. Added few more comments. Please have a look.

sources/packages/backend/apps/api/src/auth/roles.enum.ts
katrina-aot-git: removed

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: moved the defaults over to the created and updated as I'd accidentally removed the wrong ones the first time

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
katrina-aot-git: file names updated

sources/packages/backend/apps/api/src/route-controllers/announcements/models/announcements.dto.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/models/announcements.dto.ts
katrina-aot-git: updated

sources/packages/web/src/services/http/dto/Announcements.dto.ts
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated the comment too

sources/packages/backend/libs/sims-db/src/entities/announcements.model.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcements.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/services/announcements/announcements.service.ts
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  let app: INestApplication;
+  let announcementsRepo: Repository<Announcement>;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    announcementsRepo = dataSource.getRepository(Announcement);
+  });
+
+  it("Should return a saved announcement.", async () => {
```
sh16011993, line 18: Suggestion: "Should return a saved currently active announcement."

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
andrewsignori-aot: Converting to an array property in a DTO would not be a blocker since we have too many others following the same.

sources/packages/web/src/components/common/AnnouncementBanner.vue
andrewsignori-aot: @katrina-aot-git To be clear, we can use the `<strong>{{ announcement.messageTitle }}</strong>` instead of the `<b>`.
We have a simple CSS that will be applied to `strong`.
https://github.com/bcgov/SIMS/blob/f1bb49e2c447803197e9c517f0a7599075d40993/sources/packages/web/src/assets/css/base.scss#L14-L21

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: reworked

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: reworked

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: reworked

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: reworked

sources/packages/web/src/views/institution/InstitutionDashboard.vue
```diff
@@ -4,6 +4,16 @@
     layout-template="centered"
     data-cy="institutionWelcomePage"
   >
+    <announcement-banner location="InstitutionDashboard" />
```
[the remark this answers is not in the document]
katrina-aot-git: reworked

sources/packages/web/src/views/student/StudentDashboard.vue
katrina-aot-git: reworked

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: reworded

sources/packages/backend/apps/db-migrations/src/migrations/1724973005363-AddAnnouncements.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/app.aest.module.ts
katrina-aot-git: removed

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: removed

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
katrina-aot-git: Split into student and institution controllers

sources/packages/backend/apps/api/src/route-controllers/announcements/announcements.controller.ts
katrina-aot-git: Split into student and institution controllers

sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * Get system announcements.
+   * @param target location for the announcement.
+   * @returns system announcements list.
+   */
+  @Get()
+  async getAnnouncements(
+    @Query() queryString: InstitutionAnnouncementsAPIInDTO,
+  ): Promise<Announcement[]> {
+    return this.announcementService.getAnnouncements(queryString.target);
```
dheepak-aot, line 28: It is still returning the entity object on API. It must return DTO. I see the OUTDTO being created but not used here. Here is how we can do it.

```ts
  @Get()
  async getAnnouncements(
    @Query() queryString: StudentAnnouncementsAPIInDTO,
  ): Promise<AnnouncementAPIOutDTO[]> {
    const announcements = await this.announcementService.getAnnouncements(
      queryString.target,
    );
    return announcements.map((announcement) => ({
      messageTitle: announcement.messageTitle,
      message: announcement.message,
      startDate: announcement.startDate,
      endDate: announcement.endDate,
      target: announcement.target,
    }));
  }
}
``

Comment (dheepak-aot) [reply sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts:28]: image

File: sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.students.controller.ts [hunk 44]
```diff
@@ -0,0 +1,34 @@
...
+    super();
+  }
+
+  /**
+   * Get system announcements.
+   * @param target location for the announcement.
+   * @returns system announcements list.
+   */
+  @Get()
+  async getAnnouncements(
```
dheepak-aot, line 29: Same here. https://github.com/bcgov/SIMS/pull/3678/files#r1757621010

sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  /**
+   * Get system announcements.
+   * @param target location for the announcement.
+   * @returns system announcements list.
+   */
+  @Get()
+  async getAnnouncements(
+    @Query() queryString: InstitutionAnnouncementsAPIInDTO,
```
dheepak-aot, line 26: the name queryString is very generic, we can call it as systemAnnouncementOptions or something very specific to endpoint.

sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+@Controller("announcements")
+@ApiTags(`${ClientTypeBaseRoute.Institution}-announcements`)
+export class AnnouncementInstitutionsController extends BaseController {
+  constructor(private readonly announcementService: AnnouncementService) {
+    super();
+  }
+
+  /**
+   * Get system announcements.
+   * @param target location for the announcement.
```
dheepak-aot, line 21: Please fix the comment here param name and description must be updated to reflect code change.

sources/packages/backend/apps/api/src/route-controllers/announcements/models/announcement.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Announcement detail DTO. This is used for view only purpose.
+ */
+export class AnnouncementAPIOutDTO {
+  messageTitle: string;
+  message: string;
+  startDate: Date;
+  endDate: Date;
+  target: string[];
+}
+
```
dheepak-aot, line 13: As mentioned above, please create the wrapper API to wrap the array result.

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE sims.announcements (
+    id SERIAL PRIMARY KEY,
+    message_title VARCHAR(100) NOT NULL,
+    message VARCHAR(200) NOT NULL,
+    start_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    end_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    target text [] NOT NULL
```
dheepak-aot, line 7: For arrays we have been using VARCHAR() []

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,21 @@
...
+);
+
+-- ## Comments
+COMMENT ON TABLE sims.announcements IS 'The table for storing system announcements.';
+
```
dheepak-aot, line 12: comment for column id is missing.

sources/packages/backend/libs/sims-db/src/entities/announcement.model.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  @Column({
+    name: "message",
+    nullable: false,
+  })
+  message: string;
+  /**
+   * Target areas for the announcement.
+   */
+  @Column("text", {
```
dheepak-aot, line 29: Please use the column decorator the way it is used in other places. 
image

sources/packages/web/src/components/common/AnnouncementBanner.vue
sh16011993: I cannot see this change.

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <banner
+    header="System Announcements"
+    :type="BannerTypes.Warning"
+    v-if="relevantAnnouncements && relevantAnnouncements.length > 0"
```
dheepak-aot, line 5: The v-if could be simplified as 

`v-if="relevantAnnouncements.length"`

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE sims.announcements (
+    id SERIAL PRIMARY KEY,
+    message_title VARCHAR(100) NOT NULL,
+    message VARCHAR(200) NOT NULL,
```
andrewsignori-aot, line 4: @katrina-aot-git I have just talked with Carly and I suggested increasing this to 500.
The ticket is updated.
I also updated the ticket to settle on the 100 for the `message_title`.

sources/packages/web/src/components/common/AnnouncementBanner.vue
```diff
@@ -0,0 +1,46 @@
...
+  >
+    <template #content>
+      <div v-for="(announcement, index) in relevantAnnouncements" :key="index">
+        <b>{{ announcement.messageTitle }}</b> -
```
dheepak-aot, line 9: Minor. For bold fonts, we were using different font-family once, but now we at least have a distinct variable for font-family when it comes to bold fonts. 

image


I would suggest one of the 2 things. 


1. add the class `font-bold` to the element. 
2. write a generic css

image

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+describe("AnnouncementInstitutionsController(e2e)-getAnnouncements", () => {
+  let app: INestApplication;
+  let announcementsRepo: Repository<Announcement>;
+  const endpoint = `/institutions/announcements?target=institution-dashboard`;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    announcementsRepo = dataSource.getRepository(Announcement);
```
dheepak-aot, line 21: We have an object called `E2EDataSources` and `createE2EDataSources()` and we can add the announcement repo their to have easy access for this and future tests. 

As a reference, we can follow this code [Other E2E Test example](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/route-controllers/report/_tests_/e2e/report.aest.controller.exportReport.e2e-spec.ts)

This comment is not a blocker. If you do not want to make the change, feel free to ignore.

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    const announcement = new Announcement();
+    announcement.message = "test announcement";
+    announcement.messageTitle = "test title";
+    announcement.target = ["student-dashboard", "institution-dashboard"];
+    announcement.startDate = before;
+    announcement.endDate = future;
+    await announcementsRepo.save(announcement);
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeFUser,
+    );
```
dheepak-aot, line 39: Please create a factory similar to `createFakeNote` to create fake DB objects.

image

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Act/Assert
+    const response = await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK);
+
+    expect(
+      response.body.some(
```
dheepak-aot, line 81: The assert could be simplified and used like this. await request(app.getHttpServer())
      .get(endpoint)
      .auth(institutionUserToken, BEARER_AUTH_TYPE)
      .expect(HttpStatus.OK)
      .expect([
        {
          messageTitle: announcement.messageTitle,
          message: announcement.message,
          startDate: announcement.startDate.toISOString(),
          endDate: announcement.endDate.toISOString(),
          target: announcement.target,
        },
      ]);

sources/packages/web/src/components/common/AnnouncementBanner.vue
sh16011993: Resolving this since there is another comment open.

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        (announcement) =>
+          announcement.message === "already done" &&
+          announcement.messageTitle === "already done" &&
+          announcement.startDate === before.toISOString() &&
+          announcement.endDate === before.toISOString(),
+      ),
+    ).toBe(false);
+  });
+
+  it("Should not return a future announcement.", async () => {
```
dheepak-aot, line 91: For this test, we can either clear the data for "student-dashboard" before or use a fake target to test the outdated message. 

```ts
  // Arrange
    const now = new Date();
    const before = addDays(-20, now);

    const announcement = new Announcement();
    announcement.message = "already done";
    announcement.messageTitle = "already done";
    announcement.target = ["outdated-dashboard"];
    announcement.startDate = before;
    announcement.endDate = before;
    await announcementsRepo.save(announcement);
    const institutionUserToken = await getInstitutionToken(
      InstitutionTokenTypes.CollegeFUser,
    );

    const endpointWithOutdatedDashboard = `/institutions/announcements?target=institution-dashboard`;

    // Act/Assert
    await request(app.getHttpServer())
      .get(endpointWithOutdatedDashboard)
      .auth(institutionUserToken, BEARER_AUTH_TYPE)
      .expect(HttpStatus.OK)
      .expect([]);
  });
```

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: added

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Act/Assert
+    const response = await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(institutionUserToken, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK);
+
+    expect(
+      response.body.some(
+        (announcement) =>
```
dheepak-aot, line 82: For this test, we can either clear the data for "student-dashboard" before the test using `beforeEach` or use a fake target to test the outdated message. For the second approach, here is an example.

```ts
  // Arrange
    const now = new Date();
    const before = addDays(-20, now);

    const announcement = new Announcement();
    announcement.message = "already done";
    announcement.messageTitle = "already done";
    announcement.target = ["outdated-dashboard"];
    announcement.startDate = before;
    announcement.endDate = before;
    await announcementsRepo.save(announcement);
    const institutionUserToken = await getInstitutionToken(
      InstitutionTokenTypes.CollegeFUser,
    );

    const endpointWithOutdatedDashboard = `/institutions/announcements?target=institution-dashboard`;

    // Act/Assert
    await request(app.getHttpServer())
      .get(endpointWithOutdatedDashboard)
      .auth(institutionUserToken, BEARER_AUTH_TYPE)
      .expect(HttpStatus.OK)
      .expect([]);
  });
```

sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/announcement.institutions.controller.ts
katrina-aot-git: updated

sources/packages/backend/libs/sims-db/src/entities/announcement.model.ts
katrina-aot-git: updated

sources/packages/backend/apps/db-migrations/src/sql/Announcements/Create-announcements-table.sql
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: There's nothing saying that that will be the only announcement returned, but I have updated the test to refer to the faked announcement's values directly.

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: added

sources/packages/backend/libs/test-utils/src/factories/announcement.ts
```diff
@@ -0,0 +1,22 @@
+import { Announcement } from "@sims/sims-db";
+import * as faker from "faker";
+
+export function createFakeAnnouncement(
+  startDate: Date,
+  endDate: Date,
+  messageTitle?: string,
+  message?: string,
+  target?: string[],
```
dheepak-aot, line 9: To pass the values of the factory object please use initial values like this. 

```ts

export function createFakeAnnouncement(options?: {
  initialValues?: Partial<Announcement>;
}): Announcement {
  const announcement = new Announcement();
  announcement.message =
    options?.initialValues?.message ?? faker.random.words(3);
  announcement.messageTitle =
    options?.initialValues?.messageTitle ?? faker.random.words(10);
  announcement.target = options?.initialValues?.target ?? [
    "student-dashboard",
    "institution-dashboard",
  ];
  announcement.startDate = options?.initialValues?.startDate;
  announcement.endDate = options?.initialValues?.endDate;

  return announcement;
}
```

sources/packages/backend/libs/test-utils/src/factories/announcement.ts
```diff
@@ -0,0 +1,22 @@
+import { Announcement } from "@sims/sims-db";
+import * as faker from "faker";
+
+export function createFakeAnnouncement(
```
dheepak-aot, line 4: Please add a comment to the method.

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
dheepak-aot: Generally when writing tests, we make sure that API or any portion of the test returns only the data that is created for that test. 

e.g. if we create 1 single announcement which is valid for current date in any test, we must expect that announcement to be returned, so that we do not have to check if that announcement is one one among many valid ones.

katrina-aot-git: Unless I am allowed to delete / modify other announcements that may already be in the db, I can not guarantee that there will not already be relevant announcements in the db. The get for announcements does not fetch _this specific announcement_, it fetches anything in the correct time range and dashboard.

sources/packages/backend/libs/test-utils/src/factories/announcement.ts
katrina-aot-git: updated

sources/packages/backend/libs/test-utils/src/factories/announcement.ts
katrina-aot-git: updated

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: I'd been told from previous tests that one should not modify other db data in the test.

sources/packages/web/src/components/common/AnnouncementBanner.vue
katrina-aot-git: changed to a span of font-bold as seen elsewhere

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
dheepak-aot: Whatever you mentioned above is absolutely right, but only if you repeat the target. 

If you use a distinct target for each test case, then the data what we expect can be guaranteed. 
For e.g. for the first test case, I can create a target called `fake-institution-target-1` and expect a guaranteed result like this.  Please keep in mind that, we are testing the API and target(s) doesn't need to be what we use in UI.

```ts

   const now = new Date();
    const before = addDays(-10, now);
    const future = addDays(30, now);

    const announcement = createFakeAnnouncement({
      initialValues: {
        startDate: before,
        endDate: future,
        target: ["fake-institution-target-1"],
      },
    });
    const endpoint = `/institutions/announcements?target=fake-institution-target-1`;
    await announcementsRepo.save(announcement);
    const institutionUserToken = await getInstitutionToken(
      InstitutionTokenTypes.CollegeFUser,
    );

    // Act/Assert
    await request(app.getHttpServer())
      .get(endpoint)
      .auth(institutionUserToken, BEARER_AUTH_TYPE)
      .expect(HttpStatus.OK)
      .expect([
        {
          messageTitle: announcement.messageTitle,
          message: announcement.message,
          startDate: announcement.startDate.toISOString(),
          endDate: announcement.endDate.toISOString(),
          target: announcement.target,
        },
      ]);

```

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
dheepak-aot: We could definitely reset the data before each test execution if needed and we have done that in many tests. But I would suggest the approach here which could be more simple. https://github.com/bcgov/SIMS/pull/3678#discussion_r1761409785

sources/packages/backend/apps/api/src/route-controllers/announcements/_tests_/e2e/announcement.institutions.controller.getAnnouncements.e2e-spec.ts
katrina-aot-git: I have specific targets bound to specific permissions - wouldn't this also require that to change?

`
export class StudentAnnouncementsAPIInDTO {
  @IsIn(["student-dashboard"])
  target: string;
}

export class InstitutionAnnouncementsAPIInDTO {
  @IsIn(["institution-dashboard"])
  target: string;
}
`

dheepak-aot:
Thanks for doing the changes @katrina-aot-git. I just have one outstanding comment on E2E tests. https://github.com/bcgov/SIMS/pull/3678#discussion_r1761409785

dheepak-aot (approved):

Issue #3112, opened 2024-04-10
Application - Accessibility/Usability - Submit Button Location


michesmith:
**User Story**
As a student, I need the submit button to be available at the end of the form flow so I can tab and submit. 

**Acceptance Criteria**
- [ ] Add the Save draft (if visible)/Submit application buttons to the bottom right of the "Confirm Submission" tab of the application.

<img width="1283" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2VJQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e99afc1fd5d19d69eab6f15444b039af49859c30/image.png" alt="image.png" />

**Context:**
- These issues were uncovered through accessibility audit via CITZ ( 1.3.2 Submit Button- Student- Accessibility Fixes - Level A - Meaningful Sequence ) & again on the usability testing
- Detailed Spreadsheet for Student s inserted inline and also attached here: [2024_Student Application_Audit.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0xsQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--578821d4addd12c1babcafba5c861e1f5456bc9c/2024_Student%20Application_Audit.xlsx)

CarlyCotton:
Updated ticket to make it more cohesive between the accessibility and useability feedback.

dheepak-aot:
Confirmed with @CarlyCotton that `Save Draft` button at bottom is only required on the last page, unlike the `Save Draft` button on the top.

PR #3684, opened 2024-09-11
#3112 - Accessibility submit button location

bidyashish:
**Acceptance Criteria**
- [X] Add the Save draft (if visible)/Submit application buttons to the bottom right of the "Confirm Submission" tab of the application.

image

### DIFF
--- a/sources/packages/web/src/components/common/StudentApplication.vue
+++ b/sources/packages/web/src/components/common/StudentApplication.vue
@@ -19,7 +19,7 @@
     secondaryLabel="Previous section"
     class="mx-0"
   >
-    <template #primary-buttons v-if="!readonly && isLastPage">
+    <template #primary-buttons v-if="!isReadOnly && isLastPage">
       <!-- This section is loaded only for editable forms -->
       <span>
         <v-btn


dheepak-aot:
Please adjust the PR title following the pattern from other PRs. #TICKET_NUMBER - DESC

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -19,6 +19,29 @@
     secondaryLabel="Previous section"
     class="mx-0"
   >
+    <template #primary-buttons v-if="isLastPage">
```
dheepak-aot, line 22: The save draft is currently being visible in the read-only view. The reason being `notDraft` variable is not set by the parent to the component and other 2 variables `isFirstPage` and `processing` is false.

image

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -19,6 +19,29 @@
...
+      <span>
```
dheepak-aot, line 23: To address the comment above and also not to have the span(line 23) created during a read-only load, 

following solution can be followed. 

Add the condition `!readonly` to line 22(template) and remove from line 34. The solution will look like this.

image

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -19,6 +19,29 @@
...
+        <v-btn
+          color="primary"
+          v-if="!notDraft && !isFirstPage && !processing"
```
dheepak-aot, line 26: When the condition gets complex, we can simplify using a computed property like this. 

image

image

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -19,6 +19,29 @@
...
+          variant="outlined"
+          :loading="savingDraft"
+          @click="handleSaveDraft()"
```
dheepak-aot, line 29: We can directly emit on click, instead of using a method. Same for line 38.

image

dheepak-aot:
Good Job @bidyashish. Please have a look at the comments.

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -19,6 +19,30 @@
     secondaryLabel="Previous section"
     class="mx-0"
   >
+    <template #primary-buttons v-if="!readonly && isLastPage">
```
andrepestana-aot, line 22: There's no `readonly`:
image

dheepak-aot: It seems to be a typo, the property is rather called, `isReadOnly`

andrepestana-aot (commented):
I left one comment.

dheepak-aot (approved):
Thanks for making the changes. Looks good.

andrepestana-aot (approved):
Good job!

Issue #3485, opened 2024-07-04
Student: New applications: Student can no longer immediately add trusted contact when "Yes" is selected similar to Add a dependant feature


astridSABC:
### **Related to #3468**

**Describe the Bug**
Create a new application to reproduce the bug, as the edit application does not reproduce it.
Student can no longer immediately add trusted contact when "Yes" is selected (similar to Add dependants feature)

**Expected Behaviour**
Upon selecting the "Yes" option, the student can immediately enter their trusted contact's first name, last name, select the appropriate radio button option and there is a clear button.  There is a button "Add another trusted contact" available to add more.

Note: Review the use of the asterisk.  Column names with asterisk is sufficient.
Nice to have enhancements: 
- [ ] Remove red * asterisk at the top of the table since all the other column names have *.  
- [ ] Remove red * in the editable fields in the First name and last name columns
- [ ] Remove red * that are beside all the radio buttons:

     - Parent/step-parent/sponsor/legal guardian
     - Spouse/common-law partner
    - Relative
    - Other

<img width="1139" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3UxQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--78ae53e4c9267137dd8d67aeaf1cb9da24f0d789/image.png" alt="image.png" />

**Actual Behaviour**
Windows Chrome: When the student selects the "Yes" radio button, a table is displayed and they can enter information in the editable fields and no delete button.  When they select "Add another trusted contact" the information in the top row disappears. Or they select "No" and then "Yes" again, the correct layout of the table is displayed.

Note: Sometimes the same table that displays on Windows Edge (no fields) will display on Windows Chrome. 

<img width="502" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOSsxQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--aee06e3be8163d8a8fa8bbd5f266d19c0b698ac7/image.png" alt="image.png" />

Windows Edge/Mac Safari and Chrome: The student either has to select "Add another trusted contact" or they select "No", then "Yes" again and the correct table is displayed.

[...]

PR #3686, opened 2024-09-11
#3485 - Student: New applications: Student can no longer immediately add trusted contact when "Yes" is selected

andrepestana-aot:
- Added default value to the DataGrid component;
- Removed extra required field asterisks;
- Added MAXIMUM_IDLE_TIME* vars to docker-compose to fix a bug with the local API container.

image


sources/docker-compose.yml
```diff
@@ -63,6 +63,10 @@ services:
       - S3_ENDPOINT=${S3_ENDPOINT}
       - S3_DEFAULT_BUCKET=${S3_DEFAULT_BUCKET}
       - S3_REGION=${S3_REGION}
+      - MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT=${MAXIMUM_IDLE_TIME_FOR_WARNING_STUDENT}
```
andrewsignori-aot, line 66: Thanks for fixing it ❤️

andrewsignori-aot (approved):
Great work and solution for the UI. Looks good 👍

sources/packages/web/src/assets/css/formio-shared.scss
```diff
@@ -705,3 +705,13 @@
 .formio-component-file ul li.list-group-item {
   border: 1px solid rgba(0, 0, 0, 0.125) !important;
 }
+
+// Removes the required sign from the input fields with .no-required-field-sign.
+.no-required-field-sign label.field-required:after {
+  content: none;
```
sh16011993, line 711: Should we force here with `!important` as well?

sources/packages/forms/src/form-definitions/sfaa2024-25.json
```diff
@@ -11249,7 +11249,13 @@
                   "disabled": false,
                   "tableView": false,
                   "modalEdit": false,
-                  "defaultValue": [],
+                  "defaultValue": [
```
sh16011993, line 11252: Can there be at most two trusted contacts only?

sh16011993 (approved):
Great work @andrepestana-aot  👍  Thanks for the walkthrough.

Left two quick questions (one is more of a business question).

sources/packages/forms/src/form-definitions/sfaa2024-25.json
andrepestana-aot: Yes, that was the existing logic.

sources/packages/web/src/assets/css/formio-shared.scss
andrepestana-aot: No need for that one as there wasn't a class preceding and overwriting it.

Issue #3131, opened 2024-04-16
Institution-Specific Reports - COE Requests


CarlyCotton:
Follow-Up to #2580

### **User Story**
As a **BC Public Institution Admin user**, I want to be able to extract dedicated reports to support day to day operations and workload. These reports will help me understand the enrollment status of students at my institution and triage my work related to confirming or declining the enrollments.
This report is primarily to assist institutions manage their workload around confirmation of enrolment beyond functionally completing it in SIMS.

### **Acceptance Criteria**
- [x] **Build a report titled "COE Requests".** 
    - Access: Make this report available only through Institution account (to only BC Public Admin user types)
    - All records shown should only relate to the institution requesting the report.
    - Can keep database values instead of translating where they provide enough information to the user. Date format: YYYY-MM-DD
    - File Layouts should be consistent with the table in the "Sample Reports" worksheet of the attached file: [Sample Reports Worksheet](https://bcgov.sharepoint.com/:x:/r/teams/03177/Shared%20Documents/DEVS/Reports/SIMS%20Reports%20%26%20Notification%20Triggers.xlsx?d=w08758022bbde485e86f9ed065d92bcfe&csf=1&web=1&e=BgvbWd)
- [x] **Required Filter Selections:** Standard selections are required (Program Year and Intensity) 
    - The records should be all student applications that are at enrolment or completed (and not archived), are attached to the institution and meet the PY, Intensity filters selected.
    - The records should be per COE and only consider those for the most current assessment. (2 disbursements, 2 records)
- [x] **COE Actioned:** This should be blank unless the COE has been confirmed or declined. It would be the date that a user confirmed or declined the COE. **Use BC Time (Vancouver) to prevent confusion between the UTC date which is stored**
- [x] **Disb Amt:** SUM of all 'estimated award; values from each disbursement (i.e. the numbers displayed on the NoA summary card). This will not include overawards and remittance, that's fine)
- [x] **Disb Date:** Earliest date of disbursement for each disbursement unless disbursement has been sent, then use actual disbursement date.
- [x] **Default Sorting:** Required status first, among required show oldest ones first.

### **Business Context**
[Sample Reports Worksheet](https://bcgov.sharepoint.com/:x:/r/teams/03177/Shared%20Documents/DEVS/Reports/SIMS%20Reports%20%26%20Notification%20Triggers.xlsx?d=w08758022bbde485e86f9ed065d92bcfe&csf=1&web=1&e=BgvbWd)

**Technical**
[...]

PR #3695, opened 2024-09-13
#3131 - Fix content length to be based on byte lenth

dheepak-aot:
# FIX CONTENT LENTH HEADER ON REPORTS CONTROLLER

## Problem statement
Currently, in reports controller we set the `Content-Length` header value as length of the csv string.  
Due to this, we have an issue when the string involves special characters as the size of these characters are more than one byte.

e.g. When a string of length 100 characters has one special character like `â` in it, then string length is `100` but the byte length is `101`. Hence when this is set in API response it eliminates any characters which is more than set content length. 

Like this

image

## Solution
Set `content-length` header with byte length.


andrewsignori-aot (approved):
Thanks for doing the fix, looks good 👍

lewischen-aot (approved):
LGTM. Thanks for making the fix 👍

Issue #3508, opened 2024-07-11
Content - Income declaration tax year dynamic year numbers


Joshua-Lakusta:
Display dynamic year numbers for the application for income years.

**Acceptance Criteria**
Student portal / Application
Financial section
- [x] Replace the tax year 2020 for all program year options. Change it to dynamic value. Shown value = first year of the program year - 1 (example 2024-2025 program year would be  2023 (2024-1 = 2023) (see 4 spots where the updates are needed for this question shown in context image 1))
<img width="1252" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFhUQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8a30ecd1cdbb8b8e2cb20c09981c142271899bb1/image.png" alt="image.png" />

- [ ] Resolve Bug in current state that does not show the expected value. Attempted dynamic value has the wrong year but is also showing incorrect value. If the bug still exists.

Student Portal
Family section :

Dependants with disability question
- [x] Replace 20XX Tax Year with new dynamic value for all program years.  Shown value = first year of the program year - 1 (example 2024-2025 program year would be  2023 (2024-1 = 2023) (see 2nd image content image for locations in question) 
- [x] Complete the same change for the change request version of this questions document uploader using the same details as the change to the application. (Change request is under the Add/Remove dependant change request)
<img width="850" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeVhUQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f904935d959748ca413ba29c63e8f2932a384891/image.png" alt="image.png" />

Student declares Partner income

- [x] Update the content for the authorization line of supplying partner income in the application and the Partner information and income change request to the following:

> I have been authorized by my spouse/common-law partner to submit their financial information, including their total reported income from their **insert year** tax return.

 
- [x] Update the content above partner income self declare field (see image below location bullet #1) question to show the following for both the application question and Partner information and income change request

> Enter your spouse/common-law partner's reported total income from line 15000 of their "insert year" income tax return. If they did not file a "insert year" income tax return, enter their total income from all sources both inside and outside of Canada.

[...]

PR #3698, opened 2024-09-13
#3508 - 

andrepestana-aot:


sources/packages/forms/src/form-definitions/sfaa2022-23.json
andrewsignori-aot, line 1: The files for the applications are different. Can you please check?

sources/packages/forms/src/form-definitions/sfaa2024-25.json
```diff
@@ -5120,6 +5120,26 @@
           "calculateServer": true,
           "lockKey": true
         },
+        {
```
andrewsignori-aot, line 5123: Should `"redrawOn": "programYear"` be added?

sources/packages/forms/src/form-definitions/sfaa2022-23.json
andrepestana-aot: Fixed. Thanks!

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "SFAA2022-23",
   "name": "SFAA2022-23",
-  "path": "sfaa2022-23",
+  "path": "sfaa2023-23",
```
sh16011993, line 4: Isn't this supposed to be `sfaa2022-23` ?

sources/packages/forms/src/form-definitions/studentdependantsappeal.json
```diff
@@ -895,6 +895,26 @@
...
+          "persistent": true,
+          "type": "hidden",
+          "tags": [],
+          "conditional": {
+            "show": "",
+            "when": null,
+            "eq": ""
+          },
+          "properties": {},
+          "calculateValue": "const [programYear] = data.programYear ? data.programYear.split(\"-\") : [];\nvalue = programYear ? programYear -1: \"\";",
```
sh16011993, line 915: For my understanding, I see that the `value` here is extracted differently from how it is done in the `sfaa` forms. Is it because the `programYear` value received here is different from that received in the `sfaa` forms?

sh16011993 (commented):
Left 2 minor comments.

andrewsignori-aot (commented):
Nice work. I resolved all my comments. The only one pending for approval is the 2022-23 form with the different path as @sh16011993 mentioned on https://github.com/bcgov/SIMS/pull/3698/files#r1762159993

sources/packages/forms/src/form-definitions/studentdependantsappeal.json
andrepestana-aot: In the SFAA form we already extract the starting year for the program year name. Here we  need to extract the starting year and subtract 1.

andrewsignori-aot (approved):
Nice work, looks good 👍

sources/packages/forms/src/form-definitions/studentdependantsappeal.json
sh16011993: Sounds good. Thanks.

sh16011993 (approved):
Nice work @andrepestana-aot 👍 LGTM

Issue #3500, opened 2024-07-10
Student: Personal Information tab: Uploading supporting documents file uploader feature is not available for first 3 options


astridSABC:
**Describe the Bug**
On the Personal Information tab, in the "Are you a resident of B.C?" and the student selects "No", the Uploading supporting documents file upload feature is not available for first 3 options for “Which of the following best describes your situation?” section

**Expected Behaviour**
On the personal information, "Are you a resident of B.C? No option, “Upload supporting documents” feature is available for all radio button options:
•	You and your spouse….
•	Your spouse/partner has been employed in B.C….
•	You will have finished 4 years of full-time…
•	Other

<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelRNQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a3169fcfd98acb21143875e64d5cc1a86ae8f838/image.png" alt="image.png" />

**Actual Behaviour**
“Upload support documents” is only available when “Other” is selected.  
Sample below showing upload support documents feature not available for first radio button "You and your spouse/partner originally..."

<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejNNQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7a79dcb7da59953a5494e1ac2cc3eb1b6f7d7e4b/image.png" alt="image.png" />

** Steps To Reproduce**

Precondition: A student can create an application and is logged into the SIMS Student portal

1.	Click “Personal Information” tab
2.	Select “No” to the “Are you a resident of B.C.?”
3.	Select “You and your spouse….” radio button In the "Which of the following best describes your situation"
4.	Note the screen for “Upload supporting documents” feature to be displayed
5.	Repeat with all the radio button options in this section:
      a.	“Your spouse/partner has been employed in B.C….”
      b.	“You will have finished 4 years of full-time…”
      c.	“Other”

Content update:

- [ ] Revise text: "you will have finished 4 years......." to "You will have completed 4 years of full-time post-secondary study in B.C. before the first day of your study period."

[...]

PR #3700, opened 2024-09-16
#3500 - Upload supporting documents bug fix

sh16011993:

- Uploader is now shown for all the 4 radio options when either of them is selected as shown below:

image

Likewise, the uploader will show for the other radio buttons as well.

- Fixed the radio option text as below:
`You will have completed 4 years of full-time post-secondary study in B.C. before the first day of your study period.`

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2022-23.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2022-23.json
@@ -5405,8 +5405,7 @@
           },
           "properties": {},
           "lockKey": true,
-          "defaultValue": "365",
-          "isNew": false
+          "defaultValue": "365"
         }
       ],
       "input": false,
@@ -8576,8 +8575,8 @@
                       "shortcut": ""
                     },
                     {
-                      "label": "you will have finished 4 years of full-time, postsecondary study in B.C. before the first day of your study period.",
-                      "value": "youWillHaveFinished4YearsOfFullTimePostsecondaryStudyInBCBeforeTheFirstDayOfYourStudyPeriod",
+                      "label": "You will have completed 4 years of full-time post-secondary study in B.C. before the first day of your study period.",
+                      "value": "youWillHaveCompleted4YearsOfFullTimePostSecondaryStudyInBCBeforeTheFirstDayOfYourStudyPeriod",
                       "shortcut": ""
                     },
                     {
@@ -8614,7 +8613,7 @@
                   "tags": [],
                   "properties": {},
                   "conditional": {
-                    "show": true,
+                    "show": "true",
                     "when": "bcResident",
                     "eq": "no",
                     "json": ""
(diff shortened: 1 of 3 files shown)


sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -8845,12 +8844,12 @@
                   "key": "uploadSupportingDocuments",
                   "tags": [],
                   "properties": {},
-                  "customConditional": "",
+                  "customConditional": "show = data['bclivingSituation'];",
```
dheepak-aot, line 8847: Should it not be just based on the `Not a BC Resident Panel`. 

i.e as soon as No is answered then it must show up.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -8576,8 +8575,8 @@
                       "shortcut": ""
                     },
                     {
-                      "label": "you will have finished 4 years of full-time, postsecondary study in B.C. before the first day of your study period.",
-                      "value": "youWillHaveFinished4YearsOfFullTimePostsecondaryStudyInBCBeforeTheFirstDayOfYourStudyPeriod",
```
andrepestana-aot, line 8580: I thought you could choose a shorter name for this but I see the other values are also very long names. Maybe not for this PR.

dheepak-aot:
Nice work. Just one comment.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
andrepestana-aot, line 8847: I think `show = data.bclivingSituation;` is preferred.

andrepestana-aot (commented):
I left some comments.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
sh16011993: I don't think so. I feel it might confuse the user if it shows up when they haven't selected any of the 4 radio buttons.

sh16011993: Checked with @CarlyCotton on this. The expectation is the way it is currently implemented. Also follows the same from the additional transportation cost.

image

dheepak-aot: Ok, I am good if that is accepted by business.

dheepak-aot (approved):
Looks good 👍

sources/packages/forms/src/form-definitions/sfaa2022-23.json
andrewsignori-aot: > Should it not be just based on the `Not a BC Resident Panel`. ? please confirm.
> 
> i.e as soon as No is answered then it must show up.
> 
> So to achieve this technically, no condition is required here.

I agree with this point and I am not sure why was it a business decision.
The uploader is no longer dependent on which option was selected. The uploader is a required filed in the same way the options are.
I cannot see also how it would confuse the user.

andrepestana-aot (approved):
LGTM

Issue #2509, opened 2023-11-15
Update formio server and js


guru-aot:
Update formio server and js

**Acceptance Criteria**
- [x] Update formio server  to 3.3.(4) or the most updated one and check on the differences from the current one.
https://github.com/formio/formio/compare/v2.5.3...v3.3.4
- [x] Update formio js version to 4.17.3 or the latest.
https://github.com/formio/formio.js/releases/tag/v4.17.3
- [x] Update release notes instructions
- [x] Check the main form applications, designation agreement, student profiles, form.IO definitions
  - [x] Ensure that the smoking test is executed locally using the web docker to check the application deployed using nginx to validate possible security policy adjustments.

**Additional context**
Created the below wiki to help with the form.io server upgrade.
[Development ‐ Form.io Server Upgrade](https://github.com/bcgov/SIMS/wiki/Development-%E2%80%90-Form.io-Server-Upgrade)

andrewsignori-aot:
## Smoking Tests Executed on DEV

- Student
  - Created a new BCeID Student.
    - Updated student profile.
  - Created a new Part-time application with a partner.
    - Provided partner information using supporting users portal.
  - Created a new Full-time application.
- Ministry
  - Created a new institution
  - Updated the new institution
  - View a submitted designation
  - Approve a designation
- Institution Side
  - Update Institution Profile
  - Created a new location
    - Updated a new location
  - Submitted a new designation
    - View a submitted a new designation
  - Created a new program
    - Edited the new program
    - Created a new offering
     - Edited the new offering
- Approved COE

PR #3731, opened 2024-09-23
#2509 - Web content security policy and forms updates

andrewsignori-aot:
## Form.io definitions fixes
- Between versions `4.19.1` and `4.19.2` of the `formiojs`, some errors started to happen in places where the "HTML Element" was used (duplicated content in the images below).
- Student Application dependents required validation needed to be adjusted.
- Validation messages where components had HTML mixed in the label content (e.g. `strong`) were displaying it in the validation message.
- `Report a change` started to fail its submission because the submitted data no longer had the input data from inside the main panel. A new panel needed to be created and components dragged into it.

### Student Welcome Page
image

#### Welcome Page
image

#### Removed `<strong>` from labels.
image

#### Update labels mixed with class and HTML
image

#### Dependents wrong validation message
image

### Institution Report a Change
image

## Non-related changes
- Added extra space between the wizard navigation and the student application content.
image
- Adjusted some `make` commands to be in sync with others (e.g. `make api`, `make web`).

PR #3740, opened 2024-09-25
#2509 - Update formio server and js - Labels adjustment

andrewsignori-aot:
Removed the `<strong>` tags from labels content.
The change does not impact the form.io layout. As shown below the labels are still bold.
image

### DIFF
--- a/sources/packages/forms/src/form-definitions/studentdependantsappeal.json
+++ b/sources/packages/forms/src/form-definitions/studentdependantsappeal.json
@@ -583,7 +583,7 @@
                 }
               ],
               "tableView": false,
-              "label": "<strong>Dependants</strong>",
+              "label": "Dependants",
               "key": "dependants",
               "protected": false,
               "persistent": true,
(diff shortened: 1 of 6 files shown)


sources/packages/forms/src/form-definitions/studentdependantsappeal.json
```diff
@@ -583,7 +583,7 @@
                 }
               ],
               "tableView": false,
-              "label": "<strong>Dependants</strong>",
+              "label": "Dependants",
```
sh16011993, line 586: Are we removing the strong because the labels are by default bold / strong now after the formio upgrade?

andrewsignori-aot: They were already bold before the upgrade, there was no need to have them added.
After the upgrade, the validation messages started to display them inside the message.
Please see the previous PR description for further information: https://github.com/bcgov/SIMS/pull/3731

Issue #3668, opened 2024-08-29
SIMS Beta User Validation


ninosamson:
**User Story**
As a product team, we want to allow a small number of student users to apply for student funding in SIMS so we can confirm the functionality works end-to-end in production, before launching. 

**Acceptance Criteria:**
- [ ] Create a beta user authentication flow for student users. The given and last names of candidate beta users will be collected and added to a beta user table. 
- [ ] Upon login, validate against Given and Last Names from the BCSC credential. 
     - [ ] If the name matches the BCSC credential, the user can create a profile. 
     - [ ] BCeID users cannot log in. 
     - [ ] If there is a mismatch of name or BCeID user, show an error message on the login page: "Unable to login as the system is not yet available."
- [ ] Beta Validation is configurable on/off (toggle to enable validation of student user credential/name) 

**Technical**
- [x] Create a toggle to allow the behavior, default `true`, suggested name `ALLOW_BETA_USERS_ONLY`.
- [x] Create a new table to store beta users information.
  - Table name: `sims.beta_users_authorizations`.
  - Columns
    - id: 
    - given_names (varchar 100): User given names.
    - last_name (varchar 100): User last names.
    - enabled_from (date and time, default now): From when the user should be able to log in.
    - _No audit columns are needed._
   - Create a unique constraint for `given_names` and `last_name`.
- [x] If the toggle is enabled, don't allow BCeID users to reach the student API. Added the below code to the [AuthorizedPartiesGuard](https://github.com/bcgov/SIMS/blob/b49899306af5a325c794e2dbfff9aaacc8068d01/sources/packages/backend/apps/api/src/auth/guards/authorized-parties.guard.ts#L59) right before the method `return`.
```ts
if(TOGGLE_IS_ENABLED &&
  userToken.azp === AuthorizedParties.student
  && userToken.identityProvider === IdentityProviders.BCeIDBoth){
  throw new UnauthorizedException(
    new ApiProcessError(
      "The student authentication provider is not allowed at this moment.",
      INVALID_BETA_USER,
    ),
  );
}
```
- [x] If the toggle is enabled, in the [jwt.strategy](https://github.com/bcgov/SIMS/blob/4dde03d07b37a188ce39381f986f5a1452ae90e8/sources/packages/backend/apps/api/src/auth/jwt.strategy.ts#L54) execute a parallel verification with the `getUserLoginInfo` to check if the beta user is present on DB. If the user is not a registered beta user throw an exception as below.
```ts
[...]

PR #3742, opened 2024-09-25
#3668 - Send the invalid beta user immediately to login page

andrepestana-aot:
The invalid beta user was being sent to the dashboard before being redirected to the login page with an invalid beta user message.

### DIFF
--- a/sources/packages/web/src/services/AuthService.ts
+++ b/sources/packages/web/src/services/AuthService.ts
@@ -267,14 +267,13 @@ export class AuthService {
         break;
       }
       case ClientIdType.Student: {
-        // Send the user to login page with invalid beta user message instead of dashboard.
-        this.priorityRedirect = {
-          name: StudentRoutesConst.INVALID_BETA_USER,
-        };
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/services/AuthService.ts
```diff
@@ -267,6 +267,13 @@ export class AuthService {
         break;
       }
       case ClientIdType.Student: {
+        // Send the user to login page with invalid beta user message instead of dashboard.
+        this.priorityRedirect = {
+          name: StudentRoutesConst.INVALID_BETA_USER,
+        };
```
andrewsignori-aot, line 273: To do the specific redirect to `INVALID_BETA_USER` should we move this inside the `if (options?.invalidBetaUser)`?

guru-aot: yes i had the same question.

andrepestana-aot: Yes, you are right. Changed. Thanks!

guru-aot (approved):

andrewsignori-aot (approved):
Looks good 👍

PR #3747, opened 2024-09-26
#2509 - Update formio server and js - Side Effects

andrewsignori-aot:
Between version 3.55 and 4.00 the server validation changed in a way that affected the calculated inputs.

First, for calculated inputs participating in validations, the value should be set as below to avoid being manipulated by a malicious user.
For instance for a hidden input with a fixed value.
- "defaultValue": null (not required)
- "calculateValue": "value = \"100000000\";"
- "calculateServer": true
- "persistent": true | false, (it must be set to false if there is no point in saving it).
image

In version 4.0.0, the order in which the calculated input is added to the screen started to matter in some scenarios. Even declaring calculated inputs as the above, if the calculated input is declared after the component that consumes its value, it would allow the user to manipulate the hidden input for calculations (the stored value would be right either way).

### PR intended fixes
- Adjusted the Student Application and Supporting Users forms to follow the recommended approach above and also moved some hidden fields before the components using their calculated value. This change affects mostly `maxIncome` and `maxMoneyValue` validations.

_Example of some hidden fields that were just moved in the form._
image

- Some fields were not updating their content when the content had expressions. The solution was to add `"refreshOnChange": true`. The issue was noticed in appeals and added to the application as a precaution.
- Dependents table in Student Application and Supporting Users Parents were having an issue where the submission was not allowed because the entire grid was `required: true`. Removing the `required: true` from the grid allowed the form to still be validated and the form submitted.

_Please note that logics like below were not changed and are marked just because the `isNew`_.
image


sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -553,6 +553,26 @@
...
+          "tags": [],
+          "conditional": {
+            "show": "",
+            "when": null,
+            "eq": ""
+          },
+          "properties": {},
+          "lockKey": true,
+          "calculateValue": "value = \"100000000\";",
+          "calculateServer": true
```
dheepak-aot, line 575: 👍

dheepak-aot (approved):
Thanks very much for the extended analysis, testing and wiki guide. Looks Good.

lewischen-aot (approved):
Looks good to me! Thanks for taking care of the side effects.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -553,6 +553,26 @@
...
+          "type": "hidden",
+          "tags": [],
+          "conditional": {
+            "show": "",
+            "when": null,
+            "eq": ""
+          },
+          "properties": {},
+          "lockKey": true,
+          "calculateValue": "value = \"100000000\";",
```
guru-aot, line 574: 👍

guru-aot (commented):
Great work @andrewsignori-aot

Issue #18, opened 2020-12-11
All users' log in / log off needs to be logged


mzmichellesmith:
As technical support/security officer, I want the  logs in and logs out activities of users recorded for the purposes of troubleshooting & audit/security purposes.

**Acceptance Criteria**
- [ ] When a user logs into their account, call api to write log entry to console.
- [ ] When a user logs out of their account (clicks logout), call api to write log entry to console.
- [ ] When the SIMS idle timeout forces a logout call api to write entry to console.
- [ ] When a user closes the browser best effort event should be logged via api
- [ ] Events logged should contain
    - [ ] Static Prefix: SIMS Audit Event 
    - [ ] IP
    - [ ] User GUID
    - [ ] Event: [Logged In/Logged Out/Browser Closed/Session Timed Out]
    - [ ] Portal: [Student User/Supporting User/Institution User/Ministry User]

**Technical**
- [ ] Create the API calls based on the UI actions/events.
  - Once the user clicks the login button, and it is successful.
  - Once the user clicks the logoff button, and it is successful.
  - Once the login expires.
  - The best effort to log when the user closes the browser even considering that the token will be valid for the API for the upcoming token expiration time (3min-ish).
- [ ] May need to create a specific API authenticated endpoint to execute the audit event only (e.g. `api/audit`).
- [ ] Basic E2E tests for the new API endpoints.
 
**Additional Context**
- This is used for Security audit and will be reviewed in Kibana/SIEM
- Users are allowed to explicitly log-out of their respective portals.
- A user can also leave their respective portal by closing their browsing session, or session time-out.
- All entries will be written to the console so that they will be recorded in the SIEM. 
- Log entry format: [PATH} From [IP Address] | User GUID: [GUID from KeyCloak], Event: [Logged In/Logged Out/Browser Closed/Session Timed Out], Portal: [Student User/Supporting User/Institution User/Ministry User]
  - NOTE: Assuming that the log entry will already be prefixed wit the date and time as other log entries in the console already have.
  - Example: /ministry/login From 68.99.156.60 | User GUID: '04A2054629E44B1ca8e18e553da99999@idir', Event: Browser Closed, Portal: Ministry
    - Note the quotation marks around the User GUID are NOT required.  They were necessary only in the context of this ticket to prevent zenhub from shortening the GUID and making it a hyperlink.

IrenaMilosevic1:
We need to log our own logon/logoff events - this is the requirement from our Security team

[...]

PR #3755, opened 2024-09-28
#18 - All users' log in / log off needs to be logged

andrepestana-aot:
image

- Added endpoint `/audit` to log the events:
  - LoggedIn;
  - LoggedOut;
  - BrowserClosed;
  - BrowserReopened;
  - SessionTimedOut;
- Added "userLoginTriggered" boolean variable to signal that a login is in process; 
- Added "userClosedBrowser" boolean variable to signal that the user closed the browser and a "BrowserReopened" event is in process when the user is authenticated; 
- Added "userSessionTimedOut" boolean variable to signal that the user session has timed out; 
- Added E2E tests for the new endpoint.

Reference:
- Chrome is deprecating window.unload: https://developer.chrome.com/docs/web-platform/deprecating-unload
- Require user gesture for beforeunload dialogs: https://chromestatus.com/feature/5082396709879808


sources/packages/backend/apps/api/src/app.module.ts
```diff
@@ -25,6 +25,8 @@ import { ConfigModule } from "@sims/utilities/config";
 import { DatabaseModule } from "@sims/sims-db";
 import { NotificationsModule } from "@sims/services/notifications";
 import { QueueModule } from "@sims/services/queue";
+import { AuditController } from "./route-controllers/audit/audit.controller";
+import { AuditService } from "./services/audit/audit.service";
```
andrewsignori-aot, line 29: Please use the indexes.

sources/packages/web/src/services/AuthService.ts
```diff
@@ -67,6 +74,16 @@ export class AuthService {
    */
   priorityRedirect?: LocationAsRelativeRaw = undefined;
 
+  /**
+   * Function that makes a request to the API to log that the user has closed the browser/tab.
+   * @param event event received by the browser.
+   */
+  logUserClosedBrowser = async function (event: Event) {
+    event.preventDefault();
```
sh16011993, line 82: Is this added to prevent the browser default behavior of showing the confirmation dialog?

sources/packages/backend/apps/api/src/route-controllers/audit/_tests_/audit.controller.e2e-spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    auditService = await getProviderInstanceForModule(
+      module,
+      AppModule,
+      AuditService,
+    );
+  });
+
+  beforeAll(async () => {
+    auditService.logger.log = jest.fn();
+  });
```
andrewsignori-aot, line 34: Is there a reason to have `beforeAll` used twice?

sources/packages/web/src/services/AuthService.ts
andrepestana-aot: Yes, usually `beforeUnload` is used to confirm if the user wants to leave the page.

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -0,0 +1,39 @@
+import { Controller, Post, Param, Req } from "@nestjs/common";
+import { ApiTags } from "@nestjs/swagger";
+import { AuthorizedParties, IUserToken } from "../../auth";
+import { AllowAuthorizedParty, UserToken } from "../../auth/decorators";
+import BaseController from "../BaseController";
+import { Request } from "express";
+import { AuditService } from "../../services";
+import { AuditEvent } from "../../services/audit/audit-event.enum";
```
andrewsignori-aot, line 8: I guess we also add those to the index, don't we?

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Audit events log.
+   * @param event audit event.
+   * @param userToken user token.
+   * @param request request object.
+   */
+  @Post(":event")
+  async audit(
+    @Param("event") event: AuditEvent,
```
andrewsignori-aot, line 31: The parameter must be validated, please use the `ParseEnumPipe` as the below example.
```ts
@Param("supportingUserType", new ParseEnumPipe(SupportingUserType))
```

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -0,0 +1,38 @@
...
+    super();
+  }
+
+  /**
+   * Audit events log.
+   * @param event audit event.
+   * @param userToken user token.
+   * @param request request object.
+   */
+  @Post(":event")
```
andrewsignori-aot, line 28: Was there an advantage to using the `:event` as a URL parameter instead of a property in the POST body?

andrepestana-aot: I think it's simpler to implement and easier to understand what's being done this way.

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
```diff
@@ -0,0 +1,91 @@
...
+export class AuditService {
+  /**
+   * Audit logging.
+   * @param clientIp client ip.
+   * @param tokenUserName token user name.
+   * @param event audit event.
+   * @param authorizedParty authorized party.
+   */
+  async audit(
+    clientIp: string | string[],
```
andrewsignori-aot, line 17: What is the scenario where the audit method will receive a `string[]`?

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
```diff
@@ -0,0 +1,91 @@
+import { LoggerService } from "@nestjs/common";
+import { InjectLogger } from "@sims/utilities/logger";
+import { AuditEvent } from "./audit-event.enum";
+import { AuthorizedParties } from "../../auth";
+
+const SIMS_AUDIT_EVENT_PREFIX = "SIMS Audit Event";
+
+export class AuditService {
```
andrewsignori-aot, line 8: Is it missing the `@Injectable()`?

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
```diff
@@ -0,0 +1,91 @@
...
+   * @param tokenUserName token user name.
+   * @param event audit event.
+   * @param authorizedParty authorized party.
+   */
+  async audit(
+    clientIp: string | string[],
+    tokenUserName: string,
+    event: AuditEvent,
+    authorizedParty: AuthorizedParties,
+  ): Promise<void> {
```
andrewsignori-aot, line 21: No need to return a promise.

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export class AuditService {
+  /**
+   * Audit logging.
+   * @param clientIp client ip.
+   * @param tokenUserName token user name.
+   * @param event audit event.
+   * @param authorizedParty authorized party.
+   */
+  async audit(
```
andrewsignori-aot, line 16: No need for async.

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
andrepestana-aot: `request.headers["x-forwarded-for"]` can return a `string[]`.

sources/packages/web/src/services/AuthService.ts
andrewsignori-aot: So, if we do not add the `event.preventDefault();` the user will see the modal?

sources/packages/backend/apps/api/src/route-controllers/audit/_tests_/audit.controller.e2e-spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    const endpoint = `/audit/${AuditEvent.LoggedIn}`;
+    const token = await getAESTToken(AESTGroups.Operations);
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.CREATED);
+
+    expect(auditService.logger.log).toHaveBeenCalledWith(
```
andrewsignori-aot, line 43: The mock should be reset in the `beforeEach` otherwise the test result will be misleading.

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
andrewsignori-aot: I do not agree with either argument.
In terms of expanding the implementation, a body would be better.
In terms of providing arguments to a POST, the body would be preferred.
We can ask another team member, I would not say that it is a blocker.
@dheepak-aot thoughts or preferences on this?
@sh16011993 please share your opinion as a reviewer.

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
andrewsignori-aot: Yes because it deals with headers. Is there a possibility that `x-forwarded-for` be returned twice?
This method cares about receiving an IP address only and not an array of them.

sources/packages/web/src/services/http/AuditApi.ts
```diff
@@ -0,0 +1,15 @@
+import { AuditEvent } from "@/types/contracts/AuditEnum";
+import HttpBaseClient from "./common/HttpBaseClient";
+/**
+ * Audit API.
+ */
+export class AuditApi extends HttpBaseClient {
+  /**
+   * Sends an event to be logged in the API logger.
+   * @param event audit event.
+
```
andrewsignori-aot, line 10: Please remove the blank line.

sources/packages/web/src/services/http/AuditApi.ts
```diff
@@ -0,0 +1,15 @@
...
+   */
+  public async audit(event: AuditEvent): Promise<void> {
```
andrewsignori-aot, line 12: Please remove the `public`.

sources/packages/web/src/services/http/AuditApi.ts
```diff
@@ -0,0 +1,15 @@
+import { AuditEvent } from "@/types/contracts/AuditEnum";
+import HttpBaseClient from "./common/HttpBaseClient";
+/**
```
andrewsignori-aot, line 3: Usually, there is a blank line after the `imports`.

sources/packages/web/src/views/institution/Login.vue
```diff
@@ -74,6 +75,7 @@ export default defineComponent({
     },
   },
   setup(props) {
+    sessionStorage.setItem(USER_LOGIN_TRIGGERED, "true");
```
andrewsignori-aot, line 78: Should it not be added inside the login button action?
The idea is to know when the user initiated the login, right?

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
dheepak-aot: From the API perspective `event` is the payload itself here and not the context of the payload. Also with that in mind, making the event in body would make the payload extensible.
Hence my recommendation would be body.

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -71,6 +72,7 @@ export default defineComponent({
     },
   },
   setup() {
+    sessionStorage.setItem(USER_LOGIN_TRIGGERED, "true");
```
sh16011993, line 75: Should we add it on the click of the login button? Same in other `Login.vue` files.

sources/packages/web/src/services/http/AuditApi.ts
sh16011993, line 10: Minor: Please remove the empty line.

sources/packages/web/src/services/AuthService.ts
```diff
@@ -93,6 +110,20 @@ export class AuthService {
       });
 
       if (this.keycloak.authenticated) {
+        // In case of user closed browser without logout.
+        window.addEventListener("beforeunload", this.logUserClosedBrowser);
```
andrewsignori-aot, line 114: Due to the amount of checks in the session storage, I would recommend centralizing the logic in some services.
Please see the below example with the 3 methods needed for the user triggering login.
It would encapsulate the constants, the logic, the checks, and all `sessionStorage` related operations, and it would improve the code readability.
I would not see this as a blocker but I would strongly recommend to give it a try.

```ts
const USER_LOGIN_TRIGGERED = "userLoginTriggered";
const USER_CLOSED_BROWSER = "userClosedBrowser";
const USER_SESSION_TIMED_OUT = "userSessionTimedOut";

export class AuditService {
  static userLoginTriggered() {
    sessionStorage.setItem(USER_LOGIN_TRIGGERED, "true");
  }

  static wasUserLoginTriggered() {
    return sessionStorage.getItem(USER_LOGIN_TRIGGERED) === "true";
  }

  static resetLoginTriggered() {
    sessionStorage.removeItem(USER_LOGIN_TRIGGERED);
  }
}
```

sources/packages/web/src/services/AuthService.ts
```diff
@@ -67,6 +74,16 @@ export class AuthService {
...
   priorityRedirect?: LocationAsRelativeRaw = undefined;
 
+  /**
+   * Function that makes a request to the API to log that the user has closed the browser/tab.
+   * @param event event received by the browser.
+   */
+  logUserClosedBrowser = async function (event: Event) {
+    event.preventDefault();
+    sessionStorage.setItem(USER_CLOSED_BROWSER, "true");
+    await ApiClient.AuditApi.audit(AuditEvent.BrowserClosed);
```
sh16011993, line 84: I see that this gets called before clicking on the `Leave` of the dialog confirmation from the browser. Just trying to say that there is a possibility that the user cancels the page reload from the dialog and still this will be logged as user closed the browser. Under best effort, we may still be willing to keep this behavior. 

image

andrewsignori-aot (commented):

sources/packages/web/src/services/AuthService.ts
```diff
@@ -93,6 +110,20 @@ export class AuthService {
       });
 
       if (this.keycloak.authenticated) {
+        // In case of user closed browser without logout.
+        window.addEventListener("beforeunload", this.logUserClosedBrowser);
+        if (sessionStorage.getItem(USER_CLOSED_BROWSER) === "true") {
+          // In case of user reopened browser with an active session.
+          await ApiClient.AuditApi.audit(AuditEvent.BrowserReopened);
```
sh16011993, line 117: I see that this only gets called when the browser is reopened for the first time after closing. 2nd time onwards, it is not getting called.

sh16011993 (commented):
@andrepestana-aot Completed the review. Left a few comments.

sources/packages/web/src/services/AuthService.ts
sh16011993: I think it is because on closing the tab, the sessionStorage is cleared vs it is not cleared on tab/window reload.

sources/packages/web/src/services/http/AuditApi.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Audit API.
+ */
+export class AuditApi extends HttpBaseClient {
+  /**
+   * Sends an event to be logged in the API logger.
+   * @param event audit event.
+
+   */
+  public async audit(event: AuditEvent): Promise<void> {
+    this.postCall(`audit/${event}`, null);
```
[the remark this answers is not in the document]
andrepestana-aot: I use the await in every call to this method so I don't need await here, right?

sources/packages/web/src/services/AuthService.ts
andrepestana-aot: You are right. In my tests I used CTRL+SHIFT+T that reopens the browser but that also keeps the sessionStorage vars. When we close the tab and open another tab the sessionStorage is fresh and it won't find the var. I'm using localStorage for this one. Thanks!

sources/packages/web/src/services/AuthService.ts
andrepestana-aot: I removed that confirmation page that was not required. Thanks.

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
andrepestana-aot: ok. Moved it to the body.

sources/packages/web/src/services/AuthService.ts
andrepestana-aot: Actually we checked that the default behavior of `beforeUnload` is to close the browser without any confirmation. If the confirmation is required then `event. preventDefault()` can be called. As we didn't have any confirmation message I removed it from this function.

I also removed the `await` from the call to the request because with await the browser prioritizes closing the browser and cancels the request.

andrewsignori-aot: Thanks @andrepestana-aot. Sounds good to me 👍

sources/packages/backend/apps/api/src/route-controllers/audit/models/audit.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { AuditEvent } from "apps/api/src/services";
```
andrewsignori-aot, line 1: Please import using `import { AuditEvent } from "../../../services"`;

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+@Controller("audit")
+@ApiTags("audit")
+export class AuditController extends BaseController {
+  constructor(private readonly auditService: AuditService) {
+    super();
+  }
+
+  /**
+   * Audit events log.
+   * @param event audit event.
```
andrewsignori-aot, line 25: Please adjust the param name.

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+   * @param userToken user token.
+   * @param request request object.
+   */
+  @Post()
+  audit(
+    @Body() payload: AuditAPIInDTO,
+    @UserToken() userToken: IUserToken,
+    @Req() request: Request,
+  ): void {
+    const clientIp =
```
andrewsignori-aot, line 35: Minor, I guess IP should be in the upper case following other acronyms.

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
```diff
@@ -0,0 +1,92 @@
...
+export class AuditService {
+  /**
+   * Audit logging.
+   * @param clientIp client ip.
+   * @param tokenUserName token user name.
+   * @param event audit event.
+   * @param authorizedParty authorized party.
+   */
+  audit(
+    clientIp: string,
```
andrewsignori-aot, line 18: Same from https://github.com/bcgov/SIMS/pull/3755/files#r1786445283

sources/packages/backend/apps/api/src/services/audit/audit.service.ts
```diff
@@ -0,0 +1,92 @@
...
+    );
+  }
+
+  /**
+   * Gets an event friendly name to be logged given an audit event.
+   * @param event audit event.
+   * @returns audit event friendly name.
+   */
+  private getEventFriendlyName(event: AuditEvent): string {
+    let eventFriendlyName = "";
```
andrewsignori-aot, line 36: Not a blocker, as a fallback it can be initialized with the `event`.
Same for the `getPortalFriendlyName`.

sources/packages/backend/apps/api/src/route-controllers/audit/_tests_/audit.controller.e2e-spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    );
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
```
andrewsignori-aot, line 123: So far, I believe that every E2E uses the `beforeEach` at the top. Can we follow the same?

sources/packages/web/src/services/AuditService.ts
```diff
@@ -0,0 +1,69 @@
+import {
+  USER_CLOSED_BROWSER,
+  USER_LOGIN_TRIGGERED,
+  USER_SESSION_TIMED_OUT,
+} from "@/constants";
```
andrewsignori-aot, line 5: Are these constants used anywhere else?
The idea of having this service is to encapsulate also the constants. Unless I am missing something they can all be declared inside this file and not exported.

sources/packages/web/src/services/AuthService.ts
```diff
@@ -67,9 +70,20 @@ export class AuthService {
...
   priorityRedirect?: LocationAsRelativeRaw = undefined;
 
+  /**
+   * Function that makes a request to the API to log that the user has closed the browser/tab.
+   */
+  logUserClosedBrowser = function () {
+    AuditService.userClosedBrowser();
+
+    // Please note that this call is not awaiting the request to finish by design.
+    // This runs on `beforeUnload` and with the `await` the browser cancels the request in order to prioritize closing the browser.
```
andrewsignori-aot, line 80: Are the back ticks supposed to be single quotes?

sources/packages/web/src/services/AuthService.ts
```diff
@@ -67,9 +70,20 @@ export class AuthService {
...
+    ApiClient.AuditApi.audit({ event: AuditEvent.BrowserClosed });
+  };
+
   /**
    * Initializes the authentication service with the proper client type.
-   * @param clientType Keycloak client type to be used.
+   * @param clientType bKeycloak client type to be used.
```
andrewsignori-aot, line 86: Is this an intentional change?

sources/packages/web/src/components/common/IdleTimeChecker.vue
```diff
@@ -82,6 +85,7 @@ export default defineComponent({
     });
 
     const logoff = async () => {
+      sessionStorage.setItem(USER_SESSION_TIMED_OUT, "true");
```
sh16011993, line 88: Can we make it `static` as well like the others?

sources/packages/web/src/views/aest/Login.vue
```diff
@@ -73,6 +74,7 @@ export default defineComponent({
   setup() {
     const { executeLogin } = useAuth();
     const login = async () => {
+      sessionStorage.setItem(USER_LOGIN_TRIGGERED, "true");
```
sh16011993, line 77: It can use the `static` defined `userLoginTriggered`. Same for the other login files.

sources/packages/web/src/services/http/AuditApi.ts
```diff
@@ -1,15 +1,14 @@
...
 export class AuditApi extends HttpBaseClient {
   /**
    * Sends an event to be logged in the API logger.
-   * @param event audit event.
-
+   * @param payload payload.
    */
-  public async audit(event: AuditEvent): Promise<void> {
-    this.postCall(`audit/${event}`, null);
+  async audit(payload: AuditAPIInDTO): Promise<void> {
```
sh16011993, line 11: 👍

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -25,14 +26,20 @@ export class AuditController extends BaseController {
...
   audit(
-    @Param("event", new ParseEnumPipe(AuditEvent)) event: AuditEvent,
+    @Body() payload: AuditAPIInDTO,
     @UserToken() userToken: IUserToken,
     @Req() request: Request,
   ): void {
     const clientIp =
-      request.headers["x-forwarded-for"] || request.socket.remoteAddress;
-    this.auditService.audit(clientIp, userToken.userName, event, userToken.azp);
+      (request.headers["x-forwarded-for"] as string) ||
```
sh16011993, line 36: For my understanding, why did we need this? Was something fixed as a result of this?

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
```diff
@@ -25,14 +26,20 @@ export class AuditController extends BaseController {
    * @param userToken user token.
    * @param request request object.
    */
-  @Post(":event")
+  @Post()
   audit(
-    @Param("event", new ParseEnumPipe(AuditEvent)) event: AuditEvent,
+    @Body() payload: AuditAPIInDTO,
```
sh16011993, line 31: 👍

sources/packages/web/src/services/http/AuditApi.ts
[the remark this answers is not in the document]
andrewsignori-aot: Yes, you need. You need either return the promise or use `await`.

sources/packages/web/src/services/http/AuditApi.ts
andrewsignori-aot: I was not able to see this change. Was it updated?

sources/packages/web/src/views/institution/Login.vue
andrewsignori-aot: The user loading the page vs the user clicking on the button are two different things.

sources/packages/web/src/services/AuthService.ts
```diff
@@ -93,6 +107,20 @@ export class AuthService {
       });
 
       if (this.keycloak.authenticated) {
+        // In case of user closed browser without logout.
+        window.addEventListener("beforeunload", this.logUserClosedBrowser);
+        if (AuditService.hasUserClosedBrowser()) {
+          // In case of user reopened browser with an active session.
+          await ApiClient.AuditApi.audit({ event: AuditEvent.BrowserReopened });
+          AuditService.resetUserClosedBrowser();
+        }
```
andrewsignori-aot, line 116: I would say that `AuditService.hasUserClosedBrowser()` and `if (AuditService.wasUserLoginTriggered())` are mutually exclusive and an `else` can be added.

sources/packages/web/src/services/AuthService.ts
```diff
@@ -93,6 +107,20 @@ export class AuthService {
...
+        if (AuditService.wasUserLoginTriggered()) {
+          // Call audit api to log user logon.
+          await ApiClient.AuditApi.audit({ event: AuditEvent.LoggedIn });
+          AuditService.resetLoginTriggered();
+          AuditService.resetUserClosedBrowser();
```
andrewsignori-aot, line 121: Can `AuditService.resetUserClosedBrowser();` be removed from lines 115 and 121 and just be called after the `if` conditions?
I would say that even the `AuditService.resetLoginTriggered();` can be moved outside the if condition.

andrewsignori-aot (commented):

sh16011993 (commented):
Thanks for making the changes, please have a look at the comments.

sources/packages/web/src/services/AuthService.ts
andrepestana-aot: no, thanks.

sources/packages/backend/apps/api/src/route-controllers/audit/audit.controller.ts
andrepestana-aot: If the connection is direct between client and server `request.socket.remoteAddress` is used but if you're running the application behind a proxy, you might need to get the IP address from `X-Forwarded-For` headers.
The conversion of `request.headers["x-forwarded-for"]` to string is because [IncomingHttpHeaders](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v18/http.d.ts#L49) extends NodeJS.Dict<string | string[]> but in practice it's a string with comma separated IPs.

sh16011993 (approved):
Nice work @andrepestana-aot. Thanks for making the changes. LGTM 👍

sources/packages/backend/apps/api/src/route-controllers/audit/_tests_/audit.controller.e2e-spec.ts
andrewsignori-aot: Nope, I meant `beforeEach`.
If searched by `beforeEach` in the codebase there will be 37 results and none for `afterEach`. 
Both would work in this case, but we have been using the first one as far as I can see.

andrewsignori-aot:
Please find below an E2E test that I ended up creating to test the query. It can be used as a starting point for E2Es if you want.
```ts
import { createMock } from "@golevelup/ts-jest";
import { INestApplication } from "@nestjs/common";
import { QueueNames } from "@sims/utilities";
import {
  createTestingAppModule,
  describeProcessorRootTest,
} from "../../../../../test/helpers";
import {
  E2EDataSources,
  createE2EDataSources,
  saveFakeApplicationDisbursements,
  saveFakeStudent,
} from "@sims/test-utils";
import { Job } from "bull";
import {
  ApplicationStatus,
  DisabilityStatus,
  DisbursementScheduleStatus,
  NotificationMessageType,
  WorkflowData,
} from "@sims/sims-db";
import { IsNull, Not } from "typeorm";
import { StudentApplicationNotificationsScheduler } from "../../student-application-notifications/student-application-notifications.scheduler";

describe(
  describeProcessorRootTest(QueueNames.StudentApplicationNotifications),
  () => {
    let app: INestApplication;
    let processor: StudentApplicationNotificationsScheduler;
    let db: E2EDataSources;

    beforeAll(async () => {
      // Set the SFAS folder to the files mocks folders.
      const { nestApplication, dataSource } = await createTestingAppModule();
      app = nestApplication;
      db = createE2EDataSources(dataSource);
      // Processor under test.
      processor = app.get(StudentApplicationNotificationsScheduler);
    });

    beforeEach(async () => {
      // Cancel all applications to ensure tha existing data will not affect these tests.
      await db.application.update(
        { applicationStatus: Not(ApplicationStatus.Cancelled) },
        { applicationStatus: ApplicationStatus.Cancelled },
      );
    });

    it(
      "Should generate a notification for PD/PPD student mismatch close to the offering end date " +
        "when the application is completed and at least one disbursement is pending and there is a PD/PPD mismatch.",
      async () => {
        // Arrange
        // Create a student with a non-approved disability.
        const student = await saveFakeStudent(db.dataSource, undefined, {
          initialValue: { disabilityStatus: DisabilityStatus.Requested },
        });
        // Create an application with the disability as true.
        const application = await saveFakeApplicationDisbursements(
          db.dataSource,
          { student },
          {
            applicationStatus: ApplicationStatus.Completed,
            currentAssessmentInitialValues: {
              workflowData: {
                calculatedData: {
                  pdppdStatus: true,
                },
              } as WorkflowData,
            },
            createSecondDisbursement: true,
            firstDisbursementInitialValues: {
              disbursementScheduleStatus: DisbursementScheduleStatus.Sent,
            },
            secondDisbursementInitialValues: {
              disbursementScheduleStatus: DisbursementScheduleStatus.Pending,
            },
          },
        );
        // Queued job.
        const job = createMock<Job<void>>();

        // Act
        await processor.studentApplicationNotifications(job);

        // Assert
        const notification = await db.notification.findOne({
          select: {
            id: true,
            messagePayload: true,
          },
          relations: { notificationMessage: true },
          where: {
            notificationMessage: {
              id: NotificationMessageType.StudentPDPPDApplicationNotification,
            },
            dateSent: IsNull(),
            user: { id: student.user.id },
          },
        });
        expect(notification).toBeDefined();
        expect(notification.messagePayload).toStrictEqual({
          email_address: application.student.user.email,
          template_id: "7faea39f-cf8e-41ee-af02-c4790cac5b26",
          personalisation: {
            lastName: application.student.user.lastName,
            givenNames: application.student.user.firstName,
          },
        });
      },
    );
  },
);

```

sources/packages/web/src/services/AuditService.ts
andrewsignori-aot: To be clear, the suggestion was just to have the constants declared at the top of the file and not exported.
image

andrewsignori-aot (approved):
Great work and thanks for making the changes, looks good 👍

Issue #3260, opened 2024-05-09
CAS Integration 3A - Create new Supplier and Site


HRAGANBC:
## **User Story:** 
As an SABC Finance team member (Ministry user), I need all students in SIMS to have a supplier and site number attached. If we do not receive a match from the initial CAS supplier search after SIN validation, we need to create a new supplier and site for the student. 

## **Acceptance Criteria:**

- [x] Create new Supplier Status: "Manual Intervention" Set the flag isValid=FALSE

### **Get Supplier (Last Name and SIN)**
- [x] Ensure **initial** Get Supplier call uses the follow name format to search:
    - [x] ALL upper case letters
    - [x] Hyphens accepted, not dashes
    - [x] Special characters: Please use our current translation strategy to convert special characters where possible, or omit them. 
    - [x] If the student does NOT have a Given name, do NOT attempt the Get Supplier Call. Status: Manual Intervention Required
    - [x] If the student does NOT have a Canada address, do NOT attempt the Get Supplier Call. Status: Manual Intervention Required

### Get Supplier Response "None found/0 Count":
- [x] Schedule these to go once a day at noon during weekdays for now.
- [x] Call **Create New Supplier and Site (command 1)** 
- [x] Fields to include (See Field Requirements File below): SIN, AddressLine1, City, Province, Country, PostalCode, EmailAddress (Optional/Conditional fields can be left out or left null)
    - [x] **Name Field:** Use ALL CAPS. LASTNAME, GIVENNAMES (must have a space after the comma). Hyphens accepted, not dashes (CAS only translation).
    - [x] **AddressLine1:** Use first addressline from SIMS. Truncate if ours is longer than supported. Convert or omit special characters (like is done with other integrations).
    - [x] **City:** Use as-is
    - [x] **Province:** Use abbreviated versions AB,BC,MB,NB,NL,NS,NT,NU,ON,PE,QC,YT
    - [x] **Country:** For Canada: CA
    - [x] **Postal Code:** ALL CAPS, no space.
    - [x] **Email Address:** Use as-is

- [x] If successful response to Create New Site & Supplier, then update SIMS CAS table with the returned `supplierNumber`, `SiteCode` and any other required info and set `isValid`=true
- [x] If error response, retry for now. Manual intervention may be required and will be dealt with on ticket #3258
[...]

PR #3784, opened 2024-10-09
#3260 - CAS Integration 3A - Create new Supplier and Site - Get Supplier

andrewsignori-aot:
- Centralized the method to convert "special" characters into ASCII.
- Moved the `FILE_DEFAULT_ENCODING` const to the `utilities` lib to be consumed by the now shared `translateToASCII` method (around 9 files impacted because of this change). It was previously defined at the shared services level that has a dependency on the utility library.
- Created unit tests for the `convertToASCII` and for the method `getSupplierInfoFromCAS` that is mocked in E2E tests.
- Added the unit test mock helper [@suites/unit](https://github.com/suites-dev/suites), following the instructions from [here](https://github.com/suites-dev/suites). Please see the [unit test implemented](https://github.com/bcgov/SIMS/blob/2c52e6efb32c49b3be2af53eb5e2ebac6a8fb832/sources/packages/backend/libs/integrations/src/cas/_tests_/unit/cas.service.getSupplierInfoFromCAS.spec.ts) and/or the Nestjs doc to see the benefits. This was the one recommended in the [Nestjs page](https://docs.nestjs.com/recipes/automock) (please keep in mind that automock was renamed to suites).
- Create the DB migration for the new type `Manual intervention` to be used in the upcoming PR.

### Migration revert

image


sources/packages/backend/libs/integrations/src/cas/_tests_/unit/cas.service.getSupplierInfoFromCAS.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+import { HttpService } from "@nestjs/axios";
+import { ConfigService } from "@sims/utilities/config";
+
+describe("CASService-getSupplierInfoFromCAS", () => {
+  let casService: CASService;
+  let httpService: Mocked<HttpService>;
+  let configService: Mocked<ConfigService>;
+
+  beforeAll(async () => {
+    const { unit, unitRef } = await TestBed.solitary(CASService).compile();
```
andrepestana-aot, line 12: Cool!

andrepestana-aot (approved):
Looks good to me. Cool tests and nice reorganization of the code!

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -63,7 +63,8 @@ export class CASService {
     sin: string,
     lastName: string,
```
dheepak-aot, line 64: The last name uppercase conversion is happening at the consumer end. As we are doing conversions here, as a suggestion we can move the uppercase conversion into this method.

andrewsignori-aot: I am not sure if I am following. Is the suggestion to move the `toUpperCase` inside the `convertToASCII`?
The `toUpperCase` is a CAS requirement and it is happening right before CAS endpoint is called.

sources/packages/backend/libs/utilities/src/string-utils.ts
```diff
@@ -16,3 +18,127 @@ export function replaceLineBreaks(
...
+
+/**
+ * Converts input into an ASCII 7 bit buffer replacing special characters
+ * by equivalent ASCII characters when possible.
+ * @param rawContent raw content in string format.
+ * @returns a string of the input with extended ASCII characters (ISO-8859-1)
+ * converted to equivalent ASCII characters. If null or undefined is provided,
+ * null will be returned.
+ */
+export function convertToASCII(rawContent?: string): string | null {
```
dheepak-aot, line 39: What is the reason to make this method null safe?

sources/packages/backend/libs/utilities/src/string-utils.ts
```diff
@@ -16,3 +18,127 @@ export function replaceLineBreaks(
   }
   return data.replace(REPLACE_LINE_BREAK_REGEX, options?.replaceText ?? "");
 }
+
+/**
+ * List of characters to be translated before the content is converted to an
+ * ASCII buffer. These characters do not have an equivalent character in ISO8859-1
+ * and will ended up being converted to an unexpected character.
+ */
+const ASCII_INDIRECT_CONVERSIONS: Record<string, string> = {
```
dheepak-aot, line 27: 👍

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
dheepak-aot: My bad, I missed notice the `toUpperCase()` here. My point was to remove the `toUpperCase` at the consumer end. 

image

andrewsignori-aot: Ohhh I can see now.
I saw the upper case and "characters conversation" more related to a CAS-specific "payload" preparation.
For instance, doing the `toUpperCase` here would indicate the character's conversation should be happening here also, IMO.

sources/packages/backend/libs/utilities/src/string-utils.ts
andrewsignori-aot: As it is a generic method and no longer dedicated to the SFTP only, it seems better to ensure that it can handle null values also.

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
andrewsignori-aot: Remove the duplicated `toUpperCase` 😉

dheepak-aot (approved):
Great work. Thanks for the unit test library and the change.

Issue #3748, opened 2024-09-27
Content - Student current year income dynamic year / Remove COVID option for Student Current Year Income


Joshua-Lakusta:
Dynamic year for current year income was changed from original developed year and needs to be changed back.

- [ ] Student current year income **application** question the dynamic year calculation should be first year of program year.
- [ ] Student current year income **change request** question should have a dynamic year calculation for first year of program year.

example:
2024-2025 would be 2024

<img width="1271" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2tWQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--706b8b645b257f9f7f71ab4e4b0f6febf782d023/image.png" alt="image.png" />

ninosamson:
removing last two ACs as they have been implemented in previous ticket

PR #3791, opened 2024-10-11
#3748 - Content - Student current year income dynamic year

bidyashish:
- [x] Student current year income **application** question the dynamic year calculation should be first year of program year.
- [x] Student current year income **change request** question should have a dynamic year calculation for first year of program year.

Below AC were Done previously and not needed any more as part of this PR.

- [ ] Remove the "COVID-19" radio option for the student's current year income **application** question
- [ ] Remove the "COVID-19" radio option for the student's current year income **change request** question

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2022-23.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2022-23.json
@@ -5452,10 +5452,10 @@
           "input": true,
           "tableView": true,
           "key": "currentTaxYear",
-          "label": "currentTaxYear",
+          "label": "Current Tax Year",
           "protected": false,
           "unique": false,
-          "persistent": true,
+          "persistent": false,
           "type": "hidden",
           "tags": [],
           "conditional": {
--- a/sources/packages/forms/src/form-definitions/studentfinancialinformationappeal.json
+++ b/sources/packages/forms/src/form-definitions/studentfinancialinformationappeal.json
@@ -72,10 +72,10 @@
             "input": true,
             "tableView": true,
             "key": "currentTaxYear",
-            "label": "currentTaxYear",
+            "label": "Current Tax Year",
             "protected": false,
             "unique": false,
-            "persistent": true,
+            "persistent": false,
             "type": "hidden",
             "tags": [],
             "conditional": {
--- a/sources/packages/forms/src/form-definitions/sfaa2023-24.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2023-24.json
@@ -5452,10 +5452,10 @@
           "input": true,
           "tableView": true,
           "key": "currentTaxYear",
-          "label": "currentTaxYear",
+          "label": "Current Tax Year",
           "protected": false,
           "unique": false,
-          "persistent": true,
+          "persistent": false,
           "type": "hidden",
           "tags": [],
           "conditional": {
--- a/sources/packages/forms/src/form-definitions/sfaa2024-25.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2024-25.json
@@ -5452,10 +5452,10 @@
           "input": true,
           "tableView": true,
           "key": "currentTaxYear",
-          "label": "currentTaxYear",
+          "label": "Current Tax Year",
           "protected": false,
           "unique": false,
-          "persistent": true,
+          "persistent": false,
           "type": "hidden",
           "tags": [],
           "conditional": {


sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -5447,6 +5447,26 @@
...
           "defaultValue": "365"
+        },
+        {
+          "input": true,
+          "tableView": true,
+          "key": "currentTaxYear",
+          "label": "currentTaxYear",
+          "protected": false,
+          "unique": false,
+          "persistent": true,
```
andrewsignori-aot, line 5458: Please set the `persistent` to `false` to prevent saving it to the DB.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -5447,6 +5447,26 @@
...
+          "tags": [],
+          "conditional": {
+            "show": "",
+            "when": null,
+            "eq": ""
+          },
+          "properties": {},
+          "customClass": "",
+          "lockKey": true,
+          "calculateValue": "value = data.programYear ? data.programYear: \"\";"
```
andrewsignori-aot, line 5469: If `data.programYear` already has the desired value, is there a need to create a second hidden field?

sources/packages/forms/src/form-definitions/studentfinancialinformationappeal.json
```diff
@@ -68,6 +68,25 @@
             "calculateValue": "// From the program year, get the calendar start year of the the program year.\r\n// e.g. program year  = 2022-2023, calendar start year is 2022. And tax year should be 2021.\r\nconst [startYear] = data.programYear ? data.programYear.split(\"-\"):[];\r\nconst taxYear = startYear ? (+startYear)-1 : \"\";\r\nvalue = taxYear;",
             "lockKey": true
         },
+        {
+            "input": true,
+            "tableView": true,
+            "key": "currentTaxYear",
+            "label": "currentTaxYear",
```
andrewsignori-aot, line 75: Minor, the label can follow the same pattern as the exiting ones (e.g. `Current Tax Year`).
image

sources/packages/forms/src/form-definitions/studentfinancialinformationappeal.json
```diff
@@ -68,6 +68,25 @@
...
+            "protected": false,
+            "unique": false,
+            "persistent": true,
```
andrewsignori-aot, line 78: Please set the `persistent` to `false`.

andrewsignori-aot (commented):
Nice changes, please take a look at the comments.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
guru-aot: Until or unless we use the hidden field in the validation of any other input, no need to have it persistent.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -5447,6 +5447,26 @@
           "properties": {},
           "lockKey": true,
           "defaultValue": "365"
+        },
+        {
+          "input": true,
+          "tableView": true,
+          "key": "currentTaxYear",
+          "label": "currentTaxYear",
```
guru-aot, line 5455: Please have the Current Tax Year label name as descriptive and not as currentTaxYear

sources/packages/forms/src/form-definitions/studentfinancialinformationappeal.json
```diff
@@ -279,7 +298,7 @@
                             "value": ""
                         }
                     ],
-                    "content": "If you have had, or anticipate having a significant decrease in income for {{data.calculatedTaxYear}}, you may request to be assessed using this year's estimated income.",
+                    "content": "If you have had, or anticipate having a significant decrease in income for {{data.currentTaxYear}}, you may request to be assessed using this year's estimated income.",
```
guru-aot, line 301: Please use data.programYear, or please explain why this was not used.

guru-aot (commented):
Nice work @bidyashish , please have a look on the comments.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
andrewsignori-aot: About the justification "Just in case if we need know Current tax year is fetched", when that would be a benefit since this is only used to be displayed in some labels and is derived from the program year?

sources/packages/forms/src/form-definitions/studentfinancialinformationappeal.json
andrewsignori-aot: The `data.programYear` in this form is not actually the "2024-2025" program year name?
If I am right the question from @guru-aot would not make sense and and reply from @bidyashish should probably mention it. Please let me know if I am missing something.

andrewsignori-aot (approved):
Approving assuming the "Current Tax Year" hidden variable was created in all forms using the same name to avoid confusion in the future.

guru-aot (approved):
LGTM, nice work @bidyashish

Issue #3729, opened 2024-09-23
CAS Integration 3B - Create New Site (Supplier found, Multiple Sites / No matching address)


CarlyCotton:
## **User Story:** 
As an SABC Finance team member (Ministry user), I need all students in SIMS to have a supplier and site number attached. If we receive a successful response from a Get Supplier call, we need to assign that Supplier Number to the student and either use an existing Site or create a new one. We should use an existing site if the addressline/postal code match to avoid duplication.

## **Acceptance Criteria:**
- [x] If a Get Supplier Response is Successful and returns Active Supplier and NO active Sites, **Create New Site**
- [x] If a Get Supplier Response is Successful and returns Active Supplier and ONE OR MANY active Site, Compare the Response AddressLine1 and Postal Code to SIMS address (Compare against the Addressline and Postalcode as we would be submitting and NOT case sensitive. Truncated and translated / No space in postal code):
    - [x] Match FOUND: Save/Use that site. 
    - [x] NO Match: **Create New Site**

### **Create New Site**
- [x] Only use the **Create New Site** if Get Supplier Response is Successful and:
    - No Sites Returned (0 count/not found)
    - No Sites matching address are found
- [x] Fields to include (See Field Requirements File below): SupplierNumber, AddressLine1, City, Province, Country, PostalCode, EmailAddress (Optional/Conditional fields can be left out or left null)
    - [x] **AddressLine1:** Use first addressline from SIMS. Truncate if ours is longer than supported. Convert or omit special characters (like is done with other integrations).
    - [x] **City:** Use as-is
    - [x] **Province:** Use abbreviated versions AB,BC,MB,NB,NL,NS,NT,NU,ON,PE,QC,YT
    - [x] **Postal Code:** ALL CAPS, no space.
    - [x] **Email Address:** Use as-is
- [x] If successful response to Create New Site, then update SIMS CAS table with the prior **Get Supplier** returned `supplierNumber`, `SiteCode` and any other required info and set `isValid`=true
- [x] If error response, retry for now. Manual intervention may be required and will be dealt with on ticket #3258

**Business Context**
[...]

PR #3804, opened 2024-10-17
#3260 - CAS Integration 3A - Create new Supplier and Site - CAS API Call

andrewsignori-aot:
## Main Changes

- The current process is now segregated into different classes (called processors) that are responsible for taking care of the below scenarios.
  - `CASActiveSupplierNotFoundProcessor`: 🆕
    - Supplier does not exist on CAS;
    - CAS API should be called to create supplier and site;
    - Student CAS supplier should be updated.
  - `CASPreValidationsProcessor`: 🆕
    - Given name or not a Canadian address.
    - Update the student supplier for manual intervention.
  - CASActiveSupplierFoundProcessor: (🆕 refactored)
    - Supplier found on CAS
    - Student CAS supplier should be updated. 

### Address match
- Included the address match logic in the evaluation processes following the logic already decided in #3729. If an active address with a valid match is found it will be persisted and the CAS supplier will be considered valid. If the address does not exist or does not match, the supplier will be saved but the CAS record will be considered invalid.

### Error handling
API call properly reports errors using HTTP Status 400 and adding meaningful error messages. Please see below samples.
These error messages were used to define the max length validation for the submitted data.
- [0069] Length of field SupplierName exceeded. Maximum length is 80.
- [0034] SIN is already in use.
- [0069] Length of field AddressLine1 exceeded. Maximum length is 35.

### GetSupplier 404
The `GetSupplier` method from CAS API does not return a 404 error when no supplier is found as stated in the documentation.
The method seems to be implemented as a search and it will just return an empty array if no items were found.
```json
{
    "items": [],
    "hasMore": false,
    "limit": 0,
    "offset": 0,
    "count": 0,
}
```

### Secondary PR changes
- Created a small helper to identify if the student's address is from Canada. The idea is to avoid confusion about which property to use and how the value is set on DB. This is the reason the method receives the `AddressInfo` itself which contains `country` and `selectedCountry` properties.
- New CAS mocks added to allow the CAS responses mocks.

### New E2E Tests

- Should update CAS supplier table to manual intervention when student does not have a first name.
- Should update CAS supplier table to manual intervention when student address is not from Canada.                                                
- Should create a new supplier and site on CAS and update CAS suppliers table when the student was not found on CAS and the request to create the supplier and site was successful.

### Log samples

```console
[2024-10-17 22:53:31]--INFO: Log details
[2024-10-17 22:53:31]----INFO: Log details
[2024-10-17 22:53:31]----INFO: Logging on CAS.
[2024-10-17 22:53:31]----INFO: Logon successful.
[2024-10-17 22:53:31]------INFO: Log details
[2024-10-17 22:53:31]------INFO: Processing student CAS supplier ID: 69.
[2024-10-17 22:53:31]------INFO: CAS evaluation result status: NotFound.
[2024-10-17 22:53:31]------INFO: No active CAS supplier found. Reason: Supplier not found.
[2024-10-17 22:53:31]------INFO: Created supplier and site on CAS.
[2024-10-17 22:53:31]------INFO: Updated CAS supplier and site for the student.
[2024-10-17 22:53:31]INFO: CAS supplier integration executed.
```
```console
[2024-10-17 23:04:38]INFO: Log details
[2024-10-17 23:04:38]INFO: Executing CAS supplier integration...
[2024-10-17 23:04:38]INFO: Found 1 records to be updated.
[2024-10-17 23:04:38]--INFO: Log details
[2024-10-17 23:04:38]----INFO: Log details
[2024-10-17 23:04:38]----INFO: Logging on CAS.
[2024-10-17 23:04:38]----INFO: Logon successful.
[2024-10-17 23:04:38]------INFO: Log details
[2024-10-17 23:04:38]------INFO: Processing student CAS supplier ID: 86.
[2024-10-17 23:04:38]------INFO: CAS evaluation result status: PreValidationsFailed.
[2024-10-17 23:04:38]------WARN: Not possible to retrieve CAS supplier information because some pre-validations were not fulfilled. Reason(s): Non-Canadian address.
[2024-10-17 23:04:38]------INFO: Updated CAS supplier for manual intervention.
[2024-10-17 23:04:38]INFO: CAS supplier integration executed.
```
```console
[2024-10-17 22:55:57]INFO: Log details
[2024-10-17 22:55:57]INFO: Executing CAS supplier integration...
[2024-10-17 22:55:57]INFO: Found 1 records to be updated.
[2024-10-17 22:55:57]--INFO: Log details
[2024-10-17 22:55:57]----INFO: Log details
[2024-10-17 22:55:57]----INFO: Logging on CAS.
[2024-10-17 22:55:57]----INFO: Logon successful.
[2024-10-17 22:55:57]------INFO: Log details
[2024-10-17 22:55:57]------INFO: Processing student CAS supplier ID: 74.
[2024-10-17 22:55:57]------INFO: CAS evaluation result status: ActiveSupplierFound.
[2024-10-17 22:55:57]------INFO: Active CAS supplier found.
[2024-10-17 22:55:57]------INFO: Updated CAS supplier for the student.
[2024-10-17 22:55:57]INFO: CAS supplier integration executed.
```

### TODO
- Adjust the scheduler cron expression.
- Add more E2E tests.
- Run in parallel.
- If some error is present throw in the scheduler to allow it to be retried.


sources/packages/backend/libs/integrations/src/cas/models/cas-service.model.ts
andrewsignori-aot, line 1: Changes related to this PR start at line 50.
The file was renamed.

sources/packages/backend/libs/integrations/src/cas/cas-formaters.ts
sh16011993, line 1: Minor: Should it be `formatters`?

sources/packages/backend/libs/integrations/src/cas/cas-formaters.ts
```diff
@@ -0,0 +1,51 @@
...
+  ).toUpperCase();
+}
+
+/**
+ * Ensure city has the max length expected by CAS.
+ * @param address city to be formatted.
+ * @returns formatted city.
+ */
+export function formatCity(city: string): string {
+  return city.substring(0, CAS_CITY_MAX_LENGTH);
```
sh16011993, line 41: Is this not expected to be passed as uppercase?

andrewsignori-aot: The AC states "City: Use as-is" hence not a requirement.

sources/packages/backend/libs/integrations/src/cas/cas-formaters.ts
andrewsignori-aot: Renamed.

sources/packages/backend/libs/utilities/src/address-utils.ts
```diff
@@ -0,0 +1,17 @@
...
+// when 'Canada' is selected.
+export const COUNTRY_CANADA = "canada";
+
+/**
+ * Inspects the address info to determine if the address is from Canada.
+ * @param address address info to be inspected.
+ * @returns true if the address is from Canada, otherwise false.
+ */
+export function isAddressFromCanada(address: AddressInfo) {
+  return address.selectedCountry === COUNTRY_CANADA;
```
sh16011993, line 16: We are using just the `selectedCountry` to determine if the address is from Canada. Do we need the whole object passed here?

andrewsignori-aot: From the PR description. Please let me know if it is not clear
image

sources/packages/backend/libs/integrations/src/cas/models/cas-service.model.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Data used during the creation of a supplier and site on CAS.
+ * Some data transformation is needed to follow the CAS requirements.
+ * This payload is used to submit the data and also to be returned to
+ * the consumer, allowing it to be aware of the data actually submitted.
+ */
+export class CreateSupplierAndSiteSubmittedData {
+  SupplierName: string;
```
sh16011993, line 86: Just for my understanding and knowledge, why do the keys for the object start with Uppercase? Any specific reason @andrewsignori-aot that they (CAS) might be doing this?

andrewsignori-aot: I am not sure about the reason. It is not consistent in the docs either.
```json
{
  "Supplier_Number":"123456789",
  "Supplier_Site_Code":"001",
  "CAS-Returned-Messages":"SUCCESS"
}
```
vs
```json
{
  "SupplierName":"EXAMPLE, SUPPLIER 1",
  "SubCategory":"Individual",
  "Sin":"",
  "BusinessNumber":""
}
```
vs
```json
{
  "supplierprotected": null,
  "standardindustryclassification": null,
  "lastupdated": "2021-06-08 08:44:41",
}
```

sources/packages/backend/libs/utilities/src/address-utils.ts
```diff
@@ -0,0 +1,17 @@
+import { AddressInfo } from "@sims/sims-db";
+
+// 'selectedCountry' in the student contact info will have the value 'other',
+// when 'Other'(i.e country other than canada) is selected.
+export const OTHER_COUNTRY = "other";
+// 'selectedCountry' in the student contact info will have the value 'canada',
+// when 'Canada' is selected.
+export const COUNTRY_CANADA = "canada";
```
sh16011993, line 8: They are both essentially negations of each other. Do we need both of them? From what I could understand, just one is enough to resolve all the logic conditions. Please let me know if I am missing something.

sources/packages/backend/libs/integrations/src/cas/cas-formatters.ts
sh16011993, line 1: 👍

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -57,13 +69,15 @@ export class CASSupplierIntegrationService {
...
 
   /**
-   * For each pending CAS supplier, request supplier information to CAS API and update local table.
+   * For each pending CAS supplier, evaluate student current data and decide
+   * how to proceed to ensure student will have a supplier number and site
+   * code associated.
    * @param casSuppliers pending CAS suppliers.
    * @param parentProcessSummary parent log summary.
    * @param auth CAS auth details.
    * @returns true if updated a record.
```
sh16011993, line 78: returns the suppliers count?

sources/packages/backend/libs/utilities/src/address-utils.ts
andrewsignori-aot: These constants are not new and were just moved from one lib to another.
They can potentially become one but I would not see additional refactoring around this as part of this PR.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: Comment updated.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -38,10 +50,10 @@ export class CASSupplierIntegrationService {
     const summary = new ProcessSummary();
     parentProcessSummary.children(summary);
     try {
-      summary.info("Logging on CAS...");
+      summary.info("Logging on CAS.");
       const auth = await this.casService.logon();
       summary.info("Logon successful.");
-      suppliersUpdated = await this.requestCASAndUpdateSuppliers(
+      suppliersUpdated = await this.processSuppliers(
```
dheepak-aot, line 56: The silent failing is also happening inside `this.processSuppliers`.

Does this method need to fail silently w/o throwing error?

sources/packages/backend/libs/utilities/src/address-utils.ts
sh16011993: Apologies, missed it. Thanks.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -38,10 +50,10 @@ export class CASSupplierIntegrationService {
     const summary = new ProcessSummary();
     parentProcessSummary.children(summary);
     try {
-      summary.info("Logging on CAS...");
+      summary.info("Logging on CAS.");
       const auth = await this.casService.logon();
```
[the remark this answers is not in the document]
andrewsignori-aot: As per the docs, 1 hour (3600 seconds).

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -72,137 +86,129 @@ export class CASSupplierIntegrationService {
...
+      return {
+        status: CASEvaluationStatus.PreValidationsFailed,
+        reasons: preValidationsFailedReasons,
+      };
+    }
+    const supplierResponse = await this.casService.getSupplierInfoFromCAS(
+      auth.access_token,
+      casSupplier.student.sinValidation.sin,
+      casSupplier.student.user.lastName,
     );
```
dheepak-aot, line 173: I would recommend, logon and getSupplier call happening here instead upfront. May be a private method which does both together could be better,

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: The below AC is not implemented yet.
"If some error is present throw in the scheduler to allow it to be retried."
The idea is to inspect the summary (as it is already done in `getSuccessMessageWithAttentionCheck`) and throw an error if any error log message is present. Throwing an exception from here or from there would result in the same.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: The logon may not be used exclusively for the `getSupplierInfoFromCAS` and `getSupplierInfoFromCAS` may not even be the first method to be called in the future.
To be clear, is the recommendation to execute the `logon` inside the `evaluateCASSupplier`?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-evaluation-result-processor.ts
```diff
@@ -0,0 +1,29 @@
...
+   * @param evaluationResult evaluation result to be processed.
+   * @param auth authentication token needed for possible
+   * CAS API interactions.
+   * @param summary current process log.
+   * @returns processor result.
+   */
+  abstract process(
+    casSupplier: CASSupplier,
+    evaluationResult: CASEvaluationResult,
+    auth: CASAuthDetails,
```
dheepak-aot, line 26: If the logon can happen during the time of API call(create supplier or create site), then we can remove from here.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param evaluationResult evaluation result to be processed.
+   * @param _auth authentication token needed for possible
+   * CAS API interactions.
+   * @param summary current process log.
+   * @returns processor result.
+   */
+  async process(
+    casSupplier: CASSupplier,
+    evaluationResult: CASEvaluationResult,
+    _auth: CASAuthDetails,
```
dheepak-aot, line 40: If we are pursuing the auth, shouldn't it not be optional as signature?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
sh16011993: I think the idea is that the failure will be caught and then logged for info purpose as `summary.error` and then in the next run of the scheduler, the failed cas suppliers will be attempted again.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -72,137 +86,129 @@ export class CASSupplierIntegrationService {
...
+        return (
+          address.status === "ACTIVE" &&
+          address.addressline1 === casFormattedStudentAddress &&
+          address.postalcode === casFormattedPostalCode
+        );
+      });
+    return {
+      status: CASEvaluationStatus.ActiveSupplierFound,
+      activeSupplier: casResponseActiveSupplier,
+      matchedAddress: casResponseMatchedAddress,
```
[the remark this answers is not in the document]
andrewsignori-aot: That was the initial approach which would require one more processor.
I ended up reverting it because these two pieces of information are tightly coupled.
I will give it a second try in preparation for the upcoming address site creation.
- Active supplier and site found: `update supplier and site`.
- Active supplier found with no address match: create the site, and `update supplier and site`.
The `create the site` will happen inside a new processor that will call a shared method to then `update supplier and site`.
Does it sound ok?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
andrewsignori-aot: It does not matter if optional or not because the consumers must provide all parameters either way for all the processors.
If it is to make it clear to the developer looking into the processor, the "_" will achieve the same.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-evaluation-result-processor.ts
andrewsignori-aot: The idea in place was to do the logon once and take advantage of it lasts for 1h and be shared.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: If some error is logged to the summary we will throw to allow the scheduler to retry immediately.
When we have the permanent errors implemented, those should be probably logged as warnings because they will be part of the "business valid scenarios".

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -72,137 +86,129 @@ export class CASSupplierIntegrationService {
...
           suppliersUpdated++;
         }
       } catch (error: unknown) {
-        summary.error(
-          "Error while requesting or updating CAS suppliers.",
-          error,
-        );
+        // Log the error and allow the process to continue checking the
+        // remaining student suppliers.
+        summary.error("Unexpected error while processing supplier.", error);
```
dheepak-aot, line 114: Should not fail silently here. I believe only `getCASSupplierProcess` and `evaluateCASSupplier` can silently fail.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -1,26 +1,38 @@
...
+  CASEvaluationResultProcessor,
+} from "./cas-evaluation-result-processor";
 
 @Injectable()
 export class CASSupplierIntegrationService {
   constructor(
     private readonly casService: CASService,
-    private readonly systemUsersService: SystemUsersService,
     @InjectRepository(CASSupplier)
     private readonly casSupplierRepo: Repository<CASSupplier>,
```
sh16011993, line 32: May not be a part of this PR but a question here: Can we not extend this class from `RecordDataModelService<T>` and pass the `repo` to the superclass and initialize it that way?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: `summary.error` will not fail silently in upcoming PR.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -72,137 +86,129 @@ export class CASSupplierIntegrationService {
...
   }
 
   /**
-   * Updates supplier if finds an item from the response with an active address.
-   * @param supplierResponse CAS supplier response.
-   * @param summary log summary.
-   * @param casSuppliers pending CAS suppliers.
-   * @returns true if updated a record.
+   * Get the processor associated to the CAS evaluation status result.
+   * @param evaluationResult evaluation result status.
```
sh16011993, line 122: I believe the `param` needs an update here.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: `RecordDataModelService` is no longer recommended and should not be used.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
```diff
@@ -0,0 +1,118 @@
...
+   * Create the new supplier and site on CAS using the student information.
+   * @param casSupplier student supplier information from SIMS.
+   * @param evaluationResult evaluation result to be processed.
+   * @param auth authentication token needed for possible
+   * CAS API interactions.
+   * @param summary current process log.
+   * @returns processor result.
+   */
+  async process(
+    casSupplier: CASSupplier,
```
dheepak-aot, line 42: Can we just have the supplier id here? So that the processor(s) which needs more details than id can query and get it.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(
+    private readonly systemUsersService: SystemUsersService,
+    @InjectRepository(CASSupplier)
+    private readonly casSupplierRepo: Repository<CASSupplier>,
+  ) {
+    super();
+  }
+
+  /**
+   * Process the result of a pre-validation failed, setting
```
sh16011993, line 28: `failed pre-validation` ?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
sh16011993, line 40: For my understanding, what does this `_` in `_auth` represent? Does it represent that the particular parameter is not being used in the method but is being passed because it is an implementation of the abstract class?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-found-processor.ts
```diff
@@ -0,0 +1,103 @@
...
+          supplierAddress: supplierAddressToUpdate,
+          supplierStatus: SupplierStatus.Verified,
+          supplierStatusUpdatedOn: now,
+          isValid: !!supplierAddressToUpdate,
+          updatedAt: now,
+          modifier: systemUser,
+        },
+      );
+      if (updateResult.affected) {
+        summary.info("Updated CAS supplier for the student.");
```
dheepak-aot, line 89: Will it be good to identify the student with student id?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
```diff
@@ -0,0 +1,81 @@
...
+          updatedAt: now,
+          modifier: systemUser,
+        },
+      );
+      if (updateResult.affected) {
+        summary.info("Updated CAS supplier for manual intervention.");
+        return { isSupplierUpdated: true };
+      }
+      summary.error(
+        "The update of the CAS supplier for manual intervention did not result in the expected affected rows number.",
```
sh16011993, line 71: I am not able to understand this particular error message. Can you please explain this briefly @andrewsignori-aot . Thank You.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
andrewsignori-aot: I am not sure if I am following. The `casSupplier.id` is present and loaded.
`CASSupplier` as it is right is the full `CASSupplier` object returned from the "get pending suppliers query".
No specific interface was created for this one. Unless some fresh data is needed, all data required to execute the processes should be loaded in advance using the "get pending suppliers query".

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
andrewsignori-aot: It removes the warning that a parameter is declared but not used.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
```diff
@@ -0,0 +1,118 @@
...
+          supplierAddress: {
+            supplierSiteCode: result.response.supplierSiteCode,
+            addressLine1: submittedAddress.AddressLine1,
+            city: submittedAddress.City,
+            provinceState: submittedAddress.Province,
+            country: submittedAddress.Country,
+            postalCode: submittedAddress.PostalCode,
+            status: "ACTIVE",
+            lastUpdated: now,
+          },
```
dheepak-aot, line 95: what about `supplierProtected` and `siteProtected` in the update?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-found-processor.ts
andrewsignori-aot: The supplier id (which) will be related to a student is logged once to the summary to avoid repeating it.
Since the summary is grouped I do not see a need to log it again. It may need a "correlation id" for the logs, but if the id is required here it would be required in all other logs. Does it make sense?

image

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+    try {
+      const [submittedAddress] = result.submittedData.SupplierAddress;
+      const now = new Date();
+      const systemUser = this.systemUsersService.systemUser;
+      const updateResult = await this.casSupplierRepo.update(
+        {
+          id: casSupplier.id,
+        },
+        {
```
dheepak-aot, line 81: The update happening for found and not found is identical. Can we have a common method to be re-used for the update?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
```diff
@@ -0,0 +1,118 @@
...
+    evaluationResult: CASEvaluationResult,
+    auth: CASAuthDetails,
+    summary: ProcessSummary,
+  ): Promise<ProcessorResult> {
+    if (evaluationResult.status !== CASEvaluationStatus.NotFound) {
+      throw new Error("Incorrect CAS evaluation result processor selected.");
+    }
+    summary.info(
+      `No active CAS supplier found. Reason: ${evaluationResult.reason}.`,
+    );
```
sh16011993, line 52: 👍 for this info in all the processors.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
andrewsignori-aot: If the update was executed with no rows updated something went wrong.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
andrewsignori-aot: Do you mean setting them explicitly to `null`?
Since the supplier was not found, it will be a new supplier and site. These properties should never exist or be created for a new supplier and site.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
andrewsignori-aot: Do you mean the below part is identical? I would prefer to leave as it is.

image

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
dheepak-aot: To be clear, is the recommendation to execute the logon inside the evaluateCASSupplier? - Yes.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-evaluation-result-processor.ts
dheepak-aot: What about getting the auth from a shared service then?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
dheepak-aot: yes the id is already present in `CASSupplier` object. My point was the param to be just id not the whole object. 

I see 2 processors using only id and one processor using the whole object. So for the processor(s) which need more details can do a get query to find the object by  id.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
dheepak-aot: These properties should never exist or be created for a new supplier and site. - Ah, is it? I was under the impression that it is set during new supplier and site creation.

dheepak-aot:
Great Job @andrewsignori-aot. Thanks for trying it make it streamlined for any upcoming CAS updates.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
andrewsignori-aot: They are legacy properties. As far as I know, they only exist due to some SFAS requirement.
Either way, we do not have any direction about setting them.

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -82,6 +97,73 @@ export class CASService {
     return response?.data;
   }
 
+  /**
+   * Create supplier and site.
+   * @param token authentication token.
+   * @param supplierData data to be used for supplier and site creation.
+   * @returns submitted data and CAS response.
+   */
+  async createSupplierAndSite(
```
sh16011993, line 106: Can this method be in `cas-supplier.service.ts`?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: The current logon process (not part of this PR) is intended to be executed once and reused where it is needed.
We can move it inside the service and ensure that it will be reused from there but I would consider an extra refactor (that I would be willing to do in a separate PR if time allows it).
Moving it to the `evaluateCASSupplier` does not seem right. `evaluateCASSupplier` may not even be the first method to execute the login and it does not seem its responsibility.
For instance, wonder if we have a pending address to be verified and it just needs to call CAS and create a new site. This may not require a call to CAS from the `evaluateCASSupplier`.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-evaluation-result-processor.ts
andrewsignori-aot: > What about getting the auth from a shared service then?

Please see the answer here, the logon was in place and it was just resued. I would be willing to change it, but I do not see how could it be part of the same PR.
The current effort is already putting effort in a good amount of refactoring. It time allows I will do some refactor, not the ideal one but some improvement to avoid passing it as a parameter.
https://github.com/bcgov/SIMS/pull/3804#discussion_r1809328756

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
andrewsignori-aot: I am not sure if I am following it. This is the CAS API call to `createSupplierAndSite` and it is in the CAS service responsible for making the API calls. What would be the benefit of moving it?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-found-processor.ts
```diff
@@ -0,0 +1,103 @@
...
+          id: casSupplier.id,
+        },
+        {
+          supplierNumber: supplierToUpdate.suppliernumber,
+          supplierName: supplierToUpdate.suppliername,
+          status: supplierToUpdate.status,
+          supplierProtected: supplierToUpdate.supplierprotected === "Y",
+          lastUpdated: new Date(supplierToUpdate.lastupdated),
+          supplierAddress: supplierAddressToUpdate,
+          supplierStatus: SupplierStatus.Verified,
```
sh16011993, line 81: create site api call that needs to be implemented later will be created using a processor under the same umbrella of processors, correct? If that is the case, then along with `isValid` to false, don't we also need to set this to `PendingAddressVerification` or some `Pending` status so that it can be picked up again in future as a CAS supplier record for which the site information needs to be created.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/_tests_/cas-supplier-integration.scheduler.e2e-spec.ts
```diff
@@ -121,4 +161,218 @@ describe(describeProcessorRootTest(QueueNames.CASSupplierIntegration), () => {
...
+    });
+    expect(updateCASSupplier).toEqual({
+      id: savedCASSupplier.id,
+      isValid: false,
+      supplierStatus: SupplierStatus.ManualIntervention,
+      updatedAt: expect.any(Date),
+      modifier: { id: systemUsersService.systemUser.id },
+    });
+    // Ensure updatedAt was updated.
+    expect(updateCASSupplier.updatedAt.getTime()).toBeGreaterThan(
```
sh16011993, line 220: 👍

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/cas-integration/_tests_/cas-supplier-integration.scheduler.e2e-spec.ts
```diff
@@ -121,4 +161,218 @@ describe(describeProcessorRootTest(QueueNames.CASSupplierIntegration), () => {
...
+  });
+
+  it(
+    "Should create a new supplier and site on CAS and update CAS suppliers table when " +
+      "the student was not found on CAS and the request to create the supplier and site was successful.",
+    async () => {
+      // Arrange
+      const referenceDate = new Date();
+      const savedCASSupplier = await saveFakeCASSupplier(db);
+      // Configure CAS mock to return em empty result for the GetSupplier
```
sh16011993, line 286: Minor: `an`

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
sh16011993: I got the point now. The idea is to have a service that has all the API calls that we make to CAS. This is that service. Thanks.

sh16011993 (approved):
Great work @andrewsignori-aot 👍  Thank You for the walkthrough of your PR.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-found-processor.ts
andrewsignori-aot: `PendingAddressVerification` is intended to be used when the SIMS student address changes.
For now, there are no requirements to set it as pending since there is no process to check it.

sh16011993: ok

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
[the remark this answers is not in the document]
andrewsignori-aot: Created a new status and processor. The method for `update supplier and site` was kept inside each processor.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: Updated.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-pre-validations-processor.ts
andrewsignori-aot: Changed.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-evaluation-result-processor/cas-active-supplier-not-found-processor.ts
dheepak-aot: My point was conversion of CAS Supplier and CAS Site to SIMS Supplier and SIMS supplier address(and then update). May be something in one of the upcoming CAS tickets.

dheepak-aot (approved):
Great work. Thanks for doing the changes. 👍

Issue #3812, opened 2024-10-21
BUG - Content displayed twice


CarlyCotton:
**Description**
Post upgrades there are a few odd content items that have been noticed. They used to show just once and now the content is duplicated.

**Acceptance Criteria** 
- [ ] Correct the content display to show it only once:
    - [ ] Welcome page of application
    - [ ] File Uploaders (Looks like every one is duplicating the content)

**Screenshots**
<img width="539" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNDFmQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--94aa7d0cc4a005e430d3a7a8a7efc02b3266613c/image.png" alt="image.png" />

<img width="1141" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGhmQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4bc614fe2f3bc7517ce52db9bd93ac0bdb7ab1c4/image.png" alt="image.png" />
<img width="861" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVZmQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--068b1f3f8259fab490949b2a276d3655b8292824/image.png" alt="image.png" />

**Technical**
- [ ] Possible solution to add the HTML Element Tag as `div` as shown below.
  - [ ] Investigate the root cause further.
- [ ] Check all HTML Element inside the Student Application.
<img width="1160" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkJsQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4558b49c9a523b24f09f00dfd8db9b08ef34973b/image.png" alt="image.png" />

PR #3828, opened 2024-10-23
#3812 - BUG - Content displayed twice

andrepestana-aot:
The issue happened because the markup in the HTML component is invalid.  The "p" tag could not be used with children elements like: address, article, aside, blockquote, details, dialog, div, dl, fieldset, figcaption, figure, footer, form, h1, h2, h3, h4, h5, h6, header, hgroup, hr, main, menu, nav, ol, p, pre, search, section, table, or ul.

Form.io changed its default parent tag from p to div: https://github.com/formio/formio.js/pull/5792

Similar issue and developer explanation/recommendation: https://github.com/formio/formio.js/issues/5598#issuecomment-2427097530

Found some other issues not reported in the ticket as this one:
image

- Checked all form.io forms;
- Changed the "p" tag to "div" when the component had invalid children tags for paragraph;
- Changed some other invalid HTML found when doing the check.


sources/packages/forms/src/form-definitions/studentaccountactivity.json
```diff
@@ -357,7 +357,7 @@
                       "properties": {},
                       "allowMultipleMasks": false,
                       "addons": [],
-                      "tag": "p",
+                      "tag": "li",
```
andrepestana-aot, line 360: This one didn't have any issue with the rendering but I changed to make it a valid HTML.

lewischen-aot (approved):
Looks great to me 👍 Great work @andrepestana-aot

andrewsignori-aot (approved):
Great changes and investigation, looks good 👍

Issue #3620, opened 2024-08-08
Change Request - Partner Current Year Income (financial information) option


Joshua-Lakusta:
As a student I want to request the use of my partner/spouses current year income. This is important as it is a calculation impacting field and the ability to request the use of estimated current year income is a policy requirement.

### Acceptance Criteria
- [x] Add a new question to the  **Partner Information and Income** to allow students to submit partner income for the current year. This question is only relevant if they have selected Married/common-law

- [x] New question will have 2 radio options yes or no (this question will appear below the basic income declaration question for a partner.
**Content**:
> Do you want to declare your partner's estimated current year income? (If your spouse/common-law partner has had, or anticipates having a significant decrease in income for "YEAR", you may request to be assessed using your spouse/common-law partner's estimated income.)

- [x] Replace "YEAR" with dynamic outcome with following logic:

> YEAR = (dynamic year = to first year in program year (2023-2024 would be 2023))

- [x] If no is selected do nothing additional

- [x] If yes is selected expand to add an estimated income field, reasons for use of estimated income, and document uploader (this mirrors the way the student declares estimated income)

- [x] **Under Supporting Information Header the content should be:** 

> Add your spouse/common-law partner's actual income from the beginning of the calendar year up to the date of their income change and the estimated income for the remainder of that calendar year to calculate your total current year income.

- [x] Add income entry field that will represent a new value to be created and named by dev doing the ticket
- [x] Income field should follow the same rules as the estimated student income version

- [x] **Heading above the income entry field should state:**

> My spouse/common-law partner's current year income is:

- [x] There needs to be a section for the reasons for change in estimated income with five options and only one can be selected

**Content:**

> Select the reason for your spouse/common-law partner's significant decrease in income:
> - Medical illness or injury
> - Family emergency (e.g., death or injury)
> - Natural Disaster
> - Layoff, strike, lockout or other reduction in earnings beyond the family's control
> - Other

- [x] Add content for uploader content for documentation for uploader:

> Upload supporting documents for spouse/common-law partner's "DYNAMIC YEAR" estimated income.
> Please provide supporting documentation for your request to use current year estimated income for your spouse/common law partner.
> Provide the following:
> 
> A letter explaining the circumstances that meet one or more of the Criteria (see above)
> Provide at least one of the following:
> 
> All relevant supporting documentation of the circumstance(s) (medical records, letter from the physician, record of employment, layoff or strike, invoices from damages, etc.).
[...]

PR #3834, opened 2024-10-24
#3620 - Change Request - Partner Current Year Income (financial information) option (Fix Duplicate Markup)

lewischen-aot:
Fixed duplicates of the markup content in the Upload Document panel in the Form.io file.

Screenshot of the markup content before the fix
image

Screenshot of the markup content after the fix
image


dheepak-aot (approved):
Looks Good 👍

PR #3837, opened 2024-10-24
#3260 - CAS Integration 3A - Create new Supplier and Site - Process in Parallel

andrewsignori-aot:
- Adjusted supplier's processes to happen in parallel.
- Reduced the amount of allowed parallel processes to 2 to 4. Previously it was using the "os cores" as the default value which could potentially be up to 32 while running on Openshift.

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
@@ -8,12 +8,7 @@ import {
   formatAddress,
   formatPostalCode,
 } from "@sims/integrations/cas";
-import {
-  CustomNamedError,
-  isAddressFromCanada,
-  ParallelIntensity,
-  processInParallel,
-} from "@sims/utilities";
+import { CustomNamedError, isAddressFromCanada } from "@sims/utilities";
 import { CAS_AUTH_ERROR } from "@sims/integrations/constants";
 import {
   CASEvaluationResult,
@@ -28,7 +23,6 @@ import {
   CASActiveSupplierFoundProcessor,
   CASEvaluationResultProcessor,
   CASActiveSupplierAndSiteFoundProcessor,
-  ProcessorResult,
 } from "./cas-evaluation-result-processor";
 
 @Injectable()
@@ -54,9 +48,6 @@ export class CASSupplierIntegrationService {
     parentProcessSummary: ProcessSummary,
     studentSuppliers: StudentSupplierToProcess[],
   ): Promise<number> {
-    // Force the CAS token to be acquired before starting the process.
-    await this.casService.getToken();
-    // Process each supplier in parallel.
     let suppliersUpdated = 0;
     const summary = new ProcessSummary();
     parentProcessSummary.children(summary);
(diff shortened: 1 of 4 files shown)


andrepestana-aot:
> * Adjusted supplier's processes to happen in parallel.
> * Reduced the amount of allowed parallel processes to 2 to 4.

" Increased the amount of allowed parallel processes to 2 to 4."

sources/packages/backend/libs/utilities/src/parallel-processing-utils.ts
```diff
@@ -1,4 +1,19 @@
-import * as os from "os";
+/**
+ * Number of parallel processes allowed to be started at same time.
+ */
+export enum ParallelIntensity {
```
andrepestana-aot, line 4: 👍

andrepestana-aot (approved):
Looks good! Great job!

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
```diff
@@ -48,6 +54,9 @@ export class CASSupplierIntegrationService {
     parentProcessSummary: ProcessSummary,
     studentSuppliers: StudentSupplierToProcess[],
   ): Promise<number> {
+    // Force the CAS token to be acquired before starting the process.
+    await this.casService.getToken();
```
dheepak-aot, line 58: Great. When the get token fails, process is aborted right away.

andrewsignori-aot:
> > * Adjusted supplier's processes to happen in parallel.
> > * Reduced the amount of allowed parallel processes to 2 to 4.
> 
> " Increased the amount of allowed parallel processes from 2 to 4."

May comment was not clear, I meant that the default value was decreased from the "os.cores" amount (potentially 32 on Openshift) to 2 or 4 😉

sources/packages/backend/libs/utilities/src/parallel-processing-utils.ts
```diff
@@ -1,4 +1,19 @@
...
+  /**
+   * Used for regular processes that can be executed in parallel where the
+   * waiting time is expected to be low, for instance, for regular quick
+   * DB access.
+   */
+  Regular = 2,
```
dheepak-aot, line 10: It should be the other way round right for parallel processing?

I mean for short lived processes more accepted parallel threads and long processed with less parallel threads.

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
dheepak-aot, line 58: Minor and not a blocker, the method we are looking here is like `initiToken` right?

dheepak-aot:
Thank you very much for making the change. minor comments and a question.

sources/packages/backend/libs/utilities/src/parallel-processing-utils.ts
andrewsignori-aot: Longer processes (I/O processes) will leave the processor free to take more "processing" work, right?

sources/packages/backend/apps/queue-consumers/src/services/cas-supplier/cas-supplier.service.ts
andrewsignori-aot: This is the same method used internally by the method to get a token (request one or cached one).
There is no guarantee that this method will be the one initializing the token, for instance, if the scheduler is retrying due to some individual student error the token will still be cached.

The method getToken can be kept private (as it was before) and we can create something like `ensureTokenInitialization`.
I reused the same method for simplicity. Please let me know if it would be a blocker or if it requires a quick chat 😉

sources/packages/backend/libs/utilities/src/parallel-processing-utils.ts
andrewsignori-aot: Comments adjusted 😉 
Thanks for the nice conversation.

dheepak-aot (approved):
Thanks for making the change and nice conversation. Looks good 👍

Issue #3833, opened 2024-10-24
SIN Validation - Gender Field Limit 


ninosamson:
**Describe the Bug**
When creating SIN Validation request, the process is erroring out because the gender field length for "prefer not to answer" exceeds the column limit. 

**Expected Behaviour**
SIN validation should be able to handle all gender types. 

**Actual Behaviour**
The SIN validation process fails due to "prefer not to answer" 

- [ ] Increase the column length from 10 to 50 on SIN Validation to handle "prefer not to answer option" 
- [ ] validate with student profile with "prefer not to answer" as gender

PR #3839, opened 2024-10-25
#3833 - SIN Validation - Gender Field Limit

andrepestana-aot:
Altered sims.sin_validations.gender_sent to be VARCHAR(50).

Tested with a pending SIN Validation:

image

After running sin-validation-process-integration:

image

Revert: 
image

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/sql/SINValidations/Rollback-alter-column-gender-sent.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/SINValidations/Rollback-alter-column-gender-sent.sql
@@ -1,10 +1,3 @@
-UPDATE
-  sims.sin_validations
-SET
-  gender_sent = NULL
-WHERE
-  LENGTH(gender_sent) > 10;
-
 ALTER TABLE
   sims.sin_validations
 ALTER COLUMN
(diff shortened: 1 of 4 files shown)


guru-aot (approved):

sources/packages/backend/apps/db-migrations/src/sql/SINValidations/Rollback-alter-column-gender-sent.sql
```diff
@@ -0,0 +1,11 @@
+UPDATE
+  sims.sin_validations
+SET
+  gender_sent = NULL
+WHERE
+  LENGTH(gender_sent) > 10;
```
dheepak-aot, line 6: Is it desired for rollback ? IMO, we must just do the alter statement.

If at all there is any need to rollback, we can handle it manually. 

@andrewsignori-aot  ?

andrewsignori-aot: I would agree with @dheepak-aot. No values greater than 10 will be expected in any DB since they are not supported. The rollback script, if executed, should be executed at release time.

dheepak-aot (approved):
Thanks for making the change. Looks good 👍

Issue #3774, opened 2024-10-08
SFAS-SIMS Bridge File - MSFAA Implementation


CarlyCotton:
**User Story**
As a student, it is possible for me to have two MSFAA numbers (PT and FT). I don't want to have a new one issued by SIMS when I already have a valid one in SFAS/SAIL. 

**Acceptance Criteria**
- [x] On application submission during "Associate MSFAA" in the gateway, add another way to validate MSFAA (this can be done the same for PT/FT)
- [x] If a student does not have a SIGNED SIMS generated MSFAA number, search Bridge file Individual data records to find matching (SIN/DOB/LastName) student profile
    - [x] If no match found, Create New MSFAA as normal (or use the unsigned SIMS MSFAA number).
    - [x] If the most recent match found **DOES NOT** contain a MSFAA number for the intensity, Create New MSFAA as normal.
    - [x] If the most recent match found **DOES** contain a MSFAA number for the intensity, search the bridge file application data and use the most recent (or future) study end date ("300" for FT or "301" for PT) to ensure it is within 730 days of this date.  System cancel/void any other MSFAAs for that intensity. If No application or end date outside of 730 days: related to the student/MSFAA number, do not USE and create new as normal.
- [x]  In cases where we find a match in bridge file with a valid study end date: import that MSFAA number and use the most recent application end date from bridge file information as a "signed date"
 

**Business Context**
- Only VERIFIED MSFAA SAIL data will be reported. (This is a status in SAIL that is displayed based on the validity of the MSFAA number.)
- We are receiving the "Effective Date" for PT MSFAAs. As noted in the bridge file layout, this is really the only date data that we have easily accessible from SAIL for the MSFAA.  The signed date is contained within a text file in a BLOB field in another table, which is difficult to get at.  At this time, we are not going to try and get the signed date, but we will be using the effective date on this table, which is actually the date that the MSFAA request was created and should be close enough.
- We recognize that the gap between effective date and signed date could result in us declaring a MSFAA number "expired" slightly sooner than it would normally. This amounts to a small risk based on volume of students in PT with the circumstances that could end up in that situation. It results in the least number of students being impacted by having to be re-issued an MSFAA pre-maturely or worse, during an active application in the legacy systems.
- NULL date_requested/sent date would indicate that it is an imported MSFAA number

**Technical**
- [x] Create an additional process to be executed if SIMS does not have a **VALID AND SIGNED** MSFAA.
[...]

PR #3853, opened 2024-10-29
#3774 - Refactoring Import Modules

guru-aot:
Refactoring import modules in the SFASIntegrationModule.


sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.module.ts
```diff
@@ -21,21 +21,21 @@ import {
...
     SFASIntegrationService,
     SFASIntegrationProcessingService,
-    SFASIndividualService,
-    SFASApplicationService,
-    SFASRestrictionService,
-    SFASPartTimeApplicationsService,
+    SFASIndividualImportService,
+    SFASApplicationImportService,
+    SFASRestrictionImportService,
+    SFASPartTimeApplicationsImportService,
```
andrewsignori-aot, line 38: Is there a need to export the below services?
```
SFASIndividualImportService,
SFASApplicationImportService,
SFASRestrictionImportService,
SFASPartTimeApplicationsImportService,
```
The service being consumed externally is probably `SFASIntegrationProcessingService` and this would be the only one that needs to be exported. Please give it a try.

andrewsignori-aot (commented):
Thanks for the refactor, please take a look at the comments.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-integration.module.ts
guru-aot: make sense, changed now

andrewsignori-aot (approved):
Looks good 👍

andrepestana-aot (approved):
LGTM!

Issue #3840, opened 2024-10-25
Hotfix: Bulk uploads: Unexpected error while uploading the file displays when "Validate" or "Create now" are selected for valid bulk upload


astridSABC:
**Describe the Bug**
When the user has a valid .csv file, when they click "Validate" or "Create now" there is an unexpected error message displayed

**Note: Need to check a file with additional spaces or invalid information**

**Expected Behaviour**
_1. Click “Validate”_ 
(error handling)
When the file is validated, if there are any issues with the file an informative message is displayed

Example:  
 <img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBKzUyQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f9c7ec3d49008b87549dc957e732bd0eae2fbf3a/image.png" alt="image.png" />

Click "Validate" (acceptable file)
Confirmation message such as "File is validated, click "Create now" to proceed" or some user friendly message.

_2. Click "Create now"_
(error handling) - (most likely not selected if *.csv does not validate)

(acceptable file)
Confirmation message such as "Success! Your bulk offerings have been uploaded" or some user friendly message.

**Actual Behaviour**
Click "Validate" below error message is displayed

Click "Create now" below error message is displayed

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBKzkyQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2922b0afe9ac49492990ac92d521a7d5f18c0e35/image.png" alt="image.png" />

** Steps To Reproduce**
Steps to reproduce the behaviour:

**Precondition**
Note: The institution location code
A program that allows part time programs with above institution location code in the Institution Program code field and an SABC code
.xls spreadsheet has been filled out and saved as a Comma Delimited (*.csv) format

[bulk-offering-upload-template.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1IyQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4a5095b175b08a79a272023fb7c5c9c13a2d201e/bulk-offering-upload-template.xlsx)

[...]

PR #3860, opened 2024-10-30
#3840 - Hotfix: Bulk uploads: Unexpected error while uploading the file displays when "Validate" or "Create now" are selected for valid bulk upload

andrepestana-aot:
In Vuetify newer versions there's a [change on how VFileInput  deals with the model value](https://vuetifyjs.com/en/api/v-file-input/#props-model-value):
  - If `multiple` prop is undefined or false, VFileInput will return a `File`. If true, it returns a `File[]`.
 
Instead of destructuring a file from an array, the model file can be used as the blob required.


andrewsignori-aot (approved):
Great finding, looks good 👍

guru-aot (approved):
LGTM nice and quick fix @andrepestana-aot

Issue #3889, opened 2024-11-05
Increase Logging for SFTP and file handling


ninosamson:
Need to increase logging to provide more meaningful information for troubleshooting. 

- [ ] Ensure the error details for the SFTP operations are captured and logged.
- [ ] Upgrade `ssh2-sftp-client`.

PR #3901, opened 2024-11-06
#3889 - Increase Logging for SFTP and file handling - (Catch fix)

andrewsignori-aot:
Fix issues from the previous PR for better logging. The main log from the SFTP base would still work but ensuring the logs are added to the Job in some way makes it easier for troubleshooting.
The goal is to have all the methods following similar archive catch handling doing something closet to each other.

See the example below of a sample that will display errors and `causes`.
image


sh16011993 (approved):
LGTM 👍

guru-aot (approved):

dheepak-aot (approved):
Looks good 👍

Issue #3691, opened 2024-09-12
UX/UI for Ignore Restriction - Bypass Add/Remove


andrewsignori-aot:
**User Story**
As a Ministry user, I need the ability to Ignore restrictions through the interface, as well as see some limited historical details of any ignored restrictions 

**Acceptance Criteria**
- [x] Build modal for when Ministry selects "Bypass restriction" .
- [x] Build modal for when Ministry selects "View details" and the Bypass is active.
- [x] Build modal for when Ministry selects "Remove bypass".
- [x] Create role and add permissions for this feature to existing group `business-administrators`.
- [x] Applications are eligible for this ignore restriction bypass once they have been submitted and have an application number.

**Technical**
- [x] Similar modals process exists for the Ministry while adding a student restriction to the student, they can used as a reference.
- [x] "View detail" displays the "Bypass restriction" modal in the "view mode".
- [x] "Remove bypass rule?" should be enabled for active bypasses only and active student restrictions. If the student restrictions associated with the bypass gets resolved the bypass should no longer be removed.
  - [x] When enabled, display the "Bypass restriction" modal "removal mode".
    - [x] When the bypass is removed update the following DB fields.
      - `is_active` : false.
      - `bypass_removed_date`: current date.
      - `bypass_removed_by`: Ministry user.
      - `removal_note_id`: create a related note.
      - usual audit fields.
    - [x] Ensure API enforces the validation.
- [x] Header button "Bypass a restriction" displays the "Bypass restriction" modal in the "create mode".
  - [x] Dropdown should be populated as below.
    - [x] If no restrictions are available to be populated, disable the "Bypass restriction" button, disable the dropdown, and show a waning banner right above the dropdown with the message "No active restrictions available to be bypassed or all active restrictions already have an active bypass.".
    - [x] Display restriction code and creation date from the `sims.student_restrictions` table. The same restriction code may have multiple active entries in the student restrictions. Format sample `B3D added on Aug 28 2024`.
    - [x] Only populate restriction currently active in the student account.
    - [x] Use a simple dropdown to list the items. No search is needed on the client or server side.
    - [x] Populate part-time/full-time restrictions based on the application following the below.
      - For full-time application: only restriction where its action type contains `Stop full time BC funding` or `Stop full time disbursement`.
      - For part-time applications: only restriction where its action type contains `Stop part time disbursement`.
    - [x] Do not add to dropdown active restrictions that already have an active bypass. Filter to be done in the SQL query.
[...]

PR #3905, opened 2024-11-06
#3691 - UX/UI for Ignore Restriction - Bypass Add/Remove - API part

andrepestana-aot:
- Created the api endpoints:
  - GET aest/application-restriction-bypass/:id
  - GET aest/application-restriction-bypass/application/:applicationId/options-list
  - POST aest/application-restriction-bypass
  - PATCH aest/application-restriction-bypass/:id
- Added new role "aest-bypass-student-restriction" only for business administrators;
- Added e2e tests:
  - ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction
    √ Should be able to create a bypass when there is not an active bypass for the same student.                                                                                                                         
    √ Should throw an HTTP error while creating a bypass when there is an active bypass for the same active student restriction.                                                                                         
    √ Should throw an HTTP error while creating a bypass when student restriction is not active.                                                                                                                         
    √ Should throw an HTTP error while creating a bypass when the student application is in draft.                                                                                                                       
    √ Should throw an HTTP error while creating a bypass when the student application is in cancelled.            
  - ApplicationRestrictionBypassAESTController(e2e)-getApplicationRestrictionBypass.
    √ Should get an application restriction bypass for a submitted part-time application when there is an application restriction bypass with the required id.                                                           
    √ Should get an inactive application restriction bypass for a submitted part-time application when there is an application restriction bypass with the required id.         
  - ApplicationRestrictionBypassAESTController(e2e)-getAvailableRestrictionsToBypass.
    √ Should list all student active restrictions and not already bypassed for a part-time application when there is one restriction bypassed and some others not.                                       
    √ Should list all student active restrictions for a full-time application when there is one restriction bypassed and others not.    
  - ApplicationRestrictionBypassAESTController(e2e)-removeBypassRestriction
    √ Should be able to remove an active bypass when there is an active bypass.                                                                                                                          
    √ Should throw an HTTP error while removing a bypass when the bypass is not active.                                                                                                                  
    √ Should throw an HTTP error while removing a bypass when the bypass is not found.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -170,7 +170,6 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       {
         student: application.student,
         restriction: ptssrRestriction,
-        application,
         resolutionNote: null,
         creator: sharedMinistryUser,
       },
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -24,6 +24,11 @@ import {
   RestrictionBypassBehaviors,
   User,
 } from "@sims/sims-db";
+import {
+  ACTIVE_BYPASS_FOR_STUDENT_RESTRICTION_ALREADY_EXISTS,
+  APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION,
+  STUDENT_RESTRICTION_IS_NOT_ACTIVE,
+} from "../../../../constants";
 
 describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", () => {
   let app: INestApplication;
@@ -149,10 +154,9 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       .auth(token, BEARER_AUTH_TYPE)
       .expect(HttpStatus.UNPROCESSABLE_ENTITY)
       .expect({
-        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
         message:
           "Cannot create a bypass when there is an active bypass for the same active student restriction.",
-        error: "Unprocessable Entity",
+        errorType: ACTIVE_BYPASS_FOR_STUDENT_RESTRICTION_ALREADY_EXISTS,
       });
   });
 
@@ -191,10 +195,9 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       .auth(token, BEARER_AUTH_TYPE)
       .expect(HttpStatus.UNPROCESSABLE_ENTITY)
       .expect({
-        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
         message:
           "Cannot create a bypass when student restriction is not active.",
-        error: "Unprocessable Entity",
+        errorType: STUDENT_RESTRICTION_IS_NOT_ACTIVE,
       });
   });
 
@@ -228,9 +231,9 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       .auth(token, BEARER_AUTH_TYPE)
       .expect(HttpStatus.UNPROCESSABLE_ENTITY)
       .expect({
-        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
         message: "Cannot create a bypass when application is in invalid state.",
-        error: "Unprocessable Entity",
+        errorType:
+          APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION,
       });
   });
 
@@ -264,9 +267,9 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       .auth(token, BEARER_AUTH_TYPE)
       .expect(HttpStatus.UNPROCESSABLE_ENTITY)
       .expect({
-        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
         message: "Cannot create a bypass when application is in invalid state.",
-        error: "Unprocessable Entity",
+        errorType:
+          APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION,
       });
   });
 
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -1,0 +1,273 @@
+import { HttpStatus, INestApplication } from "@nestjs/common";
+import {
+  E2EDataSources,
+  RestrictionCode,
+  createE2EDataSources,
+  createFakeUser,
+  saveFakeApplication,
+  saveFakeApplicationRestrictionBypass,
+  saveFakeStudentRestriction,
+} from "@sims/test-utils";
+import {
+  AESTGroups,
+  BEARER_AUTH_TYPE,
+  createTestingAppModule,
+  getAESTToken,
+} from "../../../../testHelpers";
+import * as request from "supertest";
+import {
+  ApplicationStatus,
+  NoteType,
+  OfferingIntensity,
+  Restriction,
+  RestrictionActionType,
+  RestrictionBypassBehaviors,
+  User,
+} from "@sims/sims-db";
+
+describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", () => {
+  let app: INestApplication;
+  let db: E2EDataSources;
+  let sharedMinistryUser: User;
+  let ptssrRestriction: Restriction;
+  let endpoint: string;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    db = createE2EDataSources(dataSource);
+    sharedMinistryUser = await db.user.save(createFakeUser());
+    ptssrRestriction = await db.restriction.findOneBy({
+      restrictionCode: RestrictionCode.PTSSR,
+    });
+    endpoint = "/aest/application-restriction-bypass";
+  });
+
+  it("Should be able to create a bypass when there is not an active bypass for the same student.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+    });
+
+    const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
+      student: application.student,
+      restriction: ptssrRestriction,
+      application,
+      resolutionNote: null,
+      creator: sharedMinistryUser,
+    });
+
+    const payload = {
+      applicationId: application.id,
+      studentRestrictionId: studentRestriction.id,
+      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      note: "test note",
+    };
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    // Act/Assert
+    let applicationRestrictionBypassId: number;
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(payload)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.CREATED)
+      .then((response) => {
+        expect(response.body.id).toBeGreaterThan(0);
+        applicationRestrictionBypassId = response.body.id;
+      });
+
+    const applicationRestrictionBypass =
+      await db.applicationRestrictionBypass.findOne({
+        select: {
+          id: true,
+          application: { id: true },
+          studentRestriction: { id: true },
+          bypassBehavior: true,
+          creationNote: { noteType: true, description: true },
+          bypassCreatedDate: true,
+          bypassRemovedDate: true,
+          createdAt: true,
+          isActive: true,
+        },
+        relations: {
+          application: true,
+          studentRestriction: true,
+          creationNote: true,
+        },
+        where: { id: applicationRestrictionBypassId },
+      });
+
+    expect(applicationRestrictionBypass).toMatchObject({
+      id: applicationRestrictionBypassId,
+      application: { id: application.id },
+      studentRestriction: { id: studentRestriction.id },
+      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      creationNote: {
+        noteType: NoteType.Application,
+        description: payload.note,
+      },
+      bypassCreatedDate: expect.any(Date),
+      bypassRemovedDate: null,
+      createdAt: expect.any(Date),
+      isActive: true,
+    });
+  });
+
+  it("Should throw an HTTP error while creating a bypass when there is an active bypass for the same active student restriction ID.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+    });
+    const restrictionBypass = await saveFakeApplicationRestrictionBypass(
+      db,
+      {
+        application,
+        bypassCreatedBy: sharedMinistryUser,
+        creator: sharedMinistryUser,
+      },
+      {
+        restrictionActionType: RestrictionActionType.StopPartTimeDisbursement,
+        restrictionCode: RestrictionCode.PTSSR,
+      },
+    );
+    const payload = {
+      applicationId: application.id,
+      studentRestrictionId: restrictionBypass.studentRestriction.id,
+      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      note: "test note",
+    };
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(payload)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.UNPROCESSABLE_ENTITY)
+      .expect({
+        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+        message:
+          "Cannot create a bypass when there is an active bypass for the same active student restriction ID.",
+        error: "Unprocessable Entity",
+      });
+  });
+
+  it("Should throw an HTTP error while creating a bypass when student restriction is not active.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+    });
+
+    const studentRestriction = await saveFakeStudentRestriction(
+      db.dataSource,
+      {
+        student: application.student,
+        restriction: ptssrRestriction,
+        application,
+        resolutionNote: null,
+        creator: sharedMinistryUser,
+      },
+      {
+        isActive: false,
+      },
+    );
+
+    const payload = {
+      applicationId: application.id,
+      studentRestrictionId: studentRestriction.id,
+      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      note: "test note",
+    };
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(payload)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.UNPROCESSABLE_ENTITY)
+      .expect({
+        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+        message:
+          "Cannot create a bypass when student restriction is not active.",
+        error: "Unprocessable Entity",
+      });
+  });
+
+  it("Should throw an HTTP error while creating a bypass when the student application is in draft.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+      applicationStatus: ApplicationStatus.Draft,
+    });
+
+    const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
+      student: application.student,
+      restriction: ptssrRestriction,
+      application,
+      resolutionNote: null,
+      creator: sharedMinistryUser,
+    });
+
+    const payload = {
+      applicationId: application.id,
+      studentRestrictionId: studentRestriction.id,
+      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      note: "test note",
+    };
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(payload)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.UNPROCESSABLE_ENTITY)
+      .expect({
+        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+        message: "Cannot create a bypass when application is in invalid state.",
+        error: "Unprocessable Entity",
+      });
+  });
+
+  it("Should throw an HTTP error while creating a bypass when the student application is in cancelled.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+      applicationStatus: ApplicationStatus.Cancelled,
+    });
+
+    const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
+      student: application.student,
+      restriction: ptssrRestriction,
+      application,
+      resolutionNote: null,
+      creator: sharedMinistryUser,
+    });
+
+    const payload = {
+      applicationId: application.id,
+      studentRestrictionId: studentRestriction.id,
+      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      note: "test note",
+    };
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(endpoint)
+      .send(payload)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.UNPROCESSABLE_ENTITY)
+      .expect({
+        statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+        message: "Cannot create a bypass when application is in invalid state.",
+        error: "Unprocessable Entity",
+      });
+  });
+
+  afterAll(async () => {
+    await app?.close();
+  });
+});
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -110,7 +110,7 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
         description: payload.note,
       },
       bypassCreatedDate: expect.any(Date),
-      bypassCreatedBy: { id: sharedMinistryUser.id },
+      bypassCreatedBy: { id: expect.any(Number) },
       bypassRemovedDate: null,
       createdAt: expect.any(Date),
       isActive: true,
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -57,7 +57,6 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
     const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
       student: application.student,
       restriction: ptssrRestriction,
-      application,
       resolutionNote: null,
       creator: sharedMinistryUser,
     });
@@ -211,7 +210,6 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
     const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
       student: application.student,
       restriction: ptssrRestriction,
-      application,
       resolutionNote: null,
       creator: sharedMinistryUser,
     });
@@ -247,7 +245,6 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
     const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
       student: application.student,
       restriction: ptssrRestriction,
-      application,
       resolutionNote: null,
       creator: sharedMinistryUser,
     });
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -199,7 +199,7 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       });
   });
 
-  it("Should throw an HTTP error while creating a bypass when the student application is in draft.", async () => {
+  it("Should throw an HTTP error while creating a bypass when the student application is draft.", async () => {
     // Arrange
     const application = await saveFakeApplication(db.dataSource, undefined, {
       offeringIntensity: OfferingIntensity.partTime,
@@ -234,7 +234,7 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       });
   });
 
-  it("Should throw an HTTP error while creating a bypass when the student application is in cancelled.", async () => {
+  it("Should throw an HTTP error while creating a bypass when the student application is cancelled.", async () => {
     // Arrange
     const application = await saveFakeApplication(db.dataSource, undefined, {
       offeringIntensity: OfferingIntensity.partTime,
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
@@ -86,6 +86,7 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
           bypassBehavior: true,
           creationNote: { noteType: true, description: true },
           bypassCreatedDate: true,
+          bypassCreatedBy: { id: true },
           bypassRemovedDate: true,
           createdAt: true,
           isActive: true,
@@ -94,6 +95,7 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
           application: true,
           studentRestriction: true,
           creationNote: true,
+          bypassCreatedBy: true,
         },
         where: { id: applicationRestrictionBypassId },
       });
@@ -102,19 +104,20 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       id: applicationRestrictionBypassId,
       application: { id: application.id },
       studentRestriction: { id: studentRestriction.id },
-      bypassBehavior: RestrictionBypassBehaviors.NextDisbursementOnly,
+      bypassBehavior: payload.bypassBehavior,
       creationNote: {
         noteType: NoteType.Application,
         description: payload.note,
       },
       bypassCreatedDate: expect.any(Date),
+      bypassCreatedBy: { id: sharedMinistryUser.id },
       bypassRemovedDate: null,
       createdAt: expect.any(Date),
       isActive: true,
     });
   });
 
-  it("Should throw an HTTP error while creating a bypass when there is an active bypass for the same active student restriction ID.", async () => {
+  it("Should throw an HTTP error while creating a bypass when there is an active bypass for the same active student restriction.", async () => {
     // Arrange
     const application = await saveFakeApplication(db.dataSource, undefined, {
       offeringIntensity: OfferingIntensity.partTime,
@@ -148,7 +151,7 @@ describe("ApplicationRestrictionBypassAESTController(e2e)-bypassRestriction", ()
       .expect({
         statusCode: HttpStatus.UNPROCESSABLE_ENTITY,
         message:
-          "Cannot create a bypass when there is an active bypass for the same active student restriction ID.",
+          "Cannot create a bypass when there is an active bypass for the same active student restriction.",
         error: "Unprocessable Entity",
       });
   });
--- a/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.getApplicationRestrictionBypass.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.getApplicationRestrictionBypass.e2e-spec.ts
@@ -4,7 +4,6 @@ import {
   RestrictionCode,
   createE2EDataSources,
   createFakeUser,
-  saveFakeApplication,
   saveFakeApplicationRestrictionBypass,
 } from "@sims/test-utils";
 import {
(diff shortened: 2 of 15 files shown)


sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+  }
+
+  /**
+   * Creates a new application restriction bypass.
+   * @param payload application restriction bypass data.
+   * @param auditUserId id of the user who is creating the bypass.
+   * @returns created application restriction bypass.
+   */
+  async bypassRestriction(
+    payload: BypassRestrictionAPIInDTO,
```
andrewsignori-aot, line 230: Except for controller services, the regular services do not receive the DTOs.
A service model should be created.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+        );
+
+        const notes = new Note();
+        notes.description = payload.note;
+        notes.noteType = NoteType.Application;
+        notes.creator = auditUser;
+        notes.createdAt = now;
+        const noteObj = await transactionalEntityManager
+          .getRepository(Note)
+          .save(notes);
```
andrewsignori-aot, line 258: Please use the notes service `NoteSharedService.createStudentNote`.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+   * @param auditUserId id of the user who is creating the bypass.
+   * @returns created application restriction bypass.
+   */
+  async bypassRestriction(
+    payload: BypassRestrictionAPIInDTO,
+    auditUserId: number,
+  ): Promise<ApplicationRestrictionBypass> {
+    const auditUser = { id: auditUserId } as User;
+    const now = new Date();
+    return await this.dataSource.transaction(
```
andrewsignori-aot, line 235: No need to await.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +90,171 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+              "Could not find student restriction for the given id.",
+            );
+          case STUDENT_RESTRICTION_IS_NOT_ACTIVE:
+            throw new UnprocessableEntityException(
+              "Cannot create a bypass when student restriction is not active.",
+            );
+          case APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION:
+            throw new UnprocessableEntityException(
+              "Cannot create a bypass when application is in invalid state.",
+            );
```
andrewsignori-aot, line 208: The usual approach for the error names would be creating the necessary ones to make decisions.
If the only condition applied to different error names is the message, the usual approach is the below.
https://github.com/bcgov/SIMS/blob/7c7f20cf45757b5f6e3021b4e60cf2e184b815e1/sources/packages/backend/apps/api/src/route-controllers/application-offering-change-request/application-offering-change-request.institutions.controller.ts#L296-L312

Here, the ones that are expected to be handled by the consumer are thrown using the below.
```ts
new ApiProcessError(error.message, error.name)
```
Ones that are not expected to be processed by the consumer are thrown using the below.
```ts
throw new NotFoundException(error.message);
```

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+          payload.studentRestrictionId,
+        );
+        await this.checkForActiveStudentRestriction(
+          transactionalEntityManager,
+          payload.studentRestrictionId,
+        );
+        await this.checkForApplicationInValidState(
+          transactionalEntityManager,
+          payload.applicationId,
+        );
```
andrewsignori-aot, line 249: Please execute these three operations in parallel.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+   * @param payload application restriction bypass data.
+   * @param auditUserId id of the user who is creating the bypass.
+   * @returns created application restriction bypass.
+   */
+  async bypassRestriction(
+    payload: BypassRestrictionAPIInDTO,
+    auditUserId: number,
+  ): Promise<ApplicationRestrictionBypass> {
+    const auditUser = { id: auditUserId } as User;
+    const now = new Date();
```
andrewsignori-aot, line 234: Can this be moved closer to where it is consumed (right before line 250)?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+    return await this.dataSource.transaction(
+      async (transactionalEntityManager) => {
+        await this.checkForActiveApplicationStudentRestrictionByPasses(
```
andrewsignori-aot, line 237: Is there a benefit to executing these validations inside the transition?
If one of them fails a transaction does not even need to be started.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+   * Checks if a student restriction is active.
+   * Throws an error if the student restriction is not found
+   * or if it is not active.
+   * @param transactionalEntityManager transactional entity manager to execute the query.
+   * @param studentRestrictionId id of the student restriction to check.
+   */
+  private async checkForActiveStudentRestriction(
+    transactionalEntityManager: EntityManager,
+    studentRestrictionId: number,
+  ) {
```
andrewsignori-aot, line 322: Please add the `Promise<void>`.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+  }
+
+  /**
+   * Checks if there is an active application student restriction bypass.
+   * Throws an error if there is an active bypass for the same active student restriction id.
+   * @param transactionalEntityManager transactional entity manager to execute the query.
+   * @param applicationId id of the application to check.
+   * @param studentRestrictionId id of the student restriction to check.
+   */
+  private async checkForActiveApplicationStudentRestrictionByPasses(
```
andrewsignori-aot, line 354: Bypass is a single word hence it should be `checkForActiveApplicationStudentRestrictionBypasses`, right?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+    transactionalEntityManager: EntityManager,
+    applicationId: number,
+    studentRestrictionId: number,
+  ) {
+    const existsActiveStudentRestriction = await transactionalEntityManager
+      .getRepository(ApplicationRestrictionBypass)
+      .exists({
+        relations: { studentRestriction: true },
```
andrewsignori-aot, line 362: Is the `relations` required for `exists`?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
       order: { createdAt: "DESC" },
     });
   }
+
+  /**
+   * Gets an application restriction bypass by id.
+   * @param id id of the application restriction bypass to retrieve.
+   * @returns application restriction bypass.
+   */
+  getApplicationRestrictionBypass(
```
andrewsignori-aot, line 91: Missing the async.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+      where: {
+        id,
+        application: {
+          applicationStatus: Not(
+            In([
+              ApplicationStatus.Draft,
+              ApplicationStatus.Cancelled,
+              ApplicationStatus.Overwritten,
+            ]),
+          ),
```
andrewsignori-aot, line 128: Why would be require some extra where condition than the ID itself?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+            restrictionCode: true,
+          },
+        },
+        creationNote: true as unknown,
+        removalNote: true as unknown,
+        bypassCreatedBy: true as unknown,
+        bypassCreatedDate: true,
+        bypassRemovedBy: true as unknown,
+        bypassRemovedDate: true,
+        bypassBehavior: true,
```
andrewsignori-aot, line 110: `true as unknown` is used as a workaround only for JSONB data.
Any other table/column should have its individual column names selected.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
andrewsignori-aot: Can you please also remove the below condition
```ts
applicationStatus: Not(ApplicationStatus.Draft)
```
from the method `getApplicationRestrictionBypasses`. It can create confusion and it is not right.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+      .select([
+        "studentRestriction.id",
+        "restriction.restrictionCode",
+        "studentRestriction.createdAt",
+      ])
+      .innerJoin("studentRestriction.restriction", "restriction")
+      .innerJoin("studentRestriction.application", "application")
+      .innerJoin("application.currentAssessment", "currentAssessment")
+      .innerJoin("currentAssessment.offering", "offering")
+      .where("studentRestriction.application = :applicationId", {
```
andrewsignori-aot, line 153: Is the where condition generation the expected SQL? Is it not supposed to use the `application.id` instead of the `application = :applicationId"`?
```
studentRestriction.application.id = :applicationId
```

andrewsignori-aot, line 153: This condition should not be present in the query.
Any active student restriction should be listed to be bypassed.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+            applicationId,
+          })
+          .andWhere(
+            "applicationRestrictionBypass.studentRestriction = studentRestriction.id",
+          )
+          .andWhere("applicationRestrictionBypass.isActive = true");
+        return `NOT EXISTS (${subQuery.getQuery()})`;
+      })
+      // Restriction action type condition.
+      .andWhere((qb) => {
```
andrewsignori-aot, line 172: Please correct me if I am missing something, but knowing before query execution if the application is part-time/full-time would dramatically simplify this entire where to a simple condition. Does it make sense?
Also, the number of active student restrictions at any given time would be probably 1 to few, which means that it could be even filtered in memory to avoid a complex SQL query.

Also, the query relies on the fact the student restriction will be associated with an application when it may not.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/models/application-restriction-bypass.dto.ts
```diff
@@ -15,3 +21,41 @@ export class ApplicationRestrictionBypassSummary {
...
 }
+
+export class BypassRestrictionAPIInDTO {
+  @IsPositive()
+  applicationId: number;
+  @IsPositive()
+  studentRestrictionId: number;
+  @IsEnum(RestrictionBypassBehaviors)
+  bypassBehavior: RestrictionBypassBehaviors;
+  @IsString()
```
andrewsignori-aot, line 32: Will it ensure that a string different than `""` is not provided?
Usually we use the `@IsNotEmpty()`.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/models/application-restriction-bypass.dto.ts
```diff
@@ -15,3 +21,41 @@ export class ApplicationRestrictionBypassSummary {
...
+}
+
+export class AvailableStudentRestrictionAPIOutDTO {
+  availableRestrictionsToBypass: {
+    studentRestrictionId: number;
+    restrictionCode: string;
+    studentRestrictionCreatedAt: Date;
+  }[];
+}
+export class ApplicationRestrictionBypassAPIOutDTO {
```
andrewsignori-aot, line 50: please add a blank line.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/models/application-restriction-bypass.dto.ts
```diff
@@ -15,3 +21,41 @@ export class ApplicationRestrictionBypassSummary {
...
+  applicationRestrictionBypassId: number;
+  studentRestrictionId: number;
+  restrictionCode: string;
+  applicationRestrictionBypassCreatedDate: string;
+  applicationRestrictionBypassCreatedBy: string;
+  applicationRestrictionBypassCreationNote: string;
+  applicationRestrictionBypassRemovedDate?: string;
+  applicationRestrictionBypassRemovedBy?: string;
+  applicationRestrictionBypassRemovalNote?: string;
+  applicationRestrictionBypassBehavior: string;
```
andrewsignori-aot, line 60: In a DTO already named `ApplicationRestrictionBypass` I would suggest removing the prefix `applicationRestrictionBypass` from all these properties.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +90,171 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+          default:
+            throw error;
+        }
+      }
+      throw error;
+    }
+  }
+
+  @ApiNotFoundResponse({
```
andrewsignori-aot, line 217: Please add some comments.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +90,171 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+              "Application restriction bypass not found.",
+            );
+          case STUDENT_RESTRICTION_IS_NOT_ACTIVE:
+            throw new UnprocessableEntityException(
+              "Cannot remove a bypass when student restriction is not active.",
+            );
+          case APPLICATION_RESTRICTION_BYPASS_IS_NOT_ACTIVE:
+            throw new UnprocessableEntityException(
+              "Cannot remove a bypass when application restriction bypass is not active.",
+            );
```
andrewsignori-aot, line 252: Similar to https://github.com/bcgov/SIMS/pull/3905/files#r1833390148

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +90,171 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+          default:
+            throw error;
```
andrewsignori-aot, line 254: Is this required since it will reach line 257 either way?

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +90,171 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+          case STUDENT_RESTRICTION_IS_NOT_ACTIVE:
+            throw new UnprocessableEntityException(
+              "Cannot create a bypass when student restriction is not active.",
+            );
+          case APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION:
+            throw new UnprocessableEntityException(
+              "Cannot create a bypass when application is in invalid state.",
+            );
+          default:
+            throw error;
```
andrewsignori-aot, line 210: Same from https://github.com/bcgov/SIMS/pull/3905/files#r1833553979

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
andrewsignori-aot: Thanks for doing the change. For my understanding all the errors are now using `ApiProcessError` because there is an expectation to that they will need to be handled by the consumer/UI (showing something for the user)?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -1,17 +1,50 @@
...
+  STUDENT_RESTRICTION_IS_NOT_ACTIVE,
+  APPLICATION_RESTRICTION_BYPASS_IS_NOT_ACTIVE,
+  STUDENT_RESTRICTION_NOT_FOUND,
+} from "../../constants";
 
 /**
  * Service layer for application restriction bypasses.
  */
 @Injectable()
 export class ApplicationRestrictionBypassService extends RecordDataModelService<ApplicationRestrictionBypass> {
```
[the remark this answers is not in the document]
andrewsignori-aot: Thanks for taking care of this ❤️

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -1,13 +1,53 @@
...
+import { IUserToken, Role } from "../../auth";
+import { CustomNamedError } from "@sims/utilities";
+import {
+  ACTIVE_BYPASS_FOR_STUDENT_RESTRICTION_ALREADY_EXISTS,
+  APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION,
+  APPLICATION_RESTRICTION_BYPASS_IS_NOT_ACTIVE,
+  APPLICATION_RESTRICTION_BYPASS_NOT_FOUND,
+  STUDENT_RESTRICTION_IS_NOT_ACTIVE,
+  STUDENT_RESTRICTION_NOT_FOUND,
+} from "../../../src/constants";
```
andrewsignori-aot, line 49: Please adjust the import to `../../constants`.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.models.ts
```diff
@@ -0,0 +1,12 @@
+import { RestrictionBypassBehaviors } from "@sims/sims-db";
+
+export interface BypassRestrictionData {
+  applicationId: number;
+  studentRestrictionId: number;
+  bypassBehavior: RestrictionBypassBehaviors;
+  note: string;
+}
+
+export interface RemoveBypassRestrictionData {
```
andrewsignori-aot, line 10: Just a clarification, a model does not need to be created for every service call.
The idea is just to allow services to be reused by other services without having the controllers DTOs as a parameter.
This one as a model is not a blocker but it would also be possible to be just a single parameter in the service method.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -1,19 +1,46 @@
...
+  APPLICATION_RESTRICTION_BYPASS_NOT_FOUND,
+  APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION,
+  STUDENT_RESTRICTION_IS_NOT_ACTIVE,
+  APPLICATION_RESTRICTION_BYPASS_IS_NOT_ACTIVE,
+  STUDENT_RESTRICTION_NOT_FOUND,
+} from "../../constants";
+import {
+  BypassRestrictionData,
+  RemoveBypassRestrictionData,
+} from "apps/api/src/services/application-restriction-bypass/application-restriction-bypass.models";
```
andrewsignori-aot, line 26: Please adjuste the import. Are these not supposed to be part of some index in the services?

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
```diff
@@ -0,0 +1,279 @@
...
+  it("Should be able to create a bypass when there is not an active bypass for the same student.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+    });
+
+    const studentRestriction = await saveFakeStudentRestriction(db.dataSource, {
+      student: application.student,
+      restriction: ptssrRestriction,
+      application,
```
andrewsignori-aot, line 60: As a suggestion, the application can be removed to ensure that bypass can be applied without the relationship in place.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
andrewsignori-aot: We have expected "business" errors and some that are not.
For instance, a NOT FOUND usually is not returned using the `ApiProcessError`.
In essence we can return every single API error in a `ApiProcessError` but we are not following this.
The rule is: expected messages that are required to be shown to the user are wrapped in a `ApiProcessError`, every other will be just a regular exception.

andrewsignori-aot: It will not be a blocker, but the understanding should be clear.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.getApplicationRestrictionBypass.e2e-spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    app = nestApplication;
+    db = createE2EDataSources(dataSource);
+    sharedMinistryUser = await db.user.save(createFakeUser());
+  });
+
+  it("Should get an application restriction bypass for a submitted part-time application when there is an application restriction bypass with the required id.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      offeringIntensity: OfferingIntensity.partTime,
+    });
```
andrewsignori-aot, line 36: Suggestion, a new application in being created many times to be provided to `saveFakeApplicationRestrictionBypass` and it can be avoided if the `saveFakeApplicationRestrictionBypass` creates as an application if one is not provided.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.removeBypassRestriction.e2e-spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      id: restrictionBypass.id,
+      removalNote: {
+        noteType: NoteType.Application,
+        description: payload.note,
+      },
+      bypassRemovedDate: expect.any(Date),
+      bypassRemovedBy: { id: expect.any(Number) },
+      isActive: false,
+      updatedAt: expect.any(Date),
+      modifier: { id: expect.any(Number) },
```
andrewsignori-aot, line 98: Along all tests where the `sharedMinistryUser`, the creator and modifiers can be tested using the expected id.

andrewsignori-aot (commented):
Great work and great E2Es. Please take a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.removeBypassRestriction.e2e-spec.ts
andrepestana-aot: This `shareMinistryUser` is used for the DB operations but the user from the request is a different one.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
andrepestana-aot: IMO all this messages are meaningful for the user if that happens. I know it's unlikely to happen but at least when they happen the user might have a clue of the issue or tell it to a systems administrator. Otherwise, what do you suggest?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
andrepestana-aot: That worked without I noticed that.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -49,4 +82,380 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+        const notes = new Note();
+        notes.description = payload.note;
+        notes.noteType = NoteType.Application;
+        notes.creator = auditUser;
+        notes.createdAt = now;
+        const noteObj = await transactionalEntityManager
+          .getRepository(Note)
+          .save(notes);
+
+        return await transactionalEntityManager
```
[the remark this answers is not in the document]
andrewsignori-aot: Was this changed?
I still can see 
```ts
return await this.dataSource.transaction(
```

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
andrewsignori-aot: > We have expected "business" errors and some that are not. For instance, a NOT FOUND usually is not returned using the `ApiProcessError`. In essence we can return every single API error in a `ApiProcessError` but we are not following this. The rule is: expected messages that are required to be shown to the user are wrapped in a `ApiProcessError`, every other will be just a regular exception.
> 
> For instance, are the below ones meaningful from user perspective?
> 
> * STUDENT_RESTRICTION_NOT_FOUND
> * APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION

My suggestion is explained above and it represents what I believe the solution has been following.
From a point of view, as I mentioned earlier, we can return every single API error as an `ApiProcessError`. The system administrator should and will not rely on what the user sees. The application logs must provide the complete error and should be the source of thurh for investigation.

Please keep all of these as they are. I am closing the conversation.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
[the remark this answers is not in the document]
andrepestana-aot: Sorry, forgot that but now I changed it. Thanks!

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
andrepestana-aot: Sorry, moved to index. Thanks!

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -43,10 +67,382 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+      },
+    });
+  }
+
+  /**
+   * Gets all available student restrictions to bypass for a given application.
+   * @param applicationId application id.
+   * @returns list of application restriction bypass options.
+   */
+  async getAvailableStudentRestrictionsToBypass(
```
andrewsignori-aot, line 130: This method now executes three SQL queries to filter an amount of data that should be pretty small (1 to few active records).
My recommendation is to get from DB everything that is needed, as much as possible with a single query and do minor in-memory process to finish the filter.
For instance, the below query would return
- all active student restrictions (expect zero records to couples).
- all active bypasses (expect zero records to couples).
- only restriction with the expected action types.
- offering intensity.
From the returned query, we can eliminate the ones already in place and the ones for the action types not needed for the offering intensity.
```ts
this.applicationRepo.find({
  select: {
    id: true,
    student: {
      id: true,
      studentRestrictions: {
        id: true,
        restriction: {
          id: true,
          restrictionCode: true,
        },
        restrictionBypasses: {
          id: true,
          studentRestriction: { id: true },
        },
      },
    },
    currentAssessment: {
      id: true,
      offering: {
        id: true,
        offeringIntensity: true,
      },
    },
  },
  relations: {
    student: {
      studentRestrictions: { restriction: true, restrictionBypasses: true },
    },
    currentAssessment: { offering: true },
  },
  where: {
    id: applicationId,
    student: {
      studentRestrictions: {
        isActive: true,
        restrictionBypasses: {
          isActive: true,
        },
        restriction: {
          actionType: ArrayContains([
            RestrictionActionType.StopFullTimeBCFunding,
            RestrictionActionType.StopFullTimeDisbursement,
            RestrictionActionType.StopPartTimeBCFunding,
            RestrictionActionType.StopPartTimeDisbursement,
          ]),
        },
      },
    },
  },
});
```

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -43,10 +67,382 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+
+    const query = this.studentRestrictionRepo
+      .createQueryBuilder("studentRestriction")
+      .select([
+        "studentRestriction.id",
+        "restriction.restrictionCode",
+        "studentRestriction.createdAt",
+      ])
+      .innerJoin("studentRestriction.restriction", "restriction")
+      .andWhere("studentRestriction.student.id = :studentId", {
```
andrewsignori-aot, line 186: The first where in the sequence should be `where`.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -43,10 +67,382 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+        studentId: application.student.id,
+      })
+      .andWhere("studentRestriction.id NOT IN (:...studentRestrictionIds)", {
```
andrewsignori-aot, line 189: Is there a check needed for the case where the `studentRestrictionIds` will be empty?

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -43,10 +67,382 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+      })
+      .andWhere("studentRestriction.isActive = true");
+
+    // Restriction action type condition.
+    if (
+      application.currentAssessment.offering.offeringIntensity ===
+      OfferingIntensity.fullTime
+    ) {
+      query.andWhere(
+        "restriction.action_type::text[] && ARRAY[:...fullTimeActionTypes]::text[]",
```
andrewsignori-aot, line 200: No need for conversion on both sides of the comparison, Posgres has an operation to search arrays.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.getApplicationRestrictionBypass.e2e-spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        db,
+        {
+          bypassCreatedBy: sharedMinistryUser,
+          creator: sharedMinistryUser,
+          bypassRemovedBy: sharedMinistryUser,
+        },
+        {
+          restrictionActionType: RestrictionActionType.StopPartTimeDisbursement,
+          restrictionCode: RestrictionCode.PTSSR,
+
```
andrewsignori-aot, line 83: Empty line can be removed.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.getAvailableRestricitionsToBypass.e2e-spec.ts
```diff
@@ -0,0 +1,240 @@
...
+        {
+          application,
+          bypassCreatedBy: sharedMinistryUser,
+          bypassRemovedBy: sharedMinistryUser,
+          creator: sharedMinistryUser,
+        },
+        {
+          restrictionActionType: RestrictionActionType.StopFullTimeDisbursement,
+          restrictionCode: RestrictionCode.AF4,
+
```
andrewsignori-aot, line 166: Empty line can be removed.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +90,153 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+  @Get(":id")
+  async getApplicationRestrictionBypass(
+    @Param("id", ParseIntPipe) id: number,
+  ): Promise<ApplicationRestrictionBypassAPIOutDTO> {
+    const applicationRestrictionBypass =
+      await this.applicationRestrictionBypassService.getApplicationRestrictionBypass(
+        id,
+      );
+    if (!applicationRestrictionBypass) {
+      throw new NotFoundException(APPLICATION_RESTRICTION_BYPASS_NOT_FOUND);
```
andrewsignori-aot, line 111: Please follow the other existing `NotFoundException` and throw a message instead of the error code.

andrewsignori-aot (commented):
Thanks for making the changes, please take a look at the comments, and please let me know if the the below one or the query that it refers to needs further clarification.
https://github.com/bcgov/SIMS/pull/3905#discussion_r1838788631

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
lewischen-aot: I agree with this approach to boost performance for database querying.

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/_tests_/e2e/application-restriction-bypass.aest.controller.bypassRestriction.e2e-spec.ts
```diff
@@ -0,0 +1,275 @@
...
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.UNPROCESSABLE_ENTITY)
+      .expect({
+        message: "Cannot create a bypass when application is in invalid state.",
+        errorType:
+          APPLICATION_IN_INVALID_STATE_FOR_APPLICATION_RESTRICTION_BYPASS_CREATION,
+      });
+  });
+
+  it("Should throw an HTTP error while creating a bypass when the student application is in cancelled.", async () => {
```
lewischen-aot, line 237: There is a minor typo here: "application is in cancelled" should be "application is cancelled".

sources/packages/backend/libs/sims-db/src/entities/student-restriction.model.ts
```diff
@@ -28,4 +34,15 @@ export class StudentRestriction extends BaseRestrictionModel {
     referencedColumnName: ColumnNames.ID,
   })
   application: Application;
+
+  @OneToMany(
```
andrewsignori-aot, line 38: Please add some comments.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -43,10 +70,387 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+        )
+        .filter((bypass) => bypass.isActive)
+        .map((bypass) => bypass.studentRestriction.id);
+
+    const studentRestrictionsThatAreNotBypassed =
+      application.student.studentRestrictions.filter(
+        (studentRestriction) =>
+          !bypassedStudentRestrictionIds.includes(studentRestriction.id),
+      );
+
```
lewischen-aot, line 215: This comment is for my understanding. `studentRestrictionsThatAreNotBypassed` represent an array of student restrictions for which the associated application restriction bypass records are not active. Is it correct?

sources/packages/backend/apps/api/src/route-controllers/application-restriction-bypass/application-restriction-bypass.aest.controller.ts
```diff
@@ -50,4 +87,147 @@ export class ApplicationRestrictionBypassAESTController extends BaseController {
...
+  @Get("application/:applicationId/options-list")
+  async getAvailableStudentRestrictionsToBypass(
+    @Param("applicationId", ParseIntPipe) applicationId: number,
+  ): Promise<AvailableStudentRestrictionAPIOutDTO> {
+    const availableRestrictionsToBypass =
+      await this.applicationRestrictionBypassService.getAvailableStudentRestrictionsToBypass(
+        applicationId,
+      );
+    return {
+      availableRestrictionsToBypass: availableRestrictionsToBypass,
```
andrewsignori-aot, line 145: Service models are not directly sent as API output, they must be mapped to DTO.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
```diff
@@ -43,10 +70,387 @@ export class ApplicationRestrictionBypassService extends RecordDataModelService<
...
+        allowedRestrictionActions.some((actionType) =>
+          studentRestriction.restriction.actionType.includes(actionType),
+        ),
+      )
+      .map((studentRestriction) => ({
+        studentRestrictionId: studentRestriction.id,
+        restrictionCode: studentRestriction.restriction.restrictionCode,
+        studentRestrictionCreatedAt: studentRestriction.createdAt,
+      }))
+      .sort((a, b) => a.restrictionCode.localeCompare(b.restrictionCode));
```
lewischen-aot, line 227: 👍 This sort looks good to me, although there is no specific AC needed to sort the results.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
andrepestana-aot: Yes, the application restriction bypass records may not be active or even not exist. If they exist and are active, we're going to remove from the available student restrictions to bypass list.

sources/packages/backend/apps/api/src/services/application-restriction-bypass/application-restriction-bypass.service.ts
andrewsignori-aot: Changes looks great, thanks for making them happen.
As a suggestion, I would reoder and drop some comments as below. Not a blocker.
```ts
// Filter all active bypasses.
    const bypassedStudentRestrictionIds =
      application.student.studentRestrictions
        .flatMap(
          (studentRestriction) =>
            studentRestriction.applicationRestrictionBypasses,
        )
        .filter((bypass) => bypass.isActive)
        .map((bypass) => bypass.studentRestriction.id);
    // Filter all active student restriction that does not have an active bypass.
    const studentRestrictionsThatAreNotBypassed =
      application.student.studentRestrictions.filter(
        (studentRestriction) =>
          !bypassedStudentRestrictionIds.includes(studentRestriction.id),
      );
    // Define the specific restrictions actions allowed for the application,
    // to filter the returned student restrictions for the requested application.
    const allowedRestrictionActions =
      application.currentAssessment.offering.offeringIntensity ===
      OfferingIntensity.fullTime
        ? [
            RestrictionActionType.StopFullTimeBCFunding,
            RestrictionActionType.StopFullTimeDisbursement,
          ]
        : [RestrictionActionType.StopPartTimeDisbursement];
    return studentRestrictionsThatAreNotBypassed
      .filter((studentRestriction) =>
        allowedRestrictionActions.some((actionType) =>
          studentRestriction.restriction.actionType.includes(actionType),
        ),
      )
      .map((studentRestriction) => ({
        studentRestrictionId: studentRestriction.id,
        restrictionCode: studentRestriction.restriction.restrictionCode,
        studentRestrictionCreatedAt: studentRestriction.createdAt,
      }))
      .sort((a, b) => a.restrictionCode.localeCompare(b.restrictionCode));
```

andrewsignori-aot (approved):
Thanks for making all the changes, looks good 👍

Issue #2741, opened 2024-01-26
Ministry - Filter/Search Programs & Offerings


CarlyCotton:
**User Story**
As a _ministry user_, I need to search for programs by more than just name within each institution to be able to (quickly) assist Institutions. This involves having a way to search for all institutions or search by institution location code AND being able to filter when viewing/searching programs when viewing the programs of an institution.

**Acceptance Criteria**

- [x] Update the Programs view for ministry users within an Institution (Figure 1):
    - [ ] Add sorting options to the fields to allow ASC/DESC (Program Name, Location name, Status[Pending,Approved,Declined,Inactive])
    - [ ] Add new search options to allow better filtering: Program Name AND Location name AND Status (multiple choice option) with search button
- [ ] If this component is shared with institutions, it's okay to update the institution side too.
- [ ] Search options for Status should load with all options selected. If none are selected, show no programs.
- [x] Sorting by status will change the default when the page loads (priority) to an alphabetical sort of the status names. 

**Business Context**
_Figure 1_
<img width="1427" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1YrQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8a9c7492b4faf4ffdd0f7dd7cea314d60ddb450c/image.png" alt="image.png" />

**Technical**
- [x] Create a new input to allow the search by the location name.
- [x] Create a multi-select drop-down (or similar) to allow filter multiple statuses. By default, all should be marked.
- [x] Add the search button.
- [x] All three search inputs will behave as an "and" logic if provided.
- [x] Convert the table to use Vuetify.
- [x] Sort program statuses as below (the below order was redefined during development by new business ACs).
  - Pending
  - Approved
  - Declined
  - Inactive
- [ ] Ensure at least one E2E will be present to test the data retrieval.

CarlyCotton:
QOL: SABC code, filter by location, better searching options

CarlyCotton:
Met with IPDT Aug 21 to confirm details of the enhancements in this ticket.

andrewsignori-aot:
@CarlyCotton is there a particular order for the statuses?
- They are the below ones:
- Creation pending
- Approved
- Creation declined
- Change under review
- Change awaiting approval
- Change overwritten
- Change declined

CarlyCotton:
[...]

PR #3921, opened 2024-11-12
#2741 - Ministry Filter Search Programs and Offerings [Fix Expired Programs Sort Issue]

sh16011993:
As a part of this PR, the query to retrieve the list of programs was updated to ensure that the sort order of the retrieved list of programs has the correct order where the inactive programs comprise the programs that have `isActive = false` or the ones that have the `effectiveEndDate <= CURRENT_DATE`

### DIFF
--- a/sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
+++ b/sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
@@ -901,7 +901,7 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
     // sort
     if (paginationOptions.sortField === "programStatus") {
       paginatedProgramQuery.orderBy(
-        `CASE WHEN programs.isActive = false THEN '${INACTIVE_PROGRAM}' ELSE programs.programStatus :: text END`,
+        `CASE WHEN programs.isActive = false or (programs.effectiveEndDate is not null and programs.effectiveEndDate >= CURRENT_DATE) THEN '${INACTIVE_PROGRAM}' ELSE programs.programStatus :: text END`,
         paginationOptions.sortOrder,
       );
     } else if (paginationOptions.sortField && paginationOptions.sortOrder) {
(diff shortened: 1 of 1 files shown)


sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
```diff
@@ -914,10 +914,10 @@ export class EducationProgramService extends RecordDataModelService<EducationPro
       // TODO:Further investigation needed as the CASE translation does not work in orderby queries.
       paginatedProgramQuery.orderBy(
         `CASE           
-          WHEN programs.programStatus = '${ProgramStatus.Pending}' and programs.isActive = true THEN ${SortPriority.Priority1}
-          WHEN programs.programStatus = '${ProgramStatus.Approved}' and programs.isActive = true THEN ${SortPriority.Priority2}
-          WHEN programs.programStatus = '${ProgramStatus.Declined}' and programs.isActive = true THEN ${SortPriority.Priority3}
-          WHEN programs.isActive = false THEN ${SortPriority.Priority4}
+          WHEN programs.programStatus = '${ProgramStatus.Pending}' and programs.isActive = true and (programs.effectiveEndDate is null OR programs.effectiveEndDate > CURRENT_DATE) THEN ${SortPriority.Priority1}
```
sh16011993, line 917: If the program's effective end date is today's date, then is the program valid till today or has it expired starting today?

andrewsignori-aot (approved):
Change looks and we should go ahead with it. The logic with the active and expiration is just something to be improved. Forcing it as a regular status just brings complexities that don`t feel worth it.

On the SQL command, there are SQL reserved worked in upper case and others in lower case.
We should try to stick with one or another IMO.=, not a blocker, hence approving.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
andrepestana-aot, line 917: Please remove the TODO comment.

andrepestana-aot (approved):
Looks good. Just a minor comment to remove the TODO comment.

sources/packages/backend/apps/api/src/services/education-program/education-program.service.ts
andrewsignori-aot: This is the logic to be considered active: `effectiveEndDate > CURRENT_DATE`.

sh16011993: Sounds good. Thanks.

Issue #3896, opened 2024-11-05
Mask Institution User Income View 


ninosamson:
**User Story**
As a ministry, we need to comply with our CRA MOU.

**Acceptance Criteria**
Options: 
- [ ] Mask the Total income used for assessment purposes on the assessment with X's for BC Public Institutions (see screen shots below)

<img width="945" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd20xQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8bcdffe8df802990531e38665ba49c8e663ca08d/image.png" alt="image.png" />

<img width="1176" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHk1Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c32254617fc4721a7281b03872a7b761fafed408/image.png" alt="image.png" />

<img width="1152" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDI1Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--52f16cc6df7c9a56da43b4c8a0ebf5645080c052/image.png" alt="image.png" />

Joshua-Lakusta:
Can we make this configurable and have a toggle so when it inevitably get reversed or changed in the future its easy to turn back on?

ninosamson:
Now you're talking @Joshua-Lakusta

andrewsignori-aot:
@CarlyCotton @ninosamson I am not planning to demo this one since it is part of the 1.15 release to be validated by QA.

ninosamson:
Understood @andrewsignori-aot

PR #3932, opened 2024-11-13
#3896 - Mask Institution User Income View

andrewsignori-aot:
- Masked the `totalFamilyIncome` following the same approach used to mask the MSFAA.
- Masked only for public institutions.
- The method used to format the money in the NOA UI when it is a number has no issues with the "masked" string.
image

### Sample UI with the masked value

image


sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
```diff
@@ -88,8 +94,13 @@ export class AssessmentControllerService {
       );
     }
 
+    const assessmentDTO = assessment.assessmentData as AssessmentAPIOutDTO;
```
dheepak-aot, line 97: we can do it like this without force casting. 

```ts
const assessmentDTO: AssessmentAPIOutDTO = assessment.assessmentData;
    if (maskTotalFamilyIncome) {
      assessmentDTO.totalFamilyIncome = MASKED_MONEY_AMOUNT;
    }
```

dheepak-aot (approved):
Looks good. One suggestion. 👍

sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
andrewsignori-aot: The cast is necessary to allow it to accept a string.
image

sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.students.controller.ts
```diff
@@ -66,6 +66,7 @@ export class AssessmentStudentsController extends BaseController {
     return this.assessmentControllerService.getAssessmentNOA(assessmentId, {
       studentId: userToken.studentId,
       maskMSFAA: false,
+      maskTotalFamilyIncome: false,
```
guru-aot, line 69: 👍

guru-aot (approved):

sources/packages/backend/apps/api/src/route-controllers/assessment/assessment.controller.service.ts
andrewsignori-aot: Changed as suggested.

PR #3938, opened 2024-11-14
#3691 - Fix for the cyclic dependency

andrepestana-aot:
Fix for the circular dependency error caused by anonymous TypeScript type.
image


andrewsignori-aot (approved):

dheepak-aot (approved):
Thanks for taking care. 👍

Issue #3789, opened 2024-10-11
BUG: Student and Institution: PIR: Program Name and Program Description from student request not displayed for institution or for student (View Application)


astridSABC:
**Describe the Bug**
The PIR (Programs) program name and program description created by the student are not displayed in the Student View Application or in the Institution PIR view for the request.  

The institution won't know what program that the student is requesting.

**Expected Behaviour**

**Student View Application**
Program name that you entered is displayed
Program description that you entered is displayed
Start date and end dates that you entered are displayed
TBD: Select your study dates section is not displayed as this is a program request

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1ZBQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--edab84b9052be9830a80981b6cb71eee34f34a40/image.png" alt="image.png" />

**Institution view**
Program name that the student entered is displayed
Program description that the student entered is displayed
Start date and end dates that the student entered are displayed

**Actual Behaviour**

**Student view**:
Program name that the student entered is not displayed. ‘Enter the name of your program’ is displayed
Program description field is blank
Start date and end dates are displayed
Select your study dates and field is not displayed/disabled

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTVBQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2cf80e61cbbfd0fcb478eb452274f0894f214663/image.png" alt="image.png" />

**Institution view**:
Program name that the student entered is not displayed
Program description that the student entered is not displayed

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWRBQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4468dab59bbf1a75a86e71caf84ee399a9183216/image.png" alt="image.png" />

**Steps To Reproduce**

**Precondition**:
BC Public institution (designated)
Student account with verified SIN

1. Login to student portal
2. Click “Start Application” button
3. Select program year "(2024 – 2025 – My first…)" in the program year dropdown list
4. Click “Start Application” button
5. Select an institution from the select your institution dropdown list
[...]

PR #3944, opened 2024-11-15
#3789 - PIR: Program Name and Program Description Bug

sh16011993:
**As a part of this PR, the bug for the PIR where the Program Name and the Program Description are not showing up once the application is submitted is fixed.** 

**Fix:** The bug was happening because both the fields (Name and Description) were having a conditional set as below:

```
"conditional": {
  "show": true,
  "when": "programnotListed",
  "eq": "true"
},
```

Removed the above piece of code for both the program name and program description for all the program years.

The problem is that `programnotListed` never has the value `true`. It is a `Select Boxes Component` rather than the `Checkbox Component` which gets the value `programnotListed`. We do not need to rely on this component because both the components `Program Name` and `Program Description` are wrapped inside a container component that has a conditional setup on it. 

Tested the above for the below scenarios:

**1.  Program is not listed**

image

-------------------------------------------------------------------------------------------------------------

**2.  Study is not listed**

image


andrewsignori-aot (approved):
Nice work, looks good 👍 
Please ensure the demo will show both PIRs displaying the values 😉

guru-aot (approved):
LGTM, nice work @sh16011993

PR #3964, opened 2024-11-19
#3789 - Update the form IO conditions based on form data

dheepak-aot:
## Updated the condition to render the program name and program description

- [ ] Updated the condition to render the program name and program description fields only when a program is not selected in the student application.
- [ ] The input which says `Your program is not listed` is a select box and not check box. So we can use only custom condition to evaluate its output at it will be in an object form like this. 

## Updated the read only selected program and offering display

- [ ] Updated the read only selected program and  selected offering display only when the respective value is present.

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2022-23.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2022-23.json
@@ -1742,7 +1742,8 @@
                     "threshold": 0.3
                   },
                   "id": "eopfm7h",
-                  "defaultValue": ""
+                  "defaultValue": "",
+                  "isNew": false
                 },
                 {
                   "label": "Select your program of study:",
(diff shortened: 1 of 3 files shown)


sources/packages/forms/src/form-definitions/sfaa2022-23.json
```diff
@@ -2147,7 +2150,14 @@
...
                       "truncateMultipleSpaces": false,
                       "id": "ejhdok9",
-                      "tags": []
+                      "tags": [],
+                      "conditional": {
+                        "show": "",
+                        "when": null,
+                        "eq": ""
+                      },
+                      "customConditional": "show = data.myProgramNotListed ? data.myProgramNotListed.programnotListed : false;",
```
andrewsignori-aot, line 2159: Should the `data.myProgramNotListed.programnotListed` be converted to bool as 
```js
show = data.myProgramNotListed ? !!data.myProgramNotListed.programnotListed : false;
```

andrewsignori-aot (approved):
Juts a minor question, not a blocker hence approving it.

sources/packages/forms/src/form-definitions/sfaa2022-23.json
dheepak-aot: data.myProgramNotListed.programnotListed is a boolean itself. Still do we require a conversion?

PR #3969, opened 2024-11-20
#3691 - UX/UI for Ignore Restriction - API bugfix

andrepestana-aot:
Fix for when the student doesn't have any student restrictions in the API.


dheepak-aot (approved):
Looks good 👍

PR #3977, opened 2024-11-20
#3691 - UX/UI for Ignore Restriction - Bypass Add/Remove - Added  "Stop part time BC funding" action type

andrepestana-aot:
- Added StopPartTimeBCFunding restriction action type to the query that gets the available student restrictions to be bypassed.


andrewsignori-aot (approved):
Thanks for quickly changing it, looks good 👍

dheepak-aot (approved):
Thanks for the quick update. 👍

Issue #3667, opened 2024-08-29
Bull Scheduler increments the next scheduled job when manually promoted


dheepak-aot:
**Describe the task**
When a schedule job of the scheduler(Bull Queue) is manually promoted, then the next schedule is automatically incremented .

This seems to be the default behavior of the bulk framework. 

But our expectation is that manually promoting the job is to handle something on a manual way if there is a need and it should affect the regular schedules.

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

PR #3978, opened 2024-11-21
#3667 - Bull Scheduler increments the next scheduled job when manually promoted (Ensure Next Delayed Job)

andrewsignori-aot:
- Ensures a scheduled job will have a delayed job created with the next expected scheduled time based on the configured cron expression.
- DB row used as a lock to ensure the initialization code will not recreate delayed jobs or create them in an unwanted way if two PODs start at the same time. To be clear, the lock approach is the same used in other areas of the application and locks a single row in the table queue-consumers, please not, the table will not be locked. Please see below the sample query executed.
```sql
START TRANSACTION
select
	"QueueConfiguration"."id" AS "QueueConfiguration_id"
from
	"sims"."queue_configurations" "QueueConfiguration"
where
	(("QueueConfiguration"."queue_name" = 'archive-applications')) LIMIT 1 FOR UPDATE 
COMMIT
```
- While the queue is paused, no other delayed jobs will be created.


sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -84,29 +87,81 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
...
+   * Gets the next scheduled date for the given cron expression.
+   * @param cron cron expression string.
+   * @returns the next date the scheduler will be executed.
+   */
+  async getNexSchedulerExecutionDate(): Promise<Date> {
+    const repeatOptions = await this.queueCronConfiguration();
+    const result = cronParser.parseExpression(repeatOptions.cron, {
+      utc: true,
+    });
+    return result.next().toDate();
```
dheepak-aot, line 152: We can return milliseconds here `result.next().getMilliseconds()` and do a millisecond comparison between `nextSchedulerExecutionDate` and `delayedJobExecutionDate`

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -84,29 +87,81 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
...
+  }
+
+  /**
+   * Calculates the execution date for a delayed job based on its options.
+   * @param delayedJob job instance containing delay options.
+   * @returns calculated execution date for the delayed job.
+   */
+  getDelayedJobExecutionDate(delayedJob: Job): Date {
+    const totalMilliseconds = delayedJob.opts.delay + delayedJob.timestamp;
```
dheepak-aot, line 161: It is a great analysis to consider the delay to the job timestamp. ❤️

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
```diff
@@ -84,29 +87,81 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
...
+      const nextExecution = this.getDelayedJobExecutionDate(delayedJob);
+      return dayjs(nextExecution).isSame(nextExpectedScheduledJobDate);
+    });
+    // If the only delayed job is the expected one, no further verifications are needed.
+    if (expectedDelayedJob && delayedJobs.length === 1) {
+      return;
+    }
+    // Remove any non expected delayed job.
+    for (const delayedJob of delayedJobs) {
+      if (!expectedDelayedJob || delayedJob !== expectedDelayedJob) {
```
dheepak-aot, line 125: Asking for my understanding, will comparing the job ids alone work ? `delayedJob.id !== expectedDelayedJob.id`

sources/packages/backend/libs/services/src/queue/queue.service.ts
```diff
@@ -125,4 +126,27 @@ export class QueueService {
...
+   * a callback to be executed with a guarantee that no other code applying
+   * the the lock for the queue will be executed concurrently.
+   * @param queueName queue name to have the lock acquired.
+   * @param callback method to be executed inside the lock.
+   */
+  async acquireQueueLock(
+    queueName: QueueNames,
+    callback: () => Promise<void>,
+  ): Promise<void> {
+    await this.dataSource.transaction(async (entityManager) => {
```
dheepak-aot, line 141: 👍

dheepak-aot:
Great work and analysis. Just 2 minor comments.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
andrewsignori-aot: The idea here is to compare the object reference, which will be pretty precise.
Is there a benefit using the id?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
andrewsignori-aot: It would work but IMO it would be clear to have a proper date.
Is the idea to avoid using the dayjs?
I will give it a try to see if the readability is still good enough.

andrepestana-aot (approved):
Great analysis and implementation! Looks good to me!

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
andrewsignori-aot: The `getMilliseconds` does not contain the total milliseconds.
I would prefer to keep the current implementation unless it would be a bloker.

andrewsignori-aot: Tested with `getDate` (similar to the Date) and it seems all good.

dheepak-aot (approved):

Issue #3943, opened 2024-11-14
Bug - restriction view when access limited


Joshua-Lakusta:
**Describe the Bug**
When you view restriction you get trapped and cant close the view when you do not have access to update the restriction.

**Expected Behaviour**
The 'Close' button is not disabled for all users. Users without restriction edit access should be able to view it, not edit it and then close the modal without refreshing.

**Actual Behaviour**
The 'Close' button is disabled for users without the restriction edit access. They need to refresh the page in order to get out of the modal.

** Steps To Reproduce**
Steps to reproduce the behaviour:
1. Go to a student with restrictions as a ministry user with no restriction edit access.
2. Click on 'View' for any restriction
3. Un-closable modal reached.

<img width="1478" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeVBBQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6ff9a48850f060d97a0be6a3594790339fd53ef1/image.png" alt="image.png" />

PR #3982, opened 2024-11-21
#3943 - Bug - restriction view when access limited 

lewischen-aot:
- Removed the check-role-permission decorator to allow ministry users without designated roles to close the modal as per AC
  - The 'Close' button is not disabled for all users. Users without restriction edit access should be able to view it, not edit it and then close the modal without refreshing.

Screenshot
image

### DIFF
--- a/sources/packages/web/src/components/common/restriction/ViewRestriction.vue
+++ b/sources/packages/web/src/components/common/restriction/ViewRestriction.vue
@@ -73,14 +73,19 @@
         </template>
       </template>
       <template #footer>
-        <footer-buttons
-          :processing="processing"
-          :primaryLabel="allowUserToEdit ? 'Resolve restriction' : 'Close'"
-          secondaryLabel="Cancel"
-          @primaryClick="allowUserToEdit ? submit() : cancel()"
-          @secondaryClick="cancel"
-          :showSecondaryButton="allowUserToEdit"
-        />
+        <check-permission-role :role="allowedRole">
+          <template #="{ notAllowed }">
+            <footer-buttons
+              :processing="processing"
+              :primaryLabel="allowUserToEdit ? 'Resolve restriction' : 'Close'"
+              secondaryLabel="Cancel"
+              @primaryClick="allowUserToEdit ? submit() : cancel()"
+              @secondaryClick="cancel"
+              :disablePrimaryButton="allowUserToEdit && notAllowed"
+              :showSecondaryButton="allowUserToEdit"
+            />
+          </template>
+        </check-permission-role>
       </template>
     </modal-dialog-base>
   </v-form>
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/components/common/restriction/ViewRestriction.vue
```diff
@@ -160,6 +153,8 @@ export default defineComponent({
       return "Resolution reason is required.";
     };
 
+    console.info("props.canResolveRestriction: ", props.canResolveRestriction);
```
dheepak-aot, line 156: Please remove the console.info.

sources/packages/web/src/components/common/restriction/ViewRestriction.vue
```diff
@@ -73,19 +73,14 @@
...
-              :primaryLabel="allowUserToEdit ? 'Resolve restriction' : 'Close'"
-              secondaryLabel="Cancel"
-              @primaryClick="allowUserToEdit ? submit() : cancel()"
-              @secondaryClick="cancel"
-              :disablePrimaryButton="notAllowed"
-              :showSecondaryButton="allowUserToEdit"
-            />
-          </template>
-        </check-permission-role>
+        <footer-buttons
```
dheepak-aot, line 76: Check permission role is required and must not be removed.

sources/packages/web/src/components/common/restriction/ViewRestriction.vue
```diff
@@ -73,19 +73,14 @@
...
+          :processing="processing"
+          :primaryLabel="allowUserToEdit ? 'Resolve restriction' : 'Close'"
```
dheepak-aot, line 78: I would suggest the following quick solution which would keep the primary behavior as primary and secondary behaviour as secondary not having to juggle based on the situation.

**Solution**

```vue
<template #footer>
        <check-permission-role :role="allowedRole">
          <template #="{ notAllowed }">
            <footer-buttons
              :processing="processing"
              primaryLabel="Resolve restriction"
              :secondaryLabel="allowUserToEdit ? 'Cancel' : 'Close'"
              @primaryClick="submit"
              @secondaryClick="cancel"
              :disablePrimaryButton="notAllowed"
              :showPrimaryButton="allowUserToEdit"
            />
          </template>
        </check-permission-role>
      </template>
```

Result: 

View Active Restriction with permission

image


View resolved restriction with permission

image

View Resolved restriction without permission

image

sources/packages/web/src/components/common/restriction/ViewRestriction.vue
```diff
@@ -73,19 +73,14 @@
...
+          secondaryLabel="Cancel"
+          @primaryClick="allowUserToEdit ? submit() : cancel()"
+          @secondaryClick="cancel"
+          :showSecondaryButton="allowUserToEdit"
```
dheepak-aot, line 82: Or as a bare minimal solution. 

image

dheepak-aot:

sources/packages/web/src/components/common/restriction/ViewRestriction.vue
lewischen-aot: I agree with this solution, which is adopted in the latest commit.

dheepak-aot (approved):
Looks good @lewischen-aot 👍

andrepestana-aot (approved):
👍

sources/packages/web/src/components/common/restriction/ViewRestriction.vue
lewischen-aot: The solution looks great, and we can probably adopt this in other stories. Thanks @dheepak-aot

PR #3989, opened 2024-11-22
#3667 - Bull Scheduler - Updated ioredis and Added Logs

andrewsignori-aot:
To proceed with the investigation about the intermittent ioredis issue on DEV, the package was updated and more logs were added to try to narrow down the root cause of the issue.


dheepak-aot (approved):
👍

Issue #3258, opened 2024-05-09
CAS Integration 4A - Supplier Maintenance Errors


HRAGANBC:
### **User Story**
As a Ministry user, I need to know if SIMS could not successfully verify and assign a supplier number and site, this could be due to an identified manual intervention requirement or due to an error with the request to CAS. This needs to provide me context in which this occurred so I know if I can retry or resolve it. Retrying involves asking SIMS to attempt the action again with the current information (eventually). Resolving means that I have reviewed CAS and SIMS, making any necessary updates in either system and SIMS will not need to attempt the action again. This information will be used to create UI or populate a report.

# Acceptance Criteria
- [x] Update the CAS table column order, add new columns, update column names (ignore differences in the status label or formatting):
<img width="1001" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkdmQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a393bd5619b71669f0a2075fc2248488be9866e4/image.png" alt="image.png" />

- [x] User should see the response message for any CAS error or manual intervention pre-validation error records. 

- [x] If Manual Intervention is required the associated **Error message** based on the situation:
    - [x]  "Address outside of Canada cannot be processed." 
    - [x]  "Given name required for creation."

- [x] For error responses from CAS set the status to "Manual Intervention" include the **Error Message** received from CAS (it comes in as a string, multiple are split by bars. See CAS 2A)
   

- [x] Any of these errors should make the Supplier `isValid = false` and in cases where Supplier found but create Site failed, save any valid Supplier info

- [x] Retry button means creating a new pending CAS request  and is only available to the `business-administrator` group (can be linked to the existing cas-edit-supplier role)
    - [x] Student data may be updated (Given Name, Last Name, SIN, Address Line 1, City, etc) and we need to perform the same attempt again with the most current student info.
    - [x] On retry provide success toast pop message "A new CAS pending verification record was created. A new attempt will be made to get information from CAS."

- [x] Include all error responses, including those from GetSupplier requests.

**Business Context**
- Identification of almost certain errors are noted as Manual Interventions. These will not be able to be resolved through automatic SIMS-CAS communication and will require a user to review CAS and update CAS and/or SIMS as a result.
- Some errors are due to technical issues and not as a result of an error response from CAS. These will be automatically retried right away.
[...]

PR #3991, opened 2024-11-22
#3258 - CAS supplier errors - DB Migration

guru-aot:
image

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/sql/CASSuppliers/Add-errors-column.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/CASSuppliers/Add-errors-column.sql
@@ -1,0 +1,6 @@
+ALTER TABLE
+    sims.cas_suppliers
+ADD
+    COLUMN errors VARCHAR(300) [];
+
+COMMENT ON COLUMN sims.cas_suppliers.errors IS 'Errors received from CAS.';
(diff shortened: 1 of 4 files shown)


sources/packages/backend/apps/db-migrations/src/sql/CASSuppliers/Add-errors-column.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    sims.cas_suppliers
+ADD
+    COLUMN errors VARCHAR(300) [];
```
andrewsignori-aot, line 4: The business has a list of all possible errors. Since this is not under our control I would suggest checking the list or increasing the number.

sources/packages/backend/apps/db-migrations/src/sql/CASSuppliers/Add-errors-column.sql
```diff
@@ -0,0 +1,6 @@
...
+
+COMMENT ON COLUMN sims.cas_suppliers.errors IS 'Errors received from CAS.';
```
andrewsignori-aot, line 6: Please adjust to "Errors while execution CAS verifications.".
The errors will also be generated by SIMS even without calling CAS.

sources/packages/backend/apps/db-migrations/src/sql/CASSuppliers/Add-errors-column.sql
guru-aot: As part of the error pdf, the max length error is 110 characters, so i think we are safe.

andrewsignori-aot (approved):
Looks good 👍

PR #3992, opened 2024-11-22
#3667 - Reuse Redis Connections

andrewsignori-aot:
During the prior investigation was detected that when all the schedulers try to execute redis operations at the same time it fails after a certain amount of concurrent connections.
The possible solutions being investigated are:
1. Allow the connection sharing as mentioned here https://github.com/OptimalBits/bull/blob/develop/PATTERNS.md#reusing-redis-connections
2. Control the service initialization to prevent multiple concurrent connections.

This is PR an attempt to test the first options.


guru-aot (approved):
LGTM, nice Work @andrewsignori-aot

dheepak-aot (approved):
Looks good 👍

PR #3993, opened 2024-11-25
#3667 - Bull Scheduler - Added Global Bootstrap Synchronization

andrewsignori-aot:
Implementing the easy/fast (and stable) approach to resolve the queue initialization issues.

During the previous approach, sharing the ioredis connections (change being reverted in this PR) worked but there was also a possible false-positive memory leak warning. Further investigation will be needed if we change this approach in the feature but for now, keeping the existing approach.

The current solution extends the current  "specific queue-based" lock to an "all queues-based" lock.
The initialization is still pretty fast (around 1 second) and even if we double the number of schedulers in the future it still will be good enough (the code is executed once during queue-consumers initialization).

_Notes_: The logs before mentioned the queue-name every time and now it has changed to only the first one. The queue-name should be in the log context also but many schedulers are not "overriding" it, which should be resolved in the schedulers.

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-scheduler.ts
@@ -86,34 +86,27 @@ export abstract class BaseScheduler<T> implements OnApplicationBootstrap {
       await this.schedulerQueue.obliterate({ force: true });
       return;
     }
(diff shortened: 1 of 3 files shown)


dheepak-aot (approved):
Thanks again for the investigations and the solution looks good to me. .👍

guru-aot (approved):
LGTM

Issue #3666, opened 2024-08-28
Queue Monitoring


ninosamson:
**User Story**
As a product team, we need to monitor the queues and associated SFTP folders to diagnose errors in processing the files, and confirm archiving behaviour for failed files. 

**Acceptance Criteria**
- [ ] Review monitoring processes queues (and associated SFTP folders) for errors, failures. 
- [x] Investigate and implement basic alerting as available through bull framework. 
- [x] If not available through Bull, create new ticket for implementation of alerting. 
- [ ] Nice to have.
  - [x] Create some POC towards the acceptable plan.
  - [x] Include the team in the discussions and get the team's understanding.
  - [x] Check the possibility of using global event handlers to send notifications, for instance, create an email.
  - [x] Check if sysdig can get some status from the queues.
  - [ ] Check queues failing silently and the effort to force them to fail in the dashboard. If possible have them fixed.

**Note for IMB team:**
- Following this ticket, SIMS IMB team will develop recommended process for investigation, tracking of incidents/tickets and retries to address failed jobs/files. 
- Also need to confirm and implement desired behaviour with regards to failed files (archive or move to another folder). 
    - Example: Currently PT and FT feedback integration files are not being archived if they are erroring, even though they are non-sequential. 
    
**@michesmith to create epic to track this and subsequent tickets.**

PR #4012, opened 2024-11-28
#3666 - Queue Monitoring - Enable Prometheus Metrics

andrewsignori-aot:
- Enabled `/metrics` endpoint on `queue-consumers` to expose Prometheus metrics following [BC Gov docs](https://developer.gov.bc.ca/docs/default/component/platform-developer-docs/docs/app-monitoring/user-defined-monitoring/).
- Added [prom-client](https://github.com/siimon/prom-client) also following the BC Gov docs mentioned above and the one that seems largely adopted.
- Created a [Gauge metric](https://prometheus.io/docs/concepts/metric_types/#gauge) to capture the most recent summary of every active queue allowing to add to the metrics a summary of `active`, `completed`, `failed`, `delayed`, and 'waiting' jobs. This represents the same status observed in the Bull Board. The garage metrics rely on Redis queries and will always get the most updated values from Redis.
image
- Created a [Counter metric](https://prometheus.io/docs/concepts/metric_types/#counter) to capture all the local events triggered for a queue. This happens only in-memory and is captured every time the `metrics` endpoint is invoked. This metric is not needed to achieve queue monitoring but seems a great addition and useful data to support future analysis.
- Enable default [nodejs metrics](https://github.com/siimon/prom-client?tab=readme-ov-file#default-metrics) following also [Prometheus recommendations](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors).
- Both metrics are incremented/set using the labels `queueName`, `queueEvent`, and `queueType` to allow querying in Sysdig.

## Sysdig POCs

The Sysdig configurations are not final and were created to support the validation of the code in this PR but should not be considered final or part of the PR evaluation.

### Alerts generated for a queue with a failed job.

```
max(queue_job_counts_current_total{queueEvent="failed",kube_namespace_name="0c27fb-dev"}) by (queueName) > 0
```

image

image

### Same alerts are configured to be sent using an email channel.

image

## Sample Dashboard

The [Queues Overview](https://app.sysdigcloud.com/#/dashboards/419533?last=3600&scope=kubernetes.namespace.name%20%3D%20%3F%220c27fb-dev%22
) dashboard has some examples of data but should not be considered final or part of the PR evaluation.

image

_Note:_ the sysdig users and roles were updated and added to this PR, and it is already deployed to both tools environments. If time allows, further effort can be made to enhance the current process but any action beyond the user list update is not part of this PR.

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/queues-metrics.module.module.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/queues-metrics.module.module.ts
@@ -16,7 +16,6 @@ export class QueuesMetricsModule implements OnModuleInit {
    */
   async onModuleInit(): Promise<void> {
     this.logger.log("Associating queue events for metrics.");
-
     this.metricsService.setGlobalMetricsConfigurations();
     await this.metricsService.associateQueueEventsCountersMetrics();
   }
--- a/sources/packages/backend/apps/queue-consumers/src/queues-metrics.module.module.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/queues-metrics.module.module.ts
@@ -1,6 +1,6 @@
 import { Global, LoggerService, Module, OnModuleInit } from "@nestjs/common";
 import { InjectLogger } from "@sims/utilities/logger";
-import { MetricsService } from "apps/queue-consumers/src/services";
+import { MetricsService } from "./services";
 
 @Global()
 @Module({
--- a/sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
@@ -6,13 +6,11 @@ import { processInParallel } from "@sims/utilities";
 import { InjectLogger } from "@sims/utilities/logger";
 import {
   DEFAULT_METRICS_APP_LABEL,
-  DEFAULT_JOBS_COUNTS_GAUGE,
-  DEFAULT_JOBS_EVENTS_COUNTER,
   QueuesMetricsEvents,
   MonitoredQueue,
 } from "./metrics.models";
 import { Queue } from "bull";
-import { register, collectDefaultMetrics } from "prom-client";
+import { register, collectDefaultMetrics, Gauge, Counter } from "prom-client";
 
 @Injectable()
 export class MetricsService {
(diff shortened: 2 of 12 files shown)


sources/packages/backend/apps/queue-consumers/src/queues-metrics.module.module.ts
```diff
@@ -0,0 +1,25 @@
...
+import { InjectLogger } from "@sims/utilities/logger";
+import { MetricsService } from "./services";
+
+@Global()
+@Module({
+  providers: [MetricsService],
+  exports: [MetricsService],
+})
+export class QueuesMetricsModule implements OnModuleInit {
+  constructor(private readonly metricsService: MetricsService) {}
```
dheepak-aot, line 11: Do we require this un-used service at constructor?

andrewsignori-aot: Got it now. The comment was added to the wrong file. I will remove it from the controller.

sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Get the list of monitored queues.
+   * If not loaded yet, load it once.
+   * @returns list of monitored queues.
+   */
+  private async getMonitoredQueues(): Promise<MonitoredQueue[]> {
+    if (!this.monitoredQueueProviders) {
+      const queues = await this.queueService.queueConfigurationModel();
+      this.monitoredQueueProviders = queues
+        .filter((queueModel) => queueModel.isActive)
```
dheepak-aot, line 140: Are we not ignoring the FT queues here?

sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      labelNames: ["queueName", "queueEvent", "queueType"] as const,
+    });
+  }
+
+  /**
+   * Set global metrics configurations.
+   */
+  setGlobalMetricsConfigurations(): void {
+    register.setDefaultLabels({ app: DEFAULT_METRICS_APP_LABEL });
+    collectDefaultMetrics({ labels: { app: DEFAULT_METRICS_APP_LABEL } });
```
dheepak-aot, line 60: for collect default metrics, is it not necesscary to set the register. 

```ts
collectDefaultMetrics({
      labels: { app: DEFAULT_METRICS_APP_LABEL, register },
    });
```

sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
andrewsignori-aot: The full-time toggle is not checked here. If the queues are not executing they will not produce any event.
The events will be associated but never triggered. I do not see any harm in it. Please let me know if further discussion would be required on this.

sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
andrewsignori-aot: Yes, there is no need to set the register unless using a custom one.

image
https://github.com/siimon/prom-client?tab=readme-ov-file#default-metrics

sources/packages/backend/apps/queue-consumers/src/services/metrics/metrics.service.ts
dheepak-aot: Not required and I am good based on what we discussed.

dheepak-aot (approved):
Great work and amazing solution to share metrics to Sysdig. 

Thanks for doing the change. 👍

sh16011993 (approved):
Great Work @andrewsignori-aot 👍 Thank you for the PR walkthrough.

Issue #3981, opened 2024-11-21
Model SDPR API


ninosamson:
**User Story:** As StudentAidBC, we need to a mechanism to provide information for shared clients with SDPR, for program eligibility purposes.

**Acceptance Criteria:**
- [x] create new API and publish to separate swagger page 
    - [x] example route:   /api/external/student
- [x] parameters - SIN and DOB - NOTE: for POC we will only do SIN only. 
- [x] Response: 
    - [x] 200
        -  Given name
        -  Last name 
        -  SIN
        -  DOB
        -  Address
        -  application numbers (array)         
    - [x] 404 "Not Found" 
- [x] API authentication is clientid/secret 
- [x] Small Segregation from current API specifically for SDPR (Ministry of Social Development and Poverty Reduction)

**Additional Context**
Will need to add more data within the applications in future including data like dependent/independent & CRA income. Possibly allowing for an optional search by program year.

**Technical**
- [x] Create a new SDPR service account on Keycloak and have it documented in the [wiki](https://github.com/bcgov/SIMS/wiki/Keycloak-Config).
- [x] Create the new Nestjs module for the new endpoint. Consider creating it generic to be reusable for future clients.
- [x] E2E tests are out of the scope.

ninosamson:
After discussion, calling by SIN only is sufficient.

PR #4018, opened 2024-12-03
#3981 - Model SDPR API

andrepestana-aot:
- Created a new module for SIMS API from external access;
- Created a segregated swagger for external endpoints: http://localhost:3000/external/swagger ;
image
- Created Postman collection to authenticate and request a student by SIN (Teams folder);
image
image
- Documented on: https://github.com/bcgov/SIMS/wiki/Keycloak-Config#external-api-access-service-accounts


sources/packages/backend/apps/api/src/app.external.module.ts
```diff
@@ -0,0 +1,11 @@
+import { Module } from "@nestjs/common";
+import { StudentExternalController } from "./route-controllers";
+import { AuthModule } from "./auth/auth.module";
+import { StudentService } from "@sims/integrations/services";
+
+@Module({
+  imports: [AuthModule],
+  controllers: [StudentExternalController],
+  providers: [StudentService],
```
andrewsignori-aot, line 9: Open for discussion but we can keep those services at the API scope at first.
We can create some specific services for the external module, but I am not sure if we should start creating them in the integrations lib directly.

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+    super();
+  }
+
+  /**
+   * Searches for student student details.
+   * @param payload payload with sin to retrieve the student details.
+   * @returns student details.
+   */
+  @Post()
+  @ApiNotFoundResponse({ description: "Student not found." })
```
andrewsignori-aot, line 34: Please decorate the endpoint to ensure the right HTTP response code, since nothing is being created.
```
@HttpCode(HttpStatus.OK)
```

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+    const student = await this.studentService.getStudentBySIN(payload.sin);
+    if (!student) {
+      throw new NotFoundException("Student not found.");
+    }
+    return {
+      firstName: student.user.firstName,
+      lastName: student.user.lastName,
+      sin: student.sinValidation.sin,
+      dateOfBirth: student.birthDate,
+      address: student.contactInfo.address,
```
andrewsignori-aot, line 47: We should map also the address.

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+@Controller("student")
+@ApiTags(`${ClientTypeBaseRoute.External}-student`)
+export class StudentExternalController extends BaseController {
+  constructor(private readonly studentService: StudentService) {
+    super();
+  }
+
+  /**
+   * Searches for student student details.
+   * @param payload payload with sin to retrieve the student details.
```
andrewsignori-aot, line 30: Please use `SIN` instead of `sin`.

sources/packages/backend/libs/utilities/src/config/config.models.ts
```diff
@@ -1,4 +1,9 @@
-export type ClientType = "student" | "institution" | "aest" | "supportingUsers";
+export type ClientType =
+  | "student"
+  | "institution"
+  | "aest"
+  | "supportingUsers"
+  | "external";
```
andrewsignori-aot, line 6: I am not sure this one should be exposed in the API configs.
Was there a need for it or was it added for consistency?

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
```diff
@@ -0,0 +1,59 @@
...
+@AllowAuthorizedParty(AuthorizedParties.external)
+@Controller("student")
+@ApiTags(`${ClientTypeBaseRoute.External}-student`)
+export class StudentExternalController extends BaseController {
+  constructor(private readonly studentService: StudentService) {
+    super();
+  }
+
+  /**
+   * Searches for student student details.
```
andrewsignori-aot, line 36: These comments will be displayed on swagger, maybe it would be good to add some content on why it is a POST instead of a GET.

sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -31,6 +31,7 @@ export class ConfigService {
         institution: process.env.KEYCLOAK_CLIENT_INSTITUTION,
         aest: process.env.KEYCLOAK_CLIENT_AEST,
         supportingUsers: process.env.KEYCLOAK_CLIENT_SUPPORTING_USERS,
+        external: process.env.KEYCLOAK_CLIENT_EXTERNAL,
```
andrewsignori-aot, line 34: Is this variable going to be added to the CI/CD?

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -87,6 +88,12 @@ async function bootstrap() {
       .build();
     const document = SwaggerModule.createDocument(app, options);
     SwaggerModule.setup("swagger", app, document);
+
+    // External swagger
+    const externalDocument = SwaggerModule.createDocument(app, options, {
+      include: [AppExternalModule], // Includes only AppExternalModule.
+    });
+    SwaggerModule.setup("external/swagger", app, externalDocument);
```
andrewsignori-aot, line 96: We need to ensure this URL will be configured in the deployment config.

sources/packages/backend/libs/utilities/src/config/config.models.ts
andrepestana-aot: Yes, we don't need that. Thanks!

sources/packages/backend/apps/api/src/main.ts
andrepestana-aot: Yes, sorry. I forgot that. I'm going to add it.

sources/packages/backend/libs/utilities/src/config/config.service.ts
andrepestana-aot: Yes, sorry. I forgot that. I'm going to add it.

andrewsignori-aot: My understanding is that we will no longer use it and it was removed, hence resolving the comment 😉

sources/packages/backend/apps/api/src/app.external.module.ts
andrewsignori-aot: @dheepak-aot can you please collaborate on this?

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+import {
+  ExternalSearchStudentAPIInDTO,
+  StudentDetailsAPIOutDTO,
+} from "./models/student.dto";
+
+/**
+ * Student controller for external client.
+ */
+@RequiresUserAccount(false)
+@AllowAuthorizedParty(AuthorizedParties.external)
```
andrewsignori-aot, line 27: Adding the comment here to facilitate the discussion.
Can we please ensure that the token scopes that are not needed and also the non-needed mappers will be removed from the token?
The token should have only the necessary information.

For instance, can we remove the below.
image

sources/packages/backend/apps/api/src/auth/guards/authorized-parties.guard.ts
```diff
@@ -66,6 +66,9 @@ export class AuthorizedPartiesGuard implements CanActivate {
     authorizedParty: AuthorizedParties,
     identityProvider: IdentityProviders,
   ): boolean {
+    if (authorizedParty === AuthorizedParties.external) {
+      return true;
+    }
```
lewischen-aot, line 71: Having this condition here is fine, but we can also have it inside the switch statement, unless I'm missing the consideration not to do so:
```js
switch(authorizedParty) {
    ...
    case AuthorizedParties.external:
        return true;
    default:
        return false;
}
```

lewischen-aot:
Just a minor comment. Great work so far. Please take a look at the remaining comments.

sources/packages/backend/apps/api/src/app.external.module.ts
dheepak-aot: I agree with @andrewsignori-aot, current scope of the services that serve the external APIs can be in `API` scope.

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
andrepestana-aot: I removed all the other mappers that automatically come from the client and in `<client-id>-dedicated` there are some mappers (clientHost, clientAddress and clientId). I tried to remove them, if you refresh the page, they're still there. I think they are not optional.
image

andrepestana-aot: Even trying to remove `<client-id>-dedicated` is not allowed:
image

andrewsignori-aot (approved):
Great work, looks good 👍

lewischen-aot (approved):
LGTM 👍 Thanks for making the changes

PR #4020, opened 2024-12-03
#3666 - Queue Monitoring - Default Queue Error Handling

andrewsignori-aot:
This PR is a proposal to unify the queue behavior. It contains changes for start and cancel assessments regular queues and the idea is to expand it to all schedulers.

The below image shows these points to be shared and forced for all the queues, using the CAS scheduler as an example.
1 - Return type;
2 - Provide a `processSummary`;
3 - try/catch and default start message.
4 - Error check on `processSummary` to force the jobs to fail.
5 - Error handling on `catch`.
6 - Final log activities during `finally`.
image
_Note:_ The above will also ensure the E2E tests follow a closer approach.

Sample log for a queue using the new `BaseQueue<T>` class.

image

### Side effects
The idea is to have the BaseScheduler extend the BaseQueue which will force a change in every single scheduler (which would be the idea). To avoid the need to change every single scheduler and adapt every E2E, the methods required by BaseQueue can be implemented temporarily as below. The idea is to start the scheduler refactor still during this ticket.

```ts
/**
 * To be implemented.
 */
processQueue(_job: Job<void>): Promise<string | string[]> {
  throw new Error("Method not implemented.");
}

/**
 * To be implemented.
 */
async process(
  _job: Job<void>,
  _processSummary: ProcessSummary,
): Promise<string | string[]> {
  throw new Error("Method not implemented.");
}
```

### Error Fix
Fixed an issue with the logger service where `null` message was logged right after the regular message.

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/processors/assessment/cancel-application-assessment.processor.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/assessment/cancel-application-assessment.processor.ts
@@ -7,27 +7,32 @@ import {
   SystemUsersService,
   WorkflowClientService,
 } from "@sims/services";
-import { QueueNames } from "@sims/utilities";
+import { CustomNamedError, QueueNames } from "@sims/utilities";
 import { StudentAssessmentService } from "../../services";
 import { DataSource } from "typeorm";
 import { ZeebeGRPCError, ZeebeGRPCErrorTypes } from "@sims/services/zeebe";
-import { InjectLogger, LoggerService } from "@sims/utilities/logger";
 import {
-  QueueProcessSummary,
-  QueueProcessSummaryResult,
-} from "../models/processors.models";
+  InjectLogger,
+  LoggerService,
+  ProcessSummary,
+} from "@sims/utilities/logger";
 import {
   ApplicationStatus,
   COEStatus,
   StudentAssessment,
   StudentAssessmentStatus,
 } from "@sims/sims-db";
+import { BaseQueue } from "../../processors";
+import {
+  ASSESSMENT_NOT_FOUND,
+  INVALID_OPERATION_IN_THE_CURRENT_STATUS,
+} from "@sims/services/constants";
 
 /**
  * Process the workflow cancellation.
  */
 @Processor(QueueNames.CancelApplicationAssessment)
-export class CancelApplicationAssessmentProcessor {
+export class CancelApplicationAssessmentProcessor extends BaseQueue<CancelAssessmentQueueInDTO> {
   constructor(
     private readonly dataSource: DataSource,
     private readonly workflowClientService: WorkflowClientService,
@@ -35,22 +40,21 @@ export class CancelApplicationAssessmentProcessor {
     private readonly disbursementScheduleSharedService: DisbursementScheduleSharedService,
     private readonly assessmentSequentialProcessingService: AssessmentSequentialProcessingService,
     private readonly systemUserService: SystemUsersService,
-  ) {}
+  ) {
+    super();
+  }
 
   /**
    * Cancel the workflow instance and rollback overawards, if any.
    * @param job information to perform the process.
-   * @returns log message.
-   */
+   * @param processSummary process summary for logging.
+   * @returns processing result.   */
   @Process()
-  async cancelAssessment(
+  async process(
     job: Job<CancelAssessmentQueueInDTO>,
-  ): Promise<QueueProcessSummaryResult> {
-    const summary = new QueueProcessSummary({
-      appLogger: this.logger,
-      jobLogger: job,
-    });
-    await summary.info(
+    processSummary: ProcessSummary,
+  ): Promise<string> {
+    processSummary.info(
       `Cancelling application assessment id ${job.data.assessmentId}`,
     );
     const assessment = await this.studentAssessmentService.getAssessmentById(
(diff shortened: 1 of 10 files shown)


sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-queue.ts
```diff
@@ -0,0 +1,74 @@
...
+      const result = await this.process(job, processSummary);
+      processSummary.info(`${job.queue.name}, job ID ${job.id}, executed.`);
+      const logsSum = processSummary.getLogLevelSum();
+      if (logsSum.error) {
+        throw new CustomNamedError(
+          "One or more errors were reported during the process, please see logs for details.",
+          PROCESS_SUMMARY_CONTAINS_ERROR,
+        );
+      }
+      return result;
```
dheepak-aot, line 38: What about adding the attention message (`getSuccessMessageWithAttentionCheck`) when there is a warning like records ignored etc..

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-queue.ts
```diff
@@ -0,0 +1,74 @@
...
+   */
+  @Process()
+  async processQueue(job: Job<T>): Promise<string | string[]> {
+    const processSummary = new ProcessSummary();
+    try {
+      this.logger.log(`Processing queue ${job.queue.name}, job ID ${job.id}.`);
+      const result = await this.process(job, processSummary);
+      processSummary.info(`${job.queue.name}, job ID ${job.id}, executed.`);
+      const logsSum = processSummary.getLogLevelSum();
+      if (logsSum.error) {
```
dheepak-aot, line 32: With logSum errors now certainly resulting in retry(being thrown) we should have a clear understanding with the team about when to use warning and error messages.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-queue.ts
```diff
@@ -0,0 +1,74 @@
...
+  PROCESS_SUMMARY_CONTAINS_ERROR,
+  logProcessSummaryToJobLogger,
+} from "../../utilities";
+import { Job } from "bull";
+import { Process } from "@nestjs/bull";
+
+/**
+ * Provides basic functionality for queue processing.
+ */
+export abstract class BaseQueue<T> {
```
andrepestana-aot, line 17: It's good to have it in a abstract class. We can even have the logging for the job initialization and job ending in abstract methods if we that level of customization for some process. Good idea!

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-queue.ts
```diff
@@ -0,0 +1,74 @@
...
+   * Wrap the queue job execution in a try/catch for the global error handling.
+   * It provides also the default {@link ProcessSummary} for logging.
+   * @param job process job.
+   * @returns processing result.
+   */
+  @Process()
+  async processQueue(job: Job<T>): Promise<string | string[]> {
+    const processSummary = new ProcessSummary();
+    try {
+      this.logger.log(`Processing queue ${job.queue.name}, job ID ${job.id}.`);
```
dheepak-aot, line 28: Great initiative to centralize the common logic. ❤️

sources/packages/backend/apps/queue-consumers/src/processors/assessment/start-application-assessment.processor.ts
```diff
@@ -9,70 +9,60 @@ import {
...
+      `Starting assessment id ${job.data.assessmentId} using workflow ${workflowName}.`,
+    );
+    await this.workflowClientService.startApplicationAssessment(
+      workflowName,
+      job.data.assessmentId,
+    );
+    return "Workflow call executed with success.";
   }
 
   @InjectLogger()
```
dheepak-aot, line 68: I believe the logger is not required here anymore.

andrewsignori-aot: The logger should be added to every processor to ensure the right context is set (it is done during the property injection). Not having it causes the logs to be set to the base class context. We have many schedulers right now with this issue.

sources/packages/backend/apps/queue-consumers/src/processors/assessment/cancel-application-assessment.processor.ts
```diff
@@ -169,17 +172,16 @@ export class CancelApplicationAssessmentProcessor {
...
+          processSummary.info(
             "No impacts were detected on future applications.",
           );
         }
       }
-      await summary.info("Assessment cancelled with success.");
-      return summary.getSummary();
+      return "Assessment cancelled with success.";
     });
   }
```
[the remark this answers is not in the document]
andrewsignori-aot: Please check https://github.com/bcgov/SIMS/pull/4020#discussion_r1868424775

andrepestana-aot (approved):
Looks good to me. It will be very beneficial to the project and can be done in parts so I think we can at least start it in this sprint.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/base-queue.ts
andrewsignori-aot: It was supposed to 😉 
I am adding it now.

andrewsignori-aot: Just added it.

dheepak-aot (approved):
Great work and thanks for brining this organization of things that are done in Queue processors. Looks good 👍

PR #4038, opened 2024-12-06
#3981 - Model SDPR API - Query fix

andrepestana-aot:
- Querying the id is necessary for TypeOrm to bring multiple numbers for the `applications` array. Otherwise, it brings only one application in the array in the `student` object.

image


andrewsignori-aot (approved):
Looks good 👍

dheepak-aot (approved):
👍 Approving as it is very minor and not keep the PR waiting.

PR #4040, opened 2024-12-06
#3981 - Model SDPR API - Query improvement

andrepestana-aot:
- Added conditions to query only applications with an applicationNumber;
- Added order clause,


dheepak-aot (approved):
Looks good. Thanks for the change.

PR #4044, opened 2024-12-06
update cruncy resources

bidyashish:
Update Cruncy Resources in Production Openshift


guru-aot (approved):
LGTM, nice work @bidyashish

Issue #4013, opened 2024-11-29
Legacy Restriction Mapping


CarlyCotton:

CarlyCotton:
Based on discussion from Friday/Today - added a new AC:
- [ ] These new legacy restrictions will be shown to the student as a single Legacy restriction in Account Activity/Banners as the highest level of warning to the student. Example: 4 legacy restrictions. 1 error, 2 warning, 1 no effect. Only error level messaging would be shown.

FYI @andrewsignori-aot

CarlyCotton:
Looks good. Processed non-legacy mapping, legacy mapping with existing, added a new legacy map.

PR #4049, opened 2024-12-07
#4013 - Legacy Restriction Mapping - Student Changes

andrewsignori-aot:
- Prevent API from retuning 'No effect' notifications since they are not supposed to be notified to the student.
- Consolidated `LGCY_*` restrictions into a single `LGCY` restriction code to be returned to the student. The error code returned will be the highest one present in the list of the `LGCY_*` being returned.
- Create few `LGCY_*` in the DB seed to allow the E2E to use them.
  - Fixed the `db:seed:test:clean` that was not removing the `create_history_entry` function leading to an error when the DB was populated again.
  - Close Nestjs app (`app.close()`) to avoid the `db:seed:test:clean` and `db:seed:test` to be hanging and waiting.
 
### Ministry visualization for the new legacy map

image

### SQL to be manually added during release

```sql
INSERT INTO
    sims.sfas_restriction_maps (legacy_code, code, is_legacy_only)
VALUES
    ('TD', 'LGCY_TD', TRUE),
    ('B5', 'LGCY_B5', TRUE),
    ('B5A', 'LGCY_B5A', TRUE),
    ('B7', 'LGCY_B7', TRUE),
    ('SSD', 'LGCY_SSD', TRUE),
    ('Z2', 'LGCY_Z2', TRUE),
    ('M1', 'LGCY_M1', TRUE),
    ('SSRN', 'LGCY_SSRN', TRUE),
    ('B4', 'LGCY_B4', TRUE),
    ('Z1', 'LGCY_Z1', TRUE)
```

### DIFF
--- a/sims.code-workspace
+++ b/sims.code-workspace
@@ -113,6 +113,10 @@
       "golevelup",
       "Ilike",
       "lastupdated",
+      "LGCY",
+      "LGCYAAAA",
+      "LGCYBBBB",
+      "LGCYCCCC",
       "MBAL",
       "MSFAA",
       "NOAAPI",
@@ -134,9 +138,7 @@
       "timestamptz",
       "typeorm",
       "unparse",
-      "Zeebe",
-      "LGCY",
-      "SFAS"
+      "Zeebe"
     ],
     "[json]": {
       "editor.defaultFormatter": "vscode.json-language-features"
--- a/sources/packages/backend/apps/api/src/route-controllers/restriction/restriction.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/restriction/restriction.students.controller.ts
@@ -31,12 +31,12 @@ export class RestrictionStudentsController extends BaseController {
   async getStudentRestrictions(
     @UserToken() studentToken: StudentUserToken,
   ): Promise<StudentRestrictionAPIOutDTO[]> {
-    // TODO: check if filterNoEffectRestrictions should be passed.
     const studentRestrictions =
       await this.studentRestrictionService.getStudentRestrictionsById(
         studentToken.studentId,
         {
           onlyActive: true,
+          filterNoEffectRestrictions: true,
         },
       );
     // Separate the results between non-legacy and legacy restrictions.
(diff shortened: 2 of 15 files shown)


sims.code-workspace
```diff
@@ -113,6 +113,10 @@
       "golevelup",
       "Ilike",
       "lastupdated",
+      "LGCY",
+      "LGCYAAAA",
+      "LGCYBBBB",
+      "LGCYCCCC",
```
dheepak-aot, line 119: 😂

sources/packages/backend/apps/api/src/route-controllers/restriction/restriction.students.controller.ts
```diff
@@ -34,12 +36,38 @@ export class RestrictionStudentsController extends BaseController {
...
+        : nonLegacyRestrictions;
+      resultList.push(studentRestriction);
+    });
+    // Create the output result for non-legacy restrictions.
+    const results = nonLegacyRestrictions.map((studentRestriction) => ({
       code: studentRestriction.restriction.restrictionCode,
       type: studentRestriction.restriction.notificationType,
     }));
+    if (legacyRestrictions.length) {
+      const notificationOrder = Object.values(RestrictionNotificationType);
```
dheepak-aot, line 57: Just a suggestion. Will it be better to use a map instead of the order how enums are declared?

```ts
 const notificationOrderMap = {
        [RestrictionNotificationType.Warning]: 1,
        [RestrictionNotificationType.Error]: 2,
        [RestrictionNotificationType.NoEffect]: 3,
      };
      // If any legacy restriction is present, create a generic LGCY restriction
      // with the highest notification type.
      studentRestrictions.sort(
        (a, b) =>
          notificationOrderMap[b.restriction.notificationType] -
          notificationOrderMap[a.restriction.notificationType],
      );
```

andrewsignori-aot: I can see benefits in both ways.
Now that we do not have `NoEffect` anymore, maybe I can add a priority only to `Error` and `Warning` and the suggested code would be better and clear.

andrewsignori-aot: Code update, please take a look.

dheepak-aot (approved):
Great Job. Thanks for catch to filter `no effect` restrictions. Looks good 👍

guru-aot (approved):

Issue #4052, opened 2024-12-09
2.1 HOTFIX - Program Year Limits from Bridge File


CarlyCotton:
**User Story**
As a student, any applications from the current program year should be considered in the assessment towards my program year total. 

**Current Behaviour**
This was not happening in applications on their first app despite having a positive bridge match and application records from the current PY.

**Expected Behaviour**
Any current year application records should search for bridge file information on assessment or reassessment.

_Added 2pts after consulting with Dheepak after investigation_

CarlyCotton:
Looks good. Checked on first app and it pulled in correctly. Checked both PT/FT tables and first/second assessments of an app.

PR #4058, opened 2024-12-10
#4052 - Fix the legacy program year totals for program year limits

dheepak-aot:
## Fix the legacy program year totals for program year limits

- [x] When calculating the program year totals in the worker, if the student never had any assessment calculated before, then use `current date` as reference assessment date.
- [x] Add/Update E2E test to validate the exact scenario.


andrewsignori-aot (approved):
Great analysis and investigation, look good 👍

PR #4071, opened 2024-12-11
#3258 - Bug in the Error creation of CAS Supplier

guru-aot:
Trim function was called after the split, it should have been called before it.
image

### DIFF
--- a/sources/packages/backend/libs/integrations/src/cas/cas.service.ts
+++ b/sources/packages/backend/libs/integrations/src/cas/cas.service.ts
@@ -257,8 +257,9 @@ export class CASService {
       error.response?.status === HttpStatus.BAD_REQUEST &&
       !!error.response?.data[CAS_RETURNED_MESSAGES]
     ) {
-      const casKnownErrors =
-        error.response.data[CAS_RETURNED_MESSAGES].toString();
+      const casKnownErrors = error.response.data[
+        CAS_RETURNED_MESSAGES
+      ] as string;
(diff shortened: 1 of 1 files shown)


andrepestana-aot (approved):
Looks good. Weird how it was not shown as an error before.

dheepak-aot (approved):
Looks good.

sources/packages/backend/libs/integrations/src/cas/cas.service.ts
```diff
@@ -257,12 +257,13 @@ export class CASService {
...
-      const casKnownErrors =
-        error.response.data[CAS_RETURNED_MESSAGES].toString();
+      const casKnownErrors = error.response.data[
+        CAS_RETURNED_MESSAGES
+      ] as string;
       throw new CustomNamedError(
         "CAS Bad Request Errors",
         CAS_BAD_REQUEST,
-        casKnownErrors.split(" | ").trim(),
+        casKnownErrors.trim().split(" | "),
```
andrewsignori-aot, line 266: To be clear, that was not what I meant when I said to `trim()`.

andrewsignori-aot: The string "Error 1 | Error 2" should be splitter by "|".
Then the results will be "Erro 1 " and " Error 2", those should have the `trim()` applied.
I also mentioned to have one unit test about it before the ticket was delivered.

Issue #4019, opened 2024-12-03
Virus Scan False Positives 


ninosamson:
Since PT launch, a higher than expected number of uploaded files are failing to upload due to potential virus detection. 

Common theme seems to be PDFs. The SABC Appendix 8 form below is a PDF that will get uploaded regularly by students. Need to understand what is being detected and come up with resolution, as it may block other uploads as well.

Possible current error.
"Tue Dec 3 19:04:35 2024 -> instream(10.97.122.202@39440): Heuristics.Limits.Exceeded.MaxFiles FOUND"

https://studentaidbc.ca/sites/all/files/form-library/appendix_8.pdf

**Acceptance Criteria**
- [ ] Investigate pdfs failing virus scanning and fix
- [ ] Try to update clamav to the most updated version (nice to have)

<img width="635" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFFBQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9c510dd083d9bbe2ca578e1e50a1c1c5b2526b03/image.png" alt="image.png" />
<img width="998" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFlBQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a9b5e6a7042e8f70d6cf9e40f15af24c35a9da68/image.png" alt="image.png" />

PR #4129, opened 2024-12-17
#4019 - Virus Scan False Positives

bidyashish:
**Acceptance Criteria**
- [X] Investigate pdfs failing virus scanning and fix
- [X] Try to update clamav to the most updated version (nice to have)

**Notes**
MaxFiles 100 was causing issue with file being not scanned and using ClamAV virus Database bank to not scan file.

`
MaxFiles in ClamAV's configuration refers to the maximum number of files to be scanned within an archive, document, or any other container file. Here's a detailed explanation:
For example:
If scanning a ZIP file containing 15,000 files with MaxFiles 10000:
Only the first 10,000 files will be scanned
The remaining 5,000 files will be skipped
If AlertExceedsMax is enabled, it will trigger a "Heuristics.Limits.Exceeded.MaxFiles" alert
`

Update Clam AV Docker from BCGOV Repo
Link: https://github.com/bcgov/common-hosted-clamav-service/pkgs/container/clamav-unprivileged

Demo: Manual test in Dev using Config update.
image

### DIFF
--- a/devops/helm/clam-av/_clamav/configurations/1.0/clamd.conf
+++ b/devops/helm/clam-av/_clamav/configurations/1.0/clamd.conf
@@ -562,7 +562,7 @@ MaxRecursion 10
 # Note: disabling this limit or setting it too high may result in severe damage
 # to the system.
 # Default: 10000
-MaxFiles 10000
+# MaxFiles 10000
 
 # Maximum size of a file to check for embedded PE. Files larger than this value
 # will skip the additional analysis step.
--- a/sources/packages/clam-av/clamd.conf
+++ b/sources/packages/clam-av/clamd.conf
@@ -562,7 +562,7 @@ MaxRecursion 10
 # Note: disabling this limit or setting it too high may result in severe damage
 # to the system.
 # Default: 10000
-MaxFiles 10000
+# MaxFiles 10000
 
 # Maximum size of a file to check for embedded PE. Files larger than this value
 # will skip the additional analysis step.


devops/helm/clam-av/_clamav/configurations/1.0/clamd.conf
```diff
@@ -562,7 +562,7 @@ MaxRecursion 10
 # Note: disabling this limit or setting it too high may result in severe damage
 # to the system.
 # Default: 10000
-MaxFiles 100
+MaxFiles 10000
```
dheepak-aot, line 565: As it is set to default value please comment this line leaving the default value to be in effect.

sources/packages/clam-av/clamd.conf
```diff
@@ -562,7 +562,7 @@ MaxRecursion 10
 # Note: disabling this limit or setting it too high may result in severe damage
 # to the system.
 # Default: 10000
-MaxFiles 100
+MaxFiles 10000
```
dheepak-aot, line 565: same here https://github.com/bcgov/SIMS/pull/4129/files#r1889335483

dheepak-aot (approved):
Thanks for doing the changes @bidyashish. Looks good 👍

andrepestana-aot (approved):
Looks good. Thanks for the explanation. I just think that some explanation about why that happened and how it works would be good to be added in the description of this PR as someone can as ask one of the DEVs in the future why we change that.

Issue #4128, opened 2024-12-16
Persist uploaded documents and exception triggering for dependants with disability 


Joshua-Lakusta:
As a student and ministry user we need to be able to see the documents uploaded for dependants with disabilities declared on taxes when there is also a dependant that is not declared on taxes. We also require the exception to be triggered in this situation. 

**Acceptance Criteria**

- [ ] Ensure that when "Clear value when hidden" is not checked in form IO for this specific question so that the file information is not removed for all program years
- [ ] It is ok that this information will reappear when reactivated with future edits
- [ ] Ensure that when this change is made that the dependant declared on taxes exception triggers in scenarios where there is a disability dependant 

**Additional Context**
This was captured during another ticket by Lewis. Goal of this ticket is to ensure behaviour is the same for all program years.

<img width="690" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNDRyQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f9fe73a9b9034cf058f289fb7718a6954502f5b5/image.png" alt="image.png" />

ninosamson:
removed estimate and linked to #4132, which should resolve this ticket.

PR #4140, opened 2024-12-18
#4128 - Persist uploaded documents and exception triggering for dependants with disability 

lewischen-aot:
- Removed key "Clear value when hidden" for file component "pdDependentUpload" for program years 
  - 2022-23
  - 2023-24
  - 2024-25
- Ensured that the file will reappear when reactivated with future edits.
- Ensured that the declared on taxes exception is triggered when file is required but not provided for a dependant.

Screenshot for when the declared on taxes exception is triggered
image

Screenshot for files successfully submitted and reappeared with future edits
image

Screenshot of the dependant component with files on ministry portal
image

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2024-25.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2024-25.json
@@ -1742,8 +1742,7 @@
                     "threshold": 0.3
                   },
                   "id": "eopfm7h",
-                  "defaultValue": "",
-                  "isNew": false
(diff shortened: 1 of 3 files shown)


sources/packages/forms/src/form-definitions/sfaa2024-25.json
andrepestana-aot, line 1: What about sfaa2025-2026.json ? I didn't see anywhere mentioning that we don't need to change that one.

Issue #4094, opened 2024-12-13
File Issues - Comma in file name


CarlyCotton:
**Description**
When a student uploads a file with a comma in the name, it causes issues when trying to download it. The original file name remains the same but we remove commas when creating the unique file name. 

**Current Behaviour**
When a file has been uploaded with a comma in it, that file is unable to be downloaded by any users within SIMS.

**Expected Behaviour**
We can download files that have successfully had a virus scan.

**Steps to reproduce**
1. Upload a file with a comma in the name.
2. Have the file pass virus scan.
3. Try to download the file.

**Technical Details**
- [ ] Ensure the file name is escaped while downloading to prevent the issue with the comma and some other potential characters.
- [ ] Add an E2E test with some scenarios to test different characters.
- [ ] This should be added to v2.2 branch.

PR #4153, opened 2024-12-19
#4094 - File Issues - Comma in file name

bidyashish:
**Technical Details**
- [x] Ensure the file name is escaped while downloading to prevent the issue with the comma and some other potential characters.
- [x] Add an E2E test with some scenarios to test different characters.
- [ ] This should be added to v2.2 branch.

Demo:
Able to Download File with Comma
image
image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
@@ -16,10 +16,7 @@ import {
   ApplicationPaginationOptionsAPIInDTO,
   PaginatedResultsAPIOutDTO,
 } from "../models/pagination.dto";
-import {
-  getUserFullName,
-  encodeFileNameForContentDisposition,
-} from "../../utilities";
+import { getUserFullName } from "../../utilities";
 import {
   CustomNamedError,
   getISODateOnlyString,
@@ -159,7 +156,7 @@ export class StudentControllerService {
 
     response.setHeader(
       "Content-Disposition",
-      encodeFileNameForContentDisposition(studentFile.fileName),
+      `attachment; filename=${encodeURIComponent(studentFile.fileName)}`,
     );
 
     try {
--- a/sources/packages/backend/apps/api/src/utilities/string-utils.ts
+++ b/sources/packages/backend/apps/api/src/utilities/string-utils.ts
@@ -15,13 +15,3 @@ export const removeWhiteSpaces = (stringValue: string): string => {
 export const removeDoubleWhiteSpaces = (stringValue: string): string => {
   return stringValue.replace(/\s+/g, " ");
 };
-
-/**
- * Encodes a filename for Content-Disposition header
- * @param fileName Original filename to encode
- * @returns Object containing encoded filename
- */
-export function encodeFileNameForContentDisposition(fileName: string): string {
-  const encodedFileName = encodeURIComponent(fileName);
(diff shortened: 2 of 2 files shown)


sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -154,9 +154,11 @@ export class StudentControllerService {
       );
     }
 
+    const encodedFileName = encodeURIComponent(studentFile.fileName);
```
andrepestana-aot, line 157: I think we need to investigate it better. I saved a file with double quotes, comma and question mark. 
image

I couldn't event download that back.
image

The name saved seems to be the issue:
image

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
```diff
@@ -154,9 +154,11 @@ export class StudentControllerService {
...
+
     response.setHeader(
       "Content-Disposition",
-      `attachment; filename=${studentFile.fileName}`,
+      `attachment; filename*=UTF-8''${encodedFileName}`,
```
dheepak-aot, line 161: Great finding. 
I see that it follows the `Extended notation` example from [RFC5987](https://www.rfc-editor.org/rfc/rfc5987#section-3.2.2)

I believe the `encodeURIComponent()` sufficient not required to do further replace as stated here. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#examples 
image


@andrewsignori-aot @andrepestana-aot @guru-aot @lewischen-aot

dheepak-aot, line 161: Please move this logic of creating encoded file name to a common util.

guru-aot: Agreed @dheepak-aot

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
andrepestana-aot: Good! That's interesting it didn't work for me in my local env... Thanks for checking that!

sources/packages/backend/apps/api/src/route-controllers/student/student.controller.service.ts
andrepestana-aot: Looks good.

sources/packages/backend/apps/api/src/utilities/string-utils.ts
```diff
@@ -15,3 +15,13 @@ export const removeWhiteSpaces = (stringValue: string): string => {
...
   return stringValue.replace(/\s+/g, " ");
 };
+
+/**
+ * Encodes a filename for Content-Disposition header
+ * @param fileName Original filename to encode
+ * @returns Object containing encoded filename
+ */
+export function encodeFileNameForContentDisposition(fileName: string): string {
+  const encodedFileName = encodeURIComponent(fileName);
```
dheepak-aot, line 25: As we are just using `encodeURIComponent` please remove the util as the purpose seems to be more linear.

dheepak-aot (approved):
Thanks for making the changes @bidyashish. Looks good 👍

andrepestana-aot (approved):
Looks good!

Issue #3869, opened 2024-10-31
Unzip Federal Restrictions File 


ninosamson:
**User Story**
The federal restrictions file is passed to SIMS as a zip file. We need to unzip the file and then import the .txt file. Right now we are not expecting to unzip the file.

**Acceptance Criteria**
- [x] unzip the federal restrictions file when it comes to SIMS from the SFTP so we can process the .TXT file 
- [x] archive the zip file during this process

**Technical**
- [x] Only one file is expected per zip file.
- [ ] Use the out-of-box node lib.
- [x] Try to create the ability to decompress any zip file using the common method that reads the files.
- [ ] Nice to have, create an E2E to have a mocked zip file.

CarlyCotton:
Just a note for me to grab a sample and update ACs to be more fulsome

dheepak-aot:
While running the federal restrictions file with bulk volume of data (139MB File), we noticed that the required resource limit is going beyond that maximum limit for both CPU and Memory.

<img width="796" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2hDQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--244e66681706f8abbef3806607305c38190a7a67/image.png" alt="image.png" />

<img width="813" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2xDQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e6330bd1dc05dcb65518744ed056b33afaf558d8/image.png" alt="image.png" />

Hence to meet the demand for job(s) requiring more resources(Currently federal restrictions is the only one such job), CPU and Memory limits and requested have been increased a little.

<img width="740" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3BDQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b09146a233f8f68a3386ca092e675ead8ff8e4f8/image.png" alt="image.png" />

PR #4163, opened 2024-12-20
#3869 - Federal Restrictions - Unzip ZIP files coming from SFTP

dheepak-aot:
## Federal Restrictions - Unzip ZIP files coming from SFTP

- [x] Adjusted the existing regex to match the federal restrictions file with `.zip` and `.ZIP` format.
- [x] Used a library [adm-zip](https://github.com/cthackers/adm-zip) to extract the `.zip` file.
- [x] Updated the `sftp-integration-base` to use encoding `null` while reading only the compressed file to avoid data corruption.
- [x] For the federal restrictions integration, the first file inside the downloaded compressed archive is processed assuming there will always be only one file in the `.zip` archive.

### APPROACH 1

Based on documentation and also testing, the nodejs in built library Zlib supports archiving and extraction of only gunzip (.gz) files. 
It does not support the same operations on a .zip files.
Extracting .zip with Zlib Gunzip (Doesn't support)

image

image

Extracting .gz with Zlib Gunzip (Works Perfectly)

image

### APPROACH 2 - Third party library

https://github.com/cthackers/adm-zip 

Tested code(Not the final code)

image

It also provides non blocking method to read data. (getDataAsync)
 
It works perfectly.

image

Tested the upload with 139MB file with around 140,000 records. 
Time taken by the lib to read the file is 666ms
image


sources/packages/backend/libs/integrations/src/esdc-integration/fed-restriction-integration/fed-restriction.processing.service.ts
```diff
@@ -52,9 +52,9 @@ export class FedRestrictionProcessingService {
    */
   async process(processSummary: ProcessSummary): Promise<void> {
     const auditUser = this.systemUsersService.systemUser;
-    // Get the list of all files from SFTP ordered by file name.
+    // Get the list of all ZIP files from SFTP ordered by file name.
     const fileSearch = new RegExp(
-      `^${this.esdcConfig.environmentCode}CSLS\\.PBC\\.RESTR\\.LIST\\.D[\\w]*\\.[\\d]*$`,
+      `^${this.esdcConfig.environmentCode}CSLS\\.PBC\\.RESTR\\.LIST\\.D[\\w]*\\.[\\d]*\\.(zip|ZIP)$`,
```
andrewsignori-aot, line 57: Not enforcing the zip extension would allow the process to download the file either way, compressed or not.
I do not see an AC requesting either one or not and I do not see a reason to have it restricted.
Not a blocker.

sources/packages/backend/libs/integrations/src/services/ssh/sftp-integration-base.ts
```diff
@@ -154,10 +155,29 @@ export abstract class SFTPIntegrationBase<DownloadType> {
           return false;
         }
       }
-      // Read all the file content and create a buffer with 'ascii' encoding.
-      const fileContent = await client.get(remoteFilePath, undefined, {
-        readStreamOptions: { encoding: FILE_DEFAULT_ENCODING },
-      });
+      let fileContent: string | NodeJS.WritableStream | Buffer;
+      const fileExtension = path.parse(remoteFilePath).ext.toLocaleLowerCase();
```
andrewsignori-aot, line 159: Any particular reason to use `toLocaleLowerCase` instead of the regular `toLowerCase`?

dheepak-aot: That was a mistake 😂 intention was `toLowerCase`

andrewsignori-aot: As an alternative, the below was used in the past.
```ts
path.extname(file.originalname).toLowerCase()
```

sources/packages/backend/libs/integrations/src/services/ssh/sftp-integration-base.ts
```diff
@@ -154,10 +155,29 @@ export abstract class SFTPIntegrationBase<DownloadType> {
...
+          compressedFileContent,
+          { encoding: FILE_DEFAULT_ENCODING },
+        );
+        this.logger.log(`Extracted the first file ${fileName}.`);
+        fileContent = data;
+      } else {
+        // Read all the file content and create a buffer with 'ascii' encoding.
+        fileContent = await client.get(remoteFilePath, undefined, {
+          readStreamOptions: { encoding: FILE_DEFAULT_ENCODING },
+        });
```
andrewsignori-aot, line 179: Similar to the compressed get at the line 167 the below can be cast.
```ts
 // Read all the file content and create a buffer with 'ascii' encoding.
fileContent = (await client.get(remoteFilePath, undefined, {
  readStreamOptions: { encoding: FILE_DEFAULT_ENCODING },
})) as string;
```
This would allow the fileContent to be declared as string (which really is) instead of `let fileContent: string | NodeJS.WritableStream | Buffer;` which can be misleading.

sources/packages/backend/libs/utilities/src/compressed-file-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  fileName: string;
+  data: string;
+} {
+  const zipFile = new AdmZip(compressedFileBuffer);
+  const [firstExtractedFile] = zipFile.getEntries();
+  if (!firstExtractedFile) {
+    throw new Error("No files found in zip file");
+  }
+  // Read the first extracted file with the specified encoding.
+  const data = zipFile.readAsText(firstExtractedFile, options?.encoding);
```
andrewsignori-aot, line 23: Since there is an `async` method available (`zipFile.readAsTextAsync`), was there any reason to not use it?

sources/packages/backend/libs/utilities/src/compressed-file-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  compressedFileBuffer: Buffer,
+  options?: { encoding: string },
+): {
+  fileName: string;
+  data: string;
+} {
+  const zipFile = new AdmZip(compressedFileBuffer);
+  const [firstExtractedFile] = zipFile.getEntries();
+  if (!firstExtractedFile) {
+    throw new Error("No files found in zip file");
```
andrewsignori-aot, line 20: Please add the period.

sources/packages/backend/libs/utilities/src/compressed-file-utils.ts
```diff
@@ -0,0 +1,28 @@
+import * as AdmZip from "adm-zip";
+
+/**
+ * Reads the first extracted file from a compressed archive file.
+ * @param compressedFileBuffer
```
andrewsignori-aot, line 5: Missing parameter comment.

andrewsignori-aot (commented):

sources/packages/backend/libs/integrations/src/services/ssh/sftp-integration-base.ts
```diff
@@ -154,10 +155,29 @@ export abstract class SFTPIntegrationBase<DownloadType> {
...
-      });
+      let fileContent: string | NodeJS.WritableStream | Buffer;
+      const fileExtension = path.parse(remoteFilePath).ext.toLocaleLowerCase();
+      const isZIPFile = fileExtension === ".zip";
+      if (isZIPFile) {
+        // Read the zip file content with null encoding to avoid data corruption.
+        const compressedFileContent = (await client.get(
+          remoteFilePath,
+          undefined,
+          { readStreamOptions: { encoding: null } },
```
guru-aot, line 166: 👍

sources/packages/backend/libs/integrations/src/esdc-integration/fed-restriction-integration/fed-restriction.processing.service.ts
dheepak-aot: Not enforcing the `.zip` will also lead to not enforcing the end of file name by removing the existing `$`. I see the benefit of it working either way, but should e remove the `$`. ?

Let me know if I am missing something here.

sources/packages/backend/libs/utilities/src/compressed-file-utils.ts
dheepak-aot: To use `readAsTextAsync` in a way to wait that callback is completed, we must promisify it. As we do not have an opportunity to run the reading in parallel with any other operation, I did not use the async option.

sources/packages/backend/libs/integrations/src/esdc-integration/fed-restriction-integration/fed-restriction.processing.service.ts
andrewsignori-aot: As we talked, IMO, we do need to enforce the precise ending of the string but I am ok either way.

andrewsignori-aot (approved):
Great research, great work, looks good 👍

guru-aot (approved):
LGTM, nice work @dheepak-aot

PR #4184, opened 2024-12-27
#3869 - Increase memory and CPU limits of queue consumers to handle jobs with more resource demand

dheepak-aot:
# Issue

While running the federal restrictions file with bulk volume of data (139MB File), we noticed that the required resource limit is going beyond that maximum limit for both CPU and Memory. 

image

image


andrewsignori-aot (approved):
👍

lewischen-aot (approved):
Nice work 👍 Thanks for increasing the server performance.

Issue #4132, opened 2024-12-17
Update Form  IO 


Joshua-Lakusta:
**Describe the task**
Form IO requires an update to be completed

**Acceptance Criteria**
- [x] Update form IO
- [x] Go back and revert the fix for dependant docs and exception in tickets #4128 and #3827 if the upgrade has resolved the issues (Check to see if it has)
 Link to wiki https://github.com/bcgov/SIMS/wiki/Development-%E2%80%90-Form.io-Server-Upgrade
- [x] Application submission button (top) should only be displayed on the final page.

**Additional context**q
New update for Form IO needed 2024-12-17

CarlyCotton:
@lewischen-aot advised me about additional fix on tooltips. Looks fine and closing ticket.

CarlyCotton:
Brought back for some small impacts from upgrade:
<img width="1196" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeWxKQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e62928125e0139f43c6290bf0c84fc1fb93b54f1/image.png" alt="image.png" />
<img width="1874" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeXBKQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--444e14724d5146bf09a1321c02412ddf65816fef/image.png" alt="image.png" />

PR #4187, opened 2024-12-27
#4132 - Update Form IO (Tooltip Fix)

lewischen-aot:
- Added CSS for tooltips in formio shared css file.

Screenshot of the tooltip before and after fix
image


andrewsignori-aot (approved):
Nice find. I do not see a problem proceeding with these fixes right now because having the upgrade done is way more important than some missing icons.
Do you mind creating a tech debt to further investigate why the icons styles are being overwritten? There seems an issue between `bi` vs `fa` icons. It may be an issue with the new form.io version or something that we should look at and have configured.
Looks good 👍

andrepestana-aot (approved):
Looks good!

Issue #3596, opened 2024-07-31
Bug - unexpected error for SIN change and manual validation screen when you try to submit 


Joshua-Lakusta:
**Describe the Bug**
Some SIN numbers are triggering an unexpected error when we attempt to manually add them in the ministry portal.

**Expected Behaviour**
Be able to add the SIN with **no errors**

**Actual Behaviour**
What actually happened is that the system kicked back a unexpected error message in the top right

** Steps To Reproduce**
Steps to reproduce the behaviour:
1. Go to ministry portal
2. Search and access a student account
3. manually try to change the SIN with out overriding the sin validation. SIN is previously over ridden 
4. Use any SIN
5. add note
6. submit

<img width="798" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTBpQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8abff7f0b12d9dd8d25fcc91792648c7321b6745/image.png" alt="image.png" />

<img width="681" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTRpQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6e21620e2170236a5c19003c82e990ac2c1468d0/image.png" alt="image.png" />

New photo as of 2024-10-22
<img width="1546" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMXhtQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b00ff43ceffb6a390ea625076d9a24710626a5f0/image.png" alt="image.png" />

CarlyCotton:
Reviewed on 2024-12-31

PR #4191, opened 2024-12-31
#3596 - Bug on ministry add new SIN to student

dheepak-aot:
## Issue

The optional checkbox to `Skip the validations` returns `undefined` (When not touched) and `null` (on form reset) which is considered as bad request as the API is expecting a boolean value for `skipValidations`.

## Solution

- [x] Initialize the form-model with `skipValidations` as `false`.
- [x] On the event of model value being updated, convert the input value to boolean using the event `update:model-value`.

## Minor Refactor

- [x] Removed the call to `resetValidation()` as it has no impact after `reset()` is called.

image


guru-aot (approved):
LGTM, nice work @dheepak-aot

PR #4194, opened 2025-01-02
#4132 - Update Form IO (Fix Eligible Amount)

lewischen-aot:
- Set `"refreshOnChange": true` to display eligible amount.

image

- Removed "Submit application" button for pages before the last page.

image

### DIFF
--- a/sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
+++ b/sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
@@ -16,7 +16,7 @@
             {{ savingDraft ? "Saving..." : "Save draft" }}</v-btn
           >
           <v-btn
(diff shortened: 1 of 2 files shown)


sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
andrewsignori-aot, line 1: This is not an actual fix and should be considered a change in the actual behavior to prevent an error where the form may be submitted without triggering all the required validations.
@CarlyCotton add the below AC to track this and QAs will be advised about the change. If required a ticket will be created to investigate the validations issue.
```
The application submission button (top) should only be displayed on the final page.
```

andrewsignori-aot (approved):
Thanks for quickly taking care of these two questions. Looks good 👍

PR #4209, opened 2025-01-06
Remove Patroni and backup files and makefile commands from SIMS repository

guru-aot:
Remove Patroni deployment and backup files, makefile commands from SIMS repository.


guru-aot:
> Looks good, Nice work @guru-aot
> 
> Btw, I find few old place still referencing `Patroni`, not sure if those are to be deleted too?
> 
> image

Those are for crunchy installations and no need to be removed @bidyashish

Issue #3910, opened 2024-11-07
Show application edit history ministry view


Joshua-Lakusta:
As a ministry user I want to be able to see the edit history for a an application. This is important in order to be able to see historical changes and how they impact decision making and for setting up an understanding of logic connected to future tickets

**Acceptance Criteria**
- [x] Modify application view screen to have all versions of the app history on the left hand side top current view (#1) will not change and will be the current version
- [x] Each version of the app will be a expandable selection which starts compressed and then can be expanded to show application
- [x] We will only display the Application section for historical entries but will leave the current app ui alone.
- [x] Each app version will appear in descending order based on date then time stamp (#2).  App versions should be in descending order from most recent chronologically and descend down to original submission 
- [x] Label of each app will be Date/Time (same time stamp format as notes)
- [x] history is only for non draft applications with submitted date

<img width="1907" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTRPQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1204ec4a1d2e2ac2fd92b711a11964a5aebc9805/image.png" alt="image.png" />

**Additional Context**
This is important to be able to validate edit history for exceptions, future change requests (more tickets to come), and for verification processes. 

**Technical**
- [x] Create the menu on demand, only if some historical application exists.
- [x] Supporting users is not considered part of this effort.
- [x] Change the API to allow the overwritten application information needed to build the menu.
- [x] Ensure API would allow the overwritten applications to be retrieved.
- [x] E2E tests for the changed API endpoints.

Sample menu would be like below structure.
```txt
Application (existing)
Assessments (existing)
Restrictions Management (existing)
Application Status (existing)
- Submitted on Dec 10 2024 16:05 (new to be created on demand)
   Application (only option for now)
- Submitted on Dec 10 2024 16:05 (new to be created on demand)
   Application (only option for now)
```
_Note:_ Biz to confirm the label including the `Submitted on`.

andrewsignori-aot:
@Joshua-Lakusta DEVs just would like to ensure the supporting users are part of some upcoming tickets since it was not mentioned in this one. The idea for this ticket is to keep it simple and have the application displayed, which means that we are considering supporting users outside the scope.

PR #4213, opened 2025-01-08
#3910 - Show application edit history ministry view

lewischen-aot:
- Created the menu on demand, only if any historical application exists.
   - Modify application view screen to have all versions of the app history on the left hand side.
   - Each version of the app will be a expandable selection.
   - Each app version will appear in descending order based on date time stamp.
- Added a new API endpoint for fetching historical applications to populate the menu.
- Changed the API to allow the overwritten application information needed to build the menu for Ministry User.
   - Ensured API would allow the overwritten applications to be retrieved.
- Modified and added E2E tests for the changed API endpoints.

Screenshot of the historical application items
image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
@@ -1,0 +1,97 @@
+import { HttpStatus, INestApplication } from "@nestjs/common";
+import * as request from "supertest";
+import {
+  AESTGroups,
+  BEARER_AUTH_TYPE,
+  createTestingAppModule,
+  getAESTToken,
+} from "../../../testHelpers";
+import {
+  createE2EDataSources,
+  E2EDataSources,
+  saveFakeApplication,
+} from "@sims/test-utils";
+import { ApplicationStatus } from "@sims/sims-db";
+import { addDays, getPSTPDTDateHourMinute } from "@sims/utilities";
+
+describe("ApplicationAESTController(e2e)-getApplicationVersionHistory", () => {
+  let app: INestApplication;
+  let db: E2EDataSources;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    db = createE2EDataSources(dataSource);
+  });
+
+  it("Should get an array of application versions when there is a previous application associated with the given application.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
+      submittedDate: new Date(),
+    });
+
+    const previousApplication = await saveFakeApplication(
+      db.dataSource,
+      {
+        institution:
+          application.currentAssessment.offering.institutionLocation
+            .institution,
+        institutionLocation:
+          application.currentAssessment.offering.institutionLocation,
+        student: application.student,
+        program: application.currentAssessment.offering.educationProgram,
+        offering: application.currentAssessment.offering,
+        programYear: application.programYear,
+      },
+      {
+        applicationStatus: ApplicationStatus.Overwritten,
+        applicationNumber: application.applicationNumber,
+        offeringIntensity:
+          application.currentAssessment.offering.offeringIntensity,
+        pirStatus: application.pirStatus,
+        submittedDate: addDays(-10),
+      },
+    );
+
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    const endpoint = `/aest/application/${application.id}/versions`;
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .expect({
+        applicationVersions: [
+          {
+            applicationId: previousApplication.id,
+            submittedDate: getPSTPDTDateHourMinute(
+              previousApplication.submittedDate,
+            ),
+          },
+        ],
+      });
+  });
+
+  it("Should throw not found error when an application is not found.", async () => {
+    // Arrange
+    const endpoint = "/aest/application/99999999/progress-details";
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.NOT_FOUND)
+      .expect({
+        statusCode: HttpStatus.NOT_FOUND,
+        message: "Application id 99999999 was not found.",
+        error: "Not Found",
+      });
+  });
+
+  afterAll(async () => {
+    await app?.close();
+  });
+});
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
@@ -19,6 +19,7 @@ import {
   CompletedApplicationDetailsAPIOutDTO,
   EnrolmentApplicationDetailsAPIOutDTO,
   InProgressApplicationDetailsAPIOutDTO,
+  ApplicationVersionAPIOutDTO,
 } from "./models/application.dto";
 import {
   AllowAuthorizedParty,
@@ -37,7 +38,7 @@ import { ClientTypeBaseRoute } from "../../types";
 import { ApplicationControllerService } from "./application.controller.service";
 import { PrimaryIdentifierAPIOutDTO } from "../models/primary.identifier.dto";
 import { MSFAANumberSharedService } from "@sims/services";
-import { CustomNamedError } from "@sims/utilities";
+import { CustomNamedError, getPSTPDTDateHourMinute } from "@sims/utilities";
 import {
   APPLICATION_INVALID_DATA_TO_CREATE_MSFAA_ERROR,
   APPLICATION_NOT_FOUND,
@@ -45,6 +46,7 @@ import {
 } from "@sims/services/constants";
 import { IUserToken, Role } from "../../auth";
 import { ApplicationStatus } from "@sims/sims-db";
+import { application } from "express";
 
 @AllowAuthorizedParty(AuthorizedParties.aest)
 @Groups(UserGroups.AESTUser)
@@ -74,7 +76,7 @@ export class ApplicationAESTController extends BaseController {
   ): Promise<ApplicationSupplementalDataAPIOutDTO> {
     const application = await this.applicationService.getApplicationById(
       applicationId,
-      { loadDynamicData },
+      { loadDynamicData, allowOverwritten: true },
     );
     if (!application) {
       throw new NotFoundException(
@@ -230,5 +232,29 @@ export class ApplicationAESTController extends BaseController {
     return this.applicationControllerService.getCompletedApplicationDetails(
       applicationId,
     );
+  }
+
+  /**
+   * Get history of application versions for an application
+   * where the current application is excluded.
+   * @param applicationId application Id.
+   * @returns history of application versions.
+   */
+  @ApiNotFoundResponse({
+    description: `Application not found or not on ${ApplicationStatus.Completed} status.`,
+  })
+  @Get(":applicationId/versions")
+  async getApplicationVersionHistory(
+    @Param("applicationId", ParseIntPipe) applicationId: number,
+  ): Promise<ApplicationVersionAPIOutDTO[]> {
+    const applications =
+      await this.applicationService.applicationVersionHistory(applicationId);
+    if (!application) {
+      throw new NotFoundException("Application not found.");
+    }
+    return applications.map((application) => ({
+      id: application.id,
+      submittedDate: getPSTPDTDateHourMinute(application.submittedDate),
+    }));
   }
 }
--- a/sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
@@ -211,3 +211,8 @@ export class ApplicationWarningsAPIOutDTO {
   eCertFailedValidations: ECertFailedValidation[];
   canAcceptAssessment: boolean;
 }
+
+export class ApplicationVersionAPIOutDTO {
+  id: number;
+  submittedDate: string;
+}
--- a/sources/packages/backend/apps/api/src/services/application/application.service.ts
+++ b/sources/packages/backend/apps/api/src/services/application/application.service.ts
@@ -591,8 +591,12 @@ export class ApplicationService extends RecordDataModelService<Application> {
       loadDynamicData?: boolean;
       studentId?: number;
       institutionId?: number;
+      allowOverwritten?: boolean;
     },
   ): Promise<Application> {
+    const applicationStatus = options?.allowOverwritten
+      ? undefined
+      : Not(ApplicationStatus.Overwritten);
     return this.repo.findOne({
       select: {
         id: true,
@@ -656,7 +660,7 @@ export class ApplicationService extends RecordDataModelService<Application> {
       },
       where: {
         id: applicationId,
-        applicationStatus: Not(ApplicationStatus.Overwritten),
+        applicationStatus: applicationStatus,
         student: {
           id: options?.studentId,
         },
@@ -1833,6 +1837,34 @@ export class ApplicationService extends RecordDataModelService<Application> {
             },
           },
         },
+      },
+    });
+  }
+
+  /**
+   * Get history of application versions for an application
+   * where the current application is excluded.
+   * @param applicationId application id.
+   * @returns assessment history list.
+   */
+  async applicationVersionHistory(
+    applicationId: number,
+  ): Promise<Application[]> {
+    const application = await this.repo.findOne({
+      select: { applicationNumber: true },
+      where: { id: applicationId },
+    });
+    if (!application) {
+      return null;
+    }
+    return this.repo.find({
+      select: { id: true, submittedDate: true },
+      where: {
+        applicationNumber: application.applicationNumber,
+        applicationStatus: ApplicationStatus.Overwritten,
+      },
+      order: {
+        submittedDate: "DESC",
       },
     });
   }
--- a/sources/packages/backend/libs/utilities/src/date-utils.ts
+++ b/sources/packages/backend/libs/utilities/src/date-utils.ts
@@ -22,6 +22,7 @@ export const DATE_ONLY_ISO_FORMAT = "YYYY-MM-DD";
 export const DATE_ONLY_FORMAT = "MMM DD YYYY";
 export const DATE_ONLY_FULL_MONTH_FORMAT = "MMMM D, YYYY";
 export const DATE_TIME_FORMAT = "YYYY-MM-DD HH:mm:ss";
+export const DATE_HOUR_MINUTE_FORMAT = "MMM DD YYYY HH:mm";
 export const TIMESTAMP_CONTINUOUS_FORMAT = "YYYY-MM-DD_HH.mm.ss";
 export const TIMESTAMP_CONTINUOUS_EXTENDED_FORMAT = "YYYYMMDD-HHmmssSSS";
 export const PST_TIMEZONE = "America/Vancouver";
@@ -97,6 +98,25 @@ export const getPSTPDTDateTime = (
   local = false,
 ): string => {
   return dayjs(new Date(date)).tz(PST_TIMEZONE, local).format(DATE_TIME_FORMAT);
+};
+
+/**
+ * Convert the date to (PST: UTC−08:00/PDT: UTC−07:00).
+ * @param date date to be converted to PST.
+ * @param local, if local is set to true,
+ * then the offset will be directly append to the date without
+ * converting to the timezone actual date,
+ * if set to false, the date will be converted
+ * to the actual timezone time with offset.
+ * @returns date in MMM DD YYYY HH:mm format.
+ */
+export const getPSTPDTDateHourMinute = (
+  date: string | Date,
+  local = false,
+): string => {
+  return dayjs(new Date(date))
+    .tz(PST_TIMEZONE, local)
+    .format(DATE_HOUR_MINUTE_FORMAT);
 };
 
 /**
--- a/sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
+++ b/sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
@@ -41,6 +41,27 @@
       :title="studentMenu.applicationStatus.title"
       @click="studentMenu.applicationStatus.command"
     />
+    <v-list v-if="applicationHistory.length" density="compact" nav>
+      <v-list-group
+        v-for="application in applicationHistory"
+        :key="application.title"
+        :prepend-icon="application.props?.prependIcon"
+        :title="application.title"
+        :value="application.title"
+      >
+        <template v-slot:activator="{ props }">
+          <v-list-item v-bind="props" :title="application.title"></v-list-item>
+        </template>
+
+        <v-list-item
+          v-for="child in application.children"
+          :key="child.title"
+          :title="child.title"
+          :prepend-icon="child.props?.prependIcon"
+          @click="child.command"
+        ></v-list-item>
+      </v-list-group>
+    </v-list>
   </v-navigation-drawer>
 </template>
 
@@ -50,6 +71,7 @@ import { ref, onMounted, defineComponent } from "vue";
 import { AESTRoutesConst } from "@/constants/routes/RouteConstants";
 import { MenuItemModel, SupportingUserType } from "@/types";
 import { SupportingUsersService } from "@/services/SupportingUserService";
+import { ApplicationService } from "@/services/ApplicationService";
 
 export interface StudentApplicationMenu {
   studentApplication: MenuItemModel;
@@ -72,6 +94,7 @@ export default defineComponent({
   setup(props) {
     const router = useRouter();
     const relatedParentPartners = ref([] as MenuItemModel[]);
+    const applicationHistory = ref([] as MenuItemModel[]);
     const studentMenu = ref<StudentApplicationMenu>({
       studentApplication: {
         title: "Application",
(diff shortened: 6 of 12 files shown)


sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -231,4 +233,28 @@ export class ApplicationAESTController extends BaseController {
...
+  /**
+   * Get history of application versions for an application
+   * where the current application is excluded.
+   * @param applicationId application Id.
+   * @returns history of application versions.
+   */
+  @ApiNotFoundResponse({
+    description: `Application not found or not on ${ApplicationStatus.Completed} status.`,
+  })
+  @Get(":applicationId/versions")
```
dheepak-aot, line 246: As per technical AC, the direction is to have a more generic API to return all details of an application including the existing supporting users which is currently returned from a specific API. 

That means.

**Current API used in sidebar is:**

`aest/supporting-user/application/${applicationId}`

Returns [{supportingUserId, supportingUserType}]

**Expected Outcome**

Use this API to return both versions and supporting user details.

We can call the endpoint with more generic name instead of `/versions`

lewischen-aot: I had the same understanding when I first started the story, but I later realized that this technical AC is to modify the API [`getApplication`](https://github.com/bcgov/SIMS/blob/4fbbe08c83b5ac1e22b9cc62a841d45926025ce6/sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts#L70) to allow ministry users to fetch overwritten applications, specifically from the application service method [`this.applicationService.getApplicationById`](https://github.com/bcgov/SIMS/blob/d780ea3431b2f0310ac9f54ed863d0788424f82b/sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts#L77). On top of that, the second AC shows that supporting users is not considered as part of this effort. If we were to return the application versions data from the supporting user API, then the issue is that we would have to change the DTO for the supporting user API and it would use service method from application and application database repository. The approach seems inconsistent with how we obtain data using services and repositories in the general sense. So I made a separate API for applications versions inside the application APIs folder. I'm not very convinced that we should include application version data in the supporting user API, but I'm open for strong suggestions.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -231,4 +233,28 @@ export class ApplicationAESTController extends BaseController {
...
+  async getApplicationVersionHistory(
+    @Param("applicationId", ParseIntPipe) applicationId: number,
+  ): Promise<ApplicationVersionAPIOutDTO[]> {
```
dheepak-aot, line 249: Our dev agreement (happened before few months) was to not return arrays directly, instead wrap the array inside an object. 

Example

```ts
export class ApplicationVersionAPIOutDTO {
  id: number;
  submittedDate: string;
}

export class ApplicationDetailAPIOutDTO {
  versions?: ApplicationVersionAPIOutDTO[];
}
```

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -231,4 +233,28 @@ export class ApplicationAESTController extends BaseController {
...
+    const applications =
+      await this.applicationService.applicationVersionHistory(applicationId);
+    if (!application) {
```
dheepak-aot, line 252: This is the application from express. 

image

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -231,4 +233,28 @@ export class ApplicationAESTController extends BaseController {
...
+      throw new NotFoundException("Application not found.");
+    }
+    return applications.map((application) => ({
+      id: application.id,
+      submittedDate: getPSTPDTDateHourMinute(application.submittedDate),
```
dheepak-aot, line 257: Was there a requirement to convert the date to PDTPST? Let me know if I am missing something.

I believe we should not convert the date to PDTPST and send only the raw date object from API. 

It can be formatted at Vue to desired date/date-time format.

sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
```diff
@@ -211,3 +211,8 @@ export class ApplicationWarningsAPIOutDTO {
   eCertFailedValidations: ECertFailedValidation[];
   canAcceptAssessment: boolean;
 }
+
+export class ApplicationVersionAPIOutDTO {
+  id: number;
+  submittedDate: string;
```
dheepak-aot, line 217: https://github.com/bcgov/SIMS/pull/4213/files#r1907698007

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
lewischen-aot: Yes, I was going with this approach until the compiler returns a circular dependency error with the following code:
```js
export class ApplicationVersionAPIOutDTO {
  versions: {
    id: number;
    submittedDate: string;
  }[];
}
``` 
image

But there is no error if we declare `ApplicationVersionAPIOutDTO` to be used in `ApplicationDetailAPIOutDTO` . I'll make the change.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
dheepak-aot, line 252: To validate if the application exist, the service has a method `doesApplicationExist`. Please use the same.

```ts
const application = await this.applicationService.doesApplicationExist({
      applicationId,
    });
    if (!application) {
      throw new NotFoundException("Application not found.");
    }
``

File: sources/packages/backend/apps/api/src/services/application/application.service.ts [hunk 6]
```diff
@@ -1837,6 +1841,35 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+  async applicationVersionHistory(
+    applicationId: number,
+  ): Promise<Application[]> {
+    const application = await this.repo.findOne({
+      select: { applicationNumber: true },
+      where: { id: applicationId },
+    });
+    if (!application) {
+      return null;
+    }
```
dheepak-aot, line 1859: To check if the application exists should not be the responsibility of this method. 

Ref: https://github.com/bcgov/SIMS/pull/4213/files#r1907713751

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1837,6 +1841,35 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    return this.repo.find({
+      select: { id: true, submittedDate: true },
+      where: {
+        id: Not(applicationId),
+        applicationNumber: application.applicationNumber,
+        applicationStatus: Not(ApplicationStatus.Draft),
```
dheepak-aot, line 1865: The `applicationStatus: Not(ApplicationStatus.Draft)` does not have any effect as draft applications will not have application number and will be eliminated by previous condition.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1837,6 +1841,35 @@ export class ApplicationService extends RecordDataModelService<Application> {
     });
   }
 
+  /**
+   * Get history of application versions for an application
+   * where the current application is excluded.
+   * @param applicationId application id.
+   * @returns assessment history list.
+   */
+  async applicationVersionHistory(
```
dheepak-aot, line 1850: I would like to clarify here the rationale of this service. 

IMO it should return all the previous versions of a given application id.

Which means if there are 5 application versions and if we receive the id of 3rd version, then the method must return id of `version 2` and `id of version 1`. This is not a possible situation from UI, but possible from API perspective.

Here is an example to illustrate what I said above.
```
Application

- version 5 (Current)

- version 4

- version 3

- version 2

- version 1

Method assumed name: getPreviousApplicationVersions(applicationId:number)

Scenario 1:

getPreviousApplicationVersions(5)

expected output:

4
3
2
1

Scenario 2:

getPreviousApplicationVersions(3)

expected output:
2
1

``

File: sources/packages/backend/libs/utilities/src/date-utils.ts [hunk 9]
```diff
@@ -99,6 +100,25 @@ export const getPSTPDTDateTime = (
...
+ */
+export const getPSTPDTDateHourMinute = (
+  date: string | Date,
+  local = false,
+): string => {
+  return dayjs(new Date(date))
+    .tz(PST_TIMEZONE, local)
+    .format(DATE_HOUR_MINUTE_FORMAT);
+};
+
```
dheepak-aot, line 121: This must not be required anymore.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
dheepak-aot: I agree on the part that we do not require to return the supporting users details on this API(New application API). It is not considered as part of the ticket. 

But to be clear, my suggestion was never to return application versions from existing supporting users API. 

Rather build the new API (`applicationId/versions`) to be more generic( generic endpoint name and generic DTO) retaining the same logic. 

Suggested generic endpoint: `:applicationId/overall-details`

Suggested generic DTO: 

```ts
export class ApplicationVersionAPIOutDTO {
  id: number;
  submittedDate: string;
}

export class ApplicationOverallDetailsAPIOutDTO {
  previousVersions?: ApplicationVersionAPIOutDTO[];
}
```

Let me know if that works. 

@andrewsignori-aot @andrepestana-aot @guru-aot @sh16011993  thoughts?

lewischen-aot: Ok. I can change the endpoint url to `:applicationId/overall-details` and use the DTO as suggested.

sources/packages/backend/apps/api/src/services/application/application.service.ts
lewischen-aot: This behavior is different from the first AC in the story:
image

Can you confirm that we only display application versions with submitted dates before the current application @Joshua-Lakusta @CarlyCotton @JerPearson?

dheepak-aot: To achieve the same, I would recommend this approach. 

```ts

async getPreviousApplicationVersions(
    applicationId: number,
  ): Promise<Application[]> {
    const application = await this.repo.findOne({
      select: { applicationNumber: true, submittedDate: true },
      where: { id: applicationId },
    });
    return this.repo.find({
      select: { id: true, submittedDate: true },
      where: {
        submittedDate: LessThan(application.submittedDate),
      },
      order: {
        submittedDate: "DESC",
      },
    });

```

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -159,11 +192,31 @@ export default defineComponent({
...
         }
       });
+      const applicationVersionHistory =
+        await ApplicationService.shared.getApplicationVersionHistory(
+          props.applicationId,
+        );
+      if (applicationVersionHistory.length) {
+        applicationVersionHistory.forEach((application) => {
+          applicationHistory.value.push({
+            title: `${application.submittedDate}`,
```
dheepak-aot, line 202: I belive business confirmed that we can Prefix with `Submitted on`

image

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -41,6 +41,27 @@
...
     />
+    <v-list v-if="applicationHistory.length" density="compact" nav>
+      <v-list-group
+        v-for="application in applicationHistory"
+        :key="application.title"
+        :prepend-icon="application.props?.prependIcon"
+        :title="application.title"
+        :value="application.title"
+      >
+        <template v-slot:activator="{ props }">
```
dheepak-aot, line 52: For slot, we use generally use the shortcut `#`. like this `<template #activator="{ props }">`

sources/packages/backend/apps/api/src/services/application/application.service.ts
lewischen-aot: Never mind. Dheepak is right about the history of application versions, and the current implementation returns the desired results.

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -41,6 +41,27 @@
...
+          <v-list-item v-bind="props" :title="application.title"></v-list-item>
+        </template>
+
+        <v-list-item
+          v-for="child in application.children"
+          :key="child.title"
+          :title="child.title"
+          :prepend-icon="child.props?.prependIcon"
+          @click="child.command"
```
dheepak-aot, line 61: As per my share in DEV chat, please refrain from using the `@click` and use `to:` rather.

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -159,11 +192,31 @@ export default defineComponent({
...
+      if (applicationVersionHistory.length) {
+        applicationVersionHistory.forEach((application) => {
+          applicationHistory.value.push({
+            title: `${application.submittedDate}`,
+            props: { prependIcon: "mdi-update" },
+            children: [
+              {
+                title: "Application",
+                props: { prependIcon: "mdi-school-outline" },
+                command: () => goToApplicationVersion(application.id),
```
dheepak-aot, line 208: please remove the command and set the `props:{to:}`

```ts
if (applicationVersionHistory.length) {
        applicationVersionHistory.forEach((application) => {
          applicationHistory.value.push({
            title: `${application.submittedDate}`,
            props: { prependIcon: "mdi-update" },
            children: [
              {
                title: "Application",
                props: {
                  prependIcon: "mdi-school-outline",
                  to: {
                    name: AESTRoutesConst.APPLICATION_DETAILS,
                    params: {
                      studentId: props.studentId,
                      applicationId: application.id,
                    },
                  },
                },
              },
            ],
          });
        });
```

sources/packages/web/src/services/ApplicationService.ts
```diff
@@ -201,4 +202,16 @@ export class ApplicationService {
...
+
+  /**
+   * Get history of application versions for an application
+   * where the current application is excluded.
+   * @param applicationId, application id.
+   * @returns history of application versions.
+   */
+  async getApplicationVersionHistory(
+    applicationId: number,
+  ): Promise<ApplicationVersionAPIOutDTO[]> {
```
dheepak-aot, line 214: Please take care of DTO as per previous discussions in service and API.

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -222,4 +223,19 @@ export class ApplicationApi extends HttpBaseClient {
...
     );
   }
+
+  /**
+   * Get history of application versions for an application
+   * where the current application is excluded.
+   * @param applicationId, application id.
+   * @returns application version list.
+   */
+  public async getApplicationVersionHistory(
```
dheepak-aot, line 233: Please remove public as it is implicit.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -656,7 +660,7 @@ export class ApplicationService extends RecordDataModelService<Application> {
       },
       where: {
         id: applicationId,
-        applicationStatus: Not(ApplicationStatus.Overwritten),
+        applicationStatus: applicationStatus,
```
dheepak-aot, line 663: Minor. condition can be expressed like this. 

image

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -591,8 +591,12 @@ export class ApplicationService extends RecordDataModelService<Application> {
       loadDynamicData?: boolean;
       studentId?: number;
       institutionId?: number;
+      allowOverwritten?: boolean;
```
dheepak-aot, line 594: Please update the method comment.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -74,7 +76,7 @@ export class ApplicationAESTController extends BaseController {
   ): Promise<ApplicationSupplementalDataAPIOutDTO> {
     const application = await this.applicationService.getApplicationById(
       applicationId,
-      { loadDynamicData },
+      { loadDynamicData, allowOverwritten: true },
```
dheepak-aot, line 79: To ensure the behavior of API to be un-changed I would strongly suggest to add the query param to the API and not as hard-coded. The default value can be set as false.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+            .institution,
+        institutionLocation:
+          application.currentAssessment.offering.institutionLocation,
+        student: application.student,
+        program: application.currentAssessment.offering.educationProgram,
+        offering: application.currentAssessment.offering,
+        programYear: application.programYear,
+      },
+      {
+        applicationStatus: ApplicationStatus.Completed,
```
dheepak-aot, line 70: Previous version of any given application cannot be in any status other than `Overwritten`. Hence either remove this version for test or set the status as overwritten.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
+    const endpoint = `/aest/application/${application.id}/versions`;
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .get(endpoint)
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.OK)
+      .then((response) => {
```
dheepak-aot, line 88: We can expect response like this. 

image

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        applicationNumber: application.applicationNumber,
+        offeringIntensity:
+          application.currentAssessment.offering.offeringIntensity,
+        pirStatus: application.pirStatus,
+        submittedDate: addDays(-10),
+      },
+    );
+
+    const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+
```
dheepak-aot, line 140: Please remove this blank line.

dheepak-aot:
Great start with the history. Please have a look at my comments.

sources/packages/backend/apps/api/src/services/application/application.service.ts
sh16011993: @dheepak-aot Do we need the `applicationStatus: Not(ApplicationStatus.Draft)` in the code snippet shared above?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    db = createE2EDataSources(dataSource);
+  });
+
+  it("Should get an array of application versions when there are an overwritten application and a completed application associated with the given application.", async () => {
+    // Arrange
+    const application = await saveFakeApplication(db.dataSource, undefined, {
```
sh16011993, line 29: Shouldn't we pass the `completedStatus` for the application here. I see that default action of application creation is submitted.

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -41,6 +41,27 @@
...
+        :key="application.title"
+        :prepend-icon="application.props?.prependIcon"
+        :title="application.title"
+        :value="application.title"
+      >
+        <template #activator="{ props }">
+          <v-list-item v-bind="props" :title="application.title"></v-list-item>
+        </template>
+
+        <v-list-item
```
sh16011993, line 56: I see that the dates aren't visible and the sidebar is not adjustable. Is there a better way to display it?

image

It's ok if the business is fine with this.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
lewischen-aot: The original application's status doesn't have to be completed in order to access its historical application versions. It's possible for an application in the assessment status to have several historical application versions.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationVersionHistory.e2e-spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  let app: INestApplication;
+  let db: E2EDataSources;
+
+  beforeAll(async () => {
+    const { nestApplication, dataSource } = await createTestingAppModule();
+    app = nestApplication;
+    db = createE2EDataSources(dataSource);
+  });
+
+  it("Should get an array of application versions when there are an overwritten application and a completed application associated with the given application.", async () => {
```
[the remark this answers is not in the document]
lewischen-aot: Removed the completed application as the endpoint will return only overwritten applications for historical application versions.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -231,4 +232,34 @@ export class ApplicationAESTController extends BaseController {
...
   }
+
+  /**
+   * Get application overall details for an application.
+   * @param applicationId application Id.
+   * @returns application overall details.
+   */
+  @ApiNotFoundResponse({
+    description: `Application not found or not on ${ApplicationStatus.Completed} status.`,
+  })
```
dheepak-aot, line 243: The documentation of not found response doesn't match with the actual. Meaning we expect error which just say `Application not found.`

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
lewischen-aot: "Submitted on" is now removed upon confirming with the business team. The format now uses "MMM DD YYYY HH:mm" to display the date time of the application version, similar to the samples in the technical AC.

sources/packages/backend/apps/api/src/route-controllers/application/application.aest.controller.ts
```diff
@@ -231,4 +232,34 @@ export class ApplicationAESTController extends BaseController {
...
     );
   }
+
+  /**
+   * Get application overall details for an application.
+   * @param applicationId application Id.
+   * @returns application overall details.
+   */
+  @ApiNotFoundResponse({
+    description: `Application not found.`,
```
dheepak-aot, line 242: Backticks not required anymore `"Application not found."`

sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -41,6 +41,26 @@
...
+        :key="application.title"
+        :prepend-icon="application.props?.prependIcon"
+        :title="application.title"
+        :value="application.title"
+      >
+        <template #activator="{ props }">
+          <v-list-item v-bind="props" :title="application.title"></v-list-item>
+        </template>
+        <v-list-item
+          v-for="child in application.children"
```
dheepak-aot, line 56: I am able to see that, when the scrollbar appears in nav(which occupies some width), the time (counter 1 in screenshot) is hidden in the group headers. If we remove the Icon(counter 2 in screenshot), it will solve this problem. Is the icon really essential for business?

image

sources/packages/web/src/types/AppRoutes.ts
```diff
@@ -91,6 +91,7 @@ export enum AppRoutes {
   SINManagement = "sin-management",
   CASSupplierManagement = "cas-supplier-management",
   ApplicationDetail = "student/:studentId/application/:applicationId",
+  ApplicationVersionDetail = "version/:applicationVersionId/view",
```
dheepak-aot, line 94: `versionApplicationId` will sound more accurate IMO.

sources/packages/web/src/router/AESTRoutes.ts
```diff
@@ -303,6 +303,15 @@ export const aestRoutes: Array<RouteRecordRaw> = [
               clientType: ClientIdType.AEST,
             },
           },
+          {
+            path: AppRoutes.ApplicationVersionDetail,
+            name: AESTRoutesConst.APPLICATION_VERSION_DETAILS,
+            props: true,
+            component: StudentApplicationView,
```
dheepak-aot, line 310: Now that we send the `versionid` we have to make a minor adjustment in `StudentApplicationView.vue` otherwise the API call to get application data will always use the `applicationId` and not the `versionId`.

Issue:

image

My best suggestion on solution:
image

dheepak-aot:
Thanks for doing the changes. Almost there added few more comments.

sources/packages/web/src/views/aest/StudentApplicationView.vue
```diff
@@ -45,6 +45,10 @@ export default defineComponent({
       type: Number,
       required: true,
     },
+    versionApplicationId: {
+      type: Number,
+      required: true,
```
dheepak-aot, line 50: My bad. In the screenshot I shared, I kept the required: true , but the required must be false.

dheepak-aot (approved):
Thanks for doing the changes. Great work. 👍

sh16011993 (approved):
Great work @lewischen-aot 👍

Issue #3899, opened 2024-11-06
Small Screen Responsiveness for modal - Student 


Joshua-Lakusta:
**User Story:**
As a student, I'd like to be able to access SIMS using my mobile phone as I do not have a tablet, laptop or computer and I am completing my studies with only my phone. In general, most of SIMS student account is mobile responsive, but needs some slight tweaks to ensure users can navigate. 

**Acceptance Criteria:**
- [ ] ensure modals are resized to fit small screen - 
- [ ] Allow for the modal to fill the screen
- [ ] Try to keep current modals (desktop/fullscreen) the same 

<img width="481" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNEptQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f920e076d96b2961c30ce871aff4e1aa830ae3c4/image.png" alt="image.png" />

**Technical**
- [ ] Check the feasibility of having the modal in full screen following https://vuetifyjs.com/en/components/dialogs/#fullscreen
- [ ] Student modals are present on the profile, student application, and student edit/submit. Check other also.
- [ ] As a starting point, ensure the below devices will show full screen.
  - Test on reference iPhone 12/Samsung Galaxy 12 using the browser developers.

PR #4227, opened 2025-01-13
#3899 - Small Screen Responsiveness for Modal - Student

sh16011993:


sources/packages/web/src/components/generic/ModalDialogBase.vue
andrewsignori-aot, line 1: Nice way to achieve the conditional modal view ❤️ 
Would the below not achieve the same?
_Please disregard the 500 although it produced a nice result during my tests also._
```ts
const mediaQuerySmallToLarge = window.matchMedia("(max-width: 500px)");
function handleScreenChangeSmallToLarge(e: MediaQueryList) {
  console.log(e);
  showFullScreen.value = e.matches;
}
mediaQuerySmallToLarge.addEventListener("change", () =>
  handleScreenChangeSmallToLarge(mediaQuerySmallToLarge),
);
handleScreenChangeSmallToLarge(mediaQuerySmallToLarge);
```

andrewsignori-aot: Please note that `(max-width: 500px), (max-height: 300px)`  can also be used to check with or height. Doing this we may need to add/remove the `with="auto"` also, so, when the modal expanded is used the `with="auto"` probably need to be removed. Not a blocker since it may be not part of the ticket.

andrewsignori-aot (commented):

lewischen-aot (approved):
LGTM, as the result matches the AC. Good job 👍

andrewsignori-aot (commented):
👍

Issue #4024, opened 2024-12-04
SFAS to SIMS Bridge - CSL/BCSL Overawards


CarlyCotton:
**User Story**
The layout of the bridge file from SFAS to SIMS was updated to match the actual data being sent. This also means correcting the existing data in our environments. Any value we have previously received as BCSL Overaward was actually CSL Overaward and vice versa. This is FT impacting ONLY.

**Acceptance Criteria**
- [x] Update all current Legacy overawards to swap what we have
    - [x]  BCSL to CSLF
    - [x]  CSLF to BCSL
- [x] Current production data must also be fixed with this
- [x] **ensure that bridge file process is stopped during release process for this ticket**
- [x] "Legacy Overawards" (origin_type) from disbursement_overaward table can all be cleared in production. Next bridge process will re-associate the correct overawards with existing student profile matches

<img width="445" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMm9kQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5c4e9b36b60fc75dbb11a080c0404ea251665caa/image.png" alt="image.png" />

<img width="537" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOEVGQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--94bf52bc07b3e60932b4f234cd90f1005fc9f1a3/image.png" alt="image.png" />

**Additional Context**
Bridge file document will be updated to reflect this change (the data has always been correct, this is just a swap of the labels)
<img width="1371" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN29GQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8260923dfe0fc55bb64ddf7e840d804115779fe7/image.png" alt="image.png" />

**Technical**
- [x] Adjust the current import fields to read the right fixed position for the two fields.
- [x] Create a DB migration to switch the values, ensuring a possible rollback. A suggestion is to create columns to allow switch/rollback or rename the columns to allow the switch. Explore the possibility of locking the table to enable the safe execution, if renaming the columns.
[...]

PR #4239, opened 2025-01-14
#4024 - SFAS to SIMS Bridge - CSL/BCSL Overawards - PR 1

lewischen-aot:
- Adjusted the current import fields to read the right fixed position for the two fields.
- Created a DB migration to swap the columns to allow the switch of their values.
- Adjusted E2E tests for the new field.

### Rollback evidence
image

Screenshot of the column values before and after the swap script
image


sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Update-swap-csl-bcsl-columns.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE
```
andrepestana-aot, line 1: If you do this, you're going to lose csl_overaward values, won't you?

andrepestana-aot: We've talked about maybe renaming the columns as a safe way to not lose data. I'm not sure this would work as the values from csl_overaward will be updated with bcsl_overaward and then bcsl_overaward won't be update as the values are the same.

andrepestana-aot: I saw that you left an image showing that there's no loss and that worked. Thanks for the syncup.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Update-swap-csl-bcsl-columns.sql
```diff
@@ -0,0 +1,5 @@
...
+    sims.sfas_individuals
+SET
+    csl_overaward = bcsl_overaward,
+    bcsl_overaward = csl_overaward;
```
andrepestana-aot, line 5: Sonar is complaining about the "where" clause. Maybe adding: "WHERE 1=1" would solve it.

andrewsignori-aot: Please do not change the code. I resolved the sonar cloud issue. In this case, it is a false positive since we want the query to be executed without a where clause.

sources/packages/backend/libs/integrations/src/sfas-integration/sfas-files/sfas-individual-record.ts
```diff
@@ -136,13 +136,13 @@ export class SFASIndividualRecord extends SFASRecordIdentification {
   /**
    * BC Student Loan total overaward balance (overaward_balance.overaward_balance_amt, overaward_balance_cde = "CSL").
    */
-  get cslOveraward(): number {
+  get bcslOveraward(): number {
```
andrewsignori-aot, line 139: Do the comments need to be swapped also?

andrewsignori-aot (approved):
Great solution to achieve the swap of the columns and values. Just a minor comment, hence approving it, looks good 👍 

Just to double-check, the below AC will be executed in an upcoming PR?
```
Create a DB migration to delete the records from the sims.disbursement_overawards where the origin type is Legacy overaward.
```

lewischen-aot:
> Great solution to achieve the swap of the columns and values. Just a minor comment, hence approving it, looks good 👍
> 
> Just to double-check, the below AC will be executed in an upcoming PR?
> 
> ```
> Create a DB migration to delete the records from the sims.disbursement_overawards where the origin type is Legacy overaward.
> ```

Yes, that's right. The upcoming PR will have this DB migration.

andrepestana-aot (approved):
Looks good.

sources/packages/backend/apps/db-migrations/src/sql/SFASIndividuals/Update-swap-csl-bcsl-columns.sql
```diff
@@ -0,0 +1,9 @@
...
+
+COMMENT ON COLUMN sims.sfas_individuals.csl_overaward IS 'Canada Student Loan total overaward balance (overaward_balance.overaward_balance_amt, overaward_balance_cde = "CSL").';
```
andrewsignori-aot, line 7: Thanks for fixing the SQL also ❤️

Issue #3054, opened 2024-03-28
Email notification for blocked disbursement to ministry only once


Joshua-Lakusta:
As a ministry user I want to receive only one notification email when a student disbursement is blocked. The ministry user wants this as they will create a ticket on each notification and only one notification is necessary.

**Acceptance Criteria**
- [x] Modify the process so that the ministry email address only receives this notification one time.

**Additional Context**
- linked to ticket 2536
- Original ticket requested 3x notifications 7 days a part for both student and ministry user. We want to change it so the student still receives the 3x notifications but we would like the ministry to only receive this email once.

**Technical Context**
- [x] Seperate the logic for student and Ministry in [Reference](https://github.com/bcgov/SIMS/blob/8fb0b9a28e99016f734acac33ab66026c690be53/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts#L37) . In the calling method. 
- [ ] Create a generic method to check , if the email notifications are already sent to a particular template and metadata( disbursementId)
- [x] Existing failing e2e needs to be updated, and create new E2E tests to ensure ministry notification once.

andrewsignori-aot:
@andrewsignori-aot I brought the ticket back to in progress to fix a log issue.

PR #4245, opened 2025-01-15
#3054 - Email notification for blocked disbursement to ministry only once

andrewsignori-aot:
- Created a base class for e-Cert notifications (`ECertNotification`) and segregated the logic to generate blocked notifications into `MinistryBlockedDisbursementNotification` and `StudentBlockedDisbursementNotification`.
- Existing logic was moved into the new files.
  - Code for the method `shouldCreateNotification` for the Students was moved and no changes were done for this PR, unless critical no comments will be addressed.
- `ECertNotification` can be used for any other e-Cert notification as required, not only blocked disbursement.

_Please note that the student data required to generate the notification is now retrieved twice in the event that the disbursement is blocked and the Ministry should be notified for the first time. The benefit of having the separation between the logic for the different notifications seems greater than an occasional extra query to the student's table. Please note that the query was simplified from what it used to be._

_Please also note that the below technical AC was not executed. The queries have different approaches, sometimes checking the existence, sometimes requiring a count, and sometimes requiring a count and a date last sent. I did not see a benefit for this PR. Please share an opinion/suggestion if needed._
```
Create a generic method to check, if the email notifications are already sent to a particular template and metadata( disbursementId).
```

### DIFF
--- a/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
+++ b/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
@@ -1,169 +1,34 @@
 import { Injectable } from "@nestjs/common";
-import {
-  DisbursementBlockedNotification,
-  NotificationActionsService,
-  StudentNotification,
-} from "@sims/services";
-import {
-  BLOCKED_DISBURSEMENT_MAXIMUM_NOTIFICATIONS_TO_SEND,
-  BLOCKED_DISBURSEMENT_NOTIFICATION_MIN_DAYS_INTERVAL,
-} from "@sims/services/constants";
-import {
-  DisbursementSchedule,
-  Notification,
-  NotificationMessageType,
-} from "@sims/sims-db";
-import { dateDifference } from "@sims/utilities";
 import { ProcessSummary } from "@sims/utilities/logger";
 import { EntityManager } from "typeorm";
-
-interface NotificationData {
-  maxCreatedAt: Date;
-  notificationCount: number;
-}
+import { MinistryBlockedDisbursementNotification } from "./blocked-disbursement/ministry-blocked-disbursement-notification";
+import { StudentBlockedDisbursementNotification } from "./blocked-disbursement/student-blocked-disbursement-notification";
+import { EligibleECertDisbursement } from "../disbursement-schedule.models";
 
 @Injectable()
 export class ECertNotificationService {
   constructor(
-    private readonly notificationActionsService: NotificationActionsService,
+    private readonly ministryBlockedDisbursementNotification: MinistryBlockedDisbursementNotification,
+    private readonly studentBlockedDisbursementNotification: StudentBlockedDisbursementNotification,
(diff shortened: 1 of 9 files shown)


sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.service.ts
```diff
@@ -1,169 +1,36 @@
...
-    return (
-      !notification.notificationCount ||
-      (notification.notificationCount <
-        BLOCKED_DISBURSEMENT_MAXIMUM_NOTIFICATIONS_TO_SEND &&
-        canSendNewNotification)
+    const notifications = [
+      this.ministryBlockedDisbursementNotification,
+      this.studentBlockedDisbursementNotification,
+    ].map((notification) =>
+      notification.notify(eCertDisbursement, entityManager, log),
```
lewischen-aot, line 32: ❤️ What a great way to segregate the two services.

lewischen-aot (approved):
LGTM👍Great way to segregate different notification services ❤️

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.ts
```diff
@@ -0,0 +1,70 @@
+import { EntityManager } from "typeorm";
+import { EligibleECertDisbursement } from "../disbursement-schedule.models";
+import { ProcessSummary } from "@sims/utilities/logger";
+
+/**
+ * Provides a basic structure for creating and
+ * sending notifications related to e-Cert disbursements.
+ */
+export abstract class ECertNotification {
```
dheepak-aot, line 9: Great job! I love the way how the way how the e-cert based notifications are handled here with extensibility to add more notifications.

dheepak-aot, line 9: One minor comment to mention. As this is a base class to create notifications, the student data is a core common part that can be built from a base class method. And that base class method can be called inside `createNotification()`

dheepak-aot:
Amazing work @andrewsignori-aot. Just one comment.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-notification/e-cert-notification.ts
andrewsignori-aot: The idea was that each notification could evolve differently and the data required for each should be handled in the inherited class. For instance, a notification for the Ministry could potentially contain further information about the blocked disbursement or something else. Assuming all notifications will have the student data as a core would not be accurate IMO.

dheepak-aot: Excluding ministry emails for institutions, other ministry notifications have student first and last name as common. But not a blocker considering the present.

dheepak-aot (approved):
Great Job! Looks good👍

PR #4251, opened 2025-01-15
#4024 - SFAS to SIMS Bridge - CSL/BCSL Overawards - PR 2

lewischen-aot:
- Created a DB migration to delete the records from table `sims.disbursement_overawards` where the origin type is "Legacy overaward".
- Created an E2E test to ensure that disbursement over awards records are created during student internal creation with exact match from SFAS individual data records.
- Created a SQL script in the description for the release instruction to "ensure that bridge file process is stopped during release process for this ticket".

```sql
UPDATE
    sims.queue_configurations 
SET is_active = false
WHERE queue_name = 'sfas-integration';
```
### Rollback evidence
Screenshot 2025-01-15 154033

Please note that the `down` method for the DB migration class doesn't contain a revert script for rollback, where a comment is used instead to explain the reason and recovery process.


sources/packages/backend/apps/api/src/route-controllers/student-account-applications/_test_/e2e/student-account-application.aest.controller.approveStudentAccountApplication.e2e-spec.ts
```diff
@@ -245,6 +245,73 @@ describe("StudentAccountApplicationAESTController(e2e)-approveStudentAccountAppl
     });
   });
 
+  it("Should create disbursement overawards when an exact match is found with matching last name, birth dates and SIN for importing a student record with disbursement overawards from SFAS.", async () => {
```
andrewsignori-aot, line 248: This is a valid test and we can keep it, it is just not the code that would revert the deleted records.
The process would happen during the execution of the SFAS scheduler (sfas-integration.scheduler) during the [updateDisbursementOverawards](https://github.com/bcgov/SIMS/blob/5b9f09a6191dbd80be92cb005677e45d07e1983a/sources/packages/backend/libs/integrations/src/services/sfas/sfas-individual-import.service.ts#L144).
To create an E2E test, it needs to be under queue-consumers importing an SFAS bridge record for an individual with overawards data.

andrewsignori-aot (commented):
Changes looks good, the E2E created makes sense for the student creation scenario but the most critical to ensure the data would be recovered if deleted would be the one in queue-consumers. I added the details on the comment, please let me know if need further information.

sh16011993 (approved):
Nice work @lewischen-aot 👍

sources/packages/backend/apps/api/src/route-controllers/student-account-applications/_test_/e2e/student-account-application.aest.controller.approveStudentAccountApplication.e2e-spec.ts
andrewsignori-aot: @sh16011993 the current test is this tile is testing the API call and should stay where it is.
My comment, if not clear, is to have a second E2E test created for the queue-consumers.
legacy overawards are imported during student account creation and during the SFAS bridge process.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/sfas-integration/_tests_/sfas-integration.scheduler.e2e-spec.ts
```diff
@@ -505,6 +510,93 @@ describe(describeProcessorRootTest(QueueNames.SFASIntegration), () => {
...
+        },
+      });
+      expect(disbursementOverawards).toEqual(
+        expect.arrayContaining([
+          { overawardValue: 714.3 },
+          { overawardValue: 741.3 },
+        ]),
+      );
+    },
+  );
```
dheepak-aot, line 598: Thanks for adding the E2E test to assert the import. While validating the disbursement overawards, please also validate each overward against the code. 

```ts
const disbursementOverawards = await db.disbursementOveraward.find({
        select: { disbursementValueCode: true, overawardValue: true },
        where: {
          student: { id: student.id },
          originType: DisbursementOverawardOriginType.LegacyOveraward,
        },
      });
      expect(disbursementOverawards).toEqual(
        expect.arrayContaining([
          {
            disbursementValueCode: BC_STUDENT_LOAN_AWARD_CODE,
            overawardValue: 714.3,
          },
          {
            disbursementValueCode: CANADA_STUDENT_LOAN_FULL_TIME_AWARD_CODE,
            overawardValue: 741.3,
          },
        ]),
      );
```

dheepak-aot (approved):
Thanks for doing the changes @lewischen-aot. Looks good👍

Issue #3685, opened 2024-09-11
Deprecation of DeploymentConfig


andrewsignori-aot:
- Start with one service and create subsequent tickets for the rest

What is happening?
Plattaform Services Team is upgrading to OCP 4.14 in 2024 Q3. Coming to this new version, the DeploymentConfig object is being deprecated. There is no immediate effect after the upgrade, any existing DeploymentConfig workloads will continue to run, but eventually this feature will be removed in future upgrades. The recommendation is to use Deployment instead of DeploymentConfig.

When?
OCP4.14 upgrade is scheduled for Silver in mid September, Gold and GoldDR in October. After the upgrade, DeploymentConfig objects will be deprecated and you will start to see warning about the use of them. While DeploymentConfigs are still supported, they are not recommended for any new installations. Only security-related and critical issues will be fixed by RedHat. There is no specific timeline on when DeploymentConfig will stop working on OCP yet, but teams should start converting to Deployment now.

Do I need to do anything?
If you are creating new workloads, make sure to use Deployment instead of DeploymentConfig.
If you have existing DeploymentConfig workloads, please follow this [RedHat guidance to convert them to Deployment objects](https://developers.redhat.com/learning/learn:openshift:replace-deprecated-deploymentconfigs-deployments/resource/resources:convert-deploymentconfig-deployment?mc_cid=1a844df827&mc_eid=f049851a15).

**Technical**
- [x] Pick one of the existing deployments (suggested `web-sims`) and follow the conversion guide to establish the path to be applied to all other deployments.
- [x] If time allows, execute the conversion of all deployments. If not, create a ticket to continue the work.

dheepak-aot:
All the deployment configs are converted to deployment during this implementation.
Also the part `fetch old tags` to prune has been updated to use deployment to find current deployed image.

dheepak-aot:
Deployment instructions updated to: https://app.zenhub.com/workspaces/student-information-management-system-5fce9df5aa1b45000e937014/issues/zh/276

PR #4256, opened 2025-01-16
#3685 - Convert FormIO to deployment and update fetch old tags script

dheepak-aot:
## Convert to Deployment

- [x] Convert FormIO DC to deployment

## Update fetch old tags script

- [x] Updated fetch old tags script to get the deployed version of  `image steam` based on deployment instead of DC.
image

## Removed the `creationTimestamp` from metadata in form

- [x] Removed `creationTimestamp:null` in metadata from formio objects as it is not followed in other configs in our application.

### DIFF
--- a/devops/Makefile
+++ b/devops/Makefile
@@ -511,7 +511,7 @@ deploy-forms:
 	-p TLS_KEY=$(TLS_KEY) \
 	-p TLS_CA_CERTIFICATE=$(TLS_CA_CERTIFICATE) \
 	| oc -n $(NAMESPACE) apply -f -
-	$(call rollout_and_wait,dc/$(FORMS_NAME))
+	$(call rollout_and_wait,deployment/$(FORMS_NAME))
 
 # Remove redis and resources including secrets from openshift namespace.
 delete-redis:
--- a/devops/scripts/fetchOldTags.sh
+++ b/devops/scripts/fetchOldTags.sh
@@ -59,7 +59,8 @@ IS_NAMESPACE="${LICENSE_PLATE}-tools"
 DC_NAME="${ENV}-${APP_NAME}"
 
 # Lookup the dc to get the deployed container image tag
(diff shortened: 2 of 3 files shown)


devops/Makefile
```diff
@@ -511,7 +511,7 @@ deploy-forms:
 	-p TLS_KEY=$(TLS_KEY) \
 	-p TLS_CA_CERTIFICATE=$(TLS_CA_CERTIFICATE) \
 	| oc -n $(NAMESPACE) apply -f -
-	$(call rollout_and_wait,dc/$(FORMS_NAME))
+	$(call rollout_and_wait,deployment/$(FORMS_NAME))
```
andrepestana-aot, line 514: 👍

devops/scripts/fetchOldTags.sh
```diff
@@ -59,7 +59,8 @@ IS_NAMESPACE="${LICENSE_PLATE}-tools"
 DC_NAME="${ENV}-${APP_NAME}"
 
 # Lookup the dc to get the deployed container image tag
-DC_IMAGE=$(oc get dc/$DC_NAME -n $DC_NAMESPACE -o json | jq -r '.spec.template.spec.containers[].image')
+# TODO: Change the naming convention from DC to deployment if required in future.
+DC_IMAGE=$(oc get deployment/$DC_NAME -n $DC_NAMESPACE -o json | jq -r '.spec.template.spec.containers[].image')
```
andrepestana-aot, line 63: Not a problem: just notice that all the variables start with "DC" 😄

andrepestana-aot (approved):
Looks good to me.

devops/scripts/fetchOldTags.sh
dheepak-aot: That is the reason behind adding the TODO comment. 😊

andrewsignori-aot (approved):
Great work, nice catch taking care of the prune scrip also. Looks good 👍

Issue #3792, opened 2024-10-15
Student: FT Residency exception for dependant students


JerPearson:
Solution the residency issue where students answer residency question and are potentially put down an exception path when they are depedant

**Additional Context**
- enter text here
- enter text here

**Acceptance Criteria**
- [ ] Given (Context), When (action carried out), Then (expected outcome)
- [ ] Given (Context), When (action carried out), Then (expected outcome)

andrewsignori-aot:
@JerPearson related to the below AC, after the recent finding about the dependency on the answer "Is your parent(s)/step-parent/sponsor/legal guardian a resident of B.C.?", should we hide the "Residency information" till all the questions are answered, which means the question will only be displayed on the "Personal information" tab once the question in the "Parent information" is answered?
```
FT ONLY: Residency information will be hidden until independent = true
```

JerPearson:
@andrewsignori-aot this will be addressed on future tickets when cleaning up the overall parent process.  Two of the options included on the parent residency question that could impact DEP vs IND status will need to be moved.  They aren't related to residency and need to be there own stand alone questions.  They will likely be included on the personal information page but further analysis needed.

andrepestana-aot:
@JerPearson Please add an AC regarding the full-time labour force question to be considered to display the Parent Information Tab (it was not considering and I added it in this ticket).

JerPearson:
@andrepestana-aot AC added.  Thank you for catching this and all the work on this ticket.

PR #4261, opened 2025-01-17
#3792 - Student: FT Residency exception for dependant students - Residency information logic fix

andrepestana-aot:
Residency information logic fix:
- Part-time is always considered "independant".
- Full-time student will be considered "independant":
  - if  parent information does not need to be entered after all the questions in the personal information are answered or;
  - if parent information needs to be entered and student 
  	- has both the parents passed away or; 
  	- is estranged from both of parents. 
- Otherwise, student is considered "dependant".


lewischen-aot (approved):
The improved logic looks great 👍 Thanks for making the improvement.

dheepak-aot (approved):
Thanks for doing the changes and taking the additional step @andrepestana-aot . Looks good from the perspective of this ticket.

Issue #3929, opened 2024-11-13
Student: Parent Info: Self reporting trigger


JerPearson:
As a student I need to be asked if my parent has a valid BCSC account to determine if self reporting income is necessary.

**Acceptance Criteria**

- [x] Make these changes for the 25-26 program year application form
- [x] Revise wording from "Does your parent(s)/step-parent/sponsor/legal guardian have a valid Social Insurance Number?'" to:
> Is your parent(s)/step-parent/sponsor/legal guardian eligible for a BC Services Card account?
- [x] If above question = true, populate purple information banner directly under the question with the following content:
> **Please be advised your parent(s)/step-parent/sponsor/legal guardian will need to sign in to StudentAid BC and provide additional information**  
> Parent(s)/step-parent/sponsor/legal guardian must provide their information to StudentAid BC by logging in with a B.C. Services Card. If they are ineligible for a B.C. Services Card, select "No" to this question and provide information on their behalf.

- [ ]  Example for formatting the above.  Use content listed in above AC, not screenshot
<img width="848" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBKzY1Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7fc64016b9b002d9dcdb3bf034d4533b9027bf49/image.png" alt="image.png" />

<img width="675" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1c1Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1050c45f4df3fc6b8b961d32ddf1dde668b4d521/image.png" alt="image.png" />

Issue #3851, opened 2024-10-29
Student: Family Information Tab Revisions


CarlyCotton:
Repurpose the family information tab to only include dependant information.

**Acceptance Criteria**
- [x]  Make these changes for the 25-26 program year application form
- [x] Only show this tab if Dependant question on personal information page = True 
- [x]  Apply content changes to both FT and PT application
- [x]  Rename "Family information" tab to " Dependant information"
- [x]  Revise "Family Information" header to "Dependant information"
- [x]  Revise: "Please provide information about your family" to "Please provide information about your dependant(s)."
- [x]  Remove "Family details" sub header
<img width="792" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMitLQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--50a6caab211d18cf426ec82cd81e65b16b24b482/image.png" alt="image.png" />

- [x] Move the following two questions from the financial info tab to this page
- [x] These questions will go beneath the "Add dependants" box in their own category named "Dependant care costs".
- [x] Revise text of these questions to:

> During your study period will you have unsubsidized day-care costs during class hours for children 11 years of age or under?

> During your study period will you have unsubsidized day-care costs during class hours for disabled/infirm dependants 12 years of age or older?

<img width="755" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3FCQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a799202cc128fa4e8f78103f67221849e83b04c7/image.png" alt="image.png" />

**Additional Context**
- IMPORTANT:  For FT only the "Dependant Care costs" category will not be moved, just the questions.  There are other dependant care cost questions that need to remain on the financial page.

CarlyCotton:
Closing - slight ac shuffle into 3929

PR #4273, opened 2025-01-21
#3929 - Student Parent Self report content updates

dheepak-aot:
## Student Parent Self report content updates

- [x] Content update to the question `Eligible for BC Services Card`.
image
- [x] Content update for the banner when YES is answered to `Eligible for BC Services Card` question.
image
- [x] Moved the info and warning banner right under `Eligible for BC Services Card` question instead of the question `How many parents do you live with`.
image

## Fix for #3851 

- [x] Moved the `Dependant care costs` to it's own panel
image


andrewsignori-aot (approved):
Nice work and thanks for taking care of adding the panel, looks good 👍

lewischen-aot (approved):
LGTM 👍

Issue #3254, opened 2024-05-09
CAS AP Invoice - Invoices and Batches


HRAGANBC:
**User Story:** 
As a member of the SABC Finance team (Ministry User), I need to be able to review and approve a detailed inventory of the previous day’s PT and FT provincial grant disbursements so that SIMS can notify CAS about these disbursements and maintain up to date records of a student’s taxable income. This summary will need to have a sub total of each award and a total for the batch.
Each Invoice should relate to a single disbursement for an individual student. It will need to have two line items per award under the invoice.
Every Invoice will be included in a daily batch of invoices. 
Once approved, the entire batch is sent to CAS. We need to record the outcome for each invoice, success or error

# **Acceptance Criteria:**
## Access/Roles
- [ ] Add a new role to business-administrator group for all AP invoicing activities (view/download batch and approve)

## Invoices
- [ ] Every student who receives provincial grants (confirmed via disbursement receipt) will need an invoice PER doc/cert number. _(This should normally be one invoice per disbursement receipt but could have odd edge cases where two doc numbers on a single disbursement receipt)_
- [ ] Every invoice will need to contain two line items PER grant disbursed. (SBSD, BCAG-PT, BCAG-FT, BGPD)
    - [ ] Each award will need a Debit Line and a Credit Line
    - [ ] The Debit Line will need a Distribution Account unique to the Award
    - [ ] The Credit Line will need a Distribution Account unique to the Award
- [ ] Distribution Accounts per DR/CR per Award:
    - [ ] TEST/Staging Dist Accounts: Placeholders fine for now
    - [ ] [PROD Dist Accounts](https://bcgov.sharepoint.com/:t:/r/teams/03177/Shared%20Documents/DEVS/CAS/PROD_DistributionAccounts.txt?csf=1&web=1&e=oYjjxt)
- [ ] Every invoice needs a unique invoice number: For now, "supplier number concatenated with record id" or similar.
- [ ] Full list of fields for the invoice and invoice lines: [AP Invoice Web Service Field Requirements -General.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0lxQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8946db1f2dc7f2f0f40e6b0a3009180e0861b5de/AP%20Invoice%20Web%20Service%20Field%20Requirements%20-General.xlsx)
[...]

PR #4289, opened 2025-01-24
#3254 - CAS AP Invoice - Invoices and Batches DB Migrations and Models

andrewsignori-aot:


sources/packages/backend/apps/db-migrations/src/sql/Types/Create-cas-invoice-status.sql
```diff
@@ -0,0 +1 @@
+CREATE TYPE sims.cas_invoice_status AS ENUM ('Pending', 'Sent', 'Manual intervention');
```
dheepak-aot, line 1: Just a minor thought. Will it be more useful, if we create in the order of `'Pending', 'Manual intervention',  'Sent'` if we are trying to sort invoices by invoice status without using case just following natural enum order. ?

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
```diff
@@ -0,0 +1,44 @@
+CREATE TABLE sims.cas_distribution_accounts(
+  id SERIAL PRIMARY KEY,
+  award_value_code VARCHAR(10) NOT NULL,
+  offering_intensity sims.offering_intensity NOT NULL,
+  operation_code CHAR(2) NOT NULL,
```
dheepak-aot, line 5: What could be the benefit of not using enum here?

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
```diff
@@ -0,0 +1,44 @@
...
+  distribution_account VARCHAR(40) NOT NULL,
```
dheepak-aot, line 6: I see that this column is not mentioned in technical ACs. So is it added to specify the particular distribution account name/code ?

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
```diff
@@ -0,0 +1,44 @@
...
+  modifier INT NULL DEFAULT NULL REFERENCES sims.users(id)
+);
+
+CREATE UNIQUE INDEX cas_distribution_accounts_award_value_code_offering_intensity_operation_code ON sims.cas_distribution_accounts(
+  award_value_code,
+  offering_intensity,
+  operation_code
+)
+WHERE
+  (is_active = TRUE);
```
dheepak-aot, line 21: 👍

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
```diff
@@ -0,0 +1,44 @@
...
+  distribution_account VARCHAR(40) NOT NULL,
+  is_active BOOLEAN NOT NULL,
+  -- Audit columns
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  creator INT NOT NULL DEFAULT NULL REFERENCES sims.users(id),
+  modifier INT NULL DEFAULT NULL REFERENCES sims.users(id)
+);
+
+CREATE UNIQUE INDEX cas_distribution_accounts_award_value_code_offering_intensity_operation_code ON sims.cas_distribution_accounts(
```
dheepak-aot, line 15: Comment on Index is missing.

sources/packages/backend/apps/db-migrations/src/sql/CASInvoiceBatches/Create-cas-invoice-batches.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TABLE sims.cas_invoice_batches(
+  id SERIAL PRIMARY KEY,
+  batch_name VARCHAR(50) NOT NULL,
+  batch_date DATE NOT NULL,
```
dheepak-aot, line 4: Minor. Technical AC mentions batch date as timestamp. But date sounds good to me.

sources/packages/backend/apps/db-migrations/src/sql/Types/Create-cas-invoice-status.sql
andrewsignori-aot: There is no clear sorting requirement right now.

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
andrewsignori-aot: I didn't find a strong reason to use one, so I kept it simple.
The idea was to keep the awards associated with accounts and "operation code" where operation codes are open to any string of two characters to be sent to CAS.

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
andrewsignori-aot: It was not mentioned in the technical AC(I can add) but it represents the below business AC
```
Distribution Accounts per DR/CR per Award
```
Also in the spreadsheet provided by the business, as below.
image

sources/packages/backend/apps/db-migrations/src/sql/CASInvoiceBatches/Create-cas-invoice-batches.sql
andrewsignori-aot: Yes, there was some back and forth about where it would be used.
I will double-check, if it will be used only by SIMS I will revert to timestamp.

sources/packages/backend/apps/db-migrations/src/sql/CASInvoiceBatches/Create-cas-invoice-batches.sql
```diff
@@ -0,0 +1,37 @@
...
+  approval_status sims.cas_invoice_batch_approval_status NOT NULL,
+  approval_status_updated_on TIMESTAMP WITH TIME ZONE NOT NULL,
+  approval_status_updated_by INT NOT NULL REFERENCES sims.users(id),
+  -- Audit columns
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  creator INT NOT NULL REFERENCES sims.users(id),
+  modifier INT NULL DEFAULT NULL REFERENCES sims.users(id),
+  UNIQUE (batch_name)
```
dheepak-aot, line 13: 👍

sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Create-cas-invoices.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE sims.cas_invoices(
+  id SERIAL PRIMARY KEY,
+  cas_invoice_batch_id INT NOT NULL REFERENCES sims.cas_invoice_batches(id),
+  disbursement_receipt_id INT NOT NULL REFERENCES sims.disbursement_receipts(id),
+  cas_supplier_id INT NOT NULL REFERENCES sims.cas_suppliers(id),
```
dheepak-aot, line 5: From SIMS perspective will it be better to have student_id here? So that if there is a change in supplier between invoice generation and invoice being sent, it will be handled? Or will that approach create undesired effects?

sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Create-cas-invoices.sql
```diff
@@ -0,0 +1,43 @@
...
+  invoice_number VARCHAR(40) NOT NULL,
+  invoice_status sims.cas_invoice_status NOT NULL,
+  invoice_status_updated_on TIMESTAMP WITH TIME ZONE NOT NULL,
```
dheepak-aot, line 8: Minor. I believe the `invoice status updated by `is not added considering that the status will be updated on by system. But will it be worth considering now(expecting a manual update) or when requirement comes?

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
andrewsignori-aot: Comments added.

sources/packages/backend/apps/db-migrations/src/sql/CASInvoiceBatches/Create-cas-invoice-batches.sql
andrewsignori-aot: Converted to timestamp.

sources/packages/backend/libs/sims-db/src/entities/cas-invoice-batch.model.ts
```diff
@@ -0,0 +1,72 @@
...
+   */
+  @Column({
+    name: "approval_status_updated_on",
+    type: "timestamptz",
+  })
+  approvalStatusUpdatedOn: Date;
+  /**
+   * User that changed the status last time.
+   */
+  @ManyToOne(() => User)
```
dheepak-aot, line 59: I'm not sure about the default value of nullable in relational decorators. If that is not sure we can mention on relations explicitly. 

```ts
 @ManyToOne(() => User, { nullable: false })
  @JoinColumn({
    name: "approval_status_updated_by",
  })
  approvalStatusUpdatedBy: User;
```

sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Create-cas-invoices.sql
andrewsignori-aot: The student will be retrieved using `disbursement_receipt_id`, no need to be added as a column.
For now the idea would be to keep the `cas_supplier_id` associated during the batch creation to allow the user to review it as part of the snapshot. The `supplier_id` used to generate the batch or even the one used to call the API should be kept for history.
I can double-check if any action is expected in case the supplier number was changed between batch generation and API call. IMO it would just cause a manual intervention only.
If `the supplier_id` does not need to be part of the snapshot approval we can change this column to optional, but being the valid supplier part of the conditions to invoice generation I still feel it must be part of the snapshot.

sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Create-cas-invoices.sql
andrewsignori-aot: Since the initial status would be pending, the idea was to keep both updated upon record creation.

sources/packages/backend/libs/sims-db/src/entities/cas-invoice-status.type.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export enum CASInvoiceStatus {
+  /**
+   * Record was created and has not been sent to CAS.
+   */
+  Pending = "Pending",
+  /**
+   * Record was sent to CAS.
+   */
+  Approved = "Sent",
```
dheepak-aot, line 12: Approved = Sent ?

sources/packages/backend/libs/sims-db/src/entities/cas-distribution-account.model.ts
```diff
@@ -0,0 +1,57 @@
...
+    name: "award_value_code",
+  })
+  awardValueCode: string;
+  /**
+   * Offering intensity to allow same awards to potentially have
+   * different distribution accounts for a same award code.
+   */
+  @Column({
+    name: "offering_intensity",
+    enum: OfferingIntensity,
```
dheepak-aot, line 29: Don't we require `type: "enum"`?

sources/packages/backend/libs/sims-db/src/entities/cas-invoice.model.ts
```diff
@@ -0,0 +1,103 @@
...
+  @ManyToOne(() => DisbursementReceipt)
+  @JoinColumn({
+    name: "disbursement_receipt_id",
+    referencedColumnName: ColumnNames.ID,
+  })
+  disbursementReceipt: DisbursementReceipt;
+  /**
+   * Active CAS supplier associated with the student at the moment the invoice was created.
+   */
+  @ManyToOne(() => DisbursementReceipt)
```
dheepak-aot, line 51: CASSupplier ? (instead of DisbursementReceipt)

dheepak-aot:
Great work @andrewsignori-aot. Please take a look at the comments.

sources/packages/backend/libs/sims-db/src/entities/cas-distribution-account.model.ts
andrewsignori-aot: Yes, we do and we have it at least in 55 places. I did not find a reason why the `OfferingIntensity` was not using it in other places where the enum is present.
We had  some issues with `enum` a long time ago and if I am not wrong the understanding was to use it alongside the enum name.
It is also part of [the wiki examples](https://github.com/bcgov/SIMS/wiki/Development---Standards,-Decisions,-and-Troubleshooting#typeorm-database-models).

Found the below AC also in the PR #1036
```
Adding enumName to all properties of entities which reference a DB enum object.
```

sources/packages/backend/libs/sims-db/src/entities/cas-invoice-batch.model.ts
dheepak-aot: If the property is required. Please take care of other places as well.

andrewsignori-aot: Added the `nullable: false` since the typeorm docs state otherwise as default.

sources/packages/backend/libs/sims-db/src/entities/cas-invoice-status.type.ts
andrewsignori-aot: Fixed.

sources/packages/backend/libs/sims-db/src/entities/cas-distribution-account.model.ts
andrewsignori-aot: I added the type to be aligned with other `enum` types but I did not find a reason why only the `OfferingIntnsity` has been declared without other common properties.
```ts
 @Column({
    name: "offering_intensity",
    nullable: false,
  })
  offeringIntensity: OfferingIntensity;
```
For instance, there was a PR in the past to add `enumName` to all the enum and later it was removed only for the `offering_intensity` column. Was it because it is the only one that is shared across multiple tables?

sources/packages/backend/apps/db-migrations/src/sql/CASDistributionAccounts/Create-cas-distribution-accounts.sql
```diff
@@ -0,0 +1,46 @@
...
```
guru-aot, line 15: 👍

sources/packages/backend/libs/sims-db/src/entities/cas-distribution-account.model.ts
guru-aot: Not sure, I dont remember why we  removed it?

sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Create-cas-invoices.sql
dheepak-aot: Got it. I agree that `cas_supplier_id` being saved as a part of invoice would be critical to audit what information was sent to CAS.

dheepak-aot (approved):
Thanks for doing the changes @andrewsignori-aot. Looks good 👍

sources/packages/backend/libs/sims-db/src/entities/cas-distribution-account.model.ts
andrewsignori-aot: The plan is to add this new column with the properties that we use for other enums and run some tests to see if some issues arise.

guru-aot (approved):
LGTM, nice start @andrewsignori-aot

Issue #4014, opened 2024-11-29
Copy of Institution Read-Only User Type


Joshua-Lakusta:
**User Story**
As an institution user i need to be able to add users that have read only access. Introduce a new user role for all institution accounts which permits read-only capabilities. 

Institution stakeholders - in particular BC Public schools - have emphasized that they have large FAO offices with numerous staff, many of whom perform strictly 'advisory' roles to the student, and are not authorized to make any changes which affect a student's application or the school's list of programs and offerings. Given the risks with taking such action, they would prefer the ability to assign a 'read-only' ability to any given user. 

**Acceptance Criteria:**
- [ ] Add a new user role called 'read-only' located between user and no access 

<img width="657" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNER5Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--63ab6984990988165ee44630c52d67b0a9c10b39/image.png" alt="image.png" />

- [ ] Adjust the "Add Users" screen to describe the 'read-only' role 
<img width="322" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGZ5Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e77ab26ab0bac4e951333e9e60d9fdae7aff12d1/image.png" alt="image.png" />

- [ ] Add a new role "user type" with no other connected accesses "user roles"(see screenshot)

<img width="741" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNG55Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a486508212bf81da7b556b8a3a4f31266253c5a9/image.png" alt="image.png" />

- [ ] Use the base "User" role design and then disable the following:

- [ ]  Prevent user from modifying any program related data
- [ ]  Prevent user from changing any offering related data
- [ ]  'Complete Program Info Request' button
- [ ]  'Application Action' drop down button on the 'Confirm Enrolment' screen of a particular application
- [ ]  'Submit update' on the 'report a change' screen of a particular application
- [ ]  'Submit requested change' button on the 'Request an application change' screen of a particular application 

[...]

PR #4325, opened 2025-02-06
#4014 - Institution Read-Only User Type PART 2 (UI) (Part 1)

lewischen-aot:
As part of the story, this first PR has the following changes:
- Added a new user role called 'read-only' located between user and no access on the UI.
- Updated the UI to prevent actions in the below areas.
  - **Program and Offering**: prevent user from changing any offering related data
  - **PIR**: 'Complete Program Info Request' button
  - **Confirm enrolment**: 'Application Action' drop down button on the 'Confirm Enrolment' screen of a particular application
  - **Report a change**: 'Submit update' button on the 'report a change' screen of a particular application
  - **Request a change**: 'Submit requested change' button on the 'Request an application change' screen of a particular application
- Ensured buttons are disabled across all mentioned areas in the business ACs.
- Ensured routes are disallowed for **program-creation** and **offering-creation** pages for read-only-users.
- Adjusted the "Add Users" screen to describe the 'read-only' role

Screenshot of the assign-user-role model
Screenshot 2025-02-06 114420

Screenshot of the added item in the list
image


sources/packages/web/src/components/common/OfferingForm.vue
```diff
@@ -94,8 +94,13 @@ export default defineComponent({
       required: true,
       default: false,
     },
+    locationId: {
+      type: Number,
+      required: false,
+    },
   },
   setup(props, context) {
+    const { isReadOnlyUser } = useInstitutionAuth();
```
andrewsignori-aot, line 103: This is a base and generic component shared between Ministry and Institution. 
This component also offers a "read-only" mode through the property `formMode`.
Would not be the case to set the read-only only at institution level in the component [OfferingEdit](https://github.com/bcgov/SIMS/blob/e5b7ea01bf15f36271a9dae005a5e93b8998da7b/sources/packages/web/src/views/institution/locations/offerings/OfferingEdit.vue).
Maybe the below area would be a good place to start.
https://github.com/bcgov/SIMS/blob/e5b7ea01bf15f36271a9dae005a5e93b8998da7b/sources/packages/web/src/views/institution/locations/offerings/OfferingEdit.vue#L174-L179

sources/packages/web/src/components/common/OfferingFormSubmit.vue
```diff
@@ -4,6 +4,7 @@
     :formMode="formMode"
     :processing="processing"
     :data="data"
+    :locationId="$props.locationId"
```
andrewsignori-aot, line 7: This may not be needed if this comment is implemented.

sources/packages/web/src/components/common/OfferingForm.vue
andrewsignori-aot: Here is a sample to be tested to use the existing read-only mode for the offering form to be done in the  [OfferingEdit](https://github.com/bcgov/SIMS/blob/e5b7ea01bf15f36271a9dae005a5e93b8998da7b/sources/packages/web/src/views/institution/locations/offerings/OfferingEdit.vue).
The intention is to demonstrate what was stated in the previous comment. If accepted the changes in this file can be reverted.
I am just starting to check the overall changes.
With the changes below, the changes in the form
```ts
const loadFormData = async () => {
    const programOffering =
      await EducationProgramOfferingService.shared.getOfferingDetailsByLocationAndProgram(
        props.locationId,
        props.programId,
        props.offeringId,
      );
    initialData.value = programOffering as OfferingFormModel;
    if (isReadOnlyUser(props.locationId)) {
      // If user is readonly, set form mode to readonly.
      formMode.value = OfferingFormModes.Readonly;
      return;
    }
    // Check status to define if the form should be readonly.
    let mode = OfferingFormModes.Readonly;
    const isReadonly = hasReadonlyStatus(programOffering.offeringStatus);
    if (!isReadonly) {
      mode = programOffering.hasExistingApplication
        ? OfferingFormModes.AssessmentDataReadonly
        : OfferingFormModes.Editable;
    }
    formMode.value = mode;
  };
```

sources/packages/web/src/components/institutions/modals/EditInstitutionUserModal.vue
andrewsignori-aot, line 1: Please review the below validations. The user should be considered as having access to a location if at least one location has permission different than "No access".

image

sources/packages/web/src/components/common/ProgramOfferingDetailHeader.vue
```diff
@@ -12,8 +12,11 @@
       />
       <v-divider-vertical-opaque class="mx-2 my-0" />
       <header-title-value title="Institution name"
-        ><template #value
-          ><span class="link-primary" @click="goToInstitutionProfile()">
+        ><template #value>
+          <span v-if="isReadOnlyUser($props.locationId)">
+            {{ headerDetails.institutionName }}
+          </span>
+          <span v-else class="link-primary" @click="goToInstitutionProfile()">
```
andrewsignori-aot, line 19: Executing the check based on the `isReadOnlyUser` would not be accurate.
If a permission needed to be checked for this it must be based on the user being an admin or not, but it would be outside the scope of this PR.

sources/packages/web/src/views/institution/locations/active-applications/ViewSubmittedApplicationScholasticStanding.vue
```diff
@@ -7,7 +7,11 @@
         :routeLocation="goBackRouteParams"
       />
     </template>
-    <scholastic-standing-form :initialData="initialData" :readOnly="true" />
+    <scholastic-standing-form
+      :initialData="initialData"
+      :readOnly="true"
+      :locationId="$props.locationId"
```
andrewsignori-aot, line 13: If there is already a `readOnly` property(line 12), can you please elaborate on why any additional verification was needed that also demanded providing the `locationId`?

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -39,20 +40,32 @@ export function useInstitutionAuth(rootStore?: Store<any>) {
         auth.locationId === locationId,
     );
 
-  // User type Admin | User.
-  const userType = computed(() => {
+  // Get user type User | Read-only-user.
+  const userType = (locationId: number) => {
```
andrewsignori-aot, line 44: Can you please elaborate on why this change was required?

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
```diff
@@ -39,20 +40,32 @@ export function useInstitutionAuth(rootStore?: Store<any>) {
...
   );
 
+  const isReadOnlyUser = (locationId?: number) => {
+    if (!locationId) return true;
+    return store.state.institution.authorizationsState?.authorizations.some(
+      (auth: InstitutionUserAuthRolesAndLocation) =>
+        auth.locationId?.toString() === locationId.toString() &&
+        auth.userType === InstitutionUserTypes.readOnlyUser,
+    );
+  };
```
andrewsignori-aot, line 67: I have a couple of questions about this, maybe I am missing something.
- Is the location ID an optional parameter to allow an admin?
- I did not follow why we need `toString` to execute the comparison at line 64.

sources/packages/web/src/components/common/ScholasticStandingForm.vue
```diff
@@ -47,8 +49,13 @@ export default defineComponent({
       required: true,
       default: false,
     },
+    locationId: {
+      type: Number,
+      required: false,
+    },
   },
   setup(props, context) {
+    const { isReadOnlyUser } = useInstitutionAuth();
```
andrewsignori-aot, line 58: This component has the `readOnly` property and it is set in some places to "true" already.
The check for the read-only-user can be added at the institution-only scope where it it set to `false`, for instance at (https://github.com/bcgov/SIMS/blob/411d242f76611b392bd14137fbd5bdf20bdfe9f8/sources/packages/web/src/views/institution/locations/active-applications/ActiveApplicationReportAChange.vue)). Does it make sense?

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
lewischen-aot: This change was required because one of the technical ACs is to "ensure routes are disallowed if the entire page should not be accessed." In this case, the program-creation and offering-creation pages are not accessible for read-only-users. The original method `userType` only checks whether the user is an admin or regular user, but the current criterion is location specific. That is, if an institution user is a "regular user" in one location but "read-only-user" in another location, this user can open and view the program-creation and offering-creation pages in one location but not the other one where he is a "read-only-user". This implementation addresses this technical AC.

sources/packages/web/src/components/institutions/modals/EditInstitutionUserModal.vue
```diff
@@ -115,9 +115,15 @@ export default defineComponent({
...
-        : LocationUserAccess.NoAccess;
+      if (locationAuthorization?.authType.type === LocationUserAccess.User) {
+        return LocationUserAccess.User;
+      } else if (
+        locationAuthorization?.authType.type === LocationUserAccess.ReadOnlyUser
+      ) {
+        return LocationUserAccess.ReadOnlyUser;
+      } else {
+        return LocationUserAccess.NoAccess;
+      }
```
andrewsignori-aot, line 126: Can you please change it to an "early return" or a switch?
```ts
if (locationAuthorization?.authType.type === LocationUserAccess.User) {
  return LocationUserAccess.User;
}
if (
  locationAuthorization?.authType.type === LocationUserAccess.ReadOnlyUser
) {
  return LocationUserAccess.ReadOnlyUser;
}
return LocationUserAccess.NoAccess;
```
or
```ts
switch (locationAuthorization?.authType.type) {
  case LocationUserAccess.User:
    return LocationUserAccess.User;
  case LocationUserAccess.ReadOnlyUser:
    return LocationUserAccess.ReadOnlyUser;
  default:
    return LocationUserAccess.NoAccess;
}
```

sources/packages/web/src/components/common/ScholasticStandingForm.vue
lewischen-aot: I made the change on the scholastic standing form on the wrong file. This is the correct file to make the change.

sources/packages/web/src/components/institutions/modals/EditInstitutionUserModal.vue
lewischen-aot: The switch looks better.

sources/packages/web/src/views/institution/locations/active-applications/ViewSubmittedApplicationScholasticStanding.vue
lewischen-aot: This is a mistake. I'll change.

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
andrewsignori-aot: Thanks for the explanation, looking into the `InstitutionRouteHelper` later I was able to follow. What caught my attention was that it was a computed property and maybe had some other codes relying on being a property.
So, the only place using it was the `InstitutionRouteHelper`, right?

lewischen-aot: That is correct. The only place using this method `userType` is the `InstitutionRouteHelper` file.

sources/packages/web/src/views/institution/locations/programs/LocationProgramAddEdit.vue
```diff
@@ -74,6 +75,7 @@ export default defineComponent({
     },
   },
   setup(props) {
+    const { isReadOnlyUser } = useInstitutionAuth();
```
andrewsignori-aot, line 78: As per the conversation this morning in the daily (thanks for bringing that up) we need to execute a similar logic from the API to check if the user has access to program creation.
Adding the API code here for reference only.
https://github.com/bcgov/SIMS/blob/1a3828766a861245185ecbd16ddcfe24acca2713/sources/packages/backend/apps/api/src/services/institution-user-auth/institution-user-auth.models.ts#L120-L128

lewischen-aot: On it. The next PR will cover all the backend changes, including APIs and E2E tests.

sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/ApplicationDetailsForCOE.vue
```diff
@@ -30,7 +30,11 @@
               color="primary"
             >
               <template v-for="(item, index) in items" :key="index">
-                <v-list-item :value="index" @click="item.command">
+                <v-list-item
+                  :value="index"
+                  @click="item.command"
+                  :disabled="isReadOnlyUser($props.locationId)"
+                >
```
andrewsignori-aot, line 37: The menu is conditionally shown under the below condition, maybe it can be converted to a computed that would also include the read-only-uer check. Does it make sense? Not a blocker.
```
v-if="initialData.applicationCOEStatus === COEStatus.required"
```

sources/packages/web/src/composables/institution/useInstitutionAuth.ts
lewischen-aot: - Yes, it is to allow an admin, but now `locationId` is not optional anymore in the lastest commit.
- Using `auth.locationId === locationId` didn't give the expected result, probably because `locationId` was passed in as a string from some components. I've changed it to `auth.locationId === +locationId`, and it works.

sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/ApplicationDetailsForCOE.vue
lewischen-aot: Updated.

sources/packages/web/src/components/generic/FooterButtons.vue
andrewsignori-aot, line 1: What would be the benefit of having a property to control the entire component visibility? Can we just add the `v-if` directly in the consumer side as shown below?
```vue
<footer-buttons v-if="!isReadOnlyUser(locationId)"
  :processing="processing"
  primaryLabel="Complete program info request"
  @primaryClick="submit"
  @secondaryClick="goBack"
/>
```

lewischen-aot: Nice one.

sources/packages/web/src/views/institution/locations/active-applications/ActiveApplicationReportAChange.vue
```diff
@@ -9,7 +9,7 @@
     </template>
     <scholastic-standing-form
       :initialData="initialData"
-      :readOnly="false"
+      :readOnly="isReadOnlyUser(locationId)"
```
andrewsignori-aot, line 12: This will possibly return `true` but the previous version had it set to always `false`.
Was there a need to change this particular line?

sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/ApplicationDetailsForCOE.vue
```diff
@@ -253,6 +259,7 @@ export default defineComponent({
     });
 
     return {
+      isReadOnlyUser,
```
andrewsignori-aot, line 262: Thanks for making this change. I belive the `isReadOnlyUser` is no longer needed in the return and can be removed.

sources/packages/web/src/components/common/OfferingForm.vue
andrewsignori-aot: Please review the provided example. The below line should be moved also, otherwise the form will not load its data. Does it make sense?
```ts
initialData.value = programOffering as OfferingFormModel;
```

sources/packages/web/src/views/institution/locations/program-info-request/LocationEditProgramInfoRequest.vue
```diff
@@ -21,6 +21,7 @@
           primaryLabel="Complete program info request"
           @primaryClick="submit"
           @secondaryClick="goBack"
+          :showButtons="!isReadOnlyUser(locationId)"
```
andrewsignori-aot, line 24: Here the `@secondaryClick="goBack"` can still be available to allow the navigation back.
The `primaryClick` can be disabled using the property `disablePrimaryButton`.

sources/packages/web/src/views/institution/locations/programs/LocationProgramAddEdit.vue
andrewsignori-aot: The point of the example was to provide the logic that UI should follow related to program-related pages. To be clear, the comment is not related to the API.

The program-related check cannot rely on the method `isReadOnlyUser` in the same way other location-specific ones.
A new helper method can be create, following the same idea of what was done in the API (that is why the example was shared).

Please let me know if it needs further clarification.

sources/packages/web/src/views/institution/locations/programs/LocationProgramView.vue
```diff
@@ -17,6 +17,7 @@
           ...educationProgram,
           status: educationProgram.programStatus,
         }"
+        :locationId="$props.locationId"
```
andrewsignori-aot, line 20: Is this one still required?

sources/packages/web/src/views/institution/locations/program-info-request/LocationEditProgramInfoRequest.vue
lewischen-aot: The UI will be inconsistent where some page hides the buttons completely and some page shows disabled buttons. The original implementation was to disable the buttons, not to hide them, but after we talked about hiding the buttons, hiding buttons and disabling buttons will be consistent.

sources/packages/web/src/views/institution/locations/active-applications/ActiveApplicationReportAChange.vue
lewischen-aot: The purpose adding changing this part is to hide the footer buttons component in <scholastic-standing-form/>, which displays the footer buttons component using the property `readOnly` in the condition `v-if="!readOnly"`. That's why I modify this component to pass in the property `readOnly` based on `isReadOnlyUser()`. I could modify `ScholasticStandingForm`.

sources/packages/web/src/views/institution/locations/request-a-change/RequestAChangeFormSubmit.vue
```diff
@@ -67,6 +67,7 @@
         primaryLabel="Submit requested change"
         @primaryClick="submit"
         @secondaryClick="cancel"
+        :showButtons="!isReadOnlyUser(locationId)"
```
andrewsignori-aot, line 70: Similar to this one: https://github.com/bcgov/SIMS/pull/4325#discussion_r1947329388

andrewsignori-aot (commented):

sources/packages/web/src/views/institution/locations/active-applications/ActiveApplicationReportAChange.vue
andrewsignori-aot: Hiding vs disabling conversation is ok for buttons that have some action that should be prevented, the navigation button would not be part of it, IMO.
For the existing pages hiding buttons, were the "cancelations/navigations" buttons removed? If yes, maybe those pages must be reviewed in the future.

sources/packages/web/src/views/institution/locations/program-info-request/LocationEditProgramInfoRequest.vue
andrewsignori-aot: https://github.com/bcgov/SIMS/pull/4325#discussion_r1947336959

sources/packages/web/src/views/institution/locations/programs/LocationProgramAddEdit.vue
andrewsignori-aot: Even the programs being created at the institution level the user accesses them through their specific locations, allowing the validation to happen based on the location ID. Based on that I am resolving this comment.

sources/packages/web/src/components/common/ScholasticStandingForm.vue
```diff
@@ -73,7 +76,7 @@ export default defineComponent({
       context.emit("cancel");
     };
 
-    return { data, submitted, cancel };
+    return { data, submitted, cancel, isReadOnlyUser };
```
andrewsignori-aot, line 79: `isReadOnlyUser` can be removed now, right?

andrewsignori-aot (approved):
Thanks for doing the changes, just one minor comment left, hence approving it.

PR #4335, opened 2025-02-10
#276 - Update Prod memory for Database

bidyashish:
**Memory Update**

Update Memory to 6Gi from 2Gi only for Prod.

### DIFF
--- a/devops/helm/crunchy-postgres/values-0c27fb-prod.yaml
+++ b/devops/helm/crunchy-postgres/values-0c27fb-prod.yaml
@@ -29,10 +29,10 @@ instances:
     storageClassName: netapp-block-standard
   requests:
     cpu: "2"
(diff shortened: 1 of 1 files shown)


devops/helm/crunchy-postgres/values-0c27fb-prod.yaml
```diff
@@ -29,10 +29,10 @@ instances:
     storageClassName: netapp-block-standard
   requests:
     cpu: "2"
-    memory: 2Gi
+    memory: 6Gi
```
dheepak-aot, line 32: We also have one more prod yml on different license plate. Even though it is not used, it it better to have both prod yml in sync right?

dheepak-aot (approved):
Looks good 👍

Issue #3930, opened 2024-11-13
Student: Self Report Parent Income 


JerPearson:
As a student I want to be asked all relevant questions to accurately report my parents financial situation.

**Acceptance Criteria**
- [ ] Revise question: "How many parents do you live with?" to : TBD.  This question is not correct and need options
- [ ] Given (Context), When (action carried out), Then (expected outcome)

guru-aot:
As part of the implementation, for the new fields key added the prefix studentParents and used the existing keys of the stuedent application for the description change.
No workflow update to update the consolidatedData is taken care as part of this ticket.
@CarlyCotton FYI

ninosamson:
Moving to 2.6 as 2.5 branch has been created.

PR #4345, opened 2025-02-11
#3930 - FT Self Report Parent Income Questions

guru-aot:
Update Parents Income in 2025-2026 Full Time application.
image


sources/packages/forms/src/form-definitions/sfaa2025-26.json
```diff
@@ -19383,7 +19383,89 @@
...
+                    "strictDateValidation": false,
+                    "multiple": false,
+                    "unique": false,
+                    "min": "",
+                    "max": "",
+                    "step": "any",
+                    "integer": ""
+                  },
+                  "key": "parentalContributions",
+                  "type": "currency",
```
lewischen-aot, line 19407: I think this key should be "studentParentNetContribution" to match the key `studentDataStudentParentsTotalIncome` in the `AssessmentConsolidatedData` object.

sources/packages/forms/src/form-definitions/sfaa2025-26.json
```diff
@@ -19403,7 +19485,7 @@
                     "step": "any",
                     "integer": ""
                   },
-                  "key": "studentParentsTotalIncome",
+                  "key": "totalIncome",
```
lewischen-aot, line 19488: I think this key should be reverted to "studentParentsTotalIncome" to match the key `studentDataStudentParentsTotalIncome` in the `AssessmentConsolidatedData` object.

sources/packages/forms/src/form-definitions/sfaa2025-26.json
```diff
@@ -19486,7 +19568,7 @@
                     "step": "any",
                     "integer": ""
                   },
-                  "key": "studentParentNetAssests",
+                  "key": "foreignAssets",
```
lewischen-aot, line 19571: I think this key should be also reverted to "studentParentNetAssests" to match the key `studentDataStudentParentNetAssests` in the `AssessmentConsolidatedData` object.

lewischen-aot:
Nice work adding the new questions and fields and re-ordering the questions. I left a few comments for consideration.

sources/packages/forms/src/form-definitions/sfaa2025-26.json
andrewsignori-aot: @guru-aot it does make sense to keep the existing key names and the new ones be added following the supporting users naming convention but respecting the current prefix in the student application.
A minor difference that I noticed is that some prefixes have the parent in the plural and some singular.
I would recommend having them all prefixed as `studentParents` + `suporting user variable name` and the `AssessmentConsolidatedData` adjusted. Thoughts?

andrewsignori-aot (commented):
Nice work, overall looks good but please take a loser look at the below comment.
https://github.com/bcgov/SIMS/pull/4345#discussion_r1953166628

sources/packages/forms/src/form-definitions/sfaa2025-26.json
guru-aot: Make sense, I have updated the new fields key with the prefix `studentParents` and used the existing keys for the description change.

sources/packages/forms/src/form-definitions/sfaa2025-26.json
guru-aot: Yes Lewis, using the existing name solves the issue.

sources/packages/forms/src/form-definitions/sfaa2025-26.json
guru-aot: https://github.com/bcgov/SIMS/pull/4345#discussion_r1953316236

lewischen-aot (approved):
LGTM 👍 Thanks for making the changes and adding the "student" prefix.

andrewsignori-aot (approved):
Look good 👍

PR #4346, opened 2025-02-12
#4014 - Institution Read-Only User Type PART 2 (UI) (Part 3)

lewischen-aot:
This PR removes the two buttons related to program and offering executions for read-only users:
- Removed button "+ Create program" on the PIR page.
- Removed dropdown button "Edit offering" on the offering edit page.

Screenshots
image

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/programinformationrequest.json
+++ b/sources/packages/forms/src/form-definitions/programinformationrequest.json
@@ -2358,7 +2358,7 @@
             "eq": "",
             "json": ""
           },
-          "customConditional": "show = data['isReadOnlyUser'] !== true;",
+          "customConditional": "show = data.isReadOnlyUser !== true;",
           "logic": [],
           "attributes": {},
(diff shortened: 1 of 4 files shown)


sources/packages/forms/src/form-definitions/programinformationrequest.json
```diff
@@ -2353,12 +2353,12 @@
...
           "properties": {},
           "conditional": {
-            "show": null,
+            "show": "",
             "when": null,
             "eq": "",
             "json": ""
           },
-          "customConditional": "",
+          "customConditional": "show = data['isReadOnlyUser'] !== true;",
```
andrewsignori-aot, line 2361: There is no need to string property access, it can be `show = data.isReadOnlyUser !== true`

andrewsignori-aot (approved):
Thanks for the quick changes, looks good 👍

guru-aot (commented):
LGTM

Issue #4235, opened 2025-01-14
Bug FORM IO (2.3) Fixes errors with offering submission missing information validations


astridSABC:
**Describe the Bug**
Red warning fix the error or missing information messages displayed when they should not be. This includes when files are uploaded using the manual uploader (student), no study breaks (add offering), and change section with incomplete info (student)

- [x] Request a Change
- [ ] Student File Uploader
- [x] Application
- [x] Institution Add Offering

### **Expected Behaviour**
If valid information is included for an uploaded form or page, no red "Please fix the following errors before submitting" with links and highlights should be displayed

**Change Request**
Validation errors not being removed when submitting change request for an application.

**Student File Uploader** 
File is uploaded with no red "Please fix the following errors" and the "Document Type" area is not highlighted in red

**Student Family information tab: Yes/No to dependants**
Change section with incomplete info is only for the page that needs to be updated

**Institution Add offering**
Toast message: "Offering was validated successfully and no warnings were displayed" is displayed. Institution user click "Add offering now"

### **Actual Behaviour**

**Student File Uploader**
Warning are displayed even though the file is attached

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeGh0Q0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--df13993b2c70092292a7e1188b89bd7eb45a54c7/image.png" alt="image.png" />

**Student Dependants Yes/No**
Warning are also displayed on the financial information and confirm submission tabs.  User must refresh the page for the warnings to be removed.

Family information tab
<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVZ0Q0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3ce25c2214af583833ec411d94f91e8c0a397165/image.png" alt="image.png" />

Financial information tab
<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVp0Q0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3a30420d2b9795c81306f715c9f9c13b87e5d988/image.png" alt="image.png" />

Confirm submission tab
[...]

PR #4349, opened 2025-02-12
#4235 - Student File Upload - Reset after submit

dheepak-aot:
## Student File Upload - Reset after submit

- [x] Updated the formio object typed with `resetValue` from the formio SDK library. `resetValue()` is expected to do more than just resetting the values like setting the form to pristine .etc. 
- [x] Using reset and redraw to mitigate the error shown in student uploader after submit.

image

image


sh16011993 (approved):
LGTM @dheepak-aot 👍

andrewsignori-aot (approved):
👍

Issue #4353, opened 2025-02-13
SIMS Queue Dashboard - Access through IDIR


CarlyCotton:
**User Story**
As a ministry user, I want to be able to access the queues to troubleshoot the processes for SIMS. This should not require me to have OpenShift access and there would be an admin/queues access through the ministry portal. It will be an additional role through keycloak that can be assigned.

**Acceptance Criteria**
- [ ] Provide access to bull dashboard through a sims ministry portal for each environment
- [ ] Enable a keycloak role to control access to the bull dashboard

PR #4363, opened 2025-02-17
#4353 - SIMS Queue Dashboard - Access through IDIR

andrewsignori-aot:
Used a "JWT token exchange" concept to allow access to the "Queues Admin", considering an IDIR(Ministry) user already authenticated and with a specific role. 

> A JWT (JSON Web Token) exchange is a process where an existing JWT is used to obtain a new JWT, often with different claims or for accessing different resources.

_Note:_ Keycloak has a [token exchange preview feature](https://www.keycloak.org/securing-apps/token-exchange) that seems what was needed but is not enabled right now. Also, it would be possible to use our Keycloak js lib and authenticate a second token, but it does not seem to be recommendable and the current approach of SIMS-API issuing a token seems good enough for now.

The SIMS-API will allow access to a specific endpoint `users/queue-admin-token-exchange`, protected under a Ministry role (`aest-queue-dashboard-admin`), to generate a cookie with a SIMS-API issued token where its sole purpose is to allow access to the "Queues Admin". Queues admin is now able to validate the generated token based on a shared secret between the SIMS-API and queue-consumers that allows the token to be signed and validated.

This token is saved in a cookie that will also be accessible by the "Queues Admin", allowing its validation. The token has the security properties sets (`httpOnly`, `secure`, and `sameSite`) to prevent different security vulnerabilities. Some of those must be relaxed for local development only.
_Note:_ `credentials: true` was added under the `app.enableCors` to allow the generated cookie to be saved on the client, otherwise `Axios` would not save it.
Please see below an example of what the generated token looks like, the known [registered claims](https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-token-claims#registered-claims) were used.
```json
{
  "iss": "sims-api",
  "sub": "some-user-guid@idir",
  "aud": "queues-dashboard",
  "iat": 1739839795,
  "exp": 1739839915
}
```

The new way of accessing the "Queues Admin" is using the new link in the Ministry portal. Once accessed, it will generate the token to allow the user to access the queues dashboard.
The new Ministry role (`aest-queue-dashboard-admin`) is intended to show/hide the button to avoid confusion to other users since this is supposed to be accessible to only a few users.

image

## Possible next steps
- Share some API minor utils for the access log.
- Add a GUID to the token to allow a "session" control using a shared resource (like Redis), to invalidate the cookie after a few minutes of inactivity.

## Notes on secret generation

The secret key to be saved to the GitHub should be generated using some strong random generator, for instance, executing the `openssl rand -base64 32`. The value should be sent to the PODs as base64 which means it should be added to GitHub secrets as a base64 string.


configs/env-example
```diff
@@ -136,9 +136,11 @@ REDIS_PORT=6379
 REDIS_PASSWORD=
 REDIS_STANDALONE_MODE=true
 QUEUE_CONSUMERS_PORT=
-QUEUE_DASHBOARD_PASSWORD=
-QUEUE_DASHBOARD_USER=
 QUEUE_PREFIX={sims-local}
+# The below secret should be set with any valid base 64 string for local development.
+QUEUE_DASHBOARD_TOKEN_SECRET=
+QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS=3600
```
dheepak-aot, line 142: Do we need these props in env-example if local ignores token?

devops/Makefile
```diff
@@ -192,6 +192,7 @@ init-secrets:
 		-p ZEEBE_CLIENT_SECRET=$(ZEEBE_CLIENT_SECRET) \
 		-p CAMUNDA_OAUTH_URL=$(CAMUNDA_OAUTH_URL) \
 		-p ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS=$(ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS) \
+		-p QUEUE_DASHBOARD_TOKEN_SECRET=$(QUEUE_DASHBOARD_TOKEN_SECRET) \
```
dheepak-aot, line 195: Init secrets is missing `${QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS}`

.github/workflows/env-setup-deploy-secrets.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       ZEEBE_CLIENT_SECRET: ${{ secrets.ZEEBE_CLIENT_SECRET }}
       CAMUNDA_OAUTH_URL: ${{ secrets.CAMUNDA_OAUTH_URL }}
       ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS: ${{ secrets.ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS }}
+      QUEUE_DASHBOARD_TOKEN_SECRET: ${{ secrets.QUEUE_DASHBOARD_TOKEN_SECRET }}
```
dheepak-aot, line 50: Following env is mising here. 

`QUEUE_DASHBOARD_BASE_URL`

`QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS_KEY`

sources/docker-compose.yml
```diff
@@ -27,6 +27,9 @@ services:
       - REDIS_PASSWORD=${REDIS_PASSWORD}
       - REDIS_STANDALONE_MODE=${REDIS_STANDALONE_MODE}
       - QUEUE_PREFIX=${QUEUE_PREFIX}
+      - QUEUE_DASHBOARD_TOKEN_SECRET=${QUEUE_DASHBOARD_TOKEN_SECRET}
+      - QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS=${QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS}
```
dheepak-aot, line 31: Similar to https://github.com/bcgov/SIMS/pull/4363/files#r1962301922

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -55,6 +55,7 @@ async function bootstrap() {
   app.enableCors({
     exposedHeaders: "Content-Disposition",
     origin: allowAnyOrigin,
+    credentials: true,
```
dheepak-aot, line 58: Asking for my understanding. Why do we need to enable this cors header?

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
```diff
@@ -90,3 +90,8 @@ export const BLOCKED_DISBURSEMENT_NOTIFICATION_MIN_DAYS_INTERVAL = 7;
  * Represents the maximum number of disbursement blocked notifications to send.
  * */
 export const BLOCKED_DISBURSEMENT_MAXIMUM_NOTIFICATIONS_TO_SEND = 3;
+
+/**
+ * Bull Dashboard route.
+ */
+export const BULL_BOARD_ROUTE = "admin/queues";
```
dheepak-aot, line 97: the route needs in `/` in the end on open shift environments. 

image

sources/packages/backend/apps/api/src/route-controllers/config/config.controller.ts
```diff
@@ -38,6 +39,7 @@ export class ConfigController extends BaseController {
       maximumIdleTimeForWarningAEST:
         this.configService.maximumIdleTimeForWarningAEST,
       appEnv: this.configService.appEnv,
+      queueDashboardURL: `${process.env.QUEUE_DASHBOARD_BASE_URL}/${BULL_BOARD_ROUTE}`,
```
dheepak-aot, line 42: For the base URL, can we use the current context URL similar to the way how we get API URL in web?

image

devops/Makefile
andrewsignori-aot: It is not missing, it is just configured as a default value. I was not sure about exposing it or not.
I can add to the GitHub actions.

configs/env-example
andrewsignori-aot: They are currently used for local development. We can decide on removing it or the developer can just add a big number to the session that would last for the day.
Making it work locally simular to the deployed version has the benefit of allowing it to be tested in the same way locally.
I would be open either way leaning a bit more for keeping the local behavior close to the prod version.

devops/openshift/init-secrets.yml
```diff
@@ -198,17 +199,17 @@ parameters:
...
-    generate: expression
-    from: "[a-zA-Z0-9]{32}"
+  - name: QUEUE_DASHBOARD_TOKEN_SECRET
+    description: Bull board token secret for token generation and validation.
+    displayName: Queue dashboard token secret
+    required: true
+  - name: QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS
+    description: Bull board token expiration(seconds).
+    displayName: Queue dashboard token expiration seconds.
+    value: 3600
```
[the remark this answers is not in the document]
andrewsignori-aot: I was not willing to have it configured to avoid this to be configured way beyond 1h. I would try to have it configured as low as possible for security but I can assume that someone with access wold be responsible enough to think about it also.
I will expose it to GitHub actions.

sources/packages/backend/libs/services/src/constants/system-configurations-constants.ts
andrewsignori-aot: This configuration is not new and was just moved. I believe that if it was wrong it was already causing some issues. Does it make sense?

sources/packages/backend/apps/api/src/route-controllers/config/config.controller.ts
andrewsignori-aot: The above-mentioned one has two hardcoded values (including the port) while the one in the PR is fully configurable. I would change the one in the example to follow the same idea from the one in this PR 😄 
image

sources/packages/backend/apps/queue-consumers/src/bull-board/bull-board-authentication.middleware.ts
```diff
@@ -0,0 +1,125 @@
...
+      } else if (error instanceof JsonWebTokenError) {
+        forbiddenError = "invalid token";
+      } else {
+        forbiddenError = "authentication failed";
+        // Only add an error log if the error in unknown.
+        this.logger.error("Queues dashboard authentication error.", error);
+      }
+      response
+        .status(HttpStatusCode.Forbidden)
+        .send(`Not possible to authenticate: ${forbiddenError}.`);
```
dheepak-aot, line 80: Switch case ?

andrewsignori-aot: Is there a benefit to using it when testing `error instanceof TokenExpiredError`?

sources/tests/docker-compose.yml
```diff
@@ -25,6 +25,9 @@ services:
       - REDIS_PASSWORD=${REDIS_PASSWORD}
       - REDIS_STANDALONE_MODE=${REDIS_STANDALONE_MODE}
       - QUEUE_PREFIX=${QUEUE_PREFIX}
+      - QUEUE_DASHBOARD_TOKEN_SECRET=${QUEUE_DASHBOARD_TOKEN_SECRET}
+      - QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS=${QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS}
```
dheepak-aot, line 29: Same comment about local not requiring to authnticate.

sources/packages/backend/apps/api/src/main.ts
andrewsignori-aot: This will allow the client (Axios) to process the 'set-cookie' header and persist the cookie to the browser.

sources/tests/docker-compose.yml
andrewsignori-aot: Same answer 😄 
Is there a problem with having the local authentication enabled?
The basic auth was also enabled in the local.

devops/Makefile
andrewsignori-aot: `QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS` was changed to be exposed in GitHub vars and, if not provided, will be set at 3600 by the `make`.

.github/workflows/env-setup-deploy-secrets.yml
andrewsignori-aot: `QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS_KEY` is now configurable.
`QUEUE_DASHBOARD_BASE_URL` is populated by the `make` no need to set it.

devops/openshift/init-secrets.yml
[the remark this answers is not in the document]
andrewsignori-aot: `QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS` was changed to be exposed in GitHub vars and, if not provided, will be set at 3600 by the `make`.

sources/packages/backend/libs/auth/src/services/queues-dashboard/queue-dashboard.models.ts
```diff
@@ -0,0 +1,25 @@
...
+   * Subject of the JWT (the Ministry user guid).
+   */
+  sub: string;
+  /**
+   * Recipient for which the JWT is intended
+   */
+  aud: string;
+  /**
+   * Time after which the JWT expires
+   */
```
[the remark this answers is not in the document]
andrewsignori-aot: Added.

sources/packages/web/src/services/AuthService.ts
```diff
@@ -301,6 +306,10 @@ export class AuthService {
         break;
       }
       case ClientIdType.AEST: {
+        const { hasRole } = useAuth();
+        if (hasRole(Role.AESTQueueDashboardAdmin)) {
+          await UserService.shared.removeAdminTokenExchange();
+        }
```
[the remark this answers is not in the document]
andrewsignori-aot: Removing the cookie also "logout" the "queues admin". It allows the Ministry user to "invalidate" the "Queues Admin" login. Since it has an 1h expiration right now, even if the user logs off from the Ministry portal the "Queues Admin" would still be "accessible".

devops/Makefile
```diff
@@ -200,6 +203,7 @@ init-secrets:
 		-p S3_ENDPOINT=$(S3_ENDPOINT) \
 		-p S3_DEFAULT_BUCKET=$(S3_DEFAULT_BUCKET) \
 		-p S3_REGION=$(S3_REGION) \
+		-p QUEUE_DASHBOARD_BASE_URL=$(HOST) \
```
sh16011993, line 206: `HOST` is defined from `DOMAIN_PREFIX` and `DOMAIN`.

`DOMAIN` does not have a default value and is not passed from the `env-setup-deploy-secrets`. I believe it needs to be passed from there.

devops/openshift/init-secrets.yml
```diff
@@ -37,8 +37,9 @@ objects:
       zeebe-client-secret: ${ZEEBE_CLIENT_SECRET}
       camunda-oauth-url: ${CAMUNDA_OAUTH_URL}
       zeebe-grpc-worker-longpoll-seconds: ${ZEEBE_GRPC_WORKER_LONGPOLL_SECONDS}
-      queue-dashboard-user: ${QUEUE_DASHBOARD_USER}
-      queue-dashboard-password: ${QUEUE_DASHBOARD_PASSWORD}
+      queue-dashboard-token-secret: ${QUEUE_DASHBOARD_TOKEN_SECRET}
+      queue-dashboard-token-expiration-minutes: ${QUEUE_DASHBOARD_TOKEN_EXPIRATION_SECONDS}
```
sh16011993, line 41: LHS has minutes and RHS has seconds. Is it okay?

devops/Makefile
dheepak-aot, line 206: Adding comment here for record. As we discussed, need to validate if we have to prefix the `https://`before the HOST to make it work.

andrewsignori-aot: The `HOST` variable is the same one used to define the route for the PODs, hence it is already per environment.

devops/openshift/init-secrets.yml
andrewsignori-aot: Fixed.

devops/Makefile
andrewsignori-aot: Prefixed the host with "https" to ensure the URL will be absolute.

sh16011993: I didn't get the above comment. Will check with you tomorrow.

andrewsignori-aot: I got it now. I moved the `QUEUE_DASHBOARD_BASE_URL` to the API deploy now.

sh16011993 (approved):
Great work @andrewsignori-aot 👍  Thanks for clarifying my questions from the PR.

Issue #3249, opened 2024-05-09
Bridge Matching - Re-matching and persisting the match


HRAGANBC:
**Story Description**
As a ministry user, I want the system to recognize newly updated SFAS or SIMS info and re-engage the SFAS bridge file to look for matches and import data appropriately. This way, if I get a notification of a partial match and work with the student to align both profiles any data bridge information can then be matched to the newly made FULL match. We also want to make sure if the student data changes in SIMS, we continue to capture any PY totals on applications, study period overlap and restrictions.

**Acceptance Criteria:**
- [x] When a SIMS profile is created, continue to search for a perfect (DOB/Last name/SIN) match as normal. 
- [x] Do NOT update the student id match on an `sfas_individual` if one already exists
- [x] On any future legacy match search (msfaa/program year total/overlap/overawards/restrictions), find if the SIMS student has already had a match to a legacy profile and use that `sfas_individual.id`

- [x] Overaward imports can be handled after the linking process instead of at initial profile creation.

**Context:**
- Under the current build of the bridge file, SIMS will reference the SFAS data table _only upon SIMS account creation_, therefore if there is a lack of alignment between legacy and SIMS profile data that is resolved (e,g. by updating SFAS info) _after_ SIMS profile is created, _no_ link will be created and _no_ date will be imported. 

Solution could be to have a button available on a student profile which triggers another check with the SIMS DB SFAS_table, and allowing a new 'link' to be made. 

**Technical**
- [x] Update the SQL script [Bulk-update-students-foreign-key.sql](https://github.com/bcgov/SIMS/blob/32674e65bdb026d42967835c0aa20114ae17fb55/sources/packages/backend/libs/src-sql/sfas-individuals/Bulk-update-students-foreign-key.sql) to avoid updating the student ID on the `sfas_individuals` table if already populated.
- [x] Check the below areas (**and other possible features**) to ensure the student ID from `sfas_individuals` table will be used instead of the 3 matching fields.
  - Student Creation
    - PD/PDD (no changes required)
    - Overawards (check with biz if we can remove from student creation and leave it to the bridge only).
    - Provincial Restrictions (processed flag updated only, no changes required)
  - On-demand
    - Application Overlaps
    - Program year totals (workers)
-  [x] Check for any other area of the solution using the Last Name/DOB/SIN from `sims.sfas_individuals`.
-  [x] In case one particular area not listed above would demand more effort than what was planned, please explain and create a new ticket.
[...]

PR #4369, opened 2025-02-20
#3249 - Bridge Matching - Re-matching and persisting the match

lewischen-aot:
- Updated the SQL script [Bulk-update-students-foreign-key.sql](https://github.com/bcgov/SIMS/blob/32674e65bdb026d42967835c0aa20114ae17fb55/sources/packages/backend/libs/src-sql/sfas-individuals/Bulk-update-students-foreign-key.sql) to avoid updating the student ID on the sfas_individuals table if already populated.
- Checked the below areas (and other possible features) to ensure the student ID from sfas_individuals table will be used instead of the 3 matching fields.
  - Overawards
  - Application Overlaps
  - Program year totals (workers)
- Adapted existing E2E tests.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -316,10 +316,8 @@ export class ApplicationStudentsController extends BaseController {
     try {
       await this.applicationService.validateOverlappingDates(
         applicationId,
-        student.user.lastName,
         studentToken.userId,
-        student.sinValidation.sin,
-        student.birthDate,
+        student.id,
         referenceStudyStartDate,
         referencedStudyEndDate,
       );
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
@@ -300,9 +300,6 @@ export class ApplicationStudentsController extends BaseController {
       submissionResult.data.data.howWillYouBeAttendingTheProgram,
     );
 
-    const student = await this.studentService.getStudentById(
-      studentToken.studentId,
-    );
     // If offering is present, the selected offering's start and end dates will be used.
     let referenceStudyStartDate =
       submissionResult.data.data.selectedOfferingDate;
@@ -317,14 +314,14 @@ export class ApplicationStudentsController extends BaseController {
       await this.applicationService.validateOverlappingDates(
         applicationId,
         studentToken.userId,
-        student.id,
+        studentToken.studentId,
         referenceStudyStartDate,
         referencedStudyEndDate,
       );
       await this.applicationService.submitApplication(
         applicationId,
         studentToken.userId,
-        student.id,
+        studentToken.studentId,
         programYear.id,
         submissionResult.data.data,
         payload.associatedFiles,
--- a/sources/packages/backend/apps/api/src/services/student/student.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student/student.service.ts
@@ -749,9 +749,7 @@ export class StudentService extends RecordDataModelService<Student> {
   /**
    * Search for the student's SFAS data and update overaward values in
    * disbursement overawards table if any overaward values for BCSL or CSLF exist.
-   * @param studentId student id for the disbursement overaward record.
-   * @param lastName last name for sfas individuals.
-   * @param sinNumber student's sin number.
+   * @param studentId student id for finding a SFAS individual.
    * @param auditUserId user that should be considered the one that is causing the changes.
    * @param entityManager entityManager to be used to perform the query.
    */
--- a/sources/packages/backend/apps/api/src/services/student/student.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student/student.service.ts
@@ -10,8 +10,6 @@ import {
   SINValidation,
   StudentAccountApplication,
   StudentUser,
-  DisbursementOveraward,
-  DisbursementOverawardOriginType,
   RestrictionNotificationType,
   StudentRestriction,
   DisabilityStatus,
@@ -43,10 +41,6 @@ import {
   NotificationActionsService,
   SystemUsersService,
 } from "@sims/services";
-import {
-  BC_STUDENT_LOAN_AWARD_CODE,
-  CANADA_STUDENT_LOAN_FULL_TIME_AWARD_CODE,
-} from "@sims/services/constants";
 
 @Injectable()
 export class StudentService extends RecordDataModelService<Student> {
@@ -231,11 +225,6 @@ export class StudentService extends RecordDataModelService<Student> {
       }
 
       if (sfasIndividual) {
-        await this.importSFASOverawards(
-          sfasIndividual.student.id,
-          auditUserId,
-          entityManager,
-        );
         // For the newly created student, update the Processed status in the
         // SFAS Restrictions table to false. This will enable for the SFAS
         // Restrictions from previous imports that were originally marked as
@@ -745,49 +734,6 @@ export class StudentService extends RecordDataModelService<Student> {
 
   @InjectLogger()
   logger: LoggerService;
-
-  /**
-   * Search for the student's SFAS data and update overaward values in
-   * disbursement overawards table if any overaward values for BCSL or CSLF exist.
-   * @param studentId student id for finding a SFAS individual.
-   * @param auditUserId user that should be considered the one that is causing the changes.
-   * @param entityManager entityManager to be used to perform the query.
-   */
-  async importSFASOverawards(
-    studentId: number,
-    auditUserId: number,
-    entityManager: EntityManager,
-  ): Promise<void> {
-    const sfasIndividual = await this.sfasIndividualService.getSFASOverawards(
-      studentId,
-    );
-    if (!sfasIndividual) {
-      return;
-    }
-    const overawards: DisbursementOveraward[] = [];
-    if (sfasIndividual.bcslOveraward !== 0) {
-      overawards.push({
-        student: { id: studentId } as Student,
-        disbursementValueCode: BC_STUDENT_LOAN_AWARD_CODE,
-        overawardValue: sfasIndividual.bcslOveraward,
-        originType: DisbursementOverawardOriginType.LegacyOveraward,
-        creator: { id: auditUserId } as User,
-      } as DisbursementOveraward);
-    }
-    if (sfasIndividual.cslOveraward !== 0) {
-      overawards.push({
-        student: { id: studentId } as Student,
-        disbursementValueCode: CANADA_STUDENT_LOAN_FULL_TIME_AWARD_CODE,
-        overawardValue: sfasIndividual.cslOveraward,
-        originType: DisbursementOverawardOriginType.LegacyOveraward,
-        creator: { id: auditUserId },
-      } as DisbursementOveraward);
-    }
-    await this.disbursementOverawardService.addLegacyOverawards(
-      overawards,
-      entityManager,
-    );
-  }
 
   /**
    * Get student disability status.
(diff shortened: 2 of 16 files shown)


sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -316,10 +316,8 @@ export class ApplicationStudentsController extends BaseController {
     try {
       await this.applicationService.validateOverlappingDates(
         applicationId,
-        student.user.lastName,
         studentToken.userId,
-        student.sinValidation.sin,
-        student.birthDate,
+        student.id,
```
andrewsignori-aot, line 320: The student ID is present in the `studentToken` also, using it can eliminate the call to the method `getStudentById` at line 303.

sources/packages/backend/libs/services/src/sfas/sfas-application.service.ts
```diff
@@ -20,30 +20,23 @@ export class SFASApplicationService extends DataModelService<SFASApplication> {
...
+    studentId: number,
     studyStartDate: string,
     studyEndDate: string,
   ): Promise<SFASApplication> {
     return this.repo
       .createQueryBuilder("sfasApplication")
       .select(["sfasApplication.id"])
-      .innerJoin("sfasApplication.individual", "sfasFTstudent")
+      .innerJoin("sfasApplication.individual", "sfasIndividual")
+      .innerJoin("sfasIndividual.student", "student")
```
andrewsignori-aot, line 37: In some situations TypeOrm allow the id directly in the `where` without the need for a `innerJoin`, as shown below. Can you test it please?
```ts
return this.repo
      .createQueryBuilder("sfasApplication")
      .select(["sfasApplication.id"])
      .innerJoin("sfasApplication.individual", "sfasIndividual")
      .where("sfasApplication.applicationCancelDate IS NULL")
      .andWhere("sfasIndividual.student.id = :studentId", { studentId })
```

sources/packages/backend/libs/services/src/sfas/sfas-part-time-application.service.ts
```diff
@@ -20,30 +20,23 @@ export class SFASPartTimeApplicationsService extends DataModelService<SFASPartTi
...
-    lastName: string,
+    studentId: number,
     studyStartDate: string,
     studyEndDate: string,
   ): Promise<SFASPartTimeApplications> {
     return this.repo
       .createQueryBuilder("sfasPTApplication")
       .select(["sfasPTApplication.id"])
-      .innerJoin("sfasPTApplication.individual", "sfasPTstudent")
+      .innerJoin("sfasPTApplication.individual", "sfasIndividual")
```
andrewsignori-aot, line 36: Similar to this https://github.com/bcgov/SIMS/pull/4369/files#r1964408124

sources/packages/backend/libs/src-sql/sfas-individuals/Bulk-update-students-foreign-key.sql
andrewsignori-aot, line 1: Can you please update the comment in the file? It lost part of its meaning and can be misleading.
```ts
/*
 * After all the SFAS individuals are imported, this update
 * associates all the students ids that matches with the students
 * currently on the database.
 */

```

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -760,16 +757,11 @@ export class StudentService extends RecordDataModelService<Student> {
...
-    birthDate: string,
-    sinNumber: string,
     auditUserId: number,
     entityManager: EntityManager,
   ): Promise<void> {
     const sfasIndividual = await this.sfasIndividualService.getSFASOverawards(
-      lastName,
-      birthDate,
-      sinNumber,
+      studentId,
```
andrewsignori-aot, line 764: Please let me know if I am missing something but at this moment the `studentId` provided to `importSFASOverawards` will never be present because the student may not even be created yet.
I guess that is what triggered the below technical AC and we ended up never checking with the business. We need a follow-up on this.
```
Overawards (check with biz if we can remove from student creation and leave it to the bridge only).
```

andrewsignori-aot, line 764: If this method stays, please update the comment parameters.

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
```diff
@@ -94,28 +96,18 @@ export class SFASIndividualService {
...
-   * @param sin student's sin.
+   * @param studentId Student id.
    * @returns SFAS individual details.
    */
-  async getSFASOverawards(
-    lastName: string,
-    birthDate: string,
-    sin: string,
-  ): Promise<SFASIndividual> {
+  async getSFASOverawards(studentId: number): Promise<SFASIndividual> {
```
andrewsignori-aot, line 102: This method may removed depending on the answer that we get from the business.
Related comment: https://github.com/bcgov/SIMS/pull/4369#discussion_r1964429507

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/sfas-integration/_tests_/sfas-integration.scheduler.e2e-spec.ts
```diff
@@ -529,22 +529,30 @@ describe(describeProcessorRootTest(QueueNames.SFASIntegration), () => {
...
-      user.lastName = "INNAVOIG";
-      const student = await saveFakeStudent(
-        db.dataSource,
-        { user },
-        { initialValue: { birthDate: "1966-07-21" } },
-      );
-      const sinValidation = createFakeSINValidation(
-        {
-          student,
+      let student = await db.student.findOne({
```
andrewsignori-aot, line 532: I am not following why the student creation is no longer happening and there was it was chosen to find a student.
The test should create its own data and ensure that preexisting data will not conflict with it.

sources/packages/backend/libs/services/src/students-assessments/assessment-sequential-processing.service.ts
```diff
@@ -42,6 +43,8 @@ export class AssessmentSequentialProcessingService {
     private readonly sfasApplicationsRepo: Repository<SFASApplication>,
     @InjectRepository(SFASPartTimeApplications)
     private readonly sfasPartTimeApplicationsRepo: Repository<SFASPartTimeApplications>,
+    @InjectRepository(SFASIndividual)
+    private readonly sfasIndividualRepo: Repository<SFASIndividual>,
```
andrewsignori-aot, line 47: Please remove the not-used variable.

andrewsignori-aot (commented):
Nice work, please do the necessary follow-up with the business to have the ACs clarified/adjusted.
Please also check the below method that is still relying in the match instead of the student ID.
- [getSFASTotalUnsuccessfulCompletionWeeks](https://github.com/bcgov/SIMS/blob/83a79473d091ef06682243a8949d27fb5ce95124/sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts#L120)

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/sfas-integration/_tests_/sfas-integration.scheduler.e2e-spec.ts
lewischen-aot: Every time this test case is run, a student record with the last name, birth date and sin number is created. During local tests, this test case can create multiple student records with the same last name, birth date and sin number, but the SFAS individual mapping will only map to the first student record if it exists. Without this conditional checking, there would be some work to remove SFAS individual and student records every time this test case is run with the local test database. Unlike local database, the database in Github action is newly created every time, so creating a new student record is necessary if it is not yet created. This conditional statement makes sure this test case works both local and Github action database. It is similar to the `sharedStudent` variable declared in `beforeAll`, but it isn't really necessary because this `student` variable is used only in this test case.

sources/packages/backend/apps/api/src/services/student/student.service.ts
lewischen-aot: I think it would be fine to use student id here because it will not import over awards if the student is not yet created or 
linked from SFAS individual mapping. If the student is not created yet by the time SFAS mapping queue was running and trying to map a student SIMS, then it would not be able to link the SFAS individual to an existing student in SIMS, and it would take some extra effort to handle this scenario. That's how I understand it.

lewischen-aot: The above comment has been justified in the chat. We'll see how it goes next.

sources/packages/backend/libs/services/src/students-assessments/assessment-sequential-processing.service.ts
```diff
@@ -542,11 +531,10 @@ export class AssessmentSequentialProcessingService {
...
       .addSelect("SUM(sfasApplication.csgdAward)", "CSGD")
       .addSelect("SUM(sfasApplication.bcagAward)", "BCAG")
-      .innerJoin("sfasApplication.individual", "sfasStudent")
+      .innerJoin("sfasApplication.individual", "sfasIndividual")
+      .innerJoin("sfasIndividual.student", "student")
       .where("sfasApplication.applicationCancelDate IS NULL")
-      .andWhere("lower(sfasStudent.lastName) = lower(:lastName)", { lastName })
-      .andWhere("sfasStudent.sin = :sin", { sin })
-      .andWhere("sfasStudent.birthDate = :birthDate", { birthDate })
+      .andWhere("student.id = :studentId", { studentId })
```
andrewsignori-aot, line 537: In the same way as the other, this one can also do the where condition on the student without the need to join with the student, right?

sources/packages/backend/libs/services/src/students-assessments/assessment-sequential-processing.service.ts
```diff
@@ -592,11 +576,10 @@ export class AssessmentSequentialProcessingService {
...
       .addSelect("SUM(sfasPTApplication.bcagAward)", "BCAG")
       .addSelect("SUM(sfasPTApplication.csptAward)", "CSPT")
-      .innerJoin("sfasPTApplication.individual", "sfasStudent")
+      .innerJoin("sfasPTApplication.individual", "sfasIndividual")
+      .innerJoin("sfasIndividual.student", "student")
       .where("sfasPTApplication.applicationCancelDate IS NULL")
-      .andWhere("lower(sfasStudent.lastName) = lower(:lastName)", { lastName })
-      .andWhere("sfasStudent.sin = :sin", { sin })
-      .andWhere("sfasStudent.birthDate = :birthDate", { birthDate })
+      .andWhere("student.id = :studentId", { studentId })
```
andrewsignori-aot, line 582: Similar to https://github.com/bcgov/SIMS/pull/4369/files#r1966257261

andrewsignori-aot (commented):
Thanks for making the changes. Business is updating the ACs about the overawards during student creation and we can proceed with the changes.

sources/packages/backend/libs/services/src/sfas/sfas-individual.service.ts
```diff
@@ -128,33 +100,14 @@ export class SFASIndividualService {
...
     const sfasIndividualData = await this.sfasIndividualRepo
       .createQueryBuilder("sfasIndividual")
       .select(
         "SUM(sfasIndividual.unsuccessfulCompletion)::int",
         "totalUnsuccessfulWeeks",
       )
-      .where("sin = :sin", { sin })
-      .andWhere("birth_date = :birthDate", { birthDate })
-      .andWhere("LOWER(last_name) = LOWER(:lastName)", { lastName })
+      .innerJoin("sfasIndividual.student", "student")
```
andrewsignori-aot, line 109: Minor, this one can also use the student ID directly instead of using the `inner join`.

andrewsignori-aot (approved):
Thanks for making the changes and accommodating the AC change. Looks good 👍

Issue #4247, opened 2025-01-15
Create Application Versioning and use the versioning to get previous version(s)


dheepak-aot:
**Describe the task**
When a student application is edited and submitted one or more times, then we have previous version(s) of the same application. Currently we use the `application number`(not id) and the `submitted date` to retrieve the previous versions.

Instead we should start to version the application in DB (Similar to offerings) and use the version to retrieve the previous versions. 

**Acceptance Criteria**
- [ ] Create versioning for `sims.applications`
- [ ] Restructure the way how routes and UI are loaded in `student application view` in ministry(comes from student search).
 - The outcome must result in the current version routes(highlighted in screenshot) always loading latest version of application on-refresh.
 - Update the API loading the previous versions in sidebar to use versioning.

<img width="1004" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDkyQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c4eed740533d213a6f0bb789b5d1e9666a1ddf1a/image.png" alt="image.png" />

**Additional context**
- Add any other context about the task here.
- Or here

**Technical**
- [x] Create new columns to keep the history: `parent_application_id` and `preceding_application_id`.
- [x] Populate the new columns when an application is edited.
- [x] Create the DB migration to have the new columns populated.
- [ ] Adjust the necessary methods in the API to allow the student application view on the Ministry (image above) to retrieve information based on the new `application_parent_id`.
- [ ] Ensure the route for the student application view on the Institution is also adjusted.
- [ ] Adjust the E2Es
  - Create an E2E to ensure the current application will be the one retrieved based on the `application_parent_id`.

ninosamson:
Tagging this to 2.6 - for QA @astridSABC for 2.5, we should still test create, edit, resubmit and draft applications as a regression test. 

cc: @Joshua-Lakusta

PR #4394, opened 2025-02-26
#4247 - Create Application Versioning (Institution and Supporting User changes)

sh16011993:
### As a part of this PR, the following are completed:

- Route changes for the Institution View
- Changes for the Supporting User


sources/packages/web/src/components/layouts/aest/AESTApplicationSideBar.vue
```diff
@@ -99,6 +99,7 @@ export default defineComponent({
     const { getISODateHourMinuteString } = useFormatters();
     const relatedParentPartners = ref([] as MenuItemModel[]);
     const applicationHistory = ref([] as MenuItemModel[]);
+    const currentApplicationId = ref<number>();
```
andrewsignori-aot, line 102: There is no need to have it declared as reactive property and to be at this scope. Please change it to a regular property and use it only inside the `onMounted` scope.

sources/packages/web/src/services/http/ApplicationApi.ts
```diff
@@ -36,9 +36,18 @@ export class ApplicationApi extends HttpBaseClient {
...
+    options?: {
+      studentId?: number;
+      loadDynamicData?: boolean;
+      isParentApplication?: boolean;
+    },
   ): Promise<ApplicationDataAPIOutDTO | ApplicationSupplementalDataAPIOutDTO> {
-    let url = this.addClientRoot(`application/${applicationId}`);
+    let url = this.addClientRoot("application");
+    if (options?.studentId) {
+      url = `${url}/student/${options?.studentId}/application/${applicationId}`;
```
andrewsignori-aot, line 47: `${options?.studentId}` does not require null propagation and can have the "?" removed as `${options.studentId}`. Same for the below one that is not part of this PR.
```ts
url = `${url}?loadDynamicData=${options.loadDynamicData}`
```

sources/packages/web/src/views/institution/student/applicationDetails/ApplicationExceptions.vue
```diff
@@ -31,6 +32,17 @@ export default defineComponent({
...
   },
   setup(props) {
+    const currentApplicationId = ref<number>();
+
+    onMounted(async () => {
+      // Get current application for the parent application.
+      const currentApplication = await ApplicationService.shared.getApplication(
+        props.applicationId,
+        { loadDynamicData: false, isParentApplication: true },
+      );
```
andrewsignori-aot, line 42: Depending on how we proceed in the upcoming expansion of the history tickets, if this logic is no longer temporary we should be moved to the service to avoid duplication.
We can have a method at the ser ice level as 
`ApplicationService.shared.getCurrentApplicationVersion(parentApplicationId: number)`
Does it make sense?

lewischen-aot:
As I'm reviewing the PR by running it locally, I found out that the order of the application versions is probably incorrect, and I'm not sure if it's from the backend or the frontend. For application number 2024000007, the view page directs me to application ID 553 once I clicked on view application, but the application status of application ID 553 is `overwritten`. The current application ID should be 571. 
Screenshot 2025-02-26 110726

Screenshot 2025-02-26 110857

sources/packages/web/src/services/ApplicationService.ts
```diff
@@ -44,16 +44,46 @@ export class ApplicationService {
...
   }
 
+  /**
+   * Get current application from parent.
+   * @param parentApplicationId parent application id.
+   * @param options options for the request.
+   * - `studentId` student id for the student.
+   * @returns application information.
+   */
+  async getCurrentApplicationFromParent(
```
andrewsignori-aot, line 71: Is there a problem if the method is simplified as below?
```ts
async getCurrentApplicationFromParent(
    parentApplicationId: number,
    options?: { studentId: number },
  ): Promise<ApplicationSupplementalDataAPIOutDTO> {
  return this.getApplication(parentApplicationId, {
    studentId: options?.studentId,
    isParentApplication: true,
  });
}
```

andrewsignori-aot (approved):

lewischen-aot (approved):
LGTM 👍 Thanks for clarifying the behavior for the application viewer page.

andrewsignori-aot:
Merging the PR as agreed with @sh16011993 will allow it to reach DEV and have some tests performed on Monday, when he will be back.

Issue #4343, opened 2025-02-11
Updates to KeyCloak 


ninosamson:

ninosamson:
Hello SSO Friends, 
  
We are reaching out to share that we are moving to a new version of Red Hat Build of Keycloak 26 (RHBK 26). We plan to upgrade it in February - March 2025. 
 
In preparation for this upgrade, we highlighted some important updates regarding RHBK 26 that may require action from you to keep the integration between the SSO service and your app working. The full release note can be found here. 
 
1. The Keycloak JS library is no longer served statically from the Red Hat build of Keycloak server 
 
If your application links to Keycloak server with URLs such as: 

/js/keycloak-authz.js 

/js/keycloak-authz.min.js 

/js/keycloak.js 

/js/keycloak.min.js 

/js/{version}/keycloak-authz.js 

/js/{version}/keycloak-authz.min.js 

/js/{version}/keycloak.js 

/js/{version}/keycloak.min.js 

you will not have access to the file(s) once we upgrade to RHBK26, and your users will not be able to authenticate through your app. By Red Hat build of Keycloak 26.0 Upgrading Guide, you should now include the library in your project using a package manager such as NPM. The library is available on the NPM registry as keycloak-js. 

2. Support for legacy logout through redirect_uri parameter is removed 

Previous versions of Keycloak had supported automatic logout of the user and redirect to the application by opening logout endpoint URL with redirect_uri parameter. We communicated with the community that this functionality was deprecated in Keycloak 18 and now has been removed in RHBK 26. We recommend including id_token_hint and post_logout_redirect_uri parameters. 

3. You are already logged in message  

When a user is already logged in to one browser tab and an authentication session expired in another browser tab, RHBK 26 redirects back to the client application with an OIDC/SAML error. Note that the message You are already logged in does not appear to the end user when an authentication session expires and user is already logged in. You may consider updating your applications to handle this error. Please reach out to us if you have more questions about this error. 

4. Changes on token claims to reduce size of lightweight access token 

Claim sub and auth_time are added by protocol mapper, which are still added to the ID token and access token as before, but not to the lightweight access token

Claim nonce is added only to the ID token by default 

Claim session_state is not added to any token now by default. The other dedicated claim sid is still supported by the specification, which was available in previous versions and has exactly the same value 

For backward compatibility, through protocol mappers, SSO team adds nonce and session_state claims for standard clients. If you are a custom client, please review your token claim usage, and configure accordingly. More details can be found here. 

[...]

PR #4400, opened 2025-02-26
#4343 - Updates to KeyCloak - start after Feb25

bidyashish:
**Technical**
- [x] Review the comments added by @ninosamson below to check any possible impact.
- [x] Update keycloakjs lib can be done without any breaks. If the login is happening for all portals it should be good enough to proceed.
- [x] To be executed after Feb 25 (Dev/TEST) to check possible impacts.


andrewsignori-aot (approved):
Thanks for separating the changes, looks good 👍

andrepestana-aot (approved):
Looks good!

PR #4430, opened 2025-03-05
#4247 - Revert route changes and parent application (Vue only)

andrewsignori-aot:
Reverting changes related to the route with the parent application ID instead of the application ID itself, due to complexities moving forward with the changes that affected more areas than what was foreseen by the Team.

### DIFF
--- a/sources/packages/web/src/components/common/NoticeOfAssessmentFormView.vue
+++ b/sources/packages/web/src/components/common/NoticeOfAssessmentFormView.vue
@@ -23,7 +23,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, ref, watchEffect } from "vue";
+import { defineComponent, ref, onMounted } from "vue";
 import { StudentAssessmentsService } from "@/services/StudentAssessmentsService";
 import { AssessmentNOAAPIOutDTO } from "@/services/http/dto";
 import { ModalDialog, useFormatters } from "@/composables";
@@ -205,7 +205,7 @@ export default defineComponent({
       );
     };
 
-    watchEffect(loadNOA);
+    onMounted(loadNOA);
 
     const setMSFAAReissueProcessing = (processing: boolean) => {
       msfaaReissueProcessing.value = processing;
--- a/sources/packages/web/src/components/common/NoticeOfAssessmentFormView.vue
+++ b/sources/packages/web/src/components/common/NoticeOfAssessmentFormView.vue
@@ -150,9 +150,6 @@ export default defineComponent({
     };
 
     const loadNOA = async () => {
-      if (!props.studentId || !props.applicationId || !props.assessmentId) {
-        return;
-      }
       const assessment =
         await StudentAssessmentsService.shared.getAssessmentNOA(
           props.assessmentId,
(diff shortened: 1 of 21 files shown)


sources/packages/web/src/components/common/students/assessment/History.vue
```diff
@@ -124,13 +122,11 @@ export default defineComponent({
     // Adding watch effect instead of onMounted because
     // applicationId may not be not available on load.
     watchEffect(async () => {
-      if (props.applicationId) {
```
sh16011993, line 127: I understand that we reverted this back because now this will come instantly as it is not derived in the parent component through an api call. Shouldn't we keep this `if` check here considering that the `getAssessmentHistory` api call has `props.applicationId` as mandatory?

sh16011993: I think we can safely remove the `if` check from both the `History.vue` and `Request.vue` since it is coming directly from the parent view route.

sources/packages/web/src/views/aest/student/applicationDetails/NoticeOfAssessment.vue
```diff
@@ -20,7 +20,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, onMounted, ref } from "vue";
+import { defineComponent } from "vue";
 import NoticeOfAssessmentFormView from "@/components/common/NoticeOfAssessmentFormView.vue";
 import { AESTRoutesConst } from "@/constants/routes/RouteConstants";
 import { ApplicationService } from "@/services/ApplicationService";
```
dheepak-aot, line 26: Do we still need the `ApplicationService` here?

sources/packages/web/src/components/common/NoticeOfAssessmentFormView.vue
```diff
@@ -150,9 +150,6 @@ export default defineComponent({
     };
 
     const loadNOA = async () => {
```
dheepak-aot, line 152: The `loadNOA` is used under watch effect, so we can either revert the watch effect keep some condition under watch effect(e.g. any  of these props changing) to call the loadNOA.

sources/packages/web/src/components/common/students/assessment/History.vue
dheepak-aot: WatchEffect reacts to any change in the reactive state. So I would recommend keeping the if unless I am missing something.

andrewsignori-aot: Added it back.

dheepak-aot (approved):
Thanks for taking care of the revert. Looks good 👍

andrewsignori-aot:
> Thank You for reverting the changes @andrewsignori-aot.
> 
> I tested the following as a part of the review:
> 
> * Summitted an application, approved assessment, enrollment.
> * Submitted an appeal for an application by the student. Viewed the appeal request form in the ministry, approved it. Checked the completed appeals request form in the Completed assessments section. Checked the assessment and the NOA page for the same.
> * Raised a scholastic standing for a completed application from the institution and checked the completed request form for the same. Checked the assessment and the NOA page for the same.
> * Edited an application and raised an exception flow for an application. Approved the exception from the ministry. Checked the request form and the assessment and the NOA pages from the ministry.
> 
> All the above looks good. 👍

Amazing @sh16011993, thanks 😉

Issue #4281, opened 2025-01-23
SDPR API - Response Schema Update


CarlyCotton:
**User Story**
As a system, SIMS needs to provide more data for SDPR ISA for them to be able to implement their interface policy. Interface policy exists for SDPR to review students who are receiving assistance towards living costs and validate if they are also receiving living expenses funding from SABC financial aid.

**Acceptance Criteria**
- [x] Student details should be shown even if there is NO eligible application data to share.

- [x] Successful SIN Match Outcomes:
    - [x] SFAS Only: Send student details from `sfas_individuals` and any applications within the timeframe.
    - [x] SIMS Only: Send student details **IF** the current SIN has been validated, otherwise no match.
    - [x] SFAS & SIMS: Send student details from SIMS **IF** the current SIN has been validated, otherwise send `sfas_individuals`
---
- [x] Update incoming field in the bridge file from "Assessed costs-discretionary" into "Eligible need". Same position and length.

---
- [x] Applications to include on successful matches:
    - [x] Include only applications that are FT, AND
    - [x] Within the current program year PLUS 2 previous program years
        - If request is September 1, 2024: include current PY (24/25) and two prev (22/23 and 23/24)
        - If request is June 1, 2024: include current PY (23/24) and two prev (21/22 and 22/23)
    - [x] Show all SFAS applications that meet the above 
    - [x] Show all SIMS applications that meet the above AND:
        - [x] Application is not Overwritten, Draft, or In progress.
        - [x] Use most current assessment completed.

- [x] Institution information will always be pulled from matching SIMS institution code. If nothing found, just include institution code, everything else institution based can be empty/null. 

---
- [x] Update Schema to include more items:
```
student:
    givenNames 
    lastName
    sin
    dateOfBirth
    phoneNumber
    address:
        addressline1
        addressline2
        city
        prov
        postal
        country
applications:
    appNumber: applications.application_number
    appStatus:
    cancelDate: 
    withdrawDate: // dateOfWithdrawal 
    withdrawReason: //include "Student withdrew from program" events
    withdrawActive: //SFAS only for now until SIMS builds it
    immigrationStatus: applications.data{citizenship}
    bcResidency:
    legacyPermanentResident:
    maritalStatus: 
    legacyMaritalDate:  //only include if SFAS application data, 
    income: applications.data{taxReturnIncome}
    livingArrangement: (home/away)
    estimatedTotalAward: //sum of all disbursement amts from app without Total BC Grants similar to NOA
[...]

PR #4432, opened 2025-03-05
#4281 - Default Withdrawal Related CHAR columns to NULL

dheepak-aot:
Updated the fallback value for `withdrawalReason` and `withdrawalActiveFlag` as NULL. 

Otherwise as these columns are CHAR and NOT VARCHAR empty string is added as fillers to the value.


guru-aot (approved):
LGTM, nice work @dheepak-aot

andrewsignori-aot (approved):
Looks good 👍

Issue #3739, opened 2024-09-24
Release DB Migrations Rollback


andrewsignori-aot:
During the release process to Staging and Production, the release rollback includes reverting the DB migration already executed, if any. The current process is a temporary workaround and a better approach should be taken. 

**Acceptance criteria**
- [ ] Evaluate possible solutions for reverting the DB migrations.
- [ ] Evaluate the possible creation of a POD with the DB migrations source code to allow the `migration:revert` to be executed.
- [ ] Check alternative solutions.
- [ ] Check why the DB migration demands more CPU and memory than the usual required to execute the applications.

ninosamson:
Moving back to "In Review" for bug fix.

PR #4457, opened 2025-03-10
#3739 - Release DB Migrations Rollback - Allow migration to fail on CI/CD

andrewsignori-aot:
When CI/CD is executing the migrations, the job was considered successful even if a migration failed to be executed.
Changed the behavior of the exceptions handling when they are executed in the CI/CD pipeline or using the REPL mode.

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/db-migrations.service.ts
+++ b/sources/packages/backend/apps/db-migrations/src/db-migrations.service.ts
@@ -4,6 +4,27 @@ import { ormConfig } from "./data-source";
 import { inspect } from "util";
 
 /**
+ * Default limit to use when listing migrations.
+ */
+export const DEFAULT_LIST_LIMIT = 5;
+
+/**
+ * Behavior to use when a failure occurs.
+ * Useful to control the behavior when the migration is executed
+ * in the CI/CD pipeline or using the REPL mode.
+ */
+export enum FailureBehavior {
+  /**
+   * Only log the error and continue.
+   */
+  LogOnly = "log-only",
+  /**
+   * Rethrow the error.
+   */
+  Rethrow = "rethrow",
+}
+
+/**
  * DB migrations options available.
  * Each operation is executed in an isolated data source.
  */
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/db-migrations/src/db-migrations.service.ts
```diff
@@ -13,8 +34,10 @@ export class DBMigrationsService {
 
   /**
    * Run all pending migrations.
+   * @param failureBehavior the behavior to use when a failure occurs.
+   * Defaults to log only.
    */
-  async run(): Promise<void> {
+  async run(failureBehavior = FailureBehavior.LogOnly): Promise<void> {
```
dheepak-aot, line 40: What is the benefit of setting the default value here?

sources/packages/backend/apps/db-migrations/src/db-migrations.service.ts
```diff
@@ -25,13 +48,15 @@ export class DBMigrationsService {
...
+    }, failureBehavior);
   }
 
   /**
    * Revert the last migration.
+   * @param failureBehavior the behavior to use when a failure occurs.
+   * Defaults to log only.
    */
-  async revert(): Promise<void> {
+  async revert(failureBehavior = FailureBehavior.LogOnly): Promise<void> {
```
dheepak-aot, line 59: https://github.com/bcgov/SIMS/pull/4457/files#r1988214582

dheepak-aot (approved):
Thanks for taking care of the issue. Looks good 👍

Minor clarification.

sources/packages/backend/apps/db-migrations/src/db-migrations.service.ts
andrewsignori-aot: After further testing, I notice the RPEL is resilient to unhandled exceptions, hence I simplified the code, removing the `catch` and letting the exception be thrown all the times which would fix the migration issue for CI/CD and would also not affect the REPL mode with the benefit of simplified the code.

guru-aot (approved):

Issue #4458, opened 2025-03-11
BUG: Application versioning not displayed below Application Status on Dev and Test


astridSABC:
**Describe the Bug**
Dev and Test: When an app is edited, the original version is not displayed with the date and time below the Application status, no dropdown arrow is available.  When an app is edited again, the next version is displayed in a chronological list.

**Expected Behaviour**
Once an app is edited, the original version and subsequent versions should be displayed under the "Application Status" in the left pane in chronological order.  Each has it's own date, time stamp with a dropdown arrow.

Dropdown arrow is available for selection beside the date and time stamp
Select the trencher icon (graduation cap) or "application"
Select through all the tabs and note the information selected on the pages

**Actual Behaviour**

Version is not displayed

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBODAyQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6e302ffbb3c5084e79669070b798a9ea40204ddd/image.png" alt="image.png" />

** Steps To Reproduce**
Today: Create a single application with no PD/dependants and any income below $36,111 (original application and assessment)

First Edit: Later the same day: Edit the application (add a dependant) and submit the application. At assessment status

Steps to reproduce the behaviour:
First Edit Precondition

1. Login to the ministry portal
2. Click “search students” in the left pane
3. Enter the name of the student
4. Click “search” or press ENTER
5. Click “view” in the row for the student from above
6. Select “applications” tab
7. Click “view” beside the application that you want to review from above
8. Note the screen

Ref. TC 4327, Release 2.6

PR #4463, opened 2025-03-11
#4458 - Fix the overall status API to use application id

dheepak-aot:
## Fix the overall status API to use the application id and get the edited versions of the same

- [x] Added a quick fix (additional query) to get the `parent-application-id` from the given `application-id` and get the versions.

Application with version.
image

Application without version.

image


sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1860,19 +1860,26 @@ export class ApplicationService extends RecordDataModelService<Application> {
 
   /**
    * Get previous application versions for an application.
-   * @param parentApplicationId parent application id.
+   * @param applicationId application id.
    * @returns previous application versions.
    */
   async getPreviousApplicationVersions(
-    parentApplicationId: number,
+    applicationId: number,
```
andrewsignori-aot, line 1867: Minor, I would suggest to have the method renamed to `getApplicationVersions`, not a blocker.

andrewsignori-aot (approved):

guru-aot (approved):
LGTM, nice work @dheepak-aot

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: I will address that in the PR towards main.

Issue #4354, opened 2025-02-13
Change Request Revamp - 1st effort (database and sub event name changes)


Joshua-Lakusta:
As a ministry user I need the application to be able to have an edit post COE that allows for different states for the edit request so that their can be a ministry decision on the edit. this is important because all changes post COE must have ministry intervention. 

Acceptance Criteria

#### Database Changes
- [x] Rename application status from `Overwritten` to `Edited`, and review all impacted places.
- [x] Create a new column on `sims.applications`.
  - [x] Ensure all applications where `parent_application_id = preceding_application_id` will be updated to `Original`
  - [x] Ensure all applications where `parent_application_id != preceding_application_id` will be updated to `Edited`
  - [x] `application_edit_status`
    - `Original`: default status when the application was never edited in any way.
    - `Change in progress`: status for the new application version to be edited after COE.
    - `Change pending approval`: the workflow is being executed and did not reach the "Ministry approval" yet. May be waiting for supporting users or income.
    - `Change declined`: when the edit was requested after COE and Ministry declined it.
    -  `Change cancelled` : When the edit is cancelled by the student
    - `Changed with approval`: when the edit was requested after COE and Ministry accepted it.
    - `Edited`: when the edit was executed and no Ministry approval was needed (edit before COE).
Audit columns should be created to capture user and modified dates for the last application_edit_status update. Alternatively, a note relation can be kept also.
  - [x] `application_edit_status_updated_on`
  - [x] `application_edit_status_updated_by`
  - [x] Create a note record in preparation for having a note as a part of the approval or decline process. This will eventually be used to ID user, time, and date of the change. Note will be mandatory part of the decision process

#### Application changes
- [x] Ensure current application save draft, submit, and edit still work and set the new statuses.
- [x] Ensure the new statuses will be persisted as `Original` when a new draft application is created.
- [x] When an application is edited by a student set the new columns as below.
  - [x] `application_status`: `Edited`
  - [x] `application_edit_status`: `Edited`
  - [x] `application_edit_status_updated_on`: current date available and shared on the method.
  - [x] `application_edit_status_updated_by`: student user ID.
 
#### Workflow Changes
- [x] Ensure workflow will be executed in a backward-compatible way.
- [x] Skip exceptions approval and application status as shown below.
[...]

PR #4464, opened 2025-03-11
#4354 - Change Request Revamp - 1st effort - Workflow Changes

andrewsignori-aot:
## Workers changes

- Created a new worker `application-change-request-approval` to process the Ministry review of an application change request.
- The worker will be executed only for applications with the edit status set as `Change in progress`.
- It is the worker's responsibility to set the application's edit status to `Change pending approval` to keep the workflow waiting for some new action from the Ministry, to approve or reject it, or from the student, to cancel it.

### E2E Tests

- ApplicationController(e2e)-applicationChangeRequestApproval
  - Should update application edit status to Change pending approval when the Ministry approval is required.
  - Should not update the application edit status when it is already set as Change pending approval.
  - Should not update the application edit status when it is set as Change cancelled.

### New alternative flow to skip application status change and exceptions verifications

- The application status will remain as it is (`Edited`).
- No exceptions verifications will be executed.

image

### New `change request` approval flow

- When the application edit status is set to `Change in progress`, the new flow will be executed and wait for some action from the Ministry or the Student, as explained before.
- The edit status will be set as `Change pending approval` and the workflow will be waiting for timer to check the most recent DB status or for a message to be received (following the same approach from PIR and exceptions).
- Also following the same approach from exceptions, if the workflow will receive a "cancellation status" (`Change cancelled` or `Change declined`) and it will proceed to the `end`.

image

### PIR ignored

image

### E2E Tests
- E2E Test Workflow assessment gateway on change requests for <program year>
  - Should complete the assessment calculations when a change request was approved by the Ministry and a message was sent to the workflow.
  - Should end the assessment workflow when a change request is declined by the Ministry via message.
  - Should end the assessment workflow when a change request is declined by the Ministry via task result.
  - Should end the assessment workflow when a change request is cancelled by the student via message.
  - Should end the assessment workflow when a change request is cancelled by the student via task result.

## Non-related changes

- VS Code tasks to execute E2E tests were updated to allow the execution of files with the same name individually. Before, files with the same name were executed together, for instance, the E2E files for different program years that share the same name.
- Create two new tasks in the workspace for easy execution of Camunda and forms definitions.

## TODO
- Set the appropriate assessment status before the workflows execution ends, adapting and reusing the `Workflow wrap up` task or creating something similar, where the last seems more appropriate because we just need to execute something similar to the method [updateAssessmentStatusAndSaveWorkflowData](https://github.com/bcgov/SIMS/blob/ac0b8ca1951210e0691b03617cdd9e6940afaabc/sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts#L297).
_Note:_ the same could be reused for when the workflow is ended due to the Ministry declining exceptions.


.vscode/launch.json
```diff
@@ -98,7 +99,6 @@
       ],
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen",
-      "disableOptimisticBPs": true,
```
dheepak-aot, line 101: Asking for my understanding. What is is this property?

Also this is still present in line 80. 
image

sources/packages/backend/apps/workers/src/controllers/application/_tests_/e2e/application.controller-applicationChangeRequestApproval.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    // Act
+    const result = await applicationController.applicationChangeRequestApproval(
+      createFakeWorkerJob<
+        ApplicationChangeRequestApprovalJobInDTO,
+        ICustomHeaders,
+        ApplicationChangeRequestApprovalJobOutDTO
+      >(updateApplicationStatusPayload),
+    );
+
+    // Asserts
```
dheepak-aot, line 65: Minor. should start calling `//Assert` as `//Asserts` on multiple assertions ?

sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
```diff
@@ -168,4 +186,78 @@ export class ApplicationController {
...
+   * Processes the application change request approval.
+   * Set the application edit status to 'Change pending approval' to allow waiting
+   * till the Ministry approves or declines the changes.
+   * @returns most updated edit status.
+   */
+  @ZeebeWorker(Workers.ApplicationChangeRequestApproval, {
+    fetchVariable: [APPLICATION_ID],
+    maxJobsToActivate: MaxJobsToActivate.Normal,
+  })
+  async applicationChangeRequestApproval(
```
dheepak-aot, line 200: I see that this worker has been used for 2 different purposes, 1. Update the Edit Status 2. Return the current edit_status. 

Even though, if there is additional effort to introduce another worker, I feel using a worker to `Update Edit Status` and `Verify Approval` sounds better to me. Open for discussion.

.vscode/launch.json
andrewsignori-aot: I started to have it removed and I gave up because it was not clear to me either and I would not like to spend time on this 😄 
The short answer is that initial research did not show it was required, but I would not want to disturb the PR, and this one ended up being removed by mistake.

sources/packages/backend/apps/workers/src/controllers/application/_tests_/e2e/application.controller-applicationChangeRequestApproval.e2e-spec.ts
andrewsignori-aot: Copy and paste from other E2E tests 😉

sources/packages/backend/apps/workers/src/services/application/application.service.ts
```diff
@@ -100,4 +106,32 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    return this.repo.update(
+      {
+        id: applicationId,
+        applicationEditStatus: fromStatus,
+      },
+      {
+        applicationEditStatus: toStatus,
+        applicationEditStatusUpdatedOn: now,
+        applicationEditStatusUpdatedBy: this.systemUsersService.systemUser,
+        modifier: this.systemUsersService.systemUser,
```
dheepak-aot, line 132: Minor. We can assign a `this.systemUsersService.systemUser` to a const and use in both places.

sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
andrewsignori-aot: The idea was to follow a similar approach from `verify-application-exceptions`, which is also for "two purposes" and not to mention some other workers like `verify-assessment-calculation-order` and `workflow-wrap-up`.
In terms of a worker executing a single responsibility, yes, looks better, but we usually take into consideration the workflow cost of executing a worker call/process. Does it make sense?
I would be more inclined to create an additional worker if one of those suggested separations would be reused in more than one place.

sources/packages/backend/apps/workers/src/controllers/application/_tests_/e2e/application.controller-applicationChangeRequestApproval.e2e-spec.ts
andrewsignori-aot: Changed.

.vscode/launch.json
andrewsignori-aot: Reverted to add it back.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
```diff
@@ -789,10 +568,342 @@
...
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=programInfoStatus = "Required"</bpmn:conditionExpression>
+    </bpmn:sequenceFlow>
+    <bpmn:sequenceFlow id="Flow_1ds5r8m" sourceRef="pir-pending-event-gateway" targetRef="program-info-request-timer" />
+    <bpmn:sequenceFlow id="Flow_1sze0w6" sourceRef="program-info-request-timer" targetRef="program-info-required-task" />
+    <bpmn:sequenceFlow id="Flow_0guy66a" sourceRef="program-info-required-task" targetRef="pir-pending-gateway" />
+    <bpmn:sequenceFlow id="needs-pir-flow" name="Needs PIR" sourceRef="needs-pir-gateway" targetRef="program-info-required-task">
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=studentDataSelectedOffering = null</bpmn:conditionExpression>
+    </bpmn:sequenceFlow>
+    <bpmn:sequenceFlow id="pir-not-required-flow" name="PIR not required or Change in progress" sourceRef="needs-pir-gateway" targetRef="program-info-not-required-task">
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=studentDataSelectedOffering != null or applicationEditStatus = "Change in progress"</bpmn:conditionExpression>
```
dheepak-aot, line 690: I feel that it will be good to use a separate gateway for the deviation based on edit status. 
image

dheepak-aot: Same for PIR. 
image

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
```diff
@@ -789,10 +568,342 @@
...
+          <zeebe:input source="=&#34;load-assessment-data-submit-or-reassessment-subprocess&#34;" target="load_assessment_data_submit_or_reassessment_subprocess_passthrough" />
+        </zeebe:ioMapping>
+      </bpmn:extensionElements>
+      <bpmn:incoming>Flow_1wrj7fh</bpmn:incoming>
+      <bpmn:outgoing>Flow_0rxxozm</bpmn:outgoing>
+    </bpmn:callActivity>
+    <bpmn:sequenceFlow id="Flow_1ef3map" sourceRef="assessment-start-event" targetRef="associate-workflow-instance-task" />
+    <bpmn:sequenceFlow id="Flow_1wrj7fh" sourceRef="associate-workflow-instance-task" targetRef="load-assessment-data-submit-or-reassessment-subprocess" />
+    <bpmn:sequenceFlow id="Flow_0g0yluv" name="Original application submission" sourceRef="Gateway_1imkna6" targetRef="Gateway_139oo1r">
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=applicationEditStatus = "Original"</bpmn:conditionExpression>
```
dheepak-aot, line 822: What about `applicationEditStatus = "Edited"` ? The edit before COE.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
andrewsignori-aot: Lets check PIR first, fo you mean something like below?

image

Instead of the "OR" condition as below?

image

dheepak-aot:
Great progress towards the new change request process. Please have a look at the comments.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
andrewsignori-aot: Edit before COE never reaches the workflow as 'Edited', it follows a new submission from scratch.

sources/packages/backend/apps/workers/src/services/application/application.service.ts
dheepak-aot: Considering `.systemUser` equivalent to a method, I mentioned that. Still not a blocker.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
dheepak-aot: I see -  The new path also requires program info status to be updated. I thought it will skip the whole PIR block.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
dheepak-aot: I raised the question looking at this block of code where new application created has the edit_status as `Edited`.  Let me know if I am missing something here.

image

sources/packages/backend/apps/workers/src/controllers/application/application.controller.ts
dheepak-aot: I agree that unless we have more purpose for the new worker we can follow the current approach in PR.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
andrewsignori-aot: Got it now. I am changing that flow to be the default one from now on.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
andrewsignori-aot: As per our conversation, that is how it will look like.
image

dheepak-aot (approved):

Issue #4390, opened 2025-02-25
CAS Manual Intervention Table


ninosamson:
- [ ] Create new UI table to include individual invoice error records (Manual Intervention).
- [ ] Each record should include the response (CAS request?) date, batch name, invoice number, supplier number, error information.
- [ ] Each record can be actioned:
    - [ ] 'Resolved' to indicate user has dealt with it directly within CAS. 
    - [ ] When resolved, track which user resolved it and when it was resolved.

**Technical AC**

- [ ] Decision to reuse the same Ministry role `aest-cas-invoicing` as business did not provide one and it was ok with it.

DB
- [ ] Create a migration status `Resolved` in `invoice_status` in `cas_invoices` table.
- [ ] Create a migration to capture the user changing the status using the same naming conventions from other tables doing the same.

API
- [ ] Create a new controller named `cas-invoice`
  - [ ]  New endpoint to get `list of manual intervention invoices` from `cas-invoices` table, with lazy pagination enabled.
  - [ ]  New endpoint to update the status.
- [ ] The response date shall be `invoice_status_updated_on` from the table.
- [ ] Sort on last invoice update date ASC first.
- [ ] Create an api `patch` request to update status to `Resolve`, which validates the Ministry role to make the update and persist the user executing the update.
- [ ] Update `Audit DB` Column on update about user action.

UI
- [ ] Create a `V-Datable-Server` Datatable to view Manual Intervention below invoice batch in Ministry portal.
- [ ] Show error in expander in UI (similar to CAS supplier).
- [ ] Use the action button to resolve the status from `Manual Intervention` to `Resolved`.
- [ ] Show a Modal to confirm the action to change status, using the component `confirm-modal`.
- [ ] Action should be enabled only for a user with a valid role.
- [ ] Enable the sort on first database column `Response date`.
- [ ] Once some record is actioned, refresh the result table.

E2E
- [ ] E2E test for new API endpoints.

bidyashish:
Question for @CarlyCotton 

- which ministry role to make the update?
- Should Ddatable show `resolved` status based on filter or will it be seperate task?

CarlyCotton:
@bidyashish 

- Could have it's own or be attached to the role needed to view the page: aest-cas-invoicing
- Datatable: Can just be invoices that are manual intervention. Do not need to include a filter and resolved invoices at this time.

PR #4467, opened 2025-03-12
#4390 - CAS Manual Intervention: Add `CASInvoiceStatusUpdatedBy` Column

sh16011993:
### As a part of this PR, the following have been completed:

- Added the migration for the `CASInvoiceStatusUpdatedBy` column.

**Rollback Screenshot:**

image

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Add-cas-invoice-status-updated-by-column.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Add-cas-invoice-status-updated-by-column.sql
@@ -1,0 +1,27 @@
+ALTER TABLE
+  sims.cas_invoices
+ADD
+  COLUMN invoice_status_updated_by INT REFERENCES sims.users(id);
+
+COMMENT ON COLUMN sims.cas_invoices.invoice_status_updated_by IS 'User who updated the invoice status.';
+
+-- Update the existing records with the system user.
+UPDATE
+  sims.cas_invoices
+SET
+  invoice_status_updated_by = (
+    SELECT
+      id
+    FROM
+      sims.users
+    WHERE
+      last_name = 'system-user'
+  );
+
+-- Update the column to NOT NULL.
+ALTER TABLE
+  sims.cas_invoices
+ALTER COLUMN
+  invoice_status_updated_by
+SET
+  NOT NULL;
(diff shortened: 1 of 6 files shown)


sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Add-cas-invoice-status-updated-by-column.sql
```diff
@@ -0,0 +1,27 @@
...
+ADD
+  COLUMN invoice_status_updated_by INT REFERENCES sims.users(id);
+
+COMMENT ON COLUMN sims.cas_invoices.invoice_status_updated_by IS 'User who updated the invoice status.';
+
+-- Update the existing records with the system user.
+UPDATE
+  sims.cas_invoices
+SET
+  invoice_status_updated_by = (
```
dheepak-aot, line 12: Can we not use the creator value from same column? - Even though creator is nullable current code ensures that creator is saved during invoice creation process.

sh16011993: @dheepak-aot  I had checked with @andrewsignori-aot on this. He mentioned that for the existing records I should assign this column value to be the `system-user`.

dheepak-aot: Which is the same effect from assigning the creator. Advantage is it will simplify the SQL. Let me know if there is a reason to not go that way.

sh16011993: To me, it is just that the new column should have a meaningful value. I am okay with either of them.

sh16011993: Also, if we go with the `system-user` there will be a distinction between the pre-existing records vs the ones that will come into the database after this code push since afterwards it will be the ministry user doing the update.

guru-aot (approved):
LGTM, nice work @sh16011993

sources/packages/backend/apps/db-migrations/src/sql/CASInvoices/Add-cas-invoice-status-updated-by-column.sql
```diff
@@ -9,14 +9,7 @@ COMMENT ON COLUMN sims.cas_invoices.invoice_status_updated_by IS 'User who updat
...
 SET
-  invoice_status_updated_by = (
-    SELECT
-      id
-    FROM
-      sims.users
-    WHERE
-      last_name = 'system-user'
-  );
+  invoice_status_updated_by = creator;
```
dheepak-aot, line 12: Please update the comment `-- Update the existing records with the system user.`

dheepak-aot (approved):
Thanks for the changes. Looks good 👍

One minor comment on the sql comment

PR #4485, opened 2025-03-17
#4458 - Cherry pick the release commit and resolve conflicts and update the solution

dheepak-aot:
## Cherry Pick release commit for application version bug

- [x] Brought the commit from `Release v2.6` for application versioning bug by resolving conflicts.
- [x] Updated the service to get application versions to use ORM relation `application -> parentApplication -> versions` to get the application versions from an `applicationId` through it's parent application.

## E2E Tests

- [x] Renamed E2E test as per our regular naming convention to use the method name of the controller `getApplicationOverallDetails`. 
- [x] Updated E2E tests, added additional scenario.

image


sources/packages/backend/libs/sims-db/src/entities/application.model.ts
```diff
@@ -415,6 +415,19 @@ export class Application extends RecordDataModel {
...
+   * if the application is the parent application.
+   */
+  @OneToMany(
+    () => Application,
+    (versionApplication) => versionApplication.parentApplication,
+    {
+      nullable: true,
+    },
+  )
+  versions?: Application[];
```
andrewsignori-aot, line 430: Since every application starts with a parent I would say that this is not nullable. Does it make sense?

andrewsignori-aot: If the nullable is due the moment when we filter the application itself, an empty array will be returned if I am not wrong.

sources/packages/web/package-lock.json
andrewsignori-aot, line 1: Is it a workspace config issue? We can check together if needed.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.aest.controller.getApplicationOverallDetails.e2e-spec.ts
```diff
@@ -0,0 +1,127 @@
...
+          firstVersionApplication.currentAssessment.offering.offeringIntensity,
+        submittedDate: addDays(-1),
+      },
+    );
+
+    const currentApplication = await saveFakeApplication(
+      db.dataSource,
+      {
+        parentApplication: { id: firstVersionApplication.id } as Application,
+        precedingApplication: { id: firstVersionApplication.id } as Application,
```
andrewsignori-aot, line 58: It will not change the test output but should it not be a different `precedingApplication` for this one?

andrewsignori-aot (commented):
Great work and thanks for doing the improvement while merging it back to the main.

andrewsignori-aot (approved):
Thanks for making the changes, look good 👍

andrepestana-aot (approved):
Looks good!

PR #4491, opened 2025-03-18
#4354 - Change Request Revamp - 1st effort - Warping up workflow during "in-progress" application

andrewsignori-aot:
- Created a variation for the existing worker "workflow-wrap-up" to allow its usage before the "end workflow" for the `assessment-gateway` "Applications In Progress" lane, as shown in items 1 to 3 in the below image (image 4 is working as before).
  - The worker is backward compatible with the existing workflow, which means that once deployed, workflows executing the previous version where the new header is not provided will be able to finish its process as before. The absence of the header makes it works as before.
  - _Note:_ fixed an issue when a `job.complete()` was returned from inside a transaction block with the intention of finishing the worker task but was actually finishing only the transaction block, leaving the rest of the worker processing to still be executed.

image

The workflow task now accepts a header to define its behavior, as shown below.

image

image

### DIFF
--- a/sources/packages/backend/apps/workers/src/controllers/assessment/_tests_/e2e/assessment.controller-workflowWrapUp.e2e-spec.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/assessment/_tests_/e2e/assessment.controller-workflowWrapUp.e2e-spec.ts
@@ -30,9 +30,11 @@ import {
 import {
   AssessmentSequentialProcessingService,
   SystemUsersService,
+  WorkflowClientService,
 } from "@sims/services";
 import { addDays, getISODateOnlyString } from "@sims/utilities";
 import { StudentAssessmentService } from "../../../../services";
+import { WorkflowWrapUpType } from "../../assessment.dto";
 
 describe("AssessmentController(e2e)-workflowWrapUp", () => {
   let db: E2EDataSources;
@@ -40,6 +42,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
   let systemUsersService: SystemUsersService;
   let studentAssessmentService: StudentAssessmentService;
   let assessmentSequentialProcessingService: AssessmentSequentialProcessingService;
+  let workflowClientService: WorkflowClientService;
   let now: Date;
   let auditUser: User;
 
@@ -52,6 +55,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
     assessmentSequentialProcessingService = nestApplication.get(
       AssessmentSequentialProcessingService,
     );
+    workflowClientService = nestApplication.get(WorkflowClientService);
     auditUser = systemUsersService.systemUser;
     now = new Date();
   });
@@ -74,10 +78,9 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
 
     // Act
     const result = await assessmentController.workflowWrapUp(
-      createFakeWorkflowWrapUpPayload(
-        savedApplication.currentAssessment.id,
+      createFakeWorkflowWrapUpPayload(savedApplication.currentAssessment.id, {
         workflowData,
-      ),
+      }),
     );
 
     // Asserts
@@ -163,7 +166,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
     const result = await assessmentController.workflowWrapUp(
       createFakeWorkflowWrapUpPayload(
         currentApplicationToWrapUp.currentAssessment.id,
-        workflowData,
+        { workflowData },
       ),
     );
 
@@ -210,7 +213,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
         currentAssessmentInitialValues: {
           assessmentWorkflowId: "some fake id",
           assessmentDate: new Date(),
-          studentAssessmentStatus: StudentAssessmentStatus.InProgress,
+          studentAssessmentStatus: StudentAssessmentStatus.Completed,
         },
       },
     );
@@ -223,7 +226,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
     // Monitor the updateAssessmentStatusAndSaveWorkflowData to ensure that it will be called.
     const updateAssessmentStatusAndSaveWorkflowDataMock = jest.spyOn(
       studentAssessmentService,
-      "updateAssessmentStatusAndSaveWorkflowData",
+      "updateAssessmentWrapUpData",
     );
     // Monitor the assessImpactedApplicationReassessmentNeeded to ensure that it will not be called.
     const assessImpactedApplicationReassessmentNeededMock = jest.spyOn(
@@ -235,7 +238,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
     const result = await assessmentController.workflowWrapUp(
       createFakeWorkflowWrapUpPayload(
         currentApplicationToWrapUp.currentAssessment.id,
-        workflowData,
+        { workflowData },
       ),
     );
 
@@ -285,10 +288,9 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
 
       // Act
       const result = await assessmentController.workflowWrapUp(
-        createFakeWorkflowWrapUpPayload(
-          savedApplication.currentAssessment.id,
-          {} as WorkflowData,
-        ),
+        createFakeWorkflowWrapUpPayload(savedApplication.currentAssessment.id, {
+          workflowData: {} as WorkflowData,
+        }),
       );
 
       // Asserts
@@ -344,10 +346,9 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
 
       // Act
       const result = await assessmentController.workflowWrapUp(
-        createFakeWorkflowWrapUpPayload(
-          savedApplication.currentAssessment.id,
-          {} as WorkflowData,
-        ),
+        createFakeWorkflowWrapUpPayload(savedApplication.currentAssessment.id, {
+          workflowData: {} as WorkflowData,
+        }),
       );
 
       // Asserts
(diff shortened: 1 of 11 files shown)


sources/packages/backend/apps/workers/src/controllers/assessment/_tests_/e2e/workflow-wrap-up-factory.ts
```diff
@@ -1,29 +1,37 @@
...
+} from "../../assessment.dto";
+import { IOutputVariables, ZeebeJob } from "@camunda8/sdk/dist/zeebe/types";
 
 /**
  * Creates a fake workflow wrap up payload.
  * @param assessmentId assessment id.
- * @param workflowData workflow data.
+ * @param options customize the payload.
+ * - `workflowData` workflow data.
+ * - `wrapUpType` wrap-up type, default {@link WorkflowWrapUpType.CompleteWrapUp} if not provided.
```
dheepak-aot, line 15: 👍

sources/packages/backend/apps/workers/src/controllers/assessment/_tests_/e2e/assessment.controller-workflowWrapUp.e2e-spec.ts
```diff
@@ -210,7 +213,7 @@ describe("AssessmentController(e2e)-workflowWrapUp", () => {
         currentAssessmentInitialValues: {
           assessmentWorkflowId: "some fake id",
           assessmentDate: new Date(),
-          studentAssessmentStatus: StudentAssessmentStatus.InProgress,
+          studentAssessmentStatus: StudentAssessmentStatus.Completed,
```
dheepak-aot, line 216: Is the status required to be `Completed` here?

andrewsignori-aot: Yes, it is required and yes, the idempotency relies on it now.
The E2E is testing a scenario where the workflow data is already set, which means that an `InProgress` status was not accurate.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -290,50 +290,61 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
     entityManager: EntityManager,
+    options?: { workflowData?: WorkflowData },
   ): Promise<boolean> {
     const studentAssessmentRepo =
       entityManager.getRepository(StudentAssessment);
-    const auditUser = this.systemUsersService.systemUser;
-    const now = new Date();
     const studentAssessment = await studentAssessmentRepo.findOne({
-      select: { studentAssessmentStatus: true, workflowData: true as unknown },
+      select: { studentAssessmentStatus: true },
```
dheepak-aot, line 306: Not part of the PR. But selecting id here would make the query(sql) better as we have observed.

andrewsignori-aot: Just to ensure that we are on the same page, related to SQL/DB, the query is better without the ID since it is not required.
From Typeorm's perspective, it may fail for child objects when the parent does not have a single property as part of the select, and that leads us to include the ID in the parent object sometimes.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -290,50 +290,61 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
...
+      StudentAssessmentStatus.CancellationRequested,
+      StudentAssessmentStatus.CancellationQueued,
+      StudentAssessmentStatus.Cancelled,
+    ].includes(studentAssessment.studentAssessmentStatus);
+    const auditUser = this.systemUsersService.systemUser;
+    const now = new Date();
+    if (workflowIsCancelled) {
+      if (!options?.workflowData) {
+        // No update is executed because the data was already present or the workflow was cancelled.
+        return false;
```
dheepak-aot, line 319: May or may not be part of the PR. But this logic should be part of the controller rather than service.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -261,52 +267,67 @@ export class AssessmentController {
         jobLogger.error(message);
         return job.error(ASSESSMENT_NOT_FOUND, message);
       }
+      jobLogger.log(
+        `Workflow wrap-up type requested: ${job.customHeaders.wrapUpType}.`,
```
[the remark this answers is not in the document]
andrewsignori-aot: Yes, it will. The backward compatibility will be achieved where it matters. The workflow needs to be backward-compatible enough to allow any in-flight assessment to be finished. Any new assessment executed once this is deployed will start using the header independently if it is a change request or not.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -261,52 +267,67 @@ export class AssessmentController {
...
+          if (impactedApplication) {
+            jobLogger.log(
+              `Application id ${impactedApplication.id} was detected as impacted and will be reassessed.`,
+            );
+          }
+        },
+      );
+      if (jobActionAcknowledgement) {
+        // If there is a job result to be returned, the worker should finalize its processing.
+        return jobActionAcknowledgement;
```
dheepak-aot, line 329: 👍

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
andrewsignori-aot: It may, but the controller is already executing too much and I was not focusing on going further in any refactoring.
The controller logic should be segregated already in at least two other methods but I have no intention to address this as part of this ticket goal 😉

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.dto.ts
```diff
@@ -160,6 +160,26 @@ export interface WorkflowWrapUpJobInDTO {
...
+  /**
+   * Executes only a subset of the wrap-up operations, setting the assessment status to completed.
+   * Used when the workflow does not reach the calculations stage and must be aborted.
+   */
+  AssessmentStatusOnly = "Assessment status only",
+  /**
+   * Executes all wrap-up operations, setting the assessment status to completed and
+   * executing other processes related to sequential processing.
+   */
+  CompleteWrapUp = "Complete wrap-up",
```
dheepak-aot, line 176: Minor, it could be just `Complete` as it belongs to wrap up type.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
```diff
@@ -923,6 +919,48 @@ Note: the workflow process should continue its execution as a regular reassessme
     <bpmn:sequenceFlow id="change-in-progress-no-pir-required-flow" name="Change is in progress" sourceRef="pir-change-request-gateway" targetRef="Gateway_04uphc0">
       <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=applicationEditStatus = "Change in progress"</bpmn:conditionExpression>
     </bpmn:sequenceFlow>
+    <bpmn:sequenceFlow id="Flow_0jltxxn" sourceRef="change-request-workflow-wrap-up-task" targetRef="minitry-approval-change-declined-end-event" />
+    <bpmn:serviceTask id="change-request-workflow-wrap-up-task" name="Workflow wrap up (Assessment status only)">
+      <bpmn:extensionElements>
+        <zeebe:taskDefinition type="workflow-wrap-up" />
+        <zeebe:taskHeaders>
+          <zeebe:header key="wrapUpType" value="Assessment status only" />
```
dheepak-aot, line 927: It could be a variable initialized at start of workflow and used across many places. 

image

dheepak-aot:
Great way to wrap up the workflow ends. Added few comments. Please have a look.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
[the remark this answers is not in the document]
dheepak-aot: I get the intention more precisely when the in-flight instances are mentioned. I am good here.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.dto.ts
andrewsignori-aot: Changed.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
dheepak-aot: But what is mentioned above is not about usage of headers. It is about using a value from variable instead of repeating the value. Let me know if I am missing something here.

sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
dheepak-aot: I meant, from what I remembered in the past if we do not select the id, typeorm was creating a SQL select with id and also the selected properties(from find/findone) and was pulling the selected properties from that query output. 

I am not sure if that was the right understanding, but it looks like typeorm always uses id irrespective if it being used in select or not. 

e.g. 

```ts
await db.studentAssessment.findOne({
      select: { id: true, studentAssessmentStatus: true },
      where: { id: application.currentAssessment.id },
    });
```

results in 

```sql
select
	"StudentAssessment"."student_assessment_status" as "StudentAssessment_student_assessment_status",
	"StudentAssessment"."id" as "StudentAssessment_id"
from
	"sims"."student_assessments" "StudentAssessment"
where
	(("StudentAssessment"."id" = $1))
limit 1
```

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway.bpmn
andrewsignori-aot: I got it now that it is related to a broader conversation about the enum usage in the workflow and a possible way to have an enum-like structure type in BPMN. I will take a look outside this PR effort. Thanks for the chat.

dheepak-aot (approved):
Thanks for making the change and discussions. Looks good 👍

sh16011993 (approved):
Great work @andrewsignori-aot 👍

Issue #4454, opened 2025-03-10
Security Patching


ninosamson:
Review and resolve the findings from automated scanning for security vulnerabilities. 

- [x] https://github.com/bcgov/SIMS/security/dependabot 
- [x] https://github.com/bcgov/SIMS/security/code-scanning 
- [x] ensure axios is updated to most recent version in all packages 
- [x] time boxed effort - resolve all the High sev issues and cleanup common issues before moving on to complex, MED sev issues.

PR #4536, opened 2025-03-26
#4454 - Security Patching

guru-aot:
- Code scanning vulnerabilities are changed accordingly.
- For dependabot, packages like backend, forms and web security vulnerabilities of axios, soap, vite are installed.
- The only other package-lock.json is on the root folder, i am not sure why that is getting updated without a package.json and we can remove it. As a part of the review, please suggest if we can remove it.


sources/packages/backend/libs/integrations/src/services/atbc/atbc.service.ts
```diff
@@ -32,7 +32,7 @@ export class ATBCService {
       // (NOTE: this will disable client verification)
       // TODO: add certificate for PROD
       httpsAgent: new (require("https").Agent)({
-        rejectUnauthorized: false,
+        rejectUnauthorized: true,
```
andrewsignori-aot, line 35: This should be the default option; hence, it can probably just be removed.

sources/packages/backend/libs/integrations/src/services/atbc/atbc.service.ts
andrewsignori-aot, line 1: Can we add the below comment at the top of this service to be clear that we should not invest time on this and the plan is to have it removed, please?
```ts
/**
 * @obsolete this service is no longer in use and should be removed.
 * If not removed, should be refactored to use the new http client.
 */
@Injectable()
export class ATBCService {
....
````

andrewsignori-aot (commented):
Great work, and please remove the mentioned `package-lock.json` from the root folder that is probably a leftover from before the monorepo conversion.

andrewsignori-aot (approved):
Looks good 👍 
Once this one is merged please ensure that we will have actions for remaining open items, for instance, to upgrade Nestjs if required to have some dependent package updated.

bidyashish (approved):
LGTm @guru-aot

Issue #4459, opened 2025-03-11
Set max documents for application uploaders to 25


Joshua-Lakusta:
As a ministry user I want there to be a limit of the number of documents that can be attached to an exception document uploader.

**Acceptance Criteria**
- [x] Change the max number of documents for all application based uploaders in the student application to 25 documents.
- [x] If they remove documents they can replace those documents up to the consistent total of 25
- [x] Ensure the minimum for each uploader is one file and that they are all mandatory fields.
- [x] All program years
- [x] For manual uploader outside of the application each individual upload submission should be limited to 25 files but they can make as many submission of up to 25 files as they want
- [x] Remove the optional tag when it is present on any of the warning message content for application document uploaders

- [x] Always have the same content message for min of 1 file and max of 25 i

> You must upload at least one file, and the total number of uploaded files cannot exceed 25.

**Technical ACs:**
- [ ] Same as the ACs above.

**Additional Context**
One of the current exceptions, partener estimated current year income is only allowing one document and this raised that fact that we should have a consistent criteria and limit for all exceptions. This will resolve the issue with that one at the same time.

Joshua-Lakusta:
I set this to asset as part of this is resolving a current issue that is blocking students from submitting their info for one exception. Happy to discuss in shower stall

Joshua-Lakusta:
Added two ac's to clarify behavior when a file is removed in a app based uploader and one to explain the behavior for the manual uploader after discussion with dev working ticket.

PR #4537, opened 2025-03-26
#4459 - Set max documents for application uploaders to 25

andrepestana-aot:
- Set max number of documents to 25. Added a hidden component in each form to have it server calculated;
- All fileUploaders should be required to have at least 1 file;
- Changed some fileUploaders to allow multiple files (they allowed only one file to be uploaded);
- Removed "optional" of some labels;
- Kept same custom validation to all fileUploaders so it's easy to change it all occurrences across the forms:
```json
"validate": {
                "required": true,
                "customMessage": "You must upload at least one file, and the total number of uploaded files cannot exceed {{data.maxUploadedFiles}}.",
                "custom": "const files = data[instance.key] || []; valid = files.length > data.maxUploadedFiles ? false : true;"
              }
```

image
image

image
image

### DIFF
--- a/sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
+++ b/sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
@@ -141,6 +141,27 @@
       "lockKey": true
     },
     {
+      "input": true,
+      "tableView": true,
+      "key": "maxUploadedFiles",
+      "label": "maxUploadedFiles",
+      "protected": false,
+      "unique": false,
+      "persistent": false,
+      "type": "hidden",
+      "tags": [],
+      "conditional": {
+        "show": "",
+        "when": null,
+        "eq": ""
+      },
+      "properties": {},
+      "defaultValue": "",
+      "lockKey": true,
+      "calculateValue": "value = 25;",
+      "calculateServer": true
+    },
+    {
       "label": "HTML",
       "labelWidth": "",
       "labelMargin": "",
@@ -1689,8 +1710,8 @@
               "dir": "Current Year Income - Supporting Documents",
               "customClass": "font-weight-bold",
               "validate": {
-                "required": false,
-                "custom": "const maxUploadedFiles = 25; const files = data[instance.key] || []; valid = !files.length ? 'At least one file is required' : files.length > maxUploadedFiles ? `Maximum ${maxUploadedFiles} files allowed` : true;"
+                "required": true,
+                "custom": "const files = data[instance.key] || []; valid = files.length > data.maxUploadedFiles ? `Maximum ${data.maxUploadedFiles} files allowed` : true;"
               },
               "lockKey": true
             },
--- a/sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
+++ b/sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
@@ -1711,7 +1711,8 @@
               "customClass": "font-weight-bold",
               "validate": {
                 "required": true,
-                "custom": "const files = data[instance.key] || []; valid = files.length > data.maxUploadedFiles ? `Maximum ${data.maxUploadedFiles} files allowed` : true;"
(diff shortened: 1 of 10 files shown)


sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
```diff
@@ -1689,7 +1689,8 @@
               "dir": "Current Year Income - Supporting Documents",
               "customClass": "font-weight-bold",
               "validate": {
-                "required": true
+                "required": false,
+                "custom": "const maxUploadedFiles = 25; const files = data[instance.key] || []; valid = !files.length ? 'At least one file is required' : files.length > maxUploadedFiles ? `Maximum ${maxUploadedFiles} files allowed` : true;"
```
dheepak-aot, line 1693: Asking for my understanding. What happens when we enable the required and also have custom validation in place? 

I am asking this because, removing the `required` removes the `*` in the component.
image

dheepak-aot, line 1693: The `maxUploadedFiles` can be a hidden(calculated) variable injected into the form from Vue and API and used so that we can centralize the config. If there is a considerable amount of changes to achieve that, then at bare-minimum there must be hidden variable (calculated) assigned with value as 25 at each form that can be re-used.

dheepak-aot: Is business ok with `*` not being there for mandatory field?

andrewsignori-aot: This change can be improved in many ways. For instance, sharing the configuration from the API and having it inject into all forms that require it.
From API, I would refrain from doing the "injection" before dry run in every form, and I would add this ability to the centralized form.io dryRun API call.
That said, it was my understanding that we could take the easiest and simple route to finish this ticket to later do an improvement. 
Having the changes exactly the same already helps, and we can have a tech debt to have "File Upload Validations Centralization for UI and API", which can also include the max size for files that is also hard-coded on every form (if I am not wrong).
My suggestion, keep it simple and create a tech debit.

bidyashish (approved):
Changes looks good @andrepestana-aot 

One thought, I can see code is being repeated a lot, can it be defined once and called multiple times to avoid repetation!

sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
andrepestana-aot: I added a hidden field to calculate on server.

sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
```diff
@@ -1689,7 +1710,9 @@
               "dir": "Current Year Income - Supporting Documents",
               "customClass": "font-weight-bold",
               "validate": {
-                "required": true
+                "required": true,
+                "customMessage": "You must upload at least one file, and the total number of uploaded files cannot exceed {{data.maxUploadedFiles}}.",
+                "custom": "const files = data[instance.key] || []; valid = files.length > data.maxUploadedFiles ? false : true;"
```
andrewsignori-aot, line 1715: Minor and not a blocker, can we check if one of the below could be used instead of `data[instance.key]`?

image

andrewsignori-aot (approved):
Thanks for looking for alternatives for the form.io limitations and for getting the answer from biz. Looks good 👍

sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
andrepestana-aot: Business is ok to have one message for both "required" and custom then I added a `customMessage` and both asterisk and max number of files can be in place.

sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
andrepestana-aot: For this component input seems to be every item when they are uploaded and value is undefined:
image

PR #4542, opened 2025-03-28
#4459 - Set max documents for application uploaders to 25 (#4537) cherry pick

bidyashish:
- Set max number of documents to 25. Added a hidden component in each form to have it server calculated;
- All fileUploaders should be required to have at least 1 file;
- Changed some fileUploaders to allow multiple files (they allowed only one file to be uploaded);
- Removed "optional" of some labels;
- Kept same custom validation to all fileUploaders so it's easy to change it all occurrences across the forms:
```json
"validate": {
                "required": true,
                "customMessage": "You must upload at least one file, and the total number of uploaded files cannot exceed {{data.maxUploadedFiles}}.",
                "custom": "const files = data[instance.key] || []; valid = files.length > data.maxUploadedFiles ? false : true;"
              }
```

image

image

image

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
+++ b/sources/packages/forms/src/form-definitions/partnerinformationandincomeappeal.json
@@ -141,6 +141,27 @@
       "lockKey": true
     },
     {
+      "input": true,
+      "tableView": true,
+      "key": "maxUploadedFiles",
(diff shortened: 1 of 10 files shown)


andrepestana-aot (approved):
Looks good!

ninosamson:
I would just update the title so we know this was the cherry pick into 2.7 @bidyashish

Issue #2822, opened 2024-02-13
Create readiness probe and liveness probe for API


guru-aot:
Acceptance Criteria

- [x] Have the implemented health check created from #1580 and create readiness and liveliness probe for API
- [ ] Ensure health check api is not public

PR #4547, opened 2025-03-28
#2822 - API Readiness Liveliness

guru-aot:
- Updated the health check of API to check for the sims-db database ping.
- The public expose of the health check https://github.com/bcgov/SIMS/blob/dddb903701e8dda6c761ef74a5f8bda328c10d78/sources/packages/backend/apps/api/src/main.ts#L31 is already done, so no changes required in this PR.

### DIFF
--- a/devops/openshift/api-deploy.yml
+++ b/devops/openshift/api-deploy.yml
@@ -234,7 +234,7 @@ objects:
                   port: "${{PORT}}"
                 initialDelaySeconds: 30
                 periodSeconds: 30
-                timeoutSeconds: 1
+                timeoutSeconds: 3
               livenessProbe:
                 failureThreshold: 3
                 successThreshold: 1
@@ -244,7 +244,7 @@ objects:
                   scheme: HTTP
                 initialDelaySeconds: 30
                 periodSeconds: 30
-                timeoutSeconds: 1
+                timeoutSeconds: 3
   - apiVersion: v1
     kind: Service
     metadata:
--- a/devops/openshift/api-deploy.yml
+++ b/devops/openshift/api-deploy.yml
@@ -232,9 +232,9 @@ objects:
                 httpGet:
                   path: /health/
                   port: "${{PORT}}"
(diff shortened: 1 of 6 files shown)


sources/packages/backend/apps/api/src/route-controllers/health-check/health.controller.spec.ts
```diff
@@ -17,10 +18,7 @@ describe.skip("HealthController", () => {
 
   describe("root", () => {
     it("should return Hello world string with db connection status and version", () => {
```
sh16011993, line 20: We can update this.

sh16011993 (approved):
Great work @guru-aot 👍

devops/openshift/api-deploy.yml
```diff
@@ -232,19 +232,19 @@ objects:
...
                 successThreshold: 1
                 httpGet:
                   path: /health/
                   port: "${{PORT}}"
                   scheme: HTTP
-                initialDelaySeconds: 10
+                initialDelaySeconds: 30
                 periodSeconds: 30
-                timeoutSeconds: 15
+                timeoutSeconds: 1
```
andrewsignori-aot, line 247: Can we follow the same timeouts from queue-consumers on this one?
The DB check timeout is already 1s, which means that it could take up to 1s to reply.
Please make it 3s for both probes.

sources/packages/backend/apps/api/src/route-controllers/health-check/health.controller.spec.ts
andrewsignori-aot, line 1: I will assume this file is an outdated unit test file, not directly part of the review, and the changes here were added only to allow the tests to pass.
This file should be refactored in the future.

andrewsignori-aot (approved):
Nice work, please adjust the timeout as mentioned in [this comment](https://github.com/bcgov/SIMS/pull/4547#discussion_r2021660338) only.

Issue #4358, opened 2025-02-14
Content: Student: Institution is not Listed


AnnaPBashkatova:
**As a** *Student* **I want** *to have clear instructions on what steps to take if I can not find my institution while I am trying to submit my Financial Aid Application* **So That** *I take correct steps.*

**Acceptance Criteria**
- [x] Information need to be updated in  both PT and FT across all PY.
- [x] When Student checks "Your Institution is not listed checkbox", the following message is displayed: 
**Header:**  **Your institution is currently unable to process applications in the StudentAid BC system**
**Body:** Your institution is currently unable to process applications in the StudentAid BC system. Please have your institution contact designat@gov.bc.ca to inquire about the process for gaining access to our Student Information Management System (SIMS).

Current Behavior: 
<img width="1605" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1BoQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--40b50f91754ba98c1b77c9ec068761c54d54189a/image.png" alt="image.png" />

**Technical ACs:**

- [x] Update the banner in all the forms as per the expected behavior below: 

Expected Behavior: 
<img width="1158" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3JoQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8d658c6e17f67e3eb694e2a4d688da05922dfef9/image.png" alt="image.png" />

**Additional Context**
- The current message sends student and institution on the wrong path. **Important to fix prior to FT launch, as it will create a high volume of calls and confusion among institutions who are not yet designated.**

PR #4555, opened 2025-03-31
#4358 - Content: Student: Institution is not Listed

lewischen-aot:
- Information updated in both PT and FT across all PY.
- Updated the banner in all the forms as per the expected behavior as shown in the AC's.

Screenshot
Screenshot 2025-03-31 125023


dheepak-aot (approved):
Looks good @lewischen-aot  👍

andrepestana-aot (approved):
Looks good!

Issue #4549, opened 2025-03-28
BUG: Unexpected error: "Complete student request" for ministry user does not confirm change request


astridSABC:
**Describe the Bug**
Ministry user is not able to approve a Change Request

**Expected Behaviour**
When the ministry selects "yes", adds a note then clicks "complete student request" there is a confirmation toast message that the change request is successful.  

**Actual Behaviour**
Message "An unexpected error happened during the approval." is displayed.  The change request does not happen

<img width="506" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeGw0Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ab25240cea51f9af120a8a438002be19849b9226/image.png" alt="image.png" />

### Steps To Reproduce

Precondition: Student has a completed application and has submitted a 'request for change"

1. Login to the ministry portal
2. Click "Change requests" in the student requests section on the left pane
3. Search for the student who has requested a request for change
4. Click "view" beside the student to view the request
5. Select "view request" in the unapproved changes table
6. Select "yes"
7. Add a note
8. Click "Complete student request"

TC: 4402

PR #4557, opened 2025-03-31
#4549 - Bugfix on approving the student appeal

dheepak-aot:
## Root cause and solution
- Root cause: When the student appeal is retrieved from the database along with the relations, the offering object was not returned and hence  the reference of `offering.id` resulted in error.
- Solution: Added a join with offerings in the SQL.

## E2E Tests
- [x] Added an E2E Test for the approval of student appeal with appeal request.

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.aest.controller.approveStudentAppealRequests.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.aest.controller.approveStudentAppealRequests.e2e-spec.ts
@@ -66,7 +66,7 @@ describe("StudentAppealAESTController(e2e)-approveStudentAppealRequests", () => 
       .auth(token, BEARER_AUTH_TYPE)
       .expect(HttpStatus.OK)
       .expect({});
-
+    // Check for the appeal and appeal requests in the database.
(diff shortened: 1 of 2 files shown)


andrewsignori-aot (approved):
Great finding, thanks for adding the E2E, looks good 👍

sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.aest.controller.approveStudentAppealRequests.e2e-spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      .expect({});
+    // Check for the appeal and appeal requests in the database.
+    const updatedAppeal = await db.studentAppeal.findOne({
+      select: {
+        id: true,
+        appealRequests: {
+          id: true,
+          appealStatus: true,
+          assessedBy: { id: true },
+          note: { description: true },
```
sh16011993, line 77: Should we not select `id` as well here?

dheepak-aot: As I would not be able to assert the id and the query being in E2E test and it works without the id, I did not select the id.

sh16011993 (approved):
Nice work in addressing the bug @dheepak-aot 👍

Issue #4489, opened 2025-03-17
FormIO Changes - Split FT/PT within PY (1/2)


CarlyCotton:
**User Story**
As a ministry user, I would like the financial aid application (currently split by program year) to be split between Full-time and Part-time within each program year. This will allow a separation of logic to make each form more streamlined and to make it easier for BAs to make changes to it in the future.

**Acceptance Criteria**
- [ ] add offering intensity selection to "Apply for Funding" page before "Program Year" selection
    - [ ] This should still be limited based on environment variable currently in use for this question in the form.
- [ ] save the offering intensity alongside application to ensure dry run validation and adjust any backend code (not application form) relying on `howWillYouBeAttendingTheProgram`
- [ ] populate the intensity into the application form dropdown as label
- [ ] mockup below just for reference, follow application patterns
- [ ] Update the dropdown labels for intensity to include the dash:
    - [ ]  Part time _should be_ Part-time
    - [ ]  Full time _should be_ Full-time

---
**Additional Context**
<img width="1159" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHhXQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a138b866be78895896c8f3d35754affcfc8af14d/image.png" alt="image.png" />

CarlyCotton:
FYI @sh16011993 
Updated to add an AC re intensity dropdown labels.

PR #4564, opened 2025-04-02
#4489 - Add Offering Intensity Column Migration Update

sh16011993:

- Migration `AddOfferingIntensityColumn` was updated to add a `Part Time` default to the column to ensure that the pre-existing application data which is saved as a `Draft` without the `data ->> 'howWillYouBeAttendingTheProgram'` being population gets an `offering intensity` value and that would be `Part Time`.

**Migration Revert Screenshot:**

image


sources/packages/backend/apps/db-migrations/src/sql/Applications/Add-offering-intensity-column.sql
```diff
@@ -13,4 +13,11 @@ SET
     applications.data ->> 'howWillYouBeAttendingTheProgram'
   ) :: sims.offering_intensity
 WHERE
-  offering_intensity IS NULL;
\ No newline at end of file
+  offering_intensity IS NULL
```
andrewsignori-aot, line 16: Because now we have the `DEFAULT`, we should remove this condition.

sources/packages/backend/apps/db-migrations/src/sql/Applications/Add-offering-intensity-column.sql
```diff
@@ -13,4 +13,10 @@ SET
     applications.data ->> 'howWillYouBeAttendingTheProgram'
   ) :: sims.offering_intensity
 WHERE
-  offering_intensity IS NULL;
\ No newline at end of file
+  data ->> 'howWillYouBeAttendingTheProgram' = 'Full Time';
```
andrewsignori-aot, line 16: The query can be simplified as below now.
```sql
UPDATE
  sims.applications applications
SET
  offering_intensity = 'Full Time'
WHERE
  data ->> 'howWillYouBeAttendingTheProgram' = 'Full Time';
```

andrewsignori-aot (approved):
Thanks for the investigation and the fix. Looks good 👍 
Please be sure to execute a manual rollback and migration execution once this one is deployed to DEV.

bidyashish (approved):
LGTM

Issue #4431, opened 2025-03-05
Student View change request revamp


Joshua-Lakusta:
As a student I need to be able to start an appropriate change request for my program year, see my change requests and the status of it while I am waiting for it, and see historical changes. This is important so that I can manage my application.

**Acceptance Criteria**

## Applications page changes
- [x] Actions will be expanded to the following buttons View, Edit, Change Request, and Cancel
- [x] No change to the edit functionality. it will continue to only be available till COE is completed and disappear.
- [x] Change request action should appear once confirmation of enrollment has been completed
- [x] Change request should not be available if a cancellation has been completed
- [x] The Change Request should be a link but should trigger one of two paths that will be discussed later in this ticket
- [x] Have the submitted date appear under the submitted column (#3 image below) This is the date of the submission of the original application
- [x] Add the View button (#2 image below) that will always exist and will have the same function as clicking on Financial Aid Application (#1 image below)
- [x] The order for the options should always be the following View, Edit/Change Request, and Cancel
- [x] No changes to the behavior of the cancel option (please ignore how there is a cancel option on a completed app in the image)
- [x] Make the current behavior when a student does not have a valid SIN the edit, cancel buttons should be disabled still
- [x] When there is not a valid SIN the change request button should be disabled if the logic says it should appear
- [x] Remove the Name column from the applications screen

<img width="1503" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM285Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--06727c8ff0ef21717902d3c2a4ec661c7cd4f254/image.png" alt="image.png" />

- [x] The change request function should be disabled when application is archived

## Tracker Card
- [x] When change request is active a tracker card will exist on the current tracker card list

- [x] This tracker card will show the outstanding stage of the change request that will adapt based on the outstanding information

- [x] Base content for the tracker card

> You have a submitted change request that is still pending. Please see below for the next steps.
> 
> Currently your change request is waiting for:
> 

>    Waiting on StudentAid BC to approve the change.
>    Pending parent declaration information.
>    Pending partner declaration information.
>    Pending student income verification information.
>    Pending parent income verification information.
>    Pending partner income verification information.

- [x] The Pending parent declaration information will only appear when the change is waiting for supporting user data for a parent.
[...]

PR #4565, opened 2025-04-02
#4431 - Student View change request revamp - Application Submission

andrewsignori-aot:
## API-related changes
- Created a new endpoint `@Post(":applicationId/change-request")` to allow the creation of the new application change request.
- Application Controller Refactor
  - `validateApplicationSubmission` method created to centralize the application submissions validations and get the expected application data to be saved.
  - Method `validateSubmitApplicationData` moved from the controller to the controller service.
  
## Queue-consumers-related changes
- Adapted the method that selects the workflows to be started to also consider edited applications with change requests.

## Web-related changes
- Adapted the current student application component to handle the submission of a change request. There aren't clear content directions right now, and the goal is to allow the submission of a change request, keeping most of the validations in place.
- Created a new route speficic for the new appplication.
- The logic to display an application as read-only was adapted as a sequence of `if/else` to improve readability. if there is a better way, please share 😉 
- Submitting a change request redirects to the application details with the intention to show the tracker card to be implemented in the upcoming PR. There are no directions on it right now.
- The UI is currently preventing the user from submitting a change request when one is already in progress. No further UI direction has been given so far.

### Current UI for change request

image

_Note:_ E2E tests are yet to be included.


sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -775,4 +783,120 @@ export class ApplicationControllerService {
...
+   * @throws UnprocessableEntityException if the program year is not active.
+   * @throws BadRequestException if the application data is invalid, failing the druRun submission.
+   */
+  async validateApplicationSubmission(
+    applicationId: number,
+    studentId: number,
+    payload: SaveApplicationAPIInDTO,
+  ): Promise<ApplicationData> {
+    const application =
+      await this.applicationService.getApplicationForSubmissionValidation(
```
dheepak-aot, line 801: 👍

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -775,4 +783,120 @@ export class ApplicationControllerService {
...
     }
   }
+
+  /**
+   * Validate the application submission and get the validated result.
+   * @param applicationId application ID.
+   * @param studentId student used for authorization.
+   * @param payload payload to create the application.
+   * @throws UnprocessableEntityException if the program year is not active.
+   * @throws BadRequestException if the application data is invalid, failing the druRun submission.
```
dheepak-aot, line 793: As we are listing the potential exceptions here, what about NotFoundException?

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -775,4 +783,120 @@ export class ApplicationControllerService {
...
+        payload.data,
+      );
+    if (!submissionResult.valid) {
+      throw new BadRequestException(
+        "Not able to create an application due to an invalid request.",
+      );
+    }
+    await this.offeringIntensityRestrictionCheck(
+      studentId,
+      submissionResult.data.data.howWillYouBeAttendingTheProgram,
```
dheepak-aot, line 825: Is expected to come from `application.offeringIntensity` from @sh16011993 PR ?

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -775,4 +783,120 @@ export class ApplicationControllerService {
...
+    }
+
+    // studyStartDate from payload is set as studyStartDate
+    let studyStartDate = payload.data.studystartDate;
+    let studyEndDate = payload.data.studyendDate;
+    if (payload.data.selectedOffering) {
+      const offering = await this.offeringService.getOfferingById(
+        payload.data.selectedOffering,
+        { programId: payload.data.selectedProgram },
+      );
```
dheepak-aot, line 867: Not part of the PR or may be part of @sh16011993 work, we should also check the match for Application `Offering Intensity` vs `Offering Intensity` from actual offering right?

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: I was planning to list them all, but I am not sure if we should do it as a pattern, to be honest.
I believe that I will remove them all. Please let me know if there is a concern about keeping or removing it.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -326,30 +236,99 @@ export class ApplicationStudentsController extends BaseController {
...
+  @ApiBadRequestResponse({
+    description:
+      "Form validation failed or offering intensity type is invalid.",
+  })
+  @ApiNotFoundResponse({
+    description:
+      "Application not found or it is not in the correct status to be changed.",
+  })
+  @ApiForbiddenResponse({
+    description: "The student has a restriction on his account.",
```
dheepak-aot, line 293: Minor. `The student has a restriction.` would be sufficient and more generic.

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: I can extend this validation to also include it.

sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
andrewsignori-aot: It was changed from the current submission that state "You have a restriction on your account.".
I believe the "The student has a restriction on his account." is generic and informative enough 😉

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -307,6 +303,151 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    if (
+      applicationData.selectedLocation !== application.data.selectedLocation
+    ) {
+      throw new CustomNamedError(
+        "Change request has a different location from its original submission.",
+        APPLICATION_NOT_VALID,
+      );
+    }
+    if (
+      applicationData.selectedOffering !== application.data.selectedOffering
```
dheepak-aot, line 368: What happens if the application submitted previously is requested for PIR? `undefined === undefined ` and it works?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -307,6 +303,151 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    ) {
```
dheepak-aot, line 369: Also I have a question about PIR scenario. In this case the selected study dates and optionally program name and desc are entered in student application. And if a change request is submitted for the same application, now we have a room to manipulate the PIR related data, even though it would not impact the output of the application in terms of offering and assessment .etc. But still there is a room for the PIR information to be manipulated and displayed incorrectly right? Is business ok with that?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -307,6 +303,151 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      // Applications are expected to have the same offering value populated, even for PIRs.
+      // PIRs will not have the offering value populated and the assessment will be created
+      // based on the most recent offering associated with the current application version.
+      throw new CustomNamedError(
+        "Change request has a different offering from its original submission.",
+        APPLICATION_NOT_VALID,
+      );
+    }
+
```
dheepak-aot, line 378: I believe we are missing the check for application to be archived unless there is a reason that I am missing. Current change request process validates archived applications.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -307,6 +303,151 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    const originalAssessment = new StudentAssessment();
+    originalAssessment.triggerType = AssessmentTriggerType.OriginalAssessment;
+    originalAssessment.submittedBy = auditUser;
+    originalAssessment.submittedDate = now;
+    originalAssessment.creator = auditUser;
+    originalAssessment.offering = {
+      id: application.currentAssessment.offering.id,
+    } as EducationProgramOffering;
+    if (application.currentAssessment.studentAppeal) {
+      originalAssessment.studentAppeal = {
```
dheepak-aot, line 391: what about `originalAssessment.studentAppeal = application.currentAssessment.studentAppeal `?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -307,6 +303,151 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+    // and the assessment cannot be create at the same moment what causes a
+    // "cyclic dependency error" on Typeorm. Saving the application record and later
+    // having it associated with the assessment solves the issue.
+    await this.dataSource.transaction(async (transactionalEntityManager) => {
+      const applicationRepository =
+        transactionalEntityManager.getRepository(Application);
+      await applicationRepository.save(newApplication);
+      newApplication.modifier = auditUser;
+      newApplication.updatedAt = now;
+      newApplication.studentAssessments = [originalAssessment];
```
dheepak-aot, line 441: Minor doubt for my clarification with the observation of current edit process also follows same code, both `studentAssessments` and `currentAssessment` have cascade enabled so wondering the chances for 2 different inserts of assessment if that is possible.

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: Removed them for now.

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: Adjusted now.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -307,6 +303,151 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      );
+    }
+    // Check if there is already a change request in progress.
+    const hasChangeInProgress = application.parentApplication.versions.some(
+      (version) =>
+        [
+          ApplicationEditStatus.ChangeInProgress,
+          ApplicationEditStatus.ChangePendingApproval,
+        ].includes(version.applicationEditStatus),
+    );
```
dheepak-aot, line 348: As the versions are joined based on same criteria, we can either check only for versions or remove the criteria in the query.

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: Added the new offering intensity validation.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: The expectation is that it must work. So far, the direction is that the user will not be able to change the location/offering, which means the application previously submitted as PIR will continue to be a PIR.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: The mentioned room for manipulation exists because we do not have a final direction about how it will be handled.
For instance, if the application form presented to the student during the change request does not display these fields (or display them as non-persistent read-only), the result of the `dryRun` will not persist them. Does it make sense?
The intention of adding some validations right now was to try to ensure some critical data (e.g., location/offering/PIR) was not updated.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: There are no requirements for it.

sources/packages/web/src/router/StudentRoutes.ts
```diff
@@ -98,6 +98,20 @@ export const studentRoutes: Array<RouteRecordRaw> = [
...
       },
+      {
+        path: AppRoutes.StudentApplicationChangeRequest,
+        name: StudentRoutesConst.DYNAMIC_FINANCIAL_CHANGE_REQUEST_APP_FORM,
+        component: DynamicStudentApp,
+        props: (route: RouteLocationNormalized) => ({
+          id: Number(route.params.id),
+          programYearId: Number(route.params.programYearId),
+          selectedForm: route.params.selectedForm as string,
+          changeRequest: true,
```
dheepak-aot, line 109: is `changeRequest` a param in the route?

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: I would like to keep both independently checking it, which would allow the query (that is already shared) to be expanded in the future without the need to impact this method.
If the query was dedicated to this method only, I would be open to it. Since it is a shared query, it would introduce an unnecessary liability.

sources/packages/web/src/router/StudentRoutes.ts
andrewsignori-aot: It is not a parameter in the URL, it is a parameter provided to the component associated with the route.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -3,7 +3,9 @@
     <template #header>
       <header-navigator
         title="Applications"
-        subTitle="Financial Aid Application"
+        :sub-title="
+          changeRequest ? 'Change request' : 'Financial Aid Application'
```
dheepak-aot, line 7: 👍

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: It was the initial thought and than I changed relying on the fact that the other references were created using only the `id`, which seems safer then realying on the fact the query will return a more raw `studentAppeal` which would contains only the id and nothing else selected in a way that could interfere with the saving operation.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: I did not follow the question. Can you please elaborate on it?
Is it about avoiding calling `save` twice and using inserts instead?

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -182,13 +191,26 @@ export default defineComponent({
...
+
+      // Define if the form should be read-only.
+      if (props.readOnly) {
+        // If some value is provided in the read-only prop, then the form should be readonly.
+        isReadOnly.value = true;
+      } else if (
+        // If a change request is being made in the expected application status,
+        // then the form should be editable.
+        props.changeRequest &&
+        applicationData.applicationStatus === ApplicationStatus.Completed
```
dheepak-aot, line 203: As a UI we do not route to change request when the application status is not completed, so is the second condition on applicationStatus required here?

dheepak-aot:
Great start for the submission of change requests. Please take a look at the comments.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: But is not considered as an existing direction based on the current change request validation?

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
andrewsignori-aot: It is not to prevent internal Vue routes but from the ability of the user to change a parameter directly in the UI.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: I get the intention now. I am good.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: The doubt is regarding assigning a non persistent object(without id) for both properties, how does typeorm manage to just insert original assessment once.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: I still feel that we should remove the condition on join from the query for the same reason that it is shared.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: > But is not considered as an existing direction based on the current change request validation?

Short answer, no.

andrewsignori-aot: I sent a message Josh, but unless it is a clear requirement, it should not be part of this PR and assuming that it should be part because the current implementation of the appeals may be a mistake.

dheepak-aot (approved):
Thanks for the changes and discussions, looks good 👍

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: I tested it in three different ways, and the result is that the `originalAssessment` requires the application ID associated with it some way. Removing line 441 would not work, alternatively, the below one would also work and produce the same SQL commands as the one in the PR.

```ts
newApplication.modifier = auditUser;
newApplication.updatedAt = now;
newApplication.currentAssessment = originalAssessment;
originalAssessment.application = newApplication; // It needs the application ID associated with to allow its insert.
await applicationRepository.save(newApplication);
```

```SQL
INSERT INTO
    "sims"."student_assessments"(
        "created_at",
        "updated_at",
        "submitted_date",
        "assessment_workflow_id",
        "assessment_data",
        "assessment_date",
        "trigger_type",
        "noa_approval_status",
        "workflow_data",
        "student_assessment_status",
        "student_assessment_status_updated_on",
        "calculation_start_date",
        "reported_date",
        "creator",
        "modifier",
        "application_id",
        "submitted_by",
        "offering_id",
        "student_appeal_id",
        "student_scholastic_standing_id",
        "application_offering_change_request_id",
        "related_application_assessment_id",
        "previous_date_changed_reported_assessment_id"
    )
VALUES
    (
        DEFAULT,
        DEFAULT,
        $ 1,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        $ 2,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        $ 3,
        DEFAULT,
        $ 4,
        $ 5,
        $ 6,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        DEFAULT,
        DEFAULT
    ) RETURNING "created_at",
    "updated_at",
    "id" -- PARAMETERS: ["2025-04-04T16:19:50.649Z","Original assessment",48552,37660,48552,39088]

UPDATE
    "sims"."applications"
SET
    "updated_at" = $ 1,
    "modifier" = $ 2,
    "current_assessment_id" = $ 3
WHERE
    "id" IN ($ 4) RETURNING "updated_at" -- PARAMETERS: ["2025-04-04T16:19:50.649Z",48552,37907,37660]
```

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -262,19 +284,75 @@ export default defineComponent({
...
+          {
+            programYearId: props.programYearId,
+            data,
+            associatedFiles,
+          },
+          { isChangeRequest: true },
+        );
+        snackBar.success("Thank you, your change request has been submitted.");
+        await router.push({
+          name: StudentRoutesConst.STUDENT_APPLICATION_DETAILS,
```
sh16011993, line 313: Why do we route them to `StudentRoutesConst.STUDENT_APPLICATION_DETAILS` instead of `StudentRoutesConst.STUDENT_APPLICATION_SUMMARY` which we are doing for a regular application submit?

andrewsignori-aot: image

sh16011993 (approved):
Exemplary work @andrewsignori-aot 👍 Great job.
Thank You for clarifying my questions over the call.

PR #4573, opened 2025-04-03
#4489 - Formio Changes: Split FT/PT form Updates Fix

sh16011993:
### As a part of this PR, the following was updated:

- Removed the `calculateValue` logic for the component "Capacity in which you will be attending this program:". This logic is not be needed anymore since the value for the component is passed to the form from the Vue.

**Screenshot:**

image


andrewsignori-aot (approved):

PR #4602, opened 2025-04-14
#4431 - Student View change request revamp - Fix Partner Submission

andrewsignori-aot:
## Main Issue

An error in the SQL query (missing parentheses) was causing the incorrect data to be retrieved.
_Note:_ removed some unnecessary data from the `select` that was no longer required.

## Secondary Issue

- Fixed form.io content duplication due to the known issue of rendering HTML inside a `p` tag. Applied the know fix, replacing the `p` with a `div`.
- Minor content adjustment for the third bullet in the instructions.

image

### DIFF
--- a/sources/packages/backend/apps/api/src/services/application/application.service.ts
+++ b/sources/packages/backend/apps/api/src/services/application/application.service.ts
@@ -1556,34 +1556,31 @@ export class ApplicationService extends RecordDataModelService<Application> {
         "programYear.startDate",
         "user.userName",
         "student.id",
-        "application.data",
-        "currentAssessment",
-        "offering.offeringIntensity",
       ])
       .innerJoin("application.student", "student")
       .innerJoin("student.user", "user")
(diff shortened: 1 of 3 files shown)


sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -1556,34 +1556,31 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+            "application.applicationEditStatus = :changeInProgressApplicationEditStatus",
+            {
+              changeInProgressApplicationEditStatus:
+                ApplicationEditStatus.ChangeInProgress,
+            },
+          );
+        }),
+      )
       .getOne();
   }
```
dheepak-aot, line 1585: 👍

dheepak-aot (approved):
Thanks for taking care of the issue and adjusting the content. 👍

sh16011993 (approved):
LGTM @andrewsignori-aot 👍

Issue #4516, opened 2025-03-20
Moratorium on BC private programs with online only


Joshua-Lakusta:
As a ministry user I want the new programs made by BC private institutions that include online delivery to go to pending for ministry review

**Acceptance Criteria**
- [ ] Make it so that all new programs (or edited programs) submitted by a BC private institution that have selected online only delivery style go to pending.
- [ ] If they select online and onsite (Blended) it does not go to pending
- [ ] If they select onsite only it does not go to pending
- [ ] Include the warning banner, similar to OOP institutions selecting online only program, if the institution selects Online Only.
> **This program requires review by StudentAid BC to determine eligibility.**

**Additional Context**
- policy directed change based on program year 
- confirmed from director of policy that the target of this is only on net new 100% online programs so excluding blended.

Joshua-Lakusta:
Waiting on confirmation that we are targeting both online and blended with this ticket. If blended is to be excluded from this ticket then selecting online and onsite would not lead to the pending status only online by itself would and additional ac will need to be added to clarify.

CarlyCotton:
@CarlyCotton  find warning banner for pending program online

PR #4608, opened 2025-04-15
#4516 - Moratorium on BC Private Programs with Online Program Delivery

sh16011993:
### As a part of this PR, updated the following:

- `Education Program` form was updated to include the logic to show the program as `Pending` status when a BC Private Institution selects Online Only as their Program Delivery mode.

**Screenshots:**

image


dheepak-aot (approved):
Looks Good 👍

PR #4612, opened 2025-04-15
#4431 - Student View change request revamp - E2E Tests

andrewsignori-aot:
## Change Request Submission

ApplicationStudentsController(e2e)-applicationChangeRequest
    √ Should create a change request when the application is completed. (2383 ms)                                                                                                                                  
    √ Should throw an Unauthorized exception when the student does not have a valid SIN. (1354 ms)                                                                                                                  
    √ Should throw an UnprocessableEntity exception when a change request submission happens for a not allowed program year. (1713 ms)                                                                              
    Should throw an UnprocessableEntity exception when there is already an in-progress change request application                                                                                                    
      √ with an edit status defined as 'Change in progress'. (2839 ms)                                                                                                                                             
      √ with an edit status defined as 'Change pending approval'. (2861 ms)   

## Change Request Cancellation

ApplicationStudentsController(e2e)-applicationCancelChangeRequest
  √ Should throw a NotFoundException when the application has already been declined by the Ministry. (1422 ms)                                                                                                   
  √ Should throw a NotFoundException when the application does not belong to the student. (1358 ms)                                                                                                              
  √ Should throw a NotFoundException when the application does not exist. (59 ms)                                                                                                                                
  Should cancel a change request when the change request is still considered 'in-progress'                                                                                                                       
    √ with an edit status defined as 'Change in progress'. (1836 ms)                                                                                                                                             
    √ with an edit status defined as 'Change pending approval'. (1748 ms)      

## Other changes

- Changed the SIN validation guard to throw a specific message when the SIN validation failed, following the same approach from other guards.

## E2E Support Related Changes

- Created a new VS Code task `Database - Reset for API E2E Tests` to clean the E2E test DB and execute the migration and API seed to "refresh" the local test databases.
- Updated the `clean-db.ts` methods to allow their execution multiple times, adding the `IF EXISTS`.


sims.code-workspace
```diff
@@ -80,6 +80,15 @@
         },
         "problemMatcher": []
       },
+      {
+        "label": "Database - Reset for API E2E Tests",
+        "type": "shell",
+        "command": "npm run db:seed:test:clean && npm run test:e2e:api:seed",
+        "options": {
+          "cwd": "${workspaceFolder:Backend}"
+        },
```
bidyashish, line 89: dedicated VS Code task +1

bidyashish (approved):
LGTM @andrewsignori-aot

sources/packages/backend/apps/api/src/auth/guards/student/sin-validation.guard.ts
```diff
@@ -30,6 +35,14 @@ export class SINValidationGuard implements CanActivate {
     }
     const { user } = context.switchToHttp().getRequest();
     const userToken = user as IUserToken;
-    return this.studentService.getStudentSinStatus(userToken.userId);
+    const hasValidSin = await this.studentService.getStudentSinStatus(
+      userToken.userId,
+    );
+    if (!hasValidSin) {
+      throw new UnauthorizedException(
```
dheepak-aot, line 42: 👍

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationChangeRequest.e2e-spec.ts
```diff
@@ -0,0 +1,307 @@
...
+    app = nestApplication;
+    appModule = module;
+    db = createE2EDataSources(dataSource);
+    // Program Year for the following tests.
+    const formService = await getProviderInstanceForModule(
+      appModule,
+      AppStudentsModule,
+      FormService,
+    );
+    programYear = await ensureProgramYearExists(db, 2050);
```
dheepak-aot, line 55: Is there any particular reason to go for a program year way ahead in future?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationChangeRequest.e2e-spec.ts
```diff
@@ -0,0 +1,307 @@
...
+    const expectedUserObject = expect.objectContaining({
+      id: student.user.id,
+    });
+    expect(createdApplication).toEqual({
+      id: createApplicationId,
+      submittedDate: now,
+      applicationNumber: completedApplication.applicationNumber,
+      relationshipStatus: defaultPayload.data.relationshipStatus,
+      studentNumber: defaultPayload.data.studentNumber,
+      programYear: expect.objectContaining({
```
dheepak-aot, line 165: Do we have to use the `expect.objectContaining` here?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationChangeRequest.e2e-spec.ts
```diff
@@ -0,0 +1,307 @@
...
+    // Expected user object.
+    const expectedUserObject = expect.objectContaining({
+      id: student.user.id,
+    });
+    expect(createdApplication).toEqual({
+      id: createApplicationId,
+      submittedDate: now,
+      applicationNumber: completedApplication.applicationNumber,
+      relationshipStatus: defaultPayload.data.relationshipStatus,
+      studentNumber: defaultPayload.data.studentNumber,
```
dheepak-aot, line 164: Minor and not blocker. As the default payload does not have student number, do we need to select and validate it?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationCancelChangeRequest.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+      undefined,
+      {
+        applicationStatus: ApplicationStatus.Edited,
+        applicationEditStatus: ApplicationEditStatus.ChangeInProgress,
+      },
+    );
+    const endpoint = `/students/application/${completedApplication.id}/cancel-change-request`;
+    const token = await getStudentToken(
+      FakeStudentUsersTypes.FakeStudentUserType1,
+    );
```
dheepak-aot, line 164: A minor observation here. This test currently works because the mock is set by previous tests to use different studentId and I believe the mock is not cleared not because of `FakeStudentUserType1` user is returned at runtime. Not a blocker.

sources/packages/backend/apps/test-db-seeding/src/clean-db/clean-db.ts
```diff
@@ -8,7 +8,9 @@ export class CleanDatabase {
   async cleanDatabase(): Promise<void> {
     // Drops the database and all its data.It will erase all your database tables and their data.
     await this.dataSource.dropDatabase();
-    await this.dataSource.query("DROP EXTENSION pg_trgm");
-    await this.dataSource.query("DROP FUNCTION sims.create_history_entry");
+    await this.dataSource.query("DROP EXTENSION IF EXISTS pg_trgm");
+    await this.dataSource.query(
+      "DROP FUNCTION IF EXISTS sims.create_history_entry",
+    );
```
dheepak-aot, line 14: 👍

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -387,9 +387,13 @@ export class ApplicationService extends RecordDataModelService<Application> {
     const auditUser = { id: auditUserId } as User;
     const originalAssessment = new StudentAssessment();
     originalAssessment.triggerType = AssessmentTriggerType.OriginalAssessment;
+    originalAssessment.studentAssessmentStatus =
+      StudentAssessmentStatus.Submitted;
```
dheepak-aot, line 391: What is the reason for explicitly setting this status?

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationChangeRequest.e2e-spec.ts
andrewsignori-aot: To avoid using the existing ones. It will work with any PY 2025 on, so instead of using 2025, which will potentially be deactivated a few years from now, I just used one in the far future.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationChangeRequest.e2e-spec.ts
andrewsignori-aot: The default payload has the `studentNumber` defined below. Am I missing something?
```ts
// Application mocked data.
const data = {
  relationshipStatus: RelationshipStatus.Other,
  studentNumber: "123456789",
};
```

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationCancelChangeRequest.e2e-spec.ts
andrewsignori-aot: It works because the student associated with the application will never be the one mocked before, or the one if we reset the current mock. I can reset the mock if it would make it clear.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: The idea was to have the date submitted aligned to also use the `now`, and it seems not clear to set the date without setting the status. I can remove it if it would cause more confusion but I would also remove the date then and adjust the E2Es.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationChangeRequest.e2e-spec.ts
andrewsignori-aot: Removed `expect.objectContaining` on both E2E files. They were added by mistake due to another error in the SQL queries that were returning more data than they should. When the data is right, there is no reason to use `expect.objectContaining`.
Nice catch!

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot: That's alright. I wasn't sure about the reason and so made the comment.

sources/packages/backend/apps/api/src/route-controllers/application/_tests_/application.students.controller.applicationCancelChangeRequest.e2e-spec.ts
dheepak-aot: Reset wouldn't be required in this case, If there was a shared student then it would be required.

dheepak-aot (approved):

Issue #4619, opened 2025-04-17
BUG HOTFIX 2.7.1 - Student/Supporting User CRA Income


CarlyCotton:
**Expected**
Student CRA income response should be used in `studentDataCRAReportedIncome' 

**Actual**
Supporting user CRA income response  is used in `studentDataCRAReportedIncome' 

1. Student application in PT 24/25 application, married and partner has BCSC
2. Partner completes BCSC
3. CRA income verification comes in
4. studentDataCRAReportedIncome will reflect CRA verification income for supporting user.

CarlyCotton:
Tested in staging and CRA reported income for a BCSC supporting user  & student are being used correctly

PR #4621, opened 2025-04-17
#4619 - Hotfix - Student reported income fix

dheepak-aot:
## Issue

When an application has a supporting user and the supporting user has successful income verification, 
then the student reported income `studentDataCRAReportedIncome` is set  with partner's CRA Reported income value
by the worker job `loadAssessmentConsolidatedData` [loadAssessmentConsolidatedData](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts#L149)
  
This is because following logic receives both student and supporting user income verifications after filter BECAUSE the SQL query does not return the supporting user id.
image

## Solution
The bug has been introduced since we removed TypeOrm `RelationId` property and referenced the relational object in a way that wasn't working so left join was added to fix the issue and used the join object to select.

## Update to the income verification bypass.

- [x]  In the local environment, we bypass the CRA income verifications to allow the workflow to proceed. During this process updated the logic to save CRA Reported Income as User provided income to allow  manual tests to be easier in local which involves CRA reported income.

## Manual testing with supporting user

CRA Income verifications table

image

Camunda variable

image

### DIFF
--- a/sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
+++ b/sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
@@ -152,7 +152,7 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
         "craIncomeVerification.id",
         "craIncomeVerification.craReportedIncome",
         "craIncomeVerification.taxYear",
-        "incomeVerificationSupportingUser.id",
+        "craIncomeVerificationSupportingUser.id",
         "studentAppeal.id",
         "appealRequest.id",
         "institutionLocation.data",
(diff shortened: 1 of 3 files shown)


sources/packages/backend/apps/workers/src/services/student-assessment/student-assessment.service.ts
```diff
@@ -176,6 +176,10 @@ export class StudentAssessmentService extends RecordDataModelService<StudentAsse
         { appealStatus: StudentAppealStatus.Approved },
       )
       .leftJoin("application.craIncomeVerifications", "craIncomeVerification")
+      .leftJoin(
+        "craIncomeVerification.supportingUser",
+        "incomeVerificationSupportingUser",
```
andrewsignori-aot, line 181: Minor, since we have the `craIncomeVerification` for the student I would suggesting naming it as `craIncomeVerificationSupportingUser` to make the select looks like 
```
"craIncomeVerification.id",
"craIncomeVerification.craReportedIncome",
"craIncomeVerification.taxYear",
"craIncomeVerificationSupportingUser.id",
```

andrewsignori-aot (approved):
Thanks for the investigation and troubleshooting. Only a minor suggestion, hence approving it.

Issue #4616, opened 2025-04-17
Ministry View Change Request Revamp (Application view menu)


Joshua-Lakusta:
As a ministry user I need updates to the left side panel that contains the application and edit history to clean it up and add the pending change request static option.

**Acceptance Criteria**

### Active Section updates
- [x] Add a new header that says Active above the active version of the application
- [x] Sub header added with submitted date and time stamp
- [x] Add status under the the application link that shows the sub status.
- [x] Map Changed with approval status as Changed
- [x] Active application section will not have a collapse function and will remain static

### Pending change request section
- [x] Add new header that will appear underneath the active section when there is a change request with the sub status change pending approval or change in progress. 
- [x] If there is no change request in either of these statuses then hide this header and section.
- [x] Header content
> Pending Change Request
- [x] Change request that is in this state will display a line that is the date timestamp (status) and will have the change request application view below that(see screenshot)
- [x] Sub header added with submitted date and time stamp
- [x] This section will be static and not have a collapse function 
- [x] Add sub status beneath the Application link
- [x] Supporting user information needs to be shown (if applicable) similar to active application
- [x] When viewing the application that is pending here it should show highlighted changes in the form.

### Historical app version section update
- [x] Add new header above the historical applications with the following content

> Application History

- [x] Sub header added with following content

> Past submitted applications

- [x] Add sub status underneath the Application link for each version
- [x] Map Changed with approval status as Changed
- [x] Supporting user information needs to be shown (if applicable) similar to active application
### Reference image
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNEd4Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b41854775b7ffade064a8dbbae4b30e873b7cc0b/image.png" alt="image.png" />

**Additional Context**
-

CarlyCotton:
@Joshua-Lakusta Added ACs below in the pending change request section. Going to move forward based on this as-estimated. 

- [ ] Supporting user information needs to be shown (if applicable) similar to active application
- [ ] When viewing the application that is pending here it should show highlighted changes in the form.

Joshua-Lakusta:
[...]

PR #4628, opened 2025-04-23
#4616 - Ministry View Change Request Revamp (Application view menu)

andrewsignori-aot:
# API Related Changes

- Change the API that retrieves the application's overall information to also get the supporting users' information, which allowed the old API method to be removed as planned in the previous efforts.
  - Kept the logic to decide which application version is an "in-progress change request" in the API, which means that the API is responsible for returning a change request in an explicit way (DTO property `inProgressChangeRequest`).
  - Considered all applications as a version of itself, which means current, change request, and past versions share the same output DTO.
  - As a side effect of having the supporting users associated with change requests, every version now can easily display supporting users in the UI (it is just not enabled yet because it is not part of the ACs).
- Added the below E2E for the new scenarios
  - Should get the original application and no application versions in overall details when there is no edited application associated with the given application.
  - Should get the current application and an in-progress change request, including supporting users when the application has an in-progress change request.
  - Should get the current application and an application version when the application has a declined change request.
  - Should get the current application and versions supporting users when the application and its versions have supporting users.

# UI related changes

- Refactored the Ministry applications menu to look more like the Ministry main left menu.
  - The menu is now integrated with the routes and has the `active` behavior.
- The route for the pending change request is the same as any other application version, and the "change request" is treated as a regular past version.

## In-progress change request

image

## Parents and Partner with pending change request

image

## Active application with versions and no pending request

image

## How it was displayed on Edge (previously, there was extra white space between the icons)

image

## Local change only to allow displaying the `Changed with approval` renamed to `Changed`

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
@@ -872,12 +872,10 @@ export class ApplicationControllerService {
       id: application.id,
       submittedDate: application.submittedDate,
       applicationEditStatus: application.applicationEditStatus,
-      supportingUsers: application.supportingUsers.length
-        ? application.supportingUsers.map((user) => ({
-            supportingUserId: user.id,
-            supportingUserType: user.supportingUserType,
-          }))
-        : undefined,
+      supportingUsers: application.supportingUsers.map((user) => ({
+        supportingUserId: user.id,
+        supportingUserType: user.supportingUserType,
+      })),
     };
   }
 
--- a/sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
@@ -32,6 +32,7 @@ import {
   ApplicationOverallDetailsAPIOutDTO,
   SaveApplicationAPIInDTO,
   ChangeRequestInProgressAPIOutDTO,
+  ApplicationVersionAPIOutDTO,
 } from "./models/application.dto";
 import {
   allowApplicationChangeRequest,
@@ -810,7 +811,9 @@ export class ApplicationControllerService {
   }
 
   /**
-   * Get application overall details for the given application.
+   * Get application overall details for the given application,
+   * including the active application, in-progress change request,
+   * and previous versions of the application.
    * @param applicationId application ID.
    * @param options options.
    * - `studentId` student ID used for authorization.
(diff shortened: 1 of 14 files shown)


sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.aest.controller.ts
```diff
@@ -23,25 +20,6 @@ import { ClientTypeBaseRoute } from "../../types";
 @ApiTags(`${ClientTypeBaseRoute.AEST}-supporting-user`)
 export class SupportingUserAESTController {
   constructor(private readonly supportingUserService: SupportingUserService) {}
-  /**
-   * Get the list of supporting users respective to
-   * an application id for AEST user.
-   * @param applicationId application id.
-   * @return list of supporting users of an application.
-   */
-  @Get("application/:applicationId")
```
dheepak-aot, line 32: 👍

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -822,27 +823,62 @@ export class ApplicationControllerService {
...
+      APPLICATION_EDIT_STATUS_IN_PROGRESS_VALUES.includes(
+        version.applicationEditStatus,
+      ),
+    );
+    // Remove the in-progress change request from the previous versions list.
+    // A maximum of one in-progress change request is allowed.
+    if (inProgressChangeRequest) {
+      previousVersions = previousVersions.filter(
+        (version) => version.id !== inProgressChangeRequest.id,
+      );
```
dheepak-aot, line 853: We can do an early return if the previous versions are not present.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -2171,32 +2171,49 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+        "supportingUser.id",
+        "supportingUser.supportingUserType",
+        "versionSupportingUser.id",
+        "versionSupportingUser.supportingUserType",
       ])
       .innerJoin("application.parentApplication", "parentApplication")
+      .leftJoin(
+        "application.supportingUsers",
+        "supportingUser",
+        "supportingUser.id IS NOT NULL",
```
dheepak-aot, line 2204: What is the benefit of this condition?

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
andrewsignori-aot: The goal of the early return is to reduce nesting, complexity, and make the code easy to read.
The nesting would not be reduced in this situation. I added the early return to see how it looks and it does not seems like an improvement in this scenario. Please let me know if I am missing something at a level that it would be a blocker.

sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
```diff
@@ -263,12 +264,35 @@ export class ApplicationWarningsAPIOutDTO {
...
 
+export interface ApplicationSupportingUsersAPIOutDTO {
+  supportingUserId: number;
+  supportingUserType: SupportingUserType;
+}
+
+export class ApplicationSupportingUsersAPIOutDTO {
+  supportingUserId: number;
+  supportingUserType: SupportingUserType;
+}
```
dheepak-aot, line 275: Why do we have class and interface of same type?

andrewsignori-aot: Issue during copy and paste between API/Web DTO. It is fixed now.

dheepak-aot:
Great start @andrewsignori-aot only minor comments.

dheepak-aot (approved):

Issue #4501, opened 2025-03-19
BUG: COE Query Impacting Count and Record Visibility 


astridSABC:
**Describe the Bug**
COE table does not display correct enrolment of students on the page or might be missing students entirely

**Expected Behaviour**
If 10 items per page and there are 10 students for COE they should be displayed on one page
If 12 students, 10 should be listed on one page and 2 students on the other
All students who are required to have COE are displayed - the institution user does not need to search for them

**Actual Behaviour**
Some pages have 2 students, some has 4 and, had to search for a specific student not listed.  Note each of the pages for "Ackbar"

**Page 1** Counter beside "enrolment" displays (2) but there are maybe 7 students for COE.  The counter at the bottom of the table suggests more applications awaiting COE. No "Ackbar" student
<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUZWQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--85e7b1b0f91a233b0be5ac09d80b0e9f443e2bce/image.png" alt="image.png" />

**Page 2** Counter beside "enrolment" displays (4) but there are maybe 7 students for COE.  The counter at the bottom of the table suggests more applications awaiting COE. No "Acbar" student

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeWxWQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--786680cba5cb58a66057b512b116956b6cee4993/image.png" alt="image.png" />

**Search for a student**
<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemRWQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9f723f89a856849b55676503ee64dd47d0ed01ae/image.png" alt="image.png" />

### Steps To Reproduce

Precondition: PT offering is available to select for various dates.  12 Single student application (no PD, no dependant) and have accepted their assessments  (generally there are 10 students per page)

1. Login to the institution portal
2. Select "locations" in the left pane
3. Select "confirmation enrolment"
4. Note the page count beside the word enrolment
[...]

PR #4660, opened 2025-04-30
#4501 - BUG: COE Query Impacting Count and Record Visibility 

bidyashish:
**Expected Behaviour**
If 10 items per page and there are 10 students for COE they should be displayed on one page
If 12 students, 10 should be listed on one page and 2 students on the other
All students who are required to have COE are displayed - the institution user does not need to search for them

Issue is using 2 query in https://github.com/typeorm/typeorm/issues/320

*Note:
code to use .skip() instead of .offset() and .take() instead of .limit()

### DIFF
--- a/sources/packages/backend/apps/api/src/services/disbursement-schedule/disbursement-schedule-service.ts
+++ b/sources/packages/backend/apps/api/src/services/disbursement-schedule/disbursement-schedule-service.ts
@@ -72,8 +72,16 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
           paginationOptions.sortOrder,
         ),
       )
-      .offset(paginationOptions.page * paginationOptions.pageLimit)
-      .limit(paginationOptions.pageLimit);
+      .skip(paginationOptions.page * paginationOptions.pageLimit)
+      .take(paginationOptions.pageLimit);
+
+    // Log the generated SQL query and parameters
+    const [sql, parameters] = disbursementCOEQuery.getQueryAndParameters();
+    console.log("Executing COE query:");
+    console.log(sql);
+    console.log("Parameters:");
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/api/src/services/disbursement-schedule/disbursement-schedule-service.ts
```diff
@@ -72,8 +72,9 @@ export class DisbursementScheduleService extends RecordDataModelService<Disburse
           paginationOptions.sortOrder,
         ),
       )
-      .offset(paginationOptions.page * paginationOptions.pageLimit)
-      .limit(paginationOptions.pageLimit);
+      .skip(paginationOptions.page * paginationOptions.pageLimit)
```
dheepak-aot, line 75: The count of records is also incorrect in the page(UI). 
image

https://github.com/bcgov/SIMS/blob/33c5e6c3915dccf876a876d8cc09b9c854751ca0/sources/packages/web/src/views/institution/locations/confirmation-of-enrollment/COESummaryData.vue#L4

image

dheepak-aot:
The fix applied in the PR looks good for the bug fix. For the long term perspective as we discussed we should create a technical debt for long term solution.

sources/packages/backend/apps/api/src/services/disbursement-schedule/disbursement-schedule-service.ts
bidyashish: Thanks @dheepak-aot 

Count updated to use from api response count.

dheepak-aot (approved):
Thanks for doing the change. Looks good 👍

sh16011993 (approved):
LGTM @bidyashish 👍

Issue #4490, opened 2025-03-17
FormIO - Enable Editor in Test & PR Process


CarlyCotton:
# **User Story**
As a BA ministry user in the SIMS team, I want to be able to use form.io editor in the TEST environment to make changes. I also want traceability and the ability to have my changes peer reviewed before they go through the environments to production.

# **Acceptance Criteria**
- [x] create a new page in ministry account which is available to new keycloak role of 'form-editor' 
- [x] user can view list of forms to edit 
- [x] user can save the form, pushing the form definitions to mongodb 
- [x] user can download the form to their desktop   

Additional AC: 

- [x]  Allow the visualization of the form.io definition (e.g. form definition inspector).
- [x]  Allow a form.io definition to be pasted in the "form definition inspector" and loaded in the editor.
- [x]  Allow easy access to the GitHub repository file (the repo link can be hard-coded for now).
- [x]  Create some initial builder custom components as a starting point.

Note: Using this functionality, the ministry users can make changes, validate in the UI, and then download form files which they can then add to PRs. Business analysts to create wiki documentation after determining process. 

# **Technical Acceptance**
## Connection: 
- [x] Access formio backend using their APIs
## API: 
- [x] retrieve list of forms from mongo 
- [x] create API to save form definitions to mongo 
- [x] create API to download form definition as a file   

## UI: 
- [x] create new page to show list of forms 
- [x] use existing formio library to display form in [editor view](https://help.form.io/developers/form-development/form-builder) 
- [x] enable save form functionality

ninosamson:
Yes to including all @andrewsignori-aot - just a question on form.io definition - that's the raw JSON at the bottom of the editor that you showed? if so, I think that's very helpful, especially when someone is pulling in a definition they previously saved.

ninosamson:
Updated the AC accordingly.

andrewsignori-aot:
> Yes to including all [@andrewsignori-aot](https://github.com/andrewsignori-aot) - just a question on form.io definition - that's the raw JSON at the bottom of the editor that you showed? if so, I think that's very helpful, especially when someone is pulling in a definition they previously saved.

Yes @ninosamson, that can receive the input from a file and is also updated at every change in the editor which means it works in both way.
1. It can be edited to load the editor with a previous previous;
2. It always has the most update version from what is displayed in the editor, and its output is the same from the "Copy" button.

dheepak-aot:
[...]

PR #4674, opened 2025-05-05
#4490 - FormIO - Enable Editor in Test

andrewsignori-aot:
## API Configurations Changes

- Create a new role `aest-form-editor` to allow access to the page and API endpoints.
- Increase the `json body parser limit`  from the default `100kb` to `200kb` only for the `DynamicFormController`. Required to allow definition like the Student Application, to be updated through the API.
  - `bodyParser: false` was configured in the `main.ts` to be later configured in the `app.module`. The order of the configurations matters to allow the `200Kb` to reach the `DynamicFormController`.

## New/Updated API Endpoints for Dynamic Forms
- GET `dynamic-form/list`
- PUT `dynamic-form/:formName`
 
## UI Secondary Changes
- Bootstrap updated to version [5 to be aligned with the current form.io version](https://form.io/release-notes/formio-enterprise-9-3-0/).
- Bootstrap icons included since it is the main library used by form.io, **which fixes some icon issues throughout the application, for instance, in the NOA expanders.**
- PrimeVue style changes are required to avoid issues around the table styles and pagination components.
- Some "border-opacity" styles were affected and needed some adjustment.

## New Form Editor Page

- Added new "Dynamic Forms Editor" link conditionally displayed following the same behavior as "Queues Dashboard".

image

### Custom components

Created some custom components to be available in the builder as a starting point to be expanded as the business requires.

image


sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
```diff
@@ -23,12 +28,46 @@ export class DynamicFormController extends BaseController {
...
-    return this.formService.list();
+  async list(): Promise<FormsAPIOutDTO> {
+    const allForms = await this.formService.list();
+    return {
+      forms: allForms.map((form) => ({
+        name: form.name,
+        title: form.title,
+      })),
+    };
   }
```
dheepak-aot, line 45: Following our client based seperation, these endpoints must be under *.aest.controller

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
```diff
@@ -23,12 +28,46 @@ export class DynamicFormController extends BaseController {
     super();
   }
 
+  /**
+   * List all form definitions that contains the tag 'common' ordered by title.
+   * @returns list of form definitions.
+   */
+  @Roles(Role.AESTFormEditor)
```
dheepak-aot, line 35: This endpoint must be allowed only for AEST azp.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
```diff
@@ -23,12 +28,46 @@ export class DynamicFormController extends BaseController {
...
     return this.formService.fetch(formNameParam.formName);
   }
+
+  /**
+   * Update a form definition in Form.io.
+   * @param formName Name of the form to be updated.
+   * @param payload Form definition to be updated.
+   */
+  @Roles(Role.AESTFormEditor)
+  @Put(":formName")
```
dheepak-aot, line 63: Correct me if I am wrong. As per the form IO API specification it is the formPath and not the formName. As we have both as same it is working for us.

image

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/models/form.dto.ts
```diff
@@ -1,6 +1,23 @@
...
+
+export class FormUpdateAPIInDTO {
+  @IsObject()
+  @JsonMaxSize(JSON_200KB)
+  formDefinition: unknown;
+}
+
+export class FormAPIOutDTO {
+  title: string;
+  name: string;
```
dheepak-aot, line 18: I would say that we must retrieve the path and use the path unless I misunderstood on the expectation of PUT using the path instead of form name.

sources/packages/backend/apps/api/src/services/form/form.service.ts
```diff
@@ -43,13 +53,32 @@ export class FormService {
...
   /**
-   * Lists form definitions that contains the tag 'common'.
+   * Lists form definitions that contains the tag 'common' ordered by title.
    */
-  async list() {
+  async list(): Promise<FormDefinition[]> {
+    const authHeader = await this.createAuthHeader();
     const content = await this.httpService.axiosRef.get(
-      `${this.config.formsUrl}/form?type=form&tags=common`,
+      `${this.config.formsUrl}/form?tags=common&limit=${FORMIO_PAGE_LIMIT}&sort=${FORMIO_LIST_SORT_FIELD}`,
```
dheepak-aot, line 61: Asking for my understanding why was the param `tags=common` used here?

sources/packages/backend/apps/api/src/services/form/form.service.ts
```diff
@@ -43,13 +53,32 @@ export class FormService {
   }
 
   /**
-   * Lists form definitions that contains the tag 'common'.
+   * Lists form definitions that contains the tag 'common' ordered by title.
```
dheepak-aot, line 56: Please add @returns to the comment.

sources/packages/web/index.html
```diff
@@ -44,12 +44,6 @@
       color="#5bbad5"
       href="/safari-pinned-tab.svg"
     />
-    <link
-      rel="stylesheet"
-      href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-      integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh"
-      crossorigin="anonymous"
-    />
```
dheepak-aot, line 52: 👍

sources/packages/web/package-lock.json
```diff
@@ -641,15 +643,15 @@
       }
     },
     "node_modules/@formio/js": {
-      "version": "5.0.1",
-      "resolved": "https://registry.npmjs.org/@formio/js/-/js-5.0.1.tgz",
-      "integrity": "sha512-MVAT9lseEYmdiivVuQRvWxwXpp402YYpStrTkMqi8nvCA/AmFb9Ld5nsFUqQsfzf4jzY5NgefFmoFlSMfTk0Kg==",
+      "version": "5.0.0",
+      "resolved": "https://registry.npmjs.org/@formio/js/-/js-5.0.0.tgz",
+      "integrity": "sha512-Ly3f7q2wo5oFjlECKFtkIMy+Ah9HyAkf6ob70OrF/6MooSPF3hEu9euXouEWOoNdI+Tokr690/GdM2sDZtYzww==",
```
dheepak-aot, line 648: Did it downgrade the SDK?

sources/packages/web/package-lock.json
```diff
@@ -621,9 +623,9 @@
       }
     },
     "node_modules/@formio/core": {
-      "version": "2.3.2",
-      "resolved": "https://registry.npmjs.org/@formio/core/-/core-2.3.2.tgz",
-      "integrity": "sha512-tkH9MtDR+nZuTuizy+qlnce03C3hnd4Dm/2eFjY7cEUZdpIIh1QQ9j9s4W47Si/QOMkS7trIVxKUjxuKrt8j6Q==",
+      "version": "2.3.0",
```
dheepak-aot, line 626: Same here https://github.com/bcgov/SIMS/pull/4674/files#r2082530281

sources/packages/web/src/composables/useFormioUtils.ts
```diff
@@ -261,6 +280,32 @@ export function useFormioUtils() {
     customUtils[name] = method;
   };
 
+  /**
+   * Generate the dynamic form definition removing unwanted properties.
+   */
+  const getReadyToSaveFormDefinition = (formDefinition: unknown): unknown => {
```
dheepak-aot, line 286: Please add @parm and @returns to the comments.

sources/packages/web/src/composables/useFormioUtils.ts
```diff
@@ -261,6 +280,32 @@ export function useFormioUtils() {
...
+    // Clone the form definition to avoid modifying the original one.
+    const clonedForm = JSON.parse(JSON.stringify(formDefinition));
+    // Remove non-required properties from the form definition.
+    NON_REQUIRED_FORM_PROPERTIES.forEach((property) => {
+      delete clonedForm[property];
+    });
```
dheepak-aot, line 292: What about the same elimination at the server side?

sources/packages/web/src/services/http/DynamicForms.ts
```diff
@@ -1,11 +1,35 @@
...
     return this.getCall(`dynamic-form/${formName}`);
   }
 
-  public async getFormlist(): Promise<any> {
+  /**
+   * List all form definitions that contains the tag 'common' ordered by title.
+   * @returns list of form definitions.
+   */
+  async getFormsList(): Promise<FormsAPIOutDTO> {
     return this.getCall(`dynamic-form`);
```
dheepak-aot, line 19: Optional. Please remove the back ticks which is not required for static url.

sources/packages/web/src/views/aest/FormioEditor.vue
```diff
@@ -0,0 +1,305 @@
...
+     */
+    const viewOnRepo = async (): Promise<void> => {
+      if (selectedForm.value === undefined) {
+        snackBar.warn("Please select a form.");
+        return;
+      }
+      window.open(
+        `https://github.dev/bcgov/SIMS/blob/main/sources/packages/forms/src/form-definitions/${selectedForm.value.name.toLowerCase()}.json`,
+        "_blank",
+      );
```
dheepak-aot, line 256: Sonar use to complain about using `noopener` looks like it stopped now.

image

sources/packages/web/src/views/aest/FormioEditor.vue
```diff
@@ -0,0 +1,305 @@
...
+    };
+
+    /**
+     * Open the form definition in the GitHub repository.
+     */
+    const viewOnRepo = async (): Promise<void> => {
+      if (selectedForm.value === undefined) {
+        snackBar.warn("Please select a form.");
+        return;
+      }
```
dheepak-aot, line 252: Is this condition required?

sources/packages/web/src/views/aest/FormioEditor.vue
```diff
@@ -0,0 +1,305 @@
...
+      const shouldContinue = await saveDynamicFormModal.value.showModal();
+      if (!shouldContinue) {
+        return;
+      }
+      try {
+        saveDynamicFormLoading.value = true;
+        await ApiClient.DynamicForms.updateForm(selectedForm.value.name, {
+          formDefinition: getReadyToSaveFormDefinition(builder.schema),
+        });
```
dheepak-aot, line 224: It would be better to show a toast when form is saved to the server.

dheepak-aot:
Excellent work bring the new form builder which is going to make the form edits much more easier. Please have a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
andrewsignori-aot: I will move this one that was already misplaced, and the new ones.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
andrewsignori-aot: It will be since we are moving to AEST-specific controller as mentioned here https://github.com/bcgov/SIMS/pull/4674#discussion_r2082489483

sources/packages/backend/apps/api/src/services/form/form.service.ts
andrewsignori-aot: "Common" is a tag added to all the forms created for the application. It is used just to filter some out-of-the-box forms that come with form.io server.
It was a simple way to keep using what was already in place. The tags can be explored and organized better in the future.

sources/packages/web/package-lock.json
andrewsignori-aot: Not intentionally. The main package was never upgraded to `5.0.1`. The lock was updated 2 weeks ago when we did the nestjs upgrade (potentially not intentional).
I will revert to keep the changes to minimal.

sources/packages/web/src/composables/useFormioUtils.ts
andrewsignori-aot: These properties are not harmful, they do not pose a security risk (IMO), and we remove them because we do not use them.
I would avoid executing the same logic twice. It is important on the client side because the definition can be saved and persisted directly from the client side to the repo, for instance.
I did not see a need to replicate the same on the server side. Please let me know if we need to extend this conversation or if I am missing something.

sources/packages/web/src/views/aest/FormioEditor.vue
andrewsignori-aot: I will add.

sources/packages/web/src/views/aest/FormioEditor.vue
andrewsignori-aot: Either the condition is present, or we add the null propagation check.
I changed to use the null propagation now.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
andrewsignori-aot: Controller methods moved.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.controller.ts
andrewsignori-aot: It is the same everywhere else in the application. I will change the ones for the new implementation and see the impact of changing the existing ones, maybe in a separate PR.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/models/form.dto.ts
andrewsignori-aot: I replaced it with the path now. We have been using formName as "the path" since always.

sources/packages/backend/apps/api/src/services/form/form.service.ts
andrewsignori-aot: Added.

sources/packages/web/package-lock.json
andrewsignori-aot: https://github.com/bcgov/SIMS/pull/4674#discussion_r2085242238

sources/packages/web/src/composables/useFormioUtils.ts
andrewsignori-aot: Added.

sources/packages/web/src/views/aest/FormioEditor.vue
andrewsignori-aot: Removed and added the null propagation to keep it simple.

sources/packages/web/src/views/aest/FormioEditor.vue
andrewsignori-aot: https://github.com/bcgov/SIMS/pull/4674#discussion_r2085248784

sources/packages/web/src/composables/useFormioUtils.ts
dheepak-aot: Okay. Given that the endpoint is ministry only I am good with not replicating the logic on server side.

sources/packages/web/src/services/http/DynamicForms.ts
```diff
@@ -1,11 +1,35 @@
...
+    return this.getCall(this.addClientRoot("dynamic-form"));
+  }
+
+  /**
+   * Update a form definition in Form.io.
+   * @param formName Name of the form to be updated.
+   * @param payload Form definition to be updated.
+   */
+  async updateForm(
+    formName: string,
```
dheepak-aot, line 28: Please update the param and comment to form path.

sources/packages/backend/apps/api/src/route-controllers/dynamic-form/dynamic-form.aest.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+      forms: allForms.map((form) => ({
+        title: form.title,
+        path: form.path,
+      })),
+    };
+  }
+
+  /**
+   * Update a form definition in Form.io.
+   * @param formName Name of the form to be updated.
```
dheepak-aot, line 41: Please update the comment.

sources/packages/web/src/views/aest/FormioEditor.vue
andrewsignori-aot: Toast added.

Issue #4567, opened 2025-04-02
Ministry View Change Request Revamp (Ministry List UI)


Joshua-Lakusta:
As a ministry user I need there to be UI list that will show all the change requests. I need this to be able to easily view overall work load and access students that need adjudication 

### Ministry list UI 

- [ ] "New" Change requests when submitted should show on their own new UI list on ministry navigation menu 
- [ ] Title header will match the link that takes you to the list
- [ ] Change requests that are at the change pending approval status should show on this list
- [ ] New list name:

   > Change request (2025-2026 and later)

- [ ] Change current change request list on ministry  navigation menu name to:

   > Change request (Pre 2025-2026)

- [ ] The layout, sorting, and search functionality of the new list should match the current change request list
- [ ] Title header will match the link that takes you to the list
- [ ] When view is selected next to a change request in the new list it should direct the ministry user to the application view of the pending change request
- [ ] The new change request should **not** create a line in the unapproved change request section on the application assessment page (this is only for old-process change requests)
- [ ] Should **not** create a completed change line on the application assessment page (this is only for old-process change requests)

<img width="1601" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOVM0Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--aec9bed9317e6b1b9ca6092865e5366bff1f763f/image.png" alt="image.png" />

**Technical ACs**
- [ ] Ensure pagination is done using offset and limit instead of skip and take

**Additional Context**
-

PR #4676, opened 2025-05-05
#4567 Ministry View Change Request Revamp (Ministry List UI) - Part -1

bidyashish:
New Template API for `application-change-request`

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/application-change-request/models/application-change-request.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/application-change-request/models/application-change-request.dto.ts
@@ -1,0 +1,12 @@
+/**
+ * DTO for pending application change request summary.
+ */
+export class ApplicationChangeRequestPendingSummaryAPIOutDTO {
+  requestId: number;
+  applicationId: number;
+  studentId: number;
+  submittedDate: Date;
+  firstName?: string;
+  lastName: string;
+  applicationNumber: string;
+} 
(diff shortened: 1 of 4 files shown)


sources/packages/backend/apps/api/src/route-controllers/application-change-request/models/application-change-request.dto.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
andrewsignori-aot, line 1: We can review the DTO once it is the final version.
As a minor comment, the "change request ID" is actually the application ID itself.
The `studentId` is probably not required and can be added later, if needed.
To allow the proper redirect, the `parentApplicationId` will be required.
This would be a recommended DTO, but we need also ensure the pagination.
If this is just used as a place holder, feel free to move ahead.
```ts
export class ApplicationChangeRequestPendingSummaryAPIOutDTO {
  applicationId: number;
  parentApplicationId: number;
  submittedDate: Date;
  firstName?: string;
  lastName: string;
  applicationNumber: string;
}
```

andrewsignori-aot (approved):
Nice start, only one comment but it can be adjusted in the upcoming PR.

sh16011993 (approved):
Nice work @bidyashish 👍  Thanks for creating the controller. 😄

sources/packages/backend/apps/api/src/route-controllers/application-change-request/models/application-change-request.dto.ts
bidyashish: Thanks for the review. 
Will adjust in next PR.

Issue #4690, opened 2025-05-12
Update Deploy-All to use tag gitreference


ninosamson:
Update the deploy all workflow so that manually triggering the deploy all workflow will automatically use the selected tag for the build reference and run name. 

## Acceptance Criteria
> Generated by [Zenhub AI](https://www.zenhub.com/ai)

 - [ ] **Scenario: Update Deploy-All to use tag gitreference**
  - Given the Deploy-All workflow is configured with a tag gitreference
  - When the workflow is manually triggered
  - Then the workflow should use the selected tag for the build reference and run name

PR #4692, opened 2025-05-12
#4690 - Update deploy-all workflow to use selected tag

ninosamson:
- Update the "deploy-all" workflow so that manually triggering the deploy all workflow will automatically use the selected tag for the build reference and run name. 
- Note: used "github.ref" vs "github.ref_name" to be explicit for "actions/checkout@v4"

### DIFF
--- a/.github/workflows/release-deploy-all.yml
+++ b/.github/workflows/release-deploy-all.yml
@@ -1,13 +1,9 @@
 name: Release - Deploy All
-run-name: Release - Deploy all from ${{ inputs.gitRef }} to ${{ inputs.environment }}
+run-name: Release - Deploy all from ${{ github.ref_name }} to ${{ inputs.environment }}
 
 on:
   workflow_dispatch:
     inputs:
-      gitRef:
-        description: "Git Ref (e.g. v1.0.0 or v1.0.0-prerelease.1)"
-        required: true
-        default: ""
       environment:
         description: "Environment"
         required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/release-deploy-all.yml
andrewsignori-aot, line 1: Please ensure the steps are using the below. Some places are using `ref: ${{ github.ref }}` instead. Was that on purpose?
```yml
- name: Checkout Target Branch
  uses: actions/checkout@v4
  with:
    ref: ${{ github.ref_name }}
```

ninosamson: Yes - added so it would be more explicit, but if using github.ref_name will always be sufficient, then I can update from github.ref.

ninosamson: Updated to "github.ref_name" to be consistent with other workflows @andrewsignori-aot.

bidyashish (approved):
LGTM

.github/workflows/release-deploy-all.yml
```diff
@@ -68,7 +64,7 @@ on:
 
 env:
   NAMESPACE: ${{ secrets.OPENSHIFT_ENV_NAMESPACE }}
-  BUILD_REF: ${{ inputs.gitRef }}
+  BUILD_REF: ${{ github.ref_name }}
```
andrewsignori-aot, line 67: For information, this one would be a problem if we use the `github.ref` instead of `github.ref_name` because it is used internally to name resources deployed to Openshift, I am not wrong.

andrewsignori-aot (approved):
Looks good 👍

PR #4695, opened 2025-05-13
#4490 - FormIO - Enable Editor in Test & PR Process - Fixes and Improvements

andrewsignori-aot:
- Removes the cached form.io definition to allow immediate visualization of saved changes.
- Fix for the content submission of a large payload due to a controller change in the previous PR.
- Add missing links to the content security policy that were preventing some components in the editor not to being displayed.

image

## UI Changes

Added the form path to the auto-complete list, as shown below, to allow differentiation between forms with the same title.

image

image

### DIFF
--- a/sources/packages/backend/apps/api/src/app.module.ts
+++ b/sources/packages/backend/apps/api/src/app.module.ts
@@ -12,6 +12,7 @@ import {
   AuditController,
   ConfigController,
   DynamicFormController,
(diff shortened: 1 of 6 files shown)


dheepak-aot (approved):
Thanks for the fix and the update to the dropdown. 👍

sh16011993 (approved):

Issue #4684, opened 2025-05-08
Inject a new variable to Form and update welcome page


dheepak-aot:
## Technical

- [x] Inject a new variable to form IO to identify a change request application. The new supplementary variable will be based on all `Application Edit Status` related to the change request.
- [x] Make changes to student application form on welcome page for a change request.
- [x] During dry run submission inject this new variable to dry run submission input..

Issue #4317, opened 2025-02-04
Change request information for modified application


Joshua-Lakusta:
As a student/ministry user we want the application to be modified for post COE edits. Excluding the program section from the edit.

PR #4697, opened 2025-05-13
#4317 - Application welcome page update - Change request

dheepak-aot:
# Application welcome page update

Sub Task: #4684 

## Inject new variable to FormIO

- [x] A new variable `isChangeRequestApplication` has been injected to form to indicate if that submission was for a change request. This variable is based on the `Application Edit Status + Context of editing application on change request`  

## FormIO Updates

- [x] Welcome page content of `Student application form(2025-26)` updated with Change request related content which displays when `isChangeRequestApplication = true`
image

image

- [x] The welcome page content is now separated using Panel component for `Change Request` vs `Other application submission` 

## Dry run submission
- [x] Injected the property `isChangeRequestApplication` during the dry run submission from API.


sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
```diff
@@ -55,7 +55,7 @@ export class SaveApplicationAPIInDTO {
    */
   @IsObject()
   @JsonMaxSize(JSON_20KB)
-  data: any;
+  data: ApplicationData & Record<string, unknown>;
```
sh16011993, line 58: Since we are currently only adding `isChangeRequestSubmission`, we can do: `Record<string, boolean>` rather?

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
```diff
@@ -920,6 +923,9 @@ export class ApplicationControllerService {
       payload,
       application.offeringIntensity,
     );
+    // Inject the indicator for change request submission.
+    payload.data.isChangeRequestApplication =
+      options?.isChangeRequestSubmission ?? false;
```
andrewsignori-aot, line 928: Is this association not a responsibility of the `validateSubmitApplicationData`.
My understanding from the `validateSubmitApplicationData` is that it must resolve any payload-specific data-related task.

sources/packages/backend/apps/api/src/route-controllers/application/models/application.dto.ts
andrewsignori-aot: It can be changed to `data: ApplicationData;`.

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
```diff
@@ -534,6 +534,12 @@ export interface ApplicationData {
    * Dependant details.
    */
   dependants?: unknown[];
+
+  /**
+   *
```
andrewsignori-aot, line 539: Please remove the blank comment line.

andrewsignori-aot (commented):
Nice start, minor comments only.

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
dheepak-aot: I too started in the same way, but did the injection here trying to avoid cascading the optional flag to `validateSubmitApplicationData`. I can change.

sources/packages/forms/src/form-definitions/sfaa2025-26-ft.json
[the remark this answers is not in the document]
dheepak-aot: Anticipation was not the intention.  As it was inside a column component which had a column and 2 html components under a column, I felt it will be more simple to look and understand and also grow separately(fringe benefit not the core reason) using 2 different panels.

sources/packages/backend/apps/api/src/route-controllers/application/application.controller.service.ts
dheepak-aot: Changed.

sources/packages/backend/libs/sims-db/src/entities/application.model.ts
andrewsignori-aot: I was actually talking about 538.
image

andrewsignori-aot (approved):

PR #4698, opened 2025-05-13
#4690 - Update Deploy-All to use tag Git Reference - Build-ref Fallback

andrewsignori-aot:
Proposed solution to have the GitHub ref provided from both workflow entries: `workflow_dispatch` and `workflow_call`.
When `workflow_call` is executed, it will be provided by the parent workflow.
When `workflow_dispatch` is executed, it will use the `github.ref_name` as the default value.
_Note:_
1 - The env variable is enough to execute the jobs, but it is not accessible from the reusable workflows `with` context for `deployCamundaDefinitions` and `deployFormioDefinitions`.
2 - Reference for the ternary operation used to set the ENV variable: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example

### DIFF
--- a/.github/workflows/release-deploy-all.yml
+++ b/.github/workflows/release-deploy-all.yml
@@ -145,156 +145,156 @@ jobs:
       - name: Print env
         run: |
           echo Deploy Environment: ${{ inputs.environment }}
-          echo GIT REF: ${{ env.BUILD_REF }}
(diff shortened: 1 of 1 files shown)


dheepak-aot (approved):
Looks good. Thanks for the walkthrough. 👍

Issue #4587, opened 2025-04-08
Content: Student - Update declaration


astridSABC:
Update declaration section for FT application, year 25/26 and onward

**Acceptance Criteria**
- [ ] Update FT application Declaration content highlighted Yellow with the content that can be found here:
https://bcgov.sharepoint.com/:w:/r/teams/03177/Shared%20Documents/General/SABC%20SIMS%20Fulltime_Application%20Declaration%20-%20Final%20Draft.docx?d=w16983802f8e240e2a5eb9b5d2c1da616&csf=1&web=1&e=MyzMfK

<img width="1223" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeS9lQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6ac391c8e26d92a3ba50b8c26febd5e1b2775b4f/image.png" alt="image.png" />

- [ ] Ensure that section II.1 has a link taking student to StudentAid BC website.  Website link: https://studentaidbc.ca/
<img width="1031" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelBlQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c96e37024f3d7f9e4fe51bb17e67a21b4f13a863/image.png" alt="image.png" />

CarlyCotton:
Needs confirming with PY/SD folks - content to be confirmed

astridSABC:
@JerPearson Can you follow-up with SD for the declaration content?

Joshua-Lakusta:
@AnnaPBashkatova Can you take this one over and focus on it as you next ticket please!

AnnaPBashkatova:
Declaration: https://bcgov-my.sharepoint.com/:w:/g/personal/jeremy_pearson_gov_bc_ca/EaosOZISA4lGsUq6ZXVgQ0UB-bpJbgda8xGWQJeB51BoPA?email=Anna.Bashkatova%40gov.bc.ca&wdOrigin=TEAMS-MAGLEV.p2p_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1745946086682&web=1. Sent a follow up email to Sheena for a final approval.

PR #4737, opened 2025-05-23
#4587 - Update Declaration on FT Application

ninosamson:
- updated content on FT application, replacing form.io HTML component with content component, similar to PT declaration tab. 
image


CarlyCotton (approved):
Looks good

Issue #4445, opened 2025-03-07
BUG: Ministry user: Production, Stg, Test: "Deactivate" programs for Ministry user is no longer available


astridSABC:
**Describe the Bug**
"Deactivate" feature is not available for the Ministry user who has appropriate permissions for Production, Test, Staging

**Expected Behaviour**
Ministry user with appropriate permissions can deactivate programs in all environments

Image below is a sample from Institution user.  Ministry user has "view"

<img width="320" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0FxQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--96a7382a73076ecc1400f91946cc1b324dd6d8de/image.png" alt="image.png" />

**Actual Behaviour**
Ministry user with appropriate permissions cannot deactivate programs. 

<img width="317" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNlFxQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--dd2cfc2d28ab5117f3b75ac511df0b4be7296de2/image.png" alt="image.png" />

** Steps To Reproduce**

BC Public institution has an approved program and the ministry user has appropriate permissions to deactivate a program

1. Login to the ministry portal
2. Select "search institutions"
3. Enter institution in operating name field that has an active program (e.g. jiminy U)
4. Click "view"
5. Select "programs"
6. Select any approved program
7. Click "program actions"
8. Note the options

Additional context
Was noticed by Andre during other work and the following was provided. 
<img width="407" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejk0Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--724302dca331f0a8453ecb984a8862c0255c5159/image.png" alt="image.png" />

[...]

PR #4750, opened 2025-05-27
#4445 - Program deactivate action bug fix

dheepak-aot:
### Root Cause
- The common component has a condition to disable the `Deactivate` button based on the institution user data from institution store. This condition results in a runtime error when the component is loaded from a ministry user view because of the absence of Institution store and so the institution user data. 

## Solution
- Validation of institution user data is applied conditionally when the user type is `Institution` skipping the same for a ministry view. 
Note: We generally avoid using the auth client type based logic in the components `AuthService.shared.authClientType === ClientIdType.Institution`. But this component being very old and already using similar condition in other places, the auth client type is used here as quick solution. 

### Outcome Ministry

image

### Outcome Institution(No changes introduced here by PR)

image

### DIFF
--- a/sources/packages/web/src/components/common/ProgramDetails.vue
+++ b/sources/packages/web/src/components/common/ProgramDetails.vue
@@ -33,12 +33,7 @@
           >
             <template #="{ notAllowed }">
               <v-list-item
-                :disabled="
-                  !educationProgram.isActive ||
-                  educationProgram.isExpired ||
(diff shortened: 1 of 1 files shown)


andrewsignori-aot (approved):
Nice fix, looks good 👍

bidyashish (approved):
LGTM @dheepak-aot

Issue #4664, opened 2025-05-02
PSI Integration Changes - Filename and Folders


CarlyCotton:
**Acceptance Criteria**
- [ ] ECE Request Files:
    - [ ] Generate files in the OUT folder (not in the OUT/AUAA folder)
    - [ ] **NEW AC:** Filename change to Filename change to CON-008-AJAA.txt
        - The plan is to only keep one outgoing file as each is a snapshot of all available records. If the same filename exists in the folder, we should overwrite.
        - Need to maintain some history in the logs. 
- [ ] ECE Response Files:
    - [ ] Files will be in the IN folder (not in the IN/AUAA folder)
    - [ ] Expected filename change to CONR-008-AJAA-20230130-193312.TXT (CONR-008-[Inst Code]-YYYYMMDD-HHMMSS.TXT)
    - [ ] Change Archive folder to IN/Archive
- [ ] IER Files:
    - [ ] Generate files in the OUT folder (not in the OUT/AUAA folder)
    
*Technical Notes*
AJAA, 4 Digit Alphabet is Instituition Code 

- [ ] queue-consumers/src/processors/schedulers/institution-integration (Work on all 3 Queue Consumers)

CarlyCotton:
Spoke to Harry - plan to change file to CON-008-AJAA.txt and overwrite if file not picked up. Plan to keep only one CON file at any given time for each institution in the folder.

ninosamson:
Noting removal of this AC on Jun6/25 standup.

PR #4829, opened 2025-06-16
#4664 - PSI Integration Changes - Filename and Folders - Part 3

bidyashish:
- [x] Changes the file path for IER so that files are generated in the OUT folder instead of institution specific folders

**Demo**
image


andrewsignori-aot (approved):
Looks good 👍

Issue #4476, opened 2025-03-13
Copy of Ecert Cancellation - New Status & File Integration (1/2)


CarlyCotton:
**User Story**
As a ministry user, I need to be able to tell SIMS to ignore a previously sent ecert when generating **new** ecerts for the same application.
As a system, SIMS needs to be able to read in two flat files (pt & ft) to change the status of a sent disbursement. Recognizing that they may have been Cancelled or Rejected in the meantime (or not sent from SIMS).

### New status
- [ ] Create new disbursement status to indicate it has been 'Rejected' (This is to make a distinction between system cancellation pre-'Sent' and cancellation post-'Sent')
- [ ] Future processes for the same application should disregard those previously sent amounts similar to 'Cancelled' disbursements.
- [ ] New columns to track date of rejection and who rejected (user or system user)
- [ ] Update production `disbursement_schedules` records where `date_sent` is not null and `disbursement_schedule_status`=Cancelled
<img width="1344" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeWtFQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--423f171631ecdcdb2e978d0c757b7f87a004f003/image.png" alt="image.png" />

### File integration
- [ ] New incoming file integration for two files:
    - PEDU.PBC.CAN.PTECRT.yyyymmdd.sss
    - PEDU.PBC.CAN.ECRT.yyyymmdd.sss
- [ ] Files are received once per month.
- [ ] Records on the files are per certificate/document number and include a SIN. These could include records from SIMS or Legacy systems.
- [ ] The status of the disbursements could already have been 'Rejected'.
- [ ] Layout:
```
     03  CNEC-HEADER-RECORD.                                    
         05  CNEC-HDR-RECTYP                    PIC X(03).      
         05  CNEC-HDR-ORIGINATOR                PIC X(04).      
         05  CNEC-HDR-FILE-TITLE                PIC X(40).      
         05  CNEC-HDR-CREATE-DT                 PIC 9(08).      
         05  CNEC-HDR-CREATE-TIME               PIC X(04).      
         05  FILLER                             PIC X(41).      
     03  CNEC-DETAIL-RECORD REDEFINES CNEC-HEADER-RECORD.       
         05  CNEC-RECTYP                        PIC X(03).      
[...]

PR #4832, opened 2025-06-17
#4476 - Add overaward added date

dheepak-aot:
## Audit column assignment on overaward reversal

- During overaward reversal, while `addedBy` was populated, `addedDate` is not assigned and fixed the same in PR.


andrewsignori-aot (approved):
👍

Issue #4818, opened 2025-06-12
Parent Declare - Supporting User Portal Update


ninosamson:
## **User Story**
As a parent supporting user, I would like to report my information on the appropriate application for which my child has identified me. 

##  Update search for parent only 
- [ ]  Partner search will remain the same 
- [ ]  Remove DOB search field and replace with "Full name" of parent as provided in student application. 
- [ ]  Label for input - "Parent's Full Name" 
- [ ]  Require exact match for "Full  name" as submitted in the application. Case insensitive. 
- [ ]  Fix "Application number" error message - update to "Application number is required" 
- [ ]  Only parents reported as "isAbleToReport" will be able to provide data 

## Update date picker for partner search 
- [ ]  Update date picker to use vuetify date picker

Joshua-Lakusta:
@ astrid we will want to test a few strange names including hypenated names and other symbols and really long names for testing

PR #4891, opened 2025-07-09
#4818 - Parent Declare - Supporting User Portal Update

bidyashish:
### Refactor

- Made the workflow notification for supporting user data submission conditional and remove DB Lock.
- Ensured the completion message is sent only after the database update is successfully verified to prevent race conditions.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
@@ -27,7 +27,6 @@ import {
   STUDENT_APPLICATION_NOT_FOUND,
   SUPPORTING_USER_ALREADY_PROVIDED_DATA,
   SUPPORTING_USER_IS_THE_STUDENT_FROM_APPLICATION,
-  SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA,
 } from "../../services/supporting-user/constants";
 import { getSupportingUserFormType } from "../../utilities";
 import {
@@ -267,10 +266,7 @@ export class SupportingUserSupportingUsersController extends BaseController {
 
     if (updateResult.affected !== 1) {
       throw new UnprocessableEntityException(
-        new ApiProcessError(
-          `The application is not expecting supporting information from a ${payload.supportingUserType} to be provided at this time.`,
-          SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA,
-        ),
+        `Error while updating the supporting user details for ${supportingUser.id}. Number of affected rows was ${updateResult.affected}, expected 1.`,
       );
     }
     // Send the message to the workflow to complete the supporting user process.
--- a/sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
@@ -252,23 +252,26 @@ export class SupportingUserSupportingUsersController extends BaseController {
         address: addressInfo,
       };
 
-      const updatedUser = await this.supportingUserService.updateSupportingUser(
-        supportingUser.application.id,
-        payload.supportingUserType,
-        user.id,
-        {
-          contactInfo,
-          sin: submissionResult.sin,
-          hasValidSIN: submissionResult.hasValidSIN,
-          birthDate: userToken.birthdate,
-          supportingData: submissionResult.supportingData,
-          userId: user.id,
-        },
-      );
+      const updateResult =
+        await this.supportingUserService.updateSupportingUserReportedData(
+          supportingUser.id,
+          payload.supportingUserType,
+          user.id,
+          {
+            contactInfo,
+            sin: submissionResult.sin,
+            hasValidSIN: submissionResult.hasValidSIN,
+            birthDate: userToken.birthdate,
+            supportingData: submissionResult.supportingData,
+            userId: user.id,
+          },
+        );
 
-      await this.workflowClientService.sendSupportingUsersCompletedMessage(
-        updatedUser.id,
-      );
+      if (updateResult.affected > 0) {
+        await this.workflowClientService.sendSupportingUsersCompletedMessage(
+          supportingUser.id,
+        );
+      }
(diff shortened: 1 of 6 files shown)


sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
```diff
@@ -252,23 +252,26 @@ export class SupportingUserSupportingUsersController extends BaseController {
...
+            birthDate: userToken.birthdate,
+            supportingData: submissionResult.supportingData,
+            userId: user.id,
+          },
+        );
 
-      await this.workflowClientService.sendSupportingUsersCompletedMessage(
-        updatedUser.id,
-      );
+      if (updateResult.affected > 0) {
```
dheepak-aot, line 270: If we are checking for updated result, then we must check for `if (result.affected !== 1) ` and throw error instead.

sources/packages/backend/apps/api/src/services/supporting-user/supporting-user.service.ts
```diff
@@ -28,90 +26,48 @@ export class SupportingUserService extends RecordDataModelService<SupportingUser
...
+   * @param supportingUserType The type of the user, used for crafting the error message.
+   * @param auditUserId The user performing the update.
+   * @param updateInfo The information to be updated.
+   * @returns The result of the update operation.
    */
-  async updateSupportingUser(
-    applicationId: number,
+  async updateSupportingUserReportedData(
+    supportingUserId: number,
     supportingUserType: SupportingUserType,
```
dheepak-aot, line 38: https://github.com/bcgov/SIMS/pull/4891/files#r2196153681

sources/packages/backend/apps/api/src/services/supporting-user/supporting-user.service.ts
```diff
@@ -28,90 +26,48 @@ export class SupportingUserService extends RecordDataModelService<SupportingUser
...
+    const updateResult = await this.repo.update(
+      { id: supportingUserId, user: { id: IsNull() } },
+      {
+        contactInfo: updateInfo.contactInfo,
+        sin: sinWithNoSpaces,
+        birthDate: updateInfo.birthDate,
+        supportingData: updateInfo.supportingData,
+        user: { id: updateInfo.userId } as User,
+        personalInfo,
+        modifier: { id: auditUserId } as User,
```
dheepak-aot, line 58: To simplify we can declare an audit user variable and use it. 

`const auditUser = { id: auditUserId } as User;`

sources/packages/backend/apps/api/src/services/supporting-user/supporting-user.service.ts
```diff
@@ -28,90 +26,48 @@ export class SupportingUserService extends RecordDataModelService<SupportingUser
...
+      },
+    );
+    // If no records are updated, it means the supporting user record either
+    // did not exist or was already updated by another process.
+    if (updateResult.affected === 0) {
+      throw new CustomNamedError(
+        `The application is not expecting supporting information from a ${supportingUserType} to be provided at this time.`,
+        SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA,
+      );
```
dheepak-aot, line 67: As a service it is better to just return the update result leaving it to the consumer to decide what needs to be done on the update result.

sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
```diff
@@ -252,23 +252,26 @@ export class SupportingUserSupportingUsersController extends BaseController {
...
+        await this.workflowClientService.sendSupportingUsersCompletedMessage(
+          supportingUser.id,
+        );
+      }
     } catch (error) {
       if (error.name === SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA) {
```
dheepak-aot, line 276: Based on this comment https://github.com/bcgov/SIMS/pull/4891/files#r2196160850 the catch can be removed as there will not be any translation of error required.

sources/packages/backend/apps/api/src/services/supporting-user/supporting-user.service.ts
bidyashish: `supportingUserType `. removed.

sources/packages/backend/apps/api/src/services/supporting-user/supporting-user.service.ts
bidyashish: Error msg removed.

sources/packages/backend/apps/api/src/route-controllers/supporting-user/supporting-user.supporting-users.controller.ts
bidyashish: Hi Dheepak,
`SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA`. is not used anywhere in Backend, should error msg be updated as below instead?

```typescript
if (updateResult.affected !== 1) {
        new ApiProcessError(
          `The application is not expecting supporting information from a ${payload.supportingUserType} to be provided at this time.`,
          SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA,
        ),
      );
    }
```

dheepak-aot: SUPPORTING_USER_TYPE_ALREADY_PROVIDED_DATA is not relevant to the error that we throw here. If it is not used anywhere, then the error const can be deleted.

sources/packages/backend/apps/api/src/services/supporting-user/supporting-user.service.ts
```diff
@@ -28,90 +24,40 @@ export class SupportingUserService extends RecordDataModelService<SupportingUser
...
-      const updatedUser = await transactionRepo.save(userToUpdate);
-      await queryRunner.commitTransaction();
-      return updatedUser;
-    } catch (error) {
-      await queryRunner.rollbackTransaction();
-      throw error;
-    } finally {
-      await queryRunner.release();
-    }
+    return updateResult;
```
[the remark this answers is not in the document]
bidyashish: removed await.

Issue #4530, opened 2025-03-25
CAS Invoicing - GL Date


CarlyCotton:
**User Story**
As a ministry user, I want to ensure the GL Date being sent to CAS is the same as the date being shown on the batch report. The GL Date should be the approval date of the batch to ensure we send during an open GL period.

**Acceptance Criteria**
- [x] Change the GL Date in the batch report to **Date invoice sent to CAS** (if exists) and "n/a" if not sent yet.
- [x] Change the GL Date when we send invoices to CAS to **Date invoice sent to CAS**
- [x] Use BC Time (PST) in API Call & Report
- [x] Any invoices in the report that go through the 'Manual Intervention' path the GL Date should remain NA
- [x] For existing invoices that have already been sent:
    - [x] Any that were in the 'Manual Intervention' path, set GL Date to NA
    - [x] Any that were successfully sent, we can use status updated date (which will reflect the date sent)

**Additional Context**
GL Date will be the date that the invoice is sent to CAS by SIMS.
- This will show in the Batch report when sent (after approval), otherwise will be “n/a”
- This could result in a batch having invoices with different GL Dates.

PR #4896, opened 2025-07-10
#4530 - CAS GL date update

dheepak-aot:
# CAS GL Date update

## GL date sent to CAS

- [x] The `GL date` that is set in the payload of CAS API call has been `CAS invoice batch - creation date`, which is now updated to the date when the CAS API is called to send the invoice(the current date at the moment is converted to BC date). 
- [x] As we do not have a particular column to capture the date that an invoice is successfully sent to CAS, a new column `date_sent` is created in `sims.cas_invoices` table.
- [x] For those invoices which are already sent to CAS, the date_sent is set from `invoice_status_updated_on`(mainly for handling existing production data).

## GL date displayed in report
- [x] The GL Date displayed in the report is not updated to use `date_sent` if present and `n/a` otherwise.

Report with pending invoice. 
image

Report with Sent invoice,

image

## Rollback evidence

image

## E2E Tests

Updated E2E test for both API and Queue Consumers


sources/packages/backend/libs/sims-db/src/entities/cas-invoice.model.ts
```diff
@@ -111,4 +111,14 @@ export class CASInvoice extends RecordDataModel {
     type: "varchar",
   })
   errors?: string[];
+
```
andrewsignori-aot, line 114: Please keep the file standard and remove the blank line.

PR #4920, opened 2025-07-11
#4849 - Update Living Situation Questions 

ninosamson:
Details for this PR: 
- [ ] moves partner living situation question from "Partner" tab to "Financial Information" tab, "Living Situation" section

image

- [ ] removed conditional logic for the "Living Situation" panel and implemented logic on questions as the questions (self contained suite, partner living situation) are mutually exclusive
- [ ] no cleanup required for partner data as the removed question does not have anything mapped in consolidated data

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2025-26-ft.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2025-26-ft.json
@@ -5554,34 +5554,6 @@
               "input": true
             },
             {
-              "label": "Will your partner be living with you during your study period?",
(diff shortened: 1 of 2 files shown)


ninosamson:
> Looks good @ninosamson
> 
> just wanted to confirm if the condition to hide room and board question on marital status Married is captured in a ticket ?

Yes - the AC is actually on the "Room and Board" ticket - ZH 4610. Had to move the conditional logic from the panel to the question level as the questions are mutually exclusive.

sh16011993 (approved):
LGTM @ninosamson

PR #4926, opened 2025-07-14
#4887 - Add parent income exception questions

ninosamson:
- This PR adds the current year income question for parents. For each parent, the option of providing current year income is available. 
- Behaviour and logic is similar to partner income exception. 

image


sources/packages/forms/src/form-definitions/studentexceptions.json
```diff
@@ -411,11 +414,40 @@
             }
           ],
           "tableView": false,
-          "type": "well",
           "hideLabel": true,
-          "lockKey": true,
+          "lockKey": true
+        },
+        {
+          "label": "Current Year Parent Income",
```
sh16011993, line 421: Minor: "Current Year Parent Income Exception"

sources/packages/backend/workflow/src/workflow-definitions/load-assessment-consolidated-data.bpmn
```diff
@@ -109,6 +109,7 @@
           <zeebe:header key="studentDataStudentParentsCppLine30800" value="$.data.studentParentsCppLine30800" />
           <zeebe:header key="studentDataStudentParentsCppLine31000" value="$.data.studentParentsCppLine31000" />
           <zeebe:header key="studentDataStudentParentsEiLine31200" value="$.data.studentParentsEiLine31200" />
+          <zeebe:header key="studentDataStudentCurrentYearParentIncome" value="$.data.parents.currentYearParentIncome" />
```
dheepak-aot, line 112: `$.data.parents` is an array. Hence it does not have a property by name `currentYearParentIncome`. 

We instead should add the additional property mapped into the array here. 

image

image

sources/packages/forms/src/form-definitions/studentexceptions.json
```diff
@@ -411,11 +414,40 @@
...
+          "key": "currentYearPartnerIncomeApplicationException1",
```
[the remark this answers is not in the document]
ninosamson: Sorry - fixed it.

sources/packages/forms/src/form-definitions/studentexceptions.json
ninosamson: Updated, thanks!

sources/packages/forms/src/form-definitions/sfaa2025-26-ft.json
```diff
@@ -5994,6 +5994,448 @@
...
+                              "shortcut": ""
+                            },
+                            {
+                              "value": "other",
+                              "label": "Other",
+                              "shortcut": ""
+                            }
+                          ],
+                          "validate": {
+                            "required": true
```
dheepak-aot, line 6209: Same here. 
image

sources/packages/forms/src/form-definitions/studentexceptions.json
```diff
@@ -411,11 +414,40 @@
...
           ],
           "tableView": false,
-          "type": "well",
           "hideLabel": true,
-          "lockKey": true,
+          "lockKey": true
+        },
+        {
+          "label": "Current Year Parent Income Exception",
+          "key": "currentYearParentIncomeApplicationException",
```
dheepak-aot, line 422: In runtime the component key of the component is created with `1` appended in the end. I believe it is because it conflicts with the key of HTML component in it. It is important that the well component has the key matching with the hidden component in application. Can we change the key of html component?

image

sources/packages/forms/src/form-definitions/sfaa2025-26-ft.json
ninosamson: Updated, thanks!

sources/packages/forms/src/form-definitions/studentexceptions.json
ninosamson: Changed the html key, thanks!

dheepak-aot (approved):
Thanks for making the changes @ninosamson Looks good. 👍

PR #4936, opened 2025-07-17
#4888 - Parent CY Income: Update camunda

CarlyCotton:
Updated parent 1 / 2 total income calculation _[Camunda treats the first item as [1] and not [0]](https://docs.camunda.io/docs/components/modeler/feel/language-guide/feel-list-expressions/)_
:
image

New test cases for parent income (`calculatedDataTotalParentIncome`, `calculatedDataTotalParentDeductions`, `calculatedDataTotalFamilyIncome`, `calculatedDataTotalNetFamilyIncome`:
- 1 Parent: Parent reported income + deductions more than max
- 1 Parent: Parent reported income + deductions less than max
- 1 Parent: Parent reported income + deductions less than max + CRA income value
- 1 Parent: Parent reported income + deductions less than max + Current Year income value
- 2 Parents: Parent reported income + deductions less than max


sources/packages/backend/workflow/src/workflow-definitions/fulltime-assessment-2025-2026.bpmn
andrewsignori-aot, line 1: Is there a problem if we execute the calculations in the existing flows for each scenario? It seems that it would follow the existing approach.
If the student should be calculated either way, we can keep the student only.

image

Issue #3059, opened 2024-04-01
Software Package Updates 


andrewsignori-aot:
# **Acceptance criteria**

## Upgrade 
- [ ] Assess and upgrade the following as needed. If there are breaking changes, separate ticket may be required. 
- [ ] Create individual PRs for the individual ACs as much as possible.
    - [x] Form.io
    - [x] Zeebe/Camunda SDK
    - [ ] Typescript
    - [x] Resolve "High" severity security issues by upgrading nest/multer 
    - [ ] Nodejs - nice to have minimal 
    - [ ] Bull
    - [ ] Redis
    - [x] ClamAV 
    - [ ] esLint
    
## Clean-up 
- [x] Remove jest related packages. It is not used right now and can be added later if required.
- [x] Remove `vue-class-component`. Use in a single place and the assumption is that it is no longer needed.
- [x] Remove `@vee-validate/rules`.
- [x] Check if any other package can be removed.

## Future Update Processes (stretch) 
- [x] Create recurring alert or warning when packages are out of date - could potentially use github dependabot 
    - Reference: 
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates
- [ ] Create wiki documentation for maintaining packages

ninosamson:
Consider time-boxed effort to upgrade packages and raise ticket for remaining packages.

CarlyCotton:
Considering pulling nestjs due to flagged security level into it's own ticket for 2.9.

andrewsignori-aot:
ClamAV is already using the lastes imge available from BC Gov: https://github.com/bcgov/common-hosted-clamav-service/pkgs/container/clamav-unprivileged

andrewsignori-aot:
Redis cluster current helm package is 11.4.1 (redis-cluster 7.4.2).
Upgrade is available to helm package 12.0.13  (redis-cluster 8.0.3).

PR #4953, opened 2025-07-21
#3059 - Software Package Updates (packages cleanup)

andrewsignori-aot:
- Removed packages and test-related packages from the Web App that are not used right now. If they are required, they can be added later.
- Did a change to remove the below warning message. _Note:_ The web app uses only the NODE_ENV and no other env variable was defined in the deployment.

image

### DIFF
--- a/sources/packages/web/src/router/index.ts
+++ b/sources/packages/web/src/router/index.ts
@@ -30,7 +30,7 @@ const routes: Array<RouteRecordRaw> = [
 ];
 
 const router = createRouter({
-  history: createWebHistory(process.env.BASE_URL),
+  history: createWebHistory(),
   routes,
 });
 
--- a/sources/packages/web/src/services/http/common/HttpClient.ts
+++ b/sources/packages/web/src/services/http/common/HttpClient.ts
@@ -1,8 +1,6 @@
 import axios from "axios";
 
-const apiUrl =
-  process.env.NODE_ENV === "production" ? "/api" : "http://localhost:3000/api";
-
+const apiUrl = import.meta.env.PROD ? "/api" : "http://localhost:3000/api";
 const axiosOptions = { baseURL: apiUrl };
 const axiosInstance = axios.create(axiosOptions);
 export default axiosInstance;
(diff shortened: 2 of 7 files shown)


sources/packages/web/vite.config.ts
andrewsignori-aot, line 1: The file changes related to this PR are:
- Added the parameter `defineConfig(({ mode })`.
- The below configuration:
```ts
define: {
  "process.env.NODE_ENV": JSON.stringify(mode),
},
```

sources/packages/web/src/router/index.ts
```diff
@@ -30,7 +30,7 @@ const routes: Array<RouteRecordRaw> = [
 ];
 
 const router = createRouter({
-  history: createWebHistory(process.env.BASE_URL),
+  history: createWebHistory(),
```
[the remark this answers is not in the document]
andrewsignori-aot: The web app does not use a BASE_URL at all.

bidyashish (approved):
LGTM

sources/packages/web/src/services/http/common/HttpClient.ts
```diff
@@ -1,8 +1,6 @@
 import axios from "axios";
 
-const apiUrl =
-  process.env.NODE_ENV === "production" ? "/api" : "http://localhost:3000/api";
-
+const apiUrl = import.meta.env.PROD ? "/api" : "http://localhost:3000/api";
```
dheepak-aot, line 3: 👍

sources/packages/web/src/types/vuetify.ts
```diff
@@ -17,7 +15,7 @@ export interface ErrorMessage {
 /**
  * Vuetify v-form type definition.
  */
-export type VForm = Vue & {
```
dheepak-aot, line 20: What is the reason for removing Vue. I believe it does not add any difference right?

andrewsignori-aot: The object belonged to the removed package and did not cause any harm, hence it was removed.

Issue #4934, opened 2025-07-16
FT Assessment: Interface Policy 


CarlyCotton:
- [x] Student triggered Interface Policy
    - [x] studentDataIncomeAssistanceAmount >=1500
- [x] Partner triggered Interface Policy
    - [x] calculatedDataPartnerBCEAIncomeAssistanceAmount >=1500
- [x] Interface Policy changes Need (Fed & Prov) if triggered
- [x] Add intermediate values for EDW & IER purposes (new ticket to bring into gateway/ier)
- [x] Test Cases for Interface Policy being triggered

PR #4956, opened 2025-07-21
#4934 - FT Assessment: Interface Policy

CarlyCotton:
- Converted the interface policy path into a "flag" calculated variable to support reporting out to IER/EDW
- Variables to support reporting now it contains:
    - `calculatedDataInterfaceEducationCosts`
    - `calculatedDataInterfaceChildCareCosts`
    - `calculatedDataInterfaceTransportationAmount`
    - `calculatedDataInterfaceAdditionalTransportationAmount`
    - `calculatedDataInterfaceNeed`
image

- Test Cases added for Interface Policy being applied:
    - Student >$1500 (positive flag)
    - Student <$1500 (negative flag)
    - Student: nothing declared (negative)
    - Married: Student <$1500 + Partner >$1500 (positive)
    - Married: Student <$1500 + Partner <$1500 (negative)

### DIFF
--- a/sources/packages/backend/workflow/src/workflow-definitions/fulltime-assessment-2025-2026.bpmn
+++ b/sources/packages/backend/workflow/src/workflow-definitions/fulltime-assessment-2025-2026.bpmn
@@ -326,6 +326,7 @@ dmnFullTimeIncomeThreshold</bpmn:text>
         <bpmn:flowNodeRef>Gateway_0pktb33</bpmn:flowNodeRef>
         <bpmn:flowNodeRef>Gateway_0imfp5r</bpmn:flowNodeRef>
         <bpmn:flowNodeRef>Activity_1wciqrl</bpmn:flowNodeRef>
+        <bpmn:flowNodeRef>Event_0a81z5w</bpmn:flowNodeRef>
         <bpmn:childLaneSet id="LaneSet_0v6vyq5">
           <bpmn:lane id="Lane_0l32w49" name="Assessed Need and Eligility">
             <bpmn:flowNodeRef>Event_1ki3afc</bpmn:flowNodeRef>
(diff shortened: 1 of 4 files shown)


sources/packages/backend/workflow/src/workflow-definitions/fulltime-assessment-2025-2026.bpmn
```diff
@@ -2187,6 +2182,27 @@ The remaining limit in the program year is calculated with a floor of 0.</bpmn:d
...
+        </zeebe:ioMapping>
+      </bpmn:extensionElements>
+      <bpmn:incoming>Flow_05xztlk</bpmn:incoming>
+      <bpmn:outgoing>Flow_0eae1bf</bpmn:outgoing>
+    </bpmn:intermediateThrowEvent>
+    <bpmn:sequenceFlow id="Flow_1k0i5o3" sourceRef="Event_1fdxon1" targetRef="Gateway_0u7w3hd" />
+    <bpmn:intermediateThrowEvent id="Event_1fdxon1" name="Interface Policy Flag">
+      <bpmn:extensionElements>
+        <zeebe:ioMapping>
+          <zeebe:output source="=if (studentDataIncomeAssistanceAmount &#62;= 1500) or (calculatedDataPartnerBCEAIncomeAssistanceAmount &#62;= 1500)&#10;then &#34;yes&#34;&#10;else &#34;no&#34;" target="calculatedDataInterfacePolicyApplies" />
```
andrewsignori-aot, line 2200: We use "yes"/"no" for items coming from the forms, but from calculated variables, we can use true/false, for instance, when the eligibilities are calculated.
Is there any benefit in using "yes"/"no" as a boolean for this scenario?

andrewsignori-aot (commented):
Great work, one minor comment only.

sources/packages/backend/workflow/src/workflow-definitions/fulltime-assessment-2025-2026.bpmn
CarlyCotton: I went back and forth on this, landing on yes/no because it's reported out on IER as a Y/N flag and I believe this is the same format as expected with EDW. Happy to swap it to a true/false if that better aligns with an internal calculation though.

sh16011993 (approved):
LGTM @CarlyCotton

sources/packages/backend/workflow/src/workflow-definitions/fulltime-assessment-2025-2026.bpmn
andrewsignori-aot: IER probably has conversions in place to deal with it, and the workflow does not need to be concerned about it.
For instance, if the same data needed to be consumed by IER, EDW, SDPR, and others, there is nothing that would ensure they would consume it in the same way.
The workflow should persist the data as it is meaningful for the workflow itself.
If it is a "yes"/"no" that can become a "maybe" in the future, it would be ok. If it is a boolean, should be a boolean😉

Issue #4571, opened 2025-04-02
Aviation: Change program questions


CarlyCotton:
**User Story**
As an institution user, I need to declare if my program has aviation components to them and if so, the program needs to be reviewed by IPDT. This will also allow me to add offerings that have specific types of aviation credentials without causing a mismatch of data. Currently I would need to select 'Hours' (instead of 'Credits') to go down the aviation path.

**Acceptance Criteria**
- [ ] Move current questions in program under 'Hours' related to Aviation _(See Mockup 1 in Additional Context)_
- [ ] Add questions as standalone like WIL/ESL at bottom of form, above "Declaration":  _(See Mockup 2 in Additional Context)_ 
    - [ ] `Is this an Aviation program?` Radio options: Yes/No
    - [ ] If Aviation=Yes show two questions: 
        - [ ] `Does this program contain a minimum of 15 instructional hours per week?`
            - [ ] If Instructional hours = No: Display the warning banner that currently exists for this question (see Mockup 1) it should not depend on the course load type (hours/credits)
        - [ ] `Which credential types are included? (Select all that apply)` Options: Commercial Pilot Training, Instructor's Rating, Endorsements, Private Pilot Training
            - [ ] Minimum of one checkbox required for submission.
            - [ ] If Private Pilot Training checked: Display a warning banner
            > **This program is ineligible for StudentAid BC funding**
            > StudentAid BC does not provide any assistance to students for Private Pilot Training.
- [ ] When program has Aviation="Yes", always send to "Pending" (this happens even if no banners are triggered)
- [ ] reuse existing "is_aviation_program" column and set it as " no" when "null" 

**Additional Context**
On deploy - confirm with IPDT the old version aviation programs can be deactivated.
Program Mockup:
<img width="480" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOUdUQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a5f878aaf0fe8bacfb98c226cdafd8f46847e8f4/image.png" alt="image.png" />

Questions mockup (only for example purposes, do not include "None of the above" please see ACs)
[...]

PR #4963, opened 2025-07-22
#4571 - Aviation Change Program Questions - Part 2

sh16011993:

- Added migrations for the following:

       1. Added the new `Credential Types Aviation` column
       2. Updated the `isAviationProgram` to `no` where `null` for existing records

- Updated the `api` and `web` code logic to save, update and retrieve the values in edit and read-only mode from this newly created column `Credential Types Aviation` to be displayed to the UI

### **Screenshots:**

**Read-only mode institutions program:**

image

--------------------------------------------------------------------------------------------------------------------------------------------

**Migrations Rollback:**

image

--------------------------------------------------------------------------------------------------------------------------------------------

image

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
@@ -214,6 +214,7 @@ export class EducationProgramControllerService {
       institutionProgramCode: program.institutionProgramCode,
       minHoursWeek: program.minHoursWeek,
       isAviationProgram: program.isAviationProgram,
+      credentialTypesAvi: program.credentialTypesAviation,
       minHoursWeekAvi: program.minHoursWeekAvi,
       entranceRequirements: {
         hasMinimumAge: program.hasMinimumAge,
--- a/sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
@@ -214,7 +214,7 @@ export class EducationProgramControllerService {
       institutionProgramCode: program.institutionProgramCode,
       minHoursWeek: program.minHoursWeek,
       isAviationProgram: program.isAviationProgram,
-      credentialTypesAvi: program.credentialTypesAviation,
+      credentialTypesAviation: program.credentialTypesAviation,
       minHoursWeekAvi: program.minHoursWeekAvi,
       entranceRequirements: {
         hasMinimumAge: program.hasMinimumAge,
--- a/sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Add-credential-types-aviation-column.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Add-credential-types-aviation-column.sql
@@ -1,0 +1,13 @@
+ALTER TABLE
+  sims.education_programs
+ADD
+  COLUMN credential_types_aviation jsonb;
+
+COMMENT ON COLUMN sims.education_programs.credential_types_aviation IS 'A jsonb column to store various credential types related to aviation programs.';
+
+ALTER TABLE
+  sims.education_programs_history
+ADD
+  COLUMN credential_types_aviation jsonb;
+
+COMMENT ON COLUMN sims.education_programs_history.credential_types_aviation IS 'A jsonb column to store various credential types related to aviation programs.';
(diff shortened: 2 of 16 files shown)


sources/packages/backend/apps/api/src/route-controllers/education-program/education-program.controller.service.ts
```diff
@@ -214,6 +214,7 @@ export class EducationProgramControllerService {
       institutionProgramCode: program.institutionProgramCode,
       minHoursWeek: program.minHoursWeek,
       isAviationProgram: program.isAviationProgram,
+      credentialTypesAvi: program.credentialTypesAviation,
```
dheepak-aot, line 217: for consistency can we follow the same key in API(form) similar to entity `credentialTypesAviation` ?

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
```diff
@@ -218,6 +218,18 @@ export class EducationProgram extends RecordDataModel {
...
 
+  /**
+   * A jsonb column to store various credential types related to aviation programs.
+   */
+  @Column({
+    name: "credential_types_aviation",
+    type: "jsonb",
+    nullable: true,
+    comment:
+      "A jsonb column to store various credential types related to aviation programs.",
```
dheepak-aot, line 229: We generally do not use the comment attribute in the decorator.

sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Add-credential-types-aviation-column.sql
```diff
@@ -0,0 +1,13 @@
...
+  COLUMN credential_types_aviation jsonb;
+
+COMMENT ON COLUMN sims.education_programs.credential_types_aviation IS 'A jsonb column to store various credential types related to aviation programs.';
+
+ALTER TABLE
+  sims.education_programs_history
+ADD
+  COLUMN credential_types_aviation jsonb;
+
+COMMENT ON COLUMN sims.education_programs_history.credential_types_aviation IS 'A jsonb column to store various credential types related to aviation programs.';
```
dheepak-aot, line 13: Can we be less technical on the comment. i.e. excluding the jsonb. It can be simple as `Aviation program credential types.`

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
```diff
@@ -218,6 +218,18 @@ export class EducationProgram extends RecordDataModel {
   })
   isAviationProgram?: string;
 
+  /**
+   * A jsonb column to store various credential types related to aviation programs.
```
dheepak-aot, line 222: https://github.com/bcgov/SIMS/pull/4963/files#r2223818052

dheepak-aot:
Are the E2E tests going to be part of upcoming PR?

dheepak-aot:
Finished my review @sh16011993 Please have a look at the comments.

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
sh16011993: Isn't it a good idea to keep it? I got this as a suggestion from Copilot.

sources/packages/backend/apps/db-migrations/src/sql/EducationPrograms/Rollback-update-is-aviation-program-to-no.sql
```diff
@@ -0,0 +1,10 @@
+UPDATE
+  sims.education_programs
+SET
+  is_aviation_program = NULL
+WHERE
+  course_load_calculation = 'credit'
+  OR (
+    course_load_calculation = 'hours'
+    AND min_hours_week = 'yes'
+  );
```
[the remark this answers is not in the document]
dheepak-aot: Never mind, I can see what you have done here.

sources/packages/backend/libs/sims-db/src/entities/education-program.model.ts
dheepak-aot: If we consider to follow, then there should be a DEV agreement to follow on all the entity model properties. I do not see a benefit of using the attribute which has the same comment as variable comment. 

@andrewsignori-aot  @bidyashish do you guys think otherwise?

andrewsignori-aot: This comment property is probably intended to be used when typeorm is responsible for creating the DB. In this scenario, it would be the comment added to the DB column.
It would not bring any value in our context. Please avoid using it.

Issue #4665, opened 2025-05-02
Workflow Permissions 


ninosamson:
**As a SIMS team, we need to address code scan findings and resolve potential vulnerabilities.**

## Acceptance Criteria

- [ ] Add explicit permissions to GitHub actions jobs, as suggested by GitHub code scanning alerts

[https://github.com/bcgov/SIMS/security/code-scanning/61](https://github.com/bcgov/SIMS/security/code-scanning)

PR #4992, opened 2025-07-24
#4665 - Workflow Permissions - Fix Missing Permission

andrewsignori-aot:
Replace the `read` with the `write` permission to allow the tag creation (the write permission includes the `read` permission).


dheepak-aot (approved):
👍

Issue #4731, opened 2025-05-23
Add loading indicator for Form.io 


ninosamson:
**User story**
As a SIMS user, I need to see a loading indicator so that I can't interact with the application until the components have been loaded. 

**Acceptance Criteria**
- [x] Add a global loading indicator when loading form.io definitions and data 
- [x] Skeleton loader is displayed until both form definitions and data have been returned 
- [x] Consider the fact that forms may be cached and/or data may not be needed for a particular form

PR #5008, opened 2025-07-31
#4731 - Form-io loader

dheepak-aot:
## Form io loader refactor

- [x] Updated the `formio` and hence the `formio-container` component to receive a state of `inital data loading is complete` by the new parameter `is-data-ready` which indicates that all the data required to load the form is ready now. 
- [x] This indicator mentioned above `is-data-ready` will allow the form to be created, and until the flag is set to `true` the the form creation will be delayed. 
- [x] During the form creation using `Formio.createForm` the initial data is set during this process avoiding a form redraw.

## Loader

- [x] Form io component now uses the `skeleton` loader from Vuetify.
- [x] There is a slot provided for the consumer to use a custom skeleton loader if that is required.

## Footer Buttons

- [x] Footer buttons are not displayed until the `is-data-ready` is true as it is considered to be part of the form, even though they are technically outside form.io.

## Updated forms

- [x]  All the student portal forms which use initial data for the form loading are updated.

## Skeleton loader

image

image

### DIFF
--- a/sources/packages/web/src/components/common/SupportingUserForm.vue
+++ b/sources/packages/web/src/components/common/SupportingUserForm.vue
@@ -86,6 +86,7 @@ export default defineComponent({
 
     watchEffect(async () => {
       if (props.supportingUserId) {
+        isDataReady.value = false;
         supportingUser.value =
           await SupportingUsersService.shared.getSupportingUserData(
             props.supportingUserId,
--- a/sources/packages/web/src/views/aest/student/StudentAccountApplicationsApproval.vue
+++ b/sources/packages/web/src/views/aest/student/StudentAccountApplicationsApproval.vue
@@ -104,6 +104,7 @@ export default defineComponent({
     };
 
     const getStudentDetails = async () => {
+      isDataReady.value = false;
       const accountApplication =
         await StudentAccountApplicationService.shared.getStudentAccountApplicationById(
           props.studentAccountApplicationId,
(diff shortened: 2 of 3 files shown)


sources/packages/web/src/components/common/SupportingUserForm.vue
```diff
@@ -117,6 +119,7 @@ export default defineComponent({
           hasValidSIN: supportingUser.value.personalInfo?.hasValidSIN,
           parentFullName: supportingUser.value.parentFullName,
         };
+        isDataReady.value = true;
```
andrewsignori-aot, line 122: As a general rule, it should be set to `false` before the API call to allow it to react to the data being reloaded.
I am not sure if we should have it for this form, but for forms that potentially reload data, we can have it.

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -1,15 +1,17 @@
 <template>
   <formio
     :formName="selectedForm"
-    :data="formData"
+    :data="{ ...initialData, ...additionalFormData }"
```
andrewsignori-aot, line 4: For my understanding, I did not follow this change. Since the `StudentApplicationFormData` already has the `howWillYouBeAttendingTheProgram`, what was the benefit in assigning it using two objects.

bidyashish (approved):
LGTM @dheepak-aot

sources/packages/web/src/components/common/SupportingUserForm.vue
dheepak-aot: Added the setting to false before API call. Even if the code to load the data is moved to a separate method it will move along.

Issue #4941, opened 2025-07-17
Approved exception does not repeat - additional transportation


Joshua-Lakusta:
User Story: 
As a ministry user I want an approved additional transportation exception to not require approval again if there has been no changes made to any information connected to the.

Acceptance Criteria
Business AC

- [x] Program year 25-26 onward for ft
- [x] program year 26-27 onward for pt
- [x] If an exception is already triggered and is pending a decision we will not change that
- [x] All current applications that have a new edit submitted should have this logic apply (assumption is that we can get the historical approval information easily  for those applications)
- [x] If none of the questions related to the exception have changed and the exception has been approved previously do not retrigger the exception
- [x] If there has been no changes to any of these elements on edit submission then the exception will not trigger again and it will not be needed for it to be approved again
- [x] If a previously approved version exists and the data from the new submission matches the container exactly then do not retrigger the exception
- [x] Ensure that any files in the uploader have the same name and content
- [x] Utilize the edit comparison functionality built in the past that flags elements red to ID if anything has changed for the elements within the new container
- [ ] <img width="1207" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNTNDQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--722a6cea32662de3fa35766367fbe0b85392c7e0/image.png" alt="image.png" />
<img width="1437" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNTdDQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f9e65a25a49ede865b50f25af7fcd98d1d827da8/image.png" alt="image.png" />
<img width="1176" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkhDQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4003d020bb5239e66f5eb1aa906478f6c761a308/image.png" alt="image.png" />

Technical AC baseline
- [ ] Create a new element that maps multiple questions to a new "container" could be named whatever makes sense
- [ ] Capture all questions in the additional transportation question within the container or equivalent solution that allows us to recognize questions associated with an exception

[...]

PR #5010, opened 2025-07-31
#4941 - Approved exception does not repeat - New JSON Lib

andrewsignori-aot:
- Replaced the JSONPath with the [jsonata](https://github.com/jsonata-js/jsonata) lib to allow the filter and transformations that will be required once the new exceptions structure is in place.
- Adapt the load-consolidate-data BPMN, and a few workflows that were using the JSONPath expression.

## Manual tests
- Created a new 2025-26 full-time application including two parents (one self-declare and the other doesn't) and also an exception.
- Created a change request to remove one of the parents.
- Created a 2025-26 part-time application.
- Created a 2024-25 application and an appeal to add dependents.


sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
```diff
@@ -42,7 +42,7 @@ import {
 } from "@sims/services";
 import { SupportingUserType } from "@sims/sims-db";
 import { DataSource, EntityManager } from "typeorm";
-import * as JSONPath from "jsonpath";
+import * as jsonata from "jsonata";
```
sh16011993, line 45: I came across an editor suggestion to convert this to a default import.

sh16011993: Thanks for the discussion.

sh16011993 (approved):
Great work @andrewsignori-aot 👍  Thanks for clarifying my questions.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
```diff
@@ -170,7 +170,7 @@ export class AssessmentController {
         assessment,
         hasNOAApproval,
       );
-      const outputVariables = filterObjectProperties(
+      const outputVariables = await filterObjectProperties(
```
dheepak-aot, line 173: There is an e2e test for assessment consolidated data. Not sure how it is passing. 
image

dheepak-aot (approved):
Great work finding the better way to parse json data and thanks for the sync up. One minor observation on E2E tests.

sources/packages/backend/apps/workers/src/controllers/assessment/assessment.controller.ts
andrewsignori-aot: The "$" at the beginning is compatible with the new framework. I can remove from this to avoid confusion in the future.

PR #5017, opened 2025-08-01
#4941 - Approved exception does not repeat - Save File Hash

andrewsignori-aot:
Saving the file hash for student-uploaded files to allow future comparison of exception files.

## Migration rollback

image

### DIFF
--- a/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
@@ -20,6 +20,7 @@ import { Queue } from "bull";
 import { VirusScanQueueInDTO } from "@sims/services/queue";
 import { FILE_SAVE_ERROR } from "../../constants";
 import { ObjectStorageService } from "@sims/integrations/object-storage";
+import { createHash } from "crypto";
 
 @Injectable()
 export class StudentFileService extends RecordDataModelService<StudentFile> {
@@ -72,6 +73,10 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
     newFile.student = { id: studentId } as Student;
     newFile.creator = { id: auditUserId } as User;
     newFile.virusScanStatus = VirusScanStatus.InProgress;
+    // Generate the file hash.
+    newFile.fileHash = createHash("sha256")
+      .update(createFile.fileContent)
+      .digest("hex");
 
     summary.info(`Saving the file ${createFile.fileName} to database.`);
     let savedFile: StudentFile;
--- a/sources/packages/backend/apps/db-migrations/src/sql/StudentFiles/Add-file-hash-column.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/StudentFiles/Add-file-hash-column.sql
@@ -1,6 +1,12 @@
+-- Creates the new column with an empty default value to allow it to be defined as NOT NULL.
 ALTER TABLE
   sims.student_files
 ADD
-  COLUMN file_hash CHAR(64);
+  COLUMN file_hash CHAR(64) NOT NULL DEFAULT '';
 
-COMMENT ON COLUMN sims.student_files.file_hash IS 'SHA256 hash of the file contents.';
+ALTER TABLE
+  sims.student_files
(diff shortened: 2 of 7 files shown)


sources/packages/backend/libs/sims-db/src/entities/student-file.model.ts
```diff
@@ -101,4 +101,13 @@ export class StudentFile extends RecordDataModel {
     nullable: false,
   })
   virusScanStatusUpdatedOn: Date;
+  /**
+   * Hash of the file content (SHA256).
+   */
+  @Column({
+    name: "file_hash",
+    type: "char",
```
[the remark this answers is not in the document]
andrewsignori-aot: DB migrations are not created by typeorm in this project.

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
```diff
@@ -72,6 +73,10 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
     newFile.student = { id: studentId } as Student;
     newFile.creator = { id: auditUserId } as User;
     newFile.virusScanStatus = VirusScanStatus.InProgress;
+    // Generate the file hash.
+    newFile.fileHash = createHash("sha256")
+      .update(createFile.fileContent)
+      .digest("hex");
```
dheepak-aot, line 79: Would it be better if we consider the file name also when creating the SIN hash?

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
```diff
@@ -72,6 +73,10 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
     newFile.student = { id: studentId } as Student;
     newFile.creator = { id: auditUserId } as User;
     newFile.virusScanStatus = VirusScanStatus.InProgress;
+    // Generate the file hash.
+    newFile.fileHash = createHash("sha256")
+      .update(createFile.fileContent)
```
dheepak-aot, line 78: This could a generic operation which can be put in a util method.

andrewsignori-aot: It a single method call that I did not evaluate as complex enough, so I did not put it in a helper for now.
When I execute the upcoming hashes, I may reconsider it, but I would like to move on with it for this PR, unless it is a blocker.

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
andrewsignori-aot: The unique file name is exactly what we need to prevent to cause any change in the dash.
The idea here is to hash the file content only to allow the file content only verification.

sources/packages/backend/libs/sims-db/src/entities/student-file.model.ts
```diff
@@ -101,4 +101,13 @@ export class StudentFile extends RecordDataModel {
     nullable: false,
   })
   virusScanStatusUpdatedOn: Date;
+  /**
+   * Hash of the file content (SHA256).
```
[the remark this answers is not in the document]
andrewsignori-aot: The comment was improved, and it is also the same now on both.

sources/packages/backend/apps/db-migrations/src/sql/StudentFiles/Add-file-hash-column.sql
```diff
@@ -1,6 +1,12 @@
+-- Creates the new column with an empty default value to allow it to be defined as NOT NULL.
 ALTER TABLE
   sims.student_files
 ADD
-  COLUMN file_hash CHAR(64);
+  COLUMN file_hash CHAR(64) NOT NULL DEFAULT '';
```
sh16011993, line 5: 👍

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
dheepak-aot: I got the intention now which follows the AC. Thanks for the sync up.

Issue #5003, opened 2025-07-30
Parent form: display income tax year in dependant table


AnnaPBashkatova:
Dependent questions in Parent form need to be updated. 

Acceptance Criteria:
FT app, PY 25/26 and onward

- [x] In a parent form, reword the question to "Did you claim this dependent on your 20xx Income Tax Return due to disability"
- [x] In a parent form replace 20xx with a previous tax year: it should be updated dynamically

<img width="687" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnZrQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a3fe8115e5fb42d23468e9368df13ef8cba19870/image.png" alt="image.png" />

CarlyCotton:
Bringing back to grooming - Dynamic year isn't currently used in this form and it would be the first time bringing it in.

PR #5022, opened 2025-08-06
#5003 - Add dynamic tax year to parent form 

ninosamson:
- adds dynamic tax year to supporting user parent form for 25/26 
image


sh16011993 (approved):
LGTM @ninosamson 👍

PR #5031, opened 2025-08-06
#4731 -  Formio-loader bug fix

dheepak-aot:
## Fomio Loader issues fix

- [x] Student application form initial data has been populated with spread operator which on every reactive change re-merges and creates the value again. Because of this action, form data is reset to inital data, whenever any reactive variable is modified at Vue component. Hence the spread operator is removed, and to update a component value in formio, the formio utils have been used.
image

- [x] Due to the `submittedData` set at the create form operation, `Formio.createForm`  there are few impacts observed and we anticipate potentially more impacts. Hence we restored to the previous logic of setting form data after form create operation. 
But by doing this way, there is a visual re-rendering which is now hidden by forcefully using a timeout.

### DIFF
--- a/sources/packages/web/src/components/generic/formio.vue
+++ b/sources/packages/web/src/components/generic/formio.vue
@@ -25,6 +25,7 @@ import {
 } from "@/types";
 import { v4 as uuid } from "uuid";
 import { useFormatters, useFormioUtils, useOffering } from "@/composables";
+import { FORMIO_LOAD_DATA_PROCESSING_VIEW_DELAY } from "@/constants/system-constants";
 
 export default defineComponent({
   emits: {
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/components/generic/formio.vue
```diff
@@ -201,7 +177,13 @@ export default defineComponent({
...
         context.emit("render", form, event);
       });
-      isFormCreated.value = true;
+
+      // When the form submission data is set, the form is rendered again.
+      // To visually hide the form until the rendering to be completed, a delay is set.
+      setTimeout(() => {
+        isFormCreated.value = true;
+        context.emit("loaded", form);
+      }, 200);
```
andrewsignori-aot, line 186: Can we have the `200` created as a const?
Suggestion: `FORMIO_LOAD_DATA_PROCESSING_VIEW_DELAY`.

andrewsignori-aot (approved):
Thanks for the fix, minor suggestion for a const.
Other then that, looks good 👍

sh16011993 (approved):
Nice work @dheepak-aot 👍  Thanks for the PR walkthrough.

Issue #4877, opened 2025-07-08
Reverse Scholastic Standing Reporting: Ministry 


AnnaPBashkatova:

CarlyCotton:
Need to check order of operations between archiving / ecert to ensure intended effect.

CarlyCotton:
**to bring up with devs: UC does not generate a new assessment, just a new line.

PR #5037, opened 2025-08-07
#4877- Reverse scholastic standing -  DB migrations and UC weeks limiting

dheepak-aot:


andrewsignori-aot (approved):
Nice work, looks good 👍

Issue #5039, opened 2025-08-08
RELEASE FIX (2.13/2.14)- Resolve Restriction Error


CarlyCotton:
**Current Behaviour**
In test and staging, Ministry user with access receives an unexpected error when resolving restrictions.

**Expected Behaviour**
When Ministry user with access opens an active restriction they should be provided a description box to add text to before resolving.

```
[Nest] 16 - 08/08/2025, 5:59:44 PM LOG [AccessLoggerMiddleware] Request - PATCH /api/aest/restriction/student/1973/studentRestriction/1591/resolve From 142.32.26.60 | User GUID: e5eb4a6fe77f4ead9032b8ff2c6b1f25@idir | User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36 Edg/138.0.0.0
[Nest] 16 - 08/08/2025, 5:59:44 PM ERROR [AppAllExceptionsFilter] Unhandled exception, request path: /api/aest/restriction/student/1973/studentRestriction/1591/resolve
{
"stack": "BadRequestException: Bad Request Exception\n at ValidationPipe.exceptionFactory (/opt/app-root/src/node_modules/@nestjs/common/pipes/validation.pipe.js:107:20)\n at ValidationPipe.transform (/opt/app-root/src/node_modules/@nestjs/common/pipes/validation.pipe.js:74:30)\n at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n at async resolveParamValue (/opt/app-root/src/node_modules/@nestjs/core/router/router-execution-context.js:148:23)\n at async Promise.all (index 0)\n at async pipesFn (/opt/app-root/src/node_modules/@nestjs/core/router/router-execution-context.js:151:13)\n at async /opt/app-root/src/node_modules/@nestjs/core/router/router-execution-context.js:37:30\n at async /opt/app-root/src/node_modules/@nestjs/core/router/router-execution-context.js:46:28\n at async /opt/app-root/src/node_modules/@nestjs/core/router/router-proxy.js:9:17",
"response": {
"message": [
"noteDescription must be shorter than or equal to 1000 characters",
"noteDescription should not be empty"
]
},
"status": 400,
"options": {},
"message": "Bad Request Exception",
"name": "BadRequestException"
}
```

PR #5044, opened 2025-08-11
#5039 - Release fix: Resolve restriction error

sh16011993:


dheepak-aot (approved):
Thanks for the fix. Looks good 👍

andrewsignori-aot (approved):
Thanks for the fix, great work with the validations, and tested scenarios.

Issue #4885, opened 2025-07-09
Room and board appeal part 2 (ministry)


Joshua-Lakusta:
As a ministry user i need to be able to review the room and board appeals and wanting a landing list in order to view all outstanding appeals and action them.

#### Land spot for appeal requests
- [ ] Create an additional ministry list for appeals to land at. The formatting and overall behaviour matches the two change request lists.
- [ ] Ensure that the 25-26 appeals only show up on the appeals list and no others lists

- [ ] List name would be Appeals

<img width="342" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNzJvQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6af8026aba81a6c0563f2ce34348854e2d7e78fb/image.png" alt="image.png" />

- [ ] Content for the list (see image below for reference)

> Student requests
> Appeals
> Pending appeals  "Search bar"
> Appeals that require ministry review

- [ ] Content

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2VvQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--11a002f52fa4127a3597d2dfbb18445fa741a841/image.png" alt="image.png" />

- [ ] Update to Vuetify

**Technical ACs**
- Reuse component from Pre-2025/26 Change requests & adapt API call
- Add a new API filter for instance, `requestType`: "ChangeRequest" | "Appeals".
  - If appeals, look for records associated only with PY 2025-26 and over.
- For testing the appeals, create a 2025-26 full-time completed application. The "appeals" button should be available in the application summary table.

PR #5050, opened 2025-08-12
#4885 - Room and board appeal part 2 (ministry) E2E

bidyashish:
### As a part of this PR, the following e2e test was added:

**StudentAppealAESTController(e2e)-getAppeals**
**Technical Details:**

- Creates a program year with start date >= "2025-08-01" to match `AppealType.Appeal` filter criteria

- ✅ *Should return result count as 1 when there is one pending appeal for student application with appeal type 'appeal'*

**Demo:**
image


andrewsignori-aot (commented):
Nice work, one comment only.

PR #5056, opened 2025-08-14
#4941 - Approved exception does not repeat - Ministry Approval UI

andrewsignori-aot:
Updated the Ministry UI (also shared with institutions) to display previously approved exceptions.

## New Section: Previously Approved Requests

- Added the new section to list exceptions that were previously approved.
- Section to be displayed to the Ministry and Institutions.
- The image below shows an automatically approved exception. 

image

## Exceptions are displayed in both sections when pending

image

## Exceptions are displayed in both sections when approved

image


dheepak-aot (approved):
The ministry view looks great. Added a minor comment for isDataReady.

Issue #4871, opened 2025-07-03
Incorporate UC weeks count from legacy system into FT UC count


AnnaPBashkatova:
When student has UC weeks reported previously in Legacy system, make sure these weeks count towards sum of all FT UC weeks in SIMS

PR #5057, opened 2025-08-14
#4871 - Incorporate FT UC weeks from legacy while triggering SSR restriction

sh16011993:
###  As a part of this PR, the following were completed:

- Centralized the `getScholasticStandingSummary` method of the `student-scholastic-standings.service.ts`
- Moved `getSFASTotalUnsuccessfulCompletionWeeks` method inside the `getScholasticStandingSummary` method.
- Refactored `getScholasticStandingSummary` method and remove the `getTotalFullTimeUnsuccessfulWeeks`, replacing it with the `getScholasticStandingSummary`.


sources/packages/backend/apps/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
```diff
@@ -626,21 +610,55 @@ export class StudentScholasticStandingsService extends RecordDataModelService<St
...
+          await this.sfasIndividualService.getSFASTotalUnsuccessfulCompletionWeeks(
+            studentId,
+          );
+        return {
+          fullTimeUnsuccessfulCompletionWeeks:
+            sfasUnsuccessfulCompletionWeeks +
+            (result?.totalUnsuccessfulWeeks ?? 0),
+        };
+      }
+    } else {
```
dheepak-aot, line 638: Looks like there is a misunderstanding from dev syncup regarding the filter.

I will not consider using the filter as a blocker. But if the filter is used, it must be used only to restrict the FT legacy query which is always required as of now. 

Hence, it would recommend not to change any logic or introduce filter but just centralize the logic.

sources/packages/backend/apps/api/src/services/student-scholastic-standings/student-scholastic-standings.service.ts
```diff
@@ -629,6 +609,10 @@ export class StudentScholasticStandingsService extends RecordDataModelService<St
       })
       .groupBy("application.offeringIntensity")
       .getRawMany<ScholasticStandingSummary>();
+    const sfasUnsuccessfulCompletionWeeks =
+      await this.sfasIndividualService.getSFASTotalUnsuccessfulCompletionWeeks(
+        studentId,
+      );
```
dheepak-aot, line 615: Can we execute in parallel?

sources/packages/backend/apps/api/src/services/index.ts
```diff
@@ -32,6 +32,7 @@ export * from "./student-assessment/student-assessment.constants";
 export * from "./student-appeal/student-appeal.service";
 export * from "./student-appeal-request/student-appeal-request.service";
 export * from "./student-scholastic-standings/student-scholastic-standings.service";
+export * from "./student-scholastic-standings/student-scholastic-standings.models";
```
dheepak-aot, line 35: Can we move the import before the service at line 34. Alternatively removing this line will also work as I do not see an import of any model in the PR.

dheepak-aot (approved):
Nice work @sh16011993 . Looks good 👍

bidyashish (approved):
LHTM @sh16011993

PR #5075, opened 2025-08-19
#4886 - Convert to ASCII

sh16011993:
As a part of this PR, the following was added:

- Adapted the method `convertToASCII` to allow only printable characters from 0 to 127 (excluding newline and carriage return characters).
- Added a unit test to test it.

### DIFF
--- a/sources/packages/backend/libs/utilities/src/_tests_/unit/string-utils.translateToASCII.spec.ts
+++ b/sources/packages/backend/libs/utilities/src/_tests_/unit/string-utils.translateToASCII.spec.ts
@@ -1,10 +1,9 @@
-import {
-  CARRIAGE_RETURN,
-  convertToASCIIString,
-  LINE_FEED,
-  NON_PRINTABLE_CHARACTERS_LIMIT,
-  UNEXPECTED_CHAR,
-} from "@sims/utilities/string-utils";
+import { convertToASCIIString } from "@sims/utilities/string-utils";
+
+const NON_PRINTABLE_CHARACTERS_LIMIT = 31;
+const CARRIAGE_RETURN = 13;
+const LINE_FEED = 10;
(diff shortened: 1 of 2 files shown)


sources/packages/backend/libs/utilities/src/string-utils.ts
```diff
@@ -49,6 +49,9 @@ export function convertToASCII(rawContent?: string): Buffer | null {
   });
   const content = Buffer.from(rawContent, FILE_DEFAULT_ENCODING);
   for (const [index, char] of content.entries()) {
+    if (char < 32 && char !== 10 && char !== 13) {
+      content[index] = 63; // Replace with ? for control characters.
+    }
```
andrewsignori-aot, line 54: Can we create the following constants to improve the code readability?
- NONPRINTABLE_CHARACTERS_LIMIT = 31
- CARRIAGE_RETURN = 13
- LINE_FEED = 10
- UNEXPECTED_CHAR = 63
_Note:_ please also user the `UNEXPECTED_CHAR ` at line 141.

sources/packages/backend/libs/utilities/src/_tests_/unit/string-utils.translateToASCII.spec.ts
```diff
@@ -36,4 +36,19 @@ describe("StringUtils-convertToASCII", () => {
     // Assert
     expect(translatedData).toBeNull();
   });
+
+  it("Should replace ASCII control characters (0-31) with '?'", () => {
```
andrewsignori-aot, line 40: The test description should try to keep a "Should"/"When" pattern and have a period at the end.

andrewsignori-aot (approved):
Nice work, thanks for doing the changes, looks good 👍

dheepak-aot (approved):
Good Job!. Looks good.

Issue #4591, opened 2025-04-08
Log Person ID - External API 


ninosamson:
**User story:**
As a security officer or 3rd party, I may need to get audit records of external access. 

**Acceptance Criteria:**
- [ ] log sfas or sims ID for API calls (student ids)
- [ ] log whatever IDs are returned (student ids)

**Context**
On external sdpr controller, should return SIMS Student ID (or sfas_individual id if legacy)

PR #5084, opened 2025-08-20
#4591 - Log Person ID - External API

bidyashish:
Work Done:

- Implemented logging for all API calls to the external `sdpr controller`.
- Captured the student ID (sims_student_id or sfas_individual_id) used in each request.
- Logged the student ID returned in the API response to confirm data retrieval.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
@@ -21,6 +21,7 @@ import {
 } from "./models/student-external-search.dto";
 import { StudentExternalControllerService } from "./student.external.controller.service";
 import { SFASApplication } from "@sims/sims-db";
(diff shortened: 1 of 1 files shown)


andrewsignori-aot (approved):
Nice work, looks good 👍

sources/packages/backend/apps/api/src/route-controllers/student/student.external.controller.ts
```diff
@@ -90,10 +91,20 @@ export class StudentExternalController extends BaseController {
         studentApplications,
         legacyApplications,
       );
+    // Log found student and SFAS individual IDs.
+    if (student) {
+      this.logger.log(`Found SIMS student with ID: ${student.id}`);
+    }
+    if (sfasIndividual) {
```
dheepak-aot, line 98: The same condition, if(sfasIndividual) is at line 82 and we can make use of the same instead, adding the log for legacy after line 86.

PR #5085, opened 2025-08-20
#4886 - Converting to ASCII [Encode URL]

sh16011993:
**As a part of this PR, the URL in the `getSupplierInfoFromCAS` is encoded to ensure a valid URL is formed for the CAS api call.**


bidyashish (approved):
LGTM

Issue #5065, opened 2025-08-18
Exceptions not repeat 25-26 backwards compatibility


Joshua-Lakusta:
### User Story
As a ministry user I want the current program year to also benefit from exceptions not repeating

### Acceptance Criteria

- [ ] Create a path that allows the exception not repeating feature to be applied to 25-26 (PT) program year
- [ ] Once an edit is submitted the new functionality should be applied moving forward
- [ ] All exception decisions pre the release of this feature will not be used to confirm historical approved status
- [ ] All exception decisions on any submission post the release of this feature will be used to confirm the historical approved status

### Additional context
This work has evolved from the initial ticket 

In general this ticket makes it so that post release of this feature new decisions will be used but any previous exception decisions will always trigger the exception to be re reviewed.

andrewsignori-aot:
## Dependents list issue

Existing part-time applications with dependents exceptions will continue to work, but they will detect previously approved exceptions only if edited a third time (to be clear, the exceptions will be detected every time).

### Original submission (before new form.io definition updates)

```json
{
    "bcResident": "yes",
    "craConsent": true,
    "dependants": [
        {
            "fullName": "My son",
            "dateOfBirth": "2025-08-01",
            "declaredOnTaxes": "yes",
            "pdDependentUpload": [
                {
                    "url": "student/files/File B-e48f5f55-1d97-49e0-9541-4723cd98a45f.txt",
                    "hash": "",
                    "name": "File B-e48f5f55-1d97-49e0-9541-4723cd98a45f.txt",
                    "size": 15,
                    "type": "text/plain",
                    "storage": "url",
                    "originalName": "File B.txt"
                }
            ],
            "dependantCustodyFileUpload": [
                {
                    "url": "student/files/File C-6366aefc-b32f-4a8c-bc43-48fcba00ecde.txt",
                    "hash": "",
                    "name": "File C-6366aefc-b32f-4a8c-bc43-48fcba00ecde.txt",
                    "size": 15,
                    "type": "text/plain",
                    "storage": "url",
                    "originalName": "File C.txt"
                }
            ],
            "supportnocustodyDependants": "yes",
            "attendingPostSecondarySchool": "yes",
            "dependantsIncomeTaxApplicationException": "studentApplicationException",
            "dependantsSharedCustodyApplicationException": "studentApplicationException"
        },
        {
            "fullName": "My second Son",
            "dateOfBirth": "2025-08-04",
            "declaredOnTaxes": "yes",
[...]

PR #5091, opened 2025-08-21
#5065 - Exceptions not repeat 25-26 backwards compatibility - Fix Dependents and Custom Components

andrewsignori-aot:
- Changed the approach to generate exception information for dependents, swapping the container for a regular hidden field that produces the same JSON output to be saved to the DB. **Kudos to @dheepak-aot for providing the solution.** See below the new hidden field content, named with the same name as the previous exception container.

```js
value = {
 exceptionDescription: `Dependant's income tax - ${data.dependants[rowIndex].fullName}`,
 pdDependentUpload: data.dependants[rowIndex]["pdDependentUpload"]
}
```

- Update existing custom components and create an additional one with a complete schema.

## Updated Schema

image

## New Full Schema

image


andrewsignori-aot:
> Thanks for taking the additional step for dependants exception and creating and updating the form.io component. Looks good 👍

Thanks for looking into it and finding the alternative solution 😉

Issue #5059, opened 2025-08-15
When a partial match exists add hold restriction


Joshua-Lakusta:
As a ministry user I want a hold restriction to be automatically added to a profile when it is not bridge matched and and there are partial matches available. This is important as the manual process is delayed and can be missing students and allowing them through with the wrong information. 

**Acceptance Criteria**
- [ ] Add the "hold" restriction to the account if the following state exists at account creation
    - partial matches exist in the list
    - the account is not matched
- [ ] The hold restriction exists already
- [ ] Add the following note inside the hold restriction 

> Restriction added to prevent application completion while potential partial match exists

PR #5114, opened 2025-08-28
#5059 - When a partial match exists add hold restriction

bidyashish:
### AC
  Implements automatic addition of HOLD restrictions when partial matches are detected during student account creation. This ensures that students with potential data conflicts are prevented from completing applications until the matches are resolved manually.

### Demo 
- 2 partial data (Last name, DOB)
1. Restriction Added.
image
2.Notes Added
image

### DIFF
--- a/sources/packages/backend/apps/api/src/services/student/student.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student/student.service.ts
@@ -188,7 +188,6 @@ export class StudentService extends RecordDataModelService<Student> {
     student.user = user;
     student.sinConsent = studentInfo.sinConsent;
     let sfasIndividual: SFASIndividual;
-    let shouldAddHoldRestriction = false;
     try {
       // Get PD status from SFAS integration data.
       sfasIndividual = await this.sfasIndividualService.getIndividualStudent(
@@ -205,15 +204,6 @@ export class StudentService extends RecordDataModelService<Student> {
         student.disabilityStatusEffectiveDate = this.getDisabilityEffectiveDate(
           sfasIndividual.ppdStatusDate,
           student.disabilityStatus,
-        );
-      } else {
-        // If sfasIndividual wasn't found, check for a partial match
-        shouldAddHoldRestriction = await this.checkLegacyIndividualPartialMatch(
-          user,
-          student,
-          studentSIN,
-          auditUserId,
-          externalEntityManager,
         );
       }
     } catch (error) {
@@ -263,6 +253,15 @@ export class StudentService extends RecordDataModelService<Student> {
           this.systemUsersService.systemUser.id,
           entityManager,
         );
+      } else {
+        // If sfasIndividual wasn't found, check for a partial match
+        await this.checkLegacyIndividualPartialMatch(
+          user,
+          student,
+          studentSIN,
+          auditUserId,
+          entityManager,
+        );
       }
 
       // Create the new entry in the student/user history/audit.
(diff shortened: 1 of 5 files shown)


sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -272,6 +275,32 @@ export class StudentService extends RecordDataModelService<Student> {
...
+        holdRestriction.creator = {
+          id: this.systemUsersService.systemUser.id,
+        } as User;
+        holdRestriction.restrictionNote = holdNote;
+        holdRestriction.isActive = true;
+
+        await entityManager
+          .getRepository(StudentRestriction)
+          .save(holdRestriction);
+      }
```
dheepak-aot, line 302: It would be simple to move the restriction creation inside the `checkLegacyIndividualPartialMatch` where the notification creation takes place.
image

One the we might require is, we need persisted student(student record saved and id generated) before in hand.

For that sake, we can move the else block at line 266. 

image

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -272,6 +275,32 @@ export class StudentService extends RecordDataModelService<Student> {
...
+          "Restriction added to prevent application completion while potential partial match exists",
+          this.systemUsersService.systemUser.id,
+          entityManager,
+        );
+
+        const holdRestriction = new StudentRestriction();
+        holdRestriction.student = { id: savedStudent.id } as Student;
+        holdRestriction.restriction = await entityManager
+          .getRepository(Restriction)
+          .findOne({ where: { restrictionCode: RestrictionCode.HOLD } });
```
dheepak-aot, line 292: Please do not use the findOne without select as general.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -272,6 +275,32 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      if (shouldAddHoldRestriction && !sfasIndividual && !existingStudent) {
+        const holdNote = await this.noteSharedService.createStudentNote(
+          savedStudent.id,
+          NoteType.Restriction,
+          "Restriction added to prevent application completion while potential partial match exists",
+          this.systemUsersService.systemUser.id,
+          entityManager,
+        );
+
+        const holdRestriction = new StudentRestriction();
```
dheepak-aot, line 288: Optional. Can we move the block of creating the hold note into a private method.? It will make it easy to read and digest.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -272,6 +275,32 @@ export class StudentService extends RecordDataModelService<Student> {
...
+        holdRestriction.student = { id: savedStudent.id } as Student;
+        holdRestriction.restriction = await entityManager
```
dheepak-aot, line 290: Can we check for an existing active hold restriction before adding it? similar to this. 

image

There is a very edge case possibility that, when student switch from basic-bceid to bcsc or vice-versa then the student profile creation can happen more than once for a same student which will introduce the possibility of them already having a HOLD restriction.

sources/packages/backend/apps/api/src/services/student/student.service.ts
bidyashish: Code bloack moved.

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -394,6 +395,47 @@ export class StudentService extends RecordDataModelService<Student> {
...
+      const hasHoldRestriction = await externalEntityManager
+        .getRepository(StudentRestriction)
+        .findOne({
+          select: { id: true },
+          where: {
+            student: { id: student.id },
+            restriction: { restrictionCode: RestrictionCode.HOLD },
+            isActive: true,
+          },
+        });
```
dheepak-aot, line 409: Why did not we re-use `studentRestrictionService.hasAnyActiveRestriction` here?

sources/packages/backend/apps/api/src/services/student/student.service.ts
```diff
@@ -394,6 +395,47 @@ export class StudentService extends RecordDataModelService<Student> {
...
+        );
+
+        const holdRestriction = new StudentRestriction();
+        holdRestriction.student = { id: student.id } as Student;
+        holdRestriction.restriction = await externalEntityManager
+          .getRepository(Restriction)
+          .findOne({
+            select: { id: true },
+            where: { restrictionCode: RestrictionCode.HOLD },
+          });
```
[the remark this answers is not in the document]
bidyashish: Updated.

dheepak-aot (approved):

PR #5140, opened 2025-09-05
#5142 - Update Docker Local Network

andrewsignori-aot:
- Ensure the local network is created to be nicely shared across Docker Compose files.
- Created a make command to ensure the shared network is created before `docker-compose up` commands are executed.
- Changed network for all Docker Compose files to use `external: true` to attach to the already existing network.
- Set all the Docker Compose services to `restart: unless-stopped` to allow the container to retain its status after the local machine is restarted.


dheepak-aot (approved):
Looks good 👍

Issue #4494, opened 2025-03-18
Work flow event order change to move PIR first


Joshua-Lakusta:
As an institution and ministry user I want the PIR event in the work flow to occur first in the order and before exceptions. This is important so that 

**Acceptance Criteria**
- [ ] Move the event for PIR ahead of the exceptions
- [ ] Remove app trigger of FAED exception when **self entered** dates are within 6 weeks or after end date. Still show banner but it will have no outcome except for warning. We may need to update the content on the banner
- [ ] When a student selects and offering and its within 6 weeks of end date or after end date should still trigger FAED exception path as there will be no PIR
- [ ] When a PIR is attached and todays date is within 6 weeks of the end date or after the end date it must trigger a funding after end date
- [ ] Currently when this is entered it does flag that it needs ministry approval but is not  triggering an exception because it currently happens before the exceptions step. This would not longer be the case.
- [ ] Edit application trigger for FAED will still need to occur the same as currently as PIR may not be retriggered in the future

**Sample scenarios**

Student submits app with self declared dates that have over 6 weeks remaining + PIR has dates that have more then 6 weeks remaining  = no funding after end date exception path

Student submits app with self declared dates that are within 6 weeks and triggers FAED banner + PIR has dates that have more then 6 weeks remaining = no funding after end date exception path

Student submits app after end date of self declared dates + PIR has dates that have more then 6 weeks remaining = no funding after end date exception path

Student submits app with self declared dates that have over 6 weeks remaining + PIR has dates that have less then 6 weeks remaining  or are after end date = funding after end date exception path triggered

Student submits app with self declared dates that are within 6 weeks and triggers FAED banner + PIR has dates that have less then 6 weeks remaining  or are after end date = funding after end date exception path triggered

Student submits app after end date of self declared dates + PIR has dates that have less then 6 weeks remaining  or are after end date = funding after end date exception path triggered

Student submits app with selected offering that is within 6 weeks of end date or after end date = funding after end date exception path triggered

<img width="1233" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzlPQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--546892b7830f370a8afca0ef25d06df7aa6ec429/image.png" alt="image.png" />

**Additional Context**

Joshua-Lakusta:
[...]

PR #5155, opened 2025-09-09
#4494 - Minor PIR form variable  cleanup

dheepak-aot:
- Removed the obsolete code which is left over from the logic update for study end date based banner.

image

- Adjusted the comments as per updated code

image


sh16011993 (approved):
LGTM @dheepak-aot

Issue #2988, opened 2024-03-12
Aviation Limits - Program/Offering Limits


CarlyCotton:

Joshua-Lakusta:
This is critical as it is a mandatory type of limit and must be present for full time

Joshua-Lakusta:
Did some updates based on conversation with SME

CarlyCotton:
Program and offering updates already in progress, need to implement specific restrictions for 3 aviation types.

dheepak-aot:
SQL to check if any student restriction exist with  Federal `AV`.
```sql
select
	*
from
	sims.student_restrictions sr
where
	sr.restriction_id in (
	select
		id
	from
		sims.restrictions
	where
		restriction_type = 'Federal'
		and restriction_code = 'AV');
```

dheepak-aot:
@CarlyCotton @ninosamson During the update of Student Account Activity page, I did not include the legacy restriction `SFAS_AV` to this page to add the title and description due to the reason that currently, Student Account Activity does not have any entry for any of the specific legacy restrictions(The ones added on-demand) and only LGCY. Considering SFAS_AV similar to the legacy restrictions which are added on demand, I did not add it here.

But it was highlighted during code review that AC states that account activity must be added for each new restriction.

Can you please confirm if `SFAS_AV` must be included in the account activity?

ninosamson:
Let's not include the legacy restriction as that's not the pattern today. Updating AC @dheepak-aot .

andrewsignori-aot:
> Let's not include the legacy restriction as that's not the pattern today. Updating AC [@dheepak-aot](https://github.com/dheepak-aot) .

Not including it is ok @ninosamson, stating it is a system pattern does not seem right since we have LGCY over there.
From a system perspective, if the student has it listed, it should get some guidance or information about it.

dheepak-aot:
# Manual QA 

## Student blocked to accept assessment when already funded for aviation credential commercial pilot training

The same blocking condition will happen during disbursement as well
<img width="1869" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejJGQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--544040a2f414ff5cf95644275ab5799837cf166f/image.png" alt="image.png" />

PR #5189, opened 2025-09-15
#2988 - Aviation restriction - DB Migrations

dheepak-aot:
## DB Migrations

- [x] As we require aviation restrictions to impact applications with only certain credential types, the restrictions feature is now enhanced with a column `action_effective_conditions` which can have one or more conditions which decides the effectiveness of all restriction actions. If no conditions are provided all the actions will be implicitly effective.  If the provided conditions does not match the application related data, then the restriction actions will not take any effect.
- [x]  DB Migrations to add aviation restrictions `AVCP` , `AVIR`, `AVEN` and `SFAS_AV`.
- [x] `SFAS_AV` is added to support legacy aviation restrictions and added explicitly through migration instead of being added on-demand as it needs the effective conditions here which is different from other legacy only restrictions.
- [x] Added the legacy restriction mappings to map `AV01`, `AV02`, `AV10` to map `SFAS_AV` during legacy restriction import.

## Rollback evidence
**AddLegacyAviationRestrictionMaps**

image

**DeleteFederalRestrictionAV**

image

**AddAviationRestrictions**

image

**RestrictionsAddColActionEffectiveConditions**

image

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/sql/Restrictions/Add-aviation-restrictions.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/Restrictions/Add-aviation-restrictions.sql
@@ -17,7 +17,7 @@ VALUES
         'Other',
         ARRAY ['Stop part time disbursement',
     'Stop full time disbursement'] :: sims.restriction_action_types [],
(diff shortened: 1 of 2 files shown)


andrewsignori-aot (approved):
Great work, and nice framework addition with the `action_effective_conditions`, looks good 👍

Issue #4873, opened 2025-07-04
Ministry: ability to reverse restrictions


AnnaPBashkatova:

Joshua-Lakusta:
@CarlyCotton @ninosamson would prefer this be a part of 2.16 if possible. This feature is mainly linked to Scholastic reversal and the scholastic reversal is somewhat useless until this is done.

andrewsignori-aot:
@AnnaPBashkatova @CarlyCotton, the system currently resolves federal restrictions while importing them, hence we should at least also update the new specific column (`resolved_at`) to be created to track the resolution date, since we will no longer be able to use the modifier.

Carly, below would be the change.

<img width="863" height="596" alt="Image" src="https://github.com/user-attachments/assets/87bf449a-152e-42c1-8890-7ebb4072486a" />

andrewsignori-aot:
@AnnaPBashkatova can we have an AC to mention to QA the minor impact of the federal resolved restrictions?

> Resolved Federal restrictions should display the resolved date as they are right now.

AnnaPBashkatova:
@andrewsignori-aot Both ACs updated. Thank you, Andrew!

andrewsignori-aot:
@AnnaPBashkatova, I believe that we should have some AC that states that Federal restrictions should not be deleted.

andrewsignori-aot:
## Manual Testing 

- Used below academic restriction.

<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNXQ0Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--33faa4574d8c65ba0b74248713a4df58e9e1c605/image.png" alt="image.png" />

<img width="1382" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNlI0Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0de0a5544c1f0d871c0c959368e411dbe146805d/image.png" alt="image.png" />

[...]

PR #5194, opened 2025-09-16
#4873 - Ministry: ability to delete restrictions - DB Migrations

andrewsignori-aot:
- Added `deleted_at` to `sims.student_restrictions` to allow soft delete.
- Created new columns to capture resolution and deletion audit. _Note:_ creation will still use the `creator` and `created_at` columns.
- The entity model was changed only for students, but it has a shared class with institutions that can be enhanced once the institution restrictions are worked on.

## Migration Rollback

image

### DIFF
--- a/sources/packages/backend/libs/sims-db/src/entities/student-restriction.model.ts
+++ b/sources/packages/backend/libs/sims-db/src/entities/student-restriction.model.ts
@@ -1,13 +1,22 @@
 import {
+  Column,
+  DeleteDateColumn,
   Entity,
   JoinColumn,
   ManyToOne,
   OneToMany,
+  OneToOne,
   PrimaryGeneratedColumn,
 } from "typeorm";
 import { ColumnNames, TableNames } from "../constant";
 import { BaseRestrictionModel } from "./base-restriction.model";
-import { Student, Application, ApplicationRestrictionBypass } from ".";
+import {
+  Student,
+  Application,
+  ApplicationRestrictionBypass,
+  Note,
+  User,
+} from ".";
 
 /**
  * Entity for student restrictions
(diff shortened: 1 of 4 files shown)


sources/packages/backend/apps/api/src/auth/roles.enum.ts
```diff
@@ -45,4 +45,5 @@ export enum Role {
   StudentReissueMSFAA = "student-reissue-msfaa",
   StudentUpdateDisabilityStatus = "student-update-disability-status",
   StudentReverseScholasticStanding = "student-reverse-scholastic-standing",
+  StudentDeleteRestrictions = "student-delete-restrictions",
```
dheepak-aot, line 48: AC can be updated with this role name to avoid confusion for QA or anyone looking later. 

image

sources/packages/backend/libs/sims-db/src/entities/student-restriction.model.ts
```diff
@@ -48,4 +57,54 @@ export class StudentRestriction extends BaseRestrictionModel {
...
+  })
+  resolvedAt?: Date;
+
+  /**
+   * User who resolved the restriction.
+   */
+  @ManyToOne(() => User, { nullable: true })
+  @JoinColumn({
+    name: "resolved_by",
+    referencedColumnName: "id",
```
dheepak-aot, line 77: Can we use the id from the constant here following other places?

image

dheepak-aot (approved):
Good start for restrictions delete. 👍

One minor comment to use the constant for id.

sources/packages/backend/apps/api/src/auth/roles.enum.ts
andrewsignori-aot: I will request it to be updated.

ninosamson (approved):
Looks good!

PR #5199, opened 2025-09-17
#2988 -  Aviation restriction - ECert validation and add restriction step

dheepak-aot:
## E-Cert step

- [x] Added a new e-cert processing step `AddAviationCredentialRestrictionStep` for both full-time and part-time. This step will add the aviation restriction for the given aviation credential type if the application belongs to an aviation offering to ensure that a student is not funded more than once for a particular aviation credential type. 
**Note**: This step is executed after the e-cert calculation is done to ensure the student is funded first time.

### Manual test

- Submitted the first aviation application
image

- ECert was sent successfully but a restriction was added during e-cert generation

image

image

image

## Conditional restriction action at ECert validation(Shared for accepting assessment)

- [x] Based on the business requirement that the aviation restrictions are expected to block only that particular aviation credential funding, the existing restriction actions feature was enhanced to apply the restriction actions based on a condition.

The condition name and condition values are added to the new column `action_effective_conditions` in `sims.restrictions`.

image

These conditions will be evaluated in the code by `Evaluator` mapped to the condition name. 

image

image

### Manual test

- With the AVCP restriction added by an existing application for commercial pilot training, student submit application to same aviation credential again.

image

- With the AVCP restriction added by an existing application for commercial pilot training, student submit application to different aviation credential.
- 
image

- With the AVCP restriction added by an existing application for commercial pilot training, student submit application to NON aviation credential.
image

## E2E Tests

- [x] Added E2E tests as separate describe for aviation credentials for both full-time and part-time.

image

image

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-full-time-process-integration.scheduler.e2e-spec.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-full-time-process-integration.scheduler.e2e-spec.ts
@@ -51,9 +51,11 @@ import * as Client from "ssh2-sftp-client";
 import * as dayjs from "dayjs";
 import { FullTimeCertRecordParser } from "./parsers/full-time-e-cert-record-parser";
 import {
+  AVIATION_CREDENTIAL_TEST_INPUTS,
   awardAssert,
   createBlockedDisbursementTestData,
   loadAwardValues,
+  loadDisbursementAndStudentRestrictions,
 } from "./e-cert-utils";
 import { RestrictionCode, SystemUsersService } from "@sims/services";
 import { createFakeSFASApplication } from "@sims/test-utils/factories/sfas-application";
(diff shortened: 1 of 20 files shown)


sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/add-aviation-credential-restriction-step.ts
```diff
@@ -0,0 +1,92 @@
...
+import { EligibleECertDisbursement } from "../disbursement-schedule.models";
+import { ECertGenerationService } from "../e-cert-generation.service";
+import { getRestrictionByCode } from "./e-cert-steps-utils";
+import { RestrictionService } from "@sims/integrations/services";
+
+/**
+ * Check if the offering belongs to aviation credential types and add an aviation restriction
+ * for the given credential type if one does not already exist.
+ * By adding this restriction, it will prevent the student from being funded for the same
+ * aviation credential type again in the future.
```
andrewsignori-aot, line 20: Can we add an extra comment as
> Aviation offerings are not expected to create second disbursements due to their number of funded weeks. In case this scanrios changes, the second disbursement would be blocked if no changes are made.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/e-cert-steps-utils.ts
```diff
@@ -23,9 +24,14 @@ export function getRestrictionByActionType(
   eCertDisbursement: EligibleECertDisbursement,
   actionType: RestrictionActionType,
 ): StudentActiveRestriction {
-  return eCertDisbursement
-    .getEffectiveRestrictions()
-    .find((restriction) => restriction.actions.includes(actionType));
+  return eCertDisbursement.getEffectiveRestrictions().find(
+    (restriction) =>
+      restriction.actions.includes(actionType) &&
+      isRestrictionActionEffective(restriction.actionEffectiveConditions, {
```
andrewsignori-aot, line 30: `isRestrictionActionEffective` should be generic and receive as parameters the `restriction` and the `eCertDisbursement`. It would allow it to be expanded and handle other `actionEffectiveConditions` to also be resolved interbally by the method `isRestrictionActionEffective`.
Please let me know if I am missing something. If the assumption is right, I would consider it part of this PR.

sources/packages/backend/libs/test-utils/src/models/common.model.ts
```diff
@@ -134,4 +134,21 @@ export enum RestrictionCode {
    * Bankruptcy has been filed and there is a hold on federal and BC funding.
    */
   RB = "RB",
+  /**
+   * Aviation restriction which prevents a student from being funded
+   * for the credential commercial pilot program again in the future.
+   */
+  AVCP = "AVCP",
+
```
andrewsignori-aot, line 142: Please follow the pattern in the file for the blank lines.

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-full-time-process-integration.scheduler.e2e-spec.ts
```diff
@@ -1484,6 +1486,495 @@ describe(
...
+              },
+            },
+          });
+        },
+      );
+
+      it(
+        "Should generate e-cert and not block the disbursement for aviation credential type 'commercialPilotTraining'" +
+          ` when the student already has an active restriction ${RestrictionCode.AVCP} added due to previous funding of the same aviation credential type` +
+          " but the restriction was bypassed.",
```
andrewsignori-aot, line 1855: Is there a specific scenario to be tested for aviation and bypasses that was not already tested by the existing bypass scenarios?

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-part-time-process-integration.scheduler.e2e-spec.ts
```diff
@@ -1512,6 +1516,517 @@ describe(
         }),
       ).toBe(true);
     });
+    describe("Aviation credential part-time applications e-Cert generation", () => {
```
andrewsignori-aot, line 1519: Can we have a blank line added before the describe?

andrewsignori-aot (commented):

sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-full-time-process-integration.scheduler.e2e-spec.ts
dheepak-aot: To test the scenario that 

-> if a aviation restriction for the credential type already exist, it must create another restriction

Bypass was just the way to reach the step.

sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/e-cert-steps-utils.ts
andrewsignori-aot: Extending my point, this method should be generic and must not require modifications if new action metadata needs to be evaluated.
For instance, here we have the `aviationCredentialType` explicitly passed as a parameter when it should be known only by a specialized method evaluating the conditions.

For instance, the code below would be generic and not require further modifications.
```ts
export function getRestrictionByActionType(
  eCertDisbursement: EligibleECertDisbursement,
  actionType: RestrictionActionType,
): StudentActiveRestriction {
  return eCertDisbursement
    .getEffectiveRestrictions()
    .find(
      (restriction) =>
        restriction.actions.includes(actionType) &&
        isRestrictionActionEffective(
          eCertDisbursement,
          restriction.actionEffectiveConditions,
        ),
    );
}
```

Following this, the `ConditionEvaluator` can be changed as below.
```ts
type ConditionEvaluator = (
  eCertDisbursement: EligibleECertDisbursement,
  conditionValue: unknown,
) => boolean;
```
```ts
/**
 * Aviation credential type condition evaluator.
 * @param actionEffectiveCondition condition to be evaluated.
 * @param eCertDisbursement eCert disbursement context data required to evaluate the condition.
 * @returns true if the condition is satisfied or no aviation credential type is provided, otherwise false.
 */
function aviationCredentialTypeCondition(
  eCertDisbursement: EligibleECertDisbursement,
  conditionValue: unknown,
): boolean {
  const conditionValues = conditionValue as string[];
  return conditionValues.includes(
    eCertDisbursement.offering.aviationCredentialType,
  );
}
```
As a minor side note, I believe the `isRestrictionActionEffective` should be moved inside the method `getEffectiveRestrictions`.

If the benefits are not clear or if I am missing something, we can talk over a call.

dheepak-aot: As I said earlier, the proposed approach has an advantage and a disadvantage as per my understanding.

Advantage:

If new conditions added that work based on different context data, then the place where `isRestrictionActionEffective` is consumed remains unchanged.

Disadvantage: 

If the consumption becomes essential outside the e-cert part, then it will not be generic anymore. 


My outcome:
Considering the fact that currently it is evaluated only at e-cert I can use a type derived from `EligibleECertDisbursement` and open to modification. 

We can have a call based on my points here and your inputs.

sh16011993 (approved):
Exemplary work @dheepak-aot 👍 Thanks for the call clarifying my doubts.

andrewsignori-aot (approved):
Thanks for making the changes and for the great discussions, looks good 👍

PR #5207, opened 2025-09-23
#2988 - Aviation restriction tracker card API and UI updates

dheepak-aot:
## ECert validation to get effective restriction for an action type enhanced

- [x] Updated the condition in `ecert-steps-util` to return all restrictions that are effective for the action type provided, so that all the effective restriction codes can be recorded to additional information. 

## ECert pre-validator enhanced to provide additional information about failed validations

- [x] Updated the e-cert pre-validator framework to provide more additional information about failed validations.

## APIs returning e-cert failed validations enhanced 
- [x] APIs (Get completed application details and get application warnings) updated to return additional API context on failed validations. 

## Banners displaying messages based on e-cert failed validations updated

- [x] Updated accept assessment banner with additional message on the top.
image

- [x] Updated disbursement blocked banner with additional message on the top.

image

## Student account activity page update

image

## E2E Tests 

- [x] Added E2E tests for the completed application details API for student. (Added tests to only student API)
image

- [x] Added E2E test for the get application warnings API for student. 
image

### DIFF
--- a/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/e-cert-steps-utils.ts
+++ b/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/e-cert-steps-utils.ts
@@ -19,7 +19,7 @@ import { isRestrictionActionEffective } from "./e-cert-steps-restriction-utils";
  * will not be included in the result.
  * @param eCertDisbursement student disbursement to check student restrictions.
  * @param actionType action type.
- * @returns the first restriction of the requested action type.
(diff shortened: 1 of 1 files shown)


sources/packages/forms/src/form-definitions/studentaccountactivity.json
andrewsignori-aot, line 1: Was there a requirement not to add the legacy restriction to the list?
Its type is defined as "Error," and unless we are not displaying legacy ones to the student, it should have content. Does it make sense?

ninosamson (approved):
Looks good and thank you for providing the details on the various content additions.

Issue #4967, opened 2025-07-22
2.12: Content: Minor grammar fixes (to be updated)


astridSABC:
**Describe the task**
A clear and concise description of what the task is.

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

PR #5212, opened 2025-09-26
#4967 - Upgrade Packages - Part 1

andrewsignori-aot:
## API-related changes
- Tried to keep the configuration as simple as possible.
- Instead of disabling `@typescript-eslint/no-explicit-any` (and others), it is now set as a warning to make it more explicit and receive a potential on-demand fix.
- `webpack.config.js‎` converted to an ES module. No changes were made other than the conversion.
 
### Main packages migrated
- typescript
- ESLint migrated to version 9.
- webpack
  - copy-webpack-plugin

## Web-related changes
- Used the helper package [@vue/eslint-config-typescript](https://www.npmjs.com/package/@vue/eslint-config-typescript/v/14.1.0) for easy configuration setup for vue+ts+eslit.
- Fixed a few files with minor issues.
- Changed the Vue plugin to use `strongly-recommended` rules instead of `recommended`.
  - Kept a few rules as warnings to allow the issues to be explicit and allow a DEV to fix when touching a file with a warning. For instance, non-hyphenated properties now will be displayed as a warning.
image

### Main packages migrated
- typescript
- ESLint migrated to version 9.


sources/packages/backend/eslint.config.ts
```diff
@@ -0,0 +1,29 @@
...
+    languageOptions: {
+      globals: {
+        ...globals.node,
+        ...globals.jest,
+      },
+    },
+    rules: {
+      "no-unsafe-optional-chaining": "warn",
+      "no-prototype-builtins": "warn",
+      "@typescript-eslint/no-explicit-any": "warn",
```
dheepak-aot, line 20: 👍

sources/packages/backend/eslint.config.ts
```diff
@@ -0,0 +1,29 @@
...
+      "@typescript-eslint/interface-name-prefix": "off",
```
dheepak-aot, line 21: I could not see this rule anymore in https://typescript-eslint.io/rules/

Is it still a valid one?

andrewsignori-aot: Removed.

sources/packages/backend/package.json
```diff
@@ -120,24 +120,25 @@
...
-    "@typescript-eslint/parser": "^5.39.0",
-    "copy-webpack-plugin": "^11.0.0",
+    "copy-webpack-plugin": "^13.0.1",
     "cross-env": "^7.0.3",
-    "eslint": "^8.24.0",
-    "eslint-config-prettier": "7.0.0",
+    "eslint": "^9.36.0",
+    "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-prettier": "^3.1.4",
+    "eslint-plugin-vue": "^10.5.0",
```
dheepak-aot, line 128: Why did we add this to the backend?

andrewsignori-aot: Removed.

sources/packages/backend/package-lock.json
```diff
@@ -83,25 +83,25 @@
...
         "cross-env": "^7.0.3",
-        "eslint": "^8.24.0",
-        "eslint-config-prettier": "7.0.0",
+        "eslint": "^9.36.0",
+        "eslint-config-prettier": "^10.1.8",
         "eslint-plugin-prettier": "^3.1.4",
         "faker": "^5.2.0",
         "find-config": "^1.0.0",
         "jest": "^29.6.4",
+        "jiti": "^2.6.0",
```
dheepak-aot, line 94: Why was this package added?

andrewsignori-aot: To allow the use of a ts file without using the experimental node options for now.
Once the config is changed to ts, there is a warning asking to have this dev-only package added.
https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files

ninosamson (approved):

Issue #4964, opened 2025-07-22
Upgrade Packages


ninosamson:
Upgrade the following packages, which were not upgraded in last round of package updates. 

**Priority**
- [x] esLint
- [x] Dependabot current upgrade PRs: Minor/Patch -> Major

**Nice to Have**
- [x] Nodejs - nice to have minimal
- [x] Bull
- [ ] Redis
- [x] Typescript

andrewsignori-aot:
@ninosamson can we have a new ticket to have Redis-cluster updated and its helm updated to follow the same approach from MongoDB?

Also, we need to update the web server. The version of nginx is from 2022, and we can follow the same approach used for NodeJS, and no longer rely on the Red Hat images.

PR #5216, opened 2025-10-01
#4964 - Upgrade Packages - Nestjs Only

andrewsignori-aot:
Updated Nestjs only packages.


sources/packages/backend/libs/integrations/src/object-storage/object-storage.service.ts
```diff
@@ -65,7 +66,7 @@ export class ObjectStorageService {
     return {
       contentLength: response.ContentLength,
       contentType: response.ContentType,
-      body: response.Body,
+      body: response.Body as Readable,
```
andrewsignori-aot, line 69: Required once the `@types/node` was updated.

dheepak-aot (approved):
Good Job!. Looks good 👍

ninosamson (approved):

PR #5226, opened 2025-10-03
#4964 - Upgrade Packages - Node Image Update

andrewsignori-aot:
- Updated nodejs to [node:24.9.0-alpine3.21](https://hub.docker.com/_/node).
- Set the `npm_config_cache` to allow permissions to be added to the proper folder, since, depending on the user running the command, it may differ.
- Stop using Red Hat images, and get the node images directly from the Docker-remote artifactory.
- Web Docker files were adjusted to look more similar between prod and non-prod.
- Disabled the Camunda disk cache by configuring `CAMUNDA_TOKEN_DISK_CACHE_DISABLE` as true. Enabling the disk cache would require permission to the Camunda cache token folder, which was not considered critical. A memory cache is also used, and the disk cache is only used in case the POD restarts, which would reattach the POD volume. More about [Camunda token cache](https://github.com/camunda/camunda-8-js-sdk?tab=readme-ov-file#token-caching).
- Removed `maintainer` label that was in some of the Docker files. There was no actual reason to keep them.
- Removed `buildArgs` from docker-build that were causing warnings during the build.
- Adjusted the package names to `student-aid-bc-backend` and `student-aid-bc-frontend` for better log display during build.
- Explicitly copy web folders and files, resolving a Sonarcloud issue.

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/Dockerfile.dev
+++ b/sources/packages/backend/apps/queue-consumers/Dockerfile.dev
@@ -1,9 +1,6 @@
 # Base Image
 FROM node:24.9.0-alpine3.22
 
-# Updating container path variable with node bin
-ENV PATH=$PATH:/app/node_modules/.bin
-
 WORKDIR /app
 
 # Queue Port


sources/packages/backend/apps/load-test-gateway/Dockerfile.dev
```diff
@@ -1,15 +1,13 @@
 # Base Image
-FROM node:22.13.1-alpine3.20
+FROM node:24.9.0-alpine3.22
 
-LABEL maintainer="BC GOV"
```
dheepak-aot, line 4: 👍

sources/packages/backend/apps/queue-consumers/Dockerfile.dev
```diff
@@ -1,15 +1,13 @@
 # Base Image
-FROM node:22.13.1-alpine3.20
-
-LABEL maintainer="BC GOV"
+FROM node:24.9.0-alpine3.22
 
 # Updating container path variable with node bin
-ENV PATH $PATH:/app/node_modules/.bin
```
dheepak-aot, line 7: If the variable was never set due to invalid syntax, it is required anymore?

sources/packages/backend/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "api",
+  "name": "studentaid-backend",
   "version": "0.0.1",
   "description": "",
```
dheepak-aot, line 4: 👍

sources/packages/backend/apps/queue-consumers/Dockerfile.dev
```diff
@@ -1,15 +1,13 @@
...
+ENV PATH=$PATH:/app/node_modules/.bin
 
 WORKDIR /app
 
 # Queue Port
-ENV PORT 3001
+ENV PORT=3001
```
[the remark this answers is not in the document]
andrewsignori-aot: Updated.

PR #5236, opened 2025-10-04
#4964 - Upgrade Packages - Typescript Configs Improvements

andrewsignori-aot:


sources/packages/backend/libs/utilities/src/logger/logger.service.ts
```diff
@@ -1,13 +1,19 @@
...
 import { StringBuilder } from "@sims/utilities/string-builder";
 import { parseJSONError } from "@sims/utilities/parse-json";
+import { INQUIRER } from "@nestjs/core";
 
 /**
  * Common log across entire solution.
  */
 @Injectable({ scope: Scope.TRANSIENT })
 export class LoggerService extends ConsoleLogger {
+  constructor(@Inject(INQUIRER) parentClass?: object) {
```
dheepak-aot, line 12: Why is the parentClass set as optional?

sources/packages/backend/apps/api/src/services/bceid/bceid.service.ts
```diff
@@ -221,15 +224,12 @@ export class BCeIDService {
    * Gets WSDL authentication header needed to retrieve the WSDL from BCeID Web Service.
    * @returns Authorization header converted to base64 string.
    */
-  private getWsdlAuthHeader() {
+  private getWsdlAuthHeader(): { Authorization: string } {
```
dheepak-aot, line 227: is it necessary to use the parameter name with upper case?

sources/packages/backend/apps/api/src/app.exception.filter.ts
```diff
@@ -1,24 +1,24 @@
-import { Catch, ArgumentsHost } from "@nestjs/common";
+import { Catch, ArgumentsHost, Injectable } from "@nestjs/common";
 import { BaseExceptionFilter } from "@nestjs/core";
 import { Request } from "express";
-import { LoggerService, InjectLogger } from "@sims/utilities/logger";
+import { LoggerService } from "@sims/utilities/logger";
 
 @Catch()
+@Injectable()
```
dheepak-aot, line 7: Asking for clarification, is the injectable decorator required here? (I may have been misled seeing the nest example that `@Catch` alone must be sufficient.

andrewsignori-aot: It is redundant; just tested and removed it.

sources/packages/backend/apps/queue-consumers/src/services/application/application.service.ts
```diff
@@ -28,6 +28,12 @@ interface SecondDisbursementStillPending {
 
 @Injectable()
 export class ApplicationService {
+  /**
+   * Sub query to validate if an application has assessment already being
+   * processed by the workflow.
+   */
+  private readonly inProgressStatusesExistsQuery: string;
```
dheepak-aot, line 35: 👍

sources/packages/backend/libs/auth/src/services/keycloak/keycloak.service.ts
```diff
@@ -41,9 +39,11 @@ export class KeycloakService {
    * injection framework.
    */
   public static get shared(): KeycloakService {
+    const logger = new LoggerService();
+    logger.setContext(KeycloakService.name);
```
dheepak-aot, line 43: As we can pass the parent class to the constructor, what about declaring this way? `const logger = new LoggerService(KeycloakService);`

dheepak-aot (approved):
Great and extensive work @andrewsignori-aot. Thanks for making the change. Looks good 👍

sources/packages/backend/libs/auth/src/services/keycloak/keycloak.service.ts
andrewsignori-aot: It would not result in the same output, the constructor is expecting an instance.
I was going to create an overload to accept the string, but there are only two places not using dependency injection right now, so I did not invest time in it.

ninosamson (approved):

Issue #5166, opened 2025-09-10
Individual Exception Approval


Joshua-Lakusta:
## User story

As a ministry user I want to be able to approve exceptions individually and write an comment/message to the student within the exceptions approval ui. Currently, the Exceptions Ministry Approval UI forces ministry users to approve or deny all exceptions in a batch which does not align well with the new do not repeat approved exceptions functionality.

## Acceptance Criteria

- [x] Update UI so that Ministry users can approve/deny each exception individually.
- [x] Radio button for approve or deny for each exception
- [x] One Internal Ministry Note per submission (not per decision we still only need the one note section)
- [x] Will need to make a submit button on the bottom of the exceptions decision page that will validate all exceptions have a decision and the ministry user can only finalize request once all exceptions have a decision.
- [x] In order for the application to proceed all exceptions must have an approved status.
- [x] Process to check for previously approved exceptions must be adapted to check each individual approvals instead of the overall approval

dheepak-aot:
@ninosamson as per our discussion during the implementation, the BC public institutions can view the the individual exception status i.e. the `Approved | Declined` status on each individual exception in the institution portal view of application exceptions. @Joshua-Lakusta FYI.

<img width="864" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBenE5Q3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--32505df7e537834d7f4a7a7a8c3ff7be82ae2976/image.png" alt="image.png" />

dheepak-aot:
## Ministry submit exception decision - Toast message update

@ninosamson @Joshua-Lakusta The current toast message on ministry user approving or declining an exception is `Application exception status is now Approved | Declined.` but with the change in feature, the exception status is no more provided my the ministry user, but derived based on individual exception requests. Hence, I updated the toast message to be more generic like this. `Application exception decision submitted successfully.`  Let me know if that sounds acceptable..

PR #5243, opened 2025-10-06
#5166 - Individual exception approval - DB migration and worker updates

dheepak-aot:
# Individual exception approval - DB migration and worker updates

## DB Migration

- [x] Created a new enum `sims.application_exception_request_status` in DB for exception request status.
- [x] Created a new column `exception_request_status` in `sims.application_exception_requests` to enable individual application exception requests to be Approved/Declined.
- [x] Update query to set the status of existing application exception requests which belong to exceptions which are completed(Approved/Declined). The individual exception requests will reflect the status of it's exception.

## Worker updates

- [x] The query to select previously approved exception requests is now based on the `exceptionRequestStatus` instead of it's exception status.
- [x] When a previous approved exception request was found, then the status of current exception request is set to `Approved`. If otherwise it stays as `Pending`.
- [x] The worker that supports exception creation for previous PY `verifyApplicationExceptions` updated to set the exception request status as `Pending` during exception creation.

## E2E Tests
- [x] Adjusted the E2E tests for the workers `verifyUniqueApplicationExceptions` and `verifyApplicationExceptions`. 
- [x] Added new E2E test for `verifyUniqueApplicationExceptions` to validate the created exception request status to be based on previous exception request status, and irrespective of it's previous exception status.

image

### DIFF
--- a/sources/packages/backend/apps/db-migrations/src/sql/ApplicationExceptionRequests/Add-col-exception-request-status.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/ApplicationExceptionRequests/Add-col-exception-request-status.sql
@@ -13,7 +13,17 @@ UPDATE
 SET
     -- The value for the statuses approved and declined are same between the types sims.application_exception_request_status and sims.application_exception_status
     -- and hence we can directly cast the text value from one type to another.
-    exception_request_status = (exception.exception_status :: text) :: sims.application_exception_request_status
+    exception_request_status = (exception.exception_status :: text) :: sims.application_exception_request_status,
+    modifier = (
+        SELECT
+            id
+        FROM
+            sims.users
+        WHERE
+            -- System user.
+            user_name = '8fb44f70-6ce6-11ed-b307-8743a2da47ef@system'
+    ),
+    updated_at = NOW()
 FROM
     sims.application_exceptions exception
 WHERE
--- a/sources/packages/backend/apps/db-migrations/src/sql/ApplicationExceptionRequests/Add-col-exception-request-status.sql
+++ b/sources/packages/backend/apps/db-migrations/src/sql/ApplicationExceptionRequests/Add-col-exception-request-status.sql
@@ -1,0 +1,27 @@
+ALTER TABLE
+    sims.application_exception_requests
+ADD
+    COLUMN exception_request_status sims.application_exception_request_status NOT NULL DEFAULT 'Pending';
+
+COMMENT ON COLUMN sims.application_exception_requests.exception_request_status IS 'Status of the application exception request.';
+
+-- Update the status of existing exception requests which belong to completed exceptions(approved or declined).
+-- If the exception is approved then the status of all the associated exception requests should be set to approved.
+-- If the exception is declined then the status of all the associated exception requests should be set to declined.
+UPDATE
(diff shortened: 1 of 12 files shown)


sources/packages/backend/apps/db-migrations/src/sql/ApplicationExceptionRequests/Add-col-exception-request-status.sql
```diff
@@ -0,0 +1,27 @@
...
+    COLUMN exception_request_status sims.application_exception_request_status NOT NULL DEFAULT 'Pending';
+
+COMMENT ON COLUMN sims.application_exception_requests.exception_request_status IS 'Status of the application exception request.';
+
+-- Update the status of existing exception requests which belong to completed exceptions(approved or declined).
+-- If the exception is approved then the status of all the associated exception requests should be set to approved.
+-- If the exception is declined then the status of all the associated exception requests should be set to declined.
+UPDATE
+    sims.application_exception_requests exception_request
+SET
```
andrewsignori-aot, line 13: Should we also set the audit columns while executing this update?

sources/packages/backend/libs/sims-db/src/entities/application-exception-requests.model.ts
```diff
@@ -71,4 +72,16 @@ export class ApplicationExceptionRequest extends RecordDataModel {
...
+
+  /**
+   * Status of the application exception request.
+   */
+  @Column({
+    name: "exception_request_status",
+    type: "enum",
+    enum: ApplicationExceptionRequestStatus,
+    enumName: "ApplicationExceptionRequestStatus",
+    nullable: false,
```
andrewsignori-aot, line 84: Please remove the `nullable: false` since it is the default value and we are no longer explicitly adding them.

sources/packages/backend/apps/workers/src/services/application-exception/application-exception.service.ts
```diff
@@ -87,22 +89,33 @@ export class ApplicationExceptionService extends RecordDataModelService<Applicat
...
-        ({
+      (exceptionRequest) => {
+        const approvalExceptionRequest = this.getPreviouslyApprovedException(
+          exceptionRequest,
+          previouslyApprovedExceptionRequests,
+        );
+        // If the exception request with the same name and hash was previously approved
+        // then the status of the exception request will be set to approved.
+        // Otherwise the status will be set to pending which requires ministry assessment.
+        const exceptionRequestStatus = approvalExceptionRequest?.id
```
andrewsignori-aot, line 100: Is the "id" check required, or would the below be enough?
```ts
const exceptionRequestStatus = approvalExceptionRequest
      ? ApplicationExceptionRequestStatus.Approved
      : ApplicationExceptionRequestStatus.Pending;
```

sources/packages/backend/apps/db-migrations/src/sql/ApplicationExceptionRequests/Add-col-exception-request-status.sql
dheepak-aot: I believe it would be better to have the audit columns updated. I can update.

andrewsignori-aot (approved):

ninosamson (approved):

PR #5245, opened 2025-10-07
#4964 - Upgrade Packages - Layout fix and dependabot updates

andrewsignori-aot:
- Fixed banners that were not displayed correctly after the `@fortawesome/fontawesome-free` upgrade.

image

image

image

- Grouped form. io-related packages for dependabot PRs.
- Added monitoring of upgrades for Docker versions.
- Added a "commit-message" that should be the prefix for the PR titles, trying to make it clear where they are. We can remove it later if it does not achieve the goal.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,8 @@ version: 2
 updates:
   - package-ecosystem: "npm"
     directory: "/sources/packages/web"
+    commit-message:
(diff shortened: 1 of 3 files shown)


ninosamson (approved):

dheepak-aot (approved):
Looks good 👍 Thanks for adding the commit messages for dependabot.

Issue #5089, opened 2025-08-21
BUG: Mask required instead of blank for institution student total income used for assessment purposes


astridSABC:
**Describe the Bug**
As an institution user, I need to see the "total income used for assessment purposes" as masked ($XXXXX) instead of just blank

**Expected Behaviour**
<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNTBpQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d9cdcf7e78404c32db41107eb17b28421d673a28/image.png" alt="image.png" />

Ref #3896

**Actual Behaviour**

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNThpQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--948405d02a05f6ee3403c421d70e2f98fa76bd03/image.png" alt="image.png" />

### Steps To Reproduce

Precondition: Student has submitted an application and it is at assessment status

1. Login to the institution portal
2. Select "search students"
3. Search for the student with the application above
4. Click "view"
5. Select "Applications" tab
6. Click "view" beside the application at assessment
7. Select "assessments" in the left pane
8. Click "view assessment"
9. Click "view assessment"
10. Note the funding section
11. Note the "Total income used for assessment purposes:"

- Investigating confirmed that XXX still being returned in backend. Just need to fix UI bug. 

### NOTE
Email sent to BAA, SBA and SPA for additional context.

CarlyCotton:
Looking to get clarification on why the mask is not showing as requested ($XXXXX) and just showing blank. THis will help prioritization in case could possibly show income to institution users.

ninosamson:
Investigate further in network console to confirm if XXXX is still being returned, indicating it would be a UI issue.

PR #5328, opened 2025-10-29
#5089 - Mask required instead of blank: Institution student total income

sh16011993:
**As a part of this PR, passed the required 2nd argument mask *XXXXX* to the `currencyFormatter` utils method ensuring that the empty is replaced by the *XXXXX* mask.**

### **Screenshot:**

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/noticeofassessment.json
+++ b/sources/packages/forms/src/form-definitions/noticeofassessment.json
@@ -424,7 +424,7 @@
               "value": ""
             }
           ],
-          "content": "Total income used for assessment purposes: <span class=\"label-bold\">{{ utils.custom.currencyFormatter(data.assessment?.totalFamilyIncome, \"XXXXX\") }}</span>",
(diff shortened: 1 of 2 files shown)


sources/packages/forms/src/form-definitions/noticeofassessment.json
```diff
@@ -424,7 +424,7 @@
               "value": ""
             }
           ],
-          "content": "Total income used for assessment purposes: <span class=\"label-bold\">{{ utils.custom.currencyFormatter(data.assessment?.totalFamilyIncome) }}</span>",
+          "content": "Total income used for assessment purposes: <span class=\"label-bold\">{{ utils.custom.currencyFormatter(data.assessment?.totalFamilyIncome, \"XXXXX\") }}</span>",
```
andrewsignori-aot: If `data.assessment?.totalFamilyIncome` is null or undefined, it will display the masked "XXXXX" value for Ministry and even Students, which is not expected.
Since these "masked values" are common, we can incorporate them in the `currencyFormatter` where, if the value `isNaN` but it is "fully masked" with "X", we just return the "X".

andrewsignori-aot (commented):
Please take a look at the comment.

Issue #5183, opened 2025-09-15
New appeals UI appeal history 


Joshua-Lakusta:
## User Story
As a student I want to be able to see my appeal submission history. This will allow me to understand what appeals I have submitted and what the outcomes were.

## Acceptance Criteria
- [ ] Create a secondary tab on the appeals UI that shows appeals history 
    - [ ] Tab one is Appeal Submission
    - [ ] Tab two (new tab) Appeal History
- [ ]  Create a list on the history tab that shows all submitted or completed appeals
- [ ]  Columns 
Decision Date  |   Application  |  Appeal Name  |  Status |  View

- [ ] Decision Date format will follow MMM DD, YYYY of the decision (blank if it is pending)
- [ ] If application # exists show it under application column if it does not it should just be blank
- [ ] Appeal Name, if application appeal and there is only one selected form show name if multiple selected state Multiple. For "other" appeals it will always just be the form name.
- [ ] Status should show as approved, denied, or pending
- [ ] View button under view column should link to the appeal form for each type of appeal category submission
    - [ ] The application appeals will link to the page that currently exists 
    - [ ] The other appeals will need a view that should just show the appeal form submitted and a decision area at the bottom that mirrors the current behavior of the application appeals. (shows the form with a decision label next to the top.
<img width="1301" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMjlrQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e0326719115ea7799463f183c4572d2b21a8814f/image.png" alt="image.png" />

<img width="1319" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNDFsQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b4573b7bc859caf6aa3bcce67c020ea721d75fcd/image.png" alt="image.png" />

### Consider adding new tickets for: 
- Ministry and Institution UI to view student appeal history

**Additional Context**
General concept is to allow students to see other restriction out comes but also be able to see all appeal history.
Previous ui ticket #5153

Joshua-Lakusta:
@dheepak-aot 
Q. I believe that the appeal history table must exclude the legacy change requests(change requests before PY 2025-26) which was using the appeals framework. Please confirm.
[...]

PR #5343, opened 2025-11-04
#5183 - New appeals UI appeal history

andrewsignori-aot:
## New "Appeal History" tab in the Student Appeals

image

_Note:_ Many files were impacted by small navigation changes due to the new tabs, and some were moved.

### New Student Appeals Status Chip

- Create a new status chip status `<status-chip-student-appeal>` to allow the appeal requests status to be displayed correctly for "Approved" appeals. Please note, the "decision status" is still displayed using the `<status-chip-requested-assessment>` that is the same used for the visualization in the application details and the appeal read-only page.
- `StudentAppealSharedForm.vue` was converted to have the tabs and to be shared for the application and student routes.

## E2E Tests

- E2E tests adjusted to include new values returned from the API.
- Added an order to the appeals request to ensure consistent return of the list. There is no requirement for a particular order, so an order was chosen.

### DIFF
--- a/sources/packages/web/src/views/student/appeal/StudentAppealSharedFormSubmission.vue
+++ b/sources/packages/web/src/views/student/appeal/StudentAppealSharedFormSubmission.vue
@@ -3,7 +3,7 @@
     <template #header>
       <body-header
         title="Submission"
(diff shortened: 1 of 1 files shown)


sh16011993 (approved):
Great work @andrewsignori-aot 👍

Issue #5203, opened 2025-09-19
Modified Independent appeal: New "other" appeal to update profile


CarlyCotton:
As a ministry user and student user we want the modified independent process to be permanent when it gets completed. We want it to be an appeal that can be completed at any time. 

### Other appeal activation
- [x] Recode the other appeal option on the new appeals ui to be visible (hardcoded hidden on previous ticket)
- [x] Add the new modified independent appeal into the other appeal category drop down list

### Appeal outcome 
- [x] Outcome of the new appeal will not effect any application it will set a new status to the profile
- [x] When appeal is approved, student's modified independent profile status is changed to yes 

### Appeal form
- [x] Appeal form is named Modified independent appeal
- [x] Form content:
Is it impossible for you to provide your parents' financial information and you need to request consideration for modified independent status due to exceptional circumstances? 
Yes
No

Which of the following best describes your situation? (If none of these apply you are not eligible for modified independent status)
   - There is an irreconcilable rift between me and my parent(s), and I can provide satisfactory evidence of a severe and permanent breakdown in the relationship between you and your parent(s) that has existed for at least one year unless a sudden and serious situation has occurred.
   - You would normally be classified as a dependant student by StudentAid BC, but you are unable to contact or locate your parents, step-parent, sponsor, or legal guardian due to conflict or emergency (such as war, famine or natural disaster) in the country where they live.

- [x] Uploader if option 1 selected 

Upload supporting documents
Please provide all of the documents listed below
A statement from you describing the family rift, how long the rift has existed and why it continues.
A letter from a responsible third party testifying as to the truthfulness of your explanation and providing an independent assessment of the situation. The third party could be any of the following: counsellor, psychologist, religious advisor, doctor, teacher, and/or mature relative outside the immediate family or family friend. The status of the third party and their relationship to you must be identified in the letter.
A letter from your parent(s)/step-parent/sponsor/legal guardian describing the family rift, how long the rift has existed and why it continues. If a parental letter cannot be obtained, a second letter from another professional third party must be submitted.

- [x] Uploader if option 2 selected

Upload supporting documents
Please provide all of the documents listed below
A letter explaining the conflict or emergency that has made it impossible to contact or locate your parent(s)/step-parent/sponsor/legal guardian in the country where they live.

## QA Notes
[...]

PR #5357, opened 2025-11-12
#5203 - Modified Independent appeal - Fix Wrong Content

andrewsignori-aot:
Removed an extra bullet point that was not part of this requirement.

image


andrewsignori-aot:
@ninosamson bypassing approval in this PR to allow its deployment to TEST.

PR #5428, opened 2025-11-28
#2470 - Step parent waiver appeal - Form and consolidated data

dheepak-aot:
# Step parent waiver appeal - Form and consolidated data

Summary: The PR implementation enables the student to submit a new type of appeal call `Step parent waiver appeal` on full-time applications with both parent 1 and parent 2 reported in the application. If the application does not have both the parents reported, then API validates the request and throw error on the submission. 

## FormIO

- [x] Created new form `stepparentwaiverappeal` for the `Step parent waiver appeal`.
- [x] Form loads the parent names in dynamic way using the property `parents` injected by Vue.

image

image

## API
- [x] Updated the API `get application to request appeal` to return parent details.
- [x] During the appeal submission for step parent waiver appeal, the application is validated to have both parents reported.
- [x] During appeal submission, the property `parents` if present is re-populated at the server side to perform dry run validation. 
**Note**: The re-populated data is saved in the appeal submitted data payload.

## Workflow
- [x] Load assessment consolidated data mapping is added for the step parent waiver appeal data. As `selectedParent`(supporting user id of the waived parent) is only required for the workflow context, only that property is mapped into the appeal data instead of the whole dynamic data.
image

- [x] Added `parent1SupportingUserId` and `parent2SupportingUserId` which to consolidated data which will be used by the workflow to match with `selectedParent` in step parent waiver appeal data to identify the waived parent.
image

image

## Workers
- [x] Parent 1 and Parent 2 ordered by the supporting user id for the consistency in identifying parent 1 and parent 2.

## E2E Tests

- [x] Created E2E Test `ApplicationStudentsController(e2e)-getApplicationToRequestAppeal` with some tests including the applications with supporting user parents.
```
ApplicationStudentsController(e2e)-getApplicationToRequestAppeal
    √ Should get the application details for appeal when the application status is Completed and offering intensity is full-time. (964 ms)                                               
    √ Should get the application details for appeal when the application status is Completed and offering intensity is full-time and the application has both the parents reported as supporting users. (811 ms)                                                                                                                                                                  
    √ Should throw not found error when application is not found. (37 ms)
```
- [x] Added E2E tests for step parent waiver appeal in `StudentAppealStudentsController(e2e)-submitApplicationAppeal`
```
√ Should create step parent waiver appeal for an application when student submit the appeal for a full-time application reported with both the parents. (1201 ms)
√ Should throw unprocessable entity exception when student submit a step parent waiver appeal for an application which has only one reported parent. (1021 ms)

```
- [x] Added worker E2E test in `AssessmentController(e2e)-loadAssessmentConsolidatedData` to ensure that the parents are identified as `Parent1` and `Parent2` based on the id in the `supporting_users` table. For e.g. if application has 2 parents and say 100 and 102 are the ids of the parents, then parent with id 100 is `Parent1` and with 102 is `Parent2`.
```
 √ Should load assessment consolidated data when the student is single and a dependant with both parents reported in the application and the application offering intensity is full-time. (729 ms)
```

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
@@ -202,7 +202,7 @@ export class StudentAppealStudentsController extends BaseController {
     try {
       const dryRunPromise: Promise<DryRunSubmissionResult>[] =
         payload.studentAppealRequests.map((appeal) => {
-          // Check if the form has any inputs which is required to be populated at the server side
+          // Check if the form has any inputs which are required to be populated at the server side
           // during the dry run submission.
           if (appeal.formData.programYear) {
             appeal.formData.programYear = application.programYear.programYear;
--- a/sources/packages/forms/src/form-definitions/stepparentwaiverappeal.json
+++ b/sources/packages/forms/src/form-definitions/stepparentwaiverappeal.json
@@ -69,7 +69,7 @@
             "tableView": true,
             "validate": {
                 "required": true,
-                "custom": "let validationErrorMessage;\n\n// Get allowed values from injected property 'parents'.\nconst allowedValues = !!data.parents\n    ? data.parents.map((parent) => parent.id)\n    : [];\n// Validate and populate the validation error message.\nif (allowedValues.length < 2) {\n  validationErrorMessage = \"Invalid number of parents in the application.\";\n} else if (!allowedValues.includes(input)) {\n  validationErrorMessage = \"Invalid value for the selected parent.\";\n}\n\n// When validation error message is missing it is considered as successful validation.\nvalid = validationErrorMessage || true;\n  "
+                "custom": "let validationErrorMessage;\n\n// Get allowed values from injected property 'parents'.\nconst allowedValues = !!data.parents\n    ? data.parents.map((parent) => parent.id)\n    : [];\n// Validate and populate the validation error message.\nif (allowedValues.length !== 2) {\n  validationErrorMessage = \"Invalid number of parents in the application.\";\n} else if (!allowedValues.includes(input)) {\n  validationErrorMessage = \"Invalid value for the selected parent.\";\n}\n\n// When validation error message is missing it is considered as successful validation.\nvalid = validationErrorMessage || true;\n  "
             },
             "validateWhenHidden": false,
             "key": "selectedParent",
--- a/sources/packages/backend/apps/api/src/utilities/system-configurations-constants.ts
+++ b/sources/packages/backend/apps/api/src/utilities/system-configurations-constants.ts
@@ -146,9 +146,3 @@ export const AVIATION_RESTRICTION_CODES: RestrictionCode[] = [
   RestrictionCode.AVIR,
   RestrictionCode.SFAS_AV,
 ];
-
-/**
- * Number of parents required to be reported in the application
- * to submit the step parent waiver appeal.
- */
-export const STEP_PARENT_WAIVER_APPEAL_REQUIRED_PARENTS = 2;


sources/packages/backend/apps/api/src/route-controllers/student-appeal/_tests_/e2e/student-appeal.students.controller.submitApplicationAppeal.e2e-spec.ts
```diff
@@ -732,6 +734,224 @@ describe("StudentAppealStudentsController(e2e)-submitApplicationAppeal", () => {
...
+      // and this mock must be removed.
+      const formService = await getProviderInstanceForModule(
+        appModule,
+        AppStudentsModule,
+        FormService,
+      );
+      const dryRunSubmissionMock = jest.fn().mockResolvedValue({
+        valid: true,
+        formName: FormNames.StepParentWaiverAppeal,
+        data: { data: expectedSubmittedData },
```
andrewsignori-aot: I am not able to follow the reason why the `expectedSubmittedData` should be provided as a result of the mock. I believe that we must let the server-side override these values and assert it (as it happens at line 879).
Providing the expected data as a mock result does not seem needed.

sources/packages/forms/src/form-definitions/stepparentwaiverappeal.json
```diff
@@ -0,0 +1,197 @@
...
+            "isNew": false
+        },
+        {
+            "label": "Please indicate which parent should be waived from the application",
+            "optionsLabelPosition": "right",
+            "inline": false,
+            "tableView": true,
+            "validate": {
+                "required": true,
+                "custom": "const allowedValues = !!data.parents\n  ? data.parents.map((parent) => parent.id)\n  : [];\n  \nvalid = allowedValues.includes(input) ? true : \"Invalid value for selected parent\";"
```
andrewsignori-aot: We should have a custom validation here to enforce that two parents should be present for selection. Since the hidden fields do not allow validation, it would enforce the rule at the dry run level, keeping the form-specific validation inside the form.io itself.

andrewsignori-aot (commented):
Great work, thanks for adding the extra E2E, and nice way to handle the dynamic parents loaded into the form.io. Please take a look at the comments.

sources/packages/backend/apps/api/src/route-controllers/student-appeal/student-appeal.students.controller.ts
```diff
@@ -197,16 +200,29 @@ export class StudentAppealStudentsController extends BaseController {
...
     }
     let dryRunSubmissionResults: DryRunSubmissionResult[] = [];
+    const parents = getSupportingUserParents(application.supportingUsers);
+    const hasStepParentWaiverAppealSubmission =
+      payload.studentAppealRequests.some(
+        (appealRequest) =>
+          appealRequest.formName === FormNames.StepParentWaiverAppeal,
+      );
+    // Validate the number of parents for step parent waiver appeal.
+    if (hasStepParentWaiverAppealSubmission && parents.length < 2) {
```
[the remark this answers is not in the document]
dheepak-aot: You mean this validation being part of the custom validation executed in the form?

If so, as the custom validation is has always validated only input value provided by the component and not the count of the data set(values), I felt it should not be in the custom validation and moreover this validation is only temporary until we have the eligible appeals for an application is built. Still open for ideas.

sources/packages/forms/src/form-definitions/stepparentwaiverappeal.json
```diff
@@ -0,0 +1,197 @@
...
+            "isNew": false
+        },
+        {
+            "label": "Please indicate which parent should be waived from the application",
+            "optionsLabelPosition": "right",
+            "inline": false,
+            "tableView": true,
+            "validate": {
+                "required": true,
+                "custom": "let validationErrorMessage;\n\n// Get allowed values from injected property 'parents'.\nconst allowedValues = !!data.parents\n    ? data.parents.map((parent) => parent.id)\n    : [];\n// Validate and populate the validation error message.\nif (allowedValues.length < 2) {\n  validationErrorMessage = \"Invalid number of parents in the application.\";\n} else if (!allowedValues.includes(input)) {\n  validationErrorMessage = \"Invalid value for the selected parent.\";\n}\n\n// When validation error message is missing it is considered as successful validation.\nvalid = validationErrorMessage || true;\n  "
```
andrewsignori-aot: Minor, the validation can precisely validate what is required.
```
if (allowedValues.length !== 2)
```

andrewsignori-aot (approved):
Thanks for making the changes. The validation in the form.io looks great,
Looks good 👍

Issue #2429, opened 2023-10-17
SIMS Maintenance Page 


michesmith:
**User Story**
As a SIMS team, we need a maintenance page that we can point to in case of outages or other upgrades. 

**Acceptance Criteria**
- [ ] Create a maintenance page that can be used 
- [ ] Configuration to enable shared maintenance page for each portal 
    - Ministry
    - Student
    - Supporting User
    - Institution
    - External
- [ ] use [figma mocks](https://www.figma.com/design/9Om2tzLAVEoMQrjmrH4fhu/MOD-SIMS?node-id=0-1&p=f) as applicable 
- [ ] Static Content 
> The Student Information Management System is currently unavailable. We apologize for the inconvenience and appreciate your patience.

**Technical**
- [ ] minimum - configuration on API to display to associated user
    - [ ] when maintenance enabled, API to block other routes

michesmith:
The SABC side can post on the generic website as a workaround.

AnnaPBashkatova:
@CarlyCotton @Joshua-Lakusta Can we close this ticket as we have Banners for this purpose?

CarlyCotton:
@AnnaPBashkatova 
I think of this as a separate feature. The banners are only once a student (or institution user) log in. If SIMS is down we would need to enable a maintenance/down page as they wouldn't be able to log in to see the banners.

FYI @Joshua-Lakusta

PR #5435, opened 2025-12-01
#2429 - Update api-deploy to use maintenance mode values as string

tiago-graf:
Fixes the current deployment issue:

`error: failed to read input object (not a Template?): unable to decode "openshift/api-deploy.yml": json: cannot unmarshal bool into Go struct field Parameter.parameters.value of type string`

likely caused by a default value not being set as string


andrewsignori-aot (approved):
👍

PR #5437, opened 2025-12-02
#2429 - Update health controller to allow during maintenance mode

tiago-graf:
Allows /health endpoint to be hit while in maintenance mode is enabled by adding the custom decorator.


andrewsignori-aot (approved):
👍

PR #5443, opened 2025-12-03
#376 - Upgrade web and api dependency packages

tiago-graf:
Update packages suggested by dependabot including

- Npm packages for both web and backend applications
- Node docker images


weskubo-cgi (approved):
Looks good. I ran some containers to test the node upgrade and also did a smoke test of web, api and queues dashboard.

sources/packages/backend/package.json
andrewsignori-aot: I am really struggling to see a benefit in upgrading to these specific versions instead of the most updated ones.
Going against what we talked about in daily I believe that we should go fr the most updated versions.

Image


Image

Issue #5172, opened 2025-09-11
New out of province institution category 


Joshua-Lakusta:
As a ministry user i want there two be two catergories of oop instituions

**Acceptance Criteria**
- [ ] Add a new institution type of Out of Province Private cloned to be the same as the current out of province
- [ ] Rename current option for Out of Province to Out of Province Public
- [ ] All institutions that are currently the category of out of province will stay as Out of Province Public

## Additional Context
<img width="1860" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMXRpQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1f49b28eb75936dd358df4553d7f497bf2abc442/image.png" alt="image.png" />

Manual clean up will be required by ministry users to change those that should not be Out of Province Public to the new option.

PR #5448, opened 2025-12-03
#5172 - New "Out of Province Private" "Out of Province Public" institution types

tiago-graf:
New DB migration to insert new intitution type "Out Of Province Public" and update old institutions to the new type.

Add supporting value in enum InstitutionType

Add sorting when fetching types (not in scope but nice to have since it bundles types together)

image


sources/packages/backend/apps/db-migrations/src/sql/InstitutionType/Add-institution-type-out-of-province.sql
andrewsignori-aot: Minor, comments must have a period.

Issue #4968, opened 2025-07-22
Institution Program Restrictions: Suspension  (UI - add and display restriction)


astridSABC:
**As a** *Ministry User* **I want** *to be able to add and resolve Suspension Restrictions to Institution Program* **So That** *I can temporarily suspend a program*. 
Use case:  PTIB suspends a program but not a Location or an Institution. In this case Ministry does not want to allow any new offerings. No new applications will be allowed. Already existing applications will require Ministry review. If money were disbursed, no action. 

## Scope of this ticket: 

UI to add the restriction, UI for displaying restriction in Restrictions table

## Things to consider:
We will be building 
- [program+location] restrictions known as "Program restriction" AND
- restrictions that should be applied to location known as "Location restriction".
Make sure we can accommodate both scenarios in the near future. 

### **UI  to apply [ program + location] restriction**: 
- [ ] Create provincial [program+location] restriction SUS that Can be applied to a specific program within One Location or to the specific program across ALL Locations 

- [ ] When Ministry clicks  "Add restriction" Button on Institution _Restrictions_ Page, display the pop-up screen. This form will be dynamic in the future based on Reason, allowing to restrict Location and Institution.

<img width="804" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNUkyQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cf279b7f5503110bbdef60743fbdfe1f98b05079/image.png" alt="image.png" />

> **_Reason:_** SUS - program suspended
**_Program_** - allow to search for a program within selected location. Allow to choose one program. Only display Active Programs.
 **_Location_** - drop-down list with all locations within an Institution. Allow to choose one to all locations.  
**_Notes_** 
- [ ] All fields are mandatory fields. If User clicks "Add restriction" and the data in one of the mandatory fields is missing, display a standard warning message. 
- [ ] Once user clicks "Add restriction", display green Toast message
_Restriction was successfully added_
- [ ] Save Notes into Notes section
- [ ] Keycloak: institution-add-restriction should be available to ServiceDelivery-Limited and ServiceDelivery-Full groups (i believe, group already exists, not DEV task)

### Display restriction on Ministry portal
- [ ] Display data about added restriction in the Restriction Table. Add Location and Program Columns, remove Category.  Add View Button. Allow to sort on Location and Program
- [ ] Reason:
[...]

PR #5450, opened 2025-12-04
#4968 - Institution Program Restrictions: Suspension (UI - add and display restriction)

andrewsignori-aot:
## Restrictions Summary

- Added `Location` and `Programs` columns to the table and enabled sorting for them.
- Sorting on column `Status` was kept in order to allow it to display all `Active` items at the top, following the same behavior that was in place and the same as the students' restrictions.

image

_Note:_ resolving restrictions is not part of this ticket. The value added in the above table is a result of the DB manipulation with the sole purpose of taking the screenshot.

image

_Note:_ the only AC related to viewing a restriction is the one that mentions "Add View Button". No work was actually done on this modal, and it is here for the sake of demonstration that it is able to display the data.

## Other Changes

- Included an extra E2E for institution restriction created as [suggested during previous PR](https://github.com/bcgov/SIMS/pull/5430#discussion_r2586482905).
- Removed the hardcoded `Designation` institution type present on the API and student "add restriction".
  - Minor refactor in the student restrictions creation modal to remove errors and adjust the `processing` status (`processing` was added to the Vue template but not present at the `setup`).

## E2E Tests

RestrictionAESTController(e2e)-getReasonsOptionsList.
- √ Should throw a bad request exception when requesting federal restrictions reasons.
- √ Should throw a bad request exception when category is not provided.
- Should get reasons options list for a restriction type when the request is valid.                                                                                                  
  - √ Should get Provincial restrictions reasons list filtered by specific category when Provincial restrictions are requested for a category.                      
  - √ Should get Institution restrictions reasons list filtered by specific category when Institution restrictions are requested for a category.

RestrictionAESTController(e2e)-addInstitutionRestriction.
- √ Should throw an unprocessable entity exception when trying to add a provincial restriction.

## Outside scope but open for discussion

- The method `getAllRestrictionCategories` was not refactored beyond the minimal required to apply the accurate filter. The result is misleading because the distinct will return the first item of a distinct group to allow an ID to be returned, but the returned ID has no purpose, and returning it seems to be a mistake.
- Proposed refactors:
  - Stop returning the ID;
  - Replace the `distinct on` with a regular `distinct`.
  - Pass the restriction type as a parameter.
  - Adjust the API/Web DTOs.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/restriction/models/restriction.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/restriction/models/restriction.dto.ts
@@ -149,11 +149,6 @@ export class StudentRestrictionAPIOutDTO {
   type: RestrictionNotificationType;
 }
 
-export class RestrictionCategoryParamAPIInDTO {
-  @MaxLength(RESTRICTION_CATEGORY_MAX_LENGTH)
-  restrictionCategory: string;
-}
-
 /**
  * Query string options for filtering the restriction reasons.
  */


sources/packages/backend/apps/api/src/route-controllers/restriction/restriction.aest.controller.ts
```diff
@@ -278,13 +279,13 @@ export class RestrictionAESTController extends BaseController {
 
   /**
    * Get restrictions for an institution.
-   * @param institutionId id of the institution to retrieve its restrictions.
-   * @returns Institution Restrictions.
+   * @param institutionId ID of the institution to retrieve its restrictions.
+   * @returns Institution restrictions.
    */
   @Get("institution/:institutionId")
```
tiago-graf: shouldn't this route be "institution/:institutionId/restrictions" ?

sources/packages/backend/apps/api/src/route-controllers/restriction/restriction.aest.controller.ts
```diff
@@ -278,13 +279,13 @@ export class RestrictionAESTController extends BaseController {
...
   async getInstitutionRestrictions(
     @Param("institutionId", ParseIntPipe) institutionId: number,
-  ): Promise<RestrictionSummaryAPIOutDTO[]> {
+  ): Promise<InstitutionRestrictionSummaryAPIOutDTO[]> {
```
dheepak-aot: Can we add wrapping the array of restrictions within an Object as an item to future refactor list?

sources/packages/backend/apps/api/src/services/restriction/institution-restriction.service.ts
```diff
@@ -276,7 +282,7 @@ export class InstitutionRestrictionService extends RecordDataModelService<Instit
     }
     if (!restrictionExists) {
       throw new CustomNamedError(
-        `Restriction ID ${restrictionId} not found.`,
+        `Institution restriction ID ${restrictionId} not found.`,
```
dheepak-aot: 👍

sources/packages/backend/apps/api/src/services/restriction/restriction.service.ts
```diff
@@ -16,14 +16,16 @@ export class RestrictionService extends RecordDataModelService<Restriction> {
...
+   * @returns Restriction categories.
    */
   async getAllRestrictionCategories(): Promise<Restriction[]> {
     return this.repo
       .createQueryBuilder("restriction")
       .select(["restriction.id", "restriction.restrictionCategory"])
-      .where("restriction.restrictionCategory NOT IN ('Federal','Designation')")
+      .where("restriction.restrictionType = :restrictionType", {
+        restrictionType: RestrictionType.Provincial,
+      })
```
dheepak-aot: Thanks for the update.

sources/packages/backend/apps/api/src/route-controllers/restriction/models/restriction.dto.ts
```diff
@@ -49,6 +49,17 @@ export class RestrictionSummaryAPIOutDTO extends RestrictionBaseAPIOutDTO {
   deletedAt?: Date;
 }
 
+/**
```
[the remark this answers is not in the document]
andrewsignori-aot: I was confused because the comment was added to `RestrictionReasonsOptionsAPIInDTO`, and I missed the latest comment. I will remove the `RestrictionCategoryParamAPIInDTO`.

sources/packages/backend/apps/api/src/route-controllers/restriction/restriction.aest.controller.ts
andrewsignori-aot: I see the point, but right now it is keeping the behavior from before, and aligned with the student.
Unless there is a strong reason to be a blocker, I would keep it as it is.

dheepak-aot (approved):
Good work and thanks for making the changes and the additional refactors. Looks good 👍

Issue #5242, opened 2025-10-06
Upgrade web server


ninosamson:
**User story:** As a SIMS team, we need to upgrade our packages and frameworks. 

**Acceptance Criteria:**
- [ ] Upgrade nginx - similar to NodeJS, move away from RedHat images 
- [ ] Upgrade local (docker)

PR #5468, opened 2025-12-05
#5242 - Update nginx permissions

tiago-graf:
For OpenShift, we need to use chmod instead of chown because OpenShift runs containers with arbitrary UIDs.

Updated both Dockerfiles to use chmod with group permissions instead of chown.

### DIFF
--- a/sources/packages/web/Dockerfile
+++ b/sources/packages/web/Dockerfile
@@ -33,11 +33,12 @@ COPY --from=builder /app/default.conf /etc/nginx/conf.d/default.conf
 # Copying the main configuration file.
 COPY --from=builder /app/nginx/nginx.conf /etc/nginx/nginx.conf
 
-# Set proper permissions for nginx user
-RUN chown nginx /usr/share/nginx/html -R \
-    && chown nginx /var/cache/nginx/ -R \
-    && chown nginx /var/run -R \
-    && chown nginx /etc/nginx/ -R
+# Set proper permissions for OpenShift (arbitrary UIDs with root group)
+RUN chmod -R g+rwx /usr/share/nginx/html \
+    && chmod -R g+rwx /var/cache/nginx \
+    && chmod -R g+rwx /var/run \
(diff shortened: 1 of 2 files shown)


sources/packages/web/Dockerfile
andrewsignori-aot: For PROD, we try to fix up to the patch version, which would be `1.29.3-alpine3.22`.

sources/packages/web/Dockerfile
```diff
@@ -33,11 +33,12 @@ COPY --from=builder /app/default.conf /etc/nginx/conf.d/default.conf
 # Copying the main configuration file.
 COPY --from=builder /app/nginx/nginx.conf /etc/nginx/nginx.conf
 
-# Set proper permissions for nginx user
-RUN chown nginx /usr/share/nginx/html -R \
-    && chown nginx /var/cache/nginx/ -R \
-    && chown nginx /var/run -R \
-    && chown nginx /etc/nginx/ -R
+# Set proper permissions for OpenShift (arbitrary UIDs with root group)
+RUN chmod -R g+rwx /usr/share/nginx/html \
```
andrewsignori-aot: The following permission would be recommended following the other PROD files.

```
RUN mkdir -p /var/log/nginx && \
    chgrp -R 0 /usr/share/nginx/html /var/cache/nginx /var/run /etc/nginx /var/log/nginx && \
    chmod -R g=u /usr/share/nginx/html /var/cache/nginx /var/run /etc/nginx /var/log/nginx
````

Further explanation.
> OpenShift runs containers with an arbitrary UID that is added to the root group (GID 0). To let nginx work without a fixed user:

> mkdir -p /var/log/nginx: Ensure log directory exists before setting permissions.
chgrp -R 0 …: Make nginx’s writable paths belong to GID 0 so the arbitrary UID (in root group) can use group permissions.
chmod -R g=u …: Mirror owner permissions to the group, granting the arbitrary UID the same access (read/write/execute). For directories, this enables creating files, writing cache, PID files, and logs.

Issue #3976, opened 2024-11-20
Add a new provincially added death restriction 


Joshua-Lakusta:
As a ministry of finance user I need there to be a provincial version of the death restriction so that when there is no federal loan and only BC loan is left with RSBC I I am still able to add the death restriction where the feds ill not be notifying us.

**Acceptance Criteria**
- [ ] Add new restriction code PTD - Account closed by province
- [ ] Add the new restriction to the expandable restrictions list on the student view as PTD with the following content:

> A restriction has been placed on your file by StudentAid BC which disqualifies you from receiving further funding. For further information please contact StudentAid BC at 1-800-561-1818 (within North America) 1-778-309-4621 (outside North America).

- [ ] behavior of the restriction would be the same as the current federal death restriction (blocks everything including new apps)
- [ ] Category of the restriction will be other

**Additional Context**
TD is the federal restriction

PR #5489, opened 2025-12-09
#3976 - Add Account Closed (PTD) restriction

tiago-graf:
## Add PTD (account closure) restriction support 

Added database migration to insert PTD (Provincial Training Difference) restriction into sims.restriction table
Updated StudentAccountActivity Form to add PTD restriction to the restriction list

image

### DIFF
--- a/sources/packages/web/src/components/students/RestrictionBanner.vue
+++ b/sources/packages/web/src/components/students/RestrictionBanner.vue
@@ -31,7 +31,6 @@ import { useRouter } from "vue-router";
 import { useStudentStore } from "@/composables";
 import { StudentRoutesConst } from "@/constants/routes/RouteConstants";
 import { BannerTypes } from "@/types/contracts/Banner";
-
 export default defineComponent({
   setup() {
     const router = useRouter();


sources/packages/forms/src/form-definitions/studentaccountactivity.json
```diff
@@ -530,6 +530,28 @@
...
+          "collapsible": true,
+          "key": "codePTDTitle",
+          "type": "panel",
+          "label": "Panel",
+          "collapsed": true,
+          "input": false,
+          "tableView": false,
+          "components": [
+            {
+              "html": "<p>A restriction has been placed on your file by StudentAid BC which disqualifies you from receiving further funding. For further information please contact StudentAid BC at 1-800-561-1818 (within North America) or 1-778-309-4621 (outside North America).</p>",
```
andrewsignori-aot: The content has an extra "or" in the sentence; the AC should be either changed or the biz should be asked to adjust the AC.
If some action has already been taken to have it adjusted, please disregard this comment.

sources/packages/forms/src/form-definitions/studentaccountactivity.json
```diff
@@ -530,6 +530,28 @@
             }
           ]
         },
+        {
+          "title": "Code PTD - Account Closed by Province",
```
andrewsignori-aot: It is hard to define the pattern for the title of these contents.
There is an AC that states the title as `PTD - Account closed by province` but I am not sure if the intention was also to be the content title.
I would recommend either following the AC or asking biz.

andrewsignori-aot (approved):
I will not hold the approval on this due to the content. If there is no biz confirmation till EOD, I believe that we should move ahead.

sources/packages/forms/src/form-definitions/studentaccountactivity.json
ninosamson: I adjusted the AC for the "or".

sources/packages/forms/src/form-definitions/studentaccountactivity.json
ninosamson: Updated AC to be explicit about title in form definition.

Issue #4815, opened 2025-06-12
Rate Limit SIMS API


ninosamson:
- [ ] Add a global rate limit using nestJS framework for sims-api
- [ ] Devs to discuss appropriate limits and throttling during implementation

https://docs.nestjs.com/security/rate-limiting

PR #5498, opened 2025-12-11
#4815 - Add Rate Limit to API Endpoints

tiago-graf:
Installs npm package nestjs/throttler
Enables rate limit globally
Skips (No limit) /health endpoint
Adds e2e tests to verify limits

### DIFF
--- a/sources/packages/backend/apps/api/src/auth/_tests_/e2e/guards.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/auth/_tests_/e2e/guards.e2e-spec.ts
@@ -477,7 +477,7 @@ describe("Guards and Decorators - Authentication, Maintenance Mode, Throttler (e
 
   describe("Throttler Guard", () => {
     it("Should allow requests within the rate limit.", async () => {
-      const endpoint = "/auth-test/authenticated-student";
+      const endpoint = "/auth-test/throttle-test/success";
 
       // Act and Assert - Make requests up to the limit, all should pass
       for (let i = 0; i < throttleLimit - 1; i++) {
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/api/src/auth/_tests_/e2e/guards.e2e-spec.ts
```diff
@@ -473,6 +475,47 @@ describe("Guards and Decorators - Authentication and Maintenance Mode (e2e)", ()
...
+
+      // Assert - Next request should be throttled
+      await request(app.getHttpServer())
+        .get(endpoint)
+        .expect(HttpStatus.TOO_MANY_REQUESTS)
+        .expect((res) => {
+          expect(res.body.statusCode).toBe(HttpStatus.TOO_MANY_REQUESTS);
+          expect(res.body.message).toContain("ThrottlerException");
+        });
+    });
```
dheepak-aot: I am trying to understand how consistently this will pass. May be for this E2E suite we can try different limits (say 10 requests per second instead of 0.1 seconds) ?

tiago-graf: the exhaust test should pass pretty consistently, the multiple request in a for loop happen really fast. the potential problem is other tests making too many requests at a very fast pace and because of parallel threading it could cause a test that is expecting a success but it gets a 429 (because other tests already exausted the same endpoint) you can see copilot comment about that, the easy way to prevent that is to add a delay, but that makes our tests run a tidy bit slower

Issue #5438, opened 2025-12-02
[DEV CONNECT] Memory usage for db during backup


ninosamson:
During DB backups, memory usage increases beyond threshold in `statefulset-simsdb-ha-qmfg`, but seems to resolve itself. This is causing unnecessary alerts and may cause true incidents to be missed. 

- [ ] investigate memory quota usage during backup and increase quota as needed

> SIMS - PROD Container Memory 90% Limit is Triggered in statefulset-simsdb-ha-qmfg

CarlyCotton:
@tiago-graf  If there are any release activities related to this please post in the DEV channel

PR #5512, opened 2025-12-12
#5438 - Update pgBackRestConfig container to use less resources

tiago-graf:
Created separate resource configurations for pgBackRest sidecars: 

pgbackrest (256Mi/512Mi) handles backup operations
pgbackrestConfig (64Mi/128Mi) only watches configuration changes. 

This optimizes resource allocation by not over-provisioning the config watcher container.


weskubo-cgi (approved):
Good catch!

Issue #4879, opened 2025-07-08
Reverse Scholastic Standing Reporting: Tracker Card 


AnnaPBashkatova:
When Ministry User reverses Scholastic Standing Event reported for Student Application, display proper Tracker Cards on application.

## *Acceptance Criteria:*
All Program Years

- [ ] If at least one active UC, display "Your institution reported an unsuccessful completion of studies for your application" tracker card. 
- [ ] Otherwise, hide UC tracker cards. 
- [ ] Ensure all scholastic standing banners related to assessment will still be displayed if related to the current assessment.

Example of Tracker Card displayed on Student application when Unsuccessful Completion of Studies is reported by Institution:
<img width="1874" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHcyREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7cdc358982b7683880096a03d38cf3633c9770b7/image.png" alt="image.png" />

Q&A:
1. 
 - Create FT student application for 17 week study period
- Confirm COE
- Report 5 weeks of UC as an Institution
- Reverse UC as a Ministry User
- Log back as a Student. UC tracker card should no longer  be there. 
2. 
 - Create another FT student application for 17 week study period
- Confirm COE
- Report 5 weeks of UC as an Institution
- Report another 5 weeks of UC for the same app as an Institution
- Reverse ONE UC as a Ministry User
- Log back as a Student. UC tracker card should be there.

CarlyCotton:
@AnnaPBashkatova  Based on how Reverse Scholastic Standing ticket is functioning is this just now for tracker card and email comms? The other ACs look like they're going to be done as part of the other one.

ninosamson:
Moving back to Soak and kindly requesting that the Tracker Card be split from the Email communication - two separate tickets. Can you tweak the ACs so that our new devs can better understand the ticket @AnnaPBashkatova?

PR #5513, opened 2025-12-13
#4879 - Reverse Scholastic Standing Reporting: Tracker Card

sh16011993:

- Code is adjusted to ensure that the below tracker card is only shown when there is atleast one active unsuccessful completion weeks type of scholastic standing.

image

- Adjusted the existing failing e2e tests.


andrewsignori-aot (approved):
Nice work, looks good 👍 
Please let me know if any investigation was executed related to the query.

weskubo-cgi (approved):
Changes look good. Verified both cards now display as expected.

weskubo-cgi:
One comment for next time, please ensure that your branch name follows existing standards. E.g. `feature/#4879-reverse-scholastic-standing-reporting-tracker-card`.

Issue #5100, opened 2025-08-25
Process negative overawards on disbursement


AnnaPBashkatova:
As a system I need to be able to handle negative overawards at disbursement stage. 

## **Acceptance Criteria**
- [ ] Ensure that negative overawards cannot produce an amount greater then the assessed value of an application.
- [ ] Negative balance overawards should only counteract positive overawards that were previously deduced on an application.
- [ ] Add positive record change when negative overaward is consumed by disbursement.

<img width="1636" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOFAyQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1e893f661ded950923bb3ad783ac292f0b722164/image.png" alt="image.png" />

QA:
Full-Time
Part One
1. Log in as a Ministry User and add $500 for CSLF and another $300 from BCSL to Student account via Balances screen. 
2. Complete a FT application 
3. Confirm COE and run FT eCert and push all the way to disbursement so that we can see the overawards deducted from the disbursement 
4. Compare CSLF and BCSL numbers on assessment with numbers on eCert. Confirm that system consumed the overawards from step1 to reduce assessment numbers on disbursement.

Part 2a
1. Add a negative overaward balance of $500 CSLF and $300 BCSL
2. Manually reassess the application and run FT eCert. 
3. Compare CSLF and BCSL numbers on assessment with numbers on eCert. Confirm that system consumed the overawards from step1 to increase the assessment numbers on disbursement.

Part 2b
1. Repeat Part 1
2. Add a negative overaward balance of $1000 CSLF and $600 BCSL
3. Manually reassess the application and run FT eCert
4. Compare CSLF and BCSL numbers on assessment with numbers on eCert. Confirm that system consumed the overawards from step1 to increase the assessment numbers on disbursement. The increase should only be up to the assessed amount of the application and the amount previously reduced by the overawards.

AnnaPBashkatova:
Going to look at how this scenario is currently handled in SFAS.

andrewsignori-aot:
@Joshua-Lakusta @CarlyCotton @ninosamson I will be adding a new origin typo to the overawards table as shown below.

[...]

PR #5557, opened 2025-12-22
#5100 - Process negative overawards on disbursement - Avoid $0 process

andrewsignori-aot:
## Issue Description

When the student has or had some overaward balance, even if its SUM is $0 now, future e-Cert disbursements will create a $0 entry in the overawards table, as shown below.

## Root Cause

The balance SUM returned from DB is not properly converted to a number, and the `if` will fail to recognize the '0.00' as a false value (as they should).

image

The error has been happening for a while, as shown in DEV records.

_Note:_ With the fix, a '0.00' balance will no longer be returned.

## Fix

1 - The string returned from DB is now properly converted to a number (**required fix**).
2 - The balance does not need to be returned if it is not a positive or negative number. This can be considered optional, but I did not find a strong reason (as the method is consumed today) to not have it changed. If required in the future to return $0 balances, an option can be added to the method to conditionally include the `having` SQL clause.
3 - The DTO is now correctly returning a number, and E2E tests are correctly validating a number. Previously, DTO was mapped as a number but returned as a string that was validated as a string by E2E tests.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/overaward/models/overaward.dto.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/overaward/models/overaward.dto.ts
@@ -17,7 +17,7 @@ import {
 } from "class-validator";
 
 export class OverawardBalanceAPIOutDTO {
-  overawardBalanceValues: Record<string, number | undefined>;
+  overawardBalanceValues: Record<string, number>;
 }
 
 export class StudentsOverawardAPIOutDTO {
--- a/sources/packages/web/src/services/http/dto/Overaward.dto.ts
+++ b/sources/packages/web/src/services/http/dto/Overaward.dto.ts
@@ -4,7 +4,7 @@ import {
 } from "@/types";
 
 export interface OverawardBalanceAPIOutDTO {
-  overawardBalanceValues: Record<string, number | undefined>;
+  overawardBalanceValues: Record<string, number>;
 }
 
 export interface OverawardAPIOutDTO {


sources/packages/backend/libs/services/src/disbursement-overaward/disbursement-overaward.service.ts
```diff
@@ -69,13 +70,16 @@ export class DisbursementOverawardService {
       })
       .groupBy("student.id")
       .addGroupBy("disbursementOveraward.disbursementValueCode")
-      .getRawMany<{ studentId: number; valueCode: string; total: number }>();
+      .having("SUM(disbursementOveraward.overawardValue) <> 0")
+      // The total is returned from DB as string, needs to be converted to a number,
+      // since overawardValue is defined as numeric in the DB.
+      .getRawMany<{ studentId: number; valueCode: string; total: string }>();
```
dheepak-aot: 👍

sources/packages/backend/apps/api/src/route-controllers/overaward/models/overaward.dto.ts
```diff
@@ -17,7 +17,7 @@ import {
 } from "class-validator";
 
 export class OverawardBalanceAPIOutDTO {
-  overawardBalanceValues: Record<string, number>;
+  overawardBalanceValues: Record<string, number | undefined>;
```
dheepak-aot: When the key is present, the value cannot be undefined. The existing type is accurate.

sources/packages/web/src/services/http/dto/Overaward.dto.ts
```diff
@@ -4,7 +4,7 @@ import {
 } from "@/types";
 
 export interface OverawardBalanceAPIOutDTO {
-  overawardBalanceValues: Record<string, number>;
+  overawardBalanceValues: Record<string, number | undefined>;
```
dheepak-aot: https://github.com/bcgov/SIMS/pull/5557/changes#r2641593099

sources/packages/backend/apps/api/src/route-controllers/overaward/models/overaward.dto.ts
andrewsignori-aot: Reverted.

dheepak-aot (approved):
Great catch and fix. Looks good 👍

Issue #5418, opened 2025-11-26
Appeal - Step-Parent Waiver Assessment


CarlyCotton:
**User Story**
As a student, if I have a successful Step-Parent Waiver appeal my assessment needs to reflect that parent no longer counting towards family size and family income.

**Acceptance Criteria**
Update FT 25-26 Assessment:
- [ ] add conditions to check if step-parent appeal exists and exclude that parent's income values from the income calculation
- [ ] add conditions to check if step-parent appeal exists and exclude that parent from the family size count
- [ ] add conditions to check if step-parent appeal exists and exclude that parent's dependants when considering family size

PR #5580, opened 2026-01-06
#5418 - Appeal Step-parent waiver Assessment

CarlyCotton:
### BPMN Changes (25-26 & 26-27)
Added a new assessment flow to accept the approved step parent appeal waivers:
image
In **Waived Parent** we determine if "parent1" or "parent2" is going to be used based on comparing the supporting user IDs:
image
After that we assign the correct parent's variable to each `appealDataParent` value and following the same calculation for "Total Parent Eligible Dependants", "Parental Assets", "Parental Contribution" based on the remaining parent. Family size is also no longer including a second parent.
image
This has allowed "Calculate Family Income" to remain the same calculation with or without an appeal and still supports the original income priority (cra > parent report > student report)

---

### E2E Tests 
- Nothing should change if appeal data doesn't exist 
- Appeal applies: reduce family total income & net income, family size, parent total dependants, parent contribution dependants.

---

### BPMN Cleanup
Total Contribution Dependants was not correctly calculating the age of dependants in the 1 Parent path or the Appeal path. It was using today's date to calculate age instead of the offering start date. This has been corrected on both 25-26 and 26-27 BPMNs.


andrewsignori-aot (approved):
Great work, approving with a suggestion for now or future considerations. Looks good 👍

PR #5583, opened 2026-01-06
#5582 - Supporting Users Update Search Date of Birth Format

tiago-graf:
- Updates the supporting users search page, adds the standard ISO format for the date of birth search field.


weskubo-cgi (approved):
Looks great. Tested explicit format compared to existing version with inherited format. Can we add a screenshot to the description?

Issue #3553, opened 2024-07-23
Revise Impact of BC Lifetime Max Restriction to only block Loan


HRAGANBC:
**User Story**
As a student, I want a BC Lifetime Max to result in ineligibility for particular funding types

**Acceptance Criteria**
- [ ] Revise the impact of the BCLM restriction to block _only the BC Loan they are still eligible for BC grants. 
- [ ] Need to create a new outcome for restrictions to accommodate this as the current block BC funding is used for Bankruptcy correctly. 
- [ ] Would be ok to introduce new action or split the previous action into grant and loan separately 
- [ ] If the splitting path is chosen all restrictions other then BCLM that use (Stop full time bc funding) should have the block both created options applied.

**Technical ACs**
- [ ] Split 'Stop Full Time BC Funding' into 'Stop Full-time BC Grant Funding' and 'Stop Full-time BC Loan Funding'. (Do not need to split PT)
- [ ] Review possibility of using effective conditions

**Current Behaviour**
- BCLM currently results in blockage of all BC Funding Types associated with Full-Time. 

<img width="676" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOGNBQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3202bba05a6248784a85b7e91fab64a02928934b/image.png" alt="image.png" />

Joshua-Lakusta:
@Joshua-Lakusta review function and is critical

PR #5592, opened 2026-01-08
#3553 - Revise Impact of BC Lifetime Max Restriction to only block Loan

andrewsignori-aot:
- Split the `StopFullTimeBCFunding` into `StopFullTimeBCLoan` and `StopFullTimeBCGrants` to allow restriction to be configured with a more granular option.
- Renamed the `StopPartTimeBCFunding` to `StopPartTimeBCGrants` since part-time has no loan.
- Update restrictions `B2D`, `B3D`, `B2`, `Z1` to use the new action types, keeping their existing behavior.
- Update restrictions `BCLM` to block only BC Loan while still allowing grants disbursements.
- Restrictions bypass conditions to list restrictions were adapted to include the new restriction action types.

## E2E Tests

- Existing BCLM tests adapted to validate if grants are still disbursed.
- Created a new test to ensure different restriction IDs are correctly associated with blocked awards.
- Some E2E tests were selecting an institution restriction when the context required a student restriction due to the lack of some where conditions. Restriction provincial type was used for those scanrios.

## Migration Rollback

image

### DIFF
--- a/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/e-cert-steps-utils.ts
+++ b/sources/packages/backend/libs/integrations/src/services/disbursement-schedule/e-cert-processing-steps/e-cert-steps-utils.ts
@@ -16,7 +16,7 @@ import { isRestrictionActionEffective } from "./e-cert-steps-restriction-utils";
 /**
  * Map of restriction action types associated with the offering intensity.
  */
-const RESTRICTION_ACTION_TYPE_INTENSITY_MAP = new Map([
+const STOP_FUNDING_ACTION_INTENSITY_MAP = new Map([
   [
     OfferingIntensity.fullTime,
     [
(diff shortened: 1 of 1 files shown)


sources/packages/backend/apps/queue-consumers/src/processors/schedulers/esdc-integration/ecert-integration/_tests_/ecert-part-time-process-integration.scheduler.e2e-spec.ts
```diff
@@ -1414,9 +1416,12 @@ describe(
...
+        select: { id: true },
         where: {
+          restrictionType: RestrictionType.Provincial,
+          actionEffectiveConditions: IsNull(),
           actionType: ArrayContains([
-            RestrictionActionType.StopPartTimeBCFunding,
+            RestrictionActionType.StopPartTimeBCGrants,
           ]),
         },
       });
```
dheepak-aot: As we are seeing this very often, it seems like a good candidate to be in factory.

dheepak-aot (approved):
Great work and improvement to the granularity of the funding type restrictions. 👍

Issue #5144, opened 2025-09-05
Ministry: display Badge if Institution is restricted


AnnaPBashkatova:
As a Ministry User, I want to see a badge on Institution Page when Institution has an Active restriction on their account.

## **Acceptance Criteria**

- [ ] If institution has an active restriction, display the "Active restrictions" badge besides the "Designated" badge on Ministry site.
<img width="1613" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkZVQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0d007c50b3813997c2eac820f8d3b6652de1c75e/image.png" alt="image.png" />

Additional Context:
- Received as a part of feedback from SD while demoing feature, they currently have the Banner in SFAS, and find it very handy, especially for Tier1

CarlyCotton:
Bumping to 3 pts from 2 pts to support small investigation for a centralization of some API calls to allow them to be shared between Ministry/Institution in a way that will work best.

PR #5625, opened 2026-01-15
#5144 - Ministry: Display badge is institution is restricted

sh16011993:
### **As a part of this PR, the following were completed:**

- Used the existing `StudentRestrictionChip` to display the "Active restrictions" badge besides the "Designated" badge on Ministry site If institution has an active restriction.

**Screenshot:**

image

### DIFF
--- a/sources/packages/web/src/views/aest/institution/InstitutionDetails.vue
+++ b/sources/packages/web/src/views/aest/institution/InstitutionDetails.vue
@@ -11,6 +11,14 @@
             class="ml-4 mt-1"
             :status="institutionBasicDetail.designationStatus"
           />
+          <student-restriction-chip
+            class="ml-4 mt-1"
+            :status="
+              institutionBasicDetail.hasRestrictions
+                ? StudentRestrictionStatus.Restriction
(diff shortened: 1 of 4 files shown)


sources/packages/web/src/views/aest/institution/InstitutionDetails.vue
```diff
@@ -39,9 +47,11 @@ import { InstitutionService } from "@/services/InstitutionService";
 import { InstitutionBasicAPIOutDTO } from "@/services/http/dto";
 import { AESTRoutesConst } from "@/constants/routes/RouteConstants";
 import StatusChipDesignationAgreement from "@/components/generic/StatusChipDesignationAgreement.vue";
+import { StudentRestrictionStatus } from "@/types";
```
weskubo-cgi: **todo:** `StudentRestrictionStatus` should have a generic name now that it's used by Institutions. Note, we already have `RestrictionStatus` so we might need to get creative with the naming. `HasRestrictionType`?

weskubo-cgi (changes requested):

Issue #3625, opened 2024-08-12
Institution - Create Program Link on Banner


CarlyCotton:
**User Story**
As an institution user, I need to be taken to a fresh program page to create a new program when clicking on the link in the banner displayed on the program. Currently it's just changing the page title, leaving the name and description to be edited but all other fields are greyed out _(Fig2)_.

**Acceptance Criteria**
- [ ] When clicking on the 'Create program' button _(Fig1)_, 
    - [ ] Go to 'Create Program' page, do not show banner, and allow all fields to be editable

**Figure 1**
<img width="1245" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWxOQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6c894b821d9aeab8378c4dba4560215aa871a084/image.png" alt="image.png" />

**Figure 2**
<img width="1582" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1pOQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4c96ecbb622932ccf3126fa3efcbc84b77c6749f/image.png" alt="image.png" />

Joshua-Lakusta:
Bumped for other priorities

PR #5661, opened 2026-01-22
#3625 - Create Program not loading correctly

weskubo-cgi:
### Overview
- **[Create program]** wasn't loading correctly when navigated to from Edit Program since it's the same component
- Switched from `onMounted()` to `watchEffect()` which nicely handles initial component creation as well as subsequent navigation without program id

### Screenshots
image

image


andrewsignori-aot (approved):
Looks good 👍

Issue #5645, opened 2026-01-19
Institution types remodel


Joshua-Lakusta:
As a ministry user i want to re organize the institution types to allow for better data collection and more specific targeting of different institution types to align with policy.

**Acceptance Criteria**
### Basic information

- [ ] Change the basic institution type drown down to have only 3 options BC, Canadian out of province, and International
- [ ] Add additional questions to the institution profile subject to each institution type

- [ ] Questions for BC

> 
> Is the organization public or private?
> o Public
> o Private
> 
> Is the organization for profit or not for profit?
> o For Profit
> o Not for Profit
> 
> Is the organization medical in nature?
> o Yes
> o No

- [ ] Questions for Canadian OOP

> Is the organization public or private?
> o Public
> o Private
> 
> Is the organization for profit or not for profit?
> o For Profit
> o Not for Profit
> 
> Is the organization medical in nature?
> o Yes
> o No
> 
> In which Canadian province is the organization located? Drop down list containing following values.
> Alberta
> British Columbia
> Manitoba
> New Brunswick
> Newfoundland and Labrador
> Nova Scotia
> Ontario
> Prince Edward Island
> Quebec
> Saskatchewan
> Territories (Yukon / NWT / Nunavut)

- [ ] Questions for International

> 
> Is the organization public or private?
> o Public
> o Private
> 
> Is the organization for profit or not for profit?
> o For Profit
> o Not for Profit
> 
> Is the organization medical in nature?
> o Yes
> o No
> 
> What country is your institution located in?
> Drop down list with all countries

### Institution profile display
Please move 
### Data Alignment TBD

- [ ] For all current institutions flag no for medical unless they are currently an international medical school
- [ ] For all institutions that are currently flagged as public or private please align those factors with the new attribute
- [ ] For all international schools currently flagged as US please indicate that in the country section
- [ ] Missing data handling **To be discussed**

**Additional Context**
- enter text here
- enter text here

CarlyCotton:
@Joshua-Lakusta 
OOP:
<img width="673" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3lvREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3f185dd87bd9883550300f8cc3682d09630a36ec/image.png" alt="image.png" />
International:
[...]

PR #5691, opened 2026-01-30
#5645 - Institution profile updates - DB Migrations and Lookup data API

dheepak-aot:
# Institution profile updates - DB Migrations and Lookup data API

## DB Migrations
- [x] Migration 1: DB migration to create the system-user if the system-user is not already present in DB. 
**Need for creating this migration:** When insert scripts are executed as part of DB migrations, we require an audit-user as creator of inserted records and the audit-user can only be a system user in this case. But when these scripts execute on a clean DB, say E2E test database then there is no system-user available for the DB migrations because it is created only during nest application startup which happens after DB Migrations are completed. Hence we added a migration to create a system-user if the same does not already exist in the DB.

- [x] Migration 2: DB Migration to create the table `sims.system_lookup_configurations` to configure the lookup data under different lookup categories. This lookup service is a generic solution which is primarily created to act as a store/API to provide values to web clients that are required to load the components.  
- [x] Migration 3: DB Migration to create the STABLE function `sims.is_valid_system_lookup_key` which can be used in a check constraint for columns in tables which is based on the lookup data key. For e.g. sims.institutions has the column country which has country code value from the  `sims.system_lookup_configurations` under lookup category `Country`.
- [x] Migration 4: DB Migration to add new columns to `sims.institutions` and update one or more new columns based on the existing institution types.

image

## Rollback Evidence

image

image

image

image

## System lookup Module and API

- [x] Created service which loads lookup data on module start-up and a controller `SystemLookupConfigurationController` that is accessible to all clients (except external) similar to dynamic form configuration controller.

## API E2E Tests

- [x] Added E2E tests to for the `SystemLookupConfigurationController`.
```log
SystemLookupConfigurationController(e2e)-getSystemLookupEntriesByCategory
    √ Should get the system lookup test entries for the category country when one ore more test entries exist for the same category. (168 ms)                                                     
    √ Should throw bad error when the lookup category is invalid. (47 ms)
```

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/system-lookup-configuration/_tests_/e2e/system-lookup-configuration.controller.getSystemLookupEntriesByCategory.e2e-spec.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/system-lookup-configuration/_tests_/e2e/system-lookup-configuration.controller.getSystemLookupEntriesByCategory.e2e-spec.ts
@@ -15,13 +15,14 @@ import {
 import { SystemLookupCategory, User } from "@sims/sims-db";
 
 import { SystemLookupConfigurationService } from "@sims/services/system-lookup-configuration";
-import { ILike } from "typeorm";
+import { Like } from "typeorm";
 
 describe("SystemLookupConfigurationController(e2e)-getSystemLookupEntriesByCategory", () => {
   let app: INestApplication;
   let db: E2EDataSources;
   let auditUser: User;
   let systemLookupConfigurationService: SystemLookupConfigurationService;
+  const TEST_LOOKUP_KEY_PREFIX = "TEST_";
 
   beforeAll(async () => {
     const { nestApplication, dataSource } = await createTestingAppModule();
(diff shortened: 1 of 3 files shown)


sources/packages/backend/apps/db-migrations/src/sql/Institution/Add-additional-information-cols.sql
```diff
@@ -0,0 +1,95 @@
...
+    modifier = (
+        SELECT
+            id
+        FROM
+            sims.users
+        WHERE
+            -- System user.
+            user_name = '8fb44f70-6ce6-11ed-b307-8743a2da47ef@system'
+    )
+WHERE
```
andrewsignori-aot: I do not understand the benefit of having the check between 1 and 7. IMO, it would be equivalent to check `IF EXISTS`, which we stopped doing because if something is outside the expectation and an error happens, the error should happen.

sources/packages/backend/apps/test-db-seeding/src/clean-db/clean-db.ts
```diff
@@ -12,5 +12,8 @@ export class CleanDatabase {
     await this.dataSource.query(
       "DROP FUNCTION IF EXISTS sims.create_history_entry",
     );
+    await this.dataSource.query(
```
andrewsignori-aot: Minor, I do not recall if there was a reason not call all these DB commands in parallel.
It can be a quick improvement (outside PR scope).

sources/packages/backend/apps/api/src/route-controllers/system-lookup-configuration/_tests_/e2e/system-lookup-configuration.controller.getSystemLookupEntriesByCategory.e2e-spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        auditUser,
+      },
+      {
+        initialValues: {
+          lookupCategory: SystemLookupCategory.Country,
+          lookupKey: "TEST_CN2",
+          lookupValue: "Test Country2",
+        },
+      },
+    );
```
sh16011993: Can we add a 3rd lookup category, anything other than a country (say Province) testing the part that it should not get returned.

dheepak-aot: Adding 3rd lookup category was my first thought, but it requires updating the enum with 3rd category `SystemLookupCategory.Some3rdCategory` for the API to allow it. so i took this approach.

PR #5721, opened 2026-02-03
#5720 - Upgrade Node to 25.6

tiago-graf:
### Summary
Updates Node.js to 25.6.0-alpine3.22 across all backend and frontend services
Adjusts Dependabot Docker configuration for broader backend service scanning

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -131,28 +131,15 @@ updates:
   - package-ecosystem: "docker"
     commit-message:
       prefix: "docker"
-    directory: "/sources/packages/backend"
+    directories:
+      - "/sources/packages/backend/**/*"
+      - "/sources/packages/web/**/*"
     schedule:
       interval: "weekly"
     labels:
       - "Dependencies"
       - "Docker"
-      - "Backend"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -131,12 +131,13 @@ updates:
   - package-ecosystem: "docker"
     commit-message:
       prefix: "docker"
-    directory: "/sources/packages/backend/apps/api"
+    directory: "/sources/packages/backend"
```
andrewsignori-aot: This is nice! Do you mind sharing the documentation that would support this change?
I notice that now it does support the ability to provide `directories`, but I did not find (in a quick search) that it would support the recursive scan from the root directory.

tiago-graf: I assumed it would be recursive but i think your right, from the docs i think i have to change to include **/* in the directory, i'll do a bit more research and change this. its one of the things i can't test without pushing the changes directly

andrewsignori-aot (approved):
Looks good 👍

weskubo-cgi (approved):

Issue #5619, opened 2026-01-13
ece-response - Handle INTP & INTF records


CarlyCotton:
**User Story**
As a system, SIMS needs to process CONR files (ece-response) which include disbursements and Interest free / in-study records from SFAS (legacy) 
Currently we handle disbursements from legacy nicely, unable to find the associated disbursement.
The error created when we receive an INTP record:
> Errors:
Invalid application number at line 11.
Error processing the file IN/CONR-008-AUAA-20260108-235000.txt. Error: The file consists of invalid data and cannot be processed.
File processing aborted.

<img width="1164" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWVYREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c45edb25a932aaac28d89ad7169e8861b5f27d00/image.png" alt="image.png" />

## **Acceptance Criteria**
- [ ] INTP  & INTF Records with no application number will be received in the ECE-Response (legacy records), do not treat as an error and continue processing the file.
- [ ] List as a warning line item similar to legacy disbursements not found to allow for the summary report process to capture the details. Example of current SFAS record skipped:
    > Disbursement schedule not found for disbursement value ID: 1223811987, record at line 6 skipped. 

---
**Additional Context**
Since Jan 1, SFAS has received INTF/INTP records over 150 since Jan 1. This can be expected to continue as long as SFAS is processing IER1-11/ECE records.

CarlyCotton:
Confirmed with Harry - CON/CONR includes two types of records for the in-study reporting: INTP & INTF. Updated ACs to include both as "skippable" records.

PR #5737, opened 2026-02-05
#5619 - ECE Response Handle: INTP INTF Records

sh16011993:
### **As a part of this PR, the following were completed:**

- INTP & INTF Records with no application number will be received in the ECE-Response file, they are treated as a warning item and the file is processed as usual. The item is listed as a warning line item similar to legacy disbursements not found to allow for the summary report process to capture the details. Example:
Disbursement schedule not found for disbursement value ID: 1223811987, record at line 6 skipped.
- Added an e2e test to test the above flow.

### **Screenshot:**

**E2E Test**

image

### DIFF
--- a/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
+++ b/sources/packages/backend/apps/queue-consumers/src/processors/schedulers/institution-integration/ece-response/_tests_/ece-response-integration.scheduler.e2e-spec.ts
@@ -200,8 +200,8 @@ describe(
           "Disbursements skipped to be processed: 0",
           "Disbursements considered duplicate and skipped: 0",
           "Disbursements failed to process: 0",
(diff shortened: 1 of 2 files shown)


sources/packages/backend/libs/integrations/src/institution-integration/ece-integration/ece-response.processing.service.ts
```diff
@@ -256,21 +256,42 @@ export class ECEResponseProcessingService {
...
+    // Filter out records with warnings and track them.
+    const validRecords = eceFileDetailRecords.filter((eceDetailRecord) => {
+      const validationResult = eceDetailRecord.validate();
+      if (!validationResult.length) {
+        return true;
+      }
+      const warningMessage = validationResult.find(
+        (result) => result.validationLevel === "warning",
+      )?.message;
+      const errorMessage = validationResult
```
andrewsignori-aot: Please check the `errorMessage` after the early return at line 276 to take advantage of the early return itself.

Issue #5569, opened 2025-12-30
Estimated Award Table on Funding Summary Page: Overawards, Restrictions and Previously received payments 


AnnaPBashkatova:
As Student, Ministry and BC Public Institution, I would like to see that overawards, restrictions, previous payments on application might lower/lowered different loan/grant type in Estimated Award and Final Award Screen on Funding Summary page. 

## **Acceptance Criteria**
Any labels applied to the Award Cards should be presented identically to Student and Ministry user and BC Public Institutions.

## FT Final Award 
- [ ] Populate loan/grant numbers with numbers requested on eCert.

## Both PT and FT:
- [ ] Replace "Estimated award" in _Final award_ Card with "Final award" in the column name in both PT and FT Funding Summary Page.

## Restrictions 

In both PT and FT Estimated award card, if **restrictions** might cause final/actual funding to be lower than estimated award amount (please disregard restrictions that will stop disbursement):
- [ ] Add red label next to each impacted funding type on application
- [ ] On hover,  display tooltip with text reading: 

> Restrictions on your account might result in funding lower than the current estimated award. If you require further clarification, please contact StudentAid BC.
- [ ] Once Final Award numbers are populated, no longer display labels in Estimated Award card.

In both PT and FT Final award Card, if **restrictions** caused final/actual funding to be lower than estimated award amount:
- [ ] Add red label next to each impacted funding type on application
- [ ] On hover,  display tooltip with text reading :

> Restrictions on your account have resulted in funding that is lower than the initial estimated award.

- [ ] If restriction was removed: 
    - [ ] Estimated Award: do not display the label.
    - [ ] Final Award: If restriction impacted ecert amount, keep display label even if it was resolved.

<img width="452" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTVlREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--663a06238dd7782a4a1f9bef8fd78a166f30bff3/image.png" alt="image.png" />

<img width="900" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTllREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--07d51f328577358fbd572d7b187aef20068b76ed/image.png" alt="image.png" />

## Overawards

[...]

PR #5763, opened 2026-02-11
#5569 - Fix defect when viewing empty assessment

weskubo-cgi:
### Overview
Fixed issue where clicking View when no assessment has been created shows a blank screen. Recent update to toggled logic did not take into account that there is a prop default of `{}`. Changed default to undefined so existing logic works as expected. Note, there will be a future fix to disable or hide the View button in this scenario.

image

image

### DIFF
--- a/sources/packages/web/src/components/common/students/applicationDetails/AssessmentAward.vue
+++ b/sources/packages/web/src/components/common/students/applicationDetails/AssessmentAward.vue
@@ -53,6 +53,7 @@ export default defineComponent({
     assessmentAwardData: {
       type: Object as PropType<AwardDetailsAPIOutDTO>,
       required: true,
+      default: undefined,
     },
     noticeOfAssessmentRoute: {
       type: Object as PropType<RouteLocationRaw>,


sources/packages/web/src/components/common/students/applicationDetails/AssessmentAward.vue
```diff
@@ -53,7 +53,6 @@ export default defineComponent({
     assessmentAwardData: {
       type: Object as PropType<AwardDetailsAPIOutDTO>,
       required: true,
```
dheepak-aot: Even though it is required(which will enforce the consumer to assign variable/value to the attribute), we can set the default to undefined reflecting the initial state of the value before API is completed/completed with error. 

Image

With default removed we'll have error at consumer. 

Image

weskubo-cgi: Undefined seems like the simplest approach.

weskubo-cgi:
404 is now handled gracefully. In the future, the user shouldn't be able to navigate to the screen in this scenario (unless by direct url navigation).

image

Issue #5680, opened 2026-01-26
Partner self declare ticket 1 (student app and workflow)


Joshua-Lakusta:
## User Story
As a ministry or student user, I need to update the workflow processes to accommodate the use of the Supporting User form for partner/common-law self-declaration. This is important to help students complete their applications accurately and ensure proper data collection.

### Baseline Information

- [ ] Applies to 2026–27 program year and onward.
- [ ] Change effects both FT and PT

### Modify Application

- [ ] Replace the current partner/common-law questions with a function similar to the dependants section that allows entry of information free form.

- [ ] Only one entry will be allowed so it does not need to behave as an array that allows additional lines of info.

- [ ] Student will default have one box for adding a partner/common-law if they indicate they have one previously in the application

- [ ] Use a validation to ensure partner’s full name does not match the student’s full name (soft warning).

- [ ] Wire frame, content is exact and can me mimicked exactly

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3FsREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4f2f05eb3e08f6d8b430b2e5b2f9804814be0920/image.png" alt="image.png" />

- [ ] Remove the exception for partner current year income as an available question in the form (it will be replaced by an appeal in a coming ticket.) (If ticket number 5616 is completed first this AC is not required as the question will already be removed)
- [ ] Remove all questions that were hidden under answering "no" to the BC services card question as they will not longer be used for student self declaring partner information.
- [ ] Ensure if the answer to the BC Services card question = no then there should be no ability for the actual partner to go to the supporting user portal and submit a form

## Technical Acceptance Criteria

DATA

- [ ] Stop relying on the student partner data from the application and start to utilize the data from the student self declare utilizing the supporting user form.

### Workflow

- [ ] Assessment Gateway: (pulled from previous parent effort validate assumption is true that this is still needed for partner)

- [ ] Worker Changes 
- [ ] Use existing worker: create-identifiable-supporting-user.
- [ ] Do note remove existing worker as it will continue to be used for program years 25-26 and earlier. (backwards compatibility)
- [ ] Handles partner creation 
- [ ] Idempotent using full name + relationship type.

- [ ] Data Loading
- [ ] Reuse load-assessment-consolidated-data.
- [ ] Add mapping for studentDataPartner.

[...]

Issue #5757, opened 2026-02-10
Partner self declare part 3 tracker card


Joshua-Lakusta:
### Ticket 3
** User Story**

Follow-up effort to update the tracker card when a student needs to self declare their partner information. This is set to match the parent self declare path that already exists.

Acceptance Criteria
Update Partner/Common-law Tracker Cards
Student Declare

- [ ] If a student is self declaring for their partner, update the tracker card to show as a “warning” card with a button.

- [ ] Tracker card includes one “Student Declare” button for the student to access the form to enter the partner/common-law information.

- [ ] The “Student Declare” button will display as the full name of the partner/common-law to be declared, as provided in the application.

- [ ] Change when the Partner/Common-law information complete green tracker card shows:

   - [ ] Only display it when the student of partner/common-law declared information is provided.

- [ ] Student declare content

> Partner/Common-law information required for [partner/common-law full name]
> 
> You have indicated that [partner/common-law full name] is unable to complete their declaration. Please complete the following declaration on their behalf.
> Click on the button below to complete the declaration.
> 
> Declare
> 

- [ ] Partner/Common-law Declare

> Waiting for additional information from [partner/common-law full name]
> 
> We are waiting for supporting information from [partner/common-law full name].
> Please check your email from StudentAidBC for further instructions.
> The email includes important details and a secure link that your partner/common-law will need in order to provide their information for your application.

PR #5784, opened 2026-02-19
#5680 - Partner self declare application form and gateway (part one)

weskubo-cgi:
## Overview
- Updated PT and FT 26-27 applications to remove all partner related income/exception questions other than whether they are able to report. Also added name match warning.
- Updated assessment gateway to always create identifiable supporting users for partners for 26-27.
- Ensure backwards compatibility with 25-26 and earlier applications.
- Added E2E tests to cover updates to workers and workflow.

## Notes
- 26-27 PT and FT assessment workflows will be updated separately to avoid conflicts
- Notifications will be implemented in a future story

### Warning Message
image

### Able to Report = Yes
image

### Able to Report = No
image

### E2E Tests
image

I'm currently having some local issues with Camunda timeouts so I'm just included the new tests below:
image

 All tests are passing in GitHub:
image

### DIFF
--- a/sources/packages/backend/apps/workers/src/controllers/supporting-user/_test_/e2e/supporting-user.controller.createIdentifiableSupportingUsers.e2e-spec.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/supporting-user/_test_/e2e/supporting-user.controller.createIdentifiableSupportingUsers.e2e-spec.ts
@@ -276,10 +276,10 @@ describe("SupportingUserController(e2e)-createIdentifiableSupportingUsers", () =
     });
   });
 
-  describe("Should create supporting user for the partner and create a notification for partner declaration required when the student is married and ", () => {
+  describe("Should create supporting user for the partner and create a notification for partner declaration required when the student is married.", () => {
     const ABLE_TO_REPORT_OPTIONS = [true, false];
     for (const ableToReport of ABLE_TO_REPORT_OPTIONS) {
-      it(`partner able to report is ${ableToReport} `, async () => {
+      it(`Should create supporting user for the partner and create a notification for partner declaration required when partner able to report is ${ableToReport}.`, async () => {
         // Arrange
         const partnerFullName = faker.string.uuid();
         const savedApplication = await saveFakeApplication(
--- a/sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
+++ b/sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
@@ -257,7 +257,7 @@ export class SupportingUserController {
     isAbleToReport: boolean,
     entityManager: EntityManager,
   ): Promise<void> {
-    if (supportingUserType == SupportingUserType.Parent) {
+    if (supportingUserType === SupportingUserType.Parent) {
       const notificationPayload:
         | ParentInformationRequiredFromParentNotification
         | ParentInformationRequiredFromStudentNotification = {
@@ -280,7 +280,7 @@ export class SupportingUserController {
           entityManager,
         );
       }
-    } else if (supportingUserType == SupportingUserType.Partner) {
+    } else if (supportingUserType === SupportingUserType.Partner) {
       await this.notificationActionsService.saveSupportingUserInformationNotification(
         {
           givenNames: application.student.user.firstName,
--- a/sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway-v2.bpmn
+++ b/sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway-v2.bpmn
@@ -28,8 +28,8 @@ Note: the workflow process should continue its execution as a regular reassessme
     <bpmn:association id="Association_1wys3mp" associationDirection="None" sourceRef="TextAnnotation_17tw6q4" targetRef="minitry-approval-change-in-progress-still-pending-gateway" />
     <bpmn:textAnnotation id="TextAnnotation_1q6xzna">
       <bpmn:text>There are two conditions under which a supporting user is created:
- 1. 25-26 and earlier applications where the partner is able to report. These applications have studentDataIsYourPartnerAbleToReport = true. 26-27 and newer do not have this property.
- 2. All 26-27 and later applications. These applications have the renamed property studentDataPartnerIsAbleToReport. 25-26 and older use a different name for the form property.</bpmn:text>
+ 1. The 2025-26 and earlier applications where the partner is able to report. These applications have studentDataIsYourPartnerAbleToReport = true.
+ 2. All 2026-27 and later applications. These applications have the renamed property studentDataPartnerIsAbleToReport.</bpmn:text>
     </bpmn:textAnnotation>
     <bpmn:association id="Association_0z482ud" associationDirection="None" sourceRef="TextAnnotation_1q6xzna" targetRef="Gateway_176fysj" />
   </bpmn:collaboration>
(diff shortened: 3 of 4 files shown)


sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway-v2.bpmn
andrewsignori-aot: Minor, the comments on the below flows can be misleading, unless I am missing something.
I would suggest the following.

Image

sources/packages/backend/workflow/test/test-utils/factories/assessment-consolidated-data.ts
```diff
@@ -235,6 +237,25 @@ export function createFakeSingleIndependentStudentData(): Partial<AssessmentCons
   };
 }
 
+/**
+ * Create fake married independent student data.
+ * @param partnerIsAbleToReport indicates if the partner is able to report income.
+ * @param partner1TotalIncome total income of the partner. If not provided it will be set as 10000.
+ * @returns consolidated data for married independent student.
+ */
+export function createFakeMarriedIndependentStudentData(
```
andrewsignori-aot: Following the `createParentsData`, can the parameters be created using `options`?

sources/packages/backend/workflow/test/2026-2027/assessment-gateway/assessment-gateway.originalAssessment.e2e-spec.ts
```diff
@@ -464,6 +466,83 @@ describe(`E2E Test Workflow assessment gateway on original assessment for ${PROG
...
+      WorkflowServiceTasks.VerifyAssessmentCalculationOrderTask,
+    );
+
+    expectNotToPassThroughServiceTasks(
+      assessmentGatewayResponse.variables,
+      MultiInstanceProcesses.parent1CreateIdentifiableParent,
+      MultiInstanceProcesses.parent1RetrieveInformation,
+      MultiInstanceProcesses.parent1IncomeVerification,
+    );
+  });
```
andrewsignori-aot: I am not sure how much value this assertion is adding. I would recommend having them removed.

sources/packages/backend/workflow/test/2026-2027/assessment-gateway/assessment-gateway.originalAssessment.e2e-spec.ts
```diff
@@ -464,6 +466,83 @@ describe(`E2E Test Workflow assessment gateway on original assessment for ${PROG
...
 
+  it("Should create and load data from a supporting user when the student is married and the partner is able to report.", async () => {
+    // Arrange
+    const partnerSupportingUserId = supportingUserId++;
+
+    // Assessment consolidated mocked data.
+    const assessmentConsolidatedData: AssessmentConsolidatedData = {
+      assessmentTriggerType: AssessmentTriggerType.OriginalAssessment,
+      ...createFakeConsolidatedFulltimeData(PROGRAM_YEAR),
+      ...createFakeMarriedIndependentStudentData(),
```
andrewsignori-aot: Is there a plan to create some other E2Es with variations to test different flows?
Unlike other E2E tests, there will be nothing else testing these flows and conditions, so the idea is that the workflow E2E tests cover all necessary variations.

sources/packages/backend/apps/workers/src/controllers/supporting-user/_test_/e2e/supporting-user.controller.createIdentifiableSupportingUsers.e2e-spec.ts
```diff
@@ -275,6 +275,156 @@ describe("SupportingUserController(e2e)-createIdentifiableSupportingUsers", () =
...
+      // Validate job result.
+      expect(result).toEqual({
+        resultType: MockedZeebeJobResult.Complete,
+        outputVariables: {
+          createdSupportingUserId: partner.id,
+        },
+      });
+      // Validate the creation of notification for the partner declare information
+      // to be provided by the student.
+      // TODO Add updated assertion when SIMS #5757 is implemented.
```
andrewsignori-aot: Is there a business agreement or AC on #5757 to have the notifications for the partner customized?
Should we not use the current `saveSupportingUserInformationNotification` for partner notifications for 2025-26 and early, since we are trying to have the worker centralized?

sources/packages/backend/apps/workers/src/controllers/supporting-user/_test_/e2e/supporting-user.controller.createIdentifiableSupportingUsers.e2e-spec.ts
```diff
@@ -275,6 +275,156 @@ describe("SupportingUserController(e2e)-createIdentifiableSupportingUsers", () =
     });
   });
 
+  it(
```
andrewsignori-aot: If the only difference between the tests is the assertion for the boolean variable, we can use smilar approach from other E2E tests where an array with the input data is created,d and tests are executed in a `for` loop.

sources/packages/forms/src/form-definitions/sfaa2026-27-ft.json
```diff
@@ -5123,821 +5053,25 @@
...
-                  "type": "number",
-                  "decimalLimit": 0,
-                  "input": true
-                }
-              ],
-              "allowPrevious": false
-            }
-          ]
+          "label": "Is partner name match",
+          "calculateValue": "const studentFullName = `${data.studentGivenNames} ${data.studentLastName}`.trim().toLowerCase();\nvalue = data.partnerFullName?.trim().toLowerCase() === studentFullName;",
```
andrewsignori-aot: The logic will not work for mononym names because the `studentFullName` string will be "null jack".
Something like below may resolve the issue.
```js
const studentFullName = `${data.studentGivenNames ?? ''} ${data.studentLastName}`.trim().toLowerCase();
```

andrewsignori-aot: Can we please check if it would be possible to not save this hidden field with the dynamic data?
I do not recall if there is a need to display the warning while in read-only, if so, just ensure that it would be visible for the Ministry.

Image

sources/packages/forms/src/form-definitions/sfaa2026-27-ft.json
```diff
@@ -4935,6 +4925,68 @@
...
+                {
+                  "components": [
+                    {
+                      "label": "Full name",
+                      "applyMaskOn": "change",
+                      "tableView": true,
+                      "validate": {
+                        "required": true
+                      },
+                      "validateWhenHidden": false,
```
andrewsignori-aot: In general, I believe the UI just lets the input take all 12 cols.

andrewsignori-aot (commented):

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway-v2.bpmn
weskubo-cgi: I should have updated these to reflect the new logic.

sources/packages/backend/workflow/test/2026-2027/assessment-gateway/assessment-gateway.originalAssessment.e2e-spec.ts
weskubo-cgi: This pattern was used in some other tests as a negative assertion but I can remove.

sources/packages/backend/apps/workers/src/controllers/supporting-user/_test_/e2e/supporting-user.controller.createIdentifiableSupportingUsers.e2e-spec.ts
weskubo-cgi: The method `createIdentifiableSupportingUsers` has the following code which is specific to parents, so I assumed we would need code updates to add similar functionality for partners.
```
    if (isAbleToReport) {
      await this.notificationActionsService.saveParentInformationRequiredFromParentNotification(
        notificationPayload,
        entityManager,
      );
    } else {
      await this.notificationActionsService.saveParentInformationRequiredFromStudentNotification(
        notificationPayload,
        entityManager,
      );
```

sources/packages/backend/apps/workers/src/controllers/supporting-user/_test_/e2e/supporting-user.controller.createIdentifiableSupportingUsers.e2e-spec.ts
weskubo-cgi: I will look into it.

sources/packages/forms/src/form-definitions/sfaa2026-27-ft.json
weskubo-cgi: I confirmed that the warning should display in read-only and it works fine with `Persistent = None` and `Calculate Value on server` unselected.

sources/packages/backend/workflow/test/2026-2027/assessment-gateway/assessment-gateway.originalAssessment.e2e-spec.ts
weskubo-cgi: As discussed, I will ensure tests cover both isPartnerAbleToReport true/false scenarios.

sources/packages/backend/workflow/src/workflow-definitions/assessment-gateway-v2.bpmn
andrewsignori-aot: To my understanding, was any work done to adjust these comments?

sources/packages/backend/workflow/test/2026-2027/assessment-gateway/assessment-gateway.originalAssessment.e2e-spec.ts
```diff
@@ -464,6 +466,83 @@ describe(`E2E Test Workflow assessment gateway on original assessment for ${PROG
     );
   });
 
+  describe("Should create and load data from a supporting user when the student is married and  ", () => {
+    const ABLE_TO_REPORT_OPTIONS = [YesNoOptions.Yes, YesNoOptions.No];
+    for (const ableToReport of ABLE_TO_REPORT_OPTIONS) {
+      it(`partner able to report is ${ableToReport}`, async () => {
```
andrewsignori-aot: The `describe` should be able to describe a higher scope, grouping the test scenarios, while the `it` should be able to describe the test. In the way it is, the test scenario is broken between `describe` + `it`.

For instance, in the example below, the `describe` is used to group a context, while the `it` still describes the scenario. Please keep in mind how full sentences are created.

https://github.com/bcgov/SIMS/blob/dbf8604c23b17a9365f4e42f33611ab9abe21980/sources/packages/backend/apps/api/src/route-controllers/restriction/_tests_/e2e/restriction.aest.controller.getReasonsOptionsList.e2e-spec.ts#L39-L44

sources/packages/backend/apps/workers/src/controllers/supporting-user/supporting-user.controller.ts
```diff
@@ -260,6 +242,58 @@ export class SupportingUserController {
...
+   * @param entityManager entity manager to be used in the transaction.
+   */
+  private async saveSupportingUserNotification(
+    application: Application,
+    supportingUserType: SupportingUserType,
+    fullName: string,
+    isAbleToReport: boolean,
+    entityManager: EntityManager,
+  ): Promise<void> {
+    if (supportingUserType == SupportingUserType.Parent) {
```
andrewsignori-aot: `===` is preferred. Same for line 283.

sources/packages/backend/workflow/test/2026-2027/assessment-gateway/assessment-gateway.originalAssessment.e2e-spec.ts
weskubo-cgi: Thanks. I saw describe/for/it pattern used in many ways (as far as the description goes) so I will follow this example.

Issue #2987, opened 2024-03-12
Institution Restriction - Prevent Remittance


CarlyCotton:
**User Story**
As a ministry user, I want to restrict some institutions from being able to remit funds directly when COE is confirmed.

**Acceptance Criteria**
- [x] Add new institution LOCATION Restriction:
    - Code: REMIT 
    - Description: No direct tuition remittance allowed.
    - Category: Location
    - Notification Type: No Effect

<img width="851" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWlzREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--124da753837535abfaf0abeba977595ade63306e/image.png" alt="image.png" />

- [x] **Active Restriction:** Change COE confirmation modal to hide the remit options, treat it as "No" remittance requested.
<img width="559" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWVzREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--43f05bcbe98890a318be66018d201bbce12f727b/image.png" alt="image.png" />

- [ ] **Processing ECE:** Schools can submit remittance requests, no error if an amount is entered other than "0". Treat as $0 requested and process confirmation. Provide warning in confirmation "X records had remittance requested and due to restrictions will not be submitted"
- [x] **In-flight Apps:** Any previously requested tuition remittance can continue to be processed as requested.
- [ ] If an institution user happens to enter a remittance amount during COE when they have a restriction, show them an error _"Confirmation of enrolment was not completed. Please refresh and try again."_ (or something similar)

**Technical Criteria**
- [ ] Add some metadata to the restrictions to allow to dictate if the program or location or both are required, allowing the UI and the API to react accordingly and requiring only what is needed.
- [ ] Use restriction code to apply the actions

CarlyCotton:
@dheepak-aot I added an AC based on the discussion this morning:

>- [ ] If an institution user happens to enter a remittance amount during COE when they have a restriction, show them an error "Confirmation of enrolment was not completed. Please refresh and try again." (or something similar)

PR #5791, opened 2026-02-22
#2987 - Institution prevent remittance - Enable location only restrictions

dheepak-aot:
# Institution prevent remittance - Enable location only restrictions

## DB Migrations

- [x] Added a generic metadata column to `sims.restrictions` and updated the metadata for SUS retriction.
- [x] Created a new restriction REMIT with metadata values populated.

### Metadata framework
- [x] The metadata column in `sims.restrictions` has a property `fieldRequirements` to evaluate the field requirements of the restricted party(Student restriction | Institution restriction).
- [x] The validators are built for the metadata framework(kept in API scope as of now) to validate the field requirements based on the metadata property and actual field values.
- [x] Dropped the NOT NULL constraint for program_id and location_id in `sims.institution_restrictions`. The program and location validations are governed by restriction metadata now. 
- [x] Updated the unique constraint `institution_id_location_id_program_id_restriction_id_is_active_unique` to consider NULL as value for uniqueness for the existing combination of `institution_id,location_id,program_id and restriction_id`
### Rollback evidence

image

image

image

## API

- [x] Updated the getReasonsOptionsList to make the restriction category as optional parameter.
- [x] Updated the getInstitutionRestrictions to return institution restrictions with program or location as NULL(Will adjust the E2E for this in next PR).
- [x] Update addInstitutionRestriction to adapt the program and locations being potentially NULL and use the metadata to execute field requirement validations.

## UI
- [x] ON add restrictions page, the UI optionally displays Program and Locations component based on selected restriction.
image

image

image

image

## E2E Tests
- [x] RestrictionAESTController(e2e)-addInstitutionRestriction.
```log
 RestrictionAESTController(e2e)-addInstitutionRestriction.                                                                                                                         
    √ Should add multiple SUS institution restrictions when a valid payload with program ID and multiple locations IDs is submitted. (231 ms)                                       
    √ Should add multiple REMIT institution restrictions when a valid payload with multiple locations IDs is submitted. (96 ms)                                                     
    √ Should add a single institution only institution restriction when a valid payload without program ID and location IDs is submitted. (44 ms)                                                                                                            
    √ Should throw a bad request exception on add REMIT restriction when no locations were provided in the request payload. (22 ms)                                                 
    √ Should throw a bad request exception on add REMIT restriction when program is provided in the request payload. (40 ms)                                                        
    √ Should throw a bad request exception on add SUS restriction when no program and locations were provided in the request payload. (20 ms)                                       

```
- [x] RestrictionAESTController(e2e)-getReasonsOptionsList.

```log
RestrictionAESTController(e2e)-getReasonsOptionsList.                                                                                                                             
    √ Should get all the Institution restrictions reasons list when there is no restriction category filter applied. (17 ms)                                                                                                                                      
    Should get reasons options list for a restriction type when the request is valid.                                                                                               
      √ Should get Provincial restrictions reasons list filtered by specific category when Provincial restrictions are requested for a category. (107 ms)                           
      √ Should get Institution restrictions reasons list filtered by specific category when Institution restrictions are requested for a category. (24 ms)
```

### DIFF
--- a/sources/packages/web/src/components/institutions/modals/AddRestrictionModal.vue
+++ b/sources/packages/web/src/components/institutions/modals/AddRestrictionModal.vue
@@ -14,6 +14,7 @@
           variant="outlined"
           :rules="[(v) => checkNullOrEmptyRule(v, 'Reason')]"
           :loading="loadingData"
+          @update:model-value="resetFormModelValues()"
           hide-details="auto" />
         <v-autocomplete
           v-if="programAttributes.canShow"
@@ -138,6 +139,19 @@ export default defineComponent({
     const programAttributes = computed(() => {
       return getFieldAttributes(PROGRAM_FIELD_KEY, "Program");
     });
+
+    /**
+     * Resets the form model values based on the visibility of the fields.
+     * This is required to avoid sending invalid data to the API in case the user changes the reason after filling some fields.
+     */
+    const resetFormModelValues = () => {
+      if (!locationAttributes.value.canShow) {
+        formModel.locationIds = [];
+      }
+      if (!programAttributes.value.canShow) {
+        formModel.programId = undefined;
+      }
+    };
     const {
       showDialog,
       showModal: showModalInternal,
@@ -230,6 +244,7 @@ export default defineComponent({
       checkNullOrEmptyRule,
       locationAttributes,
       programAttributes,
+      resetFormModelValues,
     };
   },
 });
--- a/sources/packages/web/src/services/http/dto/Restriction.dto.ts
+++ b/sources/packages/web/src/services/http/dto/Restriction.dto.ts
@@ -75,8 +75,8 @@ export interface AssignRestrictionAPIInDTO extends ResolveRestrictionAPIInDTO {
  * Add restriction to an institution.
  */
 export interface AssignInstitutionRestrictionAPIInDTO extends AssignRestrictionAPIInDTO {
-  locationIds: number[];
-  programId: number;
+  locationIds?: number[];
+  programId?: number;
 }
 
 /**


sources/packages/backend/apps/db-migrations/src/migrations/1771299851366-RestrictionAddColMetadataUpdateSUSInsertREMIT.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class RestrictionAddColMetadataUpdateSUSInsertREMIT1771299851366 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Add-col-metadata-and-update-sus.sql", "Restrictions"),
+    );
+    await queryRunner.query(
+      getSQLFileData("Insert-remit-restriction.sql", "Restrictions"),
+    );
```
andrewsignori-aot: I really struggle to understand the rationale behind multiple files for these operations.
Are the assumptions below right?
- The SQL files are affecting the same table, which means both are placed under the same folder/scope.
- They are both INSERTING/UPDATING data.
- They will always be applied or reverted together.
- There is nothing special about the rollback that would demand some specific order.

sources/packages/backend/apps/api/src/route-controllers/restriction/_tests_/e2e/restriction.aest.controller.addInstitutionRestriction.e2e-spec.ts
```diff
@@ -371,6 +570,81 @@ describe("RestrictionAESTController(e2e)-addInstitutionRestriction.", () => {
...
+      .post(endpoint)
+      .send({
+        restrictionId: susRestriction.id,
+        noteDescription: "Add institution restriction note.",
+      })
+      .auth(token, BEARER_AUTH_TYPE)
+      .expect(HttpStatus.BAD_REQUEST)
+      .expect({
+        message:
+          "Field requirement error(s): program is required, location is required.",
```
andrewsignori-aot: Really great to see the details in the validations and the validation messages ❤️

sources/packages/backend/apps/api/src/route-controllers/restriction/_tests_/e2e/restriction.aest.controller.getReasonsOptionsList.e2e-spec.ts
```diff
@@ -37,59 +57,74 @@ describe("RestrictionAESTController(e2e)-getReasonsOptionsList.", () => {
...
-        message: [
-          "type must be one of the following values: Provincial, Institution",
-        ],
-        error: "Bad Request",
-        statusCode: HttpStatus.BAD_REQUEST,
+      .expect(HttpStatus.OK)
+      .then((response) => {
+        const result = response.body;
+        expect(result).toEqual(
+          expect.arrayContaining(
```
andrewsignori-aot: Is the `expect.arrayContaining` here due to possible differences in the order? If yes, can we add an order to ensure the API will be consistent?

sources/packages/backend/apps/api/src/constants/error-code.constants.ts
```diff
@@ -336,3 +336,8 @@ export const INSTITUTION_RESTRICTION_ALREADY_ACTIVE =
  */
 export const NO_LOCATION_SELECTED_FOR_DESIGNATION =
   "NO_LOCATION_SELECTED_FOR_DESIGNATION";
+
+/**
+ * Field requirements not valid.
+ */
+export const FIELD_REQUIREMENTS_NOT_VALID = "FIELD_REQUIREMENTS_NOT_VALID";
```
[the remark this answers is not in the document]
dheepak-aot: I was considering that this metadata pattern could be followed as common practice for other resources in future so kept the error name more generic. Let me know, if that shouldn't be considered.

sources/packages/backend/libs/sims-db/src/entities/restriction.type.ts
```diff
@@ -15,3 +17,12 @@ export enum RestrictionType {
    */
   Institution = "Institution",
 }
+/**
+ * Restriction metadata.
+ */
+export interface RestrictionMetadata {
```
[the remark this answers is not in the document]
dheepak-aot: Got it. I can move it to restriction.model.ts

PR #5797, opened 2026-02-23
#5800 - Dependabot - Nodejs update - 25.6.1

dheepak-aot:
- Updated nodejs docker image from `node:25.6.0-alpine3.22` to `node:25.6.1-alpine3.22`
- Added the dependabot comment to all Dockerfile with image from docker hub so that dependabot will scan them for updates and not ignore the image scanning considering artifactory as private repo.

image


weskubo-cgi (approved):
Looks good. Thanks for handling this.

Issue #5687, opened 2026-01-28
Disable View Button for Assessments (Submitted/In Progress)


CarlyCotton:
**User Story**
As a student or ministry user, I would like to see more about my assessment by clicking the 'View' button in my assessments screen. Currently, this button is available in all assessment states including when the assessment has not yet been calculated (waiting on PIR/Exception/CRA). When clicking this button, I only want to see the pages if the assessment calculation has completed because there is no actual information I need until the assessment is complete.

**Acceptance Criteria**
- [ ] When the assessment has not completed, disable the 'View' button
- [ ] Make button blue?

<img width="1102" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVhBREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f6e5f7a193b4e69cf661134b5a40609949bcfcef/image.png" alt="image.png" />

PR #5892, opened 2026-03-12
#5687 - Disable View assessment when assessment data not available

weskubo-cgi:
## Overview
- Update View logic in Assessment to disable **View** button when the assessment data isn't available (e.g. when supporting user info is pending). Note, the View Assessment screen already handles missing assessment data gracefully.
- Updated button styling
### Ministry - Supporting User declaration pending (View disabled)
image
image

### Ministry - Unsuccessful Completion (View hidden - existing logic)
image

### Ministry - View enabled
image

### Student - View enabled
image

### Institution - View enabled
image

### DIFF
--- a/sources/packages/web/src/components/common/students/assessment/History.vue
+++ b/sources/packages/web/src/components/common/students/assessment/History.vue
@@ -55,7 +55,7 @@
             <template #[`item.assessment`]="{ item }">
               <v-btn
                 v-if="!item.hasUnsuccessfulWeeks"
-                :disabled="disableViewButton(item)"
+                :disabled="!item.assessmentDate"
                 @click="$emit('viewAssessment', item.assessmentId)"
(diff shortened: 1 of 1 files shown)


sources/packages/web/src/components/common/students/assessment/History.vue
```diff
@@ -54,9 +54,8 @@
             </template>
             <template #[`item.assessment`]="{ item }">
               <v-btn
-                v-if="!item.hasUnsuccessfulWeeks"
+                v-if="showViewButton(item)"
```
dheepak-aot: The requirement was to disable the button instead of hiding it, when assessment not calculated.

weskubo-cgi: Fixed!

Issue #5858, opened 2026-03-06
FT 2 Disbursements with No Eligible Awards


CarlyCotton:
**Acceptance Criteria**
- [ ] Currently in the 2 disbursements configuration for FT, if the student is not eligible for any awards there is no disbursement schedule created. It creates an error when we get to the MSFAA process.
- [ ] Update FT Configure disbursement and add new test cases.

## QA Notes
1. Create FT Application that is fully ineligible (Small offering costs, married student, no dependants, max income each, no PD) and has 2 disbursements (>18 weeks)
2. Assessment should complete and Funding SUmmary/NOA should both show fully ineligible for all awards.

**Possible Issue:**
<img width="1129" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN05NRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--45039e6d90450f06e736ef1ea600d1a168ba9b06/image.png" alt="image.png" />

**Example of Fully Ineligible Disbursement Schedule that does not create error:**
<img width="817" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHRVRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--797a4f32211a0a264efbdfbc44019a21adefc002/image.png" alt="image.png" />

**Example of Fully Ineligible Disbursement Schedule that does not create error:**
<img width="946" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHhVRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c1c2c62ff98496c2a22ce51adccacbdc22ba1c17/image.png" alt="image.png" />

PR #5975, opened 2026-03-30
#5858 - FT 2 Disbursement - Configure Disbursement Fix

CarlyCotton:
## Configure Disbursement
- Removed the filter which prevents an empty disbursement being passed when student is fully ineligible.
- Updated the 2 disbursement paths to have consistent disbursement schedules. Previously awards were omitted when due to business rules they would never have awards based on the split of award type disbursements.
- Corrected the calculation to determine if the BCSL needs to be split between first and second disb when >18 weeks.
image

## Test Cases
- Added more test cases for 2 disbursement situations in 26/27 and 25/26
- Updated the test models for disbursement schedules

### DIFF
--- a/sources/packages/backend/workflow/test/2025-2026/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
+++ b/sources/packages/backend/workflow/test/2025-2026/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
@@ -13,7 +13,7 @@ describe(`E2E Test Workflow fulltime-assessment-${PROGRAM_YEAR}-disbursements`, 
     // Arrange
     const configureDisbursementData =
       createFakeConfigureDisbursementFullTimeData(PROGRAM_YEAR);
-    configureDisbursementData.offeringWeeks = 17;
+    configureDisbursementData.offeringWeeks = OFFERING_WEEKS;
     configureDisbursementData.offeringStudyStartDate = getISODateOnlyString(
       addDays(30),
     );
--- a/sources/packages/backend/workflow/test/2026-2027/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
+++ b/sources/packages/backend/workflow/test/2026-2027/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
@@ -13,7 +13,7 @@ describe(`E2E Test Workflow fulltime-assessment-${PROGRAM_YEAR}-disbursements`, 
     // Arrange
     const configureDisbursementData =
       createFakeConfigureDisbursementFullTimeData(PROGRAM_YEAR);
-    configureDisbursementData.offeringWeeks = 17;
+    configureDisbursementData.offeringWeeks = OFFERING_WEEKS;
     configureDisbursementData.offeringStudyStartDate = getISODateOnlyString(
       addDays(30),
     );


sources/packages/backend/workflow/test/models/assessment.model.ts
```diff
@@ -539,7 +549,7 @@ export interface CalculatedAssessmentModel {
     limitWeeklyTransportationAllowance: number;
   };
   // Disbursement schedules
-  disbursementSchedules: Array<unknown>;
+  disbursementSchedules: Array<DisbursementSchedule>;
```
andrewsignori-aot: I would recommend using the syntax below as it seems the most used in the project.
```ts
disbursementSchedules: DisbursementSchedule[];
```

sources/packages/backend/workflow/test/2026-2027/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
```diff
@@ -154,57 +269,291 @@ describe(`E2E Test Workflow fulltime-assessment-${PROGRAM_YEAR}-disbursements`,
...
+          negotiatedExpiryDate:
+            configureDisbursementData.offeringStudyStartDate,
+          disbursements: [],
+        },
+      ]),
+    );
+  });
+
+  // An array of award amounts to verify first and second disbursement value amounts.
+  const TEST_FINAL_DISBURSEMENT_AMOUNTS = [
```
andrewsignori-aot: This is becoming closer to what we suggested in the past about having a CSV-like data input and output expectation to allow biz to contribute with scenarios that can be automated.

andrewsignori-aot (commented):
Nice work, please take a look at the minor comments.

sources/packages/backend/workflow/test/2025-2026/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
```diff
@@ -11,11 +11,13 @@ describe(`E2E Test Workflow fulltime-assessment-${PROGRAM_YEAR}-disbursements`,
     // Arrange
     const configureDisbursementData =
       createFakeConfigureDisbursementFullTimeData(PROGRAM_YEAR);
+    configureDisbursementData.offeringWeeks = 17;
```
weskubo-cgi: **suggestion:** Not a required change, but code like the following would reduce the number of magic numbers used:
```
    const OFFERING_WEEKS = 17;
    configureDisbursementData.offeringWeeks = OFFERING_WEEKS;
    const START_DATE = addDays(30);
    configureDisbursementData.offeringStudyStartDate =
      getISODateOnlyString(START_DATE);
    // 17 weeks is 119 days, so add 119 days to the offering start date to set the offering end date.
    configureDisbursementData.offeringStudyEndDate = getISODateOnlyString(
      addDays(OFFERING_WEEKS * 7, START_DATE),
    );
```

sources/packages/backend/workflow/test/2025-2026/fulltime-assessment/disbursements/fulltime-assessment-disbursements.e2e-spec.ts
weskubo-cgi: It's best to avoid `arrayContaining` unless absolutely necessary as it the test will pass even if the response contains objects not in the expected array.

Issue #5878, opened 2026-03-11
Student Forms e2e Tests 


ninosamson:
**User story**
As a SIMS team, we need to ensure appropriate test coverage on all features in the application

**Acceptance Criteria**
- [ ] ensure every new end point under "forms" has e2e tests 
- [ ] ensure unit tests from appeal actions are moved to new structure

PR #5986, opened 2026-03-31
#5878 - Student Forms e2e Tests - Actions Unit Tests

andrewsignori-aot:
#### processActions

- Should execute multiple actions and pass the parameters as expected when multiple requests with different actions are part of the same form submission.
- Should execute actions uniquely when multiple requests with the same actions are part of the same form submission.
- Should throw an error when the form submission has an unknown action.
    
#### getSubmissionItemsByActionType

- Should return only the requests that include the matching action type.
- Should return only the requests that include the matching action type and decision status.
- Should not return any requests when no matching action type and decision status are found.

#### process
- Should call applyAction when appliesTo result is true.
- Should not call applyAction when appliesTo result is false.

### Changed the "Assessed date" to display a "-" instead of "Pending".

image

### Added the ":" in the validation messages.

image

### DIFF
--- a/sources/packages/web/src/components/form-submissions/FormSubmissionApproval.vue
+++ b/sources/packages/web/src/components/form-submissions/FormSubmissionApproval.vue
@@ -27,7 +27,7 @@
               v-model="decision.decisionNoteDescription"
               hide-details="auto"
               :rules="[
-                (v) => checkNotesLengthRule(v, `${decision.parentName}, notes`),
+                (v) => checkNotesLengthRule(v, `${decision.parentName}: notes`),
               ]"
               required
               :readonly="decision.decisionSaved"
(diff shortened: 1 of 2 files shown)


sources/packages/backend/apps/api/src/services/form-submission/form-submission-actions/_tests_/unit/form-submission-action-processor.processActions.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  });
+});
+
+/**
+ * Mock the database result for the form submission details retrieval to allow testing
+ * the action processor logic without relying on the database layer.
+ * @param formSubmission form submission details to be returned by the mock.
+ * @returns mocked EntityManager with the form submission details query implemented.
+ */
+function createFormSubmissionFindOneMock(
```
dheepak-aot: As the usage of this method is more generic, it can take a generic type T and mock the findOne method.

Issue #5618, opened 2026-01-13
[DEV CONNECT] Migrate to Gold and Upgrade Postgres


ninosamson:

dheepak-aot:
Adding commands executed so far: 
Execute full backup on repo2.
`pgbackrest --stanza=db backup --type=full --repo=2`

PR #6025, opened 2026-04-13
#5618 - Update sysdig teams and group deployed packages

dheepak-aot:
- [x] Added a make command for sysdig teams to include new license plates (Separate command was used as the SA token for silver will not have access to deploy in gold cluster)
 - [x] The SIMS packages which are part of release(API, Web, Workers, Queue Consumers and Migrations) are added to a group in OC `sims-packages-group`
 - [x] Added visual connections on OC from sims-api to redis and queue consumers to redis and clamav.


andrewsignori-aot (approved):
Thanks for adding the sims group, looks good 👍

weskubo-cgi (approved):
Looks good. Thanks for this improvement!

Issue #5117, opened 2025-08-28
PY 26/27 Independent Student: Remove Parent Information Tab


AnnaPBashkatova:
**As a** *Independent Student* **I do not want** *to answer questions about my parents* **as they have no impact on my assessment*

**Acceptance Criteria**
FT applications, PY 26/27 and onward
- [ ] Do not display the Parent Information section on the Personal Information tab until a student has answered all of the dependant/independant questions. 
- [ ] Only display it if the student is Dependant.

**Technical ACs**
- [ ] Capture the logic to determine estranged from parents into new hidden field. 
- [ ] Combine estranged from parents into dependantStatus to keep existing behaviour

**How to be an independant student:**
- Question: What is your marital status -- > Answers: Married/Common-law or Separated/Divorced/Widowed or Married/Common-law and unable to provide partner income due to domestic abuse ("married", "other", "marriedUnable"].includes(data.relationshipStatus))
- Question: Are you, or were you at the time of your 19th birthday a youth in the continuing care or custody of the Government of B.C -->Anwers: Yes
- Question: At time of your course study start date, will you have been out of high school for 4 years -->Answer: Yes (data.outOfHighSchoolFor4Years === "yes")
- Question: In the time since you left high school to your first day of classes, have you spent two periods of 12 consecutive months each, in the full time labour force -- > Answer: Yes (data.fulltimelabourForce === "yes")
- Question: Do you have any eligible dependants --> Answer: Yes (data.hasDependents === "yes")

PR #6035, opened 2026-04-14
#5117 - PY 26/27: Independent Student: Remove Parent Information Tab

sh16011993:
**As a part of this PR, the following was completed:**

Changed the logic for the PY 26/27 FT form to display the Parent Information only if the student is dependant.

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2026-27-ft.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2026-27-ft.json
@@ -2765,9 +2765,18 @@
           "label": "Panel"
         },
         {
-          "key": "parentInformationPanel",
-          "input": false,
           "title": "Parent information panel",
+          "collapsible": false,
+          "hideLabel": true,
+          "key": "parentInformationPanel",
+          "conditional": {
+            "show": true,
+            "when": "dependantstatus",
+            "eq": "dependant"
+          },
+          "type": "panel",
+          "label": "Panel",
+          "input": false,
           "tableView": false,
           "components": [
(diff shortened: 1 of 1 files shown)


sources/packages/forms/src/form-definitions/sfaa2026-27-ft.json
```diff
@@ -2765,9 +2765,18 @@
...
         {
+          "title": "Parent information panel",
+          "collapsible": false,
+          "hideLabel": true,
           "key": "parentInformationPanel",
+          "conditional": {
+            "show": true,
+            "when": "dependantstatus",
+            "eq": "dependant"
+          },
```
dheepak-aot, line 2776: As mentioned by the copilot, the dependantStatus is calculated based on the values that are provided from parent information, when the parent information is required and hance if the questions are answered to calculate to `independant` the panel will dissapear and most importantly the parent details will not be captured.


And on top of that, when dependantStatus becomes `dependant` the student application will display the parent residency status which is not a desired behaviour unless all the questions are answered.

sh16011993: I had read the comment from Copilot. The condition to calculate the `dependantStatus` had the  `!!data.parentsDeceased &&\r\n    !!data.estrangedFromParents && // Considered to be \"no\" at this moment.` removed. So the parent information tab would be visible if the questions are answered such that the student is a dependant student.

We can talk if the expectation is different.

sh16011993:
Creating a new PR.

Issue #5861, opened 2026-03-06
Update Camunda Deployment 


ninosamson:
## User story
As a SIMS team, we need to ensure CICD is streamlined 

## Acceptance Criteria
- [ ] Update deployment of Camunda definitions to execute in openshift 
- [ ] IP restrict access to Camunda to openshift IP egress

PR #6080, opened 2026-04-24
#5861 - Deploy Camunda Definitions in Openshift

tiago-graf:
### Summary

Updates CD to use the migration container when deploying Camunda workflow definitions


.github/workflows/release-deploy-all.yml
```diff
@@ -304,9 +304,9 @@ jobs:
 
   # Deploy Formio Definitions
   deployFormioDefinitions:
-    if: ${{ inputs.deployFormioDefinitions }}
+    if: ${{ !cancelled() && inputs.deployFormioDefinitions }}
     name: Deploy Form.io definitions
-    needs: [env-setup, deploy-sims-api]
+    needs: [env-setup, deploy-sims-api, deployCamundaDefinitions]
```
tiago-graf: Deployment of form/workflow definitions is now sequential since there's a race condition between the jobs when deleting/creating the container

devops/openshift/migrations-job.yml
```diff
@@ -28,6 +28,8 @@ objects:
               args: ["${MIGRATION_COMMAND}"]
               imagePullPolicy: Always
               env:
+                - name: HOME
+                  value: /tmp
```
tiago-graf: Needed by camunda to write to OAuth cache dir

andrewsignori-aot: Please use the `CAMUNDA_TOKEN_DISK_CACHE_DISABLE` instead.
See other areas of the source code using the same for reference.

.github/workflows/release-deploy-camunda-formio-definitions.yml
```diff
@@ -23,13 +23,13 @@ on:
         required: true
         type: string
     secrets:
-      ZEEBE_ADDRESS:
+      SA_TOKEN:
```
andrewsignori-aot, line 26: Why was it needed for Camunda deployment but not `release-deploy-formio-definitions.yml`?

devops/openshift/migrations-job.yml
```diff
@@ -105,15 +133,22 @@ parameters:
   - name: DB_PASSWORD_KEY
     value: password
   - name: FORMS_URL
-    required: true
   - name: FORMS_SECRET_NAME
-    required: true
+    value: forms-secret
```
andrewsignori-aot: For clarification, what was the motivation for this change?
This seems to be coming from HitHub Secrets as below.
```
FORMS_SECRET_NAME: ${{ secrets.FORMS_SECRET_NAME }}
```

.github/workflows/release-deploy-camunda-formio-definitions.yml
andrewsignori-aot, line 1: Do we still need distinct actions for forms and Camunda, considering they are executing the same Openshift job with a different parameter for `MIGRATION_COMMAND`?

devops/Makefile
```diff
@@ -435,8 +433,8 @@ deploy-and-run-migrations-job:
 		-p IMAGE_STREAM_TAG="$(MIGRATIONS_BUILD_REF):$(BUILD_REF)" \
 		-p BUILD_TAG=$(BUILD_REF) \
 		-p JOB_NAME=$(MIGRATIONS_JOB_NAME) \
-		-p FORMS_URL=$(FORMS_URL) \
-		-p FORMS_SECRET_NAME=$(FORMS_SECRET_NAME) \
+		$(if $(FORMS_URL),-p FORMS_URL=$(FORMS_URL)) \
+		$(if $(FORMS_SECRET_NAME),-p FORMS_SECRET_NAME=$(FORMS_SECRET_NAME)) \
```
andrewsignori-aot: If the below comment is accepted, there will be no distinction between the provided variables.
https://github.com/bcgov/SIMS/pull/6080/changes#r3140615836

sources/packages/backend/migrations/Dockerfile
```diff
@@ -22,6 +22,9 @@ COPY backend/tsconfig.json ./
 COPY backend/tsconfig.build.json ./
 COPY backend/env-setup.js ./
 
+# Copying camunda sources
+COPY backend/workflow ./workflow
```
andrewsignori-aot: Please be as specific as possible and copy only the required files to execute the code, following the other `COPY` in place.

sources/packages/backend/migrations/Dockerfile.dev
```diff
@@ -22,6 +22,9 @@ COPY backend/tsconfig.json ./
 COPY backend/tsconfig.build.json ./
 COPY backend/env-setup.js ./
 
+# Copying camunda sources
+COPY backend/workflow ./workflow
```
andrewsignori-aot: Same from here https://github.com/bcgov/SIMS/pull/6080/changes#r3140628668

andrewsignori-aot (commented):
Nice work, please take a look at the comments. Some may require a dev sync-up.

.github/workflows/release-deploy-camunda-formio-definitions.yml
tiago-graf: This is a sonar warning, since the secret is being used but not declared, I can add it to formio yml as well

devops/openshift/migrations-job.yml
tiago-graf: Since this is a optional parameter now the it will need a default value otherwise it will fail to parse the template, however if we decide to combine the deployment of definitions it won't be needed anymore

.github/workflows/release-build-all.yml
```diff
@@ -234,6 +234,5 @@ jobs:
     with:
       environment: DEV
       gitRef: ${{ needs.createTag.outputs.newTag }}
-      deployCamundaDefinitions: true
-      deployFormioDefinitions: true
+      deployCamundaFormioDefinitions: true
```
andrewsignori-aot: I would support the complete removal of this option. Not a blocker, but I do not see a reason to keep it  😉

devops/openshift/migrations-job.yml
```diff
@@ -55,13 +57,39 @@ objects:
                     secretKeyRef:
                       key: ${FORMIO_ROOT_EMAIL_NAME_KEY}
                       name: ${FORMS_SECRET_NAME}
+                      optional: true
```
andrewsignori-aot, line 60: For clarification only, are these parameters not going to be always present since the POD will execute both deployments?
Is the idea of having them `optional` to keep this YAML generic, since the deploy can happen individually?

andrewsignori-aot (approved):
Left two minor comments only.

devops/openshift/migrations-job.yml
tiago-graf: I'll double check, my understanding is that this pod will still run without those settings when running the db migrations, but you're probably right, this is not necessary

tiago-graf: Confirmed, db migrations step won't provide these parameters hence the optional flag

sh16011993 (approved):
Great work @tiago-graf 👍  Thanks for the clarification.

Issue #6092, opened 2026-04-29
BUG: 3.3:  History submission tab: Parent current year income Read-only: Radio buttons and uploaded info not displaying


astridSABC:
**Describe the Bug**
Parent current income appeal read-only does not display the radio button option selected and the associated information that was input (e.g. parent current year income, radio button selected, file upload)

**Expected Behaviour**
Selections that the student selected/entered are displayed

The upload information would be visible at some point similar to modified indepenedent

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMG05RFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4f11ca73e9f3f421edcf0c902bb134eff680f06c/image.png" alt="image.png" />

### Actual Behaviour
Radio button for the parent that was selected does not show any information selected in read-only, income field not displayed, select any radio buttons section, and file upload section

<img width="1212" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGk5RFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a62587eb4b08909370f8aa27e9ee55c0ea4f0231/image.png" alt="image.png" />

Environment: Test
App: 2026000091 (Tigress Tiger) for demo in daily standup/scrum
App: 2026000092 (Student current year income/exceptional expenses ok)

Environment: Dev
2026000040 (Pepe Green) for demo

Precondition
Create a full-time application that has a single dependant student with two parents (BC Services Card: "No"), parent information is provided and the application is at assessment status

** Steps To Reproduce**
Steps to reproduce the behaviour:
1.  Click "appeal" link in the row for the application above
2.  Select "parent current year income" checkbox
3.  Click "request appeal(s)"
4.  Select "no" for one parent
5.  Select "yes" for the other parent
6.  Enter a value in the total year gross income field
7.  Select a radio button option
8.  Attach a file
9.  Click "submit for review"
10.  Note the confirmation toast message
11.  Select the "history" tab
12.  Click "view" for the parent current year income
13.  Note the page

Ref #5722

weskubo-cgi:
@astridSABC  I also noticed a small issue with the way currency is handled that is inconsistent with other forms. I will handle the fix here as well.

Issue #5722, opened 2026-02-03
Parent current year income appeal


Joshua-Lakusta:
## User Story

Ministry users require that the exception for current year income for a parents be converted into an appeal process to allow for proper adjudication and a consistent application flow.

### General info
- [ ] 26–27 FT program year change only

- [ ] FT only

- [ ] Validations that will apply to this appeal are:
   - [ ] Application status is assessed
   - [ ] Application is not Archived
   - [ ] Add eligibility in FT 26/27 assessment when Parent data exists (`isEligible..` will be for any `dependant` students)

- [ ] Remove exception - and replace with appeal form
- [ ] Basic validation to ensure that at least one parent is yes in order to submit

- [ ] Description for submission page
Name: Parent current year income appeal
> If either of your parents has had, or is anticipated to have, a significant decrease in gross income, you may submit this appeal to request assessment using their current year estimated gross income.

### Appeal form info
- [ ] Appeal name: Parent current year income appeal
- [ ] New application form content (see below)

> #### Parent Current Year Income Appeal
> 
> If either of your parents has had, or is anticipated to have, a significant decrease in gross income for (dynamic year – e.g., 20XX if program year is 20XX–20YY), you may submit this appeal to request assessment using their current year estimated gross income.
> 
> #### Please provide supporting information
>

> Parent full name 1 
> Do you need to submit an income appeal for  "Parent full name 1" ?
o Yes (if yes expands to content below)
o No

> Enter your parents actual gross income from the beginning of the calendar year up to the date of this application, and their estimated gross income for the remainder of the calendar year, to calculate their total current year gross income.
> 
> "Parent full name 1's" total current year gross income is:
> 
> $ [Income field]
> 
> Reason for Significant Decrease in Gross Income
> 
> Select the reason that best describes the decrease in your partner/common-law partner’s gross income:
> 
> ○ Medical illness or injury
> 
> ○ Family emergency (e.g., death or injury)
> 
> ○ Natural disaster
> 
> ○ Layoff, strike, lockout, or other reduction in earnings beyond the family’s control
> 
> ○ Other
> 
> If “Other” is selected:
> Please explain the situation:
> 
> [Free-text field]
> 
> Upload Supporting Documents
> 
> Please upload documentation to support the request to use your partner/common-law partner’s current year estimated gross income.
> 
> Provide the following:
> 
[...]

PR #6094, opened 2026-04-29
#6092 - Fix read-only view

weskubo-cgi:
## Summary
- Discovered while testing #5722 
- Parent current income appeal read-only does not display the radio button option selected and the associated information that was input (e.g. parent current year income, radio button selected, file upload)
- Fix is to ensure grid is only initialized from parents supplementary-data for a new form
- Updated validation to work with empty grid to avoid issue with validation flashing briefly on form load
- Also updated income component from currency to number which is standard
- Fix will be cherry-picked to release 3.3

### Student - Appeal Submission
image

### Student - Appeal View
image

### Ministry - Read Only
image


andrewsignori-aot (approved):
Great troubleshooting and nice catch adjusting the validation when empty, looks good 👍

dheepak-aot (approved):
Great job!

Issue #6091, opened 2026-04-28
Post production gold migration tasks


dheepak-aot:
**Describe the task**
Post gold migration taks

**Acceptance Criteria**
- [ ] PR with updated prod values
- [ ] Make task and GHA for SA token validity check.
- [ ] Sysdig alerts config to gold staging.
- [ ] Local postgres upgrade to 18.
- [ ] New artifactory account setup.
- [x] Update following repository secrets and variables to gold values.
```
########## Secrets ############

SA_TOKEN=

########## Variables ############

BUILD_NAMESPACE=

OPENSHIFT_CLUSTER_URL=

OPENSHIFT_LICENSE_PLATE=e0a504
```
- [ ] Delete following repository secrets and variables from existing github environments(DEV, TEST, Staging and PROD).

```
########## Secrets ############

SA_TOKEN=

########## Variables ############

BUILD_NAMESPACE=

OPENSHIFT_CLUSTER_URL=

OPENSHIFT_LICENSE_PLATE=e0a504
```
**Additional context**
- Add any other context about the task here.
- Or here

PR #6106, opened 2026-05-05
#6091 - Update crunchy values prod and cleanup make files

dheepak-aot:
Update crunchy values prod and cleanup make files.


andrewsignori-aot (approved):
👍

weskubo-cgi (approved):
👍

PR #6109, opened 2026-05-06
#5635 - Formio Spring Upgrade - Backward Compatible Changes

andrewsignori-aot:
- Started using [API_KEYS](https://help.form.io/userguide/projects/project-settings#api-keys) instead of using user/pass/token authentication in the server-to-server communication.
  - Kept user and pass only for the server deployment.
- Forms deployment adjusted to stop using the JWT token.
- API forms service refactored to stop using the JWT token.
- New GitHub env secret `FORMS_API_KEY` to be added to the release instructions.
  - Please note, the form.io secrets entries are prefixed with `FORMIO_*`, hence the secret name was defined as `FORMIO_API_KEY` while the env variable was defined as `FORMS_API_KEY`, following the existing ones.


sources/packages/forms/src/deploy-forms.ts
```diff
@@ -30,7 +30,7 @@ interface FormDeployInfo {
     return;
   }
   console.info("Acquiring access token...");
```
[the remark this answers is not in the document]
andrewsignori-aot: Already updated.

configs/env-example
```diff
@@ -1,9 +1,11 @@
 # Set time zone for the same expected from Openshift.
+
```
weskubo-cgi, line 2: Were all the extra line breaks in this file added on purpose? It looks odd especially on multi-line comments.

weskubo-cgi (approved):
API key approach is great.

configs/env-example
andrewsignori-aot: It was added by prettier. If someone else can test, if there will be a different result in your own local, it would be good 😉

sources/packages/backend/apps/api/src/services/form/form.service.ts
```diff
@@ -22,21 +22,20 @@ const FORMIO_LIST_SORT_FIELD = "title";
 
 @Injectable()
 export class FormService {
-  private tokenCacheService: TokenCacheService = null;
+  private readonly config: FormsConfig;
+  private readonly authHeader: AxiosRequestConfig;
 
   constructor(
     private readonly configService: ConfigService,
```
dheepak-aot: Minor the config service isn't required to be global anymore.

dheepak-aot (approved):
Great work switching to API KEYS. 👍Added few minor comments.

sources/packages/backend/apps/api/src/services/form/form.service.ts
andrewsignori-aot: Pushed the change 😉

Issue #5896, opened 2026-03-13
Load Test - Formio/Application Submit


ninosamson:
**User Story**
As a SIMS team, we want to load test again before FT launch, focusing on potential bottlenecks. 

**Acceptance Criteria**
- [ ] Extend k6 load test, using load test gateway, to test application submission (anything else?) simulating peak load of 10,000/day in a short window 
- [ ] 200 concurrent users at peak?  
- [ ] confirm horizontal scalability and responsiveness of formio server

PR #6112, opened 2026-05-07
#5896 - Increase database connection limits

tiago-graf:
### Summary

Increases default pool size of pgBouncer from default (20) to 30. total connections now should be 90 instead of 60 (30 x 3 pods), still under the postgres default cap of 100.

### DIFF
--- a/devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
+++ b/devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
@@ -199,6 +199,7 @@ spec:
       config:
         global:
           client_tls_sslmode: disable
+          default_pool_size: "{{ .Values.proxy.pgBouncer.defaultPoolSize }}"
       {{ if .Values.proxy.pgBouncer.image }}
       image: {{ .Values.proxy.pgBouncer.image }}
       {{ end }}
--- a/devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
+++ b/devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
@@ -201,6 +201,7 @@ spec:
           client_tls_sslmode: disable
           default_pool_size: {{ .Values.proxy.pgBouncer.defaultPoolSize | quote }}
           max_client_conn: {{ .Values.proxy.pgBouncer.maxClientConnections | quote }}
+          pool_mode: {{ .Values.proxy.pgBouncer.poolMode }}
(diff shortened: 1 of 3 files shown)


devops/helm/crunchy-postgres/values-e0a504-dev.yaml
```diff
@@ -92,6 +92,7 @@ proxy:
   pgBouncer:
     image: # it's not necessary to specify an image as the images specified in the Crunchy Postgres Operator will be pulled by default
     replicas: 3
+    defaultPoolSize: 30
```
andrewsignori-aot, line 95: Can we have the default set at the `values.yaml`?

devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
```diff
@@ -199,6 +199,7 @@ spec:
       config:
         global:
           client_tls_sslmode: disable
+          default_pool_size: {{ .Values.proxy.pgBouncer.defaultPoolSize | default 20 | quote }}
```
andrewsignori-aot: Besides the `default_pool_size`, pgBouncer seems to expose other configurations that could be tested.

## TL;DR

- We can try changing the `pool_mode` to `transaction`.
- We can try to increase the `max_client_conn` to check if there is some impact. If there is, we can think about adjusting the number of POD connection pools to be aligned with `max_client_conn` (expected PODs vs POD connection pool size).

### `pool_mode`: session(default) / transaction / statement

Recommended for testing `transaction`. If the NestJS API uses short-lived transactions (which TypeORM typically does), switching to transaction mode would allow a much smaller default_pool_size **to serve many more concurrent API requests**.

### `max_client_conn` (default 100)

Should be set to the maximum number of concurrent app connections you expect (e.g., NestJS connection pool size × number of API pods).

Since we have 10 as a number for each POD, maybe we should increase the total so that they can be expanded, or reduce the number for each POD.

### `default_pool_size` (default 20 to 30 as set in this PR)

If `max_client_conn` = 100 but `default_pool_size` = 20, pgBouncer can queue up to 100 waiting clients but will only ever open 20 connections to PostgreSQL — the rest wait in pgBouncer's queue (governed by query_wait_timeout).

andrewsignori-aot (approved):
Great testing so far, looks good 👍

weskubo-cgi (approved):
👍

PR #6114, opened 2026-05-07
#5896 - Configure Horizontal Pod Scalers

tiago-graf:
### Summary

Reduces the number of max pods of queue-consumers and worker deployments.

### DIFF
--- a/devops/openshift/queue-consumers-deploy.yml
+++ b/devops/openshift/queue-consumers-deploy.yml
@@ -347,7 +347,7 @@ objects:
         kind: Deployment
         name: ${NAME}
       minReplicas: "${{REPLICAS}}"
-      maxReplicas: 10
+      maxReplicas: 4
       metrics:
         - type: Resource
           resource:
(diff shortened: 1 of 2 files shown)


devops/openshift/queue-consumers-deploy.yml
```diff
@@ -347,7 +347,7 @@ objects:
         kind: Deployment
         name: ${NAME}
       minReplicas: "${{REPLICAS}}"
-      maxReplicas: 10
+      maxReplicas: 4
```
andrewsignori-aot, line 350: Since we are not changing the current approach, I am ok to move ahead, since this is not the goal of this effort. That said, it would be good to have the max exposed in the way `REPLICAS` is exposed as a GitHub env var.

andrewsignori-aot (approved):
Loogks good 👍

weskubo-cgi (approved):
Thanks for doing the testing and analysis that allows us to make informed decisions here.

PR #6116, opened 2026-05-07
#5635 - Formio Spring Upgrade - Updated form.io definitions

andrewsignori-aot:
Fixed some issues for the upcoming form.io upgrade.

#### File Uploaders

- The conditional settings to display the uploaders were failing when using the reference for `dependants.declaredOnTaxes`, where, when two dependents were present, the answers for the first row would determine the condition for all rows. The condition was replaced by the `show = row.declaredOnTaxes === 'yes';` where `row` will have the right value for each dependent row, allowing the file uploader to work nicely.

image

#### Dependents Exception Description Name

- The "description name" contains the dependent name that comes from outside the container, where it is critical to allow the trigger of a new exception, forcing its hash to be different. The `rowIndex` was failing at the server side, and it is now relaced but the `row.fullName`.

### Program selection

- There were a few points accessing the `programnotListed` property without ensuring the `data.myProgramNotListed` was defined. For simplicity, the checks were changed to use the syntax as follows, which should be equivalent.

```ts
!!data.myProgramNotListed?.programnotListed
```

image

### Share information with trusted contact

- Moved the validation to ensure a maximum of two contacts can be added from a custom validation for the out-of-box form.io validation.
- This was causing a "warning" looping once the `@formio/js` was upgraded.

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/sfaa2022-23.json
+++ b/sources/packages/forms/src/form-definitions/sfaa2022-23.json
@@ -440,7 +440,7 @@
                       "refreshOnChange": false,
                       "customClass": "banner-info",
                       "key": "html12",
-                      "customConditional": "show = data.myProgramNotListed.programnotListed || (data.myStudyPeriodIsntListed && data.myStudyPeriodIsntListed.offeringnotListed);",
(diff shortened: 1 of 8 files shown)


dheepak-aot (approved):
Great findings. Looks good 👍

weskubo-cgi (approved):
Thanks for great analysis and fixes.

PR #6124, opened 2026-05-09
#5635 - Formio Spring Upgrade

andrewsignori-aot:
- Attempt to bump as many as possible the server-side and client-side Formio-related libs.
- NodeJS upgraded from 20 to 24. The idea is to try to follow the one used by the [formio Docker file](https://github.com/formio/formio/blob/main/Dockerfile).
- Form.io build changed to allow build from a commit sha defined by the env variable `FORMIO_SOURCE_REPO_REF`, which replaced the former tag variable `FORMIO_SOURCE_REPO_TAG`.

- Dependencies dynamically associated from `#main`, are now replaced by specific versions.

image

- The above dependencies are now replaced by npm package versions.
  - `@formio/core`, replaced by the env variable `FORMIO_CORE_VERSION`
  - `@formio/js` replaced by the env variable `FORMIO_JS_VERSION`
  - `@formio/vm` replaced by the env variable `FORMIO_VM_VERSION`
- `@formio/core` shows its most recent version as below, which seems outdated, but the [full version list is available](https://www.npmjs.com/package/@formio/core?activeTab=versions).

image

- The form.io build was changed to accept a "friendly" version (e.g., 9.6.8), and it will produce a combination with the appended `GitHub run_number`, as shown below.

image

image

### UI issues

- The below issues required some SCSS custom fixes due to the Bootstrap and `@fortawesome/fontawesome-free` package upgrades.

image

image

image


devops/Makefile
```diff
@@ -702,37 +707,51 @@ deploy-web-dev:
...
 	-p NAME=$(FORMS_NAME) \
 	-p FORMS_URL=$(FORMS_URL) \
 	-p TOOLS_NAMESPACE=$(BUILD_NAMESPACE) \
-	-p IMAGE_STREAM_TAG="$(FORMS_BUILD_REF):$(FORMIO_SOURCE_REPO_TAG)" \
+	-p IMAGE_STREAM_TAG="$(FORMS_BUILD_REF):$(FORMIO_BUILD_TAG)" \
 	-p FORMIO_ROOT_EMAIL=$(FORMIO_ROOT_EMAIL) \
 	-p FORMIO_ROOT_PASSWORD=$(FORMIO_ROOT_PASSWORD) \
-	-p FORMIO_API_KEY=$(FORMIO_API_KEY) \
 	-p MONGODB_URI=$(MONGODB_URI) \
 	-p CPU_REQUEST=$(FORMIO_CPU_REQUEST) \
```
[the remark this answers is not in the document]
andrewsignori-aot: Added `FORMIO_API_KEY`.

devops/Makefile
```diff
@@ -690,7 +695,7 @@ deploy-web-dev:
 	@test -n "$(TLS_KEY)" || (echo "Error: TLS_KEY is not set" && exit 1)
 	@test -n "$(TLS_CA_CERTIFICATE)" || (echo "Error: TLS_CA_CERTIFICATE is not set" && exit 1)
 	$(MAKE) deploy-web \
-		NAMESPACE=e0a504-dev \
+		NAMESPACE=0c27fb-dev \
```
dheepak-aot: why is this updated to silver namespace?

.github/workflows/env-setup-deploy-forms-server.yml
```diff
@@ -10,11 +10,11 @@ on:
...
         required: true
         type: environment
-      formioTag:
-        description: "Formio tag, please refer https://github.com/formio/formio/tags to deploy the appropriate tag."
+      formioBuildTag:
+        description: "Previously built Form.io build tag (e.g., v9.7.5-99)."
         type: string
         required: true
-        default: "v4.3.2"
+        default: "v9.7.5-0"
```
dheepak-aot: I am wondering if there is any need with default value when we use our version. Suggest to remove it.

devops/Makefile
```diff
@@ -40,7 +40,12 @@ export QUEUE_CONSUMERS_BUILD_REF := $(or ${QUEUE_CONSUMERS_BUILD_REF}, queue-con
 export LOAD_TEST_GATEWAY_BUILD_REF := $(or ${LOAD_TEST_GATEWAY_BUILD_REF}, load-test-gateway-${APP_NAME})
 export WEB_BUILD_REF := $(or ${WEB_BUILD_REF}, web-${APP_NAME})
 export FORMS_BUILD_REF := $(or ${FORMS_BUILD_REF}, forms)
-export FORMIO_SOURCE_REPO_TAG := $(or $(FORMIO_SOURCE_REPO_TAG), v4.3.2)
+# Form.io version targeting the API Server Version 9.7.5
+# see https://github.com/formio/enterprise-release/blob/master/API-Server-Change-Log.md#api-server-version-975
+export FORMIO_SOURCE_REPO_REF := $(or $(FORMIO_SOURCE_REPO_REF), f92dc5feab5cd0ff3ee30965141f11a8e552fe51)
```
dheepak-aot: As we are taking the latest commit, should it not be API server version 9.7.7 ? 

Image

sources/tests/Makefile
```diff
@@ -31,7 +31,12 @@ export ALLOW_BETA_INSTITUTIONS_ONLY :=$(or $(ALLOW_BETA_INSTITUTIONS_ONLY), fals
 
 # Formio forms integration
 export FORMIO_SOURCE_REPO_URL := $(or $(FORMIO_SOURCE_REPO_URL), https://github.com/formio/formio.git)
-export FORMIO_SOURCE_REPO_TAG := $(or $(FORMIO_SOURCE_REPO_TAG), v4.3.2)
+# Form.io version targeting the API Server Version 9.7.5
+# see https://github.com/formio/enterprise-release/blob/master/API-Server-Change-Log.md#api-server-version-975
+export FORMIO_SOURCE_REPO_REF := $(or $(FORMIO_SOURCE_REPO_REF), f92dc5feab5cd0ff3ee30965141f11a8e552fe51)
+export FORMIO_CORE_VERSION := $(or $(FORMIO_CORE_VERSION), 2.6.5)
+export FORMIO_JS_VERSION := $(or $(FORMIO_JS_VERSION), 5.3.5)
+export FORMIO_VM_VERSION := $(or $(FORMIO_VM_VERSION), 2.0.2)
```
dheepak-aot: As we have a combination of values now, we can centralize this combination.

dheepak-aot:
Great work and troubleshooting figuring out the path to form.io upgrade. Just a few minor comments.

devops/Makefile
andrewsignori-aot: I am reverting it.

.github/workflows/env-setup-deploy-forms-server.yml
andrewsignori-aot: And what would be our own version?
The idea was to try to have some sort of hint about where the current version's set was taken.
In a similar idea of what we are currently doing, we set the default for build and deploy.

devops/Makefile
andrewsignori-aot: When the last commit and last versions were taken, the API version was not updated.
I can update the PR with the newest version.

sources/tests/Makefile
andrewsignori-aot: It can be centralized in the "env" locally. These are the default values for each context.
If we centralize in the GitHub environments, it will not have an effect on the local environment, and we may also need a different one for the tests when we are upgrading.
Any ideas?

.github/workflows/env-setup-deploy-forms-server.yml
dheepak-aot: Ok. So when we have our current version, then it can be set as default value similar to what we do today. Got it.

sources/tests/Makefile
dheepak-aot: From the suggestion that I got from copilot, we can use the makefile fragment to centralize. (Disclaimer: I haven't tested it). If it is starting to block the PR(solution not working), you can move ahead.
image
image

dheepak-aot (approved):
Thanks for the changes. Looks good 👍

sources/packages/forms/Dockerfile
```diff
@@ -14,15 +17,21 @@ RUN apk update && \
   apk add git && \
   apk cache clean
 
-# Clone code into formio folder, by default.
-RUN git clone ${FORMIO_SOURCE_REPO_URL} -b ${FORMIO_SOURCE_REPO_TAG} --single-branch
+# Clone with a blobless filter to skip downloading file contents until checkout.
+# This makes it very fast while fully supporting tags, branches, and partial SHAs.
+RUN git clone --filter=blob:none ${FORMIO_SOURCE_REPO_URL} formio && \
+  git -C formio checkout ${FORMIO_SOURCE_REPO_REF}
```
weskubo-cgi, line 23: 👍

weskubo-cgi (approved):
Thanks for all the effort to get this working!

sources/tests/Makefile
andrewsignori-aot: Updated to include a shared file in a shared folder to allow other sets of envs to be shared.

tiago-graf (approved):

PR #6128, opened 2026-05-11
#5896 - Increase db max connections

tiago-graf:
### Summary

Increase db max connections from 100 to 125.
Increase db pool size from 30 to 40.
Set workers HPS max from 3 to 2.

### DIFF
--- a/devops/openshift/workers-deploy.yml
+++ b/devops/openshift/workers-deploy.yml
@@ -129,7 +129,7 @@ objects:
         kind: Deployment
         name: ${NAME}
       minReplicas: "${{REPLICAS}}"
-      maxReplicas: 3
+      maxReplicas: 2
       metrics:
         - type: Resource
           resource:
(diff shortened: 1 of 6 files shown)


devops/openshift/workers-deploy.yml
```diff
@@ -129,7 +129,7 @@ objects:
         kind: Deployment
         name: ${NAME}
       minReplicas: "${{REPLICAS}}"
-      maxReplicas: 3
+      maxReplicas: 2
```
andrewsignori-aot: What is the justification for having it set as 2?
Since the load test was not able to validate a real-world scenario, what would be the justification to keep it just as a `2`?

andrewsignori-aot (approved):
Configs look great, and thanks for the info on the load test so far.
Let's keep workers' instances at 3, looks good 👍

weskubo-cgi (approved):

PR #6129, opened 2026-05-11
#5335 - Formio Spring Upgrade - Divider Fix

andrewsignori-aot:
Minor fix for the v-divider, removing the `inset` and keeping some spacing as it was before.

### Before

image

image

### After

image

image


tiago-graf (approved):

dheepak-aot (approved):
👍thanks for taking care.

Issue #5684, opened 2026-01-28
COMMS: New SIMS Landing Page: PY 26/27


astridSABC:
**Describe the task**
As a student user, I want an update to the SIMS landing page so that it is clear where I need to login.
As an institution user, I want an update to the SIMS landing page so that it is clear where I need to login.

Updates to the SIMS landing page for PY 25/26.
Format of fonts, buttons, colour, alignment, indents addressed by IMB & SIMS when posting live.

### Acceptance Criteria

- [ ] URL for landing page [TBD]

**Student's section**

- [ ] Section Header: 
   

>  ### For Students

Frame split into three columns [TBD]

### Row 1

- [ ] Column One: Add Student icon
- [ ] Column Two: Body: Add content

> **New Student Account**

> **If applying after June XX, 2026**, use the new Student Account and Application system to create and manage your student aid applications. 

- [ ] Column Two: Add button below body content: 

> Access New Student Account

- [ ] Button on click link: https://sims.studentaidbc.ca/student/login

### Row 2

- [ ] Column One: Add Student icon for previous dashboard
- [ ] Column Two: Body: Add content

> **Old Student Dashboard Account** 

> **If you applied for full-time student aid before June XX, 2026**, access your dashboard and manage your application(s).  

- [ ] Column Two: Add button below body content: 

> Access Old Dashboard Account

- [ ] Button on click link: https://studentaidbc.ca/dashboard/login

- [ ] Column Three: Add disclaimer to the far right of the section in line with "Old Student Dashboard Account"
- [ ] Disclaimer symbol: Add ! in a circle to the right of the text

>   Not available to submit new applications 

**Institution section**

- [ ] Section Header: 
   

>  ### For Institutions

- [ ] Column One: Add Institution icon
- [ ] Column Two: Body: Add content

> **Institution Account**

> Create and manage your institution account using your Business or Basic BCeID User ID.

- [ ] Column Two: Add button below body heading content: 

> Access Institution Account

- [ ] Button onclick link: https://sims.studentaidbc.ca/institution/login 

**Additional context**

### Sample:

<img width="736" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL0M3REE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cc24fbe633f8ccc6bb90f1e0dddbbc44e5e20f3f/image.png" alt="image.png" />

[...]

PR #6147, opened 2026-05-14
#5684 - Students and Institutions Landing Page

tiago-graf:
### Summary

Adds general landing page targeting students and institutions.

image

image

### DIFF
--- a/sources/packages/web/src/views/LandingPage.vue
+++ b/sources/packages/web/src/views/LandingPage.vue
@@ -1,138 +1,145 @@
 <template>
-  <v-app-bar color="white">
-    <b-c-logo subtitle="Student &amp; Institution Accounts" />
-  </v-app-bar>
-  <v-main>
-    <v-container class="container-max-width mx-auto mt-10">
-      <!-- For Students -->
-      <v-card elevation="2" class="mb-6">
-        <v-card-text>
+  <div>
+    <v-app-bar color="white">
+      <b-c-logo subtitle="Student &amp; Institution Accounts" />
+    </v-app-bar>
+    <v-main>
+      <v-container class="landing-page-container-size mx-auto mt-6">
+        <!-- For Students -->
+        <v-card class="p-4 mb-4">
           <h1
-            class="category-header-large primary-color mb-6 text-center text-md-left"
+            class="category-header-large primary-color mb-4 text-center text-sm-left"
           >
             For Students
           </h1>
 
-          <!-- New Student Account row -->
-          <v-row align="center" class="mb-4">
-            <v-col cols="12" md="2" class="d-flex justify-center">
-              <v-img
-                height="100"
-                width="100"
-                alt="An illustration of a student."
-                src="@/assets/images/image_student_welcome.svg"
-              />
-            </v-col>
-            <v-col cols="12" md="7" class="text-center text-md-left">
-              <p class="category-header-medium mb-2">New Student Account</p>
-              <p class="mb-4">
-                <strong>If applying after June 19, 2026</strong>, use the new
-                Student Account and Application system to create and manage your
-                student aid applications.
-              </p>
-              <div class="d-flex justify-center justify-md-start">
-                <v-btn color="primary" href="/student/login">
-                  Access New Student Account
-                </v-btn>
-              </div>
-            </v-col>
-            <v-col cols="12" md="3" />
-          </v-row>
+          <!-- New Student Account -->
+          <v-card hover>
+            <v-card-text>
+              <v-row align="center">
+                <v-col cols="12" sm="2" class="d-flex justify-center">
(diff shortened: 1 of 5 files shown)


sources/packages/web/src/views/LandingPage.vue
andrewsignori-aot, line 1: Is it possible to follow a bit closer the proposed sample layout?

Image

1. Border around the options
2. Button style for the legacy system
3. Gray shadow, probably when hovering (possibility the out-of-the-box behavior of the card?)
4. Spacing between the item borders and the margin.

andrewsignori-aot, line 1: In an intermittent behavior, but not hard to reproduce after a few attempts, going back using the browser history lands the user on a blank page.

Image

## Steps to reproduce
1. Access the landing page.
2. Navigate to the student or institution login.
3. Use the browser back button to go back to the landing page.
4. Repeat a couple of times.

Tested on Chrome.

andrewsignori-aot, line 1: The view below does not seem to be taking the space nicely. For instance, the icon can still be displayed to the left of the text.

Image

sources/packages/web/src/views/LandingPage.vue
```diff
@@ -0,0 +1,137 @@
...
+                  Access Old Dashboard Account
+                </v-btn>
+              </div>
+            </v-col>
+            <v-col
+              cols="12"
+              md="4"
+              class="d-flex align-center justify-center justify-md-start"
+            >
+              <v-chip
```
weskubo-cgi: I feel like combination of the heavily rounded look of the v-chip plus it's position on the right look slightly odd. A more square "banner" style might look better.

sh16011993 (approved):
Great work @tiago-graf 👍

sources/packages/web/src/views/LandingPage.vue
tiago-graf: Updated, I think the layout is still slightly off (borders/shadows/hover) let me know what you think.

tiago-graf: Also, i moved the alert/chip to the bottom below the text, i now its not the proposed layout but it makes more sense with the mobile first mentality, i can check with josh tuesday!

tiago-graf: i'm trying to reproduce this but still no success

sources/packages/web/src/views/LandingPage.vue
```diff
@@ -0,0 +1,145 @@
...
+        <!-- For Students -->
+        <v-card class="p-4 mb-4">
+          <h1
+            class="category-header-large primary-color mb-4 text-center text-sm-left"
+          >
+            For Students
+          </h1>
+
+          <!-- New Student Account -->
+          <v-card hover>
```
andrewsignori-aot: To better reflect the flat style, I would suggest the following changes.
1. Add elevation as 0 and border as 1 to make it flattttt;
2. Move the `href` from the button to the `v-card`, which would still allow the navigation to happen from button or card click, and also add the "gray on hover" style. This seems aligned with the screenshot provided in the ticket.

```vue
 <!-- New Student Account -->
<v-card outlined border="1" :elevation="0" href="/student/login">        
```

sources/packages/web/src/views/LandingPage.vue
```diff
@@ -0,0 +1,145 @@
...
+            <v-card-text>
+              <v-row align="center">
+                <v-col cols="12" sm="2" class="d-flex justify-center">
+                  <v-img
+                    height="100"
+                    width="100"
```
andrewsignori-aot: Minor, when resizing the screen, the image seems to be getting too small.
Maybe the below can avoid it a bit.
```vue
class="mx-4"
height="150"
width="150"
min-height="150"
min-width="150"
```

tiago-graf: the margin didn't seem to work so I increased the columns from 2 to 3 (desktop version) to allow for more space between the image and text

sources/packages/web/src/views/LandingPage.vue
```diff
@@ -0,0 +1,154 @@
...
+                      prepend-icon="fa:fa fa-circle-exclamation"
+                    >
+                      Not available to submit new applications
+                    </v-chip>
+                  </div>
+                  <div class="d-flex justify-center justify-sm-start">
+                    <v-btn
+                      color="primary"
+                      variant="outlined"
+                      href="https://studentaidbc.ca/dashboard/login"
```
andrewsignori-aot: This can be removed from the `v-btn`. The event should trigger the `href` navigation from the card.

andrewsignori-aot (approved):

PR #6168, opened 2026-05-25
#6016 - Enable toggle for Bulk Withdrawal

sh16011993:
As a part of this PR, the following were completed:

- Enabled feature toggle for the Bulk Withdrawal for the BC Public Institutions (Toggle Env Variable: `BULK_WITHDRAWAL_UPLOAD`) for the sidebar

**Screenshot:**

image

### DIFF
--- a/sources/packages/web/src/composables/useFeatureToggles.ts
+++ b/sources/packages/web/src/composables/useFeatureToggles.ts
@@ -3,7 +3,7 @@ import { AppConfig } from "@/types/contracts/ConfigContract";
 import { computed, ref } from "vue";
 
 const FORMS_SUBMISSION_FEATURE_TOGGLE = "FORMS_SUBMISSION";
-export const BULK_WITHDRAWAL_UPLOAD_FEATURE_TOGGLE = "BULK_WITHDRAWAL_UPLOAD";
+const BULK_WITHDRAWAL_UPLOAD_FEATURE_TOGGLE = "BULK_WITHDRAWAL_UPLOAD";
 
 export function useFeatureToggles() {
   /**
--- a/sources/packages/web/src/composables/useFeatureToggles.ts
+++ b/sources/packages/web/src/composables/useFeatureToggles.ts
@@ -3,6 +3,7 @@ import { AppConfig } from "@/types/contracts/ConfigContract";
 import { computed, ref } from "vue";
 
 const FORMS_SUBMISSION_FEATURE_TOGGLE = "FORMS_SUBMISSION";
+export const BULK_WITHDRAWAL_UPLOAD_FEATURE_TOGGLE = "BULK_WITHDRAWAL_UPLOAD";
 
 export function useFeatureToggles() {
   /**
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/composables/useFeatureToggles.ts
```diff
@@ -3,6 +3,7 @@ import { AppConfig } from "@/types/contracts/ConfigContract";
 import { computed, ref } from "vue";
 
 const FORMS_SUBMISSION_FEATURE_TOGGLE = "FORMS_SUBMISSION";
+export const BULK_WITHDRAWAL_UPLOAD_FEATURE_TOGGLE = "BULK_WITHDRAWAL_UPLOAD";
```
dheepak-aot: Is there a need to expose the const?

sources/packages/web/src/router/InstitutionRouteHelper.ts
```diff
@@ -48,6 +52,24 @@ export async function validateInstitutionUserAccess(
...
+  to: RouteLocationNormalized,
+): Promise<boolean> {
+  if (to.name !== InstitutionRoutesConst.WITHDRAWAL_UPLOAD) {
+    return true;
+  }
+
+  const appConfig = await AppConfigService.shared.config();
+  return appConfig.featureToggles.includes(
+    BULK_WITHDRAWAL_UPLOAD_FEATURE_TOGGLE,
+  );
```
dheepak-aot: Is there a reason for not using the composable `useFeatureToggles` here?. Relevant only if we need to keep this logic.

sources/packages/web/src/router/InstitutionRouteHelper.ts
```diff
@@ -48,6 +52,24 @@ export async function validateInstitutionUserAccess(
   }
 }
 
+/**
+ * Validates access based on route-specific feature toggles.
+ * @param to the route where user is navigated to.
+ * @returns true when route is allowed by feature toggle rules.
+ */
+async function isInstitutionRouteAllowed(
```
dheepak-aot: Is it required to block the route? I am asking this question as Offerings Upload and Reports aren't blocked based on user institution being BCPublic.

sources/packages/web/src/composables/useFeatureToggles.ts
sh16011993: Yes, we are importing it in `InstitutionRouteHelper`.

sources/packages/web/src/router/InstitutionRouteHelper.ts
sh16011993: `useFeatureToggles()` loads the app config asynchronously (void loadConfig()), so isFeatureToggleEnabled(...) will return false on this synchronous execution (the local appConfig ref is still null).

sources/packages/web/src/router/InstitutionRouteHelper.ts
sh16011993: I believe so but I am not entirely sure on this.

sources/packages/web/src/router/InstitutionRouteHelper.ts
dheepak-aot: Sorry I am not able to follow why the reason stated here is applicable only this situation?

dheepak-aot (approved):
Looks good @sh16011993. Nice work.

weskubo-cgi (approved):
👍

PR #6178, opened 2026-05-27
#6091 - Postgres upgrade task

sh16011993:
As a part of this PR, the following was fixed:

- Postgres volume originally mounted at `/var/lib/postgresql/data` is incompatible with the Postgres 18 image. Changed it to `postgres:/var/lib/postgresql` for the upgrade to postgres 18.


andrewsignori-aot (approved):
👍

dheepak-aot (approved):

Issue #6187, opened 2026-05-30
BUG: DEV only: "I confirm" checkbox for exceptions, questions not reset


astridSABC:
**Describe the Bug**
This might be a FORM IO issue

Additional transportation
When "No" is selected after the "please confirm" checkbox is selected, the question does not reset.

Review: All areas that have the "I confirm" checkbox

e.g. Married/common-law and unable to provide partner income due to domestic abuse 
Select "I confirm checkbox", note the file uploaded
Select "single"
File uploader remains on the page

DEV Only
Test ok
Staging ok

**Expected Behaviour**
Additional transportation section with initial question "Do you want to be assessed for additional transportation costs?" is reset to "Yes"/"No" options when "No" is selected

**Actual Behaviour**
The question "Are you listed as a drive on the insurance for the vehicle you will be using? and the  "please confirm" banner with the checkbox are removed but the other questions related to the additional transportation exception remain

<img width="1330" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3dQRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5c6cefb0448b9b104a8c668012a116fe0237b3f5/image.png" alt="image.png" />

Select "No"

<img width="898" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL0VQRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--157dd5765581ed6f81bb2c96c57590fae1abd69e/image.png" alt="image.png" />

### Steps To Reproduce

Student can create an application (any PY any intensity)

Steps to reproduce the behaviour:
1. Fill out a happy path application to the financial information tab
2. Enter and select valid data to the additional transportation section
3. Select "Yes" to be assessed for additional transportation costs
4. Select "Yes" for listed as a driver on the insurance
5. Select  "I confirm I am eligible for this exception"
6. Select  "Yes" for are you the owner of the vehicle
7. Select "No" for do you want to be assessed (because you understand that you don't qualify)
8. Repeat with all the exceptions that have the "I confirm" checkbox

QA
- Make sure that question is reset

dheepak-aot:
@astridSABC I tested the following scenarios in local as there was a client only formio library upgrade. 

PT 2025-26 Scenarios
[...]

Issue #6163, opened 2026-05-22
BUG: Form IO: Status badges display vertically instead of horizontally


astridSABC:
**Describe the Bug**
Badge displays in a vertical format instead of horizontal
Banners are centered (terminology needed for different types of banners)
PT 25/26 Only 2nd dependant with an exception, file upload link is not available

"PD Dependent" (see the screenshot below) if minimal fix: please change to "Dependant"

**Expected Behaviour**
Badges to be displayed horizontally and clearly beside/under headings/labels
Banner content is left justified
Exception files are links

<img width="1511" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNG4rRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e44e72a95af804884ae7a949066f340e2f47935a/image.png" alt="image.png" />

<img width="751" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemdJRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--abedb66a204bc60fc6d43df5bb18bad6030bb264/image.png" alt="image.png" />

<img width="864" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeThZRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ad1dcdb5b48ddf9b18064588e39ffcc8a43af8e5/image.png" alt="image.png" />

**Actual Behaviour**

<img width="707" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMzcrRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6cf0017ecc078df3041ffe2524c8b7da91d45bc4/image.png" alt="image.png" />

[...]

PR #6192, opened 2026-06-02
#6187 - Fix Additional Transportation Question not Updating Mandatory Questions

tiago-graf:
### Summary

Update form definitions to force redrawing the transportation confirmation checkbox which was causing the form to be in a invalid state.


dheepak-aot:
The PR looks good. I am looking if there is a broader fix for all such check boxes as an outcome of this ticket and #6163 as well. If that is not an option, I am good with the PR.

tiago-graf:
Closing this as it seems the fix won't actually work

PR #6204, opened 2026-06-03
 #6163 - Form.io upgrade UI issues

weskubo-cgi:
## Summary
The following issues were fixed:
#### Vertical Chip
The chip issue is due to `width: var(--fa-width, 1.25em)` on the `fa` class that was added in FontAwesome 7. An existing style was updated to add a width.
 
#### Banner alignment
The chip issue is due to `text-align:center` on the `fa` class that was added in FontAwesome 7.  This was fixed via an update to an existing style (for warnings), adding a missing class (for nonpunitivewithdrawalform.json), and via a form.io override (for limited full-time banners).

#### Bullet alignment
UL tags were missing for various banners causing the bullet indentation to be off.

#### Missing document links
There were extra conditionals on the **PD Dependent** and **Dependant custody file upload** components which don't work in a dataGrid. They are redundant and were removed.

### Ministry - Pending Offerings
image

### Student - Non-punitive withdrawal form
image

### Student - Modified independent appeal
image

### Student - FT 2026-2027 application
image
image
image

### Student - PT 2025-2026 application
image

### Student - PT 2026-2027 application
image

### DIFF
--- a/sources/packages/forms/src/form-definitions/nonpunitivewithdrawalform.json
+++ b/sources/packages/forms/src/form-definitions/nonpunitivewithdrawalform.json
@@ -1,131 +1,132 @@
 {
-    "title": "Non-punitive Withdrawal",
-    "name": "nonpunitivewithdrawalform",
-    "path": "nonpunitivewithdrawalform",
-    "type": "form",
-    "display": "form",
-    "tags": [
-        "common"
-    ],
-    "components": [
+  "title": "Non-punitive Withdrawal",
+  "name": "nonpunitivewithdrawalform",
+  "path": "nonpunitivewithdrawalform",
+  "type": "form",
+  "display": "form",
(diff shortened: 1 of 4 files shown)


sources/packages/forms/src/form-definitions/nonpunitivewithdrawalform.json
andrewsignori-aot, line 1: Is the amount of changes in the diff due to a possible difference in the local JSON formatter being used?
In local, using the workspace, it looks the same as the modified file in this PR. Would it mean that @tiago-graf is possibly using a different formatter?

andrewsignori-aot (approved):
Nice findings and fixes, looks good 👍

sh16011993 (approved):
Nice work @weskubo-cgi 👍

Issue #6075, opened 2026-04-23
Mock API ATBC 


ninosamson:
- ticket for mock external API for ATBC to retrieve SABC PD info
- Two endpoints  
    - Get students 
    - Get PD info

ninosamson:
Updated AC - this endpoint will return  active student disability profiles over a given time frame.

dheepak-aot:
@ninosamson Here are the API details for the new external API created for disability profile. 

https://dev.sims.studentaidbc.ca/external/swagger#/external-disability-profile/DisabilityProfileExternalController_getAllDisabilityProfiles

dheepak-aot:
## Toggle

To disable the feature in production, following strategies can be followed.

- Disable the client in keycloak 
<img width="1589" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNmt3RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b9f86d6a85b2ef950a5bffa08326fb90ae3497f2/image.png" alt="image.png" />

- Not share the credentials to the API consumer
- If there is an active whitelist IPs. Not add API consumer IP to the whitelist.

PR #6210, opened 2026-06-05
#6075 - Disability profile API

dheepak-aot:
# Disability profile API

## Introducing keycloak role based access for external clients

- [x] Added a new enum `ExternalRole` to have all the roles used for external clients.
- [x] The property role is currently typed with only ministry roles i.e. `Roles`, hence created a generic type `ClientRole` that represents any client role.
- [x] For the ministry APIs, created a specific user token type `MinistryUserToken` which has the property roles with specific ministry role enum `Role`.

## External API for disability profiles
- [x] Created an external API to return all active disability profiles modified since the provided `modifiedSince` date.
- [x] `modifiedSince` is validated to be not older than 180 days and also not in future or current timestamp.

## Updated SDPR APIs to be authorized based on roles
- [x] Updated SPDR APIs with it's own specific external role for authorization.

## E2E Tests

- [x] Added E2E tests for the new disability profiles API. 
```spec
DisabilityProfileExternalController(e2e)-getAllDisabilityProfiles
    √ Should return all active student disability profiles when the profile is created or edited since the given modifiedSince date and the student has a valid SIN. (162 ms)
    √ Should throw a bad request error when the modifiedSince query parameter is missing. (12 ms)
    √ Should throw a bad request error when the modifiedSince query parameter is not within the last 180 days (9 ms)
```
- [x] Updated `getExternalToken()` to use dedicated e2e test client for external `e2e-external`.

### Happy path

image

### External client missing role
image

image


sources/packages/backend/apps/api/src/route-controllers/disability-profile/disability-profile.external.controller.ts
```diff
@@ -0,0 +1,98 @@
...
+import { DisabilityProfileService } from "../../services";
+import { ClientTypeBaseRoute } from "../../types";
+import BaseController from "../BaseController";
+import {
+  DisabilityProfileExternalAPIOutDTO,
+  DisabilityProfilesExternalAPIOutDTO,
+  DisabilityProfilesQueryExternalAPIInDTO,
+} from "./models/disability-profile.dto";
+import { ExternalRole } from "../../auth";
+import { addDays, isAfter, isBeforeDate } from "@sims/utilities";
```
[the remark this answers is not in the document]
dheepak-aot: I can use the `isSameOrAfterDate` for min modified since validation but if the `modifiedSince = modifiedUntil` then database should not be called to make a query that will not return results

sources/packages/backend/apps/api/src/testHelpers/auth/external-token-helpers.ts
```diff
@@ -1,13 +1,13 @@
...
  * @returns an external user token to be used in the tests
  */
 export async function getExternalUserToken(): Promise<string> {
-  const clientSecret = process.env.E2E_TEST_EXTERNAL_USER_CLIENT_SECRET;
-  return getCachedToken(AuthorizedParties.external, {
+  const E2E_TEST_EXTERNAL_USER_CLIENT_ID = "e2e-external";
+  const clientSecret = process.env.E2E_TEST_EXTERNAL_USER_CLIENT_SECRET!;
+  return getCachedToken(E2E_TEST_EXTERNAL_USER_CLIENT_ID, {
     clientSecret,
   });
```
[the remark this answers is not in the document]
dheepak-aot: This is not an issue as all the `*external` clients are forced to set the azp as external.

sources/packages/backend/apps/api/src/auth/userToken.interface.ts
```diff
@@ -1,21 +1,21 @@
...
+  roles: ClientRole[];
 }
 
 /**
  * Resource Access extracted from the token during the
  * authentication process on JwtStrategy validate method.
  */
 export interface ResourceAccess {
-  aest: Roles;
+  [key: string]: ClientRoles;
```
andrewsignori-aot: The type is changed from the specific `aest` to any string. Can it be declared as below?
```ts
export type ResourceAccess = Partial<Record<AuthorizedParties, ClientRoles>>;
```

sources/packages/backend/apps/api/src/route-controllers/disability-profile/models/disability-profile.dto.ts
```diff
@@ -69,6 +71,44 @@ export class StudentDisabilityProfilesAPIOutDTO {
...
+  impairmentsNotes?: string;
+  finalNotes?: string;
+}
+
+export class DisabilityProfileExternalAPIOutDTO {
+  firstName?: string;
+  lastName: string;
+  sin: string;
+  address: AddressAPIOutDTO;
+  disabilities: DisabilityExternalAPIOutDTO[];
```
andrewsignori-aot, line 105: Is there an assumption that the consumer will receive the records in order and will be aware that the order represents the disability's priority, or would it woud be better to also return the order?
Without the order being returned, and since this is for external API consumers, more comments can be added to the DTO.

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
```diff
@@ -479,4 +479,55 @@ export class DisabilityProfileService {
...
+          impairments: true,
+          impairmentsNotes: true,
+          finalNotes: true,
+        },
+      },
+      where: {
+        disabilityProfileStatus: DisabilityProfileStatus.Active,
+        completedAt: Raw(
+          (completedAt) =>
+            `${completedAt} > :modifiedSince AND ${completedAt} <= :modifiedUntil`,
```
andrewsignori-aot: Please expand the comments (in the endpoint for Swagger) to include the expected period behavior "start date excluded and end date included".

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
```diff
@@ -479,4 +479,55 @@ export class DisabilityProfileService {
...
+          diagnosis: true,
+          diagnosisNotes: true,
+          impairments: true,
+          impairmentsNotes: true,
+          finalNotes: true,
+        },
+      },
+      where: {
+        disabilityProfileStatus: DisabilityProfileStatus.Active,
+        completedAt: Raw(
```
andrewsignori-aot: We should avoid using `Raw` when there are alternatives.
I believe the below would work. For this specific query, the `And` would also be optional.
```ts
completedAt: And(
  MoreThan(modifiedSince),
  LessThanOrEqual(modifiedUntil),
),
```

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
```diff
@@ -479,4 +479,55 @@ export class DisabilityProfileService {
...
+          (completedAt) =>
+            `${completedAt} > :modifiedSince AND ${completedAt} <= :modifiedUntil`,
+          { modifiedSince, modifiedUntil },
+        ),
+        student: { sinValidation: { isValidSIN: true } },
+      },
+      relations: {
```
andrewsignori-aot: In general, I believe we have been using the `relations` before the `where,` and it would be good to have it adjusted.

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
```diff
@@ -479,4 +479,55 @@ export class DisabilityProfileService {
...
+    return this.studentDisabilityProfileRepo.find({
+      select: {
+        id: true,
+        student: {
+          id: true,
+          user: { id: true, firstName: true, lastName: true },
+          sinValidation: { id: true, sin: true },
+          contactInfo: true,
+        },
+        completedAt: true,
```
andrewsignori-aot: `completedAt` seems to be used in the `where` but not by any consumer, hence it can be removed.

sources/packages/backend/apps/api/src/auth/roles.enum.ts
```diff
@@ -50,3 +50,17 @@ export enum Role {
   StudentDeleteRestriction = "student-delete-restriction",
   StudentEditDisabilityProfile = "student-edit-disability-profile",
 }
+
+/**
+ * External user roles to control the access of external API endpoints.
+ */
+export enum ExternalRole {
+  StudentActiveSins = "student-active-sins",
```
andrewsignori-aot: Should it be named 
`StudentActiveSINs ` or
`StudentActiveSocialInsuranceNumbers`?

sources/packages/backend/apps/api/src/route-controllers/disability-profile/_tests_/e2e/disability-profile.external.controller.getAllDisabilityProfiles.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+          "modifiedSince should not be empty",
+        ],
+        error: "Bad Request",
+        statusCode: HttpStatus.BAD_REQUEST,
+      });
+  });
+
+  it(
+    "Should throw a bad request error when the modifiedSince query parameter is not within" +
+      ` the last ${DISABILITY_PROFILES_MODIFIED_SINCE_MAX_DAYS} days`,
```
andrewsignori-aot: Period is missing.

andrewsignori-aot (commented):

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
sh16011993: For my understanding, why are we using the `modifiedUntil` date since I see that it is just set to current date? Can't we just skip this check?

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
```diff
@@ -479,4 +479,55 @@ export class DisabilityProfileService {
...
+          { modifiedSince, modifiedUntil },
+        ),
+        student: { sinValidation: { isValidSIN: true } },
+      },
+      relations: {
+        student: { user: true, sinValidation: true },
+        disabilities: true,
+      },
+      order: {
+        completedAt: "ASC",
```
sh16011993, line 533: Shouldn't we keep this as `DESC` priority to get the latest added disability at the top of the returned list?

sources/packages/backend/apps/api/src/auth/userToken.interface.ts
```diff
@@ -1,21 +1,21 @@
...
-  roles: string[];
+export interface ClientRoles {
+  roles: ClientRole[];
 }
 
 /**
  * Resource Access extracted from the token during the
  * authentication process on JwtStrategy validate method.
  */
 export interface ResourceAccess {
```
sh16011993: Can we also update the comment where this interface is referenced?

sh16011993 (commented):
Primo work @dheepak-aot 👍 

Thanks for the clarification related to keycloak (clients and client scopes)

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
dheepak-aot: There is no such requirement to return recently modified profiles first, so keeping the natural ascending order.

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
dheepak-aot: Removed.

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
dheepak-aot: Moved.

sources/packages/backend/apps/api/src/services/disability-profile/disability-profile.service.ts
dheepak-aot: Comments updated. Swagger view looks like this. 

image

sources/packages/backend/apps/api/src/route-controllers/disability-profile/models/disability-profile.dto.ts
```diff
@@ -7,6 +7,7 @@ import {
   IMPAIRMENTS_NOTES_MAX_LENGTH,
   LOOKUP_KEY_MAX_LENGTH,
 } from "@sims/sims-db";
+import { AddressAPIOutDTO } from "apps/api/src/route-controllers/models/common.dto";
```
andrewsignori-aot: Please review the import to have the `apps/api/src` removed.

andrewsignori-aot (approved):
Great work, thanks for making the changes, looks good 👍

sources/packages/backend/apps/api/src/auth/userToken.interface.ts
dheepak-aot: Comment at reference updated.

Issue #6191, opened 2026-06-02
Upgrade TypeORM


ninosamson:
**Acceptance criteria**
- [ ] upgrade typeORM to latest version

PR #6213, opened 2026-06-08
#6191 - Upgrade Typeorm

sh16011993:
**As a part of this PR, the backend ORM dependency was upgraded to TypeORM 0.3.30**

Changes include:

- Bumped `typeorm` dependency from `^0.3.28` to `^0.3.30` (and refreshed `package-lock.json` accordingly).
- Updated report service to inject/use `DataSource` instead of the older and soon to be deprecated `Connection`.
- Updated the typeorm method from `exist` to `exists` which is more modern from upgrade perspective and has an identical behavior.
- Adjusted db-migrations Postgres config typing to use TypeORM’s public DataSourceOptions-based types.

**Note: At this time, upgrade to version `1.0.0` isn't possible because `typeorm` is a dependency for the package: `@nestjs/typeorm@11.0.1` and this package currently doesn't support typeorm version 1.0.0**


sources/packages/backend/package-lock.json
```diff
@@ -5547,9 +5547,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
```
andrewsignori-aot, line 5550: Just to double-check, do you have the most recent version of Nodejs updated locally? Please ensure `npm` is updated to `11.16.0` (Nodejs v26.3.0).

andrewsignori-aot (commented):
Thanks for putting the effort into looking into the version [1.0.0](https://www.npmjs.com/package/typeorm/v/1.0.0). As agreed, let's hold till `@nestjs/typeorm` supports the new version.

andrewsignori-aot (approved):
Looks good 👍

weskubo-cgi (approved):
Thanks for the updates and thorough analysis. Great work.

PR #6214, opened 2026-06-08
#6187 - Formio selected location program and the CSP fix

dheepak-aot:
### Root cause
With the recent version(s) of form.io, if a input component is hidden, then the component value is being reset to `undefined` even if the component value is supplied in the form-data(initial data). 

For e.g. In student application,  `selectedLocation` is the key for the component that is used to choose the institution location. 
image

And during the read-only view, we display a read-only input text component(No 2 in screenshot) having the location name value and hide the select component(No 1 in screenshot).
image

During such a situation where the read-only view is displayed, `data.selectedLocation` is reset to `undefined` even if the intital data has a value to it, say `{selectedLocation: 1}`. 

### Solution
Added a calculated variable which also considers the read-only variables supplied to identify if the functional value of the component was set or not. 
e.g. hidden variable name `hasSelectedLocation` with value as `!!data.selectedLocation || !!(data.selectedLocationName?.trim());` 

## CSP issue

Updated CSP to allow the content from the blocked url. 

image

### After fix
image

### Program page with selected program and offering
image

### Program page with PIR(Program present offering not present)
image

### Program page with PIR(Program and offering not present)
image

## Additional transport banner fix for PY2025-26 FT
Fix related to the ticket #6163 
image

### DIFF
--- a/sources/packages/backend/package-lock.json
+++ b/sources/packages/backend/package-lock.json
@@ -5547,9 +5547,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -5564,9 +5561,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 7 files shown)


sh16011993 (approved):
Great work @dheepak-aot 👍 
❤️'ed the explanation for the PR description.

andrewsignori-aot (approved):
Great work and fixes, looks good 👍

Issue #3844, opened 2024-10-28
PIR - Create an Offering


CarlyCotton:

astridSABC:
@astridSABC If content required above or below the new "Create offering" button, please provide in a separate enhancement ticket.

PR #6231, opened 2026-06-11
#3844 - Add Create Offering to PIR

weskubo-cgi:
## Summary
- Add button to PIR to allow creation of missing offering.
- Type cleanup in `LocationEditProgramInfoRequest.vue`.

### Institution -> Program Offering Request
image

### DIFF
--- a/sources/packages/web/src/services/http/dto/ProgramInfoRequest.dto.ts
+++ b/sources/packages/web/src/services/http/dto/ProgramInfoRequest.dto.ts
@@ -65,14 +65,3 @@ export class DenyProgramInfoRequestAPIInDTO {
   @Expose()
   otherReasonDesc?: string;
 }
-
-/**
- * Form data for Program Information Request. This is limited to user entered fields.
- */
-export interface ProgramInfoRequestFormData {
-  denyProgramInformationRequest: boolean;
-  selectedProgram?: number;
-  selectedOffering?: number;
-  pirDenyReasonId?: number;
-  otherReasonDesc?: string;
-}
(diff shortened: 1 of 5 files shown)


sources/packages/web/src/services/http/dto/ProgramInfoRequest.dto.ts
```diff
@@ -65,3 +65,14 @@ export class DenyProgramInfoRequestAPIInDTO {
   @Expose()
   otherReasonDesc?: string;
 }
+
+/**
+ * Form data for Program Information Request. This is limited to user entered fields.
+ */
+export interface ProgramInfoRequestFormData {
```
weskubo-cgi: Technically this interface only needs `denyProgramInformationRequest` for how it's used in `LocationEditProgramInfoRequest.vue` but I felt it would be better to have a complete set of user entered form data.

Also, this was considered:
```
export type ProgramInfoRequestFormData = CompleteProgramInfoRequestAPIInDTO &
  DenyProgramInfoRequestAPIInDTO & {
    denyProgramInformationRequest: boolean;
    selectedProgram: number;
  };
```
But the optionality isn't correct.

andrewsignori-aot: Non-DTO types should be moved to contracts; the `dto.ts` file must be, as much as possible, a mirror of the API side.

andrewsignori-aot (commented):
Nice work, minor comment only.

sources/packages/web/src/services/http/dto/ProgramInfoRequest.dto.ts
weskubo-cgi: Moved to existing contract `types\contracts\institution\ProgramInfoRequest.ts`.

sh16011993 (approved):
Nice work @weskubo-cgi 👍

andrewsignori-aot (approved):
Looks good 👍

Issue #5969, opened 2026-03-27
PIR outcome visibility


Joshua-Lakusta:
### User Story

As a student or ministry user, I want to view the outcome of a completed Program Information Request (PIR) within the associated application, so that I can understand what program and offering details were provided by the institution.

### Basic Criteria

- [ ] This change should be enabled for all versions and should be a functional change that is backwards compatible (if possible)
- [ ] Show PIR outcome in a new container of information on the read only view and edits of the application
- [ ] PIR outcome details are only available when a PIR is Completed 
- [ ] PIR is related to its version 
- [ ] Display the PIR from the completion of the PIR, from the original assessment
- [ ] Display is always read-only and cannot be edited by the student
- [ ] If no PIR was ever completed do not show this field
- [ ] If PIR is in progress do not show the new field

- [ ] Still display this for change request

- [ ] Information is visible to:

- Student
- Ministry users 
- Institution users that can view the application (BC publics)

- [ ] New field should show at the bottom of the Program and offering info container right above where the retrigger pir bubble is (see images below)

- [ ] The content shown should be the following

> **Program information request summary**

>  Program name:
>  Offering name: _(Use the version displayed in the dropdown w/name, year of study and dates)_

<img width="783" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeVdsRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6b005a8cf7e00cc0d7d7af6109c41453f2a5a8a1/image.png" alt="image.png" />

PR #6260, opened 2026-06-16
#5969 - Update PIR outcome Styling

tiago-graf:
### Summary

Update form definitions to use a header styling for the PIR outcome panel.

image


weskubo-cgi (approved):
Thanks for the fix.

dheepak-aot (approved):
👍

Issue #6261, opened 2026-06-16
BUG: INS: Offerings: "Flashing numbers" in "Total break days" and "Total eligible break days"


astridSABC:
**Describe the Bug**
As an institution user, I observe, Numbers flashing (or do a random count from 0 to 10) in the "Total break days" and "Total eligible break days".

Offerings can be created

Dev, Test, Staging

**Expected Behaviour**
Calculated numbers in the "Total break days" , "Total eligible break days" fields are static.

**Actual Behaviour**
Numbers flash (or do a random count from 0 to 10) in the "Total break days" and "Total eligible break days"

<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNWc2RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--76abd27c4f53a12324eeaac739acb1a052f8a47d/image.png" alt="image.png" />

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjA2RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c53c7d78e10351941bcbc23e9c8ca1702c30a4d5/UI_Offering_Staging.mp4)

### Steps To Reproduce

PRECONDITION:
Program has been created that allows for FT/PT, blended
User is logged into the Institution account and on the Program Detail page

1. Click "Add offering" in the right corner
2. Enter any name
3. Select any year of study
4. Select "Full-time" for will this be offered
5. Select "Onsite"
6. Select "No" for aviation
7. Select "No" for WI;l
8. Enter study start date: June 7, 2026
9. Enter end date November 27, 2026
10. Enter break start date August 19, 2026
11. Enter break end date August 28, 2026
12. Note the "Total break days" field
13. Note the "Total eligible break days" field
14. Enter all other information
15. Click "Validate" button (note the field validation)

andrewsignori-aot:
@astridSABC @Joshua-Lakusta is there any issue with initializing the study breaks as below instead of with empty controls? The main difference the user will need to click on the button to start adding the values.

[...]

PR #6263, opened 2026-06-16
#6261 - BUG: INS: Offerings: "Flashing numbers" in "Total break days" and "Total eligible break days"

andrewsignori-aot:
## Issue 1: "Flashing numbers" 

- The below highlighted components were calculating their values based on the `Break start date` and `Break end date` and also on each other's output, causing an infinite loop of changed values events.
- This behavior is present in the [@formio/js@5.4.0](https://www.npmjs.com/package/@formio/js/v/5.4.0) and there is no reason to believe that it is the expected behavior (not fixed by its newly created 5.4.1 version).

### Proposed solution

- A hidden field was introduced to have the three values calculated and displayed in an HTML `p` component that emulates the visual of an input.
- A validation for the minimum amount of break days, which was previously in the `Total break days` column, was moved to the `Break end date` column.
- The logic for the validations was refactored to use an "early return" pattern due to the hard-to-follow nested `if` conditions.

image

image

## Issue 2: Cannot read properties of undefined (reading 'indexOf')

- The `validationWarnings` and `validationInfos` hidden fields were accessed before they were set, causing the `indexOf` method to be called from an `undefined` object.
- Moving the hidden field to the top resolved the issue (as it had resolved in some past experiences).

## Related validations

image


sources/packages/forms/src/form-definitions/educationprogramoffering.json
```diff
@@ -8,6 +8,22 @@
     "common"
   ],
   "components": [
+    {
```
andrewsignori-aot, line 11: These hidden fields were just moved to the top to prevent the `indexOf` error, as explained in the PR description.

sources/packages/forms/src/form-definitions/educationprogramoffering.json
```diff
@@ -1141,67 +1156,66 @@
...
+                      "key": "eligibleDays1",
+                      "type": "htmlelement",
+                      "input": false,
+                      "tableView": false
+                    },
+                    {
+                      "label": "study Breaks Summary",
+                      "persistent": "client-only",
+                      "calculateValue": "// Total days.\r\nlet totalDays = 0;\r\nif (row.breakStartDate && row.breakEndDate) {\r\n  const diffDays = moment(row.breakEndDate, \"YYYY-MM-DD\").diff(\r\n    moment(row.breakStartDate, \"YYYY-MM-DD\"),\r\n    \"days\"\r\n  ) + 1 ;\r\n  totalDays = Math.max(diffDays, 0);\r\n}\r\n// Maximum of 21 days per break are eligible.\r\n// Any days over 21 consecutive days must be unfunded\r\nconst eligibleDays = (totalDays > 21) ? 21 : totalDays;\r\n// Breaks days (each) over 21 consecutive days must be unfunded.\r\nconst ineligibleDays = totalDays - eligibleDays;\r\n\r\nvalue = { totalDays, eligibleDays, ineligibleDays }\r\n\r\n",
+                      "key": "summary",
```
andrewsignori-aot: Tried to use only `summary` as the hidden field name in an attempt (not so successful) to prevent the designer from being too clunky.

Image

sources/packages/forms/src/form-definitions/educationprogramoffering.json
```diff
@@ -1779,38 +1793,6 @@
...
-      "tableView": false
-    },
-    {
-      "label": "applicationStatus",
-      "persistent": "client-only",
-      "key": "applicationStatus",
-      "type": "hidden",
-      "input": true,
-      "tableView": false
-    },
```
andrewsignori-aot, line 1813: Legacy hidden fields from a previous framework that are no longer required.

sources/packages/forms/src/form-definitions/educationprogramoffering.json
```diff
@@ -1122,11 +1135,13 @@
...
...                  "custom": "// Validation logic.\r\nif(!input){\r\n  // The required validation will be handled by the form field properties, so we can return true here.\r\n  return true;\r\n}\r\n\r\nif (!data.studyStartDate || !data.studyEndDate) {\r\n  valid = \"Study start date and study end date can't be empty.\";\r\n  return valid;\r\n}\r\n\r\nif (!row.breakStartDate) {\r\n  valid = \"Please provide break start date\";\r\n  return valid;\r\n}\r\n\r\nif (input < data.studyStartDate || input > data.studyEndDate) {\r\n  valid = \"Break end date should be in between study start date and study end date.\";\r\n  return valid;\r\n}\r\n\r\nconst startDate = moment(row.breakStartDate, \"YYYY-MM-DD\");\r\nconst endDate = moment(input, \"YYYY-MM-DD\");\r\nif (endDate.isBefore(startDate)) {\r\n  valid = \"Break end date must be after break start date\";\r\n  return valid;\r\n}\r\n\r\nconst totalDays = endDate.diff(startDate, \"days\") + 1 ;\r\nif(totalDays < 6){\r\n  valid = \"Breaks must be a minimum of 6 days.\";\r\n  return valid;\r\n}\r\n\r\nif (data.studyBreaksOverlappingDetection) {\r\n  valid = data.studyBreaksOverlappingDetection;\r\n  return valid;\r\n}\r\n\r\nvalid = true;\r\n"
```
dheepak-aot, line 1142: 👍

dheepak-aot, line 1142: As per the co-pilot, on the early return is it not expected to set `valid = true` instead of return?

dheepak-aot (approved):
Good work and great research. Looks good 👍

sh16011993 (approved):

Thanks for the call.

Issue #6240, opened 2026-06-13
BUG: 3.5 NPWD: Appeal cannot submit: Error message displays for app selected in dropdown list


astridSABC:
**Describe the Bug**
NPWD appeal cannot be submitted. Error message: "The selected item is not present in the list of items." Application number is dropdown list is not recognized as a qualifying application for NPWD

<img width="1244" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelkwRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--25714fd1c48c6098c4d7cf7bc46ee633284d7b70/image.png" alt="image.png" />

**Expected Behaviour**
Qualifying applications in the dropdown list are available and can be submitted without any errors

**Actual Behaviour**
Error message: "The selected item is not present in the list of items." displays under the Application Number dropdown list.
You cannot select the application and cannot submit an appeal

Dev Tool Snippet
<img width="533" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemcwRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b7fcd4d87ef8fedbdea527c1921527a784c646f0/image.png" alt="image.png" />

Ministry user:
App: 2025001748 (was previously approved in previous release)

### Steps To Reproduce

FT student application was submitted, assessment accepted, enrolment confirmed, full time-e-cert funds disbursed, withdrew the application (< 10 days) in the study period to qualify for NPWD. Record the application number (e.g. 2025001923)

E.g. Study start date: May 4 2026 - Aug 21 2026
Withdrawal date: May 11 2026

Steps to reproduce the behaviour:
1. Login to the student portal for the student with the application above
2. Select "Applications" in the top right pane
3. Note the application that qualifies for the NPWD
4. Select "Forms" in the top right of the page
5. Select "Non-punitive withdrawal" check box
6. Click "Fill form" button
7. Note the "Application Number" dropdown list (No application number is displayed)
8. Select the dropdown list for applicable applications that qualify (only qualifying apps display)
9. Select application number

Video displays a previous NPWD submitted and approved, then it shows the new item

[...]

PR #6264, opened 2026-06-16
#6240 - Non punitive withdrawal formio fix

dheepak-aot:
## Root cause

The custom validation executed for the select component `nonPunitiveWithdrawalId` was always evaluating to false as the script variable `value` is being returned as undefined. 

```ts
// Ensure the selected value is one of the available ones to be selected.
const isAvailableItem = data.scholasticStandingWithdrawals.filter(item => item.scholasticStandingId === value).length === 1;
valid = isAvailableItem || "The selected item is not present in the list of available items."
```

## Solution

Using script variable `input` instead of `value` returns the correct user input. 

image

The script was also formatted and array safe check added. 

```ts
// Ensure the selected value is one of the available ones to be selected.
const isAvailableItem =
  Array.isArray(data.scholasticStandingWithdrawals) &&
  data.scholasticStandingWithdrawals.filter(
    (item) => item.scholasticStandingId === input
  ).length === 1;
valid =
  isAvailableItem ||
  "The selected item is not present in the list of available items.";
```

### Submission
image

### Read only
image

image

### DIFF
--- a/sources/packages/forms/src/form-definitions/nonpunitivewithdrawalform.json
+++ b/sources/packages/forms/src/form-definitions/nonpunitivewithdrawalform.json
@@ -58,7 +58,7 @@
       "customOptions": "",
       "validate": {
         "required": true,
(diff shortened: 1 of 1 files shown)


andrewsignori-aot (approved):
👍

sh16011993 (approved):
Nice catch @dheepak-aot 👍

PR #6266, opened 2026-06-16
#4199 - Prune Stale Image Tags

tiago-graf:
### Summary

Updates prune images workflow to use new typescript app.
Updates prune process to include deletion of feature branch images (manual builds)

### Manual Test / Dry Run Result

image


.github/workflows/prune-images.yml
```diff
@@ -58,34 +61,21 @@ jobs:
           echo "Jobs: ${OCJOBS}"
           echo "Prefix: ${PREFIX}"
           echo "Minimum Tags: ${MIN_TAGS}"
-          echo "OC CLI Version: $(oc version)"
-
-      - name: Checkout source code
-        uses: actions/checkout@v6
+          echo "Node Version: $(node --version)"
```
andrewsignori-aot: It is not printed in other similar actions using `node-version` and I do not believe it is needed here.

.github/workflows/prune-images.yml
```diff
@@ -58,34 +61,21 @@ jobs:
...
+        run: npm run prune
+        working-directory: devops/image-pruner
+        env:
+          SA_TOKEN: ${{ secrets.SA_TOKEN }}
+          OPENSHIFT_URL: ${{ vars.OPENSHIFT_CLUSTER_URL }}
+          LICENSE_PLATE: ${{ vars.OPENSHIFT_LICENSE_PLATE }}
+          ENVIRONMENT: ${{ env.ENVIRONMENT }}
+          APPLICATIONS: ${{ env.APPLICATIONS }}
+          OCJOBS: ${{ env.OCJOBS }}
+          PREFIX: ${{ env.PREFIX }}
```
andrewsignori-aot: Is there any benefit in having these variables exposed?

devops/image-pruner/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "Node16",
```
andrewsignori-aot: Can we use "target": "ES2024", following other configs in the project?

devops/image-pruner/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "image-pruner",
```
andrewsignori-aot: Please follow the other project conventions:
- "name": "studentaid-image-pruner"
- Add a description

devops/image-pruner/src/main.ts
```diff
@@ -0,0 +1,104 @@
...
+  initializePruner(client, config);
+
+  console.log("Starting image tag pruning...");
+  console.log(`Environment       : ${config.environment}`);
+  console.log(`License plate     : ${config.licensePlate}`);
+  console.log(`Applications      : ${config.applications.join(", ")}`);
+  console.log(`Jobs              : ${config.ocJobs.join(", ")}`);
+  console.log(`Tag prefix        : ${config.prefix}`);
+  console.log(`Min prefix tags   : ${config.minTags}`);
+  console.log(`Dry run           : ${config.dryRun}`);
```
andrewsignori-aot: `console.table` can be used. Examples can be found in the source code.

devops/image-pruner/src/main.ts
```diff
@@ -0,0 +1,104 @@
...
+  console.log(
+    `Image tag pruning completed. ${config.dryRun ? "DRY RUN - NO CHANGES WERE MADE TO THE CLUSTER!" : ""}`,
+  );
+  console.log("\n" + "=".repeat(80));
+}
+
+void runPruner().catch((error: unknown) => {
+  console.error("Image tag pruning failed:", error);
+  process.exit(1);
+});
```
andrewsignori-aot: Please use an auto-invoked function similar to the one in the `deploy-forms.ts`. The error handling can still happen inside the method.
If there is a benefit in this approach, please let me know.

andrewsignori-aot:
Taking a closer look at the [openshift-rest-client ](https://www.npmjs.com/package/openshift-rest-client), it is heavily dependent on the [kubernetes-client](https://www.npmjs.com/package/kubernetes-client) that does not seem frequently maintained (Published 6 years ago).

A possible alternative would be the [@kubernetes/client-node](https://www.npmjs.com/package/@kubernetes/client-node) that is not Openshift, but would support the common Kubernetes operations and apparently could support custom calls for the Openshift specific resources. 

The idea is to have a SIMS `OpenshiftClient` that would consume "a library" and provide the methods required for the pruning, or other possible Openshift interactions in the future.

In that direction, is it possible to have the `openshift.client.ts` as a class to be consumed as below, also bringing the `runtimeContext` into the class?

```ts
const openshiftClient= new OpenshiftClient({config});.
ocClient.getDeployment();
ocClient.getJob();
ocClient.getImageStream();
ocClient.deleteImageStreamTag();
```

Also, can we have the auto-executed method in the `pruner.ts` in a similar way as the `deploy-forms.ts`, which means
- The auto-executed method should create a new `openshiftClient` and use it throughout its execution.
- The helper methods can be in a separate class, if not in `OpenshiftClient`, or some intermediate helper class.

_Plase note:_ we got a confirmation that it would be ok to have this effort extended till the end of the sprint.

The above represents the outputs from the dev-sync up.

.github/workflows/prune-images.yml
sh16011993: We can remove this if not needed.

.github/workflows/prune-images.yml
andrewsignori-aot: Was this comment accepted, or is there any reason why we are keeping input parameters like `applications`, `ocjobs`, `prefix`, `minTags`, and  `ENVIRONMENT`?

devops/image-pruner/src/main.ts
```diff
@@ -0,0 +1,82 @@
...
+    throw new Error("Missing required environment variable: LICENSE_PLATE.");
+  }
+  if (!environment) {
+    throw new Error("Missing required environment variable: ENVIRONMENT.");
+  }
+  if (!["dev", "test", "prod"].includes(environment)) {
+    throw new Error(
+      `ENVIRONMENT must be one of: dev, test, prod. Got: ${environment}.`,
+    );
+  }
```
andrewsignori-aot: As mentioned [here](https://github.com/bcgov/SIMS/pull/6266#discussion_r3455182953), `ENVIRONMENT` is there a need to have it as an input?

devops/image-pruner/src/main.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const minTags = Number.parseInt(process.env.MIN_TAGS ?? "2", 10);
+  if (Number.isNaN(minTags) || minTags < 0) {
+    throw new Error(
+      `MIN_TAGS must be a non-negative integer. Got: ${process.env.MIN_TAGS}.`,
+    );
+  }
```
andrewsignori-aot: Similar to [this one](https://github.com/bcgov/SIMS/pull/6266#discussion_r3455182953), there is no reason to have it as an input. If there is a need to be customized, it must be converted to a GitHub environment; it will need a code change, either way, to update the default value.
My recommendation is to keep it as an internal constant.

devops/image-pruner/src/main.ts
```diff
@@ -0,0 +1,82 @@
...
+      .filter(Boolean),
+    ocJobs: (process.env.OCJOBS ?? "")
+      .split(",")
+      .map((value) => value.trim())
+      .filter(Boolean),
+    prefix: process.env.PREFIX ?? "main",
+    minTags,
+    dryRun: !process.argv.includes("--confirm"),
+  };
+}
```
andrewsignori-aot, line 49: This [comment](https://github.com/bcgov/SIMS/pull/6266#discussion_r3455182953) would allow this to be simplified.

devops/image-pruner/src/main.ts
```diff
@@ -0,0 +1,71 @@
...
+      licensePlate: config.licensePlate,
+      saToken: config.saToken,
+    },
+  });
+
+  const pruner = new ImagePruner(config, openshiftClient);
+  await pruner.pruneImages();
+} catch (error: unknown) {
+  console.error("Image tag pruning failed:", error);
+  process.exit(1);
```
andrewsignori-aot: The following would be equivalent, also providing the "exit 1" and the full stack trace to help with troubleshooting.
```ts
throw new Error("Image tag pruning failed", { cause: error });
````

File: devops/image-pruner/src/clients/openshift.client.ts [hunk 11]
```diff
@@ -0,0 +1,146 @@
...
+      plural: "imagestreamtags",
+      name: `${imageStreamName}:${tag}`,
+    });
+  }
+
+  /**
+   * Returns the configured OpenShift API URL.
+   * @returns The OpenShift API URL.
+   */
+  get openShiftUrl(): string {
```
andrewsignori-aot: Is this method used or intended to be used in the near future?

devops/image-pruner/tsconfig.json
andrewsignori-aot, line 1: I believe the other projects are not using imports with the extension `.ts`. Unless there is a benefit. Can we change the tsconfig as below (`"moduleResolution": "bundler"`)?
```ts
{
  "compilerOptions": {
    "target": "ES2024",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "types": ["node"],
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "noEmit": true
  },
  "include": ["src/**/*.ts"]
}
```

devops/image-pruner/.env.example
andrewsignori-aot, line 1: Please remove the ones that are no longer environment variables.

devops/image-pruner/src/clients/openshift.client.ts
```diff
@@ -0,0 +1,138 @@
...
+  BatchV1Api,
+  CustomObjectsApi,
+  KubeConfig,
+  V1Deployment,
+  V1Job,
+} from "@kubernetes/client-node";
+import type {
+  ImageStreamResource,
+  OpenshiftClientOptions,
+} from "../models/openshift.model.ts";
```
andrewsignori-aot: This [comment](https://github.com/bcgov/SIMS/pull/6266/changes#r3455355717) is intended to remove the extensions while importing to be aligned with the other project areas.

devops/image-pruner/src/clients/openshift.client.ts
```diff
@@ -0,0 +1,138 @@
...
+   * Retrieves an image stream resource from the OpenShift cluster.
+   * @param namespace The namespace of the image stream.
+   * @param name The name of the image stream.
+   * @returns The image stream resource.
+   */
+  async getImageStream(
+    namespace: string,
+    name: string,
+  ): Promise<ImageStreamResource> {
+    return await this.customApi.getNamespacedCustomObject({
```
andrewsignori-aot: Minor, no need to `await`.
Same for the other occurrences in the same file.

devops/image-pruner/src/utils/image.utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  return parts.at(-2) || "";
+}
+
+/**
+ * Extracts the image tag name from a full image reference string.
+ * @param image The full image reference string.
+ * @returns The image tag name.
+ */
+export function extractImageTagName(image: string): string {
+  const imageWithoutDigest = image.split("@")[0];
```
andrewsignori-aot: Minor, preferred syntax would be:
```ts
const [imageWithoutDigest] = image.split("@");
```

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    for (const application of this.config.applications) {
+      await this.pruneDeploymentApp(application);
+    }
+
+    for (const job of this.config.ocJobs) {
+      await this.pruneJobApp(job);
+    }
+
+    console.log("\n" + "=".repeat(80));
```
andrewsignori-aot: Code at lines 50 and 54 is optional.

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+      ) ?? [];
+
+    if (featureTagsToDelete.length === 0 && prefixTagsToDelete.length === 0) {
+      console.log("No tags to delete.");
+      return;
+    }
+
+    for (const tag of prefixTagsToDelete) {
+      console.log(
+        `  ${appIdentifier}:${tag.tag} (Older than deployed tag, created ${getTagCreatedAt(tag)})`,
```
andrewsignori-aot: Minor: `\t` can be used instead of whitespace.

devops/image-pruner/src/pruner.ts
andrewsignori-aot, line 1: Please try to add comments to all methods.

andrewsignori-aot (commented):
Great work, thanks for taking forward the lib change, minor comments left.

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+      );
+      await this.deleteTag(tag, imageStream);
+    }
+
+    for (const tag of featureTagsToDelete) {
+      console.log(
+        `  ${appIdentifier}:${tag.tag} (Feature tag, created ${getTagCreatedAt(tag)})`,
+      );
+      await this.deleteTag(tag, imageStream);
```
sh16011993: Why can't the deletion of the tags be done in parallel here?

```
await Promise.all(prefixTagsToDelete.map((tag) => this.deleteTag(tag, imageStream)));
```

Likewise, for the `prefixTagsToDelete`

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+        setting: "Apps",
+        value: this.config.applications.join(", ") || "(none)",
+      },
+      { setting: "Jobs", value: this.config.ocJobs.join(", ") || "(none)" },
+      { setting: "Tag prefix", value: this.config.prefix },
+      { setting: "Min prefix tags", value: String(this.config.minTags) },
+      { setting: "Dry run", value: String(this.config.dryRun) },
+    ]);
+
+    for (const application of this.config.applications) {
```
sh16011993, line 42: Can we run these in parallel? But on making this parallel, logs from the different apps will need to be taken care of as they will get mixed due to concurrency. To address that, you can collect them and flush them together. 

Also, if running in parallel, `Promise.allSettled` can be used as it will resolve when all the promises will settle (irrespective of success or failure) ensuring that the failure in one of the apps does not cause the execution termination of the other applications.

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    if (!deployedTag || !imageStream) {
+      console.warn(
+        `Skipping ${appName} (deployed tag was not found on ImageStream).`,
+      );
+      return;
+    }
+
+    await this.pruneTags(imageStream, deployedTag, appName);
```
sh16011993: I see that getting the deployment tag is a part of the try/catch but this step of pruning isn't. If one of the tag deletion errors out, we can still let the process complete for the deletion of other tags and log the unsuccessful deletion.

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+    const prefixTagsToDelete =
+      oldPrefixTags?.slice(
+        0,
+        Math.max(0, oldPrefixTags.length - this.config.minTags),
+      ) ?? [];
+
+    const featureTagsToDelete: ImageStreamTag[] =
+      imageStream.status?.tags?.filter(
+        (tag) =>
+          !tag.tag.startsWith(this.config.prefix) && !tag.tag.startsWith("v"),
```
sh16011993: I did not get this part where the code is trying to filter out the feature tags starting with `v`. What is prefixed with `v` that we do not want to delete?

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+      console.log(
+        `  ${appIdentifier}:${tag.tag} (Feature tag, created ${getTagCreatedAt(tag)})`,
+      );
+      await this.deleteTag(tag, imageStream);
+    }
+
+    console.log(
+      `  Total of ${featureTagsToDelete.length + prefixTagsToDelete.length} tag(s) deleted.`,
+    );
+  }
```
sh16011993, line 268: Should we delegate this to the `deleteTag` method doing the work and keep this inside the `if (!this.config.dryRun) {` code block ensuring that this is printed only during the actual prune operation from the workflow?

devops/image-pruner/src/pruner.ts
```diff
@@ -0,0 +1,247 @@
...
+    );
+
+    return { deployedTag, imageStream };
+  }
+
+  private async getJobTag(appName: string): Promise<{
+    deployedTag: ImageStreamTag | undefined;
+    imageStream: ImageStreamResource;
+  }> {
+    const jobName = `${this.config.environment}-${appName}`;
```
sh16011993, line 171: I am not getting this part. Since the images are in the tools and shared across different environments inside a license plate, can we not hardcode the environment here?

devops/image-pruner/src/pruner.ts
andrewsignori-aot: Running in parallel without proper control over how many executions can be problematic.
I would recommend keeping this as it is since it is the first effort, and once running regularly, it should delete just a few resources per day.

devops/image-pruner/src/pruner.ts
andrewsignori-aot: https://github.com/bcgov/SIMS/pull/6266#discussion_r3456142674

devops/image-pruner/src/pruner.ts
andrewsignori-aot: Release versions for now.

sh16011993 (commented):
Great work @tiago-graf 👍 

Left a few comments.

devops/image-pruner/src/pruner.ts
tiago-graf: environment is right now hardcoded to dev

devops/image-pruner/src/pruner.ts
tiago-graf: This is a total counter so inside the deleteTag would not work, are you suggesting it should not say they were deleted during the dry-run?

sh16011993: Yes, I meant that the dry-run shouldn't say that.

sh16011993 (approved):
Great work @tiago-graf 👍

andrewsignori-aot (approved):
Great work, thanks for the extra effort to switch the library. Looks good 👍

Issue #6121, opened 2026-05-08
MIN: Prod: Request URI Too Long: capacity limit for server: Dev, Test, Staging, now on Prod


astridSABC:
**Describe the Bug**
When the Ministry user logs out of the Ministry portal a "Request URI Too Long" is displayed

> Request-URI Too Long
The requested URL's length exceeds the capacity limit for this server.

Was a 500: Internal server error page is displayed instead of main login page for Ministry user

Dev/Test/Staging/Prod:
Ministry logout: Fail

Staging: 
Student: Ok
Institution: Ok
Supporting user: ok

Test environment
Student: Ok
Institution: Ok
Supporting user: Ok

**Expected Behaviour**
SIMS Ministry portal login is displayed on logout

<img width="1465" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemZaRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f749ddf4061be8ca4053d9f5283302c22fc7c622/image.png" alt="image.png" />

**Actual Behaviour**
<img width="656" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUV0RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7f3ae49e6c103888dd1811e6abd45f9fa8c82174/image.png" alt="image.png" />

<img width="1099" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHpaRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ea9eeff68a5c14c998b73d4bb9b3ff125db726eb/image.png" alt="image.png" />

**Steps To Reproduce**
Steps to reproduce the behaviour:
1. Login to the Ministry portal
2. Search for a student
3. Select "Logout"

## Developer Notes
 - This is caused by the length of the id_token_hint in the logout URL. As we've added form-specific roles, the side of the id token has grown and resulted in the error. If you go back to a small set of roles, then the logout works as expected.
 - The default length (`LimitRequestLine`) for Apache is 8190
[...]

PR #6271, opened 2026-06-17
#6121 - Ministry logout issue

weskubo-cgi:
## Summary
- Ministry logout was failing with a 414 Request-URI Too Long error because the size of the id_token used in logout has greatly increased since form roles were added.
- The keycloak js adapter doesn't have a way to request specific scopes when refreshing the token so a custom axios call was used.
- To minimize risk, the fix was only applied to Ministry users. Other SiteMinder logouts were unchanged.
- Also cleaned up the existing BCLogo component which had some unused logout code
- The client-roles client scope for the aest client needs to be updated to Optional for this update to work. This is documented in the release notes.

### DIFF
--- a/sources/packages/web/src/components/generic/BCLogo.vue
+++ b/sources/packages/web/src/components/generic/BCLogo.vue
@@ -19,26 +19,18 @@
   </v-container>
 </template>
 
-<script lang="ts">
-import { useAuth } from "@/composables";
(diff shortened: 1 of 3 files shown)


weskubo-cgi:
Code update not required. This was fixed via a Keycloak configuration change.

Issue #5837, opened 2026-03-03
Upgrade Vuetify


ninosamson:
## Acceptance criteria
- [ ] Upgrade vuetify to version 4

andrewsignori-aot:
Vuetify MCP Running and prompted with: 

> Analyse the package under the workspace project Web that contains a Vuetify 3 project. Using the vuetify-mcp server, scan this project for Vuetify 3 to 4 breaking changes. List each issue found with the file, line number, and recommended fix.

```console
I’m starting with the official Vuetify MCP v4 breaking-change catalog, then I’ll map those specific patterns against files under the Web package and return a concrete findings list.

Ran `get_v4_breaking_changes` 
Completed with input: {}

I have the official break list; now I’m running targeted searches in Web source files for concrete, detectable v3→v4 risks (removed props, renamed slot args, and behavior-dependent patterns).

Searched for text `multi-line` (`**/sources/packages/web/src/**/*.vue`), no results

Searched for text `v-snackbar` (`**/sources/packages/web/src/**/*.vue`), no results

Searched for regex `v-slot:.*\{\s*item\s*\}|#.*\{\s*item\s*\}` (`**/sources/packages/web/src/**/*.vue`), no results

Searched for regex `v-select|v-autocomplete|v-combobox` (`**/sources/packages/web/src/**/*.vue`), no results

Searched for text `v-form` (`**/sources/packages/web/src/**/*.vue`), no results

Search exclude settings blocked the previous pass; I’m rerunning against the absolute Web folder path so we get real results from the Vue codebase.

Searched for text `multi-line` (`Web UI` · `src/**`), no results

Searched for text `v-snackbar` (`Web UI` · `src/**`), 5 results

Searched for regex `v-slot:.*\{\s*item\s*\}|#.*\{\s*item\s*\}` (`Web UI` · `src/**`), 200 results

Searched for regex `v-select|v-autocomplete|v-combobox` (`Web UI` · `src/**`), 35 results

Searched for text `v-form` (`Web UI` · `src/**`), 89 results

I’m narrowing this to high-confidence breakage only by searching for exact deprecated patterns (for example `v-slot:item="{ item }"` on select-like components and any `v-snackbar` `multi-line` usage), then I’ll report precise file-level findings.

Searched for regex `v-slot:item="\{\s*item\s*\}"|#item="\{\s*item\s*\}"` (`Web UI` · `src/**/*.vue`), 1 result

[...]

PR #6279, opened 2026-06-18
#5837 - Upgrade Vuetify (Part 3)

andrewsignori-aot:
## Landing page adjustments

image

image

## Login pages

- Login pages were adjusted to also use `body-header-container`.

image

image

image

image

## Fix header navigator

image

image

image

## NOA Adjustments

- Some spacing classes, such as `pb-5`, `pt-3`, etc., were causing too much extra spacing.
- Table lines were not being displayed.

image

image

## "Error" Pages Component

- Created a common component for the following pages.
- These were requirement spacing adjustments, and where they were adjusted.

image

image

image

image

### DIFF
--- a/sources/packages/web/src/components/common/ErrorPageContainer.vue
+++ b/sources/packages/web/src/components/common/ErrorPageContainer.vue
@@ -1,0 +1,27 @@
+<template>
+  <div class="fill-height d-flex align-center justify-center">
+    <v-card elevation="0">
+      <v-card-text>
+        <v-row density="compact">
+          <v-col cols="12" md="8" class="d-flex align-center">
+            <body-header-container
+              :title="props.title"
+              :sub-title="props.subtitle"
+            >
+              <slot></slot>
+            </body-header-container>
+          </v-col>
+          <v-col cols="12" md="4" class="d-flex justify-center align-center">
+            <slot name="image"></slot>
+          </v-col>
+        </v-row>
+      </v-card-text>
+    </v-card>
+  </div>
+</template>
+<script setup lang="ts">
+const props = defineProps<{
+  title: string;
+  subtitle: string;
+}>();
+</script>
--- a/sources/packages/web/src/components/common/SharedLogin.vue
+++ b/sources/packages/web/src/components/common/SharedLogin.vue
@@ -34,26 +34,31 @@
       <template v-else>
         <v-row no-gutters density="compact">
           <v-col md="9">
-            <body-header-container
-              :title="props.title"
-              :sub-title="props.subtitle"
-            >
+            <body-header-container>
+              <template #header>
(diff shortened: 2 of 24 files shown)


sources/packages/web/src/components/generic/HeaderNavigator.vue
```diff
@@ -24,7 +24,9 @@
       </span>
     </v-col>
     <v-col class="d-flex justify-end">
-      <slot name="buttons"></slot>
+      <div>
```
andrewsignori-aot, line 27: Adding a `div` here allows the buttons to be aligned to the right, as intended by the `justify-end` at the 26th line, without the need for the consumer to add a `row` or other style.
This change allowed some `row` to be removed from the consumers (e.g., `StudentAccountApplicationsApproval.vue`).

sources/packages/web/src/components/common/SharedLogin.vue
```diff
@@ -34,17 +34,18 @@
...
-              {{ props.title }}
-            </h1>
-            <p class="my-2">
-              <slot name="subtitle">{{ props.subtitle }}</slot>
-            </p>
+            <body-header-container
+              :title="props.title"
+              :sub-title="props.subtitle"
+            >
+            </body-header-container>
```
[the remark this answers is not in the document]
andrewsignori-aot: Fixed.

sources/packages/web/src/components/common/SharedLogin.vue
```diff
@@ -1,25 +1,24 @@
 <template>
   <v-card elevation="2" class="mx-auto mt-12" max-width="670px" outlined>
     <v-card-text>
-      <template v-if="isMobile">
```
dheepak-aot, line 4: I see that isMobile is removed. Does it mean that using `body-header-container` is making it naturally responsive?

sources/packages/web/src/components/common/ErrorPageContainer.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div class="fill-height d-flex align-center justify-center">
```
dheepak-aot, line 2: Thanks for adding that layout for error pages.

tiago-graf (approved):
Looks great!

dheepak-aot (approved):
Thanks for the additional efforts!. Looks good 👍

sources/packages/web/src/components/common/SharedLogin.vue
andrewsignori-aot: The `isMobile` was added in the previous PR, but it caused too much duplication.
Removing it just made the image move to a different position, keeping the layout "the same".
If we need a different mobile layout, we can adapt.

Issue #6277, opened 2026-06-18
BUG: Cannot create International institution (Unexpected error) HOTFIX 3.5


astridSABC:
**Describe the Bug**
Unexpected error creating an international institution
Cannot validate if this occurs for all institution profile creation for Business BCeID

**Expected Behaviour**
An international institution is created without any errors

Occurs on Dev/Test/Staging

**Actual Behaviour**

Unexpected error is displayed

<img width="514" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWxERGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--aaf3f8823d5321d0d679202032ccb30fcad5d471/image.png" alt="image.png" />

If you try to edit the field, it becomes read-only
<img width="1235" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWRERGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f1b2d9e458fbe4f9d1cb63eaedcd7dd06bb916e1/image.png" alt="image.png" />

<img width="829" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFJERGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--48a624d04e58747c4fb1382426b583a7bd04173a/image.png" alt="image.png" />

Other
<img width="1034" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFZERGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d34fe4ddd05fe958d9f5271e6dc29435fae1d5af/image.png" alt="image.png" />

** Steps To Reproduce**
Steps to reproduce the behaviour:
1. Login to the ministry portal
2. Click "Create Institution" on the main page
3. Enter valid information for all the fields up to Country
4. Select any country other than Canada
5. Select "Public"
6. Select "Not for profit"
7. Select "No" for medical
8. Enter valid data for all the other fields to "Country"
9. Select "Other"
10. Enter country name in the "Country" list
[...]

PR #6284, opened 2026-06-19
#6277 - Create or update international institution hot fix

dheepak-aot:
## Root cause

This issue started since `release v3.5.0` due to the formiojs upgrade as the formio changed the default state of a select component from `undefined`(not including the property) to empty string `""`. 

This impacted the value that was sent for the property `province` in institution profile payload.

image

image

Due to this change in payload, the empty string was sent as lookup key to DB for province which is evaluated to not accepted value.

## Solution
Added a transformer in institution profile DTO class that converts any falsy value(empty string or undefined) to null and the payload is refined before consumption.

### Ministry create international institution
image
image

### Institution setup for international institution(business bceid)

image

image


andrewsignori-aot (approved):
Great analysis, looks good 👍

weskubo-cgi (approved):
Thanks for the quick turnaround on this.

Issue #6275, opened 2026-06-18
BUG: Prod: Too many requests (Failed to load resources)


astridSABC:
**Describe the Bug**
As a ministry user, when they are working with SIMS sometime there is a blank page, buttons freeze, or selecting a different section on the page freezes

The has been a 500 error or automatic logout

**Expected Behaviour**
When a ministry user selects a button, different section on the left panel, they can continue with their work

**Actual Behaviour**
SIMS does not process the request, the page/button freezes
Sometimes a blank page is displayed

Dev tools:
Date: June 18, 2024
Time: ~08:45

<img width="566" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN1ZDRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f85d0d2358af87c856711a03805156a5029166e9/image.png" alt="image.png" />

** Steps To Reproduce**

The steps are random

Steps to reproduce the behaviour:
1. Login to the ministry portal
2. Select "Forms" link under student requests
3. Select "View" for any forms that were submitted
4. Return to the main page
5. Select "Change requests" link

ninosamson:
Temporary workaround implemented to increase IP rate limit but this ticket remains open for full investigation and potential resolution.

CarlyCotton:
`THROTTLE_LIMIT`
`THROTTLE_TIME`

PR #6306, opened 2026-06-25
#6275 - Fix "Too Many Requests (429)" Error

tiago-graf:
### Summary

Enables trust proxy on api nestjs to make client ip available without the need of dynamic code.
Add throttle limit configuration by module instead of only global.

### Tests

Trust proxy disabled:
image

Trust proxy enabled:
image


sources/packages/backend/apps/api/src/auth/_tests_/e2e/throttler.e2e-spec.ts
```diff
@@ -67,9 +72,31 @@ describe("Guards and Decorators - Throttler (e2e)", () => {
...
+        message: "ThrottlerException: Too Many Requests",
+      });
+    });
   });
 
   afterAll(async () => {
-    await app.close();
+    if (app) {
+      await app.close();
+    }
```
andrewsignori-aot: The most common use is the following (which was missed here).
```ts
await app?.close();
```

sources/packages/backend/apps/api/src/auth/guards/client-route-throttler.guard.ts
andrewsignori-aot, line 1: When I noticed the like 103 using the `authorizedParties[0]` it was clear to me the azp would not work for the shared endpoints, and getting the first one seems misleading also.
Another point to note is that all the throttles would be inspected for every single request.

A possible solution would be executing the test by a combination of the URL prefixes and the client IP, setting the `ttl` and the `blockDuration` for each request.

First, a default setting can be set.
```ts
static createThrottlerOptions(config: ConfigService): ThrottlerModuleOptions {
  const { throttleConfig } = config;
  return {
    throttlers: [
      {
        name: "default",
        ttl: throttleConfig.default.time,
        limit: throttleConfig.default.limit,
      },
    ],
  };
}
```
A handleRequest can set the `ttl` and `limit`.
```ts
/**
 * Handles each request by overriding the limit and time-to-live with the
 * settings configured for the client area that owns the current controller.
 * @param requestProps throttler request properties provided by the base guard.
 * @returns true when the request is allowed to proceed.
 */
protected handleRequest(requestProps: ThrottlerRequest): Promise<boolean> {
  const clientRoute = this.getControllerClientRoute(requestProps.context);
  const { time, limit } = this.getThrottleSettingsForRoute(clientRoute);
  return super.handleRequest({
    ...requestProps,
    ttl: time,
    limit,
    blockDuration: time,
  });
}
```
Here, the `getControllerClientRoute` would need to choose the config based on the URL patterns below. In this way, when a shared controller is accessed from different clients, my assumption is that it would count them separately.
- `api/student`
- `api/aest`
- `api/institutions`
- `api/external`

sources/packages/backend/apps/queue-consumers/src/main.ts
```diff
@@ -5,10 +5,15 @@ import { QueueConsumersModule } from "./queue-consumers.module";
...
 import { SystemUsersService } from "@sims/services";
 import cookieParser from "cookie-parser";
+import { NestExpressApplication } from "@nestjs/platform-express";
 
 (async () => {
-  const app = await NestFactory.create(QueueConsumersModule);
-  const config = app.get<ConfigService>(ConfigService);
+  const app =
+    await NestFactory.create<NestExpressApplication>(QueueConsumersModule);
+  const config = app.get<ConfigService>(ConfigService);
```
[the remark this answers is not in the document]
weskubo-cgi: Might want to check your VSCode settings.

sources/packages/backend/apps/api/src/utilities/index.ts
```diff
@@ -11,6 +11,5 @@ export * from "./datatable-database-mapping-utils";
 export * from "./datatable-config";
 export * from "./pagination-utils";
 export * from "./address-utils";
-export * from "./http-utils";
```
weskubo-cgi, line 14: Thanks for cleaning this up.

sources/packages/backend/libs/utilities/src/config/config.service.ts
```diff
@@ -511,24 +513,108 @@ export class ConfigService {
   }
 
   /**
-   * Gets the throttle time in milliseconds.
-   * @returns throttle time in milliseconds.
+   * Gets all throttle settings for the API.
+   * @returns throttle settings grouped by module route.
    */
-  get throttleTime(): number {
+  get throttleConfig(): ThrottleConfig {
```
weskubo-cgi: Not a blocker but the naming here is a little inconsistent with other getters. How about:

```
  get throttleSettings(): ThrottleConfig {
    return this.getCachedConfig(
      "throttleSettingsConfig",
      this.createThrottleConfig(),
    );
  }
```
Though this now gets confusing with the `ThrottleConfig`/`ThrottleSettings` interfaces. What about if we use `ThrottleOptions` for the interface which avoid overloaded usage of the term config. So:
```
  get throttleOptions(): ThrottleOptions {
    return this.getCachedConfig(
      "throttleOptionsConfig",
      this.createThrottleConfig(),
    );
  }
```

sources/packages/backend/apps/api/src/auth/_tests_/e2e/throttler.e2e-spec.ts
```diff
@@ -43,17 +48,17 @@ describe("Guards and Decorators - Throttler (e2e)", () => {
...
 
       // Act and Assert - Make requests that exceeds the limit, all should pass
-      for (let i = 0; i < throttleLimit + 1; i++) {
+      for (let i = 0; i < defaultThrottleLimit + 1; i++) {
         await request(app.getHttpServer()).get(endpoint).expect(HttpStatus.OK);
       }
     });
 
-    it("Should block requests exceeding the rate limit (429 Too Many Requests).", async () => {
+    it("Should block requests exceeding the default throttler limit (429 Too Many Requests).", async () => {
```
weskubo-cgi: Can we make sure all test descriptions follow Should/when? E.g.

**Should block requests with a 429 Too Many Requests error when the default throttler limit is exceeded.**

sources/packages/backend/apps/api/src/auth/guards/client-route-throttler.guard.ts
```diff
@@ -0,0 +1,102 @@
...
+          // No authorized party is declared, so the controller uses the default throttle policy.
+          // routes include: /config
+          return throttleConfig.default;
+      }
+    };
+    // Caches the resolved configuration per controller class. The policy depends
+    // only on the controller's static decorator, so it is computed once and
+    // reused for both the ttl and limit resolvers across all requests, avoiding
+    // repeated reflection.
+    const configByController = new WeakMap<object, ThrottleSettings>();
```
andrewsignori-aot: Controllers are created at application bootstrap and are never removed. I did not see much benefit in using the `WeakMap` here.

sources/packages/backend/apps/api/src/auth/guards/client-route-throttler.guard.ts
andrewsignori-aot, line 1: Items to note:
- Dependencies are, as much as possible, injected. The `const reflector = new Reflector();` is not injected.
- The method `resolveThrottleConfig` is called twice to define `ttl` and `limit`.

The code below seems a bit closer to the style of the project, but it was not tested. The idea was to eliminate the above-mentioned notes.

```ts
import { ExecutionContext, Injectable } from "@nestjs/common";
import { Reflector } from "@nestjs/core";
import {
  InjectThrottlerOptions,
  InjectThrottlerStorage,
  ThrottlerGuard,
  ThrottlerModuleOptions,
  ThrottlerRequest,
  ThrottlerStorage,
} from "@nestjs/throttler";
import { ConfigService, ThrottleSettings } from "@sims/utilities/config";
import { AuthorizedParties } from "../authorized-parties.enum";
import { AUTHORIZED_PARTY_KEY } from "../decorators/authorized-party.decorator";

/**
 * Throttler guard for route-level client modules.
 *
 * The throttler runs before the authentication layer, so the decoded token is
 * not available at this point. To support independent rate limits per client
 * type the configuration is resolved from the controller's @AllowAuthorizedParty
 * decorator instead of the token.
 *
 * A single throttler is registered and the effective `ttl` and `limit` are
 * resolved together, once per request, in `handleRequest`. This collapses what
 * would otherwise be multiple throttlers each performing their own reflection
 * check on every request. Controllers without the decorator fall back to the
 * default throttle policy.
 */
@Injectable()
export class ClientRouteThrottlerGuard extends ThrottlerGuard {
  /**
   * Caches the resolved configuration per controller class. The policy depends
   * only on the controller's static decorator, so it is computed once and
   * reused across all requests, avoiding repeated reflection.
   */
  private readonly configByController = new WeakMap<object, ThrottleSettings>();

  constructor(
    @InjectThrottlerOptions() options: ThrottlerModuleOptions,
    @InjectThrottlerStorage() storageService: ThrottlerStorage,
    reflector: Reflector,
    private readonly appConfig: ConfigService,
  ) {
    super(options, storageService, reflector);
  }

  /**
   * Creates throttler options with a single throttler definition. The effective
   * limits are resolved per request in `handleRequest`, so only fallback default
   * values are defined here.
   * @param config app configuration service.
   * @returns throttler module options.
   */
  static createThrottlerOptions(config: ConfigService): ThrottlerModuleOptions {
    const { throttleConfig } = config;
    return {
      throttlers: [
        {
          name: "default",
          ttl: throttleConfig.default.time,
          limit: throttleConfig.default.limit,
        },
      ],
    };
  }

  /**
   * Handles each request by resolving the throttle configuration a single time
   * and applying it as both the time-to-live and the limit, avoiding a second
   * resolution that separate `ttl` and `limit` resolver functions would require.
   * @param requestProps throttler request properties provided by the base guard.
   * @returns true when the request is allowed to proceed.
   */
  protected handleRequest(requestProps: ThrottlerRequest): Promise<boolean> {
    const { time, limit } = this.resolveThrottleConfig(requestProps.context);
    return super.handleRequest({
      ...requestProps,
      ttl: time,
      limit,
      blockDuration: time,
    });
  }

  /**
   * Resolves and caches the throttle configuration for the controller handling
   * the current request.
   * @param context request execution context.
   * @returns throttle settings to apply to the request.
   */
  private resolveThrottleConfig(context: ExecutionContext): ThrottleSettings {
    const controller = context.getClass();
    const cachedConfig = this.configByController.get(controller);
    if (cachedConfig) {
      return cachedConfig;
    }
    const authorizedParties = this.reflector.get<AuthorizedParties[]>(
      AUTHORIZED_PARTY_KEY,
      controller,
    );
    const resolvedConfig = this.resolvePolicy(authorizedParties);
    this.configByController.set(controller, resolvedConfig);
    return resolvedConfig;
  }

  /**
   * Resolves the throttle configuration for a controller based on its
   * authorized parties.
   *
   * Since the throttler runs before authentication, the actual authorized
   * party (azp) of the request is unknown. A client-specific limit can only be
   * applied with certainty when the controller declares a single authorized
   * party. Controllers that allow multiple authorized parties (or none) use the
   * default policy, which is intentionally permissive as it also covers shared,
   * high-traffic endpoints accessed by different client types.
   * @param authorizedParties authorized parties declared by the controller.
   * @returns throttle settings to apply.
   */
  private resolvePolicy(
    authorizedParties: AuthorizedParties[],
  ): ThrottleSettings {
    const { throttleConfig } = this.appConfig;
    if (authorizedParties?.length > 1) {
      // Multiple authorized parties are allowed, so the controller is a shared,
      // high-traffic endpoint that uses the permissive default throttle policy.
      // Routes include: /dynamic-form-configuration/*, /dynamic-form/*, /system-lookup-configuration/*, /audit.
      return throttleConfig.default;
    }
    switch (authorizedParties?.at(0)) {
      case AuthorizedParties.aest:
        return throttleConfig.aest;
      case AuthorizedParties.institution:
        return throttleConfig.institutions;
      case AuthorizedParties.student:
        return throttleConfig.students;
      case AuthorizedParties.supportingUsers:
        return throttleConfig.supportingUsers;
      case AuthorizedParties.external:
        return throttleConfig.external;
      default:
        // No authorized party is declared, so the controller uses the default throttle policy.
        // routes include: /config
        return throttleConfig.default;
    }
  }
}
```

tiago-graf: It seems to work great, pushed the changes

sources/packages/backend/libs/utilities/src/config/config.models.ts
```diff
@@ -151,3 +151,23 @@ export interface QueueDashboardAccess {
   tokenExpirationSeconds: number;
   baseURL: string;
 }
+
+/**
+ * Represents one throttle settings pair.
+ */
+export interface ThrottleSettings {
+  time: number;
```
weskubo-cgi, line 163: If it's not done elsewhere it might be worth describing what each of these mean, especially since time is actually mapped to ttl. Even a copy/paste from the nest docs would be good.

sources/packages/backend/apps/api/src/app.module.ts
```diff
@@ -93,16 +94,15 @@ import { DEFAULT_METRICS_APP_LABEL } from "./route-controllers/metrics/metrics.m
...
-          ttl: config.throttleTime,
-          limit: config.throttleLimit,
-        },
-      ],
-    }),
+    ThrottlerModule.forRoot([
+      {
+        name: "default",
+        // Throttler settings ttl and limit are set to 0 as placeholders
+        // These values are overridden at runtime based on the controller in ClientRouteThrottlerGuard.
```
weskubo-cgi, line 101: 👍

sources/packages/backend/apps/api/src/auth/_tests_/e2e/throttler.e2e-spec.ts
```diff
@@ -67,9 +72,29 @@ describe("Guards and Decorators - Throttler (e2e)", () => {
         message: "ThrottlerException: Too Many Requests",
       });
     });
+
+    it("Should block requests exceeding the AEST throttler limit (429 Too Many Requests).", async () => {
```
weskubo-cgi: It might be worth adding a few more tests:

1) Fallback (default) for AEST when no env variables are available.
2) Normal and fallback scenarios for one additional AZP, since they use a different fallback (restrictive).

weskubo-cgi (changes requested):
Looks great. Just a few minor suggestions.

weskubo-cgi (approved):

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -19,6 +19,9 @@ async function bootstrap() {
     bodyParser: false,
   });
 
+  // Trust upstream proxy headers to resolve client IP addresses correctly.
+  app.set("trust proxy", 1);
```
dheepak-aot: Can we have it as const as it is used in queue consumers as well? `system-configurations-constants.ts` in lib can be the place to have const.

sources/packages/backend/apps/api/src/main.ts
```diff
@@ -19,6 +20,9 @@ async function bootstrap() {
     bodyParser: false,
   });
 
+  // Trust upstream proxy headers to resolve client IP addresses correctly.
+  app.set("trust proxy", TRUSTED_PROXY_HOPS);
```
dheepak-aot: I meant to declare constant for the property name `trust proxy` not the value. Sorry for the confusion.

devops/openshift/api-deploy.yml
```diff
@@ -635,6 +655,26 @@ parameters:
...
+  - name: STUDENTS_THROTTLE_LIMIT
+    required: false
+  - name: SUPPORTING_USERS_THROTTLE_TIME
+    required: false
+  - name: SUPPORTING_USERS_THROTTLE_LIMIT
+    required: false
+  - name: EXTERNAL_THROTTLE_TIME
+    required: false
+  - name: EXTERNAL_THROTTLE_LIMIT
+    required: false
```
dheepak-aot: The required can either be set to true or it can have value as default similar to  THROTTLE_LIMIT and THROTTLE_TIME

devops/Makefile
dheepak-aot, line 462: Other throttle environments  must follow the same path.

dheepak-aot (approved):
Thanks for making the changes @tiago-graf. Approved with some minor comments.

Issue #6282, opened 2026-06-19
Institution under review restriction part 1


Joshua-Lakusta:
## User story
As a ministry user i want to have a restriction that blocks applications and accepting but not disbursement for applications at and institution.

## Acceptance Criteria
- [x] Restriction name is
> IUR -  Institution under review.

### Restriction effects
- [x] This restriction will block all new applications at the institution
   - [x] All locations for this institution do not appear on the institution drop down list
- [x] The restriction will also affect all students that have applications in flight connected to that institution in the following way
   - [x] It will block them from accepting assessment
- [x] Behavior for the location list will mirror the de designation behavior for new applications, edits, and change requests.

### Communication

- [x] If a student is applying for funding and the restriction is added before they submit the application update the toast pop warning to the following content 

> The application cannot be submitted at this time because the institution associated with your application is currently restricted.

- [x] Add to the current accept assessment banner with the following content
- [x] This content should be shown if the IUR restriction exists and has effected the students application

> Your assessment cannot be accepted at this time because the institution associated with your application is currently under review.

- [x] Create a banner for the institution when the IUR restriction exists
- [x] Should appear at the top of the dashboard page when an institution user logs
- [x] Red banner
- [x] Can mimic the design of student based restriction banners with error, warning and no effect
   - [x] Error= red
   - [x] warning = yellow
   - [x] no effect = no banner

> **Institution Restricted**

> Your institution is currently under review.

## Technical

- [x] Create a new restriction
  - [x] Include proper metadata for the UI in the upcoming ticket.
  - [x] Use the metadata below
 ```json
{
  "fieldRequirements": {
    "program": "not allowed",
    "location": "not allowed"
  }
}
```
- [x] Allow the Ministry DTO to optionally accept no locations. This is not the main goal for this ticket, but it seems that after a preliminary test, the ministry UI may be fully prepared to submit an institution-only restriction. If these are the only required changes to allow the Ministry to submit the restriction, it can be delivered together.
- [x] Generic method to check if the application has a restriction with one of the below actions.
  - [x] Create a new restriction action: 
    - [x] Stop full-time accept assessment
    - [x] Stop part-time accept assessment
- [x] Consider expanding the current DTO `ApplicationWarningsAPIOutDTO` to return an extra property (e.g., `acceptAssessmentRestrictions`) as an array of restriction codes.
- [x] Ensure the API will validate the "accept assessment".

andrewsignori-aot:
[...]

PR #6313, opened 2026-06-29
#6282 - Institution under review restriction - Student Application

andrewsignori-aot:
- Excluded the locations from the list of available locations to be selected by the student.
  - Excluded the locations associated with a `Stop Application Eligibility` institution restriction using an `exists` where the `where clause` follows the same approach as the method `getEffectiveInstitutionRestrictions`. Both are kept at the `RestrictionSharedService` intentionally, and having both merged into one was considered, but it seemed the complexity would not be worth it.
- Prevented the application submission (using the toast already in place) in case the student had already selected the location, but the institution was put under review.
  - The backend verification is executed right after the submitted location is validated.
  - The toast notifications were "emulating" a header, and this behavior was kept. A minor refactor was executed with the intention of not affecting the current behavior while allowing the new message to be displayed without a "header".

image

- While preventing the student from selecting the now restricted institution, the application edit mode displayed the location ID, as shown below.

image

- To prevent it, a validation was implemented in the form.io utils to validate the loaded values to ensure the loaded value in the dropdown is still a valid option to be selected. The same should currently happen if an institution loses its designation status.

_Please note:_ E2E tests will be added in an upcoming PR.


sources/packages/backend/apps/api/src/route-controllers/application/application.students.controller.ts
```diff
@@ -268,6 +271,10 @@ export class ApplicationStudentsController extends BaseController {
             throw new UnprocessableEntityException(
               new ApiProcessError(error.message, error.name),
             );
+          case ACTIVE_INSTITUTION_RESTRICTION:
+            throw new ForbiddenException(
+              new ApiProcessError(error.message, error.name),
```
andrewsignori-aot, line 276: Used `ForbiddenException` following the same approach for the `ACTIVE_STUDENT_RESTRICTION`.

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -152,22 +156,36 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
+    const restrictionActionType =
+      offeringIntensity === OfferingIntensity.fullTime
+        ? RestrictionActionType.StopFullTimeApplicationEligibility
+        : RestrictionActionType.StopPartTimeApplicationEligibility;
+    designatedLocationsQuery.andWhere(
+      `NOT EXISTS(${this.restrictionSharedService.getEffectiveInstitutionRestrictionsExistsQuery(
+        designatedLocationsQuery,
+        [restrictionActionType],
+      )})`,
+    );
```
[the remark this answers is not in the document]
andrewsignori-aot: `offeringIntensity` is not optional.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -180,6 +184,12 @@ export class ApplicationService extends RecordDataModelService<Application> {
         INSTITUTION_LOCATION_NOT_VALID,
       );
     }
+    // Validate possible institution restrictions.
+    await this.institutionRestrictionCheck(
+      institutionLocation.institution.id,
+      institutionLocation.id,
+      application.offeringIntensity,
+    );
```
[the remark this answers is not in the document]
andrewsignori-aot: To be added in an upcoming PR.

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
```diff
@@ -81,20 +91,22 @@ export class RestrictionSharedService extends RecordDataModelService<Restriction
...
       );
+    if (options?.programId) {
+      query.andWhere(
+        new Brackets((qb) => {
+          qb.where("institutionRestriction.program.id = :programId", {
+            programId: options.programId,
+          }).orWhere("institutionRestriction.program.id IS NULL");
+        }),
+      );
+    }
```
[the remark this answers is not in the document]
andrewsignori-aot: Changed.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -383,6 +393,40 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
 
+  /**
+   * Check for existing restrictions at the institution and/or location level.
+   * Please note, specific program level restrictions are not checked at this level
+   * because they are also not checked at the PIR completion time.
+   * @param institutionId institution ID.
+   * @param locationId location ID.
+   * @param offeringIntensity offering intensity.
+   */
+  async institutionRestrictionCheck(
```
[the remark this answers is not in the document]
andrewsignori-aot: Fixed.

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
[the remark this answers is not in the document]
andrewsignori-aot: To be added in an upcoming PR.

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -152,22 +156,36 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
    * Gets all locations available and return just
    * a subset of available data.
+   * @param offeringIntensity application offering intensity to load designated locations for.
    * @param onlyBetaInstitutionLocations if true then return only beta institution locations.
    * @returns all locations.
    */
   async getDesignatedLocations(
+    offeringIntensity: OfferingIntensity,
     onlyBetaInstitutionLocations: boolean,
   ): Promise<Partial<InstitutionLocation>[]> {
```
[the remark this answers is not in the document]
andrewsignori-aot: Changing it to be mandatory.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -33,6 +33,16 @@ export function useFormioDropdownLoader() {
...
+    // If the current value is not in the options, reset the value to empty.
+    // Not validating the current value against the new options will allow the UI to
+    // display the value (usually the ID) of the option instead of a user-friendly label.
+    const currentValue = dropdown.getValue();
+    if (
+      currentValue &&
+      !optionsItems.some((item) => item.id === currentValue)
+    ) {
+      dropdown.setValue("");
+    }
```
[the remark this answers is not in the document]
andrewsignori-aot: Changed.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
```diff
@@ -381,24 +382,24 @@ export default defineComponent({
...
           switch (error.errorType) {
             case STUDY_DATE_OVERLAP_ERROR:
-              errorLabel = "Invalid submission.";
-              errorMsg = error.message;
+              snackBar.error(`Invalid submission! ${error.message}`);
               break;
             case ACTIVE_STUDENT_RESTRICTION:
-              errorLabel = "Active restriction!";
-              errorMsg = error.message;
+              snackBar.error(`Active restriction! ${error.message}`);
```
weskubo-cgi, line 391: Just a thought...should we be consistent and rely solely on the API error message like was done for `ACTIVE_INSTITUTION_RESTRICTION`?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -33,6 +33,18 @@ export function useFormioDropdownLoader() {
         label: item.description,
       }),
     );
+    // If the current value is not in the options, reset the value to empty.
+    // Not validating the current value against the new options will allow the UI to
```
weskubo-cgi: Maybe "...will cause the UI..."?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -33,6 +33,18 @@ export function useFormioDropdownLoader() {
...
+    // display the value (usually the ID) of the option instead of a user-friendly label.
+    const currentValue = dropdown.getValue();
+    if (
+      currentValue !== undefined &&
+      currentValue !== null &&
```
weskubo-cgi: Using `currentValue != null` will handle both these cases.

sources/packages/web/src/components/common/StudentApplication.vue
```diff
@@ -184,6 +185,20 @@ export default defineComponent({
               props.programYearId,
               props.isReadOnly,
             );
+          } else {
+            // If the form is editable and no location is selected, then reset the programs and offerings dropdowns.
```
weskubo-cgi, line 189: 👍

weskubo-cgi (approved):
Looks great. Just a few non-blocking comments.

sources/packages/backend/apps/api/src/services/application/application.service.ts
dheepak-aot, line 192: We should also add the validation to change request.

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -383,6 +393,40 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
 
+  /**
+   * Check for existing restrictions at the institution and/or location level.
+   * Please note, specific program level restrictions are not checked at this level
+   * because they are also not checked at the PIR completion time.
+   * @param institutionId institution ID.
+   * @param locationId location ID.
+   * @param offeringIntensity offering intensity.
+   */
+  private async institutionRestrictionCheck(
```
dheepak-aot, line 404: Shouldn't this method also include an optional program id?

sources/packages/backend/apps/api/src/services/application/application.service.ts
```diff
@@ -383,6 +393,40 @@ export class ApplicationService extends RecordDataModelService<Application> {
...
+      offeringIntensity === OfferingIntensity.fullTime
+        ? RestrictionActionType.StopFullTimeApplicationEligibility
+        : RestrictionActionType.StopPartTimeApplicationEligibility;
+    const institutionRestriction =
+      await this.restrictionSharedService.getEffectiveInstitutionRestrictions(
+        institutionId,
+        locationId,
+        {
+          actionTypes: [stopApplicationEligibilityAction],
+          limitOne: true,
```
dheepak-aot, line 419: 👍

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -160,14 +163,29 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
...
-      .orderBy("location.name")
+      .select(["location.id", "location.name"])
+      .innerJoin("location.institution", "institution")
       .where(
         `EXISTS(${this.designationAgreementLocationService
           .getExistsDesignatedLocation()
           .getSql()})`,
-      );
+      )
+      .orderBy("location.name");
```
dheepak-aot: Minor. Is there a reason to end the query builder statement here? (the andWhere can continue)

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
```diff
@@ -81,20 +91,25 @@ export class RestrictionSharedService extends RecordDataModelService<Restriction
...
+        new Brackets((qb) => {
+          qb.where("institutionRestriction.program.id = :programId", {
+            programId: options.programId,
+          }).orWhere("institutionRestriction.program.id IS NULL");
+        }),
+      );
+    } else {
+      // If the program is not provided, then only restrictions that are not specific to a program are considered.
+      query.andWhere("institutionRestriction.program.id IS NULL");
+    }
```
dheepak-aot, line 113: 👍

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
```diff
@@ -105,9 +120,48 @@ export class RestrictionSharedService extends RecordDataModelService<Restriction
...
+          qb.where("institutionRestriction.location.id = location.id").orWhere(
+            "institutionRestriction.location.id IS NULL",
+          );
+        }),
+      )
+      .andWhere("institutionRestriction.program.id IS NULL")
+      .andWhere(`restriction.actionType @> :${actionTypesParam}`)
+      .limit(1)
+      .getQuery();
+    parentQueryBuilder.setParameter(actionTypesParam, actionTypes);
```
dheepak-aot: I would suggest that the param name can be passed as input to the method instead of passing the query builder for that purpose. 

Input could also be an object like `{actionTypesParam:string, institutionIdAlias:string, locationIdAlias:string }` passing the details required for the subquery.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
```diff
@@ -33,6 +33,18 @@ export function useFormioDropdownLoader() {
         label: item.description,
       }),
     );
+    // If the current value is not in the options, reset the value to empty.
+    // Not validating the current value against the new options will allow the UI to
+    // display the value (usually the ID) of the option instead of a user-friendly label.
+    const currentValue = dropdown.getValue();
+    if (
```
dheepak-aot, line 40: Should this part go to the else of this condition? 

Image

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: Can you please clarify what the motivation would be to have it added to the change request?

sources/packages/web/src/composables/useFormioDropdownLoader.ts
andrewsignori-aot: Yes, they are equivalent. The idea of having them explicitly is for "readability/intent" only.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: As agreed, this will mimic the designation agreement behavior to be explored in the upcoming PR. Findings to be shared with the Team.

sources/packages/backend/apps/api/src/services/application/application.service.ts
andrewsignori-aot: Please see the method comments for the current intention.
```
Please note, specific program level restrictions are not checked at this level
because they are also not checked at the PIR completion time.
```

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
andrewsignori-aot: The current approach, it allows the `designatedLocationsQuery` to be provided as a parameter to the method `getEffectiveInstitutionRestrictionsExistsQuery`.

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
andrewsignori-aot: I am not sure if I can see any benefit. There are drawbacks to both ways.
In the current approach, which I believe should be the project pattern from now on, the helper method (e.g., `getEffectiveInstitutionRestrictionsExistsQuery`) has the ability to receive the method parameters and build its query parameters or adjust the query itself in an easy-to-extend manner.

sources/packages/web/src/composables/useFormioDropdownLoader.ts
andrewsignori-aot: I am not sure if I followed this. The `dropdown.component.data.values` should always be populated, regardless of whether the current value is present in the list.
If that is not the point being raised, please let me know.

dheepak-aot: Oh yeah this is the option values. Im good.

sources/packages/web/src/views/student/financial-aid-application/FullTimeApplication.vue
andrewsignori-aot: We should, but keep the existing behavior for now, since an effort is underway to review all toast notification messages.

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
andrewsignori-aot: As discussed, I created an interface `QueryAndParamsForExecution` to allow the parameters to be returned, being set in the method that generates the sub-query, allowing the parent to set them without the need for the parent query itself to be provided as a parameter.

I would like to keep the `QueryAndParamsForExecution` at the `simsdb` lib level, but I was not sure where exactly. Please share some insights if a better place would be better.

sources/packages/backend/apps/api/src/services/institution-location/institution-location.service.ts
```diff
@@ -158,20 +161,34 @@ export class InstitutionLocationService extends RecordDataModelService<Instituti
   async getDesignatedLocations(
     onlyBetaInstitutionLocations: boolean,
   ): Promise<Partial<InstitutionLocation>[]> {
+    // Create the query to exclude locations that have restrictions that
+    // prevent applications from being created for them.
+    // Currently the actions are specifically for full-time and part-time application eligibility,
+    // but this is reserved for future use. It is not possible to filter location based on the
+    // offering intensity, hence the query will exclude locations that have any of the actions that
+    // prevent applications from being created.
+    const { query: institutionRestrictionsExistsQuery, parameters } =
```
dheepak-aot, line 170: Thanks for the change.

sources/packages/backend/libs/services/src/restriction/restriction-shared.service.ts
dheepak-aot: The current place where the QueryAndParamsForExecution is kept seems to be perfect!

dheepak-aot (approved):
Great work. Thanks for making the change. Looks good 👍

Issue #6137, opened 2026-05-12
MIN: Pending Programs: Change "Effective end date" to "Expiry date"


astridSABC:
**Describe the task**
As a ministry user, we need to have clearer language with pending programs so that we do not accidentally use today's date and inactivate the program

**Acceptance Criteria**
- [ ] Move the content under "Approve program", to above the notes freeform section
- [ ] Below "Approve program", add "Please select the **expiry date** for the program
   

- [ ] Change "Effective end date" to "Expiry date"
    - [ ]  Add validation that this needs to be a future date

**Additional context**
- Currently the content has "Effective end date" and users could misinterpret it to "effective start date".  Adding the current date as today's date, makes the program 'Inactive'

<img width="610" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUxoRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9b3cf535768df416b0542b126f975fbe4cb52a26/image.png" alt="image.png" />

Something like the following

<img width="700" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBekV0RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--36b4ab9c4388f51abdf1b07c3af731cf7fab48f8/image.png" alt="image.png" />

tiago-graf:
@astridSABC when implemeting this ticket we decided to update the date picker component to use the new standard, so feel free to test that sceen as much as you can :)

New:
<img width="1171" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOHhuRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5f716531e079d89cbf762d66373ba8662f4302ae/image.png" alt="image.png" />

Old:
<img width="1288" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBODFuRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--11d59cab448c18915d70d9868757e39a5af497a5/image.png" alt="image.png" />

PR #6317, opened 2026-06-30
#6137 - Pending Program Expiry Date

tiago-graf:
### Summary

Update contents to display "Expiry Date" instead of "Effective End Date".
Add future date validation to expiry date.
Replace v-text-field with v-date-input

image


sources/packages/web/src/composables/useRules.ts
```diff
@@ -64,10 +64,28 @@ export function useRules() {
     if (dateString) {
       return (
         !!dateString.match(/^\d{4}-\d{2}-\d{2}$/) ||
-        "Expiry end date is not in right format."
+        "Expiry date is not in right format."
       );
     }
-    return "Expiry end date is required.";
+    return "Expiry date is required.";
```
dheepak-aot: As mentioned by co-pilot, it is used in multiple places and we can use a message parameter to be provided at consumer level. 
```ts
  const checkStringDateFormatRule = (
    dateString: string,
    fieldName = "Expiry end date",
  ) => {
    if (dateString) {
      return (
        !!dateString.match(/^\d{4}-\d{2}-\d{2}$/) ||
        `${fieldName} is not in right format.`
      );
    }
    return `${fieldName} is required.`;
  };
```

sources/packages/web/src/components/aest/institution/modals/ApproveProgramModal.vue
```diff
@@ -5,19 +5,24 @@
...
-          label="Effective end date"
+          label="Expiry date"
           class="mt-2"
           type="date"
           v-model="formModel.effectiveEndDate"
           variant="outlined"
-          :rules="[checkStringDateFormatRule]"
+          :rules="[checkFutureDateRule]"
         />
+        <div class="pb-2 mt-4">
```
weskubo-cgi: I think we can do without the `mt-4`.

sources/packages/web/src/components/aest/institution/modals/ApproveProgramModal.vue
```diff
@@ -5,19 +5,24 @@
         <error-summary :errors="approveProgramForm.errors" />
         <div class="pb-2">
           <span class="label-value"
-            >Outline the reasoning for approving this program. This will be
-            stored in the institution profile notes.</span
+            >Please select the expiry date for the program</span
           >
         </div>
         <!-- TODO Date picker is not available in the vuetify 3 version, so temporary usage of textfield and regex-->
```
weskubo-cgi: Is it possible that we can make this change to `v-date-input` now? The current implementation is heavily dependent on browser settings. This is what it looks like for me:

Image

Note, `AddExpiryDate.vue` and `ApproveDenyDesignationModal.vue` also uses the older style.

sources/packages/web/src/composables/useRules.ts
```diff
@@ -64,10 +64,28 @@ export function useRules() {
...
+  /**
+   * Check if a date string is in the future.
+   * @param dateString date string in YYYY-MM-DD format.
+   * @returns true if the date is in the future, otherwise a validation message.
+   */
+  const checkFutureDateRule = (dateString: string) => {
+    if (!dateString) {
+      return "Expiry date is required.";
+    }
+    const selectedDate = new Date(dateString);
```
dheepak-aot: We can use the dayjs utitlity to make it easier. 

```ts
// If the input date component returns only date
dayjs(dateString).startOf("day").isAfter(dayjs().startOf("day"))
// If the date component returns date and time
dayjs(dateString).isAfter(dayjs().startOf("day"))
```

sources/packages/web/src/composables/useRules.ts
```diff
@@ -64,10 +64,28 @@ export function useRules() {
...
-    return "Expiry end date is required.";
+    return "Expiry date is required.";
+  };
+
+  /**
+   * Check if a date string is in the future.
+   * @param dateString date string in YYYY-MM-DD format.
+   * @returns true if the date is in the future, otherwise a validation message.
+   */
+  const checkFutureDateRule = (dateString: string) => {
```
weskubo-cgi: I think the comment on fieldName as a parameter also applies here.

sources/packages/web/src/composables/useRules.ts
weskubo-cgi: I'm hoping we can remove any usage of this function by getting rid of the older style date inputs and replacing with `v-date-input`. See https://github.com/bcgov/SIMS/pull/6317/changes#r3515789264.

weskubo-cgi (changes requested):

sources/packages/web/src/composables/useRules.ts
dheepak-aot: Please also perform a server side validation on the same. And for the same we can use the current date-time in BC Time(We have a date util method for that.)

sources/packages/web/src/components/aest/institution/modals/ApproveProgramModal.vue
```diff
@@ -76,6 +89,17 @@ export default defineComponent({
       effectiveEndDate: "",
       approvedNote: "",
     } as ApproveProgramAPIInDTO);
+    // Two-way binding between the v-date-input (Date) and the string date
+    // stored in the form model, keeping the model as the single source of truth.
+    const effectiveEndDate = computed<Date | undefined>({
```
tiago-graf: This is not a practice used anywhere else in the app (we use a separate variable instead) but i like this better, let me know your thoughts

sources/packages/backend/apps/api/src/utilities/class-validation/custom-validators/is-date-after-today.ts
```diff
@@ -0,0 +1,47 @@
...
+} from "class-validator";
+import { getPSTPDTDateFormatted, isAfter } from "@sims/utilities";
+
+/**
+ * Checks if the decorated date is after the current date in PST time,
+ * ensuring the value represents a future date regardless of the server
+ * timezone.
+ */
+@ValidatorConstraint()
+class IsDateAfterTodayConstraint implements ValidatorConstraintInterface {
```
dheepak-aot: We can use the existing validator IsDateAfter.
 `@IsDateAfter(() => getPSTPDTDateFormatted(new Date()), "Expiry date")`

sources/packages/web/src/composables/useRules.ts
```diff
@@ -64,12 +65,28 @@ export function useRules() {
...
+   * @param date date value to be validated.
+   * @param fieldName friendly field name to be added to the validation message.
+   * @returns true if the date is in the future, otherwise a validation message.
+   */
+  const checkFutureDateRule = (date: Date | string, fieldName = "Date") => {
+    if (!date) {
+      return `${fieldName} is required.`;
+    }
+    if (!isBeforeDateOnly(new Date(), date)) {
+      return `${fieldName} must be a future date.`;
```
weskubo-cgi: Not a blocker but I'm wondering if "must be in the future." would sound better. Just a personal thought.

sources/packages/web/src/composables/useRules.ts
```diff
@@ -64,12 +65,28 @@ export function useRules() {
     if (dateString) {
       return (
         !!dateString.match(/^\d{4}-\d{2}-\d{2}$/) ||
-        "Expiry end date is not in right format."
+        "Expiry date is not in right format."
       );
     }
     return "Expiry end date is required.";
```
weskubo-cgi, line 71: We should update the field name here as well.

sources/packages/web/src/components/aest/institution/modals/ApproveProgramModal.vue
dheepak-aot: Unless it is not working can we keep it the way it is for consistency?

sources/packages/web/src/composables/useRules.ts
dheepak-aot: This part of change must be reverted as it is no more pertinent to Approve Program model.

sources/packages/web/src/components/aest/institution/modals/ApproveProgramModal.vue
tiago-graf: Since we moved from the v-text-field to v-date-input the value is now a date which prevents me from keeping it the way it was, the solution used in other components is to have a local date variable and map it to the model but i like the idea of keeping the form values centralized in the model, hence this proposed solution.

weskubo-cgi (approved):
Thanks for the final updates.

dheepak-aot (approved):
Thanks for making the changes. Looks good!

Issue #6230, opened 2026-06-11
Program suspension restriction amendments


Joshua-Lakusta:
## User story
As a ministry user i want the form of notification and the blocking conditions for this restriction to be ammended.

### Restriction impact:

- [x] Remove the logic preventing the creation of new offerings on the program
- [x] Remove any logic that is not the following in regards to the student application
   - [x] Stop the student from accepting assessment
   - [x] Stop a disbursement 

### Communication changes

- [x] Remove the banner created on the institution portal when this restriction exists (no banner will inform the institution of anything)
   - [x] This would be set to no effect for institution communication 
- [x] Change the content on the banner displayed to the student to the following

> **Your application cannot be accepted yet**
> Your application is currently pending further review by StudentAid BC. 

### New notification

- [x] Create a new notification to the ministry user email designat@gov.bc.ca
- [x] This notification will trigger when a students assessment is complete and they are blocked from accepting or when the application is blocked at ecert due to the suspension restriction.
- [x] This message is sent once for each time these triggers occur for that disbursement 
- [x] We need an separate check for each the accept assessment trigger and the ecert/disbursement trigger
- [x] Use the disbursement id for the notification of blocking disbursement/ecert
- [x] The accept trigger message can be delayed by up to 24 hours
- [x] The ecert trigger message needs to be sent as soon as possible
- [x] The template for this is available on gc notify 
`0f756338-670d-4fee-bba4-fd69f8210d17`
    - [x] Need the data values for program name and institution operation name from devs to be added to template

## Release instructions 
Confirm email addresses have been added to the new template and notification behavior

## QA Notes
-Offering can be created on the suspended program
-Student cannot accept assessment
-Stop the disbursement
Added by @dheepak-aot 
-Previously submitted applications(before the current ticket implementation is in effect) with program(and location) selected by student with SUS restriction will be displaying the program banner in the application read-only view.
<img width="956" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3BvRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2857bb99615decf54c8afd937664a19e7570e681/image.png" alt="image.png" />

dheepak-aot:
## Student application - Program restriction banner

@Joshua-Lakusta Here is my understanding of the student application program restriction banner as per our discussion.
[...]

PR #6366, opened 2026-07-14
#6230 - Minor SQL adjustment to get preliminary list of applications

dheepak-aot:
Fix: If the number of non-archived applications pending accept assessment exceed 65,535 which is the postgres limit for values passed into IN(), the sql breaks.

Solution: Adjusted the query return non-archived applications pending accept assessment only from restricted institutions.

Improvement: Added a log with count of application ids for non-archived applications pending accept assessment from restricted institutions.

image


andrewsignori-aot (approved):
Nice catch, looks good 👍

PR #6369, opened 2026-07-14
#6230 - Fix all settled promise

dheepak-aot:
The usage of `await Promise.allSettled` in student application notifications is silently ignoring the errors. 
Updated the code to fix the silent failure.


andrewsignori-aot (approved):
👍

andrewsignori-aot:
We can have it merged and continue the investigation; no secondary required.

Issue #4839, opened 2025-06-18
Remove Files: Soft Delete


CarlyCotton:
**User Story**
As a ministry user, I want to take action on a file that should not be accessible to other users (including students). This can be due to a ministry user uploading a file or a student uploading a file that may contain inappropriate content including personal information for other individuals. This is an existing step in the process when this occurs which may include contacting OCIO regarding a privacy investigation, where the file information needs to remain but should not be accessible anymore.

**Acceptance Criteria**
- [ ] Allow uploaded files to be 'soft deleted' (including date/time and user information)
- [ ] Soft deleted files "Removed" will no longer be visible through SIMS to any users
- [ ] **TO BE DETERMINED:** 
    - [ ] Ministry users can do this action on any uploaded files (new role to control this - ServiceDelivery-Full/Business Admin)
    - [ ] Ministry users who 'Remove' the file will be required to leave a note. Due to possibly sensitive nature of why this is happening the content should remind the user that it can be seen by others.
    - [ ] The note is type 'General'

**Additional Context**
This step supports further viewing of the file. If an investigation determines it can or should be fully deleted we can manually remove from S3 storage after this initial step.

dheepak-aot:
@CarlyCotton @Joshua-Lakusta What is the action that needs to be done here? I do not see any content related to this AC. 

<img width="796" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNW1DRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4e3882ac91e2b5e6b3ee26ee9bc6a949ee78f9d2/image.png" alt="image.png" />

CarlyCotton:
@dheepak-aot 
I was thinking similar to this modal that already exists with a note about the content being visible:
<img width="756" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDJGRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--09c62fc9257159c141649ac60bce54fac7dfc708/image.png" alt="image.png" />

dheepak-aot:
Ministry role to delete file: `student-delete-uploaded-file`

JerPearson:
@dheepak-aot Ministry user group that can access this feature should be: ServiceDelivery-Full.
[...]

PR #6374, opened 2026-07-16
#4839 - Remove Files: Soft Delete [backend]

sh16011993:

- Created new role student file deletion: `student-delete-uploaded-file`

Release Instruction update: This role needs to be added to the keycloak set of roles and assigned to the relevant ministry users.

- Created the db migration for the new column: `deletedAt` and the new column `deletion_note_id` linking the deleted student file to the corresponding added note.

**Rollback Screenshot:**

image

- Added the controller method to delete an uploaded file if the ministry user has the above mentioned role and saved a note for the same as a part of the same transaction

- Added e2e tests to test the above added controller method

`StudentAESTController(e2e)-deleteStudentUploadedFile.`
  √ Should delete a non-temporary file when the file exists and is not deleted.
  √ Should throw a NotFoundException when trying to delete a file that does not exist for the student.
  √ Should throw an UnprocessableEntityException when trying to delete a file that is already deleted.
  √ Should throw a NotFoundException when trying to delete a temporary file.
  √ Should throw a BadRequestException when the payload is missing the mandatory note description.
  
  **Screenshot:**

image

### DIFF
--- a/sources/packages/backend/apps/api/src/auth/roles.enum.ts
+++ b/sources/packages/backend/apps/api/src/auth/roles.enum.ts
@@ -18,6 +18,7 @@ export enum Role {
   StudentAddRestriction = "student-add-restriction",
   StudentResolveRestriction = "student-resolve-restriction",
   StudentUploadFile = "student-upload-file",
+  StudentDeleteUploadedFile = "student-delete-uploaded-file",
   StudentApproveDeclineAccountRequests = "student-approve-decline-account-requests",
   StudentEditProfile = "student-edit-profile",
   StudentApproveDeclineExceptions = "student-approve-decline-exceptions",
--- a/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
@@ -50,8 +50,7 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
       ])
       .where("studentFile.uniqueFileName = :uniqueFileName", {
         uniqueFileName,
-      })
-      .andWhere("studentFile.deletedAt IS NULL");
+      });
 
     if (studentId) {
       query.andWhere("studentFile.student.id = :studentId", { studentId });
--- a/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
@@ -50,7 +50,8 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
       ])
       .where("studentFile.uniqueFileName = :uniqueFileName", {
         uniqueFileName,
-      });
+      })
+      .andWhere("studentFile.deletedAt IS NULL");
 
     if (studentId) {
       query.andWhere("studentFile.student.id = :studentId", { studentId });
@@ -170,7 +171,6 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
         student: {
           id: true,
         },
-        deletedAt: true,
       },
       relations: {
         student: true,
--- a/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
+++ b/sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
@@ -1,5 +1,12 @@
 import { Injectable } from "@nestjs/common";
-import { DataSource, EntityManager, In, Not, UpdateResult } from "typeorm";
+import {
+  DataSource,
+  EntityManager,
+  In,
+  IsNull,
+  Not,
+  UpdateResult,
+} from "typeorm";
 import {
   RecordDataModelService,
   StudentFile,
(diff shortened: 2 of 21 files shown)


sources/packages/backend/apps/api/src/auth/roles.enum.ts
```diff
@@ -18,6 +18,7 @@ export enum Role {
   StudentAddRestriction = "student-add-restriction",
   StudentResolveRestriction = "student-resolve-restriction",
   StudentUploadFile = "student-upload-file",
+  StudentDeleteUploadedFile = "student-delete-uploaded-file",
```
andrewsignori-aot, line 21: It would be good to keep the role on the web in sync for this PR, also, but not a blocker.

sources/packages/backend/apps/db-migrations/src/sql/StudentFiles/Add-deleted-at-column.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+  sims.student_files
+ADD
+  COLUMN deleted_at TIMESTAMPTZ;
```
andrewsignori-aot: Please follow the other similar `deleted_at` columns defined below.
`deleted_at TIMESTAMP WITH TIME ZONE,`

sources/packages/backend/libs/sims-db/src/entities/student-file.model.ts
andrewsignori-aot, line 1: The AC below is requesting that the note_id be linked to the deleted file.
We should have a dedicated column to store the created note ID. The goal is to keep a strong link in the DB between the user performing the action, the file, and the note. This is not intended to have any impact in the UI for now.

- [ ] Note should link to file in the database

Suggestion: `deletion_note_id` (same as used for `sims.student_restrictions`).

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
```diff
@@ -132,10 +143,72 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
...
+    await this.dataSource.transaction(async (transactionalEntityManager) => {
+      const now = new Date();
+      await this.noteSharedService.createStudentNote(
+        studentId,
+        NoteType.General,
+        noteDescription,
+        auditUserId,
+        transactionalEntityManager,
+      );
+      await transactionalEntityManager.getRepository(StudentFile).update(
```
andrewsignori-aot: To ensure concurrency at this level, please consider executing the check as below, removing the current `STUDENT_FILE_IS_DELETED`.

```ts
const updateResult = await transactionalEntityManager
  .getRepository(StudentFile)
  .update(
    { id: studentFile.id, deletedAt: IsNull() },
    {
      deletedAt: now,
      modifier: { id: auditUserId } as User,
      updatedAt: now,
    },
  );
if (!updateResult.affected) {
  throw new CustomNamedError(
    "Student file is already set as deleted.",
    STUDENT_FILE_IS_DELETED,
  );
}
```

andrewsignori-aot (commented):
Great work, nice E2E tests. Please take a look at the comments with focus first in the below one:
https://github.com/bcgov/SIMS/pull/6374/changes#r3598752734

sources/packages/backend/apps/api/src/auth/roles.enum.ts
andrewsignori-aot: To be taken care of in the upcoming PR for the same ticket.

sources/packages/backend/apps/api/src/services/student-file/student-file.service.ts
```diff
@@ -132,10 +150,82 @@ export class StudentFileService extends RecordDataModelService<StudentFile> {
...
+        "Student file not found.",
+        STUDENT_FILE_NOT_FOUND,
+      );
+    }
+    if (studentFile.deletedAt) {
+      throw new CustomNamedError(
+        "Student file is already set as deleted.",
+        STUDENT_FILE_IS_DELETED,
+      );
+    }
```
andrewsignori-aot: Minor, this can be removed since the check on line 223 was added.

sources/packages/backend/apps/db-migrations/src/migrations/1784221801453-AddStudentFileDeletion.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getSQLFileData } from "../utilities/sqlLoader";
+
+export class AddStudentFileDeletion1784221801453 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      getSQLFileData("Add-deleted-at-column.sql", "StudentFiles"),
+    );
+    await queryRunner.query(
+      getSQLFileData("Add-deletion-note-id-column.sql", "StudentFiles"),
+    );
```
andrewsignori-aot: There is no reason to create two SQL files to affect the same table.
Please execute the changes to `StudentFiles` in a single file.

andrewsignori-aot (commented):
Thanks for doing the changes, please check the comments.

andrewsignori-aot (approved):
Since this one was raised a while ago, if there is no secondary approval, please feel free to have it merged to allow the upcoming UI one to be raised.

Issue #6302, opened 2026-06-24
Post Gold Migration Tasks


ninosamson:
**Acceptance Criteria**
- [x] ensure artifactory registry is associated with gold tools namespace. may need to rebuild bitnami images and change URL in make files 
- [x] create git workflow to retrieve SA token expiry date - @dheepak-aot  to share scripts

dheepak-aot:
Make scripts created on initial efforts.

```cmd
get-token-expiry:
    @DECODED=$$(echo $$SA_TOKEN | cut -d '.' -f 2 | { base64 -d 2>/dev/null || base64 -D 2>/dev/null; }); \
    EXP_VALUE=$$(echo $$DECODED | grep -o '"exp":[0-9]*' | cut -d ':' -f 2); \
    EXPIRY_DATE=$$(date -d "@$$EXP_VALUE" +"%Y-%m-%d %H:%M:%S %Z (UTC%z)"); \
    echo "Expiry: $$EXPIRY_DATE"
    
validate-token-expiry:
    @THRESHOLD_SECONDS=$$((7 * 24 * 60 * 60)); \
    if [ -z "$$SA_TOKEN" ]; then \
        echo "Error: SA_TOKEN is not set."; \
        exit 1; \
    fi; \
    DECODED=$$(echo $$SA_TOKEN | cut -d '.' -f 2 | { base64 -d 2>/dev/null || base64 -D 2>/dev/null; }); \
    EXP_VALUE=$$(echo $$DECODED | grep -o '"exp":[0-9]*' | cut -d ':' -f 2); \
    if [ -z "$$EXP_VALUE" ]; then \
        echo "Error: Unable to extract 'exp' from SA_TOKEN."; \
        exit 1; \
    fi; \
    NOW_TS=$$(date +%s); \
    REMAINING_SECONDS=$$((EXP_VALUE - NOW_TS)); \
    EXPIRY_DATE=$$(date -d "@$$EXP_VALUE" +"%Y-%m-%d %H:%M:%S %Z (UTC%z)"); \
    echo "Expiry: $$EXPIRY_DATE"; \
    if [ $$REMAINING_SECONDS -lt $$THRESHOLD_SECONDS ]; then \
        echo "Error: Token expires in less than 7 days."; \
        exit 1; \
    fi
```

dheepak-aot:
When the Github action is failing indicating the expiry date is less than 2 weeks, follow the Wiki to generate new SA Token. 

[...]

PR #6420, opened 2026-07-31
#6302 - GHA to validate SA Token expiry

dheepak-aot:
## GHA to validate SA Token expiry

- [x] Added a new github action which is scheduled weekly(Runs every Monday at 12:00 a.m UTC) to read and log the expiry  date and raise error(fail the job) if the expiry date is less than 14 days.
- [x] Re-organized the folder structure to bring all the scripts/tasks under devops into `task` folder.

image


.github/workflows/validate-sa-token-expiry.yml
```diff
@@ -0,0 +1,26 @@
+name: Validate SA Token Expiry
+run-name: Validate SA Token Expiry
```
andrewsignori-aot: Suggestion to have a name that explicitly mentions what the token is: 
"Validate OpenShift SA Token Expiry"

devops/task/src/sa-token/main.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // JWTs are in 3 parts separated by dots. We need the payload (2nd part).
+  const parts = token.split(".");
+  if (parts.length < 2) {
+    throw new Error("Invalid token format.");
+  }
+
+  // Decode Base64Url payload.
+  const payloadJson = Buffer.from(parts[1], "base64url").toString("utf8");
+  const payload: { exp: number } = JSON.parse(payloadJson);
```
andrewsignori-aot: The idea of using TypeScript is also to allow the usage of packages that would allow DevOps to be at the same level as the other areas of the solution.
For instance, to decode and validate the token, `jwt-decode` can be used.

andrewsignori-aot (approved):
Nice work, no blocker comments.

.github/workflows/prune-images.yml
```diff
@@ -17,11 +17,11 @@ jobs:
 
       - name: Install dependencies
         run: npm ci
-        working-directory: devops/image-pruner
+        working-directory: devops/task
```
weskubo-cgi: Definitely not blocker but I'd prefer the plural `tasks` for a grouping of tasks.

weskubo-cgi (approved):
Thanks for handling this.

Issue #4379, opened 2025-02-24
Graceful Handling of POD Apps When Pods Terminated 


ninosamson:
**User Story:** 
As a SIMS team, we need to ensure that deployments are efficient in resource consumption and graceful in shutting down services. 

**Acceptance Criteria:**
- [ ] When one or more pods are terminated during an update, the application should attempt to finalize and handle the termination gracefully, minimizing impact on users
- [ ] Check for any resources that have methods to be terminated, for instance, zeebe, redis, DB.

tiago-graf:
rollout test in DEV:

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM2pGRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--189e778f09d043b5dd6e1780f8b2b899a6b3c003/image.png" alt="image.png" />

uhelm:
seeing the same in DEV. Looks good to me:

<img width="1757" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOFRSRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6d8740fa5a27f9f62fd71af6e1a9d23bce49099c/image.png" alt="image.png" />

PR #6430, opened 2026-08-04
#4379 - Gracefully terminate applications

tiago-graf:
### Summary

Enables shutdown hooks in nestjs
Implements and adds logs to modules onApplicationShutdown for troubleshooting purposes.

### Tests

Tests with a transaction and added temporary logs/await for 40 seconds results in an error and rollback as per below:

image
image

Queue-consumer logs when testing docker stop command:
image
image

Workers logs when testing docker stop command:
image

API logs when testing docker stop command:
image

### DIFF
--- a/sources/packages/backend/apps/load-test-gateway/src/main.ts
+++ b/sources/packages/backend/apps/load-test-gateway/src/main.ts
@@ -35,7 +35,7 @@ async function bootstrap(): Promise<void> {
   await app.listen(loadTestAPIPort);
 
   // Logging node http server error
-  app.getHttpServer().on("error", (error) => {
+  app.getHttpServer().on("error", (error: unknown) => {
     logger.error(`Load test server received ${error}`, undefined, "Bootstrap");
     exit(1);
   });
@@ -44,7 +44,7 @@ async function bootstrap(): Promise<void> {
     "Bootstrap",
   );
 }
-bootstrap().catch((error) => {
+bootstrap().catch((error: unknown) => {
   const logger = new LoggerService();
   logger.error(
     `Load test gateway bootstrap exception: ${error}`,
--- a/sources/packages/backend/apps/load-test-gateway/src/main.ts
+++ b/sources/packages/backend/apps/load-test-gateway/src/main.ts
@@ -3,13 +3,16 @@ import { NestFactory } from "@nestjs/core";
 import { LoadTestModule } from "./load-test.module";
 import { KeycloakConfig } from "@sims/auth/config";
 import { LoggerService } from "@sims/utilities/logger";
-import { exit } from "process";
+import { exit } from "node:process";
 import { NestExpressApplication } from "@nestjs/platform-express";
 import { ValidationPipe } from "@nestjs/common";
 
-async function bootstrap() {
+async function bootstrap(): Promise<void> {
   await KeycloakConfig.load();
   const app = await NestFactory.create<NestExpressApplication>(LoadTestModule);
+
+  // Listens to termination signals so open connections can be gracefully closed.
+  app.enableShutdownHooks();
 
   // Get the injected logger.
   const logger = await app.resolve(LoggerService);
@@ -32,7 +35,7 @@ async function bootstrap() {
   await app.listen(loadTestAPIPort);
 
   // Logging node http server error
-  app.getHttpServer().on("error", (error: unknown) => {
+  app.getHttpServer().on("error", (error) => {
     logger.error(`Load test server received ${error}`, undefined, "Bootstrap");
     exit(1);
   });
@@ -41,7 +44,7 @@ async function bootstrap() {
     "Bootstrap",
   );
 }
-bootstrap().catch((error: unknown) => {
+bootstrap().catch((error) => {
   const logger = new LoggerService();
   logger.error(
     `Load test gateway bootstrap exception: ${error}`,
--- a/sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
+++ b/sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
@@ -1,4 +1,4 @@
-import { Injectable, Logger, OnApplicationShutdown } from "@nestjs/common";
+import { Injectable, Logger } from "@nestjs/common";
 import {
   CustomTransportStrategy,
   Server,
@@ -22,7 +22,7 @@ import {
 @Injectable()
 export class ZeebeTransportStrategy
   extends Server
-  implements CustomTransportStrategy, OnApplicationShutdown
+  implements CustomTransportStrategy
 {
   constructor(
     private readonly zeebeClient: ZeebeGrpcClient,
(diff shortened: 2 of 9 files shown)


sources/packages/backend/libs/services/src/zeebe/zeebe.module.ts
```diff
@@ -21,4 +31,19 @@ export class ZeebeModule {
...
+   * closed the same client.
+   * @param signal signal that triggered the shutdown.
+   */
+  async onApplicationShutdown(signal?: string): Promise<void> {
+    this.logger.log(
+      `Signal (${signal}) received: Closing Zeebe client connection...`,
+    );
+    await this.zeebeClient.close();
+    this.logger.log("Zeebe client connection closed.");
+  }
```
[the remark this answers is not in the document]
weskubo-cgi: This would also be consistent with the implementation for other modules.

sources/packages/backend/libs/services/src/zeebe/zeebe.module.ts
```diff
@@ -21,4 +31,19 @@ export class ZeebeModule {
...
+  /**
+   * Closes the shared Zeebe client connection while the application is
+   * gracefully shutting down. The close call is idempotent, so it is safe
+   * even when a transport strategy (e.g. the workers app) has already
+   * closed the same client.
+   * @param signal signal that triggered the shutdown.
+   */
+  async onApplicationShutdown(signal?: string): Promise<void> {
+    this.logger.log(
+      `Signal (${signal}) received: Closing Zeebe client connection...`,
```
weskubo-cgi, line 48: We should format this consistenly across `onApplicationShutdown` implementations. Personally I prefer a period to ellipsis.

sources/packages/backend/apps/load-test-gateway/src/main.ts
```diff
@@ -32,7 +35,7 @@ async function bootstrap() {
   await app.listen(loadTestAPIPort);
 
   // Logging node http server error
-  app.getHttpServer().on("error", (error: unknown) => {
```
dheepak-aot, line 35: If we remove the unknown the error is typed as any. Is there a reason for this change? same in other places.

tiago-graf: Just addressing some of the warnings: "'error' will use Object's default stringification format ('[object Object]') when stringified.sonarqube(typescript:S6551)
(local var) error: unknown"

dheepak-aot: Considering that we will be updating the way that we log error messages, I would recommend not to change the type.

sources/packages/backend/libs/services/src/queue/queue.module.ts
```diff
@@ -29,7 +32,51 @@ import { DatabaseModule } from "@sims/sims-db";
...
+
+          // Closing the queue stops processing and closes all underlying Redis clients
+          await queue.close();
+
+          this.logger.log(`Queue '${queueName}' successfully closed.`);
+        } catch (error) {
+          this.logger.error(`Error closing queue '${queueName}':`, error);
+        }
+      }),
+    );
```
dheepak-aot: Considering the nestjs/bull source code that it handles the cleanup automatically when `app.enableShutdownHooks()` is active, is it required to manually do the same?

https://github.com/nestjs/bull/blob/998aeeff01f162805183be656e56e79614ec02c6/packages/bull/lib/bull.providers.ts#L43

We also have a roadmap to move to bullMq and nestjs/bullmq also handles the cleanup automatically.

sources/packages/backend/apps/workers/src/zeebe/zeebe-transport-strategy.ts
```diff
@@ -97,7 +97,7 @@ export class ZeebeTransportStrategy
         await lastValueFrom(jobResult);
       }
       return jobResult;
-    } catch (error: unknown) {
```
dheepak-aot, line 100: We can keep the error typed as unknown.

sources/packages/backend/libs/services/src/zeebe/zeebe.module.ts
```diff
@@ -21,4 +31,23 @@ export class ZeebeModule {
...
+    );
+    try {
+      await this.zeebeClient.close();
```
dheepak-aot, line 51: As per the recommendation to gracefully shutdown the Zeebe, the workers must also be gracefully stopped using `worker.stopGracefully()` from my quick research. 

Can we see how we can use that to drain the workers?

sources/packages/backend/libs/sims-db/src/database.module.ts
```diff
@@ -14,4 +15,27 @@ import { DBEntities, ormConfig } from "./data-source";
...
+        // explicitly ensures it happens cleanly before module disposal finishes.
+        await this.dataSource.destroy();
+        this.logger.log("TypeORM connection pool successfully closed.");
+      } catch (error) {
+        this.logger.error("Error during TypeORM connection teardown:", error);
+      }
+    } else {
+      this.logger.log("TypeORM DataSource was already disconnected.");
+    }
+  }
```
dheepak-aot: Similar to Redis, As we use the standard way of `TypeOrmModule.forRoot` the shutdown hooks are expected to be handled automatically.

sources/packages/backend/libs/services/src/zeebe/zeebe.module.ts
tiago-graf: it looks like stopGracefully is an older deprecated method, as per the documentation the .close should stop connections and drain the workers.

sources/packages/backend/libs/sims-db/src/database.module.ts
tiago-graf: Removed the redundant code since closing the connection is handled automatically

sources/packages/backend/libs/services/src/queue/queue.module.ts
tiago-graf: removed redundant code

sources/packages/backend/libs/services/src/zeebe/zeebe.module.ts
dheepak-aot: I looked up the recent version documentation and couldn't see anything about this being deprecated. I believe they moved the method to different path in the SDK. Let me know if I am missing something.

https://docs.camunda.io/docs/apis-tools/typescript/api-reference/#graceful-shutdown

weskubo-cgi (approved):

sources/packages/backend/libs/services/src/zeebe/zeebe.module.ts
tiago-graf: Sorry confirmed not deprecated, I'm having a hard time finding how to implement the stopGracefully and it turns out its not part of the package/implementation we use

dheepak-aot: Confirming here from the zeebe sdk documentation that `zeebeClient.close()` will gracefully shutdown the workers and do a safe exit. I am good with the comment.

Image

dheepak-aot (approved):
Great work. Thanks for making the changes and discussion. Looks good. 👍

Issue #6436, opened 2026-08-05
INS: Student Details: File Uploads: Hide "Delete" button for soft delete


astridSABC:
**Describe the task**
As a BC Public institution user, for student details file uploads tabs, hide the "Delete" action in the file uploads section

**Acceptance Criteria**
- [ ] Hide or Remove the "Delete" action button

**Additional context**
Current state: 
The "Delete" button can be selected and the delete file modal displays.  When "Delete File" is selected there is an unexpected error.  The institution user cannot delete a file

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeCtxRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6971c3e8987bcd27a23d19680ecb1a122c3c08b8/image.png" alt="image.png" />

andrewsignori-aot:
## Demo

@astridSABC, this fix is deployed to DEV, as shown below.

### Institutions

- No longer display the "Action" column.

<img width="1450" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeXF1RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ab01c5d20c20025e185bef82ee7135e1b70a91a9/image.png" alt="image.png" />

### Ministry

- No changes are expected.

<img width="1584" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeXV1RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--259a38c42f2bec5e1399c5afe52ca496409dbfff/image.png" alt="image.png" />

astridSABC:
@andrewsignori-aot 

Thank you for the quick fix.  

For biz reference:

Institution does not have the action column for file uploads
<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzZ2RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0985b53f5e4dc4e86554b9a38cdcc4b08a936929/image.png" alt="image.png" />

[...]

PR #6442, opened 2026-08-06
#6436 - Conditionally hiding the delete option for institutions

andrewsignori-aot:
- Hiding the column "Action" that contains the "Delete" option, which should be available only for the Ministry.
- Used a similar approach that was already in place, adding some constants to create a code connection between column definitions and the place where the conditions are applied.

## Institution

image

## Ministry

image

### DIFF
--- a/sources/packages/web/src/components/common/students/StudentFileUploads.vue
+++ b/sources/packages/web/src/components/common/students/StudentFileUploads.vue
@@ -138,6 +138,8 @@ import {
   StudentFileUploadsHeaders,
   StudentFileUploadsDetails,
   ApiProcessError,
+  STUDENT_FILE_UPLOADS_UPLOADED_BY_HEADER_KEY,
+  STUDENT_FILE_UPLOADS_ACTION_HEADER_KEY,
 } from "@/types";
 import { StudentService } from "@/services/StudentService";
 import {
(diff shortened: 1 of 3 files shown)


sources/packages/web/src/components/common/students/StudentFileUploads.vue
```diff
@@ -252,11 +259,19 @@ export default defineComponent({
...
+          (header) =>
+            header.key !== STUDENT_FILE_UPLOADS_UPLOADED_BY_HEADER_KEY,
+        );
+      }
+      if (!props.canDeleteFiles) {
+        headers = headers.filter(
+          (header) => header.key !== STUDENT_FILE_UPLOADS_ACTION_HEADER_KEY,
+        );
+      }
+      return headers;
```
dheepak-aot, line 274: 👍

dheepak-aot (approved):
Looks good 👍Thanks for taking care of the issue.

sources/packages/web/src/types/contracts/DataTableContract.ts
```diff
@@ -656,16 +656,33 @@ export const ScholasticStandingHistoryHeaders = [
   { title: "Action", sortable: false, key: "action" },
 ];
 
+/**
+ * Student File Uploads header key for the uploadedBy that is conditionaly hidden.
```
weskubo-cgi: Should be **conditionally**. Same comment applies below.

weskubo-cgi (approved):

sources/packages/web/src/types/contracts/DataTableContract.ts
andrewsignori-aot: Adjusted to something similar.

Issue #1927, opened 2023-05-05
Caching typeorm queries in Used for Auth


dheepak-aot:
**Describe the task**
Implement cache for queries used in following methods.
![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/f6df2589-cea0-427a-98d1-58ef9a2cc49d)

![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/c6bf982f-4a08-447a-8c5a-f31e5418f88c)

The default lifetime of the ORM Cache across the application is 10 minutes.

When eviction/removal of cache must be instantaneous and cache cannot live for 10 minutes. One of such scenario is when an institution user is disabled by the admin and the disabled user should **_NOT_** be able to access any information right after them being disabled. What it means is, the cache holding the query result pertaining to isActive status of user must be removes immediately when isActive is updated by some other operation.

In such scenario, we could use the the cache id to remove the cache like this. 
The cache id must be a combination of cache_id_name + all the parameters of the query.

![image.png](https://images.zenhubusercontent.com/60e72aa247c95cd873ed8ce1/8da6fff3-640d-41ca-b4e5-137848f776a0)

Ref to typeorm Cache: https://github.com/typeorm/typeorm/blob/master/docs/caching.md

Create a centralized provide to manage the cache keys and cache removal across the application.

```ts
export class ORMCacheManager {
  getUserLoginCacheName(userName:string):string {

  }
  clearUserLoginCache(userName:string) {

  }
}
```

**Acceptance Criteria**
- [ ] Enable cache for the queries mentioned above
- [ ] Create centralized method for cache key and cache removal
- [ ] Wherever the cache is required to be removed immediately on an update/insert, implement cache removal as mentioned.
- [ ] Enable the cache for the query that had the cache removed in ticket #2043
- [ ] Ensure that typeorm will query DB directly if redis is down

PR #6444, opened 2026-08-06
#1927 - TypeORM Cache

tiago-graf:
### Summary

Enable cache for Auth related queries.
Adds ORM cache manager to help manage cache ids.


sources/packages/backend/libs/sims-db/src/orm-cache-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Centralizes the creation and invalidation of the TypeORM query result
+ * cache identifiers so cached queries can be found and removed by their
+ * business purpose instead of relying on ad-hoc strings spread across
+ * the code.
+ */
+@Injectable()
+export class ORMCacheManager {
+  private readonly logger = new Logger(ORMCacheManager.name);
```
andrewsignori-aot: Please refrain from using the logger from `@nestjs/common`.
Use the injected `LoggerService` from `@sims/utilities/logger`, unless there is a reason to not use it in this scenario.

sources/packages/backend/libs/sims-db/src/orm-cache-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  constructor(@InjectDataSource() private readonly dataSource: DataSource) {}
```
andrewsignori-aot: `InjectDataSource` should not be required.

sources/packages/backend/libs/sims-db/src/orm-cache-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Builds the `cache` option object accepted by TypeORM's `find`/`findOne`
+   * methods, using the default ORM cache lifetime.
+   * @param cacheId cache identifier, usually generated by one of the
+   * `get*CacheId` methods above.
+   * @returns cache option to be assigned to a `find`/`findOne` call.
+   */
+  toFindOptionsCache(cacheId: string): { id: string; milliseconds: number } {
```
andrewsignori-aot: Minor: `to` seems to be a method from a `builder`, which is not the case.
I would stick with something like `getOptionsCache`.

weskubo-cgi: It looks to me like the `cache` method signature when using find/findOne or queryBuilder is the same so I'm not sure why why this method was created specifically for find/findOne. Also, shouldn't all caches use `ORM_CACHE_LIFETIME`?

sources/packages/backend/apps/api/src/route-controllers/institution/_tests_/e2e/institution.aest.controller.updateInstitution.e2e-spec.ts
```diff
@@ -47,6 +52,12 @@ describe("InstitutionAESTController(e2e)-updateInstitution", () => {
     payload.classification = InstitutionClassification.Public;
     const endpoint = `/aest/institution/${institution.id}`;
     const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+    // Warms up the institution type cache with the original institution type.
+    const originalInstitutionType =
+      await institutionService.getInstitutionTypeById(institution.id);
+    expect(originalInstitutionType.institutionType.id).not.toBe(
```
weskubo-cgi: I get the intention here but wouldn't it just be simpler to assert the original insitution type?

sources/packages/backend/apps/api/src/services/institution-user-auth/institution-user-auth.service.ts
andrewsignori-aot, line 61: Should the `getInstitutionLocationsIds` not be cached also?

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -566,6 +568,18 @@ export class InstitutionService extends RecordDataModelService<Institution> {
...
+    const institutionUser = await this.institutionUserRepo.findOne({
+      select: { id: true, user: { userName: true } },
+      relations: { user: true },
+      where: { id: institutionUserId },
+    });
+    if (institutionUser) {
+      await this.ormCacheManager.clearUserAuthorizationsCache(
+        institutionUser.user.userName,
+      );
+    }
```
andrewsignori-aot: Should this be executed right after the successful DB update, right after the `await queryRunner.commitTransaction();`?

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateUserStatus.e2e-spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      institutionUserRole,
+    );
+    await db.institutionUserAuth.save(institutionUserAuth);
+    const { userName } = institutionUser.user;
+    // Warms up the login information cache with the user still active.
+    const activeLoginInfo = await userService.getUserLoginInfo(userName);
+    expect(activeLoginInfo.isActive).toBe(true);
+
+    const token = await getInstitutionToken(InstitutionTokenTypes.CollegeFUser);
+    const endpoint = `/institutions/institution-user/${institutionUser.id}/status`;
```
weskubo-cgi: Not a blocker since there's only a single test but we've been moving to a `getEndpoint(...)` function so that the endpoint is only defined a single time. There are existing examples in the code.

sources/packages/backend/apps/api/src/services/institution/institution.service.ts
```diff
@@ -566,6 +568,18 @@ export class InstitutionService extends RecordDataModelService<Institution> {
     } finally {
       await queryRunner.release();
     }
+    // Ensures the change takes effect immediately instead of waiting for the
+    // cached authorizations to expire.
+    const institutionUser = await this.institutionUserRepo.findOne({
+      select: { id: true, user: { userName: true } },
```
weskubo-cgi: Do we need to select `id` here?

weskubo-cgi (changes requested):
Nice work. Just a few questions/comments.

sources/packages/backend/apps/api/src/route-controllers/institution/_tests_/e2e/institution.aest.controller.updateInstitution.e2e-spec.ts
```diff
@@ -19,16 +20,20 @@ import {
 } from "@sims/sims-db/constant";
 import { ONTARIO_PROVINCE_CODE } from "@sims/test-utils/constants";
 import { getInstitutionProfilePayload } from "./institution.utils";
+import { InstitutionService } from "../../../../../src/services";
```
andrewsignori-aot: Please import directly from `../../../../services` instead of using `src`.

sources/packages/backend/apps/api/src/route-controllers/institution/_tests_/e2e/institution.aest.controller.updateInstitution.e2e-spec.ts
```diff
@@ -47,6 +52,12 @@ describe("InstitutionAESTController(e2e)-updateInstitution", () => {
     payload.classification = InstitutionClassification.Public;
     const endpoint = `/aest/institution/${institution.id}`;
     const token = await getAESTToken(AESTGroups.BusinessAdministrators);
+    // Warms up the institution type cache with the original institution type.
+    const originalInstitutionType =
+      await institutionService.getInstitutionTypeById(institution.id);
+    expect(originalInstitutionType.institutionType.id).not.toBe(
+      OUT_OF_PROVINCE_PUBLIC_INSTITUTION_ID,
+    );
```
andrewsignori-aot: Please move the assertion inside the `// Act/Assert` area.

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateUserStatus.e2e-spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    await db.institutionUser.save(institutionUser);
+    const institutionUserAuth = createFakeInstitutionUserAuth(
+      institutionUser,
+      institutionUserRole,
+    );
+    await db.institutionUserAuth.save(institutionUserAuth);
+    const { userName } = institutionUser.user;
+    // Warms up the login information cache with the user still active.
+    const activeLoginInfo = await userService.getUserLoginInfo(userName);
+    expect(activeLoginInfo.isActive).toBe(true);
```
andrewsignori-aot, line 67: Please move the assertion outside the arrange.

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateUserStatus.e2e-spec.ts
andrewsignori-aot, line 1: If I am not wrong, we do not have `redis` available in the E2E tests. If that is true, the E2E tests would not validate the cache eviction.

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateInstitutionUserWithAuth.e2e-spec.ts
```diff
@@ -185,6 +192,12 @@ describe("InstitutionUserInstitutionsController(e2e)-updateInstitutionUserWithAu
     );
     await db.institutionUserAuth.save(institutionUserAuth);
     const user = institutionUser.user;
+    // Warms up the authorizations cache with the user's original permissions.
+    const originalAuthorizations =
+      await institutionUserAuthService.getAuthorizationsByUserName(
+        user.userName,
+      );
+    expect(originalAuthorizations.authorizations).toHaveLength(1);
```
andrewsignori-aot: Please move the assertion outside the arrange.

andrewsignori-aot (commented):
Nice improvements; please taka a look at the comments.
For the below AC, lets ensure to execute a test once deployed to DEV to ensure the API will continue to work if the Redis cluster (localhost does not uses cluster) goes down.
> - [ ] Ensure that typeorm will query DB directly if redis is down

sources/packages/backend/apps/api/src/services/institution-user-auth/institution-user-auth.service.ts
andrewsignori-aot: Was this comment missed or considered not part of the ticket?

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateUserStatus.e2e-spec.ts
andrewsignori-aot: Can this comment have a reply for history?

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateInstitutionUserWithAuth.e2e-spec.ts
```diff
@@ -203,11 +210,25 @@ describe("InstitutionUserInstitutionsController(e2e)-updateInstitutionUserWithAu
     const endpoint = `/institutions/institution-user/${institutionUser.id}`;
 
     // Act/Assert
+    // Warms up the authorizations cache with the user's original permissions.
+    const originalAuthorizations =
+      await institutionUserAuthService.getAuthorizationsByUserName(
+        user.userName,
+      );
+    expect(originalAuthorizations.authorizations).toHaveLength(1);
```
andrewsignori-aot, line 218: To be clear, the `Act` in this method is the API call.
Data that needs to be pre-read or acquired prior to the act goes into the "Arrange".
Assertions go after the "Act".

sources/packages/backend/libs/sims-db/src/orm-cache-manager.ts
weskubo-cgi: Thanks for making this generic and applying it consistently.

sources/packages/backend/apps/api/src/route-controllers/institution-user/_tests_/e2e/institution-user.institutions.controller.updateUserStatus.e2e-spec.ts
tiago-graf: we are enabling redis in our E2E tests!

sources/packages/backend/apps/api/src/_tests_/e2e/orm-cache-resilience.e2e-spec.ts
tiago-graf, line 1: not sure if this is the best location for this test, I put it here since it a app wide test and not controller specific

sources/packages/backend/apps/api/src/services/institution-user-auth/institution-user-auth.service.ts
tiago-graf: locations are now cached too

sources/tests/docker-compose.yml
```diff
@@ -226,7 +229,7 @@ services:
       - ESDC_ENVIRONMENT_CODE=${ESDC_ENVIRONMENT_CODE}
       - APPLICATION_ARCHIVE_DAYS=${APPLICATION_ARCHIVE_DAYS}
       - QUEUE_CONSUMERS_PORT=${QUEUE_CONSUMERS_PORT}
-      - DISABLE_ORM_CACHE=${DISABLE_ORM_CACHE}
+      - DISABLE_ORM_CACHE=true
```
andrewsignori-aot: Can you please clarify why the ENV variable was removed and why `DISABLE_ORM_CACHE` is now set to `true`?

sources/tests/docker-compose.yml
```diff
@@ -236,6 +239,7 @@ services:
     restart: always
     depends_on:
       - postgres
+      - redis
```
andrewsignori-aot: Queue-consumers and workers do not currently depend on Redis for the tests, right?

sources/tests/Makefile
```diff
@@ -44,7 +44,7 @@ backend-unit-tests:
 
 e2e-tests-api:
 	@echo "+\n++ Make: Starting infrastructure containers for API E2E tests \n+"
-	@docker compose up -d --wait mongo formio postgres
+	@docker compose up -d --wait mongo formio postgres redis
```
andrewsignori-aot, line 47: For clarification, when the `depends_on` is specified, do we still need to explicitly set up the services?

sources/tests/.env-e2e-tests
```diff
@@ -1,26 +1,35 @@
 # This file is intended to override some values from the global .env file
+
```
weskubo-cgi: I think your formatter is being overzealous as this comment shouldn't be split. In fact, most of the blank lines aren't required.

sources/packages/backend/apps/api/src/testHelpers/auth/institution-auth-helpers.ts
```diff
@@ -30,6 +32,23 @@ import {
...
+/**
+ * Reuses a single ORMCacheManager instance per data source instead of
+ * creating a new one on every cache invalidation call.
+ * @param dataSource data source the manager will be bound to.
+ * @returns cached ORMCacheManager instance for the given data source.
+ */
+function getOrmCacheManager(dataSource: DataSource): ORMCacheManager {
+  let ormCacheManager = ormCacheManagers.get(dataSource);
+  if (!ormCacheManager) {
+    ormCacheManager = new ORMCacheManager(dataSource, new LoggerService());
```
andrewsignori-aot: Although it would work for now, it would be preferable to get a reference from the NestJS application and provide it to the required methods.
For instance, a new dependency injected into the ORMCacheManager would cause this helper to be updated.

sources/packages/backend/apps/api/src/route-controllers/institution/_tests_/e2e/institution.aest.controller.updateInstitution.e2e-spec.ts
```diff
@@ -49,12 +55,27 @@ describe("InstitutionAESTController(e2e)-updateInstitution", () => {
...
+
     await request(app.getHttpServer())
       .patch(endpoint)
       .send(payload)
       .auth(token, BEARER_AUTH_TYPE)
       .expect(HttpStatus.OK);
 
+    // The cached institution type must reflect the new type right away instead
+    // of continuing to report the previous one until the cache expires.
+    const updatedInstitutionType =
```
weskubo-cgi, line 72: My understanding is that this test (and other similar ones) will also pass if the cache isn't enabled. I think we also need a test as follows:
1) Query the record to initialize the cache
2) Make a direct db update to the record (i.e. not through the controller/service).
3) Query the record to retrieved the cached version

This test will only pass if the cache is enabled since our cache won't be automatically evicted over this small timeframe.

weskubo-cgi (changes requested):
Added a couple additional comments.

sources/tests/Makefile
tiago-graf: Yes but we run the compose up command below explicitly saying no dependencies:
```
@docker compose run --no-deps
```

sources/tests/docker-compose.yml
tiago-graf: Youre right! i'll leave redis for the API only

sources/tests/docker-compose.yml
tiago-graf: this is a global env that is being now flipped to "enabled" the problem is that we don't want to enable it for all services but only to one of them, we can discuss this during dev connect

weskubo-cgi (approved):
Thanks for making the requested updates.

sources/packages/backend/apps/api/src/auth/_tests_/e2e/guards.e2e-spec.ts
```diff
@@ -29,7 +29,8 @@ import {
   mockUserLoginInfo,
   resetMockUserLoginInfo,
 } from "../../../testHelpers";
-import { Student, User } from "@sims/sims-db";
+import { registerOrmCacheManager } from "../../../testHelpers/auth/institution-auth-helpers";
```
andrewsignori-aot: Please consume it from the index.

sources/packages/backend/apps/api/src/route-controllers/institution/_tests_/e2e/institution.aest.controller.updateInstitution.e2e-spec.ts
```diff
@@ -11,24 +12,29 @@ import {
...
+  let institutionService: InstitutionService;
   const OUT_OF_PROVINCE_PUBLIC_INSTITUTION_ID = 3;
 
   beforeAll(async () => {
-    const { nestApplication, dataSource } = await createTestingAppModule();
+    const { nestApplication, module, dataSource } =
+      await createTestingAppModule();
     app = nestApplication;
     db = createE2EDataSources(dataSource);
+    institutionService = (module as TestingModule).get(InstitutionService);
```
andrewsignori-aot: Minor, no cast required. The following should work.
```ts
institutionService = module.get(InstitutionService);
```

andrewsignori-aot (approved):

Issue #6235, opened 2026-06-12
Student: NOA: Add more items to "You are eligible to receive" section


astridSABC:
**As a** *Ministry user* **I want the You are eligible to receive: section of the Notice of Assessment in the student application** *to have weeks of study and course load* **So That** *when I view the students application will help ministry users respond more quickly*

### Acceptance Criteria

Min/Institution: Assessment: Notice of Assessment page
Student: View Assessment: Assessment page

In the "You are eligible" section,:

- [ ] Create a new column beside Intensity and add the following rows
   - [ ] Add "Course load:" [bold black font]
       - [ ] Include number and % [unstyled]
       - [ ] Condition: Only if application is Part-time

- [ ] Create a new column beside Intensity/Course load and add the following rows
     - [ ] Add "Funded Weeks of Study:"  [bold black font]
         - [ ]  Funded weeks and display number only [unstyled]
  
- [ ]  Viewable for Student, Ministry, BC Public institutions

- [ ] Accommodate for mobile view

        
 <img width="1561" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNStPRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--87254a57ec979e05700e8bcaf5fa3a648474803f/image.png" alt="image.png" />

**Additional Context**
- SBST - 4702

Current:
<img width="1225" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnhrRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4cec7bc1e6a15c0598f8d1f5ef2ca84ec47b8c28/image.png" alt="image.png" />

SFAS
<img width="585" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNUZ1RGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--695c32f5fe5469eb23dcfdb0ce310bcbeb3721eb/image.png" alt="image.png" />

astridSABC:
@astridSABC Add disability status after application number

astridSABC:
[...]

PR #6463, opened 2026-08-11
#6235 - Add Course load to NOA

tiago-graf:
### Summary

Add course load and funded weeks to notice of assessment.

image
image

### DIFF
--- a/sources/packages/forms/src/form-definitions/noticeofassessment.json
+++ b/sources/packages/forms/src/form-definitions/noticeofassessment.json
@@ -342,7 +342,7 @@
                       ],
                       "content": "Course load:",
                       "refreshOnChange": false,
-                      "key": "html114",
+                      "key": "courseLoadLabel",
                       "type": "htmlelement",
                       "input": false,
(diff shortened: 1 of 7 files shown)


sources/packages/forms/src/form-definitions/noticeofassessment.json
```diff
@@ -321,12 +321,114 @@
...
+                      "label": "HTML",
+                      "tag": "span",
+                      "className": "label-bold d-block mt-2",
+                      "attrs": [
+                        {
+                          "attr": "",
+                          "value": ""
+                        }
+                      ],
+                      "content": "Course load:",
```
dheepak-aot: While it is provided in the ACs, usage of colon `:` was not in the existing labels and the image in the ticket. I would suggest check with business one to avoid doing a PR after demo 😊

Image

dheepak-aot (approved):
Nice work. Looks good. 👍Added a clarification question.

weskubo-cgi (approved):
Looks good.

Issue #6471, opened 2026-08-13
Analysis: CRA integration file for negative value


Joshua-Lakusta:
## User story
As a applicant, parent, or partner I want my income to reflect a negative value reported from CRA as a $0.

## Acceptance Criteria
- [ ] Analysis on the federal file on how it shows a negative value and how SIMS is consuming this information (We have seen a situation where we have validated a negative CRA noa but when CRA is run it is showing as a positive of that same number.)
- [ ] Based on analysis assign an income value of $0 to be used in the assessment for that individual for all negative values.

Context
There is a current example of this in production and the PI details can be provided one on one to whoever is working this ticket.

andrewsignori-aot:
@Joshua-Lakusta, please send me the extra information required to execute the initial analysis.

andrewsignori-aot:
@Joshua-Lakusta can you please provide an example (from non-PROD) of the point in the application where the CRA income was noticed as negative? **The example does not need to be from the negative scenario**, just as an indication of where the issue is present.

andrewsignori-aot:
# Demo

## Sample File that can be used on TEST

[ABCSA99992.TXT](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFRHRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c98dbe8297a426993e8a65b2b9f3f3054dfe939c/ABCSA99992.TXT)

### File has two records with income to be validated

_Noe:_ A footer must be present; it is not validated, but it is expected to be present.

<img width="943" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUxHRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--eb8bd4e400c5d282b7e071b04890aa230a11cfbf/image.png" alt="image.png" />

### File Structure

1. SIN numbers must be the same between the two CRA income verifications. They do not need to match with any record in the system. They are used only to connect records `0022` and `0150`.
2. Transaction codes. These two lines are connected by the SIN number (1).
  - `0022`: record identification;
  - `0150`: income record (a.k.a. 15000).
3. IDs from the `sims.cra_income_verifications` that should be updated with the income.

[...]

PR #6494, opened 2026-08-19
#6471 - CRA integration file for negative value

andrewsignori-aot:
# Root Cause

The value associated with the CRA code `15000`, which is the income consumed by SIMS, can potentially receive a number appended with a negative sign.

For instance, an income of `40999` would be received as `000040999-` where currently only the 9 characters are read.

## Changes

- Updated the class reading the values from the file to also consider a possible negative value. No changes were required to allow the number to be saved to the DB.
- Updated the data sent to workflow calculations to ensure that a number, if present, will be at least 0(zero), as per business requirements.
- Updated E2E tests to ensure:
  - a negative can be received and saved to the DB,
- if a negative number is present, it must be sent to workflow as 0(zero),

### Other Changes

- Persisting missing audit column during creation and updates.
- Update some E2E tests to validate that such auditing values are being persisted.
- Create a new E2E file for income request creation.


weskubo-cgi (approved):

dheepak-aot (approved):
👍

Issue #6418, opened 2026-07-30
BUG: Resolve Restriction: View Restriction Modal: Notes to be blank instead of autopopulated with previous response


astridSABC:
**Describe the Bug**
Resolve Restriction: View Restriction modal notes are autopopulated to the next view restriction modal when one is actioned after the other.

Restrictions: Student
Restrictions: Institution

If there is a delay between resolving the restrictions the note is not autopopulated.

**Expected Behaviour**
Notes field is not autopopulated with the previous response.

**Actual Behaviour**
The "Notes" text field is autopopulated with the previous information

<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOUdnRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b357a1fa5138d4a52dc631a914410a5c299e9999/image.png" alt="image.png" />

### Steps To Reproduce

**Preconditions:** 

Student has multiple restrictions on their account.

Institution has multiple restrictions on their account.

### Steps to reproduce the behaviour:

Student Restriction

1. Login as the Ministry user
2. Search for the student with the restriction
3. Select "Restrictions" tab
4. Click "View" in the actions column
5. Enter a note
6. Click "Resolve Restriction"
7. Immediately click "View" to the next row to add a comment to another restriction
8. Note the content of the Notes free form text field

Repeat with the institution profile

 As above, if there is a delay, then the note field remains blank.

Additional context:
SD needs to add information such as the application number, so we want to ensure that the Notes text field is empty.

**Technical**
- [ ] Ensure the modal is updated to use the most recent modal improvements, also considering using the shared common modal with a note.

PR #6496, opened 2026-08-19
#6418 - Resolve restriction notes

weskubo-cgi:
## Summary
- Updated `ViewRestrictions` component so that note can be reset without impacting form submission.
- Updated to use callback to modal stays open after backend failure.
- Converted `ViewRestrictions` to script setup.
- Minor cleanup in `Restrictions`.

#### View
image

#### API failure
image

#### First Resolution
image
image

#### Second Resolution
image

#### View
image

#### First Resolution
image
image

#### Second Resolution
image

### DIFF
--- a/sources/packages/web/src/components/common/restriction/ViewRestriction.vue
+++ b/sources/packages/web/src/components/common/restriction/ViewRestriction.vue
@@ -1,131 +1,113 @@
 <template>
-  <v-form ref="viewRestrictionForm">
-    <modal-dialog-base title="View restriction" :showDialog="showDialog">
-      <template #content>
-        <error-summary :errors="viewRestrictionForm.errors" />
-        <h3 class="category-header-medium">Restriction information</h3>
-        <content-group>
+  <user-note-confirm-modal
+    title="View restriction"
+    ref="viewRestrictionModal"
+    :allowed-role="allowedRole"
+    :ok-label="allowUserToEdit ? 'Resolve restriction' : 'Close'"
+    :show-secondary-button="allowUserToEdit"
+    notes-label="Resolution reason"
+    :show-notes="allowUserToEdit"
+  >
+    <template #content>
+      <h3 class="category-header-medium">Restriction information</h3>
+      <content-group>
+        <title-value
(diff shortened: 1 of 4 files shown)


sources/packages/web/src/components/common/restriction/ViewRestriction.vue
```diff
@@ -108,118 +108,82 @@
...
 
-    const cancel = () => {
-      viewRestrictionForm.value.reset();
-      viewRestrictionForm.value.resetValidation();
-      resolvePromise(false);
-    };
+  formModel.restrictionId = props.restrictionData.restrictionId;
+  const payload = { ...formModel };
+  resolvePromise(payload);
+  viewRestrictionForm.value.reset();
```
weskubo-cgi: The issue was the missing form reset. However, since the reset clears the formModel, it is first spread to a new const. This is a common pattern. The alternative approach is to keep note as a ref and not part of the formModel (another pattern) but since there are multiple form fields anyways I left it as is.

sources/packages/web/src/views/aest/institution/Restrictions.vue
```diff
@@ -70,7 +70,6 @@
       <view-restriction-modal
         ref="viewRestriction"
         :restriction-data="institutionRestriction"
-        @submit-resolution-data="resolveRestriction"
```
dheepak-aot, line 73: Was this intentionally removed?

sources/packages/web/src/components/common/restriction/ViewRestriction.vue
dheepak-aot: As discussed can reset based on the resolvePromise response.

sources/packages/web/src/views/aest/institution/Restrictions.vue
weskubo-cgi: Yes. Unless I'm missing something this isn't emitted.

dheepak-aot (approved):

uhelm (approved):
Looks good. Thanks for the screenshots!

Issue #3958, opened 2024-11-18
INS: Study Dates and Year disappear at the end of the offering after a submitted requested change is completed


astridSABC:
**Describe the Bug**
Study Dates and Year disappear off the end after request a change is submitted requested change is completed

**Expected Behaviour**
Study Dates and Year are still displayed in the offering name after the request a change is submitted for In Progress and Completed status
Read-only name displays the offering name, study dates and year of study like it does on selection.

Sample of study dates

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnpKQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--631a77a55363e2ec312c222b1f6c4301a5ff8827/image.png" alt="image.png" />

**Actual Behaviour**
Study Dates and Year disappear

Note: This occurs for any character length offering name

App: 2025000395
<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd0lwRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8b7df7e5dd02181905256fab9d523a2f3a45e34a/image.png" alt="image.png" />

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnJKQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ee1398c1cb5c7da5f8777fcb1e73b3239b439747/image.png" alt="image.png" />

<p></p>
<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNmJKQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--67eef7b91aa6b196511b32b7dc4857140d5c47b5/image.png" alt="image.png" />

<p></p>
[...]

PR #6508, opened 2026-08-21
#3958 - E2E Tests for get application offering change API

dheepak-aot:
## E2E Tests
```spec
ApplicationOfferingChangeRequestInstitutionsController(e2e)-getApplicationOfferingChangeRequest
    √ Should get the application offering change request when application offering change request exists. (235 ms)
    √ Should throw not found exception when application offering change request does not exist. (18 ms)
    √ Should throw forbidden exception when the institution user does not have access to application location. (127 ms)
```


weskubo-cgi (approved):
Thanks for adding these tests.

uhelm (approved):
Looks good to me. Thank you!

PR #6512, opened 2026-08-24
#6509 - Revert openshift-tools-installer to previous version

uhelm:
## Summary

The redhat/openshift-tools-installer @v2 and @v3 contain a bug that causes them to take 3 minutes for a successful run as opposed to 5s as expected. There are no workarounds or switches for this issue so reverting to an old version is the best action for now. RedHat claims that this issue was fixed in v3 but it doesn't appear to be the case. I opened an upstream bug report and hope this should get fixed in a future release. If not, we can investigate alternative actions or something custom.

## Screenshots

v2 and v3 with skip_cache: false or skip_cache: true

image

v1.13

image

### DIFF
--- a/.github/workflows/clamav.yml
+++ b/.github/workflows/clamav.yml
@@ -36,9 +36,10 @@ jobs:
         with:
           ref: ${{ github.ref_name }}
       - name: Install CLI tools from OpenShift Mirror
-        uses: redhat-actions/openshift-tools-installer@v3
+        uses: redhat-actions/openshift-tools-installer@144527c7d98999f2652264c048c7a9bd103f8a82
         with:
           oc: "4"
+          skip_cache: true
       - name: Log in to OpenShift
         run: |
           oc login --token=${{ secrets.SA_TOKEN }} --server=${{ vars.OPENSHIFT_CLUSTER_URL }}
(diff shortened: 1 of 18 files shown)


.github/workflows/clamav.yml
```diff
@@ -36,9 +36,10 @@ jobs:
         with:
           ref: ${{ github.ref_name }}
       - name: Install CLI tools from OpenShift Mirror
-        uses: redhat-actions/openshift-tools-installer@v3
+        uses: redhat-actions/openshift-tools-installer@144527c7d98999f2652264c048c7a9bd103f8a82
```
dheepak-aot, line 39: @uhelm this commit goes back to V1.13.1 which uses Node 16. The issue started with actions using node 20 
I wasn't aware that we have to go back so far during the discussion. Hence considering the 3 minute delay can be digested temporarily I would suggest to wait a while to see if they provide a new version instead of we downgrading it.

Also by looking into the actions, the issue seems to have started from V3 only as I can see the GHAs executed in V2 are still taking only few seconds.

image

I will be ok, if going to v2 is considered as temporary solution until we wait for the new version.

uhelm: @dheepak-aot It didn't start with the v3 upgrade, there are many v2 jobs where this happened as well, i.e. https://github.com/bcgov/SIMS/actions/runs/31039583025/job/92420471304

It doesn't seem to be related to caching / not caching and doesn't seem to follow a specific pattern. All node <22 jobs are now executed with node24 regardless so I don't think there is a major difference in the runtime. But I'm fine to wait or revert to v2 if you think that's better.

uhelm:
RedHat released a new version that fixes the keep-alive issue. Will close for now and submit a different PR

Issue #6507, opened 2026-08-21
Pods are taking a long time to enter Ready state


uhelm:
ReadinessProbe initial delay is too long. It takes the 'api' pod ~1m to become Ready.

Fix this by adding a startupProbe with a much shorter interval

PR #6514, opened 2026-08-24
#6507 - Fix health probes for openshift deployments

uhelm:
## Summary

- Pods are taking a long time to enter Ready state which delays deployment and GitHub Actions completion
- Adds startup probes to the API, Forms, Queue Consumers, and Web OpenShift deployments.
- Removes default probe values.
- Removes invalid `path` and `scheme` values from the Forms TCP socket probes.
- Allow startup probes to gate readiness and liveness checks instead of relying on initial delays.

General Logic:
- Only Startup Probe is active initially with a much shorter interval than the regular probes. If the startup probe fails for 60 seconds (periodSeconds * failureThereshold) the pod is restarted
- Once the startup probe is healthy the regular livenessprobe and readinessprobe take over. These run for the same periods and timeouts as before (30s)
## Validation

- Pods enter healthy status in 5s instead of 60s

```
old pod status (61s difference between initialize and ready):

    - type: Initialized
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-08-18T16:15:39Z'
    - type: Ready
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-08-18T16:16:40Z'
```

```
new pod status (7s difference between initialize and ready):

    - type: Initialized
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-08-24T23:28:25Z'
    - type: Ready
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-08-24T23:28:32Z'
```

### DIFF
--- a/devops/openshift/forms-deploy.yml
+++ b/devops/openshift/forms-deploy.yml
@@ -85,23 +85,23 @@ objects:
                   memory: "${MEMORY_REQUEST}"
                 limits:
                   memory: "${MEMORY_LIMIT}"
+              startupProbe:
+                tcpSocket:
+                  port: 3001
+                periodSeconds: 5
+                timeoutSeconds: 1
+                failureThreshold: 20
               livenessProbe:
                 tcpSocket:
-                  path: "/ops/healthz"
                   port: 3001
-                  scheme: HTTP
                 timeoutSeconds: 1
                 periodSeconds: 10
-                successThreshold: 1
                 failureThreshold: 3
               readinessProbe:
                 tcpSocket:
-                  path: "/ops/readyz"
                   port: 3001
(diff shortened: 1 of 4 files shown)


devops/openshift/forms-deploy.yml
andrewsignori-aot, line 1: This one is not part of the regular release. We should have it deployed at least to DEV once the PR is merged.

andrewsignori-aot (approved):
Nice work, great improvement, looks good 👍

sh16011993:
Nice work @uhelm 👍  Thanks for the call discussing the details.

sh16011993 (approved):

PR #6524, opened 2026-08-25
#6509 - Update openshift-tools-installer from v3 to v3.1

uhelm:
## Summary
- Fixes a longstanding bug where the action step takes 3 minutes instead of 5s
- Setting skip_cache: true because caching does not save any execution time and requires elevated permissions (actions: write) that introduce a security risk


andrewsignori-aot (approved):
Nice work, thanks for taking care of it, looks good 👍

sh16011993:
Nice work @uhelm 👍  Thanks for the call discussing the details.

sh16011993 (approved):

Issue #6528, opened 2026-08-25
Hotfix: BUG Parental dependant calculations


Joshua-Lakusta:
### Issue

- [ ] This issue effects 25-26 and 26-27 program years

- [ ] Parent 2 calculation for contribution based on dependants is referencing parent 1 instead of parent 2 leading to incorrect calculations. I have bolded the lines that contain the parent1 errors and should be 2.

### Camunda

```
calculatedDataTotalEligibleParent2ContributionDependants
// 19-22 years of age: post-sec must be Yes
  if parent1DependentTable = null then 1 else
  count(
  parent1DependentTable[
  date(item.dateOfBirth) < (date(offeringStudyStartDate) - duration("P18Y"))
  and date(item.dateOfBirth) >= (date(offeringStudyStartDate) - duration("P22Y"))
  and item.attendingPostSecondarySchool = "yes"
]
  )+1
```

### Data pull
Once fix is complete we need a csv of all students where there is 2 parents and parent 2 entered 1 or more dependants and parent 2 dependants is greater than parent one.

### Reference image 
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2pVRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2f15c1ddddfd9b421b520cf1a1043bc7a7820fa5/image.png" alt="image.png" />

andrewsignori-aot:
# Data Extraction Query

- Find applications that:
  - Not edited or cancelled;
  - Have two parents;
  - Both parents submitted their data;
  - Second parent declared `parentsOtherDependants` as `yes`;
  - Second parent has more dependents declared than the first parent.

```sql
WITH two_parents_applications AS (
    SELECT
        application_id,
        COUNT(*) AS record_count,
        MIN(id) AS parent1_id,
        MAX(id) AS parent2_id
    FROM
        sims.supporting_users
    GROUP BY
        application_id
    HAVING
        COUNT(*) = 2
)
SELECT
    applications.id "Application Id",
    applications.application_number "Application Number",
    applications.application_status,
    two_parents_applications.parent1_id "Parent 1 Supporting User Id",
    two_parents_applications.parent2_id "Parent 2 Supporting User Id"
FROM
    two_parents_applications
    INNER JOIN sims.applications applications ON applications.id = two_parents_applications.application_id
[...]

PR #6531, opened 2026-08-25
#6528 - Hotfix: BUG Parental dependant calculations

andrewsignori-aot:
While calculating the `calculatedDataTotalEligibleParent2ContributionDependants` for Parent 2, the Parent 1 data was used by mistake.

The issue would affect students with two parents, where the "second parent" would declare dependents.

image

## Workflow calculation for parent 2

image

### Parent 2 with one dependent (ID 1234)

image

### Parent 1 with no dependent (ID 1233, the lesser of the two IDs is Parent 1)

image


dheepak-aot (approved):
Great catch! Looks good 👍

weskubo-cgi (approved):
Approved. Thanks for the quick turnaround.

Issue #6547, opened 2026-08-31
BUG - breaks in study


Joshua-Lakusta:
**Describe the Bug**
Offering breaks in study should allow the submission of a study break when the no breaks in study option is selected (when there is no study break info)

**Expected Behaviour**
It should not try to validate the hidden fields when no study breaks is selected and should allow submission

**Actual Behaviour**
Offering breaks in study trying to validate the hidden fields and it creates a toast pop (see picture at the bottom)

** Steps To Reproduce**
Steps to reproduce the behaviour:
1. Go to an instiution account
2. go to a program
3. create an offering
4. fill offering with generic info
5. click no study breaks option
6. Validate the offering
7. Submit the offering 

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNmprRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c239ff70105baf167fef49ebf5dff3363943d815/image.png" alt="image.png" />

Joshua-Lakusta:
Must be resolved before release based on criticality.

Joshua-Lakusta:
Tested and no issues. Closing

PR #6554, opened 2026-09-01
#6547 - BUG - breaks in study

andrewsignori-aot:
# The issue

- Even when no study breaks are provided, the form.io submission contains an empty entry for study breaks that correctly fails validation.
- Other validators in the offering model are passing even while validating the empty study break.

## The fix

- Some attempts were made to force form.io to remove the hidden empty fields from the submission, such as:
  -  using `clearOnHide`;
  - moving the conditional from the form.io UI to the custom condition;
  - changing the default value to an empty array when `lackStudyBreaks' was `true`.

_Note:_ fixed an E2E where `lackStudyBreaks' was `true` and the API was wrongly saving the provided study breaks.

## Final fix

- Sanitizing the received payload to prevent any study break information from passing over to the next layers if `lacksStudyBreaks` is `true`.

_Note:_ A possible fix would be to include a validator to ensure no study breaks are present if `lackStudyBreaks' is `true`, but it would demand further investigation to ensure form.io data does not contain empty study breaks, or they must be removed prior to the API submission. The current fix was considered "good enough"; it ensures data sanitization similar to the CSV bulk upload, and it is centralized on the server side.

image

## E2E Test

- An E2E test will be added for the main sync-up.

### DIFF
--- a/sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.controller.service.ts
+++ b/sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.controller.service.ts
@@ -217,7 +217,7 @@ export class EducationProgramOfferingControllerService {
       operatingName: institutionLocation.institution.operatingName,
       legalOperatingName: institutionLocation.institution.legalOperatingName,
       primaryEmail: institutionLocation.institution.primaryEmail,
-      // Ignore any study breaks submitted if not study breaks are indicated in the payload.
+      // Ignore any study breaks submitted if no study breaks are indicated in the payload.
       studyBreaks: payload.lacksStudyBreaks ? [] : payload.studyBreaks,
       programContext: program,
       institutionContext: {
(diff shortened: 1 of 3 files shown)


sources/packages/backend/apps/api/src/route-controllers/education-program-offering/education-program-offering.controller.service.ts
```diff
@@ -217,7 +217,8 @@ export class EducationProgramOfferingControllerService {
       operatingName: institutionLocation.institution.operatingName,
       legalOperatingName: institutionLocation.institution.legalOperatingName,
       primaryEmail: institutionLocation.institution.primaryEmail,
-      studyBreaks: payload.studyBreaks,
+      // Ignore any study breaks submitted if no study breaks are indicated in the payload.
+      studyBreaks: payload.lacksStudyBreaks ? [] : payload.studyBreaks,
```
dheepak-aot, line 221: 👍

dheepak-aot (approved):
Looks good 👍

sh16011993 (approved):
LGTM @andrewsignori-aot 👍

Issue #6527, opened 2026-08-25
Database backups take up too much space


uhelm:
We're averaging ~800MB for every 2h incremental backup, even though our database size is only 3.5GB. This is throughout the entire day, even at off hours. Points to an issue with overly frequent field updates. Will need some investigation.

**Acceptance Criteria**
- [ ] Database incremental backups should reflect the actual database changes for that time period.
- [ ] Test the different options from `pgbackrest` to allow backups by changed blocks instead of the entire file, which is probably what is generating ~800MB for every 2h incremental backup.
- [ ] Execute the tests on DEV and then apply the solution to all environments.
- [ ] Ensure the PVC and S3 backups are updated to use the new configs.

**Additional context**
- pgbackrest docs: https://pgbackrest.org/user-guide.html#backup/bundle

uhelm:
Analyzed database operations for a 12h period 8pm (Aug 25) to 8am (Aug 26). There are significant database inserts/updates for the SFAS and FRL integrations. However, those do not explain consistent large incrementals every 4 hours.
Most likely cause for this is how pgbackrest backs up data. Small changes in tables cause the entire file (not sure which file - WAL?) to be backed up. The repo1-bundle and repo1-block options for pgbackrest/crunchy solve this.

## Recommendation
Implement repo-bundle and repo-block for both repos in the PostgresCluster object

PR #6571, opened 2026-09-03
#6527 - pgbackrest incremental backup changes

uhelm:
## Summary
Incremental backups are taking up too much space because entire DB files are backed instead of changed blocks.

This change contains:
- Adding the repo.bundle and repo.block options for repo1 and repo2 to enable block-based incremental backups
- Removing outdated Helm namespace values file from the SILVER cluster
- Removing two outdated CPU limit settings for pgbackrest

Has been applied manually in DEV


devops/helm/crunchy-postgres/values-e0a504-prod.yaml
```diff
@@ -65,14 +65,12 @@ pgBackRest:
         cpu: 500m
         memory: 550Mi
       limits:
-        cpu: 500m
```
dheepak-aot, line 68: As discussed, if the cpu limit is no longer effective, we must remove it starting from postgesCluster yaml.

devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
```diff
@@ -121,11 +121,15 @@ spec:
         # Support both PVC and s3 backups
         repo1-retention-full: {{ .Values.pgBackRest.retention | quote }}
         repo1-retention-full-type: {{ .Values.pgBackRest.retentionFullType }}
+        repo1-bundle: 'y'
+        repo1-block: 'y'
```
dheepak-aot: Minor. Can we have these values coming through values file. Just having in values.yaml will do.

dheepak-aot: Minor. Can we have these values coming through values file. Just having in values.yaml will do.

devops/helm/crunchy-postgres/values-e0a504-prod.yaml
uhelm: You are right. I missed the main object properties. Fixed now

devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
uhelm: I moved both repo1 and repo2 settings to be configurable

devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
```diff
@@ -121,11 +119,15 @@ spec:
         # Support both PVC and s3 backups
         repo1-retention-full: {{ .Values.pgBackRest.retention | quote }}
         repo1-retention-full-type: {{ .Values.pgBackRest.retentionFullType }}
+        repo1-bundle: {{ .Values.pgBackRest.repos.bundle | squote }}
+        repo1-block: {{ .Values.pgBackRest.repos.block | squote }}
```
dheepak-aot: Can we use the quote instead of single quote to follow the pattern of the existing file?

dheepak-aot (approved):
Thanks for the investigation and work done. Looks good 👍
Added a minor comment.

devops/helm/crunchy-postgres/templates/PostgresCluster.yaml
uhelm: OpenShift turns the quotes into single quotes so I chose those. We can stick to the double quotes to keep it consistent though. Fixed that.

andrewsignori-aot (approved):
Thanks for testing and make it move forward. Looks good 👍

PR #6590, opened 2026-09-08
#6343 - Program Vuetify conversion - 1

dheepak-aot:
# Evaluators and System lookups

## Create system lookups
- [x] Created DB migrations to move the lookup data from form definitions into `sims.system_lookup_configurations`. 
Note: The columns in education programs table does not have a constraint for the lookup key due to existing structure(some of the lookup keys and individual columns) and also trying to keep the efforts close to the vuetify migration only.
 - [x]  Created system lookups for `Program length` `Institution regulatory body` `Program entrance requirement` and `Program aviation credential`
## Evaluators
- [x] Create evaluator framework that evaluates `Program status` and `Field of study code`. 
- [x] Create an API to evaluate the program calculated values and exposed only `Field of study code` for now.

## Unit tests
- [x] Created unit tests to validate the calculated `Field of study code`.
```spec
EducationProgramFieldOfStudyCodeEvaluator-evaluate
    √ Should return 25 when credential type is undergraduateCertificate and CIP code is 12.1234. (5 ms)
    √ Should return 55 when credential type is graduateDiploma and CIP code is 12.1234.
    √ Should return 70 when credential type is postGraduateOrDoctorate and CIP code is 12.1234. (1 ms)
    √ Should return 63 when credential type is postGraduateOrDoctorate and CIP code is 31.1234.
    √ Should return 33 when credential type is undergraduateDegree and CIP code is 31.1234.
    √ Should return 20 when credential type is undergraduateDegree and CIP code is 77.1234. (1 ms)
    √ Should return 20 when CIP code is not provided.
    √ Should return 20 when credential type is not provided.
```
- [x] Created unit tests to validate the calculated `Program status`.
```spec
EducationProgramStatusEvaluator-evaluate
    √ Should return Approved when all conditions are met for approval. (3 ms)
    √ Should return Pending when program is BC private and delivered online only, making it pending. (1 ms)
    √ Should return Pending when program has a joint institution, making it pending.
    √ Should return Pending when program is delivered without academic credits, making it pending.
    √ Should return Pending when program ESL eligibility is 20 or more, making it pending. (1 ms)
    √ Should return Pending when program has no entrance requirements, making it pending.
    √ Should return Pending when program has less than the minimum expected hours per week, making it pending.
    √ Should return Pending when program is an aviation program, making it pending.
    √ Should return Pending when program has a WIL component that is not approved, making it pending.
    √ Should return Pending when program has an approved WIL component that is not eligible, making it pending.
[...]

Issue #6343, opened 2026-07-08
Convert Pages to Vuetify - Program


CarlyCotton:
The NOA page and others are built using FormIO and the number of dynamic features and values on the page could be better supported by using Vuetify. 

**Acceptance Criteria**
- [ ] Convert Program page
- [ ] All existing business logic should be maintained

**QA**
- Stylistic changes can be expected
- no business logic should be changed
- Review page in various forms:
    - create program
    - edit program
    - read-only program 
- BC Public vs BC Private vs Other institution types
- Field formatting
    - CIP code format
    - SABC program code format
- Business logic (beyond being required fields):
    - Field of study code calculation
    - No duplicate SABC program code 
    - Online-only question
    - 100% on site question
    - Entrance requirements
    - ESL <20%
    - Program partnerships
    - WIL approved
    - WIL eligibility requirements
    - Field trip eligibility requirements
    - international exchange eligibility requirements
    - Aviation
    - credential types
    - instructional hours
- Pending or Auto-Approve conditions still in place
- Duplicate SABC program code

Joshua-Lakusta:
2 points remaining for 4.3 ui updates

PR #6599, opened 2026-09-10
#6343 - Convert program to Vuetify - Backend

dheepak-aot:
# Convert program to Vuetify - Backend

### Class validators
- [x] Most of the API payload validation is now handled through class validators.
- [x] Similar to offerings, the class validators use the `ValidateIf` and `AllowIf` decorators to ensure the sanity of the payload when the existence of a value is subject to another value in the payload. for e.g. other regulatory body can have value only if the regulatory body is `other`.

### API validations post class validation
- [x] The lookup values from payload are validated by the API.
- [x] The entrance requirements are validated to not have any other values when `none of the above entrance requirement` is selected.
- [x] The values `isBCPrivate` and `isBCPublic` sent by the client is validated at server side to match the institution type.

## Form.IO
- [x] The checkbox component(select boxes)  values returned by form.io and vuetify are different. Form.io returns an object model e.g. `{key1: boolean, key2:boolean}` where vuetify values are array of string e.g. `["key1", "key2"]`. Hence considering the format of vuetify for long-term, a calculated value was created which has all converted checkbox values to be mapped to API payload. This converted format is now sent to API and processed.
**Note:** This is only for a short term, until front end is converted to Vuetify. 

## DB Migrations
- [x] The lookup values for program credential types are missed in the  [previous PR](https://github.com/bcgov/SIMS/pull/6590). Hence added here. Once the code is merged and deployed to DEV the rollback can be done to ensure all the lookup values are up to date.

## E2E Tests

- [x]  Extended the E2E Tests for create program API to validate different areas, like program status, class validator errors and other API validation errors. 

```spec
EducationProgramInstitutionsController(e2e)-createEducationProgram
    √ Should create an education program when valid data is passed. (134 ms)
    √ Should create an education program in status Pending when program is offered jointly or in partnership. (50 ms)
    √ Should create an education program in status Pending when program has more than 20 percent ESL Content. (32 ms)
    √ Should create an education program in status Pending when program has no entrance requirements. (33 ms)
    √ Should create an education program in status Pending when program belongs to BC Private institution and delivered only online. (110 ms)
    √ Should create an education program pending notification when the program is pending. (33 ms)
    √ Should not create an education program when user is read-only. (75 ms)
    √ Should throw duplicate SABC code for education program when there is already an active education program with the same SABC code. (30 ms)
    √ Should create an education program when SABC code passed exists for an expired education program. (38 ms)
    √ Should create an education program when SABC code passed exists for an education program flagged as inactive. (36 ms)
    √ Should throw bad request error when CIP code format is invalid. (14 ms)
    √ Should throw bad request error when program declaration is not true. (12 ms)
    √ Should throw bad request error when international exchange program eligibility is missing for an international program. (13 ms)
    √ Should throw bad request error when aviation credential type is provided for a non-aviation program. (20 ms)
    √ Should throw bad request error when none of the entrance requirements is provided along with other entrance requirements. (18 ms)
    √ Should throw bad request error when invalid lookup values are provided. (17 ms)
    √ Should unprocessable entity error when BC Private and BC Public status does not match with the institution type. (18 ms)
```

### Edit Program
image
image


sources/packages/backend/apps/api/src/route-controllers/education-program/_tests_/e2e/education-program.institutions.controller.createEducationProgram.e2e-spec.ts
```diff
@@ -366,6 +426,115 @@ describe("EducationProgramInstitutionsController(e2e)-createEducationProgram", (
...
+    // Arrange
+    const payload = { ...getPayload(), isBCPrivate: false, isBCPublic: true };
+    // BC Private institution user.
+    const institutionUserToken = await getInstitutionToken(
+      InstitutionTokenTypes.CollegeCAdminLegalSigningUser,
+    );
+
+    // Act/Assert
+    await request(app.getHttpServer())
+      .post(ENDPOINT)
```
andrewsignori-aot: Not a blocker, minor, some other tests are using a `getEndpoint` method even for constant URIs for consistency.

sources/packages/web/src/services/http/dto/EducationProgram.dto.ts
```diff
@@ -81,6 +82,49 @@ export interface EducationProgramsSummaryAPIOutDTO {
   isExpired: boolean;
 }
 
+export interface ProgramFormModel {
```
andrewsignori-aot, line 85: Should it not be in the contracts instead of DTO?

andrewsignori-aot (approved):
Great work, and thanks for great effort to have it done. Looks good 👍

sh16011993 (approved):
@dheepak-aot I went through the PR description. Great work 👍

Issue #6614, opened 2026-09-16
HOTFIX: Students unable to submit edited applications - staff unable to approve change requests


Joshua-Lakusta:
## Description
Students and PSFS staff are not able to submit or approve an application modification / change request. Observed in PROD and TEST. 

<img width="503" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNDBERHc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ee05ba2ce485c0dbb169ece1d9a1325c48565936/image.png" alt="image.png" />

## Expected Behaviour

No errors should be presented and the request should be submittable. 

## Logs

```text
Student editing their application:

[Nest] 18 - [TIMESTAMP] LOG [AccessLoggerMiddleware] Request - PATCH /api/students/application/[APPLICATION_ID]/submit From [IP_ADDRESS] | User GUID: [USER_GUID] | User Agent: [USER_AGENT]
query failed: UPDATE "sims"."supporting_users" SET "application_id" = $1, "updated_at" = CURRENT_TIMESTAMP WHERE "id" = $2 -- PARAMETERS: [null,[SUPPORTING_USER_ID]]
error: error: null value in column "application_id" of relation "supporting_users" violates not-null constraint
[Nest] 18 - [TIMESTAMP] ERROR [AppAllExceptionsFilter] Unhandled exception, request path: /api/students/application/[APPLICATION_ID]/submit
{
[...]

PR #6618, opened 2026-09-17
#6614 - Fix to the application edit and change request

dheepak-aot:
## Root cause
During application edit and change request approval actions, the application is first retrieved through [ApplicationService.getApplicationById()](https://github.com/bcgov/SIMS/blob/main/sources/packages/backend/apps/api/src/services/application/application.service.ts) and the retrieved object is persisted later using typeorm save after updating some properties w.r.t creating new application version. 
Since the retrieved object has supporting user(s) without id being fetched, at the typeorm save fails trying to update the supporting user(s) present in the object but fails to find the object as the id is absent.

## Solution
Supporting user id was included in the fetch sql to fix the issue. 
image

### Edit application

image

image

image

## Change Request approval
image

image


andrewsignori-aot (approved):
Looks good, thanks for taking care of it 👍

weskubo-cgi (approved):

Issue #6516, opened 2026-08-25
Remove the offering column added for rollback purpose


dheepak-aot:
**Describe the task**
During the implementation of the hotfix ticket #6161 a column was added to offerings table to copy the study_breaks calculated values to perform rollback. After few sprints from `Hotfix v4.0.1` this column must be remove as a part of cleanup.

**Acceptance Criteria**
- [ ] Remove the column `study_breaks_before_update` from sims.education_programs_offerings and sims.education_programs_offerings_history.

PR #6619, opened 2026-09-17
#6516 - Remove offering column added for rollback purpose

sh16011993:

- Column `study_breaks_before_update` was removed from the `education_programs_offerings` and the `education_programs_offerings_history` table.


dheepak-aot (approved):
Looks good 👍

weskubo-cgi (approved):
Thanks for taking care of this.

Issue #6617, opened 2026-09-17
Remove the student_disability_profile_disabilities column added for rollback purpose


weskubo-cgi:
**Describe the task**
During the implementation of the #6525 a column was added to the student_disability_profile_disabilities table to copy the impairments to perform rollback. After 4.2 is deployed, this column must be removed as a part of cleanup.

**Acceptance Criteria**
- [ ] Remove the column `impairments_before_update` from sims.student_disability_profile_disabilities.

PR #6620, opened 2026-09-17
#6617 - Remove the impairments before update column

sh16011993:
### As a part of the PR, the following was completed:

- Column `impairments_before_update` was removed from the `student_disability_profile_disabilities` table.


dheepak-aot (approved):
Looks good 👍

weskubo-cgi (approved):
Thanks for taking care of this.

